PR #3, opened 2023-04-18
Fix logging URLs in multihost_job

gobbleturk:
1. Fix logging URLs to the public console.cloud.google.com instead of internal URL
2. Improve documentation for multihost_job:
      * Emphasize new TPUs are created
      * Tell user that they need to include installation command
      * Include cloud logging information


rwitten (approved):
Nice work!

PR #12, opened 2023-04-24
Checkpointing XLML Tests

priyanka-ganesha:
* Checkpointing tests for single host, multi host, and multi slice configurations.
* Custom GCS Bucket parameterization

### DIFF
--- a/end_to_end/eval_assert.py
+++ b/end_to_end/eval_assert.py
@@ -24,6 +24,8 @@ def read(metrics_file, target):
 def test_tflops(metrics_file, target, threshold):
   """Asserts over tflops values"""
   avg_tflops = read(metrics_file, target)
+  # Checks for acceptable performance by asserting that average tflops value
+  # is greater than or equal to threshold
   assert avg_tflops >= threshold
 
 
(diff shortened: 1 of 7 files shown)


end_to_end/eval_assert.py
```diff
@@ -20,13 +21,32 @@ def read(metrics_file, target):
...
+def test_checkpointing(metrics_file, target):
+  """Asserts over loss values from loaded checkpoint"""
+  metrics_file_saved = 'saved_' + metrics_file
+  metrics_file_restored = 'restored_' + metrics_file
+
+  with open(metrics_file_saved, 'r', encoding='utf8') as saved,\
+    open(metrics_file_restored, 'r', encoding='utf8') as restored:
+    saved_loss = json.loads(saved.readlines()[-1])[target]
+    restored_loss = json.loads(restored.readlines()[0])[target]
+    assert isclose(saved_loss, restored_loss, rel_tol=0.5)
```
gobbleturk, line 39: Can we include a comment what this assert is? Something like we are comparing that the first step of restore matches the last step of save (or maybe is one step difference) - for the purpose of testing that save/restoring works, not actually a performance assert.

end_to_end/eval_assert.py
```diff
@@ -20,13 +21,32 @@ def read(metrics_file, target):
...
+
+def test_checkpointing(metrics_file, target):
+  """Asserts over loss values from loaded checkpoint"""
+  metrics_file_saved = 'saved_' + metrics_file
+  metrics_file_restored = 'restored_' + metrics_file
+
+  with open(metrics_file_saved, 'r', encoding='utf8') as saved,\
+    open(metrics_file_restored, 'r', encoding='utf8') as restored:
+    saved_loss = json.loads(saved.readlines()[-1])[target]
+    restored_loss = json.loads(restored.readlines()[0])[target]
```
gobbleturk, line 38: Why does the saved_loss grab index [-1] but the restored_loss grabs index [0]? Should both grab index [-1]?

end_to_end/eval_assert.py
gobbleturk, line 39: Can we decrease the rel_tol here? I assume the two losses are either identical or just one step different, I would expect rel_tol at 0.1 or smaller should also work?

end_to_end/eval_assert.py
```diff
@@ -20,13 +21,32 @@ def read(metrics_file, target):
...
 
   _, metrics_file, threshold, target = argv
 
-  avg_val = read(metrics_file, target)
-
-  assert avg_val >= float(threshold)
+  if target == 'perf/per_device_tflops_per_sec':
+    test_tflops(metrics_file, target, float(threshold))
+  elif target == 'learning/loss':
+    test_checkpointing(metrics_file, target)
```
gobbleturk, line 49: I think future tests may use/re-use different targets. Can we instead add a new arg named something like "test_scenario" (example inputs "checkpointing", "performance", etc) to the main of eval_assert which tells us which tests to run?

gobbleturk (changes requested):

gobbleturk (approved):

PR #14, opened 2023-04-24
Gradient cutting by local comparisons

gobbleturk:
This helps with unstable training (large training loss spikes), however it is by no means a complete fix. There are still several spikes seen with this change, but we have seen better loss over time, with fewer spikes that are recovered from faster.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -109,3 +109,6 @@ prompt: "I love to "
 enable_profiler: False
 enable_checkpointing: True
 
+grad_window: 100
+grad_max_growth: 1.2
+
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -116,5 +116,6 @@ adam_eps: 1.e-8 # A small constant applied to denominator outside of the square 
 adam_eps_root: 0. # A small constant applied to denominator inside the square root.
 
 # Gradient clipping parameters
-grad_clip_moving_decay_weight: 0.9 # Gradient norms moving average decay weight.
-grad_clip_max_growth_factor: 1.1 # Gradient is clipped by moving average times this factor.
+rolling_gradient_clipping_start_step: 100 # Step to start clipping the gradients.
+rolling_gradient_clipping_factor: 1.1 # Gradient is clipped by moving average times this factor.
+rolling_gradient_decay_weight: 0.9 # Gradient norms' moving average decay weight.
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -118,4 +118,4 @@ adam_eps_root: 0. # A small constant applied to denominator inside the square ro
 # Gradient clipping parameters
 rolling_gradient_clipping_start_step: 100 # Step to start clipping the gradients.
 rolling_gradient_clipping_factor: 1.1 # Gradient is clipped by rolling average times this factor.
-rolling_gradient_decay_weight: 0.9 # Gradient norm's exponentially weighted rolling decay weight.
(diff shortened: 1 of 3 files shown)


MaxText/configs/base.yml
```diff
@@ -109,3 +109,9 @@ prompt: "I love to "
 enable_profiler: False
 enable_checkpointing: True
 
+grad_cut_factor: 1.2 # Gradient capped by mean of previous gradients times this factor.
```
rwitten, line 112: `grad_cut` is a bad name unfortunately. My best is `rolling_gradient_clipping`.

MaxText/configs/base.yml
```diff
@@ -109,3 +109,9 @@ prompt: "I love to "
...
+grad_cut_window_len: 100 # The window length of running gradients to compare to.
+grad_cut_start_step: 100 # Gradient starts being cut at this step.
```
rwitten, line 114: Probably safe to assume these are the same for now?

MaxText/train.py
```diff
@@ -203,13 +204,18 @@ def loss_fn(params):
 
   grad_fn = jax.value_and_grad(loss_fn, has_aux=True)
   (loss, intermediate_outputs), grads = grad_fn(state.params)
+  pre_cut_grad_norm = jnp.array(max_utils.l2norm_pytree(grads))
+  max_grad_norm = jnp.array(config.grad_cut_factor * grad_cut)
+  grad_scale_factor = jnp.minimum(max_grad_norm, pre_cut_grad_norm) / pre_cut_grad_norm
+  grads = jax.tree_map(lambda g: g * grad_scale_factor, grads)
```
rwitten, line 210: I think it is best to call this `scaled_grads`.

MaxText/train.py
```diff
@@ -203,13 +204,18 @@ def loss_fn(params):
...
+  grad_norm = max_utils.l2norm_pytree(grads)
   new_state = state.apply_gradients(grads=grads)
-  metrics = {'scalar': {'learning/loss': loss, 'learning/grad_norm' : max_utils.l2norm_pytree(grads),
+  metrics = {'scalar': {'learning/loss': loss, 'learning/grad_norm' : grad_norm,
```
rwitten, line 213: as metrics we probably want `raw grad norm`, `applied grad norm` and `rolling gradient clipping threshold`

MaxText/configs/base.yml
rwitten: So we'd call the various quantities lower in the code `rolling_gradient_clipping_window_len`, `rolling_gradient_clipping_start_step`, `rolling_gradient_clipping_scale_factor`, `rolling_gradient_clipping_current_threshold`, etc.

MaxText/train.py
```diff
@@ -318,23 +324,31 @@ def train_loop(config, state=None):
...
   example_batch = None
   last_step_completion = datetime.datetime.now()
+  grad_norms = deque(maxlen=config.grad_cut_window_len)
+  LARGE_GRADIENT_NORM = 1.0 * 10**6 # Used when gradient cutting is not desired.
 
   local_metrics_file = open(config.metrics_file, 'a', encoding="utf8") if config.metrics_file else None
 
   for step in np.arange(get_first_step(state), config.steps):
     example_batch = load_next_batch(train_iter, example_batch, config)
+    if step < config.grad_cut_start_step or not grad_norms:
```
rwitten, line 343: put this into its own function! and don't define `LARGE_GRADIENT_NORM`, I think `float('inf')` has the right semantics here.

rwitten (commented):
Excited to see the results for this.

PR #17, opened 2023-04-27
Add configurable Adam parameters, set beta2 to 0.95

gobbleturk:
This change:
* Allows users to overwrite default Adam optimizer parameters
* Sets the Maxtext default beta2 to a much stabler (less training loss spikes) value of 0.95 instead of default 0.999

Before this change 17B model with default beta2=0.999
Pre-beta-instability

After this change 17B model with beta2=0.95
post-beta-stability

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -109,7 +109,8 @@ prompt: "I love to "
 enable_profiler: False
 enable_checkpointing: True
 
+# Adam optimizer parameters
 adam_b1: 0.9 # Exponential decay rate to track the first moment of past gradients.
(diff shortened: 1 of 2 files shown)


rwitten (approved):
nice!

gobbleturk:
This change:

* Allows users to overwrite default Adam optimizer parameters
* Sets the Maxtext default beta2 to a much stabler (less training loss spikes) value of 0.95 instead of default 0.999

Before this change 17B model with default beta2=0.999
Pre-beta-instability

After this change 17B model with beta2=0.95
post-beta-stability

PR #23, opened 2023-05-01
Update readme

raymondzouu:


gobbleturk:
Are we also planning to include an 8xv4-128.sh file that repros your performance?

raymondzouu:
That's a good idea, added it.

gobbleturk (approved):

PR #24, opened 2023-05-01
replace `absl` with `argparse` in `multihost_job.py`

tonyjohnchen:
Replaced `absl` with `argparse`, so `multihost_runner.py` will be fully `dependency free`

Local test:
image


gobbleturk (approved):
The PR name references `multihost_runner.py` but it should be `multihost_job.py`

PR #26, opened 2023-05-03
Tensorboard instructions frequency

priyanka-ganesha:
Print tensorboard instructions every log_period steps, default 10.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -94,15 +94,15 @@ def write_metrics(writer, metrics, step, config):
       for metric_name in metrics.get("scalars",[]):
         writer.add_scalars(metric_name, metrics["scalars"][metric_name], step)
 
-    full_log = step % 5 == 0
+    full_log = step % config.log_period == 0
 
     max_logging.log(f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
           f"TFLOP/s: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
           f"loss: {metrics['scalar']['learning/loss']:.3f}")
 
-    if full_log and config.enable_profiler:
+    if full_log:
       max_logging.log(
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -94,15 +94,15 @@ def write_metrics(writer, metrics, step, config):
...
 
     max_logging.log(f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
           f"TFLOP/s: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
           f"loss: {metrics['scalar']['learning/loss']:.3f}")
 
-    if full_log:
+    if full_log and config.enable_profiler:
       max_logging.log(
-          f"To see full metrics 'tensorboard --logdir={config.tensorboard_dir}'"
+          f"PROFILING: To see full metrics 'tensorboard --logdir={config.tensorboard_dir}'"
```
rwitten, line 105: Tensorboard produces data even with profiling off! It shouldn't be tied to the profiler.

MaxText/train.py
```diff
@@ -94,15 +94,15 @@ def write_metrics(writer, metrics, step, config):
       for metric_name in metrics.get("scalars",[]):
         writer.add_scalars(metric_name, metrics["scalars"][metric_name], step)
 
-    full_log = step % config.log_period == 0
+    full_log = step % 5 == 0
```
rwitten, line 97: I don't love adding more constants scatter around the codebase and not sure I fully understand this particular one. Does the `log_period` only influence this or are there other consumers? If there aren't any other consumers, maybe we just set the default `log_period` to say 10?

rwitten (changes requested):

gobbleturk (approved):
As we're only changing the frequency of the tensorboard command print, this is fine

rwitten (approved):

PR #29, opened 2023-05-05
Reinstate internal-ip and remove gcloud auth login

gobbleturk:
* Reinstate explicit internal-ip flag, the previous auto-determination only works using "gcloud auth login" with an @google.com address
* Various permission and API improvements in the readme


rwitten (approved):
Looks good!

PR #31, opened 2023-05-09
Decode xlml

gobbleturk:
XLML test for decoder.py, which tests that decode.py works via counting number of decoded tokens

Also: 
* Fixes decoder.py by removing the config.use_pjrt
* Refactors write_metrics_locally to max_utils so it can be shared by both train.py and decoder.py
* Rename eval_assert.test_flops to eval_assert.assert_metric_average to indicate it be used more generally such as this test

### DIFF
--- a/end_to_end/test_decode_v4_8_one_slice.sh
+++ b/end_to_end/test_decode_v4_8_one_slice.sh
@@ -1,24 +1,22 @@
 #!/bin/bash
 set -e
 
-USER=${1}
-NUM_TOKEN_THRESHOLD=${2}
-OUTPUT_PATH=${3}
-DATASET_PATH=${4}
-# Run name is optional 5th input - our daily XLML tests will use one.
+NUM_TOKEN_THRESHOLD=${1}
+OUTPUT_PATH=${2}
+# Run name is optional 3rd input - our daily XLML tests will use one.
 
 
-if [ -z ${5} ]
+if [ -z ${3} ]
 then 
     RUN_NAME=${USER}_$(date +%Y-%m-%d-%H-%M-%S)
 else
-    RUN_NAME=${5}_$(date +%Y-%m-%d-%H)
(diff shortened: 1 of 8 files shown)


end_to_end/test_tflops_v4_8_one_slice.sh
```diff
@@ -17,7 +17,7 @@ fi
 #Setup and Train
 bash setup.sh
 python3 MaxText/train.py MaxText/configs/base.yml run_name=$RUN_NAME\
-    steps=150 reuse_example_batch=1 remat_policy='full' enable_checkpointing=False metrics_file='metrics.txt'\
+    steps=15 reuse_example_batch=1 remat_policy='full' enable_checkpointing=False metrics_file='metrics.txt'\
```
priyanka-ganesha, line 20: I think you may have missed changing this back to 150 after testing

end_to_end/test_decode_v4_8_one_slice.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+
+NUM_TOKEN_THRESHOLD=${1}
+OUTPUT_PATH=${2}
```
priyanka-ganesha, line 5: We may need DATASET_PATH as well since base.yml has a default value for it

priyanka-ganesha (approved):

end_to_end/test_tflops_v4_8_one_slice.sh
gobbleturk: Fixed.

end_to_end/test_decode_v4_8_one_slice.sh
gobbleturk: Good catch! I think decode.py doesn't use the training examples - but it does rely on the vocab/tokenizer so we do need to pass in the dataset path.

PR #32, opened 2023-05-11
Added vocab creation test

priyanka-ganesha:
Checks for the existence of a vocab object in multislice configuration.

### DIFF
--- a/end_to_end/eval_assert.py
+++ b/end_to_end/eval_assert.py
@@ -2,6 +2,7 @@
 from absl import app
 from typing import Sequence
 from math import isclose
+from google.cloud import storage
 import json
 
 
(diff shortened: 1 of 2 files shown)


gobbleturk (approved):
This change LGTM. However after writing my own eval_assert test and looking at this one, I have a better idea how to generalize the inputs to eval_assert--

I think we should refactor the

`_, metrics_file, threshold, target, test_scenario = argv`

to 

`_, test_scenario, test_vars* = argv`

and then something like:
```
if test_scenario == 'checkpoint_save_restore':
   test_vocab_creation(test_vars*)
elif test_scenario == 'vocab_creation':
   test_vocab_creation(test_vars*)
```

This way each test only passes in its relevant variables. Wdyt? If you like it we could submit it as a separate PR followup soon

PR #34, opened 2023-05-12
Allow user to install jax-head,jaxlib-nightly,libtpu-nightly or customized libtpu

tonyjohnchen:
Allow user to install jax-head,jaxlib-nightly,libtpu-nightly.
Allow user to install customized libtpu from GCS bucket.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -18,8 +18,12 @@
 pip3 show jax && pip3 uninstall -y jax 
 pip3 show jaxlib && pip3 uninstall -y jaxlib
 
-# Install dependencies from requirements.txt
-pip3 install -r requirements.txt -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+libtpu_path="/lib/libtpu.so"
+
+# Delete libtpu if it exists
+if [ -e "$libtpu_path" ]; then
+    sudo rm "$libtpu_path"
+fi
 
 # Set environment variables
 for ARGUMENT in "$@"; do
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -14,4 +14,31 @@
...
 # limitations under the License.
 
-pip3 install -r requirements.txt -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
\ No newline at end of file
+# Uninstall existing JAX and JAXlib
+pip3 show jax && pip3 uninstall -y jax 
+pip3 show jaxlib && pip3 uninstall -y jaxlib
+
+# Install dependencies from requirements.txt
+pip3 install -r requirements.txt -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
```
rwitten, line 22: Not sure I understand this. We install `requirements.txt` -- but doesn't taht still contain `jax[tpu]`? Shouldn't we remove `jax[tpu]` from requirements.txt and only install it if necessary?

setup.sh
```diff
@@ -14,4 +14,31 @@
...
+done
+
+if [[ $MODE == "nightly" ]]; then
+    echo "Installing jax-head, jaxlib-nightly, and libtpu-nightly"
+    # Install JAX from GitHub head
+    pip3 install git+https://github.com/google/jax
+    # Install jaxlib-nightly
+    pip3 install --pre -U jaxlib -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
+    # Install libtpu-nightly
+    pip3 install libtpu-nightly -f https://storage.googleapis.com/jax-releases/libtpu_releases.html -U --pre
```
rwitten, line 37: We shouldn't install a `libtpu-nightly` if we're given a gcs path, that seems confusing.

setup.sh
```diff
@@ -14,4 +14,31 @@
...
+fi
+
+if [[ -n "$LIBTPU_GCS_PATH" ]]; then
+    sudo pip3 install -U crcmod
+    echo "Copying libtpu.so from $LIBTPU_GCS_PATH to /lib/libtpu.so"
+    sudo gsutil cp "$LIBTPU_GCS_PATH" /lib/libtpu.so
+fi
```
rwitten, line 44: Having now read to this point, I recommend the following order.

(1) Uninstall jax, jaxlib and libtpu and delete `/lib/libtpu.so`
(2) Install jax, jaxlib and libtpu
- if `stable`, via jax[tpu]. Assert the user doesn't pass in a libtpu separately.
- if `nightly`, install jax and jaxlib as you do. Install libtpu from either GCS or libtpu nightly as appropriate. 
(3) install the rest of requirements.

rwitten (changes requested):
Thanks for pulling this together, I have some ideas

setup.sh
tonyjohnchen: Just made the change based on your comments.

setup.sh
```diff
@@ -14,4 +14,50 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-pip3 install -r requirements.txt -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
\ No newline at end of file
+# Uninstall existing JAX and JAXlib
+pip3 show jax && pip3 uninstall -y jax 
+pip3 show jaxlib && pip3 uninstall -y jaxlib
+
```
rwitten, line 20: shouldn't we uninstall libtpu?

gobbleturk (approved):

setup.sh
```diff
@@ -14,4 +14,50 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
rwitten, line 16: Can we `set -e` this script? So if there is an error in the code, it propagates to the top level error status?

rwitten (changes requested):
Two tiny things!

setup.sh
tonyjohnchen: Done

setup.sh
tonyjohnchen: Done

rwitten (approved):

PR #35, opened 2023-05-15
Add mode 'head' and 'libtpu-only' mode for installation script

tonyjohnchen:
'head' mode will install jax, jaxlib, and custom libtpu from head.
'libtpu-only' mode will only install custom libtpu without re-install jax,jaxlib from head. (Noted: build jaxlib from head will take about 10 mins)

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -128,7 +128,7 @@ elif [[ $MODE == "head" ]]; then
     python3 build/build.py --enable_tpu --bazel_options="--override_repository=xla=$HOME/xla"
     pip3 install dist/jaxlib-*-cp*-manylinux2014_x86_64.whl --force-reinstall --no-deps
 else
-    echo -e "\n\nError: You can only set MODE to [stable,nightly,head].\n\n"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -17,32 +17,72 @@
 # Enable "exit immediately if any command fails" option
```
rwitten, line 17: Can you put a comment at the top describing how to use the module? Looks great otherwise!

rwitten (changes requested):

setup.sh
tonyjohnchen: Done.

I will add a README with more details for internal developer next.

rwitten (changes requested):
Just want to give clearer feedback for how to write the documenation.

setup.sh
```diff
@@ -14,35 +14,96 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# How to use:
```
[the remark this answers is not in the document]
tonyjohnchen: Makes sense!
Done.

rwitten (approved):

PR #43, opened 2023-05-25
-1 Parallelism Logic

priyanka-ganesha:
* Changes default of dcn_data_parallelism and ici_fsdp_parallelism to unspecified values (-1)
* Additional asserts

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -67,6 +67,11 @@ def create_device_mesh(config):
   """Creates a device mesh with each slice in its own data parallel group. If there is only one slice, uses two replicas """
   devices = jax.devices()
   num_devices = len(devices)
+  try:
+    num_slices = 1 + max([d.slice_index for d in devices])
+  except:
+    num_slices = 1
+  num_devices_per_slice = num_devices//num_slices
   max_logging.log(f"Devices: {devices} (num_devices: {num_devices})")
   assert len(devices) > 1, "You must have at least two devices"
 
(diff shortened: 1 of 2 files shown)


rwitten (changes requested):
Little nits! Excited to land this.

MaxText/max_utils.py
```diff
@@ -75,8 +96,15 @@ def create_device_mesh(config):
   dcn_parallelism = [config.dcn_data_parallelism, config.dcn_fsdp_parallelism, config.dcn_tensor_parallelism]
   ici_parallelism = [config.ici_data_parallelism, config.ici_fsdp_parallelism, config.ici_tensor_parallelism]
 
-  assert np.product(dcn_parallelism) * np.product(ici_parallelism) == num_devices, f"Number of devices {num_devices} \
-        does not match the product of the parallelism {np.product(dcn_parallelism) * np.product(ici_parallelism)}"
+  # Validate parallelisms
+  dcn_parallelism = validate_unspecified(dcn_parallelism, num_slices, 'DCN')
+  ici_parallelism = validate_unspecified(ici_parallelism, num_devices_per_slice, 'ICI')
+
+  assert np.product(dcn_parallelism) == num_slices, f"Number of slices {num_slices} does not match the product\
```
rwitten, line 103: These asserts really should be inside of the function that generated the data!

MaxText/max_utils.py
```diff
@@ -63,10 +64,30 @@ def write_metrics_locally(metrics, step, last_step, file):
   if step == last_step:
     file.close()
 
+def validate_unspecified(parallelism_vals, target_product, parallelism_type):
```
rwitten, line 67: Nit: probably not the best function name given a lack of context.

Perhaps `fill_unspecified_mesh_axes`.

There is no need to mention validate typically -- it is an expectation that functions validate their data! (This can be different if for example we're operating in an environment where for performance reasons we'd be not validating for correctness. But this codebase always validates for correctness as it goes.)

MaxText/max_utils.py
rwitten, line 103: These asserts really should be inside of the function that generated the data!

rwitten (changes requested):
Last nits

rwitten (approved):

PR #44, opened 2023-05-25
Update import for `partition_spec` and `Mesh`

tonyjohnchen:
Update import since `partition_spec` and `Mesh` is no longer in `experimental` folder.


rwitten (approved):
Thanks for taking care of this!

PR #48, opened 2023-06-02
More accurate TFLOP math. Now counts attention!

rwitten:
Please double check the math re: attention!


gobbleturk:
The expression LGTM!

I'm assuming the layers.py variables q_length and kv_length are both max_target_length, and
the factor of 12 comes from 2 * [standard 6] since there is both the QK matmul and  the attn_weights * V matmul

gobbleturk (approved):

gobbleturk:
Nit: After asking Lukasz which matmuls we should speed up with AQT I realize it is the same set as those you are calling "matmul_tflops" - I think a better name may be "learnable_weights_matmul_tflops", to distinguish from the attention matmuls

PR #50, opened 2023-06-06
Removed bash setup from e2e tests

priyanka-ganesha:
Taken care of in XLML maxtext mixin


rwitten (approved):
Looks great. Please don't merge until the XLML tests are also approved.

PR #51, opened 2023-06-07
CQR via gcloud API in multihost_job

gobbleturk:
* Use gcloud instead of curl for CQR request
* Change inputs to accept CQR_EXTRA_ARGS, which are verbatim CQR args

Tested by running multihost_job with `--CQR_EXTRA_ARGS=--best-effort`


rwitten (approved):
OMG does look a lot better!

PR #52, opened 2023-06-07
Delete the QR using --force in multihost_job

gobbleturk:
* Use gcloud delete qr --force to delete TPUs and QR at end of command
* Change default runtime-version to tpu-ubuntu2204-base, which has a recent cloud SDK version (the --force flag was only recently added), is actively maintained/updated, and is the version we generally recommend for cloud users. 
* Remove endpoint input (it was unusable due to the recent curl -> gcloud CQR change). It will take some effort to add back, and isn't very valuable at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,26 +144,11 @@ either be a TPUVM or not, but it cannot be one of the workers. If your runner ma
 
 6. Clean up TPUs and QR when finished.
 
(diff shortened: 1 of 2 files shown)


rwitten (approved):
Maybe include a message that this worker is waiting 10 minutes before doing the tear down to allow other workers to gracefully tear down?

PR #53, opened 2023-06-07
Add configuration to collect stack traces

SurbhiJainUSC:
This PR is to add the configuration to collect python stack traces when a fault such as Segmentation fault, Floating-point exception, Illegal operation exception occurs in the program. Additionally, it will also periodically collect stack traces to help debug when a program running on Cloud TPU is stuck or hung somewhere.

PyPI package: https://pypi.org/project/cloud-tpu-diagnostics/

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -119,5 +119,5 @@ adam_eps_root: 0. # A small constant applied to denominator inside the square ro
 
 # Stack trace parameters
 collect_stack_trace: True
-stack_trace_to_cloud: False  # display stack trace on console
-stack_trace_interval_seconds: 600  # 10 minutes
+stack_trace_to_cloud: False  # Uploads to cloud logging if True, else to the console if False.
+stack_trace_interval_seconds: 600  # Stack trace collection frequency in seconds.
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -116,3 +116,8 @@ adam_b1: 0.9 # Exponential decay rate to track the first moment of past gradient
 adam_b2: 0.95 # Exponential decay rate to track the second moment of past gradients.
 adam_eps: 1.e-8 # A small constant applied to denominator outside of the square root.
 adam_eps_root: 0. # A small constant applied to denominator inside the square root.
+
+# Stack trace parameters
+collect_stack_trace: True
+stack_trace_to_cloud: False  # display stack trace on console
+stack_trace_interval_seconds: 600  # 10 minutes
--- a/README.md
+++ b/README.md
@@ -46,6 +46,14 @@ bash download_dataset.sh {GCS_PROJECT} {GCS_BUCKET_NAME}
 
 To run maxtext the TPUVMs must have permission to read the gcs bucket. These permissions are granted by service account roles, such as the `STORAGE ADMIN` role. 
 
+## Getting Started: Configure Diagnostic Settings
+Here is the related PyPI package: https://pypi.org/project/cloud-tpu-diagnostics.
+### Debugging
+The following configurations will help to debug a fault or when a program is stuck or hung somewhere by collecting stack traces. Change the parameter values accordingly in `MaxText/configs/base.yml`:
+1. Set `collect_stack_trace: True` to enable collection of stack traces on faults or when the program is hung. To disable this, set `collect_stack_trace: False`.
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -46,6 +46,14 @@ bash download_dataset.sh {GCS_PROJECT} {GCS_BUCKET_NAME}
 
 To run maxtext the TPUVMs must have permission to read the gcs bucket. These permissions are granted by service account roles, such as the `STORAGE ADMIN` role. 
 
+## Getting Started: Configure Diagnostic Settings
+Here is the related PyPI package: https://pypi.org/project/cloud-tpu-diagnostics.
+### Debugging
+The following configurations will help to debug a fault or when a program is stuck or hung somewhere by collecting stack traces. Change the parameter values accordingly in `MaxText/configs/base.yml`:
+1. Set `collect_stack_trace: True` to enable collection of stack traces on faults or when the program is hung. To disable this, set `collect_stack_trace: False`.
+2. Set `stack_trace_to_cloud: False` to display stack traces on console. `stack_trace_to_cloud: True` will create a temporary file in `/tmp/debugging` in the TPUs to store the stack traces. There is an agent running on TPU VMs that will periodically upload the traces from the temporary directory to cloud logging in the gcp project.
```
gobbleturk, line 54: The /tmp/debugging logging is cool, I didn't realize this. For my own understanding, how is this logging achieved? Are you implementing an agent  in cloud-tpu-diagnostics to /tmp/debugging or is there a pre-existing agent that does this? Is it an ops agent?

MaxText/configs/base.yml
```diff
@@ -116,3 +116,8 @@ adam_b1: 0.9 # Exponential decay rate to track the first moment of past gradient
 adam_b2: 0.95 # Exponential decay rate to track the second moment of past gradients.
 adam_eps: 1.e-8 # A small constant applied to denominator outside of the square root.
 adam_eps_root: 0. # A small constant applied to denominator inside the square root.
+
+# Stack trace parameters
+collect_stack_trace: True
+stack_trace_to_cloud: False  # display stack trace on console
```
gobbleturk, line 122: Nit: I think these comments should describe the variable, not the default value. I.e. for stack_trace_to_cloud I think "Uploads to cloud logging if true, else to the console if false"

stack_trace_interval_seconds is pretty self-explanatory (no need for comment) or could be "Stack trace collection frequency"

README.md
```diff
@@ -46,6 +46,14 @@ bash download_dataset.sh {GCS_PROJECT} {GCS_BUCKET_NAME}
 
 To run maxtext the TPUVMs must have permission to read the gcs bucket. These permissions are granted by service account roles, such as the `STORAGE ADMIN` role. 
 
+## Getting Started: Configure Diagnostic Settings
```
gobbleturk, line 49: Thank you for adding this to the Readme!!

gobbleturk (changes requested):
This looks great, thanks for adding this to Maxtext!

I added a few small comments, the only thing I think we should change is `pip3 install cloud-tpu-diagnostics` should get moved to [requirements.txt](https://github.com/rwitten/MaxText/blob/main/requirements.txt)

README.md
SurbhiJainUSC: We have a module in `cloud-tpu-diagnostics` that collects stack traces and dump it to a temp file in TPU VMs. Then, there is a runtime monitor agent running on VMs that periodically uploads the traces from VMs to cloud logging.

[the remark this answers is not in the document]
SurbhiJainUSC: `stack_trace_to_cloud=True` - create a temp file in `/tmp/debugging` directory and dump the traces to it. So, you can view the traces in VM itself. Additionally, runtime monitor agent will upload the traces from temp file to a dedicated log name in cloud logging. And, that log name will only have traces and no other information.
`stack_trace_to_cloud=False` - stack traces are displayed on console and moved to cloud logging but it will be present inside stdout log name along with other standard output.

SurbhiJainUSC:
> 

Thank you for the review. I have addressed the comments.

README.md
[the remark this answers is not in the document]
gobbleturk: Oh thank you for this explanation this is much clearer. I see the value in having a log name with only the traces - I think you should include that in the comment for stack_trace_to_cloud - that there is a dedicated log for only the stack traces. Is it obvious to the user where to find this log?

gobbleturk (approved):
Great looks good to me!

README.md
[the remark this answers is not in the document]
SurbhiJainUSC: Good point. I will add this information to PyPI README.

bvandermoon:
LGTM, great to see this

PR #54, opened 2023-06-08
Int8 clean

gobbleturk:
Int8 training via AQT

Instead of adding AQT to pip installs, this PR copies the two necessary files directly into maxtext. We use the aqt package folder structure so that when we later remove these files and switch to `pip install aqtp` the imports statements should still work.


rwitten (approved):
Looks great!

PR #59, opened 2023-06-13
Fix decode with sync checkpoint option

gobbleturk:
The recent API change allowing a sync checkpointer has to be plumbed into decode.py as well.


gobbleturk:
The recent API change allowing a sync checkpointer has to be plumbed into decode.py as well.

PR #60, opened 2023-06-13
Auto sharding base.yaml comment

priyanka-ganesha:


MaxText/configs/base.yml
```diff
@@ -63,11 +63,13 @@ logical_axis_rules: [
                     ]
 data_sharding: [['data', 'fsdp', 'tensor']]
 
-dcn_data_parallelism: -1
+# One axis for each parallelism type may hold a placeholder (-1)
```
gobbleturk, line 66: Can we mention that the default is to set the product of DCN axis (e.g. dcn_data_parallelism) to the number of slices and the ICI axis (e.g. ici_fsdp_parallelism) to the number of devices per slice?

gobbleturk (approved):
Great thank you! I left one comment requesting we explicitly say DCN==slices and ICI==devices per slice

PR #61, opened 2023-06-13
Unsetting GCS Bucket paths by default

priyanka-ganesha:
* Unset GCS Bucket paths in base.yml file and additional checks to make sure it is not empty before training
* Removed references to setting dcn parallelism value in README
* Changed hardcoded reference to dataset_path to use config value
* Edited multihost_dataloading_test.py to have gcs bucket params

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -89,7 +89,7 @@ class _HyperParameters():
     base_output_directory = raw_keys["base_output_directory"]
     assert base_output_directory, "Erroring out, need a real base_output_directory"
     assert raw_keys['dataset_path'], "Erroring out, need a real dataset_path.\n\
-      See instructions for dowloading the c4 dataset here:\n\
+      See instructions for downloading the c4 dataset here:\n\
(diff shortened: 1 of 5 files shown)


MaxText/pyconfig.py
```diff
@@ -87,6 +87,8 @@ def user_init(raw_keys):
     run_name = raw_keys["run_name"]
     assert run_name, "Erroring out, need a real run_name"
     base_output_directory = raw_keys["base_output_directory"]
+    assert base_output_directory, "Erroring out, need a real base_output_directory"
+    assert raw_keys['dataset_path'], "Erroring out, need a real dataset_path"
```
gobbleturk, line 91: I think it would be nice to also include a link to the readme section for dowloading the dataset here, something like "Erroring out, need a real dataset_path. See  instructions for dowloading the c4 dataset here: https://github.com/google/maxtext/blob/main/README.md#getting-started-download-dataset-and-configure."

gobbleturk (approved):
LGTM, left a comment suggesting you add a helpful link in the error message

PR #62, opened 2023-06-13
Add end to end tests for 16B and 32B param configs

raymondzouu:


gobbleturk (approved):
LGTM! Why have profiler enabled though?

raymondzouu:
> LGTM! Why have profiler enabled though?

The profiler would be useful for analyzing the performance.

PR #69, opened 2023-06-15
Cleaned eval_assert to have test_vars as list

priyanka-ganesha:
Based on: https://github.com/google/maxtext/pull/32#pullrequestreview-1423079565


gobbleturk (changes requested):
I LOVE the addition of the "test succeeded" statements. 

This is the right idea, but I've left comments for how to better pack/unpack lists in python

gobbleturk (approved):

PR #72, opened 2023-06-20
Avoid deprecated imports

rwitten:


tonyjohnchen (approved):
Thank you for pointing it out!

PR #82, opened 2023-07-07
Add version option and custom LIBTPU to stable

gobbleturk:


tonyjohnchen (approved):
Looks good! Thank you for adding it!

PR #84, opened 2023-07-10
Update 16b 32b tests to use full remat and larger batch size

gobbleturk:
Update 16b and 32b configs to use full remat and larger batch size

Also:
Add a config that runs these tests with additional optimizied rto
Remove datestring from the RUN_NAME so checkpointing works.


rwitten (approved):
This looks very reasonable!

PR #86, opened 2023-07-14
Code pieces must be re-written to transition MaxText to Support PyTorch/XLA

miladm:
Added comments on what files and code pieces must be re-written to transition MaxText to Support PyTorch/XLA

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -14,6 +14,9 @@
 
 # This sentinel is a reminder to choose a real run name.
 # If there is already a checkpoint under this run, that checkpoint will auto-resume.
+
+# we need to port this script to the pytorch model
+
 run_name: ""
 
 # If we aren't resuming from an existing checkpoint, load parameters from this path if provided.
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -129,6 +129,7 @@ def dot_product_attention(query: Array,
   # Normalize the attention weights across `kv_length` dimension.
   attn_weights = jax.nn.softmax(attn_weights).astype(dtype)
 
+  # Looks like this section is skipped in MaxText (skipping dropout). correct?
   # Apply attention dropout.
   if not deterministic and dropout_rate > 0.:
     keep_prob = 1.0 - dropout_rate
@@ -222,6 +223,8 @@ class DenseGeneral(nn.Module):
 
     contract_ind = tuple(range(0, len(axis)))
 
+    # quantization - maxtext specific? for short term purposes, bf16 should be
+    # sufficient?
     if not cfg.use_int8_training:
       return lax.dot_general(inputs, kernel, ((axis, contract_ind), ((), ())))
     else:
@@ -332,6 +335,7 @@ class MultiHeadDotProductAttention(nn.Module):
       #pylint: disable=no-value-for-parameter
       return self.kernel_init(*args) / depth_scaling
 
+    # kv cache is needed in the pytorch implementation
     # Project inputs_q to multi-headed q/k/v
     # dimensions are then [batch, length, num_heads, head_dim]
     query = projection(kernel_init=query_init, name='query')(inputs_q)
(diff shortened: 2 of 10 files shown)


MaxText/sequence_packing.py
```diff
@@ -16,6 +16,9 @@
 
 """Packed Sequence Op."""
 
+# can we use this script out of the box for pytorch? if we want to stay away
+# from tf.data, we should explore further on how to repurpose this helper scrpit
```
miladm, line 20: if we were to skip going through tf.data, how much work is repurposing this script for pytorch?

khatwanimohit: This script essentially contains map operations for packing multiple rows into one until the length matches the max_sequence_length, This can be done on huggingface's datasets. Should be 3 days of work

MaxText/input_pipeline.py
```diff
@@ -162,7 +168,7 @@ def filter_fn(x):
   # Return multi-host jax.Array prep iterator
   return multihost_gen
 
-
+# this method needs a rewrite in pytorch land (assuming no use of tf.data)
```
khatwanimohit, line 171: we can use huggingface datasets library to get C4: https://huggingface.co/datasets/c4

MaxText/input_pipeline.py
```diff
@@ -84,6 +88,8 @@ def _normalize_features(features):
 # -----------------------------------------------------------------------------
 
 
+# this method needs a rewrite in pytorch land (assuming no use of tf.data and
+# jax api)
```
khatwanimohit, line 92: Most of the pre-process operations like map, filter, padding are easily transferrable as they are available for huggingface datasets as well.

MaxText/multihost_dataloading.py
```diff
@@ -20,6 +20,9 @@
 Adapted from Sholto's:
 https://github.com/sholtodouglas/multihost_dataloading
 """
+
+# I expect this script must be rewritten for PyTorch/XLA
```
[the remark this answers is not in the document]
khatwanimohit: This script deals with how every host handles its shards of dataset and divides it among its local devices, further creating a global tensor representing a batch with the right sharding_spec. I have recently started this work and might take 1,2 weeks to land ( might be less)

MaxText/tokenizer.py
```diff
@@ -16,6 +16,8 @@
 
 """Provides op for tokenizing a dataset."""
 
+# I expect this script must be rewritten for PyTorch/XLA
+
```
khatwanimohit, line 20: We can easily re-write this script: https://pytorch.org/text/main/transforms.html#sentencepiecetokenizer. Should be 1,2 days of work.

MaxText/configs/base.yml
```diff
@@ -14,6 +14,9 @@
 
 # This sentinel is a reminder to choose a real run name.
 # If there is already a checkpoint under this run, that checkpoint will auto-resume.
+
+# we need to port this script to the pytorch model
```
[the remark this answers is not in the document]
gobbleturk: The logical axis rules dictate how different tensors are sharded (what axis are sharded via DP or FSDP or TP). Sharding is necessary for PyTorch, altho I think it would be okay if it's not configurable and just uses the Maxtext default (which will still take significant work that Mohit is more familiar with)

MaxText/layers.py
```diff
@@ -222,6 +223,8 @@ def __call__(self, inputs: Array) -> Array:
 
     contract_ind = tuple(range(0, len(axis)))
 
+    # quantization - maxtext specific? for short term purposes, bf16 should be
+    # sufficient?
```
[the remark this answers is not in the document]
gobbleturk: We are currently performing int8 experiments now to see the potential benefits. All perf numbers are based on bf16 - I think its okay if pytorch doesn't support int8 to start, but at some point it should since its likely there is a large performance benefit.

MaxText/multihost_dataloading.py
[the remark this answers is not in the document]
miladm: do you have an estimate of effort?

MaxText/layers.py
```diff
@@ -350,6 +355,7 @@ def query_init(*args):
     value = checkpoint_name(value, 'value_proj')
 
     if decode:
+      # pytorch needs kv cache
```
miladm, line 358: @jonb377 wdyt?

PR #88, opened 2023-07-18
Add MLP remat poicy

raymondzouu:


rwitten (changes requested):
Let's close this because it isn't actively in flight at this point. Additionally `remat_saved_tensors` can't be an array if we want it to be passed at the command line (which I think we do?)

PR #91, opened 2023-07-21
Remove all sudo

tonyjohnchen:
Remove all `sudo` from `setup.sh`.
Update the custom_libtpu_path from `/lib/libtpu.so` to `$HOME/custom_libtpu/libtpu.so`.

This would be helpful for GKE since GKE run in privileged mode as root and it doesn't allow sudo.
It is also a good practice to not use `sudo`

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -37,7 +37,7 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE) ]]; then
      exit 1
 fi
 
-libtpu_path="/lib/libtpu.so"
+libtpu_path="$HOME/custom_libtpu/libtpu.so"
 
 if [[ "$MODE" == "libtpu-only" ]]; then
     # Only update custom libtpu.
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -37,17 +37,17 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE) ]]; then
...
     # Only update custom libtpu.
     if [[ -n "$LIBTPU_GCS_PATH" ]]; then
         # Install custom libtpu
         echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
         # Install required dependency
-        sudo pip3 install -U crcmod
+        pip3 install -U crcmod
         # Copy libtpu.so from GCS path
-        sudo gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
+        gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
```
khatwanimohit, line 50: Do you think we should set `TPU_LIBRARY_PATH` in this script, whenever user provides `LIBTPU_GCS_PATH`. This way we can avoid user making the \$HOME vs. $HOME mistake. 
WDYT ?

tonyjohnchen: Good idea, but I don't think it help too much since `TPU_LIBRARY_PATH` will be reset each time you SSH into the TPU VM.

gobbleturk (approved):

khatwanimohit (approved):

PR #92, opened 2023-07-21
Add project and zone arguments to multihost_runner

gobbleturk:
* Add project and zone arguments to multihost_runner like in multihost_job
* Make sure a useful message is printed when the project/zone don't match or the TPU isn't found in the region.
* Use global ARGS instead of parsing and passing into individual ones like multihost_job

### DIFF
--- a/multihost_runner.py
+++ b/multihost_runner.py
@@ -97,19 +97,25 @@ def get_slices():
   """ Returns a list of slices matching TPU_PREFIX """
   command = [
       "gcloud", "alpha", "compute", "tpus", "tpu-vm", "list",
-      f"--filter=name~{args.TPU_PREFIX}", "--format=csv(name,accelerator_type)", f"--project={project}", f"--zone={zone}"
+      f"--filter=name~{args.TPU_PREFIX}", "--format=csv(name,accelerator_type)", f"--project={args.PROJECT}", f"--zone={args.ZONE}"
   ]
-  completed_command = subprocess.run(command, capture_output=True, check=True)
+  try:
+    completed_command = subprocess.run(command, capture_output=True, check=True)
+  except subprocess.CalledProcessError as e:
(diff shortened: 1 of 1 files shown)


multihost_runner.py
```diff
@@ -342,37 +356,34 @@ def main() -> None:
...
 
   #### Parse flags ####
-  tpu_prefix = args.TPU_PREFIX
-  script_dir = args.SCRIPT_DIR
-  main_command = args.COMMAND
-  run_name = args.RUN_NAME
-  use_existing_folder = args.USE_EXISTING_FOLDER
-  internal_ip = args.INTERNAL_IP
+  if not args.PROJECT:
+    args.PROJECT = get_project()
```
michelle-yooh, line 360: IMO, it's not the best practice to set the args values backward. we can consider setting these as global vairables instead?

multihost_runner.py
```diff
@@ -200,23 +227,24 @@ def execute_main_command(main_command,slices, local_log_dir, run_name, zip_name,
     for worker_num in range(cur_slice.num_workers):
       output_filename = os.path.join(local_log_dir, f"output_slice_{cur_slice.slice_num:04d}_worker_{worker_num:04d}.txt")
       output_logs.append(output_filename)
-      mkdir_command = f"mkdir -p {run_name}"
-      mv_zip_command = f"mv {zip_name} {run_name}"
-      cd_command = f"cd {run_name}"
+      mkdir_command = f"mkdir -p {args.RUN_NAME}"
```
michelle-yooh, line 230: Is there a chance we will call this function from another module? Calling this as `args.RUN_NAME` won't make it reusable by other modules.

michelle-yooh: discussed offline

multihost_runner.py
gobbleturk: Changed to set the default in the flag definition. In much older versions of the code we used a different flag parser where I think this was undesirable because the --help printed the default values (which would run these weird functions in this case), but with argparse the default values are not printed with --help, so using a complex function in the default seems fine

michelle-yooh (approved):
lgtm

PR #93, opened 2023-07-22
Bucket name normalize

gobbleturk:
We choose to normalize the bucket name (so it can include the "gs://" or not in multihost_job since it is clearly a bucket by the variable name, however in maxtext we instead choose to validate and enforce the gs:// since the base_output_directory and dataset_path are not obviously buckets (perhaps they should be renamed, but this validator will quickly instruct users what to do).

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -89,9 +89,12 @@ class _HyperParameters():
     assert run_name, "Erroring out, need a real run_name"
     base_output_directory = raw_keys["base_output_directory"]
     assert base_output_directory, "Erroring out, please set a real base_output_directory"
-    assert raw_keys['dataset_path'], "Erroring out, please set a real dataset_path in configs/base.yml\n\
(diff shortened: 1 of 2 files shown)


MaxText/pyconfig.py
```diff
@@ -88,10 +88,13 @@ def user_init(raw_keys):
     run_name = raw_keys["run_name"]
     assert run_name, "Erroring out, need a real run_name"
     base_output_directory = raw_keys["base_output_directory"]
-    assert base_output_directory, "Erroring out, please set a real base_output_directory"
-    assert raw_keys['dataset_path'], "Erroring out, please set a real dataset_path in configs/base.yml\n\
+    assert base_output_directory, "please set a real base_output_directory"
```
Obliviour, line 91: nit: it looks like the `set a real {config_var} assert` can also be nicely added into the `validate_gcs_bucket_name` function. This would remove a whole line of duplicated code!

Obliviour (approved):

MaxText/pyconfig.py
gobbleturk: Done!

PR #97, opened 2023-07-26
Modify multihost_runner kill script for multiple PIDs

gobbleturk:
Rarely either you can have multiple orphaned PIDs or PIDs stranded on multiple devices, this change now handles both.


bvandermoon:
Thanks Matt. This LGTM. Just want to understand if there did end up being some way to test this beforehand? I suspect that you just tested killing a process in general as discussed, but just want to confirm

bvandermoon (approved):

gobbleturk:
I couldn't think of a way to create multiple hanging PIDs - such a state seems hard to get into. I did test running it clean (so we hit "No existing processes found, so your TPU is ready to use!") and with one job running, which gets correctly de-duped as a single PID and killed

PR #99, opened 2023-07-27
Int8 update

gobbleturk:


gobbleturk:
Moving to PR https://github.com/google/maxtext/pull/106 instead

PR #102, opened 2023-07-31
Add support for curl command

tonyjohnchen:
Example usage:

```
python3 multihost_job.py --COMMAND_TYPE=curl --NUM_SLICES=$NUM_SLICES --RUN_NAME=$RUN_NAME --BUCKET_NAME=$BUCKET_NAME --PROJECT=${PROJECT} --ZONE=${ZONE} --TPU_TYPE=$TPU_TYPE --VERSION=$VERSION --COMMAND=""
```

### DIFF
--- a/create_vlp_stable_fleet.sh
+++ b/create_vlp_stable_fleet.sh
@@ -1,0 +1,18 @@
+# PROJECT=tpu-prod-env-multipod
+PROJECT=tpu-prod-env-vlp-2nic
+ZONE=us-east5-b
+NUM_SLICES=2
+TPU_TYPE=v5litepod-256
+VERSION=v2-alpha-tpuv5-lite
+BUCKET_NAME=tonyjohnchen-maxtext
+
+for (( i=0; i<$iteration; i++ ));
+do
+    RUN_NAME=startup_time_${NUM_SLICES}slice_${TPU_TYPE}_$(date +%Y-%m-%d-%H-%M-%S)_stable_fleet
+    QR_ID=$RUN_NAME
+    python3 multihost_job.py --COMMAND_TYPE=curl --NUM_SLICES=$NUM_SLICES --RUN_NAME=$RUN_NAME --BUCKET_NAME=$BUCKET_NAME --PROJECT=${PROJECT} --ZONE=${ZONE}\
+    --TPU_TYPE=$TPU_TYPE --VERSION=$VERSION \
+    --COMMAND="echo hello"
+done
+
+gcloud alpha compute tpus queued-resources list --project=${PROJECT} --zone=${ZONE} --filter=startup_time 


multihost_job.py
```diff
@@ -154,6 +159,49 @@ def run_create_resources(startup_script_file):
...
+                        "runtime_version": args.VERSION,
+                        "network_config": {
+                            "network": "default",
+                            "subnetwork": "default",
+                            "enable_external_ips": True
+                        },
+                        "scheduling_config": {
+                            "maintenance_interval": 2
+                        },
+                        "metadata": {
```
Obliviour, line 180: Is the metadata needed? Looks startup related?

multihost_job.py
```diff
@@ -154,6 +159,49 @@ def run_create_resources(startup_script_file):
   captured_output = subprocess.run(command, check=False, shell=True, capture_output=True)
   return captured_output
 
+def run_create_resources_curl(startup_script):
+  """ Run the Create Queued Resources (CQR) request with curl command"""
+  data = {
+        "tpu": {
+            "node_spec": [
```
gobbleturk, line 166: Remove the bracket on this line and the closing one on line 188. It is unecessary and removing it allows it to make it easier flexiblity

multihost_job.py
```diff
@@ -154,6 +159,49 @@ def run_create_resources(startup_script_file):
...
+                            "enable_external_ips": True
+                        },
+                        "scheduling_config": {
+                            "maintenance_interval": 2
+                        },
+                        "metadata": {
+                            "startup-script": startup_script
+                        }
+                    },
+                    "multi_node_params": {
```
gobbleturk, line 184: Remove multi_node params - this should not be a field in single slice, so we should only add it after first checking num_slices > 1

multihost_job.py
```diff
@@ -154,6 +159,49 @@ def run_create_resources(startup_script_file):
...
+                        "node_count": args.NUM_SLICES
+                    }
+                }
+            ]
+        }
+    }
+
```
gobbleturk, line 191: After removing the list format of node_spec and the multi_node_params field, you can add 

```
  if args.NUM_SLICES > 1:
    multi_node_dict = {"node_count": args.NUM_SLICES, "node_id_prefix": args.RUN_NAME}
    data["tpu"]["node_spec"]["multi_node_params"] = multi_node_dict
  else:
    data["tpu"]["node_spec"]["node_id"] = args.RUN_NAME
```
    
  To support both single and multi slice
  
  With these changes the lines should look like https://paste.googleplex.com/6241383198228480

raymondzouu:
I was trying out these changes with the curl command and a couple things I noticed so far are
* it did not return an error when using the curl command when there was an existing qr with the same name
* if I wanted to use a custom network like mtu9k there is no way to set it from the command

gobbleturk:
This looks okay to me, but I will re-iterate my earlier point that Raymond also brought up - the API of multihost_job no longer supports customization of the curl command arguments, e.g. custom network, or best-effort/on-demand/reserved.

We could add back all of the args  in https://github.com/google/maxtext/commit/5796aa61a3a83d93cf9c39f6c6f01fc2db386bc0#diff-dfaa2b9b91181624bf5f2a6647006e1abea1ea43c1dd409fe89e6a45958834b1L193 which allows the curl to be customizable, but I think this is may be more confusing to external users who have no reason to curl.

The only reason to use curl instead of CLI is for new features that haven't rolled out to the CLI yet - for this reason it is only internal users and not external customers who need curl, so I think this change is really just for us. Thats why at first I advocated keeping this in a local branch, but I think enough of us depend on CURL that it is convenient for us to push this lightweight curl change to main. If you need to customize the curl (e.g. custom network), you can look at the previous implementation I linked earlier or the curl documentation to figure out how to modify the multihost_job curl command here for your use-case.

gobbleturk (approved):

PR #104, opened 2023-08-03
Shuffle dataset with seed

khatwanimohit:
Tested with `end_to_end/test_determinism.sh` on V4-8.

### DIFF
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -96,7 +96,8 @@ def preprocessing_pipeline(
   shift: bool = True,
   drop_remainder: bool = True,
   prefetch_size = tf.data.experimental.AUTOTUNE,
-  data_sharding = None
+  data_sharding = None,
+  seed = 0,
 ):
   """Shuffle and batch/pack the given dataset."""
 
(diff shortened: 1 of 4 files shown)


MaxText/train.py
```diff
@@ -281,14 +282,20 @@ def train_loop(config, state=None):
   mesh = Mesh(devices_array, config.mesh_axes)
 
   # Set up datasets.
+  print("check type ", type(init_rng))
```
gobbleturk, line 285: Is this left on purpose? Is it just a debugging statement?

MaxText/input_pipeline.py
```diff
@@ -96,7 +96,8 @@ def preprocessing_pipeline(
   shift: bool = True,
   drop_remainder: bool = True,
   prefetch_size = tf.data.experimental.AUTOTUNE,
-  data_sharding = None
+  data_sharding = None,
+  seed = 0,
```
gobbleturk, line 100: I am also planning to add an initial random seed for the model weights, so we need to be more specific than "seed", can we call this "data_shuffle_seed" instead?

gobbleturk (approved):
This looks great, thank you for adding this!

Just one main comment - can we rename seed to "data_shuffle_seed" or similar?

PR #105, opened 2023-08-04
Add seed for initial model weights

gobbleturk:
* Add seed for initial model weights
* Remove un-used method predict_step in train.py. It is only used decode.py and called locally there.
      * We are removing predict_step in this PR since the predict_step also has a source of PRNG which could be a source of 
         confusion


khatwanimohit (approved):
LGTM

PR #106, opened 2023-08-05
Update int8 8-4-23

gobbleturk:
Update int8 code to reflect latest AQT (taken from google's codebase).

In the future we will use AQT from pip instead, but as of now AQT is not updated in pip.


michelle-yooh (approved):
LGTM

PR #111, opened 2023-08-11
init commit for GKE docker image

tonyjohnchen:
This scripts should build GKE docker image for maxtext.

### DIFF
--- a/GKE_docker_image_tool/build_docker_image.sh
+++ b/GKE_docker_image_tool/build_docker_image.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Description:
+# bash build_docker_image.sh MODE=stable
+# bash build_docker_image.sh MODE=nightly
+# bash build_docker_image.sh MODE=head
+
+# bash build_docker_image.sh MODE=stable PROJECT=cloud-tpu-multipod-dev
+# bash build_docker_image.sh MODE=stable JAX_VERSION=0.4.13 PROJECT=cloud-tpu-multipod-dev
+# bash build_docker_image.sh MODE=stable JAX_VERSION=0.4.13 PROJECT=cloud-tpu-multipod-dev IMAGE_PREFIX=jax0.4.13
+
+
+# Enable "exit immediately if any command fails" option
+set -e
+
+export PROJECT=tpu-prod-env-multipod
+export IMAGE_PREFIX=gke_img
+
+# Set environment variables
+for ARGUMENT in "$@"; do
+    IFS='=' read -r KEY VALUE <<< "$ARGUMENT"
+    export "$KEY"="$VALUE"
+    echo "$KEY"="$VALUE"
+done
+
+if [[ "$MODE" == "stable" || ! -v MODE ]]; then
+    # Stable mode
+  if [[ ! -v JAX_VERSION ]]; then
+    docker build --build-arg MODE="stable" -f ./maxtext.Dockerfile -t ${IMAGE_PREFIX}_maxtext_stable .
+    docker tag ${IMAGE_PREFIX}_maxtext_stable gcr.io/$PROJECT/${IMAGE_PREFIX}_maxtext_stable:latest
+    docker push gcr.io/$PROJECT/${IMAGE_PREFIX}_maxtext_stable:latest
+  else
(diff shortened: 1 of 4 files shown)


GKE_docker_image_tool/build_docker_image.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# bash build_docker_image.sh MODE=stable PROJECT=cloud-tpu-multipod-dev
+# bash build_docker_image.sh MODE=stable JAX_VERSION=0.4.13 PROJECT=cloud-tpu-multipod-dev
+# bash build_docker_image.sh MODE=stable JAX_VERSION=0.4.13 PROJECT=cloud-tpu-multipod-dev IMAGE_PREFIX=jax0.4.13
+
+
+# Enable "exit immediately if any command fails" option
+set -e
+
+export PROJECT=tpu-prod-env-multipod
```
rwitten, line 30: Can this one come from the environment? Something like:
https://source.corp.google.com/piper///depot/google3/experimental/users/vbarr/multipod/xpk/xpk.py;l=231-248?q=xpk

GKE_docker_image_tool/maxtext.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG JAX_VERSION
+ENV ENV_JAX_VERSION=$JAX_VERSION
+
+# Set the working directory in the container
+WORKDIR /app
+
+RUN git clone https://github.com/google/maxtext.git
+
+RUN echo "Running command: bash setup.sh MODE=$ENV_MODE JAX_VERSION=$ENV_JAX_VERSION"
+RUN cd maxtext && bash setup.sh MODE=${ENV_MODE} JAX_VERSION=${ENV_JAX_VERSION}
```
rwitten, line 32: This workflow is a good start but doesn't seem to address the typical user flow.

Imagine the user's flow is (1) make a change to MaxText locally, (2) test it locally either (2a) in a docker image or (2b) in on the raw VM and then (3) upload a docker image to GKE.

This script assumes that you're grabbing code from MaxText off of Git (so it doesn't work with local changes) and the rebuilds will be slow because you assume we want to reinstall dependencies each time (even though 95+ of changes don't change the dependencies).

Instead, we want:
* use the local version of MaxText (so the user can easily make changes). (Notice that if the user wants the one on github they can just `git clone` it!
* support gracefully either rebuilding with dependencies or just copying over the local MaxText. (usually there isn't a change to the dependencies so there is no reason to setup.sh) I believe Docker handles this via layering where the dependencies would be a different layer.
* [More a question] How would the user develop locally in their docker container? This could either be running a train.py or doing an XAOT and it might be invoked as part of building or it might be interactive.
* [Totally a nit] Can the MaxText directory be the root directory in the docker image? It is super annoying that scripts need to `cd maxtext` in the Docker image but not outside. I wonder if docker can be set to have a specific default working directory?

rwitten (changes requested):
Great start!

GKE_docker_image_tool/build_docker_image.sh
tonyjohnchen: Done

GKE_docker_image_tool/maxtext.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG MODE
+ENV ENV_MODE=$MODE
+
+ARG JAX_VERSION
+ENV ENV_JAX_VERSION=$JAX_VERSION
+
+# Set the working directory in the container
+WORKDIR /app
+
+RUN git clone https://github.com/google/maxtext.git
```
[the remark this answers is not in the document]
tonyjohnchen: Done

GKE_docker_image_tool/maxtext.Dockerfile
tonyjohnchen: 1) 2) 4) done

bvandermoon (commented):
Could you add a quick README with some recommended steps to run this?

tonyjohnchen:
> Could you add a quick README with some recommended steps to run this?

Hi Branden, 
I am planning to write a readme as internal doc first. I will update the public readme for maxtext once GKE and docker image are fully ready.

build_docker_image.sh
```diff
@@ -0,0 +1,62 @@
...
+    export "$KEY"="$VALUE"
+    echo "$KEY"="$VALUE"
+done
+echo "PROJECT=$PROJECT"
+
+if [[ "$MODE" == "stable" || ! -v MODE ]]; then
+    # Stable mode
+  if [[ ! -v JAX_VERSION ]]; then
+    docker build --build-arg MODE="stable" -f ./maxtext.Dockerfile -t ${IMAGE_PREFIX}_maxtext_stable .
+    docker tag ${IMAGE_PREFIX}_maxtext_stable gcr.io/$PROJECT/${IMAGE_PREFIX}_maxtext_stable:latest
```
bvandermoon, line 45: Nit: Not a blocker but if these two lines are the same as the else, they can be pulled outside of this if/else

bvandermoon:
> 

Sounds good, thanks Tony

bvandermoon:
> Hi Branden,
I am planning to write a readme as internal doc first. I will update the public readme for maxtext once GKE and docker image are fully ready.

Sounds good, thanks Tony

bvandermoon (approved):

build_docker_image.sh
tonyjohnchen: Good call!

Done.

PR #115, opened 2023-08-17
add an option of using aqt dq dg

ychzhang:


MaxText/configs/base.yml
```diff
@@ -40,6 +40,7 @@ gcs_metrics: False
 # Activation dtypes.
 dtype: "bfloat16"
 int8_training: False
+use_dqdg: False
```
lukaszlew, line 43: Add a comment that it has effect only if int8_training=True on this and 2 flags belof

MaxText/layers.py
```diff
@@ -225,12 +226,16 @@ def __call__(self, inputs: Array) -> Array:
     if not cfg.int8_training:
       return lax.dot_general(inputs, kernel, ((axis, contract_ind), ((), ())))
     else:
-      aqt_cfg = maxtext_sweeps.sweep1(cfg.fwd_int8, cfg.bwd_int8)
-      aqt_dot_general = aqt.make_dot_general(aqt_cfg)
       aqt_key = self.make_rng('aqt')
-      context = aqt.Context(key=aqt_key, train_step=None)
+      if cfg.use_dqdg:
+        aqt_dq_dg = aqt_dq.make_aqt_dq_dg()
+        return aqt_dq_dg(aqt_key, inputs, kernel, ((axis, contract_ind), ((), ())))
```
lukaszlew, line 232: how come it does not need rng?

lukaszlew (changes requested):

MaxText/layers.py
ychzhang: It needs. Rng is the first argument of this function (not in Context). Whereas rng for aqt v2 is wrapped in a context object.

lukaszlew: I'm blind :)

PR #116, opened 2023-08-17
do not quantize logits matmul

ychzhang:


MaxText/configs/base.yml
```diff
@@ -40,10 +40,11 @@ gcs_metrics: False
 # Activation dtypes.
 dtype: "bfloat16"
 int8_training: False
-# The following three flags will only have effect when int8_training is True
+# The following four flags will only have effect when int8_training is True
 use_dqdg: False
 fwd_int8: True
 bwd_int8: True
+quantize_logits: False
```
lukaszlew, line 47: change to true.

PR #124, opened 2023-08-24
Add support for per_device_batch_size<1

raymondzouu:
Support `per_device_batch_size<1` by loading the data as if `per_device_batch_size=1` and then decimating proportion of the data based on `per_device_batch_size` in `train_step`.

### DIFF
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -217,11 +217,7 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       tokenizer.TokenizeOp(sp_tokenizer), num_parallel_calls=AUTOTUNE)
 
   # Set global batch size.
-  if config.per_device_batch_size < 1:
-    # For per_device_batch_size<1, we load the data as if per_device_batch_size=1
-    batch_size = global_mesh.size
-  else:
-    batch_size = max_utils.calculate_raw_input_batch_size(config.per_device_batch_size)
+  batch_size = config.batch_size_to_load
 
   if config.eval_per_device_batch_size > 0:
     eval_batch_size = config.eval_per_device_batch_size * global_mesh.size
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -29,7 +29,6 @@ from jax.sharding import PartitionSpec as P
 import tokenizer
 import multihost_dataloading
 import sequence_packing
-import max_utils
 
 AUTOTUNE = tf.data.experimental.AUTOTUNE
 
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -216,7 +216,11 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       tokenizer.TokenizeOp(sp_tokenizer), num_parallel_calls=AUTOTUNE)
 
   # Set global batch size.
-  batch_size = config.per_device_batch_size * global_mesh.size
+  if config.per_device_batch_size < 1:
+    batch_size = global_mesh.size
+  else:
+    batch_size = int(config.per_device_batch_size) * global_mesh.size
+
   if config.eval_per_device_batch_size > 0:
     eval_batch_size = config.eval_per_device_batch_size * global_mesh.size
   else:
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -216,12 +216,12 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       tokenizer.TokenizeOp(sp_tokenizer), num_parallel_calls=AUTOTUNE)
 
   # Set global batch size.
-  batch_size = config.batch_size_to_load
+  global_batch_size_to_load = config.global_batch_size_to_load
 
   if config.eval_per_device_batch_size > 0:
     eval_batch_size = config.eval_per_device_batch_size * global_mesh.size
   else:
-    eval_batch_size = batch_size
+    eval_batch_size = global_batch_size_to_load
 
   def filter_keys(record):
     return {'inputs': record['inputs'], 'targets': record['targets']}
@@ -230,7 +230,7 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
 
   train_iter = preprocessing_pipeline(
       train_ds,
-      batch_size,
+      global_batch_size_to_load,
       global_mesh,
       shuffle=config.enable_data_shuffling,
       num_epochs=None,
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -29,6 +29,7 @@ from jax.sharding import PartitionSpec as P
 import tokenizer
 import multihost_dataloading
 import sequence_packing
+import max_utils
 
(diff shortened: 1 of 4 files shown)


MaxText/input_pipeline.py
```diff
@@ -216,7 +216,11 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       tokenizer.TokenizeOp(sp_tokenizer), num_parallel_calls=AUTOTUNE)
 
   # Set global batch size.
-  batch_size = config.per_device_batch_size * global_mesh.size
+  if config.per_device_batch_size < 1:
+    batch_size = global_mesh.size
```
rwitten, line 220: Hmmm this code looks right but it is a little hard to read because the batch size doesn't have the normal meaning. Can you consider how to make this easier for future people to follow?

MaxText/max_utils.py
```diff
@@ -167,10 +167,16 @@ def init_train_state(model, tx, config, key):
 
   Args: model, tx, config, key
   """
-  input_shape = (
-      len(jax.devices()) * config.per_device_batch_size,
-      config.max_target_length
-  )
+  if config.per_device_batch_size < 1:
```
rwitten, line 170: since you keep doing this calculation, should it be shared with the above. Something like "calculate_raw_input_batch_size()" as a function.

MaxText/train.py
```diff
@@ -148,14 +148,15 @@ def record_activation_metrics(output_metrics, intermediate_outputs, config):
       output_metrics['scalar'][f'activ_mean/layer_{layer_num:03d}'] = layer["activation_mean"][0]
       output_metrics['scalar'][f'activ_stdev/layer_{layer_num:03d}'] = layer["activation_stdev"][0]
 
-def train_step(model, config, state, data, dropout_rng):
+def train_step(model, config, state, data, dropout_rng, batch_size):
```
rwitten, line 151: hmmm shouldn't be adding random arguments. Can config calculate this? (If not, let's have config derive this.)

MaxText/train.py
```diff
@@ -260,7 +266,7 @@ def train_loop(config, state=None):
                        data_pspec,
                        None),
     out_shardings=(state_mesh_annotations, None, None),
-    static_argnums=(0,1,),
+    static_argnums=(0,1,5,),
```
rwitten, line 269: why we shouldn't have new args floating around.

rwitten (changes requested):
Looks like it works to me! Does it work?

Some code nits.

MaxText/input_pipeline.py
raymondzouu: Added a comment to clarify why we use that batch_size calculation.

MaxText/max_utils.py
raymondzouu: Done.

MaxText/train.py
raymondzouu: I changed it so that it will call `max_utils.calculate_raw_input_batch_size` to calculate the `batch_size`.

MaxText/train.py
raymondzouu: Removed this

MaxText/train.py
```diff
@@ -167,6 +168,12 @@ def train_step(model, config, state, data, dropout_rng):
   rng1, gen_aqt_rng = jax.random.split(dropout_rng)
   aqt_rng, rng2 = jax.random.split(gen_aqt_rng)
 
+  if config.per_device_batch_size < 1:
+    # decimate data by proportion of per_device_batch_size
+    batch_size = max_utils.calculate_raw_input_batch_size(config.per_device_batch_size)
```
raymondzouu, line 173: Would it be better to store `batch_size` into `config` before calling `train_step` so that we don't have to keep calling `max_utils.calculate_raw_input_batch_size` to recalculate the value? I saw that `__setattr__` https://github.com/google/maxtext/blob/50f9300c4f4d7f6cbc52db538b62d0397262d663/MaxText/pyconfig.py#L148-L150 to modify `config` just raises an error. Is there a reason why `config` should not be modified?

MaxText/max_utils.py
raymondzouu: Done.

MaxText/train.py
rwitten: Yes that would be MUCH better. `config` can't be edited once given to a user but it can do stuff before it is used.

Generally you should calculate a `batch_to_load` and a `batch_to_train_on`.

Also there is no need for an `if` here. It is always safe to slice, right?

rwitten (commented):
Just nits.

MaxText/train.py
raymondzouu: Changed it so that `batch_size_to_load` and `batch_size_to_train_on` are calculated in `pyconfig.py` and then stored in `config`. Yes it should be safe to always slice but would it be better to only slice when necessary?

MaxText/input_pipeline.py
```diff
@@ -216,7 +216,8 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       tokenizer.TokenizeOp(sp_tokenizer), num_parallel_calls=AUTOTUNE)
 
   # Set global batch size.
-  batch_size = config.per_device_batch_size * global_mesh.size
+  batch_size = config.batch_size_to_load
```
rwitten, line 219: `global_batch_size_to_load`

MaxText/pyconfig.py
```diff
@@ -109,6 +109,8 @@ def user_init(raw_keys):
     raw_keys['mlp_dim'] = 2**mlp_dim_scale * raw_keys['base_mlp_dim']
     raw_keys['num_decoder_layers'] = 2**layer_scale * raw_keys['base_num_decoder_layers']
 
+    raw_keys['batch_size_to_load'], raw_keys['batch_size_to_train_on'] = get_batch_sizes(raw_keys['per_device_batch_size'])
```
rwitten, line 112: `global_batch_size_to_train_on`

MaxText/pyconfig.py
```diff
@@ -134,7 +136,16 @@ def get_individual_scales(scale):
   layer_scale = base_scale
   return emb_scale, num_head_scale, mlp_dim_scale, layer_scale
 
+def get_batch_sizes(per_device_batch_size):
```
rwitten, line 139: `calculate_global_batch_sizes` since you're deriving this one. But totally a nit, even nittier than my other nits and I wouldn't make you redo just for that.

MaxText/train.py
```diff
@@ -167,6 +167,11 @@ def train_step(model, config, state, data, dropout_rng):
   rng1, gen_aqt_rng = jax.random.split(dropout_rng)
   aqt_rng, rng2 = jax.random.split(gen_aqt_rng)
 
+  if config.per_device_batch_size < 1:
```
rwitten, line 170: Again do we need to branch here? Can't we always slice?

rwitten (commented):
Minor points, starting to look quite polished.

MaxText/train.py
raymondzouu: Sure I removed it, it should be safe to always slice

PR #129, opened 2023-08-31
Add end to end tests for 64B params configs

michelle-yooh:


gobbleturk:
This looks great to me! Could you confirm two things:
1. The 16B/32B still get about the same performance with this change
2. The default maxtext setting still works (doesn't OOM) on a v4-8? You can decrease the default batch size if so.

rwitten (approved):
LGTM

michelle-yooh:
> This looks great to me! Could you confirm two things:
> 
> 1. The 16B/32B still get about the same performance with this change
> 2. The default maxtext setting still works (doesn't OOM) on a v4-8? You can decrease the default batch size if so.

1. The existing 16B/32B are not affected as they overwrite base.yaml. Do we want to `scale` for 16B/32B as well? They perform slightly better in terms of MFU with scale. See row#11-15 [here](https://docs.google.com/spreadsheets/d/1aNJOvAfszxr_4TOvmnmfKstoUV2D1s3iiMJfCRuWhu4/edit?resourcekey=0-Y-kmap2mMW_hawB3AAc6FQ#gid=153403776)
2. The new base.yaml runs fine on v4-8. See the logs [here](https://paste.googleplex.com/6205457944281088)

gobbleturk (approved):
Great, thank you for confirming!

PR #130, opened 2023-09-01
Add network rto setup for GKE runs

raymondzouu:


gobbleturk (approved):

rwitten (approved):
Are we sure we can't have one path for GCE and GKE? If not this looks ok.

gke_rto_setup.sh
```diff
@@ -0,0 +1,15 @@
+echo "Running gke_rto_setup.sh..."
+
+echo "install iproute2 ethtool lsof"
+apt-get install -y iproute2 ethtool lsof
```
rwitten, line 4: Please move this into docker create.

rwitten (changes requested):

gke_rto_setup.sh
raymondzouu: Done.

maxtext_dependencies.Dockerfile
```diff
@@ -4,6 +4,9 @@ FROM python:3.10-slim
 # Install system dependencies and Git
 RUN apt-get update && apt-get install -y curl gnupg git
 
+# Install iproute2 ethtool lsof
+RUN apt-get update && apt-get install -y iproute2 ethtool lsof
```
rwitten, line 8: Shoudl we add these to line #5? Just thinking out loud, not sure which is better or clearer.

rwitten (approved):
Probably ok one random question

maxtext_dependencies.Dockerfile
raymondzouu: Yeah it could be added to that line but I think it is clearer to keep it separate as the dependencies for adjusting network?

rwitten: Probably should have the unifying comment then. But nbd this looks pretty good.

PR #136, opened 2023-09-07
Add 128B config with modified automatic scale

michelle-yooh:
I manually verified # of params to be 128B.
MFU is to be computed

### DIFF
--- a/MaxText/configs/largest_job/128b.sh
+++ b/MaxText/configs/largest_job/128b.sh
@@ -1,0 +1,28 @@
+echo "Running 128b.sh"
+# Example command to invoke this script
+# bash MaxText/configs/largest_job/128b.sh
+
+export OUTPUT_PATH="gs://maxtext-experiments-multipod"
+export DATASET_PATH="gs://maxtext-dataset/"
+
+# Set environment variables
+for ARGUMENT in "$@"; do
+    IFS='=' read -r KEY VALUE <<< "$ARGUMENT"
+    export "$KEY"="$VALUE"
+done
+
+# Set up network
+bash gke_rto_setup.sh
+
+# For DNS lookup when running on large number of VMs
+echo '142.250.123.95 www.googleapis.com' | tee -a /etc/hosts
+echo '142.251.4.128 storage.googleapis.com' | tee -a /etc/hosts
+
+# Train
+export LIBTPU_INIT_ARGS="--xla_tpu_enable_data_parallel_all_reduce_opt=true --xla_tpu_data_parallel_opt_different_sized_ops=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true --xla_tpu_overlap_compute_collective_tc=true --xla_enable_async_all_gather=true"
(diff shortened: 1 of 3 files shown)


MaxText/configs/largest_job/128b.sh
```diff
@@ -0,0 +1,28 @@
+echo "Running 128b.sh"
+# Example command to invoke this script
+# bash MaxText/configs/largest_job/128b.sh
+
+export OUTPUT_PATH="gs://maxtext-experiments-multipod"
+export DATASET_PATH="gs://maxtext-dataset/"
+
+# Set environment variables
+for ARGUMENT in "$@"; do
```
rwitten, line 9: dead code ?

rwitten (approved):
Tiny nit but LGTM otherwise.

MaxText/configs/largest_job/128b.sh
```diff
@@ -0,0 +1,28 @@
...
+# Set up network
+bash gke_rto_setup.sh
+
+# For DNS lookup when running on large number of VMs
+echo '142.250.123.95 www.googleapis.com' | tee -a /etc/hosts
+echo '142.251.4.128 storage.googleapis.com' | tee -a /etc/hosts
+
+# Train
+export LIBTPU_INIT_ARGS="--xla_tpu_enable_data_parallel_all_reduce_opt=true --xla_tpu_data_parallel_opt_different_sized_ops=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true --xla_tpu_overlap_compute_collective_tc=true --xla_enable_async_all_gather=true"
+python3 MaxText/train.py MaxText/configs/base.yml run_name=$RUN_NAME\
```
rwitten, line 23: also no run name right? This is coming from XPK automatically?

rwitten (changes requested):

MaxText/configs/largest_job/128b.sh
raymondzouu: I added this to the other scripts as well in case you wanted to change any of the env variables (`DATASET_PATH`, `OUTPUT_PATH`, `RUN_NAME`)

MaxText/configs/largest_job/128b.sh
raymondzouu: This is if you wanted to change the `RUN_NAME`. If it isn't specified then it'll fall back to the XPK name

MaxText/configs/largest_job/128b.sh
rwitten: ok but can you give sensible defaults? things should work without them. Also you shouldn't need to set any of these while debugging.

michelle-yooh: The script already has the default values for `OUTPUT_PATH` and `DATASET_PATH`. The `RUN_NAME` is supposed to come from xpk if not set. Do you mean that I should add a default value for `RUN_NAME` for non-XPK case? and Could you clarify what you mean by it should work without setting those env variables? I tried running without them, and I was getting errors

rwitten: My mistake thanks Michelle. This is good to merge.

rwitten (approved):

PR #139, opened 2023-09-08
Cosine Learning Rate Schedule

gobbleturk:
Change learning rate schedule to two or three parts:
1) Linear warmup
2) cosine to 1/10 of value
3) Fill in remaining steps with zero

Add option to have schedule longer than number of steps (schedule ends with value higher than final decay final), or shorter (learning rate of zero padded until end)

lr_schedule < steps:
image

lr_schedule > steps:
image


MaxText/max_utils.py
```diff
@@ -230,20 +230,51 @@ def setup_initial_state(model, tx, config, rng, mesh, checkpoint_manager):
...
+  """
+  def make_cos_schedule(init_lr, final_lr, len_steps):
+    def schedule(step):
+      pct = (step) / len_steps
+      a = 0.5 * (jnp.cos(jnp.pi*pct) + 1)
+      lr = init_lr * a + final_lr * (1 - a)
+      return lr
+    return schedule
+
+  steps = config.steps
```
lukaszlew, line 251: We need 2 params in config:
config.schedule_steps # this one is controlling when the LR schedule ends (goes to zero after cooldown)
config.run_steps # this one is controlling when the training should finish - if we want to finish early or if we want to train longer to get 0 LR for some more steps to get good final measurement.
If either one is -1, it should inherit value from the other one (just like in int8-sweeps) branch.

MaxText/configs/base.yml
```diff
@@ -111,8 +111,17 @@ max_corpus_chars: 10_000_000
...
+# 1) Linear warmup from 0 to [learning_rate]
+# 2) Cosine from [learning_rate] to [learning_rate * cosine_learning_rate_final_fraction]
+# 3) Linear cooldown from [learning_rate * cosine_learning_rate_final_fraction] to 0
+# 4) Constant learning rate of 0 
 learning_rate: 3.e-5
-warmup_steps: 15_000
+cosine_learning_rate_final_fraction: 0.1 
+warmup_steps_fraction: 0.1 # First schedule section length
+cosine_steps_fraction: 0.8 # Second schedule section length
+cooldown_steps_fraction: 0.05 # Third scedhule section length
```
lukaszlew, line 123: Let's keep cooldown at 0 default. and cosine at 0.9

rwitten:
I view this change as unnecessary -- I was planning for MaxText to replicate the Llama2 science for areas in which we don't have a preference. Is there a reason we specifically need this? Feel free to DM me as necessary.

rwitten (approved):
This has my approval but please also get Lukasz's

PR #140, opened 2023-09-08
Switch from adam to adamW

gobbleturk:


rwitten (commented):
This CL is great. But can we (1) validate on a 1B model that this doesn't create a loss regression and (2) cite Llama2 section 2.2 as our reference.

gobbleturk:
Added citation. Ran a 1B experiment and found AdamW had a ~1% better training loss than Adam

image

rwitten (approved):

PR #142, opened 2023-09-11
Enable Flash Attention

khatwanimohit:


khatwanimohit:
Loss curves:

image
image

MaxText/layers.py
```diff
@@ -429,16 +435,49 @@ def swap_dims(x):
...
-        query,
-        key,
-        value,
-        bias=attention_bias,
-        dropout_rng=dropout_rng,
-        dropout_rate=self.dropout_rate,
-        deterministic=deterministic,
-        dtype=self.dtype,
-        float32_logits=self.float32_logits)
+    if cfg.enable_flash_attention:
```
rwitten, line 438: Can we encapsulate this switch logic into a function?

rwitten (commented):
Some nits but the main very serious issue is convergence.

rwitten:
Convergence is so bad that I wonder if we have flash attention being applied significantly incorrectly.

khatwanimohit:
image

rwitten (approved):
LGTM thanks

PR #148, opened 2023-09-12
Run MaxText with stable dependencies for 2 steps in UnitTest.yml

khatwanimohit:


.github/workflows/UnitTests.yml
```diff
@@ -46,5 +46,6 @@ jobs:
     - name: Test with pytest
       run: |
         source venv/bin/activate
+        python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset steps=2
```
rwitten, line 49: Shouldn't this be a second test? To make it clearer what is failing?

rwitten (changes requested):
Quick nit.

.github/workflows/UnitTests.yml
khatwanimohit: Agreed, added unit tests for decode as well

rwitten (approved):

PR #149, opened 2023-09-12
Update main to latest int8 config

gobbleturk:
All files under MaxText/aqt/jax/v2 have been copied from the [int8-sweeps branch](https://github.com/google/maxtext/tree/int8-sweeps)

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -68,7 +68,7 @@ mlp_activations: ["relu"]
 dropout_rate: 0
 logits_via_embedding: True  # NOTE: this is True just for testing.
 # proj, minimal, full, or none
-remat_policy: 'minimal'
+remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
 enable_flash_attention: False # WARNING: Does not support attention_bias, requires JAX>=0.4.16
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -40,19 +40,20 @@ gcs_metrics: False
 # Activation dtypes.
 dtype: "bfloat16"
 int8_training: True
-# The following ten flags will only have effect when int8_training is True
+# The following flags will only have effect when int8_training is True
 use_dqdg: False
 fwd_int8: True
 dlhs_int8: True
-drhs_int8: True
-
+drhs_int8: False
 fwd_int8_qk: False
 dlhs_int8_qk: False
 drhs_int8_qk: False
-
-fwd_int8_pv: False
-dlhs_int8_pv: False
+fwd_int8_pv: True
+dlhs_int8_pv: True
 drhs_int8_pv: False
+aqt_use_dummy_static_bound: False
+aqt_use_fwd_quant: False
+aqt_rng_type: "jax.uniform"
 
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes 
 # then you should explicitly set base_embed_dim, base_num_heads, base_mlp_dim, base_num_decoder_layers and/or head_dim.
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -39,9 +39,20 @@ gcs_metrics: False
 
 # Activation dtypes.
 dtype: "bfloat16"
-int8_training: False
+int8_training: True
+# The following ten flags will only have effect when int8_training is True
+use_dqdg: False
 fwd_int8: True
-bwd_int8: True
+dlhs_int8: True
+drhs_int8: True
+
+fwd_int8_qk: False
+dlhs_int8_qk: False
+drhs_int8_qk: False
+
+fwd_int8_pv: False
+dlhs_int8_pv: False
+drhs_int8_pv: False
 
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes 
 # then you should explicitly set base_embed_dim, base_num_heads, base_mlp_dim, base_num_decoder_layers and/or head_dim.
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -39,7 +39,7 @@ gcs_metrics: False
 
 # Activation dtypes.
 dtype: "bfloat16"
-int8_training: True
+int8_training: False
 # The following flags will only have effect when int8_training is True
 use_dqdg: False
 fwd_int8: True
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -348,7 +348,7 @@ class MultiHeadDotProductAttention(nn.Module):
           deterministic=deterministic,
           dtype=self.dtype,
           float32_logits=self.float32_logits,
-          cfg=cfg)
+          cfg=self.config)
     return x
 
   @nn.compact
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -136,6 +136,24 @@ def dot_product_attention(query: Array,
       attn_weights = jnp.einsum('bqhd,bkhd->bhqk', query, key, _dot_general=aqt_dot_general)
     return attn_weights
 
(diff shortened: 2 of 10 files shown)


MaxText/configs/base.yml
```diff
@@ -40,8 +40,20 @@ gcs_metrics: False
...
+dlhs_int8_qk: False
+drhs_int8_qk: False
+fwd_int8_pv: True
+dlhs_int8_pv: True
+drhs_int8_pv: False
+aqt_use_dummy_static_bound: False
+aqt_use_fwd_quant: False
+aqt_rng_type: "jax.uniform"
 
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes
```
lukaszlew, line 58: These parameters were different in our int8 runs:
- remat_policy
- per_device_batch_size
- max_target_length
- enable_dropout

Also, we have `config.global_batch_size_to_load` now.

What should we do about it?

MaxText/layers.py
```diff
@@ -123,8 +125,15 @@ def dot_product_attention(query: Array,
   query = LayerNorm(dtype=dtype, name='query_layer_norm', kernel_axes = ('heads',))(query)
   key = LayerNorm(dtype=dtype, name='key_layer_norm', kernel_axes = ('heads',))(key)
 
-  # `attn_weights`: [batch, num_heads, q_length, kv_length]
-  attn_weights = jnp.einsum('bqhd,bkhd->bhqk', query, key)
+  # QK Product, a.k.a `attn_weights`: [batch, num_heads, q_length, kv_length]
+  if not cfg.int8_training:
+    attn_weights = jnp.einsum('bqhd,bkhd->bhqk', query, key)
+  else: 
+    aqt_cfg = maxtext_sweeps.sweep1(cfg.fwd_int8_qk, cfg.dlhs_int8_qk, cfg.drhs_int8_qk, use_fwd_quant=cfg.aqt_use_fwd_quant)
```
lukaszlew, line 132: Why are you not passing 
      use_dummy_static_bound=cfg.aqt_use_dummy_static_bound,
      rng_type=cfg.aqt_rng_type,

lukaszlew (changes requested):

MaxText/configs/base.yml
gobbleturk: I don't see the `global_batch_size_to_load` parameter you're referring to in int8-sweeps base.yml, what are you referring to?

gobbleturk: Remat policy: Our best medium/large configs do use remat policy of full, and since int8 works much better with remat full I think we should change to the default to full, wdyt @rwitten?

per_device_batch_size: This is expected to change with each model, the default here is optimized for the default size of 1B. Users should expect (or will quickly learn) to decrease it as the model size increases. We publish optimized configs for larger models with chosen batch sizes in MaxText/configs, e.g. [64b.sh](https://github.com/google/maxtext/blob/main/MaxText/configs/64b.sh) which uses this [config](https://github.com/google/maxtext/blob/main/end_to_end/test_tflops_64b_params.sh)

max_target_length: This was changed when flash attention was added, I believe we should keep it at 2k moving forward. Again we publish optimized configs of various sizes where we explicitly set this parameter.

enable_dropout: I changed this to false for int8 so we had deterministic runs, but the dropout_rate is zero by default as well so its also deterministic when true until you change the dropout_rate. I don't think the default value is particularly important here

MaxText/layers.py
gobbleturk: Great catch! Fixed in both places it was missing.

MaxText/configs/base.yml
rwitten: Yes lets move everything to full remat.

MaxText/layers.py
```diff
@@ -123,8 +127,22 @@ def dot_product_attention(query: Array,
   query = LayerNorm(dtype=dtype, name='query_layer_norm', kernel_axes = ('heads',))(query)
   key = LayerNorm(dtype=dtype, name='key_layer_norm', kernel_axes = ('heads',))(key)
 
-  # `attn_weights`: [batch, num_heads, q_length, kv_length]
-  attn_weights = jnp.einsum('bqhd,bkhd->bhqk', query, key)
+  # QK Product, a.k.a `attn_weights`: [batch, num_heads, q_length, kv_length]
+  if not cfg.int8_training:
+    attn_weights = jnp.einsum('bqhd,bkhd->bhqk', query, key)
+  else:
+    aqt_cfg = maxtext_sweeps.sweep1(
```
rwitten, line 134: I don't know what sweep1 is but I have a feeling it shouldn't be here?

Also this might conflict with FlashAttention in which you need to rebase on top of that.

rwitten: Finally can this be encapsulated in a function that computes attention? Because we're scattering the chaos when folks try to read the code.

MaxText/layers.py
```diff
@@ -146,8 +164,20 @@ def dot_product_attention(query: Array,
     attn_weights = attn_weights * multiplier
 
   # Take the linear combination of `value`.
-  return jnp.einsum('bhqk,bkhd->bqhd', attn_weights, value)
-
+  if not cfg.int8_training:
+    return jnp.einsum('bhqk,bkhd->bqhd', attn_weights, value)
+  else:
+    aqt_cfg = maxtext_sweeps.sweep1(cfg.fwd_int8_pv,
```
rwitten, line 170: sweep1? again let's encapsulate

MaxText/layers.py
```diff
@@ -229,12 +259,22 @@ def __call__(self, inputs: Array) -> Array:
...
       return lax.dot_general(inputs, kernel, ((axis, contract_ind), ((), ())))
     else:
-      aqt_cfg = maxtext_sweeps.sweep1(cfg.fwd_int8, cfg.bwd_int8)
-      aqt_dot_general = aqt.make_dot_general(aqt_cfg)
       aqt_key = self.make_rng('aqt')
-      context = aqt.Context(key=aqt_key, train_step=None)
-
-      return aqt_dot_general(inputs, kernel, ((axis, contract_ind), ((), ())), context=context)
+      if cfg.use_dqdg:
+        aqt_dq_dg = aqt_dq.make_aqt_dq_dg()
```
rwitten, line 264: same same

MaxText/layers.py
gobbleturk: Renamed sweep1 to `quantization_config`. However I couldn't think of a quick way to better encapsulate the aqt computation -- possibly in the future we can have an aqt einsum wrapper that takes in
* The einsum equation (e.g. ab, bc -> ac)
* The inputs
* The aqt config
But that takes some refactoring and is unclear to me its any simpler than what exists

gobbleturk: After discussing with @rwitten offline we have moved each int8 einsum and dot general into their own methods so users may instead simply read only the method invocation instead of the method definition.

Future work may include create a parametrized structure of aqt_config options (a dict hierarchy in base.yml) so that the three aqt configs can be created from a single method

MaxText/layers.py
```diff
@@ -113,6 +117,43 @@ def dot_product_attention(query: Array,
   assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
   assert query.shape[-1] == key.shape[-1], 'q, k depths must match.'
 
+  def compute_qk_attn_weights(query, key, cfg, aqt_rng):
```
lukaszlew, line 120: These 2 functions are almost identical, factor out common code?
You can put cfg.fwd_int8_qk and alike into params.

lukaszlew (approved):

MaxText/layers.py
rwitten: I actually don't mind this duplication. This hides complexity fully from the reader and they might diverge more in the future. Unless you feel strongly Lukasz, let's say this is ok.

rwitten (approved):

PR #151, opened 2023-09-13
Pass mesh to Transformer in decode.py

gobbleturk:
Need to pass mesh to Transformer in decode.py after changing Transfomer init API in this [PR](https://github.com/google/maxtext/pull/142)


khatwanimohit (approved):
LGTM

rwitten (approved):
Looks good but how do we catch this earlier? Go off thread.

PR #152, opened 2023-09-13
Fix sizes to match main

gobbleturk:
Change default sizes to match main branch


gobbleturk:
Added by https://github.com/google/maxtext/pull/156

PR #153, opened 2023-09-13
Fix comment to match order of axis doublings

gobbleturk:
Recently we [changed](https://github.com/google/maxtext/pull/136) for the order of doubling axis to achieve global_parameter_scale, but we need to update the method comment to match this


rwitten (approved):
You're a hero. My bad!

PR #161, opened 2023-09-18
fix ici_fsdp_parallelism in pedagogical example of shardings.py

yashs97:


gobbleturk:
I think we either want the user to have to set the shardings intentionally here (so an error for the default configuration is fine) or we should choose defaults that match train.py which is [here](https://github.com/google/maxtext/blob/main/MaxText/max_utils.py#L96C17-L96C17)

rwitten (approved):

PR #162, opened 2023-09-18
Add tests for profiling

yashs97:


MaxText/pytest.ini
```diff
@@ -3,4 +3,4 @@
 testpaths =
     tests
 python_files = *_test.py
-addopts = -rf --import-mode=importlib --ignore=tests/input_pipeline_test.py
\ No newline at end of file
+addopts = -rf --import-mode=importlib --ignore=tests/input_pipeline_test.py --ignore=tests/profiler_tests.py
```
rwitten, line 6: Why? Let's match style unless there is a problem.

rwitten (changes requested):

MaxText/pytest.ini
yashs97: Do we need to collect a profile and run tests when there's a change in MaxText? 
From my understanding, the point of these tests is to run them when there a change in the profiler code. The reason why I think they should exist in MaxText is because it is a package that our team controls and easy to run on a TPU VM.

rwitten (approved):

PR #163, opened 2023-09-19
Int8 sweeps logits lax parens

gobbleturk:
Initial logits PR

### DIFF
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -706,7 +706,7 @@ class Embed(nn.Module):
     """
     dtype = self.attend_dtype if self.attend_dtype is not None else self.dtype
     if not self.config.int8_training:
-      return maxtext_dot(query, jnp.asarray(self.embedding, dtype).T, lax.dot_general)
+      return maxtext_dot(query, jnp.asarray(self.embedding, dtype).T)
     else:
       aqt_cfg = maxtext_sweeps.sweep1(
         self.config.fwd_int8_logits,
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -709,9 +709,9 @@ class Embed(nn.Module):
       return maxtext_dot(query, jnp.asarray(self.embedding, dtype).T, lax.dot_general)
     else:
       aqt_cfg = maxtext_sweeps.sweep1(
-        self.config.fwd_int8_qk,
-        self.config.dlhs_int8_qk,
-        self.config.drhs_int8_qk,
+        self.config.fwd_int8_logits,
+        self.config.dlhs_int8_logits,
+        self.config.drhs_int8_logits,
         use_dummy_static_bound=self.config.aqt_use_dummy_static_bound,
         rng_type=self.config.aqt_rng_type,
         use_fwd_quant=self.config.aqt_use_fwd_quant,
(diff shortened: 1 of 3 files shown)


MaxText/layers.py
```diff
@@ -704,8 +705,26 @@ def attend(self, query: Array) -> Array:
       in NLP models.
     """
     dtype = self.attend_dtype if self.attend_dtype is not None else self.dtype
-    return jnp.dot(query, jnp.asarray(self.embedding, dtype).T)
```
lukaszlew, line 707: I think it is worthwhile to just fork jnp.dot and add _dot_general parameter to it. That will make the review easier and will be easier to go back to jnp.dot once _dot_general param lands in jax

MaxText/layers.py
```diff
@@ -704,8 +705,26 @@ def attend(self, query: Array) -> Array:
...
-
+    if not self.config.int8_training:
+      #return jnp.dot(query, jnp.asarray(self.embedding, dtype).T)
+      return lax.dot_general(query, jnp.asarray(self.embedding, dtype).T, (((query.ndim - 1,),(0,)),((),())))
+    else:
+      aqt_cfg = maxtext_sweeps.sweep1(
+        self.config.fwd_int8_qk,
```
lukaszlew, line 713: why qk? why not _logits

MaxText/layers.py
```diff
@@ -704,8 +705,26 @@ def attend(self, query: Array) -> Array:
...
+        use_fwd_quant=self.config.aqt_use_fwd_quant,
+      )
+      aqt_dot_general = aqt.make_dot_general(aqt_cfg)
+      aqt_key = self.make_rng('aqt')
+      context = aqt.Context(key=aqt_key, train_step=None)
+      aqt_dot_general = functools.partial(aqt_dot_general, context=context)
+      # !!! AQT requires LHS dtype equals RHS dtype
+      # return aqt_dot_general(query, jnp.asarray(self.embedding).T, (((query.ndim - 1,),(0,)),((),())))
+      # The below fails with jnp.bfloat16
+      return aqt_dot_general(jnp.asarray(query, jnp.float32), jnp.asarray(self.embedding, jnp.float32).T, (((query.ndim - 1,),(0,)),((),())))
```
lukaszlew, line 727: We need to find common type, if both are f32, we are good, if bothr are bf16 we are good.
If they are bf16+f32 we need to cast both to f32

lukaszlew (changes requested):

MaxText/layers.py
gobbleturk: Fixed.

MaxText/layers.py
gobbleturk: Done.

MaxText/layers.py
gobbleturk: Done.

MaxText/my_jnp_dot.py
```diff
@@ -0,0 +1,35 @@
+from jax._src.typing import Array, ArrayLike, DuckTypedArray, DTypeLike, Shape
+from typing import (Any, Callable, Optional, TypeVar, Union,
+                    cast as type_cast, overload)
+
+def dot(lhs, rhs, _dot_general, precision = None,
```
lukaszlew, line 5: _dot_general = lax.dot_general default argument would work well

lukaszlew (approved):

MaxText/my_jnp_dot.py
gobbleturk: Done.

PR #165, opened 2023-09-19
Fix profiler for shardings.py 

khatwanimohit:
Tested on stable `JAX==0.4.16` and nightly `0.4.17.dev20230919`

### DIFF
--- a/pedagogical_examples/shardings.py
+++ b/pedagogical_examples/shardings.py
@@ -240,11 +240,11 @@ pjit_gen_layers = pjit(
         out_shardings=parameter_sharding
       )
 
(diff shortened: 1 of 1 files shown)


yashs97 (approved):

rwitten (changes requested):
I need some context! Why is this change needed? What problem does it solve?

In the past we put it where we put it because we didn't want to profile the compile.

khatwanimohit:
> What problem does it solve?

Currently profiling throws a segmentation fault if start_trace is called within `with` statement

rwitten (approved):
THanks!

PR #166, opened 2023-09-19
Enable Autocheckpoint and exit after saving an autocheckpoint

xuefgu:
This PR does two things:

1) Enable Autocheckpoint in MaxText. Along the way, the checkpoint manager is initialized in a more canonical way, with the option `save_interval_steps` passed as an option during creation.

2) Make the program exit after the checkpoint has been saved.

I have tested this PR to make sure that the full functionality of Autocheckpoint works as desired on two v4-8: a checkpointing process is automatically started upon a maintenance event, the program exits after saving the checkpoint, and the TPU goes through maintenance thereafter.

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -51,21 +51,32 @@ def _multislice_distribute_initialize():
                              num_processes=jax.process_count(),
                              process_id=jax.process_index())
 
-def create_orbax_checkpoint_manager(checkpoint_dir: str, enable_checkpointing: bool, use_async: bool):
+def create_orbax_checkpoint_manager(
+    checkpoint_dir: str,
+    enable_checkpointing: bool,
+    use_async: bool,
+    save_interval_steps: int
+):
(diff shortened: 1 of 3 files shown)


gobbleturk (approved):
This looks great to me, thank you for adding this!

rwitten (approved):
This looks great to me and see Matt has already approved.

I assume you already have it but has Colin Gaffney confirmed (I don't need an approval in the system).

Has this been tested? (Or are we comfortable with our XLML tests -- I'll let you and Matt decide.)

xuefgu:
> This looks great to me and see Matt has already approved.
> 
> I assume you already have it but has Colin Gaffney confirmed (I don't need an approval in the system).
> 
> Has this been tested? (Or are we comfortable with our XLML tests -- I'll let you and Matt decide.)

xuefgu:
Colin has reviewed this PR.

Matt and I confirmed that this should not change XLML tests behavior.

PR #167, opened 2023-09-19
segment_ids in flash attention

khatwanimohit:
* Adding segment_ids in flash_attention 
* add unit test for attention

### DIFF
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -23,7 +23,6 @@ from aqt.jax.v2 import aqt_dq_dot_general as aqt_dq
 from aqt.jax.v2.google import aqt_config
 from jax.experimental.shard_map import shard_map
 from jax.sharding import Mesh
-from jax.sharding import PartitionSpec as P
 
 import functools
 import operator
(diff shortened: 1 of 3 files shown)


MaxText/layers.py
```diff
@@ -328,40 +326,49 @@ class MultiHeadDotProductAttention(nn.Module):
 
 
   def apply_attention(self, query, key, value, enable_flash_attention,
-                      attention_bias, dropout_rng, deterministic):
+                      decoder_segment_ids, attention_bias, dropout_rng, deterministic, decode):
     """ Apply Attention
     """
-    if enable_flash_attention:
+    if enable_flash_attention and not decode:
```
rwitten, line 332: "If enable_flash_attention and decode" should be asserting, probably here if it is unsupported.

MaxText/tests/attention_test.py
```diff
@@ -0,0 +1,107 @@
...
+from jax.sharding import Mesh
+from layers import MultiHeadDotProductAttention
+from jax.sharding import PartitionSpec as P
+import numpy as np
+
+import pyconfig
+import sys
+
+class AttentionTest(unittest.TestCase):
+  """Test for the RoPE implementation """
```
rwitten, line 31: RoPE?

rwitten (approved):
Broadly looks good just some little notes.

MaxText/layers.py
khatwanimohit: Yes, decode is not supported with flash_attention

MaxText/tests/attention_test.py
khatwanimohit: fixed

rwitten (approved):

PR #169, opened 2023-09-20
Refactor tokenizer

khatwanimohit:
- Remove training from tokenizer.py and create train_tokenizer.py
- only load tokenizer in input_pipeline.py
- default path for tokenizer is `assets/tokenizer`


rwitten (commented):
Missing tests! We should be able to:
(1) create a new tokenizer file
(2) use that new tokenizer file
(3) use the default tokenizer file

(If you add an assert when doing (1) that it matches the version in source control you don't need to test (2) and (3) separately)

rwitten (approved):

PR #171, opened 2023-09-22
Hardcode gcs IPs

gobbleturk:
For a large number of hosts we have seen errors reading checkpoints.

The error occurs when looking up the GCS IP using DNS. This fixes the issue by hardcoding the GCS IP.


rwitten (changes requested):
What scale are you seeing this at? We have a bug working on fixing this but we can't merge this code because it is totally unmaintainable -- those IPs are subject to change.

gobbleturk:
I've seen this error on 8 pods, whereas 4 pods works fine

PR #172, opened 2023-09-22
Add 1b convergence test

gobbleturk:
Add a 1B convergence test config

### DIFF
--- a/end_to_end/test_convergence_1b_params.sh
+++ b/end_to_end/test_convergence_1b_params.sh
@@ -2,7 +2,7 @@
 set -e
 
 USER=${1}
-TFLOP_THRESHOLD=${2}
+LOSS_THRESHOLD=${2}
 OUTPUT_PATH=${3}
 DATASET_PATH=${4}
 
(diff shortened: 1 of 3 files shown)


rwitten (commented):
I'm a little confused -- what size machine is this test supposed to run on? What loss is it supposed to achieve? Can you document that so subsequent people have a clear benchmark to run against? (Alternatively, your first user can do that but we shouldn't merge this CR until they have.)

(For example, this is only Chinchilla if I assume something about the machine it is running on, right?)

end_to_end/test_convergence_1b_params.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ -z ${5} ]
+then 
+    RUN_NAME=${USER}
+else
+    RUN_NAME=${5}
+fi
+
+# Train
+export LIBTPU_INIT_ARGS="--xla_tpu_enable_data_parallel_all_reduce_opt=true --xla_tpu_data_parallel_opt_different_sized_ops=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true --xla_tpu_overlap_compute_collective_tc=true --xla_enable_async_all_gather=true"
+python3 MaxText/train.py MaxText/configs/base.yml run_name=$RUN_NAME\
```
rwitten, line 19: Also does this work? Isn't our loss ramp set to 15000 steps by default?
https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml#L130

rwitten (changes requested):

rwitten (approved):

PR #175, opened 2023-09-25
Reshard train state to checkpoint efficiently with multiple slices

gobbleturk:


MaxText/train.py
```diff
@@ -282,10 +283,14 @@ def train_loop(config, state=None):
     last_step_completion = new_time
 
     if checkpoint_manager is not None:
-      if checkpoint_manager.save(step, state):
```
rwitten, line 285: this code is more complex and not any better than just writing the normal way -- sharding should be ~free.

rwitten (changes requested):

gobbleturk:
This solution does not work, probably because it is too many QPS to GCS still since each host talks to GCS

PR #178, opened 2023-09-26
Add T5X xent and unit test

raymondzouu:
Merge in @rwitten's T5X cross entropy implementation along with unit test. 

Tested the change on the 128B config on 16 x v5e-256 and it increased TFLOP/s by ~10 TFLOP/s and increased memory by 131.48 MiB.

Tested convergence on the 1B config on 1 x v5e-256 for 3400 steps. The achieved loss is about the same.

new t5x xent: (last_10_losses_mean=2.637)
image
original optax xent: (last_10_losses_mean=2.641)
image

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -247,3 +247,80 @@ def create_learning_rate_schedule(learning_rate: float, warmup_steps: int):
       rsqrt_schedule(init_value=learning_rate, shift=warmup_steps),
       ], boundaries=[warmup_steps],
       )
+
+from typing import Tuple, Mapping, Optional, Union
+
+
+
+@jax.custom_vjp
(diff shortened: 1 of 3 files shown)


MaxText/max_utils.py
```diff
@@ -274,4 +275,77 @@ def schedule(step):
     boundaries.append(warmup_steps + cos_steps + constant_zero_steps)
 
   return optax.join_schedules(pieces, boundaries)
-  
\ No newline at end of file
+
+
+@jax.custom_vjp
```
rwitten, line 280: Just make a note that this is from original t5x code as a ncomment. Looks great otherwise.

rwitten (changes requested):

MaxText/max_utils.py
raymondzouu: Done.

rwitten (approved):

PR #187, opened 2023-09-29
Loading real data on subset of hosts 

khatwanimohit:


rwitten (commented):
Just unassigning myself. We will talk live to discuss landing this CR.

MaxText/pyconfig.py
```diff
@@ -317,12 +317,16 @@ def get_individual_scales(scale):
 def calculate_global_batch_sizes(raw_keys):
   """ Calculates target global batch size from target devices and per_device_batch"""
   per_device_batch_size = raw_keys['per_device_batch_size']
+  expansion_factor_real_data = raw_keys['expansion_factor_real_data']
   num_devices = get_num_target_devices(raw_keys)
   if per_device_batch_size < 1.0:
     # For per_device_batch_size<1, we load the data as if per_device_batch_size=1
     global_batch_size_to_load = num_devices
```
rwitten, line 324: But couldn't fewer than that number of hosts load the data?

rwitten: Like I think we should still ramp the data?

MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -121,16 +119,72 @@ def raw_generate_synthetic_data(config):
...
+  return process_loading_real_data
+
+def make_mixed_train_iterator_and_tokenizer(config, mesh, add_bos, add_eos):
+  process_indices = get_process_loading_real_data(config, mesh)
+  print(len(process_indices),"hosts out of ",jax.process_count()," are loading real data")
+  if jax.process_index() in process_indices:
+    if config.dataset_type == "c4":
+      return make_c4_train_iterator_and_tokenizer(config, mesh, add_bos, add_eos)
+    elif config.dataset_type == "c4-array_record":
+      return make_grain_train_iterator_and_tokenizer(config, mesh, add_bos, add_eos)
```
rwitten, line 176: don't we need to pass in how many shards are really loading data? The data iterators care about this!

MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -121,16 +119,72 @@ def raw_generate_synthetic_data(config):
...
+    elif config.dataset_type == "c4_mlperf":
+      print("Overwrite both add_bos and add_eos to False")
+      return make_c4_mlperf_train_iterator_and_tokenizer(config, mesh, add_bos=False, add_eos=False)
```
rwitten, line 179: p.s. -- why is tokenizer coming from the dataset? I think that was a mistake in hindsight.

rwitten (changes requested):

MaxText/input_pipeline/input_pipeline_interface.py
khatwanimohit: Done

MaxText/input_pipeline/input_pipeline_interface.py
khatwanimohit: Will take care of this in another PR

rwitten (approved):
This has my approval but please discuss with Aireen and Roshani before merging.

@aireenmei @RoshaniN -- this CR changes it so MaxText can make balancing decisions about the number of hosts that read from GCS. We find in practice this is a useful lever because the thundering horde of VMs can crush GCS but the aggregate data isn't too much.

aireenmei:
Thanks for adding me to the thread. I think I missed some context so not sure I understand the whole idea. I see we only have a subset of host loading data, are they going to pass the data to hosts that are not loading real data? Why the rest of hosts are returning synthetic data?

RoshaniN:
I don't see any issues as standalone_dataloader would be using the same input pipeline as train. I would like to understand the recommendations on the expansion_factor_real_data, will do that offline.

aireenmei:
Could you share some convergence results when expansion_factor_real_data != -1 ?

khatwanimohit:
> Could you share some convergence results when expansion_factor_real_data != -1 ?

Convergence test with expansion_factor_real_data=4 (i.e. 16 hosts out of 64 hosts will load the real data)
https://cloudlogging.app.goo.gl/DPDSXu2tSM3ga8hG8

aireenmei:
Thanks! Could you also run a convergence test with grain? `bash end_to_end/test_convergence_1b_params.sh DATASET_TYPE="c4-array_record" ...`

khatwanimohit:
> Thanks! Could you also run a convergence test with grain? `bash end_to_end/test_convergence_1b_params.sh DATASET_TYPE="c4-array_record" ...`

Convergence run with c4-array_record: https://cloudlogging.app.goo.gl/jiFMzAx8SDRw4nM46

@aireenmei  I will also add airflow tests for both the convergence test

MaxText/configs/base.yml
```diff
@@ -164,6 +164,7 @@ dataset_name: 'c4/en:3.0.1'
 eval_dataset_name: 'c4/en:3.0.1'
 eval_split: 'validation'
 per_device_batch_size: 12.0
+expansion_factor_real_data: -1 # if -1 then all hosts will load real data
```
RoshaniN, line 167: Nit: Could you detail this comment to say something like "-1 then all hosts will load real data >0 subset of hosts load real data and share with other hosts"?
This would help future users understand when to enable this feature.

MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -121,16 +125,74 @@ def raw_generate_synthetic_data(config):
                                                     dtype=jax.numpy.int32)
     return output
 
+class BadSyntheticDataIterator():
+  """Creates a Dummy synthetic data iterator for loading on subset of hosts"""
```
RoshaniN, line 129: Nit: I got to know recently - go/piper-pi-terms-list#avoid-dummy

MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -121,16 +125,74 @@ def raw_generate_synthetic_data(config):
...
+    output['targets'] = jax.numpy.full( (config.global_batch_size_to_load,
+                                          config.max_target_length), -1, dtype=jax.numpy.int32)
+    output['targets_position'] = jax.numpy.full( (config.global_batch_size_to_load,
+                                                  config.max_target_length), -1, dtype=jax.numpy.int32)
+    output['targets_segmentation'] = jax.numpy.full( (config.global_batch_size_to_load,
+                                                      config.max_target_length), -1, dtype=jax.numpy.int32)
+    return output
+
+def get_process_loading_real_data(config, mesh):
+  sharding = jax.sharding.NamedSharding(mesh, P(*config.data_sharding))
```
RoshaniN, line 166: Nit: please add header strings for functions in this class.

RoshaniN (approved):
Thanks Mohit!

aireenmei (approved):
Thanks Mohit!

PR #188, opened 2023-09-29
Use `jax.jit` in Maxtext and remove `pjit`. Do this by converting PartitionSpec to NamedShardings.

michelle-yooh:
sync based on cl/568965332
I'm temporarily adding this cl manually until copybara is set up.


rwitten (approved):
Can you search for all pjits and make sure they don't exist in the code any more at all?

PR #191, opened 2023-09-29
Read checkpoint steps from GCS only once, broadcast to rest

gobbleturk:
Tested on 2x VLP


rwitten (commented):
Thank you for the heroic effort here! I cherrypicked this CR in and I'm unassigning myself while you decide with Colin how to land it.

PR #194, opened 2023-10-03
Add support for installing nightly version of`tensorboard_plugin_profile`

yashs97:
Helpful for ensuring that a profile collected with nightlies work with `tbp-nightly`

### DIFF
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -25,7 +25,6 @@ from jax.experimental.shard_map import shard_map
 from jax.sharding import Mesh
 from jax.sharding import PartitionSpec as P
 
-import dataclasses
 import functools
 import operator
 from typing import Any, Callable, Iterable, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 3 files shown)


bvandermoon:
I'm generally ok with this change. But just wondering if there is a need for it? For example, is it helpful for your workflow? Is it true that JAX nightly users also want tbp-nightly?

bvandermoon (commented):
I'm generally ok with this change. But just wondering if there is a need for it? For example, is it helpful for your workflow? Is it true that JAX nightly users also want tbp-nightly?

shahyash10:
> I'm generally ok with this change. But just wondering if there is a need for it? For example, is it helpful for your workflow? Is it true that JAX nightly users also want tbp-nightly?

It is helpful for the nightly XLML tests and ensures that a profile collected with nightly version of JAX works with nightly version of `tbp`

bvandermoon:
> > I'm generally ok with this change. But just wondering if there is a need for it? For example, is it helpful for your workflow? Is it true that JAX nightly users also want tbp-nightly?
> 
> It is helpful for the nightly XLML tests and ensures that a profile collected with nightly version of JAX works with nightly version of `tbp`

Got it, makes sense. Thanks Yash. Please add that to the description for context

bvandermoon (approved):

rwitten (changes requested):
you should uninstall ahead of time like we do for Jax/Jaxlib. That way we can change.

rwitten (approved):

PR #197, opened 2023-10-04
Add a flag in multihost_job.py to enable Autocheckpoint

xuefgu:
This change allows the users to launch a production-ready MaxText job with the Autocheckpoint feature enabled on Cloud TPU.

It is tested by first launching a job with:
```
NODE_COUNT=2
python3 multihost_job.py --NUM_SLICES=$NODE_COUNT --RUN_NAME=$RUN_NAME --TPU_TYPE=v4-8 --VERSION=tpu-ubuntu2204-base --BUCKET_NAME=$BUCKET_NAME --ENABLE_AUTOCHECKPOINT=true \
--COMMAND="bash setup.sh && python3 MaxText/train.py MaxText/configs/base.yml run_name=$RUN_NAME global_parameter_scale=2 per_device_batch_size=6 save_period=50 base_output_directory=$OUTPUT_DIR dataset_path=$DATASET_PATH"
```
and then observing that the checkpoint does get saved upon a disruption.

### DIFF
--- a/multihost_job.py
+++ b/multihost_job.py
@@ -75,17 +75,18 @@ def normalize_gcs_bucket_name(args):
 def print_flags(args):
   """ Print configuration values after defaults have been filled in. """
   print("Running multihost_job with the following configuration:")
-  print(f"Project             (--PROJECT)         = {args.PROJECT}")
-  print(f"Zone                (--ZONE)            = {args.ZONE}")
-  print(f"TPU type            (--TPU_TYPE)        = {args.TPU_TYPE}")
-  print(f"TPU runtime version (--VERSION)         = {args.VERSION}")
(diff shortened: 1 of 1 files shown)


rwitten (changes requested):
Do we need this. Isn't this covered by --CQR_EXTRA_ARGS?

xuefgu:
> Do we need this. Isn't this covered by --CQR_EXTRA_ARGS?

We could have the user pass the `--autocheckpoint-enabled` flag in `--CQR_EXTRA_ARGS`. This PR is to make MaxText support the feature from the top-level, without the user having to look into how to enable the feature in the CQR request. I can see the pros/cons both ways.

rwitten (commented):
If we want to merge this, let's do it when ENABLE_AUTOCHECKPOINT is ready to be true by default.

rwitten (approved):

PR #200, opened 2023-10-09
Add support to XPK for v4

yashs97:
XPK should support TPU v4, adding support for all slice listed at https://cloud.google.com/tpu/docs/supported-tpu-configurations#large-topologies.


Obliviour (approved):
Thank you for making this change!!

Obliviour (approved):

xpk/xpk.py
```diff
@@ -1450,6 +1459,18 @@ def workload_name_type(value, pat=re.compile(r'[a-z]([-a-z0-9]*[a-z0-9])?')):
...
+  '--cluster-type',
+    type=str,
+    default='regional',
+    help='regional or zonal cluster type.'
+)
+cluster_create_optional_arguments.add_argument(
+  '--cluster-machine-type',
+    type=str,
+    default='e2-standard-32',
+    help='cluster machine type.'
```
Obliviour, line 1472: help='Set the machine tpu within the default cpu node pool. For zonal clusters, make sure that the zone supports the machine type, and for regional clusters, the all zones in the region supports the machine type.'

xpk/xpk.py
```diff
@@ -607,17 +613,23 @@ def get_zone():
   return zone_outputs[-1]  # The zone name lives on the last line of the output
 
 
-def zone_to_region(zone) -> str:
-  """Helper function converts zone name to region name.
+def get_zone_from_zone_or_region(name, cluster_type) -> str:
```
Obliviour, line 616: Hmm I considering what other names we can pick here:

I like to describe the function by its purpose. This function will return the command needed to create a zonal or regional cluster. I think `get_cluster_type_argument(zone_name, cluster_type)` or something similar is clear to me. Let me know what you think!

xpk/xpk.py
```diff
@@ -1450,6 +1459,18 @@ def workload_name_type(value, pat=re.compile(r'[a-z]([-a-z0-9]*[a-z0-9])?')):
     help='The number of slices to run the job on, defaults to 1.',
     required=True,
 )
+cluster_create_optional_arguments.add_argument(
+  '--cluster-type',
```
Obliviour, line 1463: --cluster-type also needs to be an argument for 'xpk cluster list'. Or actually maybe cluster list should show both zonal and regional clusters. That would be nicer?

I don't think there is a single argument for this, so the ask would be to run two commands one with --zone and one with --region in `xpk cluster list'

xpk/__init__.py
Obliviour, line 1: Can we add the License in this file as well? It sounds good to have?

Obliviour (commented):
Overall LGTM, I have some comments about the zonal cluster / regional cluster things. We basically want to make sure zonal and regional is supported in any command that uses that. I believe that is cluster create, cluster list, and cluster delete.

I think we 1) make cluster_type an argument in both cluster create and cluster delete. 

2) make cluster list show both zonal and regional clusters.

xpk/xpk.py
Obliviour: Also to reiterate, `cluster delete` will need the argument as well

MaxText/__init__.py
```diff
@@ -13,5 +13,3 @@
  See the License for the specific language governing permissions and
  limitations under the License.
  """
-
-# adhd
```
rwitten, line 17: <3

unit_test_and_lint.sh
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-pylint MaxText/
+python3 -m pylint $(git ls-files '*.py')
```
rwitten, line 17: Why was this change necessary?

xpk/xpk.py
```diff
@@ -607,17 +613,23 @@ def get_zone():
   return zone_outputs[-1]  # The zone name lives on the last line of the output
 
 
-def zone_to_region(zone) -> str:
```
rwitten, line 610: Why are we support zonal and regional clusters instead of just regional clusters? To me it adds complexity and removes sensible defaults. Is there a use case where we'd recommend the zonal cluster?

rwitten (changes requested):
There are a lot of nice improvements in this CR. However, why are we adding complexity to support zonal clusters?

xpk/xpk.py
yashs97: v4 isn't available in a lot of the zones like `us-central2-a` etc. I repeatedly ran into errors when trying to regional cluster for v4. The default however is still a regional cluster, I haven't changed any defaults.

unit_test_and_lint.sh
yashs97: It makes sure that pylint runs on all files in maxtext repo if you're using unit_test_and_lint.sh before committing code changes. Also this is how pylint.yml uses pylint, don't really see a reason why it should be different in 2 places in the same repo

rwitten: Seems fine to me! Just curious.

xpk/xpk.py
rwitten: Can we understand the root cause of the issues with the regional cluster? Not being available in a zone is no reason you shouldn't be able to make a regional cluster.

If we understand why this is necessary, we might decide we want to make this change. But I don't want to include  these zonal support without reason because it will permanently increase complexity.

xpk/xpk.py
```diff
@@ -244,6 +241,18 @@
...
+    ),
+    'v4-16': SystemCharacteristics(
+      '2x2x2', 2,'tpu-v4-podslice', 'ct4p-hightpu-4t', 4
+    ),
+    'v4-32': SystemCharacteristics(
+      '2x2x4', 4,'tpu-v4-podslice', 'ct4p-hightpu-4t', 4
+    ),
+    'v4-64': SystemCharacteristics(
+      '2x4x4', 8,'tpu-v4-podslice', 'ct4p-hightpu-4t', 4
+    ),
```
rwitten, line 255: Also if you're doing this, can you support all listed sizes?

rwitten (changes requested):
Little nit re: v4 and really need to answer why zonal is required.

xpk/xpk.py
yashs97: I think @Obliviour is better suited to answer this than me

Obliviour: The issue we saw was with creating the cpu node pool that needs cpus from all zones in the region for regional clusters.

only a small portion of cpu types are available in all the zones in the region and there are some potential quota limits.

rwitten: I had a conversation with Victor and I don't see this change as important because there is a workaround (use the compatible CPU instances). Let's keep the instance flag but get rid of this one.

rwitten (changes requested):

xpk/xpk.py
yashs97: removed zonal support flag!

Obliviour (approved):

rwitten (approved):

PR #201, opened 2023-10-09
[xpk] Use kueue github repo manifest file instead of local one

Obliviour:
TESTED=yes, ran xpk cluster create, and ran simple workloads that waited in the kueue.


Obliviour:
> Have you checked the diff between these files? I do think there is a diff...

Here is the diff.

```
diff kueue_manifests.yml kueue_manifests_to_upgrade_to.yml
9859,9861c9859,9860
<     waitForPodsReady:
<       enable: true
<       timeout: 20m
---
>     #waitForPodsReady:
>     #  enable: true
9877,9894d9875
< data:
<   controller_manager_config.yaml: |
<     apiVersion: config.kueue.x-k8s.io/v1beta1
<     kind: Configuration
<     namespace: kueue-system
<     health:
<       healthProbeBindAddress: :8081
<     metrics:
<       bindAddress: :8080
<     webhook:
<       port: 9443
<     waitForPodsReady:
<       enable: true
<       timeout: 20m
<     integrations:
<       frameworks:
<       - "batch/job"
<       - "jobset.x-k8s.io/jobset"
```

PR #202, opened 2023-10-09
add ici_tensor_parallelism=4 to test_decode

michelle-yooh:
This change is to fix the dimension issue in XLML test which was earlier seen in unit tests.  https://pantheon.corp.google.com/kubernetes/job/us-central2/xl-ml-test-1vm-us-central2/automated/mp-jax-nightly-maxtext-decode-func-v4-8-1-slices-6zhfz/logs?project=cloud-tpu-multipod-dev&mods=allow_workbench_image_override&e=13803378

### DIFF
--- a/end_to_end/test_decode.sh
+++ b/end_to_end/test_decode.sh
@@ -17,6 +17,7 @@ fi
 #Train
 python3 MaxText/decode.py MaxText/configs/base.yml run_name=$RUN_NAME\
     steps=50 enable_checkpointing=False metrics_file='metrics.txt'\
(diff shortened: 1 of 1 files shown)


gobbleturk:
@rwitten This is a fix required after https://github.com/google/maxtext/pull/193.

I believe that PR has a slight mis-speficiation of the sharding in [decode.py](https://github.com/google/maxtext/pull/193/files#diff-168aa6c3635f444399658f5a1e49418997bd9833f77a0a284d565920b1a0954e) after changing from the pjit to jit API - why does FSDP sharding no longer work and instead we must use TP?

PR #203, opened 2023-10-10
[xpk] Add deletion prompt and force flag on cluster create

Obliviour:
TESTING=yes

[y] checked that cluster create with deletion returns the prmpt [y] y / yes continues the prompt
[y] n / no stops the deletion
[y] other words stops the deletion
[y] force skips the prompt


Obliviour:
Apologies for doing one of the less useful ones (did this in the past before we talked)...

rwitten (approved):

PR #204, opened 2023-10-10
Fix installing jax and jaxlib from HEAD

yashs97:
- Fix installing jax and jaxlib from HEAD.
- Link JAX and custom libtpu binaries correctly
- Add `SKIP_REQUIREMENTS` flag (default=false) to setup.sh so that installing packages from requirements.txt aren't installed if running `shardings.py` from pedagogical examples

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -130,6 +130,13 @@ elif [[ $MODE == "nightly" ]]; then
     pip3 install tbp-nightly --upgrade
 elif [[ $MODE == "head" ]]; then 
 # Head mode
+    echo "Installing jax and jaxlib from head"
+    cd $HOME && git clone https://github.com/google/jax.git
+    cd $HOME && git clone https://github.com/openxla/xla.git
+    cd $HOME/jax
+    pip3 install numpy wheel build
+    python build/build.py --bazel_options=--override_repository=xla=$HOME/jaxlib
+    pip install dist/*.whl  # installs jaxlib (includes XLA)
     if [[ -n "$LIBTPU_GCS_PATH" ]]; then
         # Install custom libtpu
         echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -40,6 +42,10 @@ if [[ $LIBTPU_GCS_PATH == NONE ]]; then
   unset LIBTPU_GCS_PATH
 fi
 
+if [[ $SKIP_REQUIREMENTS == NONE ]]; then
+    SKIP_REQUIREMENTS="false"
```
rwitten, line 46: This code pattern doesn't look correct to me.  Because it leaves a state where SKIP_REQUIREMENTS can either be false or unset which seems undesirable? (The code does seem OK but it weird we expose so many options.)

setup.sh
```diff
@@ -130,36 +136,45 @@ elif [[ $MODE == "nightly" ]]; then
     pip3 install tbp-nightly --upgrade
 elif [[ $MODE == "head" ]]; then 
 # Head mode
+    echo "Installing jax and jaxlib from head"
```
rwitten, line 139: Why are you using this mode? Elsewhere we use nightly. (Tony and I were talking about getting rid of this.)

setup.sh
rwitten: But I like the feature a lot.

rwitten (changes requested):
You also need to update the docker packager:
https://github.com/google/maxtext/blob/main/maxtext_dependencies.Dockerfile

Pro-tip: send a message before a CR for things like this! Generally eager for improvements to shared machinery but discussion up front can save churn.

setup.sh
yashs97: done, following the same pattern as other args!

setup.sh
```diff
@@ -130,36 +136,45 @@ elif [[ $MODE == "nightly" ]]; then
...
+    if [ -d "$HOME/jax" ]; then
+        cd "$HOME/jax" && git pull origin main
+    else
```
rwitten, line 142: This part of the CR is not mergeable -- it is duplicative but different from the nightly case.

setup.sh
```diff
@@ -130,36 +136,45 @@ elif [[ $MODE == "nightly" ]]; then
...
+        cd $HOME && git clone https://github.com/openxla/xla.git
+    fi
+    cd $HOME/jax
+    pip3 install numpy wheel build
+    python3 build/build.py --enable_tpu --bazel_options=--override_repository=xla=$HOME/xla
+    pip3 install dist/*.whl --force-reinstall --no-deps # installs jaxlib (includes XLA)
+    pip3 install -e .  # installs jax
     if [[ -n "$LIBTPU_GCS_PATH" ]]; then
+        installed_libtpu_path=$(python3 -c '
+        import os; import jax; print(os.environ["TPU_LIBRARY_PATH"])
```
rwitten, line 157: This isn't mergeable because it ccnflicts wiht the other conventions.

rwitten (changes requested):
Nits related to conflicting with alternative flows.

setup.sh
yashs97: I researched this more and we need to do this differently from nightly because jaxlib lives in JAX and I need to link it to local openxla and pass in the `--enable-tpu` flag as well (https://jax.readthedocs.io/en/latest/developer.html#building-jaxlib-from-source-with-a-modified-xla-repository - the command in this link is being run from the jax directory)

rwitten: ok sounds good.

shahyash10:
closing this PR as head mode from setup.sh will be removed soon!

PR #219, opened 2023-10-20
Fix profiler tests

yashs97:
Removing a few `absltest` only methods that `unittest` doesn't have. `unittest` replaced `abslttest` a few PRs ago.


rwitten (approved):
Please be sure to squash as you merge.

PR #227, opened 2023-10-26
Enable preemption and add preemption instructions to readme

Obliviour:


xpk/README.md
```diff
@@ -147,6 +147,34 @@ cleanup with a `Cluster Delete`.
     xpk-test --tpu-type=v5litepod-16
     ```
 
+### Workload Priority and Preemption
+* Set the priority level of your workload with `--priority=LEVEL`
+
+  We have five priorities defined: [`verylow`, `low`, `medium`, `high`, `very-high-non-preempt`].
```
rwitten, line 153: should we also have a `very-high`? My feeling is that some people have workloads that they don't want to have preempted. So it would be helpful for workloads to be not preemptible (i.e. if high could have that property). It is less helpful for other workloads to not want to preempt.

rwitten (changes requested):
Broadly looks good but I don't think we want this type of non-preemption? What do you think?

(I will 100% be convinced by users telling us what we want. But users have talked to me about some workloads not being preempted not submitting workloads that are polite.)

Obliviour:
> Broadly looks good but I don't think we want this type of non-preemption? What do you think?
> 
> (I will 100% be convinced by users telling us what we want. But users have talked to me about some workloads not being preempted not submitting workloads that are polite.)

Talked offline, going with strict priority. No non-preemption for starters.

xpk/README.md
Obliviour: Changed to very-high

xpk/xpk.py
```diff
@@ -207,10 +210,10 @@
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: veryhigh
```
rwitten, line 210: Why is very-high and verylow different? Otherwise looks great.

rwitten (changes requested):

xpk/xpk.py
Obliviour: great catch thank you.

rwitten (approved):

PR #231, opened 2023-10-27
Add Llama style RoPE

A9isha:


rwitten (changes requested):
Looks good but I want to make sure that you're also deleting relpos / demonstrating this in the model. I think @khatwanimohit needs that urgently!

A9isha:
Thank you Rafi! I have replaced the existing RelativePositionBiases with the new RoPE implementation

MaxText/layers.py
```diff
@@ -440,6 +440,11 @@ def query_init(*args):
     key = projection(kernel_init=self.kernel_init, name='key')(inputs_kv)
     value = projection(kernel_init=self.kernel_init, name='value')(inputs_kv)
 
+    #Apply RoPE
+    query = LLaMARotaryEmbedding(embedding_dims = self.head_dim, name='query_rotary')(query)
```
khatwanimohit, line 444: Since we are sequence_packing our inputs, I think we should also pass `inputs_position`. You can get this information after you plumb it in the Transformer class from train.py

khatwanimohit (changes requested):

MaxText/layers.py
A9isha: Excellent suggestion - thanks Mohit! I made the change

MaxText/tests/llama_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+""" Tests for Llama """
+import jax
+import unittest
+import jax.numpy as jnp
+from typing import Tuple
+import layers
+import numpy as np
+
+jax.config.update('jax_platform_name', 'cpu')
```
khatwanimohit, line 25: Can you remove this and let your tests run on TPU. This config is global and all the other unit tests in MaxText assume they are running on TPU.

khatwanimohit (changes requested):

MaxText/tests/llama_test.py
```diff
@@ -0,0 +1,103 @@
...
+    freqs_cis = precompute_freqs_cis(dim_per_head, seq_len * 2)
+    freqs_cis = jnp.take(
+        freqs_cis, jnp.arange(seq_len, dtype=np.int32)[None, :], axis=0
+    )
+
+    llama_output = apply_rotary_emb(
+    jnp.asarray(x_q), jnp.asarray(x_k), freqs_cis
+    )
+
+    # Calculate RoPE embeddings from MaxText implementation
```
rwitten, line 93: BTW this is a VERY cool idea (have a real implementation and have a reference implementation and keep them locked together) but we need to explain the design. I.e. users should feel free to change the "real" implementation in MaxText as long as it passes our tests. But they shouldn't change the "reference" implementation.

MaxText/tests/llama_test.py
A9isha: As discussed, we have increased the rtol for the llama_test to 1e-02

MaxText/tests/llama_test.py
A9isha: Thank you Rafi! I added a description.

rwitten (approved):

khatwanimohit (approved):

PR #232, opened 2023-10-27
Improve README.md

parambole:
- Rephrase multihost_runner.py script section to highlight that current state of the directory will be copied to TPUVM
- Highlight the configs need to be set before running multihost_runner.py

#campground


parambole:
Created a new PR https://github.com/google/maxtext/pull/233

PR #234, opened 2023-10-30
adding debug-dump-gcs to dump hlo and upload to the gcs bucket

ssusie:
The flag sets the  XLA_FLAGS to dump to /tmp/xla_dump dir then uploads to provided GCS bucket. This however may overwrite the XLA_FLAGS passed by user through env file


xpk/xpk.py
```diff
@@ -473,6 +473,13 @@ def add_env_config(args):
               'environment, a value must be specified.'
           )
           env[variable] = os.environ[variable]
+
+  if args.debug_dump_gcs:
+    if 'XLA_FLAGS' in env:
+      # Caveat: might be overwriting something passed by user
+      print('overwriting XLA_FLAG from env-file')
```
Obliviour, line 480: Nit 'xpk_print'

xpk/xpk.py
```diff
@@ -1998,6 +2010,14 @@ def directory_path_type(value):
         " Defaults to 0."
     ),
 )
+workload_create_parser_optional_arguments.add_argument(
+    '--debug-dump-gcs',
+    type=str,
+    default=None,
+    help=(
+        "gcs bucket where debugging info like hlo dump is uploaded to"
```
Obliviour, line 2018: Nit to add an example of an argument to help guide the user. Also nit to add period. Ty!

xpk/xpk.py
```diff
@@ -1455,6 +1462,11 @@ def workload_create(args) -> int:
     xpk_exit(setup_docker_image_code)
 
   add_env_config(args)
+  if args.debug_dump_gcs:
+    args.command += ('; WORKER_ID=$(python3 -c "import jax; print(jax.process_index())");'
```
Obliviour, line 1466: Curious if the process_index print is needed

rwitten: The print here is being used to capture the process index into args.command. Interesting idea.

Two thoughts:
* args should be immutable (so unless this is inside of initializing args we need a new variable and stitch them togeter)
* this code will run jax.process_index() which if I recall correctly retriggers slice builder and MXLA. That is probably pretty unexpected to the user. Obviously you do want an identified but how about $HOSTNAME?

rwitten (changes requested):
Little nits but broadly good!

xpk/xpk.py
ssusie: Thanks, changed to raising an error as per Rafi's suggestion

xpk/xpk.py
ssusie: Done, thanks.

xpk/xpk.py
ssusie: Modified and used the full $HOSTNAME, thanks for the suggestion!

xpk/xpk.py
```diff
@@ -473,6 +473,13 @@ def add_env_config(args):
               'environment, a value must be specified.'
           )
           env[variable] = os.environ[variable]
+
+  if args.debug_dump_gcs:
+    if 'XLA_FLAGS' in env:
+      # Caveat: might be overwriting something passed by user
+      print('overwriting XLA_FLAG from env-file')
+    env['XLA_FLAGS'] = '--xla_dump_to=/tmp/xla_dump/'
```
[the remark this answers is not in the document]
ssusie: Great point, modified to raise an error.

xpk/xpk.py
```diff
@@ -473,6 +473,14 @@ def add_env_config(args):
               'environment, a value must be specified.'
           )
           env[variable] = os.environ[variable]
+
+  if args.debug_dump_gcs:
+    if 'XLA_FLAGS' in env:
```
rwitten, line 478: Shouldn't we check if `xla_dump_to` is in the XLA_FLAGS? That way the user can set `XLA_FLAGS` (for other things).

xpk/xpk.py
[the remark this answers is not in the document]
rwitten: This is now saying the user can't use XLA_FLAGS which seems wrong. We want them to be able to use `XLA_FLAGS` but not `xla_dump_to`, right?

rwitten (changes requested):

rwitten (approved):

xpk/xpk.py
Obliviour: +1 we can support more adding `xla_dump_to` the XLA_FLAGS if `xla_dump_to` is not specified by the env file.

Obliviour (approved):

PR #239, opened 2023-11-01
Use AQT PIP Package

gobbleturk:


ZacCranko:
Hey there, Just wondering what the status is on this?

gobbleturk:
> Hey there, Just wondering what the status is on this?

The refactoring was done in https://github.com/google/maxtext/pull/253, I will close this one, thanks for pointing this out. If you sync past PR 253 MaxText now depends the AQT package - it should be bit identical behavior, but you do have to install the AQTP package

PR #241, opened 2023-11-01
avoid accessing private jax.random APIs

jakevdp:


rwitten (approved):

jakevdp:
It looks like these files will be deleted soon, so no need to make this change after all: https://github.com/google/maxtext/pull/239

PR #254, opened 2023-11-09
Add support for manually capturing a profile

yashs97:


MaxText/configs/base.yml
```diff
@@ -155,6 +155,10 @@ eos_id: 2  # sentencepiece default
 prompt: "I love to "
 
 enable_profiler: False
+# Profiles can either be captured programmatically or manually, programmatic capture is the default.
+programmatic_capture: True
```
bvandermoon, line 159: Thoughts on making this field something like profile_capture_mode? Not sure if a string would be best. But as an example, if it were a string it could be set to something like "programmatic" or "manual". Might be more clear to users. Any thoughts on how this could be represented as an either or choice between the two? Is an ENUM possible here?

rwitten: +1 on Branden's comment.

Overall though why are we proposing to make this change and what does it accomplish?

rwitten (changes requested):

MaxText/configs/base.yml
yashs97: The reason I prefer a boolean because AFAIK Maxtext is not validating strings that are passed into the config file and it always gets messy when user input type is string and input validation logic is absent.

The reason I'm making this change is because manual capture is what a few customers have used before and I feel like we should support it out of the box. Also, manual capture will eventually be the default way to automatically capture profiles.

bvandermoon: Is an enum possible for the validation? Otherwise some type of argument check could work if needed.

For default way of capturing - not sure I would go that far. But we are expecting both options will be used by customers. If you want to manual capture now, you have to insert the start server snippet manually

yashs97: User input validation is entirely missing for maxtext and is outside the scope of this PR. 
But I believe pydantic  is the goto solutions for data validation in python

shahyash10:
Closing this PR as it doesn't make sense to add this feature right now

PR #255, opened 2023-11-09
fixed integer check bug

grad0s:
the assertion doesn't check for determined_val being integer, missing function call.


rwitten (approved):

rwitten:
Thank you good samaritan! This code looks good but annoyingly you need to sign the CLA for us to be allowed to merge it.

PR #210, opened 2023-10-17
Adding support for NVIDIA GPUs to setup.sh

abhinavgoel95:
@rwitten Please have a look and let me know if you'd like some changes.

PR #265, opened 2023-11-17
Set up CI tests for GPU

michelle-yooh:
This PR does the following:
 * [Rafi's change request](https://github.com/google/maxtext/pull/210#pullrequestreview-1754162612) in #210 is resolved here.
   * The suggested use of Jax-toolbox image was substituted by manual installation of nvidia driver so that the base image can be used for both TPU and GPU. Although it might be possible that TPU runs fine on Jax-toolbox image, it is not fully tested. 
 * CI test is set up so that it runs GPU tests on a GPU runner as well. 
   * Some flags, configs and env var were added accordingly. 
   * pytest marker for TPU is introduced to distinguish the tests that can only be run on TPUs. 
 * AddLabel workflow is fixed up as a side


michelle-yooh:
`TPU test (v4-8)` will be removed from the required status checks. Instead, `Unit test on (tpu-v4-8)` and `Unit test on (gpu-a100-40gb-4)` will be added

.github/workflows/UnitTests.yml
```diff
@@ -31,47 +31,50 @@ jobs:
...
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c 'cd MaxText;python3 -m pytest'
+    - name: Test with pytest on GPU
+      if: ${{ startsWith(matrix.accelerator-type, 'gpu') }}
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c 'cd MaxText;python3 -m pytest -m "not tpu"'
     - name: Test train.py
       run: |
-        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
-        'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset steps=2 enable_checkpointing=false'
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged --shm-size=1g maxtext_base_image bash -c \
```
rwitten, line 63: It seems like we need separate things we test on GPU and TPU (and perhaps a shared file we test on both?). We shouldn't be eroding the usefulness of the TPU tests to support GPU's -- we should be testing the full capabilities of both systems..

docker_build_dependency_image.sh
```diff
@@ -45,10 +45,18 @@ if [[ -z ${MODE} ]]; then
...
+if [[ -z ${DEVICE} ]]; then
+  export DEVICE=tpu
+  echo "Default DEVICE=${DEVICE}"
+fi
+
 if [[ -z ${LIBTPU_GCS_PATH+x} ]] ; then
   export LIBTPU_GCS_PATH=NONE
   echo "Default LIBTPU_GCS_PATH=${LIBTPU_GCS_PATH}"
-  docker build --network host --build-arg MODE=${MODE} --build-arg JAX_VERSION=$JAX_VERSION --build-arg LIBTPU_GCS_PATH=$LIBTPU_GCS_PATH -f ./maxtext_dependencies.Dockerfile -t ${LOCAL_IMAGE_NAME} .
+  docker build --network host --build-arg MODE=${MODE} --build-arg JAX_VERSION=$JAX_VERSION --build-arg LIBTPU_GCS_PATH=$LIBTPU_GCS_PATH --build-arg DEVICE=$DEVICE -f ./maxtext_dependencies.Dockerfile -t ${LOCAL_IMAGE_NAME} .
```
rwitten, line 56: If DEVICE == GPU don't we build twice?

docker_build_dependency_image.sh
```diff
@@ -45,10 +45,18 @@ if [[ -z ${MODE} ]]; then
...
+  if [[ ${DEVICE} == "gpu" ]]; then
+    docker build --network host -f ./maxtext_gpu_dependencies.Dockerfile -t ${LOCAL_IMAGE_NAME} .
```
rwitten, line 58: Any chance we can respect Jax version? If not, we should design this so we assert if specific jax version was set but our GPU installer can't install specific versions.

end_to_end/test_decode_checkpoint.sh
```diff
@@ -58,7 +58,7 @@ echo
 $cmd python3 MaxText/train.py MaxText/configs/base.yml \
 run_name=${training_ckpt_run_id} \
 base_output_directory=${base_output_directory} \
-dataset_path=${dataset_path} \
+dataset_path=${dataset_path} attention=mha \
```
rwitten, line 61: Again, we need to test TPU limits on TPU and GPU limits on GPU. An interesting option for you would be to add a flag called "recommended" which chooses whatever we recommend and is platform aware.

maxtext_gpu_dependencies.Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG BASEIMAGE=maxtext_base_image
+FROM $BASEIMAGE
+
+
+RUN sed -i -e "s/main/main contrib/" /etc/apt/sources.list.d/debian.sources
+RUN curl https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64/cuda-keyring_1.1-1_all.deb -o cuda-keyring_1.1-1_all.deb
```
rwitten, line 6: Can we ask our CUDA friends how we keep this up to date for whatever is recommended for Jax? (Like do they update the .deb or do they author new ones? Is there one that is always "last stable"?)

rwitten (changes requested):
Some feedback but looks like a huge improvement!

end_to_end/test_decode_checkpoint.sh
michelle-yooh: I understand your point but I still think choosing mha over flash doesn't affect what this test is testing. But adding something like `attention=auto` sounds good to me.

.github/workflows/UnitTests.yml
rwitten: Ok --shm-size=1g seems fine but then make sure you set that in https://github.com/google/maxtext/blob/main/docker_build_dependency_image.sh#L64

Michelle, it is really annoying to no longer test the frontier settings. Can I suggest you refactor so it can run "flash_tpu" on TPUs and "flash_gpu" on GPUs. I imagine we can detect if we have GPUs or TPUs attached....

docker_build_dependency_image.sh
rwitten: Ah nice. Thanks this makes sense.

end_to_end/test_decode_checkpoint.sh
rwitten: Ok looks like we're aligned but I think we should just support `attention=mha` and `attention=flash` for now. And `attention=flash` can dispatch to TPU/GPU as needed.

rwitten (changes requested):
Little nits

rwitten (approved):

PR #267, opened 2023-11-20
Added a basic not-SPMD example.

rwitten:


ssusie (approved):
Great example. Thanks Rafi.

PR #271, opened 2023-11-28
added support for pallas flash attention for GPU

abhinavgoel95:
@rwitten please have a look at this PR

### DIFF
--- a/MaxText/layers.py
+++ b/MaxText/layers.py
@@ -43,8 +43,8 @@ import os
 import math
 
 try:
-  from jax.experimental.pallas.ops import attention
-  from jax.experimental.pallas.ops import layer_norm
+  from jax.experimental.pallas.ops import attention as pallas_attention
+  from jax.experimental.pallas.ops import layer_norm as pallas_layernorm
 except ImportError:
   logging.warning('jax_triton not found, please `pip install jax-triton`')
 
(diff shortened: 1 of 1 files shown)


MaxText/layers.py
```diff
@@ -39,8 +39,14 @@
 from jax.ad_checkpoint import checkpoint_name
 import jax.numpy as jnp
 from jax.experimental.pallas.ops.tpu import flash_attention
+import os
+import math
 
-
+try:
+  from jax.experimental.pallas.ops import attention
```
rwitten, line 46: I think we should be more careful when namespacing here, say calling these `as pallas_attention` and `as layer_norm`?

MaxText/layers.py
```diff
@@ -39,8 +39,14 @@
...
+  from jax.experimental.pallas.ops import layer_norm
+except ImportError:
+  logging.warning('jax_triton not found, please `pip install jax-triton`')
```
rwitten, line 49: I'd really prefer to not crash for all users here! Can we crash when we use these things if they're not installed.

MaxText/layers.py
```diff
@@ -348,6 +354,29 @@ def wrap_flash_attention(query, key, value, decoder_segment_ids):
             )
       x = wrap_flash_attention(query, key, value, decoder_segment_ids)
       x = jax.numpy.transpose(x, axes = (0,2,1,3))
+      
+    elif gpu_fast_attention and not decode:
+      query = LayerNorm(dtype=self.dtype, name='query_layer_norm', kernel_axes = ('heads',))(query)
+      key = LayerNorm(dtype=self.dtype, name='key_layer_norm', kernel_axes = ('heads',))(key)
+      b, s, n, h = key.shape
+      bwd_pass_impl = os.getenv(
+        'maxtext_flash_attention_backward_pass_impl', default='xla'
```
rwitten, line 363: Can this just go into config? gpu_flash_attention_backward_pass_impl?

rwitten (changes requested):

rwitten (changes requested):
There is also a conflict here with changes that are landing in a separate CR. It isn't a big deal to fix. Happy for @khatwanimohit to fix them for you.
https://github.com/google/maxtext/pull/261

PR #275, opened 2023-12-04
Initialize Jax Distributed System First

gobbleturk:
Initialize the jax distributed system before the device backend.

Tested:

Jax 0.4.20 (basically a refactor here, just calling the initialize earlier in code):
* Single slice QR

Jax nightly 0.4.21.dev-xxx (hits the new jax.distributed.initialize() without args):
* Single slice QR
* Multi slice QR
* Single slice GKE
* Multi slice GKE

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -19,6 +19,7 @@ from collections import OrderedDict
 
 import accelerator_to_spec_map
 import math
+import max_utils
 import os
 import sys
 import yaml
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -18,7 +18,6 @@
 from collections import OrderedDict
 
 import accelerator_to_spec_map
-import max_utils
 import math
 import os
 import sys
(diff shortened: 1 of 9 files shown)


rwitten (approved):
tiny nit, but feel free to ignore it.

MaxText/pyconfig.py
```diff
@@ -94,6 +95,11 @@ def _load_kwargs(self, argv: list[str], **kwargs):
   @staticmethod
   def user_init(raw_keys):
     '''Transformations between the config data and configs used at runtime'''
+
+    # We initialize the jax distributed system here because it must be done before device backend is initialized.
+    if raw_keys["enable_checkpointing"] and raw_keys["async_checkpointing"] and raw_keys["compile_topology_num_slices"]==-1:
```
rwitten, line 100: Why can't we always do it here?

rwitten (approved):
Giving you approvals so you can merge as needed. Thanks for landing this one.

MaxText/pyconfig.py
gobbleturk: We will work on making it so that it can always be done - but currently arguments only automated for cloud TPU environments - not CPU. To allow fast tests and development on CPU or GPU environments we can skip the jax distributed system by disabling async checkpointing - but in the future we should create initializers for CPU and GPU environments

PR #277, opened 2023-12-06
Maxtext commit number check

Obliviour:
This will provide the following error when there is more than one commit in the PR.

image


rwitten (approved):
OMG <3

PR #279, opened 2023-12-06
Minor update, allow the jax cache directory to be changed via environ…

obrienadam:
…ment variable. For context, on some platforms (e.g. Borg with an in-RAM filesystem) this directory will not exist.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -318,11 +318,11 @@ def main(argv: Sequence[str]) -> None:
   jax.config.update('jax_default_prng_impl', 'unsafe_rbg')
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
   os.environ["LIBTPU_INIT_ARGS"] = os.environ.get("LIBTPU_INIT_ARGS","") + " --xla_tpu_spmd_rng_bit_generator_unsafe=true"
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -318,7 +318,7 @@ def main(argv: Sequence[str]) -> None:
   jax.config.update('jax_default_prng_impl', 'unsafe_rbg')
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
   os.environ["LIBTPU_INIT_ARGS"] = os.environ.get("LIBTPU_INIT_ARGS","") + " --xla_tpu_spmd_rng_bit_generator_unsafe=true"
-  cc.initialize_cache(os.path.expanduser("~/jax_cache"))
+  cc.initialize_cache(os.environ.get("JAX_CACHE_DIR", os.path.expanduser("~/jax_cache")))
```
rwitten, line 321: Nice, but let's do this  via MaxText's config?

rwitten: We can move this lower and then write to say `config.jax_cache_dir`?

rwitten (changes requested):

MaxText/train.py
obrienadam: Sure that sounds reasonable, will update.

rwitten (approved):

PR #280, opened 2023-12-07
Add V5P support in maxtext

Obliviour:


MaxText/accelerator_to_spec_map.py
```diff
@@ -79,6 +81,295 @@ class SystemCharacteristics:
     'v4-4096': SystemCharacteristics(
       'tpu', 'v4:8x16x16', 'megacore', (2, 2, 1), 2048
     ),
+    # v5p
+    'v5p-8': SystemCharacteristics(
+        'tpu', 'v5p:2x2x1', 'default', (2, 2, 1), 4
```
Obliviour, line 86: is it default or megacor?

rwitten: pretty sure it is megacore but @gobbleturk ?

rwitten (approved):

gobbleturk (approved):

MaxText/accelerator_to_spec_map.py
Obliviour: we are megacores now

PR #281, opened 2023-12-07
Support decode checkpoint with null opt_state.

singh-mitali:
Currently training checkpoints store opt_state which is not needed for decode.
This change creates a checkpoint without opt state to reduce memory used for decode.

### DIFF
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -23,7 +23,6 @@ import os
 from absl import app
 from flax.linen import partitioning as nn_partitioning
 import numpy as np
-import optax
 
 from layers import Transformer
 import pyconfig
@@ -133,15 +132,13 @@ def decode_loop(config, state=None):
 
   # Model and Optimizer definition
   model = Transformer(config, mesh = mesh)
-
-  tx = optax.adamw(
-    max_utils.create_learning_rate_schedule(config)
-  ) # TODO: we need an optax.GradientTransformation to form a TrainState, but we don't use it when decoding
-
-
   _, sp_tokenizer = create_data_iterator_with_tokenizer(config, mesh)
-
-  state, state_mesh_annotations = max_utils.setup_initial_state(model, tx, config, rng, mesh, checkpoint_manager)
+  state, state_mesh_annotations = max_utils.setup_decode_state(
+    model, config, rng, mesh, checkpoint_manager
+    )
+  assert state.opt_state == {}, "non null opt_state in checkpoint"
+  num_params = max_utils.calculate_num_params_from_pytree(state.params)
+  max_logging.log(f"Number of model params={num_params/10**9:.3f} billion")
 
   state_mesh_shardings = jax.tree_map(
       lambda p: jax.sharding.NamedSharding(mesh, p), state_mesh_annotations)
--- a/end_to_end/test_decode_checkpoint.sh
+++ b/end_to_end/test_decode_checkpoint.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+set -ue #x for debugging
+
+helpFunction()
+{
+  echo ""
+  echo "Usage: $0 <dataset_path> <output_path>"
+  echo -e "\t-n dry_run is true "
+  echo -e "\t-r runid: run_test_model_0b"
+  echo -e "\t-d dataset_path: gs://test-maxtext-dataset"
+  echo -e "\t-o output_path: gs://test-maxtext-output"
+  echo -e "\t-t num_token_threshold: 40"
+  echo -e "\t-i ici_tensor_parallelism: 8"
+  exit 1 # Exit script after printing help
+}
+
+# Default option values
+dry_run=false
+run_id=test_model_0b_$(date +%Y-%m-%d-%H)
+dataset_path=gs://test-maxtext-dataset
+base_output_directory=gs://test-maxtext-output
+ici_tensor_parallelism=8
+num_tokens_threshold=40
+
+while getopts "nr:d:o:t:i:" opt
+do
+  case "$opt" in
+      n ) dry_run=true ;;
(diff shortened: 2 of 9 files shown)


end_to_end/test_decode_checkpoint.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+
+set -ue #x for debugging
+
+helpFunction()
```
rwitten, line 5: Is this being run as an integ test somewhere? It would probably be best to run these directly on the github runner?

MaxText/decode.py
```diff
@@ -134,14 +134,14 @@ def decode_loop(config, state=None):
   # Model and Optimizer definition
   model = Transformer(config, mesh = mesh)
 
-  tx = optax.adamw(
```
rwitten, line 137: <3

MaxText/decode.py
```diff
@@ -134,14 +134,14 @@ def decode_loop(config, state=None):
...
-    max_utils.create_learning_rate_schedule(config)
-  ) # TODO: we need an optax.GradientTransformation to form a TrainState, but we don't use it when decoding
-
-
   _, sp_tokenizer = create_data_iterator_with_tokenizer(config, mesh)
 
-  state, state_mesh_annotations = max_utils.setup_initial_state(model, tx, config, rng, mesh, checkpoint_manager)
+  state, state_mesh_annotations = max_utils.setup_initial_state(
```
rwitten, line 139: Why don't we have "setup_state_for_inference" and "setup_initial_state" and only pass the right things into the two of them. Going through one interface for the two difference pretty different tasks seems unncessary.

MaxText/decode.py
```diff
@@ -134,14 +134,14 @@ def decode_loop(config, state=None):
...
+    model, None, config, rng, mesh, checkpoint_manager, True
+    )
+  assert state.opt_state == {}, "non null opt_state in checkpoint"
+  num_params = max_utils.calculate_num_params_from_pytree(state.params)
```
rwitten, line 143: I think it is weird we have a strange TrainState, that really seems not ideal. Wouldn't it be better to just not have a train state? Is that hard for some reason?

MaxText/decode_checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+import pyconfig
+
+from absl import app
+from etils import epath
+from jax.sharding import Mesh
+from jax import random
+from typing import Sequence
+from layers import Transformer
+
+def read_checkpoint(config, checkpoint_manager, is_decode=False):
```
rwitten, line 36: What does "is_decode" mean in this context?

MaxText/decode_checkpoint.py
```diff
@@ -0,0 +1,123 @@
+"""
```
rwitten, line 1: Did you discuss with Colin Gaffney if we could instead load just the params from a bigger checkpoint? I would have thought (possibly naively) that we could. (Also maybe this file would still be useful?)

rwitten (changes requested):

MaxText/decode.py
singh-mitali: Done.

MaxText/decode_checkpoint.py
singh-mitali: updated var naming to make it easier to comprehend.

MaxText/decode.py
singh-mitali: updated functions to make it cleaner.

MaxText/decode.py
singh-mitali: Not sure what the suggestion was here.

MaxText/decode_checkpoint.py
rwitten: Totally reasonable! But wouldn't we still want to execute a partial read?

MaxText/decode.py
rwitten: yes looks amazingly clear!

MaxText/decode.py
```diff
@@ -133,15 +132,13 @@ def decode_loop(config, state=None):
...
-    max_utils.create_learning_rate_schedule(config)
-  ) # TODO: we need an optax.GradientTransformation to form a TrainState, but we don't use it when decoding
-
-
   _, sp_tokenizer = create_data_iterator_with_tokenizer(config, mesh)
-
-  state, state_mesh_annotations = max_utils.setup_initial_state(model, tx, config, rng, mesh, checkpoint_manager)
+  state, state_mesh_annotations = max_utils.setup_decode_state(
+    model, config, rng, mesh, checkpoint_manager
+    )
```
rwitten, line 138: I think our linters might complain unless the ")" is moved one left here but I don't know!

MaxText/decode_checkpoint.py
```diff
@@ -0,0 +1,131 @@
...
+  assert decode_state.opt_state == {}, "non null opt_state in decode checkpoint"
+  num_training_params = max_utils.calculate_num_params_from_pytree(training_state.params)
+  num_decode_params = max_utils.calculate_num_params_from_pytree(decode_state.params)
+  assert num_training_params == num_decode_params, "count mismatch in params in checkpoints"
+
+  max_logging.log(f"Successfully generated decode checkpoint at: {config.checkpoint_dir}0/default")
+  return True
+
+
+def main(argv: Sequence[str]) -> None:
```
rwitten, line 124: I believe this function's only job is to generate_decode_checkpoint's in which case I think we should call it generate_decode_checkpoint.py.

rwitten (approved):
Looks solid. I'm still a little uneasy about decode_checkpoint not dropping the keys it doesn't need as it loads the checkpoint -- that would avoid some OOMs but I don't think it is a huge deal.

PR #282, opened 2023-12-07
Added support for pallas flash attention for GPU

khatwanimohit:
For merging changes from @abhinavgoel95 in https://github.com/google/maxtext/pull/271

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -57,8 +57,9 @@ logits_via_embedding: True  # NOTE: this is True just for testing.
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: flash, mha
+attention: 'flash' # Supported attention: flash, mha, gpu_flash
 
+gpu_flash_attention_backward_pass_impl: "xla" # allowed values (xla, triton)
 record_internal_nn_metrics: 0
 
 # Output directory
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -15,6 +15,7 @@
 """Attentions Layers."""
 
 import functools
+import math
 from typing import Optional
 
 from flax import linen as nn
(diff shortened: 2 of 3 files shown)


MaxText/configs/base.yml
```diff
@@ -57,8 +57,9 @@ logits_via_embedding: True  # NOTE: this is True just for testing.
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: flash, mha
+attention: 'flash' # Supported attention: flash, mha, gpu_mha
```
rwitten, line 60: gpu_flash, right?

rwitten (changes requested):
nits!

MaxText/configs/base.yml
```diff
@@ -57,8 +57,9 @@ logits_via_embedding: True  # NOTE: this is True just for testing.
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: flash, mha
+attention: 'flash' # Supported attention: flash, mha, gpu_flash
 
+gpu_flash_attention_backward_pass_impl: "xla" # allowed values (xla, triton)
```
rwitten, line 62: I remember there was some discussion of if this controlled both the backwards pass and forwards pass or just the backwards pass. If it is just the backwards pass, this is great but if it is controlling both can we update accordingly?

MaxText/layers.py
```diff
@@ -32,14 +32,19 @@
...
 from jax import lax
 from jax import random
 from jax.ad_checkpoint import checkpoint_name
 import jax.numpy as jnp
 from jax.experimental.pallas.ops.tpu import flash_attention
-
+try:
+  from jax.experimental.pallas.ops import attention as pallas_attention
+except ImportError:
+  max_logging.log("WARNING: jax_triton not found, please `pip install jax-triton`")
```
rwitten, line 47: Really should be `please install jax-triton to get best performance on GPU`.

rwitten (approved):
Tiny nit but want to give approval.

MaxText/layers.py
rwitten: Like the two points are (1) are they really supposed to install jax-triton via pip and (2) TPU users shouldn't worry about this.

PR #283, opened 2023-12-08
Data loader for Maxtext training emulator for storage

RoshaniN:
The standalone data loader, sets up the model and data iterator similar to the train_loop of train.py. The data loader iterates through batches of data, to log step time of the first step and time taken to load all the batches, but does not actually train the model.

The data loader mimics the interactions of the TPU VM hosts with the storage systems such as GCS.

I will add Github workflow to run standalone_dataloader.py separately.


RoshaniN:
Missing CLA is due to pulling in commits from main - 
❌ https://github.com/google/maxtext/commit/89012c22a4ff3fc7e0ba5c0a3075df54038d18f2 Author: @a-googler <no****ly​@google.com>

rwitten (changes requested):
This CR looks great! But I strongly encourage you to add a quick command line test into UnitTests.yml. Folks will (accidentally) break `standalone_dataloader.py` SUPER easily.

P.S. - are the timing results right? You might need a jax.block_until_ready depending on some details about how the data loading pipeline is designed.

MaxText/standalone_dataloader.py
```diff
@@ -0,0 +1,77 @@
...
+
+  example_batch = None
+
+  start = datetime.datetime.now()
+  start_step = get_first_step(state)
+
+  # Actual data loading steps
+  for step in np.arange(start_step, config.steps):
+    example_batch = load_next_batch(data_iterator, example_batch, config)
+    # print("Step ", step, " finished in ", new_time - last_step_completion)
```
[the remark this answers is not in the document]
RoshaniN: Done, thank you!

MaxText/standalone_dataloader.py
```diff
@@ -0,0 +1,77 @@
...
+  # Actual data loading steps
+  for step in np.arange(start_step, config.steps):
+    example_batch = load_next_batch(data_iterator, example_batch, config)
+    jax.block_until_ready(example_batch)
+    if step==0:
+      new_time = datetime.datetime.now()
+      print("First step completed in ", new_time-start," seconds")
+
+  end = datetime.datetime.now()
+  print(config.steps," batches loaded in ", end-start ," seconds, on host ", jax.process_index())
```
rwitten, line 56: Mostly a nit but will leave the code a bit clearer.

What are you trying to time?

I think you're trying to time separately:
(a) the time to first batch
(b) once initialized the time to load N batches.

If that case, you need should:
* time (a) outside of the for loop (and end the timer on (a) when `block_until_ready`)
* time (b) by starting the timer, loading n batches and `block_until_ready` on all of them at the end.

I actually don't think it matters much in this case that you have block_until_ready's in the for loop, but you're adding a curious dependency between the data.

rwitten (approved):
There's a bit of cleanup I'd prefer here, otherwise looks good! But approving since it really isn't a big deal.

MaxText/standalone_dataloader.py
RoshaniN: Thanks for the comment!
I am trying to time separately:

1. The time to load the first batch 
Is your suggestion to do the following outside of the for loop?
```
start = datetime.datetime.now()
start_step = get_first_step(state)
jax.block_until_ready(start_step)
first_end = datetime.datetime.now()
time_to_load_first_batch = first_end-start
```
In this case, I think the iterator has not moved, only jax spmd initialization has happened.

2. Time to load n batches, including the first one. In xpk runs, I was able to see the time taken to load the batches on the slowest host at the end.

```
start = datetime.datetime.now()
for step in np.arange(start_step, config.steps):
jax.block_until_ready(example_batch) # waiting for the last batch
end = datetime.datetime.now()
```

Do you think adding ` jax.block_until_ready(example_batch)` outside of the for loop,  serves my intent of waiting until last batch is loaded?

rwitten: No, in 1 we should start when the program starts and include a 
```
```

Yes 2 is fine.

PR #284, opened 2023-12-08
Updated README.md

chiefkarlin:
Fixed startup guide paths


rwitten (approved):

Obliviour (approved):
Hi @chiefkarlin I think the commits need to be squashed into one and then it can move to the next step!

PR #285, opened 2023-12-11
llama2 ckpt convert

A9isha:
This PR includes:

1. A script to convert Llama2 7B model checkpoint to MaxText orbax format
2. Convert layers.py to match Llama2 architecture


MaxText/layers.py
```diff
@@ -609,11 +609,22 @@ def __call__(self, inputs, decode: bool = False, deterministic: bool = False):
...
+          name="ffn_layer1",
+          config=cfg)(
+              inputs)
+    x = jnp.multiply(x, up_proj_x)
+
+    # No Dropout in Llama
+    # x = nn.Dropout(
+    #     rate=self.intermediate_dropout_rate, broadcast_dims=(-2,))(
+    #         x, deterministic=deterministic)  # Broadcast along length.
+    # x = nn.with_logical_constraint(x, ('activation_batch', 'activation_length', 'activation_mlp'))
```
hanzlfs, line 627: I think we still need this sharding constraint.

MaxText/layers.py
```diff
@@ -566,7 +567,6 @@ def swap_dims(x):
 
 
 
-
 class MlpBlock(nn.Module):
```
hanzlfs, line 570: Can we either rename or inherit this to LLamaMLPBlock?

MaxText/layers.py
```diff
@@ -1176,9 +1200,10 @@ def __call__(self,
                 max_decode_length)
 
     y = LayerNorm(dtype=cfg.dtype, name='decoder_norm', kernel_axes = ('embed',))(y)
-    y = nn.Dropout(
-        rate=cfg.dropout_rate, broadcast_dims=(-2,))(
-            y, deterministic=deterministic)
+    #No Dropout in Llama
```
hanzlfs, line 1203: Instead of commenting out dropout, could we do
if cfg.dropout_rate is not None:
   y = nn.Dropout(...)

And configure llama model with dropout rate is None?

MaxText/layers.py
```diff
@@ -1038,6 +1049,8 @@ def __call__(self,
     inputs = nn.with_logical_constraint(inputs, ('activation_batch', 'activation_length', 'activation_embed'))
 
     # inputs: embedded inputs to the decoder with shape [batch, length, emb_dim]
+    residual = inputs
+    #input_layernorm aka pre_self_attention_layer_norm
```
hanzlfs, line 1053: Layernorm is confusing here, we probably need rename it as input_norm. But it can be address in another PR.

MaxText/layers.py
hanzlfs: Same as other droput layers.

MaxText/configs/base.yml
hanzlfs, line 1: do we need a new config file? or rename this as llama_7b?

MaxText/layers.py
```diff
@@ -1176,9 +1200,10 @@ def __call__(self,
                 max_decode_length)
 
     y = LayerNorm(dtype=cfg.dtype, name='decoder_norm', kernel_axes = ('embed',))(y)
-    y = nn.Dropout(
-        rate=cfg.dropout_rate, broadcast_dims=(-2,))(
-            y, deterministic=deterministic)
```
rwitten, line 1181: Can't we just set dropout rate to zero?

MaxText/max_utils.py
```diff
@@ -378,3 +379,11 @@ def get_abstract_state(model, tx, config, rng, mesh):
   with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
     state_mesh_annotations = nn.logical_to_mesh(state_logical_annotations)
   return unboxed_abstract_state, state_mesh_annotations
+
+def convert_state_to_jnp(state_dictionary):
```
rwitten, line 383: nit: probably a treemap is cleaner here?

MaxText/max_utils.py
```diff
@@ -225,6 +225,7 @@ def setup_initial_state(model, tx, config, rng, mesh, checkpoint_manager):
                                                 mesh,
                                                 state_mesh_annotations)
 
+
```
rwitten, line 228: nit: please don't mess with whitespace

MaxText/layers.py
```diff
@@ -1116,9 +1139,10 @@ def __call__(self,
 
     # [batch, length] -> [batch, length, emb_dim]
     y = self.shared_embedding(decoder_input_tokens.astype('int32'))
-    y = nn.Dropout(
-        rate=cfg.dropout_rate, broadcast_dims=(-2,))(
-            y, deterministic=deterministic)
+    #No Dropout in Llama
```
rwitten, line 1142: Can't we just set dropout rate to zero?

MaxText/layers.py
```diff
@@ -1071,16 +1093,17 @@ def __call__(self,
...
-            next_layer_addition, deterministic=deterministic)
+    layer_output = mlp_lnx + residual
+    
+    #No Dropout in Llama
+    # next_layer_addition_dropped_out = nn.Dropout(
+    #     rate=cfg.dropout_rate, broadcast_dims=(-2,))(
+    #         next_layer_addition, deterministic=deterministic)
 
-    layer_output = next_layer_addition_dropped_out + inputs
+    # layer_output = next_layer_addition_dropped_out + inputs
```
rwitten, line 1106: nit: dead code?

MaxText/layers.py
```diff
@@ -1071,16 +1093,17 @@ def __call__(self,
...
     mlp_lnx = nn.with_logical_constraint(mlp_lnx, ('activation_batch', 'activation_length', 'activation_embed'))
 
-    next_layer_addition = mlp_lnx + attention_lnx
-
-    next_layer_addition_dropped_out = nn.Dropout(
-        rate=cfg.dropout_rate, broadcast_dims=(-2,))(
-            next_layer_addition, deterministic=deterministic)
+    layer_output = mlp_lnx + residual
+    
+    #No Dropout in Llama
```
rwitten, line 1101: Can't we just set dropout rate to zero? (Sorry I'm copy and pasting this one haha)

MaxText/layers.py
```diff
@@ -441,9 +441,10 @@ def query_init(*args):
     # Layer norms here prevent (near) one-hot softmaxes, which can lead to
     # unstable training loss and nans, see the "QK Normalization" subsection in
     # https://arxiv.org/pdf/2302.05442.pdf.
-    query = LayerNorm(dtype=self.dtype, name='query_layer_norm', kernel_axes = ('heads',))(query)
-    key = LayerNorm(dtype=self.dtype, name='key_layer_norm', kernel_axes = ('heads',))(key)
-    value = LayerNorm(dtype=self.dtype, name='value_layer_norm', kernel_axes = ('heads',))(value)
+    #Llama architecture doesn't have these layernorms
```
rwitten, line 444: Probably make these configurable?

MaxText/configs/base.yml
```diff
@@ -44,20 +44,20 @@ int8_training: False
...
-head_dim: 256
+base_emb_dim: 4096
+base_num_heads: 32
+base_mlp_dim: 11008
+base_num_decoder_layers: 32
+head_dim: 128
 # activation functions are .
-mlp_activations: ["relu"]
+mlp_activations: ["silu"]
 dropout_rate: 0
```
hanzlfs, line 54: There is another norm config eps to be added. As we should rename layernorm etc, I can address this issue in another PR.

MaxText/configs/base.yml
```diff
@@ -44,20 +44,20 @@ int8_training: False
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes 
 # then you should explicitly set base_embed_dim, base_num_heads, base_mlp_dim, base_num_decoder_layers and/or head_dim.
 global_parameter_scale: 1
-base_emb_dim: 2560
```
rwitten, line 47: Just to make sure we're on the same page, are you working with Mohit to add forking logic where if the model is set to say "llama2-7b" these overrides go into effect? The overrides can be hard coded into pyconfig.

rwitten (changes requested):
Very cool! Some feedback inline. Have you been able to run forwards passes in MaxText and in an existing Llama2 codebase?

MaxText/layers.py
hanzlfs: +1

MaxText/configs/base.yml
A9isha: There is a epsilon param in the LayerNorm computation, would that suffice?
https://github.com/google/maxtext/blob/main/MaxText/layers.py#L635

hanzlfs: It's different from llama2's config, their default eps is 1e-5
https://huggingface.co/meta-llama/Llama-2-7b-hf/blob/main/config.json#L18

hanzlfs: Let's not change in the layerNorm class, we need rename this class to RmsNorm

MaxText/configs/base.yml
```diff
@@ -44,20 +44,20 @@ int8_training: False
...
-mlp_activations: ["relu"]
+mlp_activations: ["silu"]
 dropout_rate: 0
 logits_via_embedding: True  # NOTE: this is True just for testing.
 # proj, minimal, full, or none
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: flash, mha
+attention: 'mha' # Supported attention: flash, mha
```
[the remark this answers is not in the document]
hanzlfs: 2cents of mine not related to this PR: it may be OK to not use flash attention by default. Also we should support more both flash and non flash attention. As we are also need flash_mha_gpu and flash_mha_tpu (this current `flash`).

Also to extend current code to support llama-70b and mixtral in future, we need (GQA and local attention).
And may need flash_gqa_gpu(tpu) and flash_sliding_widow extra.

For inference mode, we may need to show performance of MQA as well. So In fact we need a more attention layers.

MaxText/layers.py
A9isha: Done

MaxText/max_utils.py
A9isha: Removed - sorry about that!

MaxText/layers.py
A9isha: Done.

MaxText/layers.py
A9isha: Brought back all the Dropout removals as suggested by Rafi since

by default dropout_rate=0

https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml#L54
https://github.com/google/maxtext/blob/main/MaxText/layers.py#L1070

MaxText/layers.py
A9isha: Brought it back - thanks!

MaxText/layers.py
A9isha: Removed - thanks!

MaxText/convert_llama_ckpt.py
```diff
@@ -0,0 +1,306 @@
+r"""Convert weights from a llama/vicuna model to a pax one.
```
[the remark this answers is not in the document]
A9isha: I removed the references to Pax.

The trouble with adding a test is that the script needs Meta's checkpoint - which we cannot add to MaxText repo though.

MaxText/configs/base.yml
rwitten: @hanzlfs @A9isha @khatwanimohit I think it will be better to just expose a config flag that does overrides for llama_7b. Mohit is going to throw up a prototype for this.

MaxText/configs/base.yml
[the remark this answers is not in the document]
rwitten: 100% agreed on supporting all of these things in teh future. However, I think flash should be the default unless you have a specific concern. (Is that concern "GPU won't work out of the box" -- that is a good one!)

MaxText/layers.py
rwitten: Yes that seems great but I don't think we want to change types (i.e. None vs. 1e-5). Happy to either have a control flag (qkv_rms_norm_enabled) or use values <0 to mean disabled.

MaxText/configs/base.yml
A9isha: utilizing Mohit's prototype now.

MaxText/configs/base.yml
A9isha: utilizing Mohit's prototype

MaxText/configs/base.yml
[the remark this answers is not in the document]
A9isha: reverted this back.

PR #286, opened 2023-12-11
Allow the start step for profiling to be configured

obrienadam:
For Megascale benchmarking purposes we sometimes want to delay profiling start for a few steps until the iteration time has stabilized.

We were rolling a custom training loop for this but we'd like to avoid that as it's been broken a few times.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -273,8 +273,12 @@ def train_loop(config, state=None):
   local_metrics_file = open(config.metrics_file, 'a', encoding="utf8") if config.metrics_file else None
   running_gcs_metrics = [] if config.gcs_metrics else None
 
-  start_step = get_first_step(state)
+  start_step = get_first_step(state) # this is the start_step for training
+  first_profiling_step = start_step + config.skip_first_n_steps_for_profiler
   for step in np.arange(start_step, config.steps):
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -273,8 +273,11 @@ def train_loop(config, state=None):
   local_metrics_file = open(config.metrics_file, 'a', encoding="utf8") if config.metrics_file else None
   running_gcs_metrics = [] if config.gcs_metrics else None
 
-  start_step = get_first_step(state)
+  start_step = max(get_first_step(state), config.profiling_start_step)
```
rwitten, line 276: Not sure I 'get it' -- isn't this changing our first step because we use `start_step` on 277?

I think you want this notion of start step on the check in 278? On this line we could instead write:

```
start_step = get_first_step(state) # this is the start_step for training
first_profiling_step = start_step + config.skip_first_n_steps_for_profiler
```

rwitten (changes requested):
I might be missing something but I don't htink this CR is quite what you want?

MaxText/train.py
obrienadam: Ah I see, yeah I will update. Thanks!

rwitten (approved):

PR #287, opened 2023-12-12
refactor layers.py

hanzlfs:
Refactor layers.py into
- attentions.py, which in future we will extend to MHA MQA, GQA local attention
- linears.py includes dense general and MLP, in future we will extend to MOE
- normalizations.py
- quantizations.py which will incudes fp8 in future for gpu
- embeddings.py


hanzlfs (commented):
change attention tests.

rwitten (approved):
This looks amazing to me. I noticed we changed from LayerNorm to RMSNorm -- are there any other substantive changes other than the refactoring.

PR #289, opened 2023-12-13
Reorganize MaxText/configs/ directory and documentation, and add v5p results

raymondzouu:
This PR contains the following changes:
* organize `MaxText/configs/` directory into `v4`, `v5e`, and `v5p` directories
* Add README for `MaxText/configs/` directory with repro instructions that point to `get_started/`and add READMEs for `v4`, `v5e`, and `v5p` directories with runtime results
* add `preflight.sh` script that handles initial setup like rto network setup
* condense testing scripts `test_convergence_1b_params.sh`, `test_tflops_16b_params.sh`, `test_tflops_32b_params.sh`, `test_tflops_64b_params.sh` into single scripts

To do after these changes are submitted:
* delete `MaxText/configs/experimental` after user guide is updated to point to `MaxText/configs/v5p`
* delete `v5e_high_performance.md` in copybara/g3


preflight.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "Error: PLATFORM flag is missing."
+    exit 1
+fi
+
+if [[ $PLATFORM == "gce" ]]; then
+    # Set up network for running on gce
+    echo "Setting up network for GCE"
+    bash rto_setup.sh
+    # For DNS lookup when running on large number of VMs on gce
+    echo '142.250.123.95 www.googleapis.com' | sudo tee -a /etc/hosts
```
gobbleturk, line 27: @rwitten is this still needed? I vaguely recall this being fixed at some point but maybe not

gobbleturk (approved):
Awesome change, I was imagining cleaning this up at some point, very happy to see it here!

preflight.sh
rwitten: Ya I think we should get rid of this IP pinning for now. It is needed only in a very weird setting.

rwitten (approved):
One tiny nit but looks like an amazing step forward.

preflight.sh
raymondzouu: Sounds good, I removed it.

rwitten (approved):

PR #290, opened 2023-12-14
Fix broken import statements

michelle-yooh:
- layers.py is deleted because it seems it is left unintentionally #287. 
 - types.py is renamed to common_types.py because it caused confusion when the import is resolved.


hanzlfs:
HI Michelle, thanks for fixing I also found these two lines need to be changed 
https://github.com/google/maxtext/blob/dd263510224aa44708a7ab47d68538497df6ba2f/MaxText/generate_decode_checkpoint.py#L34-L36C6

from layers import model -> from layers import models
model.Transformer -> models.Transformer.

Thanks for fixing this!

hanzlfs:
Also in llama_test.py need to add
from layers import embeddings

L100 and L101
layers.LLaMARotaryEmbedding -> embeddings.LLaMARotaryEmbedding
https://github.com/google/maxtext/blob/main/MaxText/tests/llama_test.py#L100-L101

rwitten (approved):

PR #291, opened 2023-12-15
More refactor of attentions  & Add MQA GQA 

hanzlfs:
1. More MHA refactor (remove attention norm)
2. support MQA GQA(for Llama 70B)

### DIFF
--- a/MaxText/tests/attention_test.py
+++ b/MaxText/tests/attention_test.py
@@ -1,34 +1,40 @@
-"""
- Copyright 2023 Google LLC
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
 
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
+"""Tests for Attentions."""
 
-      https://www.apache.org/licenses/LICENSE-2.0
-
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- See the License for the specific language governing permissions and
- limitations under the License.
- """
-
-""" Tests for Attention """
-import jax
+import sys
 import unittest
+
+from flax.core import freeze
+import jax
 import jax.numpy as jnp
 import max_utils
-from jax.sharding import Mesh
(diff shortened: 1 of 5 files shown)


MaxText/tests/attention_test.py
```diff
@@ -1,110 +1,402 @@
...
+    )
+
+    self.assertTrue(
+        jax.numpy.allclose(
+            flash_output, mha_output, rtol=1e-01, atol=1e-01, equal_nan=False
+        )
+    )
+
+  def test_multihead_attention(self):
+    """Test MultiHeadDotProductAttention."""
```
rwitten, line 155: I find the reimplementation kind of weird! Shouldn't we just test that MHA == FLASH_MHA?

MaxText/tests/attention_test.py
```diff
@@ -1,110 +1,402 @@
...
+    weights = jax.nn.softmax(logits, axis=-1)
+    combined = jnp.einsum('bnts,bsnd->btnd', weights, value)
+    expected = jnp.einsum('btnd,ndf->btf', combined, out_kernel)
+
+    out = np.asarray(out).astype(np.float16)
+    expected = np.asarray(expected).astype(np.float16)
+    np.testing.assert_allclose(out, expected, atol=1e-6)
+
+  def test_multiquery_attention(self):
+    """Test MultiQueryDotProductAttention."""
```
rwitten, line 235: Again I don't love the reimplementation. Shouldn't we test that MQA is the same as MHA (with queries replicated)?

MaxText/tests/attention_test.py
```diff
@@ -1,110 +1,402 @@
...
+    weights = jax.nn.softmax(logits, axis=-1)
+    combined = jnp.einsum('bnts,bsd->btnd', weights, value)
+    expected = jnp.einsum('btnd,ndf->btf', combined, out_kernel)
+
+    out = np.asarray(out).astype(np.float16)
+    expected = np.asarray(expected).astype(np.float16)
+    np.testing.assert_allclose(out, expected, atol=1e-6)
+
+  def test_groupedquery_attention(self):
+    """Test GroupedQueryDotProductAttention."""
```
rwitten, line 317: Again, I would just test that GQA recovers MHA where the keys replicated.

rwitten (changes requested):
LOL you've made me OK with the code but I still find the tests annoying. The tests are all a "reimplementation" which will be annoying when the code inevitably changes. Wouldn't it be better to tests MHA == FlashMHA and then GQA and MQA use MHA for their tests (by replicating their KV caches?)

MaxText/tests/attention_test.py
hanzlfs: Then how we know MHA or Flash MHA is correct?

hanzlfs: Who is our ground truth then?

MaxText/tests/attention_test.py
hanzlfs: Stacking them into a method, is not readable.

MaxText/tests/attention_test.py
hanzlfs: same as above

MaxText/tests/attention_test.py
hanzlfs: Refactoring this to equivalent test between GQA = MHA when num_kv_head = num_head

rwitten (approved):
let's go! :-)

PR #292, opened 2023-12-15
Support multiple models

khatwanimohit:
Update model config variables through pyconfig.py to add support for multiple models

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -16,6 +16,10 @@
 # If there is already a checkpoint under this run, that checkpoint will auto-resume.
 run_name: ""
 
+# Types of models supported.
+# default model (palm like model with no overrides in config)
+model_name: "default"
+
 # At most one of load_parameters_path and load_from_other_directory should be set.
 #   * load_parameters_path is used when you only care about the parameters, not the optimizer state or
 #      the step count
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -17,6 +17,8 @@
 # pylint: disable=missing-module-docstring
 from collections import OrderedDict
 
+import max_logging
+
 import accelerator_to_spec_map
 import math
 import max_utils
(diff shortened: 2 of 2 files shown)


MaxText/configs/base.yml
```diff
@@ -16,6 +16,10 @@
 # If there is already a checkpoint under this run, that checkpoint will auto-resume.
 run_name: ""
 
+# Types of models supported.
+# Currently supported models: gpt3
+model_name: "gpt3"
```
rwitten, line 21: Current model is "palm" like.

MaxText/pyconfig.py
```diff
@@ -129,6 +139,33 @@ def user_init(raw_keys):
...
+      raw_keys = validate_and_update_keys(raw_keys, llama_7b_model_vars)
+
+def validate_and_update_keys(raw_keys, model_keys):
+  max_logging.log("Updating following parameters in config\n")
+  for k in model_keys:
+    max_logging.log(f"{k}: {model_keys[k]}")
+    if k in raw_keys and isinstance(raw_keys[k], type(model_keys[k])):
+      raw_keys[k] = model_keys[k]
+    else:
+      raise ValueError(f'Key {k} does not exist in config/base.yml.')
```
rwitten, line 166: Give a more polite error message! I.e. it might be the wrong type. A standard way to write this beautifully is:

```
if k not in raw_keys:
  raise ValueError("Blah1")
elif  not isinstance
  raise ValueError("Blah2")
else:
  do the thing

File: MaxText/configs/base.yml (hunk 1 above)
Comment (rwitten) [reply MaxText/configs/base.yml:21]: But I think we should call it `default` for now -- it is really saying that if you use this model we don't use any overrides.

File: MaxText/pyconfig.py [hunk 3]
```diff
@@ -129,6 +139,33 @@ def user_init(raw_keys):
 
     validate_attention_type(raw_keys['attention'])
 
+  @staticmethod
+  def update_model_vars(raw_keys):
+    ''' Update model config variables
+    '''
+    validate_model_name(raw_keys['model_name'])
```
rwitten, line 146: I don't think we quite are asserting that the models are the same yet. I think for now a polite solution would be to add a ValueError for if you used llama-7b.

MaxText/configs/base.yml
rwitten: (And if you say anything else we will mix in some overrides for you.)

rwitten (changes requested):
Little nits.

rwitten (approved):

PR #294, opened 2023-12-18
modify maxtext to use changes directly from orbax

ssusie:
Added an option `enable_single_slice_checkpointing` to base.yaml. When set to True, will use `SingleSliceArrayRestoreArgs` directly from orbax to do broadcasting on multiple slices.


.gitignore
```diff
@@ -10,6 +10,7 @@ __pycache__/
 *.so
 
 # Distribution / packaging
+orbax/
```
rwitten, line 13: Let's land this feature in Orbax first and then land it in MaxText. I'm totally agree in principle in MaxText for now.

rwitten (changes requested):

.gitignore
ssusie: Sounds good, thanks.

PR #296, opened 2023-12-20
Fix v5p name to v5 for AOT API

gobbleturk:
Fix internal device names from v5p to v5 for AOT API - the user facing name will remain v5p

### DIFF
--- a/MaxText/accelerator_to_spec_map.py
+++ b/MaxText/accelerator_to_spec_map.py
@@ -83,292 +83,292 @@ UserFacingNameToSystemCharacteristics = {
     ),
     # v5p
     'v5p-8': SystemCharacteristics(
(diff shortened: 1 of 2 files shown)


Obliviour (approved):

rwitten (approved):
This looks fine but is this in sync with the XPK kind of equivalent list? (Or are we OK with them falling out os sync?)

gobbleturk:
> This looks fine but is this in sync with the XPK kind of equivalent list? (Or are we OK with them falling out os sync?)

For v5p the external names still match - I think its okay that this internal field doesn't match exactly since they are used for different APIs - for AOT we pass "v5:8x8x8" to `get_topology_description` whereas XPK probably passes "v5p" and "8x8x8" as separate arguments to some GKE slice creation. I think it makes sense to keep the external names for each size in sync but internal fields can diverge

PR #297, opened 2023-12-21
Profile for fixed number of steps

gobbleturk:
Profile for a configurable number of steps.

This decouples the number of steps to train and the number of steps to profile - allowing us to run for a large number of of steps while still profiling for a small number of steps.

Tested with steps = 1, 2, 5, 8

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -279,6 +279,9 @@ def train_loop(config, state=None):
 
   start_step = get_first_step(state) # this is the start_step for training
   first_profiling_step = start_step + config.skip_first_n_steps_for_profiler
+  if config.enable_profiler and first_profiling_step >= config.steps:
+    raise ValueError("Profiling requested but initial profiling step set past training final step")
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -279,6 +279,8 @@ def train_loop(config, state=None):
 
   start_step = get_first_step(state) # this is the start_step for training
   first_profiling_step = start_step + config.skip_first_n_steps_for_profiler
+  last_profiling_step = np.min([first_profiling_step + config.profiler_steps - 1, config.steps - 1])
```
rwitten, line 282: please use https://numpy.org/doc/stable/reference/generated/numpy.clip.html

rwitten (changes requested):
I love the change, just making life easier and easier!

MaxText/train.py
gobbleturk: Changed, glad to learn of numpy clip. Although numpy.clip uses min(a_max, max(a,min_bound)) which is the opposite order of what I had before. This changes the behavior of the edge case when first_profiling_step >= config.steps (e.g. steps = 1) - so I've added an assert for it

rwitten (approved):

PR #304, opened 2023-12-27
import pjit correctly in shardings.py

yashs97:


khatwanimohit (approved):

rwitten (approved):
Even better would be adding a unit test that just ran shardings.py to keep it running!

PR #309, opened 2024-01-03
Anisha llama2

A9isha:
Llama2 architecture in MaxText. Inference only support at the moment. 
Finetuning is soon to be supported


.github/workflows/UnitTests.yml
```diff
@@ -80,6 +80,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_decode_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M) -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4'
+    - name: Test llama2_decode.py
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
+        'python3 MaxText/decode.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset load_from_other_directory=gs://maxtext-llama/llama2-7b/decode-ckpt-scaled-maxtext load_from_other_directory_step=0 per_device_batch_size=1 model_name='llama2-7b' ici_tensor_parallelism=4 steps=1 max_prefill_predict_length=4  max_target_length=16 async_checkpointing=false'
```
rwitten: Can you move your "I love to" overrides here? There aren't really good in base.yml. (Amazing to see this running  :-))

MaxText/max_utils.py
```diff
@@ -480,3 +480,11 @@ def init_kv_cache(model, config):
   with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
     state_mesh_annotations = nn.logical_to_mesh(state_logical_annotations)
   return state_mesh_annotations
+
+def convert_state_to_jnp(state_dictionary):
+  for key, value in state_dictionary.items():
```
rwitten, line 485: oh wow this is super dangerous as an impl. can you do this as a `jax.tree_map(lambda x : jnp.array(x), state_dictionary)`?

MaxText/layers/llama2.py
```diff
@@ -0,0 +1,159 @@
...
+AttentionOp = attentions.AttentionOp
+RMSNorm = normalizations.RMSNorm
+
+
+#-----------------------------------------
+# The Decoder Layer specific for Llama2
+#-----------------------------------------
+
+
+class DecoderLayer(models.DecoderLayer):
```
rwitten, line 57: Do you really want to subclass here? (Are you using that subclassing?) I think it would be best not to encourage subclassing because we're saying there isn't reuse. If there is reuse... let's talk about it!

rwitten: Please name your class LlamaDecoderLayer

MaxText/checkpointing.py
```diff
@@ -118,8 +119,11 @@ def map_to_pspec(data, pspec):
     else:
       step = load_from_other_directory_step
       max_logging.log(f"restoring state from {load_from_other_directory} step {step}")
-    return mngr_loader.restore(step, abstract_unboxed_pre_state,
-                                      {"restore_args" : restore_args}), None
+    state = mngr_loader.restore(step, abstract_unboxed_pre_state,
+                                      {"restore_args" : restore_args})
+
+    max_utils.convert_state_to_jnp(state.params)
```
rwitten, line 125: Why did you feel a need to do this? If we're going to do this (which is arguably a good thing!), there is some other simplifications we should be making

MaxText/checkpointing.py
```diff
@@ -118,8 +119,11 @@ def map_to_pspec(data, pspec):
...
+    return state,None
```
rwitten, line 126: nit: I think `state, None`

MaxText/decode.py
```diff
@@ -70,13 +71,13 @@ def encode_strings(strs, max_len, tokenizer, mesh):
 
   for i, s in enumerate(strs):
     toks = tokenizer.tokenize(s).numpy()
-    assert toks.shape[0] < max_len, f"We aren't able to tokenize input {i}, it is too long"
+    assert toks.shape[0]-1 <= max_len, f"We aren't able to tokenize input {i}, it is too long"
```
rwitten, line 74: I hadn't understood what you meant by off by two. Are we sure we still want to remove EOS here? (Is it because we aren't setting config.use_eos to false?

MaxText/layers/models.py
```diff
@@ -145,6 +146,16 @@ class Decoder(nn.Module):
   shared_embedding: nn.Module
   mesh: Mesh
 
+  def get_decoder_layer(self):
+    if self.config.model_name == "default":
+      return DecoderLayer
+    elif self.config.model_name[0:5] == "llama":
```
rwitten, line 152: Sorry to be the fun police but llama2 please! (or if we only support llama2-7b let's say it.

assets/llama2-7b-goldens.pickle
rwitten, line 1: Do you use this? (If so, where?)

MaxText/pyconfig.py
```diff
@@ -145,18 +145,26 @@ def update_model_vars(raw_keys):
...
+        'base_num_kv_heads': 32,
         'base_mlp_dim': 11008,
         'base_num_decoder_layers': 32,
         'head_dim': 128,
-        'mlp_activations': ['silu'],
+        'mlp_activations': ['silu','linear'],
+        'vocab_size': 32000,
+        'enable_dropout': False,
+        'attention':'dot_product',
+        'vocab_relative_path':'tokenizer.llama2',
```
rwitten, line 161: where does this tokenizer come from? (Might need to be in assets or something?)

rwitten (commented):
Magic! Little changes throughout.

MaxText/pyconfig.py
A9isha: yes it is put in a GCS bucket

A9isha: should we put a comment saying explaining this?

assets/llama2-7b-goldens.pickle
A9isha: I don't use it, wanted to keep it for future reference. But you are right, perhaps doesn't need to be in github. removing it now.

.github/workflows/UnitTests.yml
rwitten: (I mean the golden evaluation and the prompt should both be replicated here.)

MaxText/decode.py
A9isha: right, we are leaving `config.use_eos=True` like it was previously. Please let me know if you think that's not the right thing.

MaxText/pyconfig.py
rwitten: But a random GCS bucket? Should we jsut put it into MaxText's assets?

MaxText/checkpointing.py
A9isha: Done - thanks!

MaxText/checkpointing.py
A9isha: If we make the weights into jax.numpy, the checkpoint converter code (MaxText/convert_llama_ckpt.py ) will not run on a v4-8 (for decoding we are using tensor parallelism to make it work on a v4-8)

Reference saxml converter:
https://github.com/google/saxml/blob/2e59631958d03b8f0858f33da185a96c7a9ddb1d/saxml/tools/convert_llama_ckpt.py

rwitten: What isn't jax.numpy.array? Everything we serialize really should be a jax.numpy.array already...

MaxText/decode.py
rwitten: I think you're saying for inference we should set config.use_eos=false?

MaxText/max_utils.py
A9isha: Done - thank you!

MaxText/convert_llama_ckpt.py
```diff
@@ -0,0 +1,340 @@
...
+  #scale the query weights
+  self_attention['query']['kernel'] = self_attention['query']['kernel']/np.sqrt(head_dim)
+
+  jax_weights['decoder']['decoder']['self_attention'] = self_attention
+
+
+
+
+
+  layer_weight['mlp']['wi_0']['kernel'] = np.array(layer_weight['mlp']['wi_0']['kernel'])
```
rwitten, line 231: Sorry my mistake was I didn't realize that these were numpys instead of jax numpy's. I think the checkpoints should be saved as jax numpys...

A9isha: if we try to do that then the checkpoint conversion will not run on a v4-8, too much memory requirement

MaxText/layers/llama2.py
A9isha: Done - thanks!

MaxText/convert_llama_ckpt.py
rwitten: I commented somewhere else but I think the orbax checkpoint should just be jnp.array's instead of np.array's.

.github/workflows/UnitTests.yml
A9isha: Done - thanks!

MaxText/convert_llama_ckpt.py
rwitten: Right now, I don't believe we will properly "load sharded"!

MaxText/pyconfig.py
A9isha: As discussed offline, we are skipping this.

MaxText/decode.py
A9isha: done - thanks!

MaxText/checkpointing.py
A9isha: converted to jax.numpy.array

MaxText/convert_llama_ckpt.py
A9isha: converted to jax.numpy.array

MaxText/convert_llama_ckpt.py
```diff
@@ -0,0 +1,346 @@
...
+  if save_training_ckpt:
+    print("Creating a training state")
+    learning_rate_schedule = max_utils.create_learning_rate_schedule(config)
+    tx = maxtext_utils.get_optimizer(config, learning_rate_schedule)
+    state_new, _ = max_utils.setup_training_state(
+      model, tx, config, init_rng, mesh, checkpoint_manager
+      )
+  else:
+    print("Creating a decoding state")
+    state_new, _ = max_utils.setup_decode_state(
```
rwitten, line 307: Can't we always save the decoding state? Like we're just hallucinating the trainstate?

rwitten: (The decoding state should work during training as a parameter state, right?)

MaxText/convert_llama_ckpt.py
```diff
@@ -0,0 +1,346 @@
...
+  jax_weights['decoder']['decoder']['pre_self_attention_layer_norm'] = layer_weight['pre_self_attention_layer_norm']
+  jax_weights['decoder']['decoder']['post_self_attention_layer_norm'] = layer_weight['post_self_attention_layer_norm']
+
+  #convert all weights to jax.numpy
+  jax_weights = jax.tree_map(jnp.array, jax_weights)
+
+  print(f"jax_weights = {jax_weights}")
+
+  #dummy configs for the checkpoint_manager
+  step_number_to_save_new_ckpt = 0
```
rwitten, line 260: Why are we involving MaxText at all at this point? It seems brittle. Can we just create an orbax checkpoint manager and checkpoint?

MaxText/decode.py
```diff
@@ -43,6 +43,7 @@
 
 import max_logging
 
+
```
rwitten, line 46: nit probably delete

MaxText/decode.py
```diff
@@ -173,6 +174,8 @@ def prefill_or_load(config, model, state, rng, sp_tokenizer, mesh, state_mesh_sh
 
 def decode_loop(config, state=None):
   """Decoding loop for the Transformer model."""
+  assert config.add_eos is False,\
+    "For decoding, we set eos to be False"
```
rwitten, line 178: "For decoding, we must set add_eos=False"

MaxText/decode.py
```diff
@@ -232,11 +237,15 @@ def decode_loop(config, state=None):
 
   new_text, _ = decode_tokens([int(x[0,0]) for x in outputs], sp_tokenizer)
   max_logging.log(f"Completion: `{config.prompt}` -> `{new_text}`")
+  if config.prompt == "I love to" and config.model_name == "llama2-7b":
```
rwitten, line 240: I don't understand this? Shouldn't it be

```
if config.autoregressive_decode_assert != "":
  assert new_text==config.autoregressive_decode_assert, f"generated text mismatch {new_text=} {config.autoregressive_decode_assert=}"
```

rwitten (changes requested):
Some nits and a more substantive comment about checkpointing that I'm happy to discuss live.

MaxText/decode.py
A9isha: done

MaxText/decode.py
A9isha: done

MaxText/decode.py
A9isha: Done

MaxText/convert_llama_ckpt.py
A9isha: removed - thank you!

MaxText/convert_llama_ckpt.py
A9isha: modified according to offline discussion - thank you!

MaxText/convert_llama_ckpt.py
```diff
@@ -0,0 +1,288 @@
...
+  layer_weight['post_self_attention_layer_norm']['scale'] = np.transpose(
+                                    layer_weight['post_self_attention_layer_norm']['scale'],
+                                    axes=(1, 0))
+
+  jax_weights['decoder']['decoder']['mlp'] = layer_weight['mlp']
+  jax_weights['decoder']['decoder']['pre_self_attention_layer_norm'] = layer_weight['pre_self_attention_layer_norm']
+  jax_weights['decoder']['decoder']['post_self_attention_layer_norm'] = layer_weight['post_self_attention_layer_norm']
+
+  #convert all weights to jax.numpy
+  jax_weights = jax.tree_map(jnp.array, jax_weights)
```
rwitten, line 238: This will likely break on llama 70B because it requires at least 2x the memory.

MaxText/decode.py
```diff
@@ -232,11 +236,16 @@ def decode_loop(config, state=None):
...
   max_logging.log(f"Completion: `{config.prompt}` -> `{new_text}`")
+  if config.autoregressive_decode_assert != "":
+    assert new_text==config.autoregressive_decode_assert, \
+    f"generated text mismatch {new_text=} {config.autoregressive_decode_assert=}"
 
 def main(argv: Sequence[str]) -> None:
   pyconfig.initialize(argv)
   os.environ["TFDS_DATA_DIR"] = pyconfig.config.dataset_path
   os.environ["TF_FORCE_GPU_ALLOW_GROWTH"] = "true"
+
```
rwitten, line 247: nit if you're going back to not add the whitespace

rwitten (approved):

PR #316, opened 2024-01-05
Fix build commands in setup.sh for head mode

SurbhiJainUSC:
This PR fixes the commands to build jaxlib from source based on the JAX developer guide: https://jax.readthedocs.io/en/latest/developer.html#building-or-installing-jaxlib


khatwanimohit (approved):
Thanks!

rwitten (approved):

PR #319, opened 2024-01-08
Support model Gamma

khatwanimohit:


rwitten (changes requested):

MaxText/layers/gamma.py
```diff
@@ -0,0 +1,135 @@
...
+
+from layers import models
+from layers import normalizations
+from layers import attentions
+from layers import initializers
+from layers import embeddings
+from layers import linears
+from layers.linears import _convert_to_activation_function
+
+Embed = embeddings.Embed
```
rwitten, line 18: BTW I don't love this thing on lines 18 to 32. No reason for you to have to do rework because of it. But please don't replicate it in the future.

MaxText/layers/models.py
```diff
@@ -151,6 +152,9 @@ def get_decoder_layer(self):
     elif self.config.model_name[0:6] == "llama2":
       from layers import llama2
       return llama2.LlamaDecoderLayer
+    elif self.config.model_name[0:5] == "gamma":
```
rwitten, line 155: Can you look for gamma- (and above for llama2-). Just imaginging when we version up gamma.

end_to_end/test_gamma.sh
```diff
@@ -0,0 +1,5 @@
+set -e
+# Test Gamma 2.5B decode
```
rwitten, line 2: Can we test the converter here?

rwitten (changes requested):

rwitten (approved):
LGTM! You can also close out my comments as you go!

PR #321, opened 2024-01-09
Simplify Jax Distributed System Initialization

rwitten:


MaxText/max_utils.py
```diff
@@ -117,44 +115,13 @@ def upload_blob(destination_gcs_name, source_file_name):
...
-    jax.distributed.initialize(coordinator_address=coordinator_address,
-                              num_processes=jax.process_count(),
-                              process_id=jax.process_index())
+  """ The best recipe to initialize the Jax Distributed System has varied over time. We keep a layer of
+      indirection in MaxText to avoid breaking the call sites unnecessarily.
 
+      Currently jax.distributed.initialize() fully works as expected!
+  """
   max_logging.log("Attempting to initialize the jax distributed system...")
-  if jax.__version__ >= '0.4.21':
```
gobbleturk, line 154: This change drops MaxText support for old versions of Jax - just confirming this is what we want?

gobbleturk (approved):

MaxText/max_utils.py
rwitten: yes -- maxtext only supports ~HEAD.

PR #322, opened 2024-01-09
Simplified RNG handling (https://github.com/google/maxtext/issues/276)

rwitten:


MaxText/train.py
```diff
@@ -305,13 +303,16 @@ def train_loop(config, state=None):
...
     raise ValueError("Profiling requested but initial profiling step set past training final step")
   last_profiling_step = np.clip(first_profiling_step + config.profiler_steps - 1, first_profiling_step, config.steps - 1)
+
   for step in np.arange(start_step, config.steps):
     if step == first_profiling_step:
       max_utils.activate_profiler(config)
 
     example_batch = load_next_batch(data_iterator, example_batch, config)
+
+    nextrng = jax.random.fold_in(init_rng, start_step)
```
gobbleturk, line 313: I think we want jax.random.fold_in(init_rng, step) so it generates new rng each step

MaxText/train.py
```diff
@@ -209,20 +208,20 @@ def loss_fn(params):
   if config.record_internal_nn_metrics:
     record_activation_metrics(metrics, intermediate_outputs, config)
 
-  return new_state, metrics, rng2
+  return new_state, metrics
```
gobbleturk, line 211: You should change `[get_functional_train_with_signature](https://github.com/google/maxtext/blob/65d1d164716be5ab820b1f0d9a2d946297056399/MaxText/maxtext_utils.py#L42C32-L42C32)` to remove the rng output - I think AOT needs this. Maybe we need IF CHANGE THIS CODE THEN CHANGE THAT CODE comments for changing train step signature

gobbleturk (approved):

MaxText/train.py
rwitten: OMG thank you.

PR #327, opened 2024-01-11
Upload config to GCS and tensorboard

raymondzouu:
* Adding flag `save_config_to_gcs`. When it is set to `True`, MaxText will upload the config params to GCS as `config.yml`.
* Write config params and libtpu_init_args to tensorboard

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -52,6 +52,9 @@ metrics_file: "" # for testing, local file that stores scalar metrics. If empty,
 # If true save metrics such as loss and TFLOPS to GCS in {base_output_directory}/{run_name}/metrics/
 gcs_metrics: False
 
+# If true save config to GCS in {base_output_directory}/{run_name}/
+save_config_to_gcs: False
+
 # Activation dtypes.
 dtype: "bfloat16"
 int8_training: False
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -29,6 +29,7 @@ from jax.experimental import mesh_utils
 
 
 import json
+import yaml
 import flax
 from flax.training import train_state
 from flax import linen as nn
@@ -104,6 +105,17 @@ def write_metrics_locally(metrics, step, config, file):
   if step == config.steps - 1:
     file.close()
 
+def add_config_to_summary_writer(config, summary_writer):
+  """Writes config params to tensorboard"""
+  if jax.process_index() == 0:
+    for key, value in config.get_keys().items():
+      add_text_to_summary_writer(key, str(value), summary_writer)
+
+def add_text_to_summary_writer(key, value, summary_writer):
+  """Writes given key-value pair to tensorboard as text/summary"""
+  if jax.process_index() == 0:
+    summary_writer.add_text(key, value)
+
 def write_metrics_for_gcs(metrics, step, config, running_metrics):
   """Writes metrics to gcs"""
   metrics_dict_step = _prepare_metrics_for_json(metrics, step, config.run_name)
@@ -122,6 +134,23 @@ def write_metrics_for_gcs(metrics, step, config, running_metrics):
     max_logging.log(f"File {metrics_filename} moved successfully!")
     running_metrics = [] # reset running_metrics to empty list
   return running_metrics
+
+def write_config_raw_keys_for_gcs(raw_keys):
+  """Writes config raw keys to GCS"""
+  if not raw_keys["save_config_to_gcs"] or jax.process_index() != 0:
+    return
+  max_logging.log("Writing config to GCS...")
+
+  raw_keys_dict = dict(raw_keys)
+  filename = "config.yml"
+  with open(filename, 'w', encoding="utf8") as config_for_gcs:
+    yaml.dump(raw_keys_dict, config_for_gcs)
+  config_for_gcs.close()
+
+  gcs_filename=os.path.join(raw_keys["base_output_directory"], raw_keys["run_name"], filename)
+  max_logging.log(f"Moving file {filename} to GCS...")
+  upload_blob(gcs_filename, filename)
+  max_logging.log(f"File {filename} moved successfully!")
 
 def parse_gcs_bucket_and_prefix(destination_gcs_name):
   path_parts = destination_gcs_name.replace("gs://", "").split("/")
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -315,6 +315,11 @@ def train_loop(config, state=None):
   max_logging.log(f"number parameters: {num_model_parameters/10**9:.3f} billion")
   per_device_tflops = calculate_training_tflops(num_model_parameters, config)
 
+  # Write train config params, num model params, and XLA flags to tensorboard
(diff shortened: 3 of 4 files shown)


MaxText/max_utils.py
```diff
@@ -100,6 +100,19 @@ def write_metrics_for_gcs(metrics, step, config, running_metrics):
     running_metrics = [] # reset running_metrics to empty list
   return running_metrics
 
+def write_train_config_for_gcs(config):
+  """Writes train config to gcs"""
+  train_config_filename = "train_config.txt"
+  with open(train_config_filename, 'w', encoding="utf8") as train_config_for_gcs:
```
gobbleturk, line 106: Nit: Is it possible to instead write the config in yaml format so it can be used in place of `configs/base.yml` for future runs? IMO this would slightly preferrable - although perhaps format is more convenient for downstream tasks you have in mind.

gobbleturk:
I've been wanting this feature for some time, thank you!

gobbleturk (approved):

MaxText/max_utils.py
gobbleturk: I think `.txt` is fine, thanks for looking into it.

MaxText/configs/base.yml
```diff
@@ -42,6 +42,9 @@ metrics_file: "" # for testing, local file that stores scalar metrics. If empty,
 # If true save metrics such as loss and TFLOPS to GCS in {base_output_directory}/{run_name}/metrics/
 gcs_metrics: False
 
+# If true save train config to {base_output_directory}/{run_name}/
+gcs_train_config: False
```
rwitten, line 46: This change is very reasonable but should we just print to stdout (as well or maybe instead?) Like for 90% of use cases that is what the people need.

MaxText/max_utils.py
rwitten: Are we sure this is hard? What did you try?

rwitten (commented):
A couple ideas.

MaxText/configs/base.yml
rwitten: actually I'll take care of this, NVM

MaxText/max_utils.py
raymondzouu: I changed it so that it outputs as a yml now. Here is an example of what the output yml looks like gpaste/5947072564428800. When it exports to yml, it uses a slightly different notation for specifying lists but I confirmed that it generates the same config when passed back into MaxText.

MaxText/train.py
```diff
@@ -289,6 +289,12 @@ def train_loop(config, state=None):
   max_logging.log(f"number parameters: {num_model_parameters/10**9:.3f} billion")
   per_device_tflops = calculate_training_tflops(num_model_parameters, config)
 
+  # Write train config params and number of model params to tensorboard
```
raymondzouu, line 292: Is it ok to write the train config params to tensorboard by default or should it be put behind a flag?

rwitten: Seems fine.

MaxText/train.py
```diff
@@ -289,6 +289,13 @@ def train_loop(config, state=None):
   max_logging.log(f"number parameters: {num_model_parameters/10**9:.3f} billion")
   per_device_tflops = calculate_training_tflops(num_model_parameters, config)
 
+  # Write train config params, num model params, and XLA flags to tensorboard
+  if jax.process_index() == 0:
```
rwitten, line 293: please hide this in a funciton.

MaxText/train.py
```diff
@@ -359,6 +366,9 @@ def main(argv: Sequence[str]) -> None:
   pyconfig.initialize(argv)
   config = pyconfig.config
   validate_train_config(config)
+  if config.gcs_train_config and jax.process_index() == 0:
```
rwitten, line 369: hide this in a function! you can put the if in `write_train_config_for_gcs`

MaxText/max_utils.py
```diff
@@ -115,6 +119,23 @@ def write_metrics_for_gcs(metrics, step, config, running_metrics):
     running_metrics = [] # reset running_metrics to empty list
   return running_metrics
 
+def write_train_config_for_gcs(config):
```
rwitten, line 122: lol this is hilarious and pragmatic. but can we just copy.deepcopy before user_init right?

rwitten (changes requested):
Almost perfect but this `_tuples_to_lists` inversion is a bit too spicy for my taste. Can you just intercept earlier in pyconfig?

MaxText/train.py
raymondzouu: Done!

MaxText/train.py
raymondzouu: Done!

MaxText/max_utils.py
raymondzouu: Haha yeah it is. We could upload the config [here](https://github.com/google/maxtext/blob/a54a940d3a2b7bf2851289ef8b7335ffbeacb751/MaxText/pyconfig.py#L174) before the `_list_to_tuples` conversion happens so that we don't have to convert back however then the config that gets uploaded would not have the initialized values like `emb_dim`, `num_query_heads`, etc.

MaxText/configs/base.yml
```diff
@@ -52,6 +52,9 @@ metrics_file: "" # for testing, local file that stores scalar metrics. If empty,
...
```
rwitten, line 56: I don't understand this flag name? Don't we want to save decode there as well?

raymondzouu: Thanks for pointing this out, I changed the flag name and added the config uploading to `decode.py` as well.

MaxText/decode.py
```diff
@@ -266,6 +266,7 @@ def main(argv: Sequence[str]) -> None:
   os.environ["TF_FORCE_GPU_ALLOW_GROWTH"] = "true"
 
   validate_config(pyconfig.config)
+  max_utils.write_config_for_gcs(pyconfig.config)
```
rwitten, line 269: Sorry this should be in pyconfig right? We don't want to duplicate it everywhere?

rwitten (approved):
Approved with a tiny nit.

MaxText/max_utils.py
raymondzouu: Sounds good, I changed it to upload the config first as `base_config.yml` and then again as `initialized_config.yml` after the values get initialized.

MaxText/decode.py
raymondzouu: Done. Changed it to upload in pyconfig.

MaxText/max_utils.py
raymondzouu: Actually, I changed it to just upload once in pyconfig after the values are initialized but before the type conversion. So now the config we upload will have all of the values and we won't have to do a `_tuples_to_lists`.

raymondzouu:
These changes were merged in https://github.com/google/maxtext/commit/2fdf8a2dfa553b928d52052711219ca9cbcf4ec7. Closing out this PR now.

PR #328, opened 2024-01-12
Letting users pass args via M_*

rwitten:
Now you can pass args via setting an env var M_ARG_NAME.


MaxText/pyconfig.py
```diff
@@ -61,9 +65,17 @@ def _lists_to_tuples(l: list[Any]) -> Union[tuple[Any],list[Any]]:
 
 class _HyperParameters():
   # pylint: disable=missing-class-docstring
+  def _validate_env_variables(self, raw_data_from_yaml):
+    for environment_var in os.environ:
+      if environment_var[:4] == _MAX_PREFIX:
```
raymondzouu, line 70: I think this is supposed to be `environment_var[:2]` to check for `"M_"`

rwitten: yes thank you Raymond, fixed on new revision. I changed the name late. Nice eagle eyes calling out both the unnecessary constant and the wrong constant!

(p.s. -- I woke up at midnight because I dreamt of this bug. amazing that you had already caught it when I woke up!)

gobbleturk (approved):

raymondzouu (approved):

PR #335, opened 2024-01-12
[MAXTEXT] use split and concat directly instead of reshape and transpose and slice and concat

michelle-yooh:
[MAXTEXT] use split and concat directly instead of reshape and transpose and slice and concat


rwitten:
Thank you for demonstrating this Michelle.

PR #337, opened 2024-01-13
[DO NOT MERGE] Goodput prototype v1

dipannita08:
Quick prototype to compute Goodput based on total step time and job time on MaxText.

### DIFF
--- a/MaxText/goodput.py
+++ b/MaxText/goodput.py
@@ -1,0 +1,127 @@
+"""
+Goodput Prototype (TODO: Move to packaged library)
+"""
+
+import os
+import max_logging
+import time
+import jax
+import datetime
+import json
+from google.cloud import storage
+import logging
+logger = logging.getLogger(__name__)
+step_time_start_dict = {}
+step_time_filename = f"goodput_step_times.txt"
+job_runtime_filename = f"goodput_job_runtime.txt"
+
+def should_log_goodput(config):
+  """Determine whether or not to enable Goodput logging"""
+  if jax.process_index == 0 and config.enable_goodput_metrics:
+      return True
+  return False
+
+def get_goodput_dir(config):
+  """Derive the goodput directory where logs will be stored"""
+  goodput_dir = os.path.join(config.base_output_directory, config.run_name, "goodput", "")
+  return goodput_dir
+
+def get_gcs_bucket_and_blob_name(destination_gcs_name):
+  """Derive the GCS blob and bucket name"""
+  path_parts = destination_gcs_name.replace("gs://", "").split("/")
+  bucket = path_parts.pop(0)
+  blob = "/".join(path_parts)
+  return bucket, blob
+
+def upload_file_to_gcs(destination_gcs_name, source_file_name):
+  """Upload a file to a GCS location"""
+  try:
+    bucket_name, blob_name = get_gcs_bucket_and_blob_name(destination_gcs_name)
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(bucket_name)
+    blob = bucket.blob(blob_name)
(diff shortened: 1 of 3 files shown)


MaxText/goodput.py
```diff
@@ -0,0 +1,127 @@
+"""
+Goodput Prototype (TODO: Move to packaged library)
```
rwitten, line 2: Also why is this writing to GCS? Is there some metric sservice we can use?

MaxText/train.py
```diff
@@ -310,10 +312,13 @@ def train_loop(config, state=None):
       max_utils.activate_profiler(config)
 
     example_batch = load_next_batch(data_iterator, example_batch, config)
+    step_start_time = datetime.datetime.now()
     with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
       state, metrics, nextrng = p_train_step(
           state, example_batch, nextrng
       )
+    step_stop_time = datetime.datetime.now()
```
rwitten, line 320: This code doesn't do what you think it does -- Jax calls are non blocking. I think it would better to provide a goodput.goodput_log_step_time call that took step time (and inside of it you can get a datetime).

MaxText/goodput.py
rwitten: Writing to GCS is a lot more complicated that it seems so we'd need to think about that carefully.

MaxText/goodput.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.error("Error in uploading file to GCS location.", e)
+
+def download_file_from_gcs(source_gcs_name, destination_file_name):
+   """Download a file from a GCS location"""
+   try:
+      bucket_name, blob_name = get_gcs_bucket_and_blob_name(source_gcs_name)
+      storage_client = storage.Client()
+      bucket = storage_client.bucket(bucket_name)
+      blob = bucket.blob(blob_name)
+      blob.download_to_filename(destination_file_name)
```
rwitten, line 54: BTW this type of logic is unwelcome on our control thread! We need to be writing asynchronously to something that gracefully handles appends.

(GCS really doesn't like appends. And this code seems to be synchronous on our control thread.)

MaxText/train.py
```diff
@@ -356,9 +361,13 @@ def main(argv: Sequence[str]) -> None:
       stack_trace_to_cloud = config.stack_trace_to_cloud,
       stack_trace_interval_seconds = config.stack_trace_interval_seconds))
   diagnostic_config = diagnostic_configuration.DiagnosticConfig(debug_config)
+  job_start_time = datetime.datetime.now()
   with diagnostic.diagnose(diagnostic_config):
     train_loop(config)
-
+  job_end_time = datetime.datetime.now()
+  goodput.goodput_log_job_runtime(config, job_end_time - job_start_time)
```
rwitten, line 368: Also you really don't want to pass a config outside of MaxText into your generic interface! Pass what you need!

MaxText/train.py
```diff
@@ -356,9 +361,13 @@ def main(argv: Sequence[str]) -> None:
...
+  total_goodput = goodput.get_job_goodput(config)
+  max_logging.log(f"Total job goodput: {total_goodput:.2f}%")
```
rwitten, line 370: Are we sure we want to print this? Because this run doesn't know the overall goodput -- the goodput depends on fusing multiple runs together.

rwitten (changes requested):
Excited for goodput! But lots to discuss here.

MaxText/goodput.py
xuefgu: +1, I think we can readily use Cloud Logging to stream logs: https://cloud.google.com/logging/docs/write-query-log-entries-python#linux

PR #338, opened 2024-01-13
Single v4-8 step time from 1.209 to 1.199 secs. (No convergence data …

rwitten:
…since it is actually bit-wise identical locally)

Optimiization From Blake!


rwitten:
weird mnkey patch doesn't work.

PR #341, opened 2024-01-14
Checkpointing Organization

rwitten:
Modified Checkpoint Loading So A Full State Checkpoint or a Parameter Checkpoint Are Compatible With Both Train and Decode.


.github/workflows/UnitTests.yml
```diff
@@ -76,11 +76,11 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset int8_training=true steps=2 enable_checkpointing=false'
-    - name: Test test_decode_checkpoint
+    - name: Test generate_parameter_checkpoint
```
singh-mitali, line 79: Nit - Change this to generate_params_only_checkpoint?

MaxText/checkpointing.py
```diff
@@ -56,9 +56,8 @@ def create_orbax_checkpoint_manager(
 
 
 def load_state_if_possible(checkpoint_manager: CheckpointManager,
```
singh-mitali, line 58: It will be cleaner to split this function  - read from checkpoint_manager or read from a path. Also do not like the name if_possible. Should be handled cleanly by caller code.

MaxText/checkpointing.py
```diff
@@ -82,44 +82,47 @@ def load_state_if_possible(checkpoint_manager: CheckpointManager,
...
+      return checkpoint_manager.restore(latest_step, abstract_unboxed_pre_state,
+                                        {"restore_args" : restore_args}), None
     else:
-      step = load_from_other_directory_step
-      max_logging.log(f"restoring state from {load_from_other_directory} step {step}")
-    return mngr_loader.restore(step, abstract_unboxed_pre_state,
-                                      {"restore_args" : restore_args}), None
+      max_logging.log("failed to find preexisting checkpoint for this run")
+
+  if load_parameters_from_path != "":
```
singh-mitali, line 107: Maybe we only have checkpoint_path as variable for inference and decode scripts. Another parameter checkpoint_converter_output_path for any conversion scripts.

Additionally code can pass checkpoint_type as an enum to distinguish between the various types of checkpoints and the read/write code can handle accordingly.

singh-mitali (commented):
Added some comments on overall better organization.

MaxText/checkpointing.py
rwitten: I think if_possible is a pretty common convention! There are lots of call sites so this gives clear semantics. Either way, I don't think you're commenting on the CR here?

MaxText/checkpointing.py
rwitten: Agreed checkpoint type as an enum will be a good idea if the complexity keeps rising. But you're not really commenting on changes in this CR right?

singh-mitali: Yes can be part of a subsequent change.

MaxText/checkpointing.py
singh-mitali: Yes - not related to this particular CR - but overall code.

singh-mitali (approved):
Overall LGTM - not familiar with the scan/unrolling of params. So skipping that for now.

PR #346, opened 2024-01-17
Fix M_ error handling

raymondzouu:


MaxText/pyconfig.py
```diff
@@ -70,8 +70,8 @@ def _validate_env_variables(self, raw_data_from_yaml):
       if environment_var[:len(_MAX_PREFIX)] == _MAX_PREFIX:
         proposed_key = environment_var[len(_MAX_PREFIX):].lower()
         if proposed_key not in raw_data_from_yaml:
-          raise ValueError(f"We received env `{environment_var}` but it doesn't match a key, so it is aassumed a mistake")
-        if proposed_key.upper() != proposed_key:
+          raise ValueError(f"We received env `{environment_var}` but it doesn't match a key, so it is assumed a mistake.")
+        if not environment_var[len(_MAX_PREFIX):].isupper():
           raise ValueError(f"We received env `{proposed_key}` but it isn't all uppercase.")
```
rwitten, line 75: Is this is the right error message? I think we want:
f"We received env `{environment_var}` but it isn't all uppercase."

rwitten (changes requested):
Thanks for fixing this! But I think the error message isn't quite right?

MaxText/pyconfig.py
raymondzouu: Done.

rwitten (approved):

PR #347, opened 2024-01-17
Standalone checkpointer for Maxtext training emulator for storage

RoshaniN:
- The standalone checkpointer sets up the model and checkpoint manager, similar to train_loop of train.py. 
- The checkpointer saves the checkpoint at the designated steps and restores them (if needed).
- It mimics the storage interactions of the TPU VM hosts with storage systems such as GCS and logs the time for checkpoint saves and restores.
- Added a github workflow unit test for running standalone_checkpointer.py separately.

Similar to https://github.com/google/maxtext/pull/315

### DIFF
--- a/MaxText/standalone_checkpointer.py
+++ b/MaxText/standalone_checkpointer.py
@@ -1,0 +1,107 @@
+"""
+ Copyright 2023 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+# pylint: disable=g-bad-todo, abstract-method, consider-using-with, ungrouped-imports
+"""Standalone checkpointer - only saves and restores checkpoints at regular intervals, accesses storage needs."""
+
+# Calling jax.device_count here prevents a "TPU platform already registered" error.
+# See github.com/google/maxtext/issues/20 for more
+
+import datetime
+import os
+
+from typing import Sequence
+from absl import app
+from flax.linen import partitioning as nn_partitioning
+import jax
+import numpy as np
+
+import checkpointing
+import max_utils
+import max_logging
+import pyconfig
+from train import setup_mesh_and_model, get_first_step, validate_train_config
+
+from layers import models
+
+from jax.experimental.compilation_cache import compilation_cache as cc
+
+Transformer = models.Transformer
+
+def checkpoint_loop(config, state=None):
+  """Main Checkpointing loop.
+  Saves checkpoints.
+  Args:
+    config:
+    state:
+    ckpt_path:
+  Returns:
+  """
+  init_rng, writer, checkpoint_manager, mesh, model, _, tx = setup_mesh_and_model(config)
+
+  unboxed_abstract_state, state_mesh_annotations = max_utils.get_abstract_state(model, tx,
+                                                config, init_rng, mesh, is_training=True)
+  # A barrier to sync all hosts before starting to restore checkpoint
+  jax.experimental.multihost_utils.sync_global_devices("dummy1")
+  checkpoint_load_start = datetime.datetime.now()
+  with nn_partitioning.axis_rules(config.logical_axis_rules):
+    state, _ = checkpointing.load_state_if_possible(checkpoint_manager,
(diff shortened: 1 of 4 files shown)


MaxText/standalone_checkpointer.py
```diff
@@ -0,0 +1,99 @@
...
+  init_rng = random.PRNGKey(config.init_weights_seed)
+  ( writer, checkpoint_manager, _, _, _, _, _, state) = setup_train_loop(config, init_rng, log_checkpoint_restore_time=True)
+  start_step = get_first_step(state) # this is the start_step for training
+
+  for step in np.arange(start_step, config.steps):
+    start_time = datetime.datetime.now()
+    if checkpoint_manager is not None:
+      if checkpoint_manager.save(step, state):
+        max_logging.log(f"saved a checkpoint at step {step}")
+        checkpoint_manager.wait_until_finished()
```
RoshaniN, line 72: @rwitten and @gobbleturk - need some help here. I was hoping that this function is the Orbax equivalent of jax.block_until_ready(), but I am not quite sure. I see multiple logs , for each of the hosts.

rwitten: wait_until_finished() shoudl block (and DEFINITELY should block if async_checkpointing=false). Is there a problem?

MaxText/checkpointing.py
```diff
@@ -75,6 +77,7 @@ def load_state_if_possible(checkpoint_manager: CheckpointManager,
     mesh: a physical TPU mesh
     state_mesh_annotation: a PyTree of sharding rules, matching
       abstract_unboxed_pre_state.
+    log_checkpoint_restore_time: True, to log checkpoint restore time.
```
rwitten, line 80: I don't love making changes that make key functions hrader to read and more complicated.

Can't the wait_until_finished give you this feature?

rwitten (changes requested):
This would be a last resort solution! Let's understand why you can't wait_until_completed in orbax...

MaxText/standalone_checkpointer.py
RoshaniN: I was hoping save() to be blocking . I found out that wait_until_finished() blocks in case of async checkpoint. 
Even with jax.experimental.multihost_utils.sync_global_devices("dummy") before the start of checkpointing, I see differences in the order of microseconds between the hosts. 

I decided not to be nit-picky about this at the moment.

MaxText/checkpointing.py
RoshaniN: I removed the intrusive logging comments. Thanks for your suggestions offline, to try and use load_state_if_possible. I was able to get it working.

rwitten (approved):

RoshaniN:
Rafi, took your suggestion to add a barrier before starting and using wait_until_finished after starting. Thanks!

MaxText/standalone_checkpointer.py
```diff
@@ -0,0 +1,104 @@
...
+  else: # Checkpoint was unavailable, state needs to be initialized
+    state, state_mesh_annotations = max_utils.setup_training_state(model,
+          tx, config, init_rng, mesh, checkpoint_manager)
+
+  start_step = get_first_step(state) # this is the start_step for training
+  for step in np.arange(start_step, config.steps):
+    if checkpoint_manager is not None:
+      start_time = datetime.datetime.now()
+      # A barrier to sync all hosts before starting to checkpoint
+      jax.experimental.multihost_utils.sync_global_devices("dummy")
```
RoshaniN, line 78: Added the barrier here

MaxText/standalone_checkpointer.py
RoshaniN: Added now!

MaxText/standalone_checkpointer.py
```diff
@@ -0,0 +1,107 @@
...
+  jax.experimental.multihost_utils.sync_global_devices("dummy")
+  checkpoint_load_start = datetime.datetime.now()
+  with nn_partitioning.axis_rules(config.logical_axis_rules):
+    state, _ = checkpointing.load_state_if_possible(checkpoint_manager,
+                                                config.load_parameters_path,
+                                                config.load_full_state_path,
+                                                unboxed_abstract_state,
+                                                mesh,
+                                                state_mesh_annotations)
+  jax.experimental.multihost_utils.sync_global_devices("dummy")
```
rwitten, line 67: No a sync doesn't work here -- you need a block_until_ready. Sync makes sure everyone is executing the same python, block_until_ready means the tensor is actually ready. Jax will by default not worry about it.

Also, you need to use unique strings for all your different sync_global_devices -- not super important in this case, but you don't want to accidentally sync different barriers.

RoshaniN: Thanks for explaining, done!

rwitten (approved):
In the future, please give descriptive names in sync_global_devices. For example "Before Loading State", etc.

Feel free to fix this next time you're touching the file.

PR #348, opened 2024-01-17
Collect metrics for GCS tessellation

llukai:
Creating a new branch for GCS Tessellation.


MaxText/standalone_dataloader.py
```diff
@@ -47,15 +47,15 @@ def data_load_loop(config, state=None):
   example_batch = load_next_batch(data_iterator, example_batch, config)
   jax.block_until_ready(example_batch)
   first_end = datetime.datetime.now()
-  time_to_load_first_batch = first_end-start
+  time_to_load_first_batch = (first_end-start).total_seconds()
```
rwitten, line 50: Do you want to print this number on ALL VMs? And you might want to only log this on `jax.process_index()==0`, that is up to you!

Also, I would recommend you put a jax.mulithost_utils.sync_global_devices before your start (and before you start data loading initialization) so you don't count "jitter" in the difference from the first host to boot to the last host to boot against you. Imagine if 1 host joins the job 60 seconds late but then the data is ready 1 second after that.

rwitten (changes requested):
Broadly good but some thoughts.

PR #351, opened 2024-01-18
Add Mistral 7B inference

RissyRan:
# Description

Add Mistral 7B inference:
* Refactor `convert_llama2_ckpt.py` to `llama_or_mistral_ckpt.py` and add mistral-7b config
* Refactor `pyconfig.py` file and add yml config files for models
* Add unit test for the model
* Other small changes, i.e. exclude hidden files when finding checkpoints 

# Test

I tested by running following commands:

1. Run script below to download the checkpoint, convert it, and run decoding script

```
export M_ENABLE_CHECKPOINTING=true
export M_BASE_OUTPUT_DIRECTORY=gs://runner-maxtext-logs
export M_DATASET_PATH=gs://maxtext-dataset
export M_ASYNC_CHECKPOINTING=false

pip3 install torch
gsutil -m cp -r gs://maxtext-external/mistral-7B-v0.1 /tmp
python3 MaxText/llama_or_mistral_ckpt.py --base-model-path /tmp/mistral-7B-v0.1 --model-size mistral-7b --maxtext-model-path gs://maxtext-mistral/test/${idx}/decode-ckpt-maxtext/
python3 MaxText/decode.py MaxText/configs/base.yml load_parameters_path=gs://maxtext-mistral/test/${idx}/decode-ckpt-maxtext/0/default run_name=runner_direct_${idx} per_device_batch_size=1 model_name='mistral-7b' assets_path=gs://maxtext-external/mistral-7B-v0.1 ici_tensor_parallelism=4 max_prefill_predict_length=4  max_target_length=16 prompt="I love to" autoregressive_decode_assert="read. I love to read about the Bible. I love" attention=dot_product
```

2. See successful result on v4-8

```
...
Config param vocab_size: 32000
Config param warmup_steps_fraction: 0.1
I0118 08:21:50.833148 140425765660672 mesh_utils.py:320] _create_device_mesh_for_nd_torus assignment: [(), (), (), (0, 1), (2,)]
Num_devices: 4, shape (1, 1, 1, 4, 1)
I0118 08:21:50.998879 140425765660672 dataset_info.py:610] Load dataset info from gs://maxtext-dataset/c4/en/3.0.1
I0118 08:21:51.225100 140425765660672 dataset_info.py:702] For 'c4/en/3.0.1': fields info.[splits] differ on disk and in the code. Keeping the one from code.
I0118 08:21:51.417470 140425765660672 reader.py:261] Creating a tf.data.Dataset reading 1024 files located in folders: gs://maxtext-dataset/c4/en/3.0.1.
I0118 08:21:51.549226 140425765660672 logging_logger.py:49] Constructing tf.data.Dataset c4 for split train, from gs://maxtext-dataset/c4/en/3.0.1
I0118 08:21:51.657563 140425765660672 dataset_info.py:610] Load dataset info from gs://maxtext-dataset/c4/en/3.0.1
I0118 08:21:51.878630 140425765660672 dataset_info.py:702] For 'c4/en/3.0.1': fields info.[splits] differ on disk and in the code. Keeping the one from code.
I0118 08:21:51.913564 140425765660672 reader.py:261] Creating a tf.data.Dataset reading 8 files located in folders: gs://maxtext-dataset/c4/en/3.0.1.
I0118 08:21:51.951487 140425765660672 logging_logger.py:49] Constructing tf.data.Dataset c4 for split validation, from gs://maxtext-dataset/c4/en/3.0.1
Model path: gs://maxtext-external/mistral-7B-v0.1/tokenizer.model
normalizer.cc(51) LOG(INFO) precompiled_charsmap is empty. use identity normalization.
restoring params from load_parameters_from_path='gs://maxtext-external/mistral-7B-v0.1-maxtext/0/default'
I0118 08:21:53.870312 140425765660672 checkpointer.py:164] Restoring item from gs://maxtext-external/mistral-7B-v0.1-maxtext/0/default.
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1705566114.153335  115786 gcs_resource.cc:99] Using default AdmissionQueue with limit 32
I0000 00:00:1705566114.159234  118364 google_auth_provider.cc:179] Running on GCE, using service account 630405687483-compute@developer.gserviceaccount.com
I0118 08:22:33.818854 140425765660672 checkpointer.py:167] Finished restoring checkpoint from gs://maxtext-external/mistral-7B-v0.1-maxtext/0/default.
/home/ranran/.local/lib/python3.10/site-packages/jax/_src/numpy/lax_numpy.py:2206: FutureWarning: None encountered in jnp.array(); this is currently treated as NaN. In the future this will result in an error.
  return array(a, dtype=dtype, copy=False, order=order)  # type: ignore
/home/ranran/.local/lib/python3.10/site-packages/jax/_src/ops/scatter.py:96: FutureWarning: scatter inputs have incompatible types: cannot safely cast value from dtype=float32 to dtype=int32 with jax_numpy_dtype_promotion='standard'. In future JAX releases this will result in an error.
  warnings.warn(
Number of model params=7.242 billion, memory usage=13.489GB, bytes per param=2.000
Computed prefill cache 
Number of cache entries=0.004 billion, memory usage=0.008GB, bytes per cache=2.001
Total memory (for cache and params) 13.497 GB
Completion: `I love to` -> `read. I love to read about the Bible. I love`
Did 10 steps in 0.134 seconds for 4.0 sequences with a total memory footprint of 13.497 GB
Therefore, a per-generate time of 0.0134 seconds, a throughput of 298.9 tok/s and 252.1 GB/s/device
```


MaxText/max_utils.py
```diff
@@ -163,6 +163,7 @@ def create_device_mesh(config, devices=None):
   except:
     num_slices = 1
   num_devices_per_slice = num_devices//num_slices
+  max_logging.log(f"{num_devices=}, {num_slices=}, and {num_devices_per_slice=}")
```
rwitten, line 166: Probably don't do this!

MaxText/convert_ckpt.py
```diff
@@ -11,16 +11,16 @@
  limitations under the License.
  """
```
rwitten, line 12: Can this one be llama_or_mistral_ckpt? Because other checkpoints have other paths.

MaxText/max_utils.py
rwitten: Or if you do, let me know why and possibly make a clearer summary

requirements.txt
```diff
@@ -17,3 +17,4 @@ tensorflow-text
 tensorboardx
 tensorflow
 tensorboard-plugin-profile
+torch
```
rwitten, line 20: This can't go in our dependency closure -- we don't want folks to install torch to dev. It is useful only in this weird situation. Might be worth adding a check to the convert script though?

rwitten (changes requested):
Can you add a test (similar to the llama_decode test) to ensure this is working in on some golden data?

MaxText/max_utils.py
WendyShang: ah sorry i printed the devices out to debug and i think that confused Ran. sorry my bad!

RissyRan:
> Can you add a test (similar to the llama_decode test) to ensure this is working in on some golden data?

Sure, I could add that.

MaxText/max_utils.py
RissyRan: I don't have a strong reason, but I followed the format [here]( https://github.com/google/maxtext/blob/cf100d56c548e4fed8d92557c1c86d5b2e50e640/MaxText/layers/models.py#L155). Is there any reason against it? Thanks!

MaxText/convert_ckpt.py
RissyRan: Yes, of course. Will update it in next commit.

requirements.txt
RissyRan: Makes sense. I will add a check to do that.

MaxText/max_utils.py
RissyRan: > ah sorry i printed the devices out to debug and i think that confused Ran. sorry my bad!

No worries. I thought this will be helpful. But I could just delete it if you agree.

MaxText/pyconfig.py
```diff
@@ -194,6 +198,25 @@ def update_model_vars(raw_keys):
         'add_eos': False
       }
       raw_keys = validate_and_update_keys(raw_keys, llama2_7b_model_vars)
+    elif raw_keys['model_name'] == _MISTRAL_7B:
```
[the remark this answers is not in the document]
RissyRan: I see there are some scripts to overwrite those keys, and one example [here](https://github.com/google/maxtext/blob/main/MaxText/configs/v4/22b.sh#L47). I will verify if this works for Mistral/Llama case, if so, the configs here could be the default one, and any param passing in will overwrite them.

MaxText/convert_ckpt.py
A9isha: and we could override any config including say tokenizer path here: https://github.com/google/maxtext/blob/main/MaxText/pyconfig.py#L176-L195

MaxText/pyconfig.py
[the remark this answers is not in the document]
A9isha: The commandline override is for temporary overriding during a run/test for example. We would need to persist the configs for models like Llama2/Mistral. I agree with Wendy that adding more models will need us to add another list of params. Basically, today the default model's configs are listed in [base.yml](https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml) and the overrides for other models are listed in this pyconfig.py and is determined by the `model_name` param. Do you recommend creating separate yml files for separate models and read them into pyconfig.py? That should be possible.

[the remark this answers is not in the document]
WendyShang: yep i think that is a cleaner way than changing pyconfig.py! Thanks!!

RissyRan:
> Can you add a test (similar to the llama_decode test) to ensure this is working in on some golden data?

Sure. Added unit test for Mistral to test decode part, and also add TODO to add training and fine-tuning tests. Please let me know if we want to add other tests at this moment.

MaxText/convert_ckpt.py
RissyRan: Hi @WendyShang, do you mind explaining a little bit more when you say `give a decode.py example to signify that we are using mistral's tokenizer`? Thank you!

end_to_end/mistral_load_and_test.sh
```diff
@@ -0,0 +1,18 @@
...
+idx=$(date +%Y-%m-%d-%H-%M)
+
+export base_ckpt_path=gs://maxtext-external/mistral-7B-v0.1-maxtext/0/default
+export M_ENABLE_CHECKPOINTING=true
+export M_BASE_OUTPUT_DIRECTORY=gs://runner-maxtext-logs
+export M_DATASET_PATH=gs://maxtext-dataset
+export M_ASYNC_CHECKPOINTING=false
+
+#pip install torch
+#wget -P /tmp https://files.mistral-7b-v0-1.mistral.ai/mistral-7B-v0.1.tar
```
rwitten, line 11: Look we don't want the wget -- it would be best to use our own mirror. But otherwise we want to generate this checkpoint. The promise to the customer is we can generate this checkpoint!

rwitten (commented):
Looks really good, just a question of waht we're testing.

MaxText/convert_ckpt.py
RissyRan: Thanks Wendy! Like we discussed offline, I will keep this in mind, and see how we could improve further when adding Mixtral inference. Resolve this for now.

end_to_end/mistral_load_and_test.sh
RissyRan: Will discuss offline for action item here. Thanks Rafi!

RissyRan:
> Looks really good, just a question of waht we're testing.

Thanks! Updated the test, and all checks passed.

rwitten (approved):
I'm giving approval but please move the test itself to XLML

RissyRan:
> I'm giving approval but please move the test itself to XLML

Sounds good. I assume this will include both Llama2 and Mistral tests end-to-end tests. 

cc @jonb377

PR #352, opened 2024-01-18
trim instead of drop long samples

aireenmei:
https://github.com/google/maxtext/issues/274

### DIFF
--- a/MaxText/input_pipeline.py
+++ b/MaxText/input_pipeline.py
@@ -79,6 +79,19 @@ def normalize_features(ds):
       _normalize_features,
       num_parallel_calls=AUTOTUNE)
 
+def length_trim(ds, max_len):
+  """"Trim to Max length"""
+  def _trim_fn(features):
+    if tf.shape(features['inputs'])[0] > max_len:
+      features['inputs'] = features['inputs'][:max_len]
+    if tf.shape(features['targets'])[0] > max_len:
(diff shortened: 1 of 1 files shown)


rwitten:
Have you done convergence check here?

aireenmei:
Done. Convergence looks ok, after 20400 steps, "Mean of last 10 losses is 2.481567859649658", log: "https://pantheon.corp.google.com/kubernetes/service/us-east5/v5e-256-bodaborg/default/aireen-v5e256-0119-0155/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod"

(main commit convergence for comparison: https://pantheon.corp.google.com/kubernetes/service/us-east5/v5e-256-bodaborg/default/aireen-v5e256-0119-1202/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod )

A separate note: I noticed in the convergence recipe, per_device_batch_size was increased from 2 to 8, which increased the step time, but took less steps to reach ~2.6. Should we decrease the 20400 steps for convergence test? Now it took ~4.5 h compared with ~2.2 h earlier this month

rwitten (approved):
Thanks for the CR!

Matt is thinking through the convergence test recipe, let's not worry about it right now.

aireenmei:
I've stashed the commits and it should be ready to merge. Could you merge?

PR #357, opened 2024-01-18
Add kubectl and gke-gcloud-auth-plugin to maxtext dependency dockerfile

raymondzouu:
This change is needed to be able to run MaxText/XPK-based XLML tests on the new Airflow-based infra. The infra executes the XPK command using KubernetesPodOperator on the docker image https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/a0287d3730ccd9d2b24ac2339dbb1cd42386f9e7/xlml/utils/xpk.py#L50-L57. KubernetesPodOperator needs `kubectl` and `google-cloud-sdk-gke-gcloud-auth-plugin`.

### DIFF
--- a/maxtext_dependencies.Dockerfile
+++ b/maxtext_dependencies.Dockerfile
@@ -14,6 +14,10 @@ RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyri
 # Install the Google Cloud SDK
 RUN apt-get update && apt-get install -y google-cloud-sdk
 
+# Install kubectl and gke-gcloud-auth-plugin components needed for KubernetesPodOperator
+RUN apt-get install -y kubectl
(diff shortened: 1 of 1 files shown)


tonyjohnchen (approved):

rwitten:
This docker image is building a runner -- a.k.a. an env where MaxText runs. Why does that env need to know about Kubectl etc?

rwitten (changes requested):
I don't understand this change! Doesn't seem like a MaxText runner (i.e. a GPU/TPU ndoe) should need to know about kubectl.

raymondzouu:
The MaxText runner docker image needed kubectl since XLML currently uses the runner docker image to execute the XPK command with KubernetePodsOperator. Instead of it installing it here for all use cases, I made a change in the XLML codebase such that the XLML test will install it in the image only when needed. I will close out this PR as it is no longer needed.

PR #359, opened 2024-01-19
Splash Attention

rwitten:
Splash works (including MQA/GQA). Not yet using sliding window but I'll turn that on when we have non-kernel reference, will be a one line change.

To merge, I need Jax 0.4.24 to land which will let me avoid copy and pasting the kernels.

Perf was a tiny speedup on my box (8 ms / step faster) for 1B model with seqlen=2048.


enijkamp:
@rwitten This looks great! Any update on "when we have non-kernel reference"? Thank you!

gobbleturk (approved):

PR #360, opened 2024-01-20
Fix preprocessing pipeline to use tokenizer's bos and eos

A9isha:
Enables Llama2 training


rwitten (commented):
Nice!

But where are we ensuring the right BOS and EOS are used for Llama during training? Like isn't the Llama BOS == 2? I don't even see that in a config. (Or is it directly coming out of the tokenizer?)

Also random but can you delete this (dead) line of code:
https://github.com/google/maxtext/blob/ec9cdce1928fb36f93e498c8f74f8f130bad3dc2/MaxText/configs/base.yml#L163

Finally, can you do a convergence run using the convergence recipe to ensure nothing is broken.

rwitten (approved):
<3

PR #362, opened 2024-01-21
Separated the notion of model_name (param overrides) from decoder_block

rwitten:
…ck (what decoder block to use)


A9isha (approved):
Thanks Rafi!

PR #363, opened 2024-01-22
[MLPerf GPT3] Data pipeline, Eval step

ZhiyuLi-goog:
As titled, Data pipeline, Eval step and compiliance log for MLPerf

* eval step
* new c4_mlperf dataset for mlperf
* mllog i.e. compiliance log for MLPerf
* switch to iterator class `MultiHostDataLoadIterator` to support for loop in evaluation step
  - `reset` iterator at each evaluation step
  - iterator class supports for loop naturally

### DIFF
--- a/MaxText/input_pipeline/_tfds_data_processing.py
+++ b/MaxText/input_pipeline/_tfds_data_processing.py
@@ -24,6 +24,7 @@ import tensorflow as tf
 import tensorflow_datasets as tfds
 import jax
 
+import multihost_dataloading
 import tokenizer
 import sequence_packing
 
@@ -161,7 +162,10 @@ def preprocessing_pipeline(
   if prefetch_size:
     dataset = dataset.prefetch(prefetch_size)
 
-  return iter(dataset.as_numpy_iterator())
+  multihost_gen = multihost_dataloading.MultiHostDataLoadIterator(dataset, global_mesh)
+
+  # Return multi-host jax.Array prep iterator
+  return multihost_gen
 
 
 def get_datasets(
@@ -187,7 +191,7 @@ def get_datasets(
   # eval_data = get_raw_dataset(eval_ds_builder, config.eval_split)
   eval_ds = eval_ds_builder.as_dataset(split=config.eval_split,
                                           read_config = read_config,
-                                          shuffle_files=config.enable_data_shuffling)
+                                          shuffle_files=False)
   eval_ds = eval_ds.shard(num_shards = jax.process_count(), index = jax.process_index())
   eval_ds = normalize_features(eval_ds)
 
@@ -244,10 +248,11 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       eval_ds,
       eval_batch_size,
       global_mesh,
-      shuffle=config.enable_data_shuffling,
+      shuffle=False,
       pack_examples=False,
       max_length=config.max_target_length,
       shift=False,
+      drop_remainder=False,
       data_shuffle_seed = data_shuffle_seed,)
 
   predict_iter = preprocessing_pipeline(
@@ -262,4 +267,3 @@ def preprocess_dataset(config: ml_collections.ConfigDict,
       data_shuffle_seed = data_shuffle_seed,)
 
   return train_iter, eval_iter, predict_iter, sp_tokenizer
-
--- a/MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
+++ b/MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
@@ -1,0 +1,285 @@
+"""
+ Copyright 2023 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+"""Input pipeline for gpt3 c4 mlperf dataset."""
+
+from typing import Optional
+
+import functools
+
+import numpy as np
+
+import ml_collections
+import tensorflow as tf
+import tensorflow_datasets as tfds
+import jax
+import jax.numpy as jnp
+from jax.experimental import multihost_utils
+
+import tokenizer
+import multihost_dataloading
+import sequence_packing
+
+AUTOTUNE = tf.data.experimental.AUTOTUNE
+
+# data processing functions:
+#   _shift_left_and_pad, rekey, reduce_concat_tokens and split_tokens_to_targets_length
+# Adapted from:
+#   https://github.com/google-research/text-to-text-transfer-transformer/blob/ba171b6f94eafcee60d0714fd6d60749b572d1f2/t5/data/preprocessors.py
+# -----------------------------------------------------------------------------
+def _shift_left_and_pad(tensor, pad_val):
+  """Shift the input to the left with pad_val"""
+  # Expand dims here so that the below code can work with 1-d tensors.
+  v = tf.expand_dims(tensor, 0)
+  # Make sure we keep tensor as ragged to allow for uneven concat.
+  if isinstance(v, tf.Tensor):
+    v = tf.RaggedTensor.from_tensor(v)
+
+  # Append padding to the last item of every sequence.
+  pad_shape = tf.concat([v.bounding_shape()[:-2], [1, 1]], axis=0)
+  pad_tensor = tf.broadcast_to(pad_val, pad_shape)
+  last_in_sequence = tf.concat([v[..., -1:, 1:], pad_tensor], axis=-1)
+  # Concat back the newly modified final sequence item.
+  v = tf.concat([v[..., :-1, :], last_in_sequence], axis=-2)
+  # Un-expand outer dimension.
+  v = v[0]
+  return v
+
+def rekey(ds, key_map=None):
+  """normalization with key mapping"""
+  def _rekey(x, key_map=None):
+    """Replace the feature keys according to the mapping in `key_map`.
+    For example, if the dataset returns examples of the format:
+    {'foo': 'something', 'bar': 'something else', 'zoo': 'others'}
+    and key_map = {'boo': 'foo', 'spar': 'bar', 'zoo': None} then this function will return
+    examples with the format
+    {'boo': 'something', 'spar': 'something else'}
+    If a mapping is to None, then the key will be dropped.
+    Args:
+      x: an example to process.
+      key_map: dictionary mapping new keys to original keys
+    Returns:
+      A preprocessed example with the format listed above.
+    """
+    if key_map:
+      return {
+          new_key: x[old_key]
+          for new_key, old_key in key_map.items() if old_key
+      }
+    return x
+
+  return ds.map(
+      functools.partial(_rekey, key_map=key_map),
+      num_parallel_calls=AUTOTUNE)
+
+def reduce_concat_tokens(dataset,
+                         feature_key='targets',
+                         batch_size=128,
+                         ):
+  """Token-preprocessor to concatenate multiple unrelated documents.
+  If we want to generate examples of exactly the right length,
+  (to avoid wasting space on padding), then we use this function, folowed by
+  split_tokens.
+  Args:
+    dataset: a tf.data.Dataset with dictionaries containing the key feature_key.
+    feature_key: an string
+    batch_size: an integer - how many documents to concatenate into one
+  Returns:
+    a dataset
+  """
+  dataset = dataset.map(
+      lambda x: {feature_key: x[feature_key]}, num_parallel_calls=AUTOTUNE)
+  dataset = dataset.padded_batch(batch_size, padded_shapes={feature_key: [-1]})
+  def _my_fn(x):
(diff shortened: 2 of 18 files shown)


rwitten (changes requested):
I believe there are CR's missing from this so temporarily removing myself as assignee.

ZhiyuLi-goog:
> I believe there are CR's missing from this so temporarily removing myself as assignee.

Hi, @rwitten You are right. 
This CR is a follow-up one which depends on [the first change](https://github.com/google/maxtext/pull/293). We can go back to this one later.

ZhiyuLi-goog:
The first PR https://github.com/google/maxtext/pull/293 has been merged in and this CL is ready for review.

MaxText/multihost_dataloading.py
```diff
@@ -105,3 +105,21 @@ def get_next_batch_sharded(
   input_gdas = jtu.tree_map_with_path(partial(_form_global_array, global_mesh = global_mesh), local_data)
 
   return input_gdas
+
+
+class MultiHostDataLoadIterator:
```
rwitten, line 110: @aireenmei be aware.

@ZhiyuLi-goog looks great!

MaxText/pyconfig.py
```diff
@@ -149,13 +149,21 @@ def __init__(self, argv: list[str], **kwargs):
...
     keys_from_env_and_command_line = self._update_from_env_and_command_line(raw_keys, raw_data_from_yaml, argv, **kwargs)
     max_logging.log(
         f"Updating keys from env and command line: {keys_from_env_and_command_line}")
+
     keys_from_model = _HyperParameters.update_model_vars(raw_keys)
     max_logging.log(f"Updating keys from model: {keys_from_model}")
     validate_no_keys_overwritten_twice(keys_from_env_and_command_line, keys_from_model)
 
+    # We initialize the jax distributed system here because it must be done before device backend is initialized.
+    if raw_keys["enable_checkpointing"] and raw_keys["async_checkpointing"] and raw_keys["compile_topology_num_slices"]==-1:
```
rwitten, line 162: @gobbleturk I'm really confused that this is spreading but we need to figure out how to make it stop. Should the three of us get together and discuss this?

MaxText/train.py
```diff
@@ -196,33 +197,55 @@ def train_step(model, config, state, data, dropout_rng):
...
+      data[k] = v[:config.global_batch_size_to_train_on,:]
 
-  def loss_fn(params):
+  def loss_fn(params, is_train=True):
     logits, intermediate_outputs = model.apply({'params': params},
                          data['inputs'],
                          data['inputs_position'],
                          decoder_segment_ids=data['inputs_segmentation'],
-                         enable_dropout=config.enable_dropout,
+                         enable_dropout=config.enable_dropout if is_train else False,
```
rwitten, line 209: Why? Can't you set Dropout= 0?

MaxText/train.py
```diff
@@ -301,16 +324,30 @@ def train_loop(config, state=None):
   Returns:
   """
   ( init_rng, writer, checkpoint_manager, state_mesh_annotations, model,
-  mesh, learning_rate_schedule, data_iterator, state) = setup_train_loop(config)
+  mesh, learning_rate_schedule, data_iterator, eval_data_iterator, state) = setup_train_loop(config)
 
-  functional_train, in_shard, out_shard, static_argnums, donate_argnums = maxtext_utils.get_functional_train_with_signature(
+  # pylint: disable=line-too-long
+  functional_train, in_shard, out_shard, static_argnums_train, donate_argnums_train = maxtext_utils.get_functional_train_with_signature(
```
rwitten, line 330: Not good to have a "functional_train_with_signature" that takes "is_train". Happy to discuss solutions or for you to propose one!

MaxText/train.py
```diff
@@ -345,6 +391,10 @@ def train_loop(config, state=None):
 
   example_batch = None
   last_step_completion = datetime.datetime.now()
+  if config.enable_mllog:
```
rwitten, line 394: probably wrap this in a maxtext specific version to preserve readability in train.py.

MaxText/train.py
```diff
@@ -371,6 +421,23 @@ def train_loop(config, state=None):
...
+          _, eval_metrics = p_eval_step(
+            state, eval_batch, nextrng
+          )
+        cumulative_eval_metrics['total_loss'] += float(eval_metrics['scalar']['evaluation/total_loss'])
+        cumulative_eval_metrics['total_weights'] += float(eval_metrics['scalar']['evaluation/total_weights'])
+      eval_loss = cumulative_eval_metrics['total_loss'] / (cumulative_eval_metrics['total_weights'] + EPS)
+      max_logging.log(f"average loss after {step=}: {eval_loss=}, total_weights={cumulative_eval_metrics['total_weights']}")
+      if config.enable_mllog:
+        mllog_utils.early_stop_check(config, step, eval_loss)
+      if eval_loss <= config.target_eval_loss:
```
rwitten, line 438: Shouldn't we log here?

MaxText/train.py
ZhiyuLi-goog: Hi, @rwitten 

`Dropout = 0`(`dropout_rate=0`) is equivalent to `enable_dropout = False`, while `model` allows us to pass `enable_dropout` only and `dropout_rate` is fixed in config. Considering we enable dropout with a valid `dropout_rate` in training while disabling it evaluation.

Looking into the code in depth, `deterministic` is simply designed to disable dropout in evaluation:
```
# flax https://flax.readthedocs.io/en/latest/_modules/flax/linen/stochastic.html#Dropout
x = nn.Dropout(0.5, deterministic=not train)(x)

# ours
deterministic=not enable_dropout,
...
nn.Dropout(
        rate=cfg.dropout_rate, broadcast_dims=(-2,)
    )(x, deterministic=deterministic)
```

MaxText/train.py
ZhiyuLi-goog: Thank you @rwitten for the comments.

Instead of fusing train and evaluation in the same `functional_train_with_signature`, I am considering to add a `functional_eval_with_signature` in addition to `functional_train_with_signature`, what do you think?

```
def get_functional_eval_step(train_step, model, config):
  # train and eval step share the same api for simplicity:
  #   train:  new_state, metrics = train_step(state, data, is_train=True)
  #   eval:  _, metrics = train_step(state, data, is_train=False)
  return functools.partial(train_step, model, config, is_train=False)

def get_functional_train_with_signature(train_step, mesh, state_mesh_annotations, model, config):
  ...

# add a new independent method
def get_functional_eval_with_signature(train_step, mesh, state_mesh_annotations, model, config):
  ...

```

MaxText/multihost_dataloading.py
aireenmei: Hi @ZhiyuLi-goog , my grain changes were in this morning. I had some refactoring on tfds input pipeline as well. Looks like it may cause conflicts with your change. Let me know if you have questions / want to discuss any change

MaxText/train.py
rwitten: yes this makes sense.

MaxText/train.py
rwitten: that is a good idea and an improvement. But you've pushed the problem to:
```
return functools.partial(train_step, model, config, is_train=False)
```

MaxText/train.py
rwitten: I mean maxtext print!

MaxText/train.py
ZhiyuLi-goog: Hi, Rafi, thank you for your comment. Currently, I combined train and eval [under the same function](https://github.com/google/maxtext/blob/73665db73eb108a23e845b5d8220672d0afd6d71/MaxText/train.py#L181-L252). `is_train` flag is used to differentiate the branches. This is mainly to reuse loss_fn to ensure a compact code style.  
Looks like you are favoring more towards 2 separate functions instead. How about we do this way:
```
# we can almost reuse 99% of loss_fn for both train_step and eval_step
def loss_fn(params, is_train=True):
    logits, intermediate_outputs = model.apply({'params': params},
                         ...
                         enable_dropout=config.enable_dropout if is_train else False,  # only difference
    ...

# independent functions elsewhere.
def train_step
  ...
functional_train = get_functional_train_with_signature
p_train_step = jax.jit(functional_train)

def eval_step
  ...
functional_eval = get_functional_eval_with_signature
p_eval_step = jax.jit(functional_eval)
```

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
```diff
@@ -0,0 +1,286 @@
+
```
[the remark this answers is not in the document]
ZhiyuLi-goog: Gotcha, @gobbleturk is convergence step appropriate metrics for comparison? I expect the packing way in MLPerf is helpful in the number of convergence since there is more meaningful tokens with less paddings.  

If we need to collect all data points for different models, it might be a long period. Not sure if we can go back to it in March after getting best gpt3 training perf.

MaxText/train.py
ZhiyuLi-goog: Good suggestion! Will do.

MaxText/train.py
ZhiyuLi-goog: Good suggestion, will add maxtext print to notify early stop.

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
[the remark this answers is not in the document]
rwitten: No we don't need to collect all data points for different models, just Matt's convergence sweep.

MaxText/train.py
rwitten: SGTM

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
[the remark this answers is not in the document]
ZhiyuLi-goog: * [Visualization of different data format](https://colab.sandbox.google.com/drive/16Idie6590780l__4txwyLpkdSIIBZEcF#scrollTo=CJyHgzfncFUi)
* [TODO]convergence test data points

MaxText/train.py
ZhiyuLi-goog: Done.

MaxText/train.py
ZhiyuLi-goog: Disabled all mllogs in main branch and will cover it in a specific version/branch.

MaxText/input_pipeline/_tfds_data_processing.py
```diff
@@ -187,7 +191,7 @@ def get_datasets(
   # eval_data = get_raw_dataset(eval_ds_builder, config.eval_split)
   eval_ds = eval_ds_builder.as_dataset(split=config.eval_split,
                                           read_config = read_config,
-                                          shuffle_files=config.enable_data_shuffling)
+                                          shuffle_files=False)
```
rwitten, line 194: Wait we want to offer this still right but set to False? am I missing something?

MaxText/standalone_dataloader.py
```diff
@@ -38,22 +36,20 @@ def data_load_loop(config, state=None):
...
-  _, _, _, _, _, _, _, data_iterator, state = setup_train_loop(config)
+  _, _, _, _, _, _, _, data_iterator, _, state = setup_train_loop(config)
 
-  devices_array = max_utils.create_device_mesh(config)
-  mesh = Mesh(devices_array, config.mesh_axes)
   example_batch = None
 
   start = datetime.datetime.now()
   start_step = get_first_step(state)
-  example_batch = load_next_batch(data_iterator, example_batch, config, mesh)
```
rwitten, line 49: please talk to @aireenmei -- not sure why you all are taking and removing mesh :-)

rwitten (commented):
Hold on a tiny nit, looks almost awesome.

MaxText/input_pipeline/_tfds_data_processing.py
ZhiyuLi-goog: Thank you Rafi, if I am understanding correctly, evaluation dataset is usually not recommended to be shuffled since it won't help in back prop and it might introduce subtle differences from different orderings between 2 eval runs. Not an issue for c4_mlperf eval dataset which loads processed data directly.

I found it when checking if we can get the same first eval batch between 2 different eval runs.

MaxText/standalone_dataloader.py
ZhiyuLi-goog: Thank you @aireenmei & @rwitten.

`load_next_batch` function is folded in [`MultiHostDataLoadIterator`](https://github.com/google/maxtext/blob/de24bde15c264704c99a12c433a14d8aa75cebed/MaxText/multihost_dataloading.py#L96) which accepts both `tf.data.dataset` and `grain.DataLoader`. `mesh` is not needed since it is part of the class. 

The main goal is to ensure repeatable evaluation loop. We may trigger eval run multiple times and need a way to [reset evaluation iterator](https://github.com/google/maxtext/blob/de24bde15c264704c99a12c433a14d8aa75cebed/MaxText/tests/tfds_data_processing_test.py#L128)

Feel free to let me know what you think.

aireenmei: Thanks for the explanation. LGTM

MaxText/input_pipeline/_tfds_data_processing.py
rwitten: my mistake, thank you

MaxText/pyconfig.py
rwitten: @ZhiyuLi-goog I think this changed a little upstream but I have no objection.

rwitten (approved):
LGTM

PR #364, opened 2024-01-22
CPU Inference Configs

priyanka-ganesha:
Inference decode configurations intended for CPU for model sizes 1B, 4B, 8B, and 16B parameters.


MaxText/configs/cpu/16b.sh
```diff
@@ -0,0 +1,28 @@
+echo "Running 16b.sh"
```
rwitten, line 1: Please put these in end to end and run some as CPU only end to end tests.

I also don't htink ou need the weird argument copy paaa (or even the dupliction at all nowadays with the export M_whatever.

rwitten (changes requested):

PR #367, opened 2024-01-23
refactor input_pipeline.py and fix test

aireenmei:
1. the original input_pipeline.py was refactored into input_pipeline/tfds_data_processing.py and input_pipeline/input_pipeline.py
2. add flag enable_file_shuffling, because I don't see a way to set seed for file_shuffling when calling "train_ds_builder.as_dataset", so when it is set to true through the flag enable_data_shuffling, we don't get determinism. With the two flags, we can achieve determinism by setting enable_file_shuffling=False, enable_data_shuffling=True, data_shuffle_seed=<the same seed>
3. fix input_pipeline_test.py and rename to tfds_data_processing_test.py

convergence tested: https://pantheon.corp.google.com/kubernetes/service/us-east5/v5e-256-bodaborg/default/aireen-v5e256-0119-1503/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod

### DIFF
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -29,7 +29,7 @@ import numpy as np
 
 import pyconfig
 import max_utils
-from input_pipeline import create_data_iterator_with_tokenizer
+from input_pipeline.input_pipeline_interface import create_data_iterator_with_tokenizer
 from layers import models
 
 import common_types
--- a/MaxText/input_pipeline/_tfds_data_processing.py
+++ b/MaxText/input_pipeline/_tfds_data_processing.py
@@ -23,8 +23,6 @@ import ml_collections
 import tensorflow as tf
 import tensorflow_datasets as tfds
 import jax
-import jax.numpy as jnp
-from jax.sharding import PartitionSpec as P
 
 import tokenizer
 import multihost_dataloading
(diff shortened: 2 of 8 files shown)


MaxText/configs/base.yml
```diff
@@ -177,6 +177,7 @@ profiler_steps: 5
 # data_shuffle_seed and init_weights_seed (i.e. reproducible losses)
 enable_dropout: True
 enable_data_shuffling: True
+enable_file_shuffling: False
```
rwitten, line 180: Sorry I'm a little confused here? Did you add a new field for the same meaning as the above field?

MaxText/decode.py
```diff
@@ -29,7 +29,7 @@
 
 import pyconfig
 import max_utils
-from input_pipeline import create_data_iterator_with_tokenizer
+from input_pipeline.input_pipeline import create_data_iterator_with_tokenizer
```
rwitten, line 32: nit: input_pipeline/input_pipeline.py is a slightly sad name but I htink it is fine.

rwitten (changes requested):
I'm confused by `enable_data_shuffling` and `enable_file_shuffling` (looks like you renamed but didn't remove the old name?)

MaxText/configs/base.yml
aireenmei: I added a new flag. These are shuffling at different levels. "enable_file_shuffling" correspondings to "shuffle_files" here: https://github.com/google/maxtext/blob/main/MaxText/input_pipeline.py#L183. "enable_data_shuffling" is used here: https://github.com/google/maxtext/blob/main/MaxText/input_pipeline.py#L133-L134. I separate these two flags because when enable_file_shuffling=True, "shards are shuffled for each epoch, so reading is not deterministic anymore" (https://www.tensorflow.org/datasets/determinism) and there's no way to set a fixed seed for file shuffle. While with "enable_data_shuffling=true", determinism can be achieved by setting data_shuffle_seed

MaxText/decode.py
aireenmei: I'm open to a name suggestion!

MaxText/configs/base.yml
aireenmei: I added some comments to config, hopefully it's more clear now

rwitten: Thank you for the deep explanation. Are we sure we want to expose both of them?

Separately, if you are sure about this, the names/comments need to be crisper.

MaxText/decode.py
rwitten: Are we designing the system so that input_pipeline/input_pipeline.py is the only interface? If so input_pipeline_interface.py is a helpful name (and hiding the other names as prefaced with an underscore is best).

aireenmei: done

MaxText/configs/base.yml
aireenmei: I just took another look and found that shuffle_file can also be deterministic when read_config is set. Initially I saw my unit test for determinism was failing so I changed the flag. But I then realized the end-to-end determinism is passing and my unit test failure was due to not setting read_config. I revert the config change and fix unit tests now.

rwitten (approved):
Thanks for bearing with me!

PR #369, opened 2024-01-23
[Draft] mixtral inference

WendyShang:


MaxText/layers/llama2.py
```diff
@@ -26,11 +26,13 @@
 
 
 import jax.numpy as jnp
+import jax
```
RissyRan, line 29: Hi @rwitten and @WendyShang, for layers, do you suggest we have a separate one, i.e. mistral.py or rename this to something like `llama2_or_mistral.py`? I remember we will need to add sliding window attention soon. So wondering if it makes sense to separate the layer here. Thanks!

PR #370, opened 2024-01-24
Splitting kv cache - prefill

patemotter:
First part of changes to split the kv cache into prefill and AR portions. This code just takes care of the prefill portion and defaults to not being used for the time being. When the AR portion of the kv cache split is complete we can remove the newly added `combined_kv_cache` boolean config entry and the related `if/else` statements.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -80,6 +80,7 @@ attention: 'flash' # Supported attention: dot_product, flash, gpu_flash_xla, gpu
 # Combine matmuls for QKV and MLP
 fused_qkv: False
 fused_mlp: False
+combined_kv_cache: True
 
 record_internal_nn_metrics: 0
 
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -444,16 +444,36 @@ def get_kv_cache_annotations(model, config, rng, mesh):
       config.global_batch_size_to_load,
       config.max_target_length
     )
+    model_vars = model.init({'params': rng, 'dropout': rng, 'aqt': rng},
+                            jnp.ones(input_shape),
+                            jnp.ones(input_shape),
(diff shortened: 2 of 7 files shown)


MaxText/configs/base.yml
```diff
@@ -80,6 +80,7 @@ attention: 'flash' # Supported attention: dot_product, flash, gpu_flash_xla, gpu
 # Combine matmuls for QKV and MLP
 fused_qkv: False
 fused_mlp: False
+combined_kv_cache: True
```
rwitten, line 83: BTW MaxText usually doesn't bother with this kind of thing. We just implement what we think is best!

(Not sure either way on this CR but just stylistically we are focused on "only the right path" wherever possible

rwitten (changes requested):
Ok some thoughts:
* I don't think we want to support both a joined cache and an unjoined cache. I think we just do an unjoined cache.
* I think in a single CR we should make this whole change -- I don't want to leave MaxText in a weird incomplete state.

MaxText/max_utils.py
```diff
@@ -440,20 +440,42 @@ def get_kv_cache_annotations(model, config, rng, mesh):
   """ Get a shaped abstraction of the state (including optimizer)"""
 
   def init_kv_cache(model, config):
+    if config.combined_kv_cache:
```
rwitten, line 443: Oh interesting -- I think your point here is that model init was kind of wrong before? Because in either case we should be init-ing with the config.max_prefill_predict_length as input?

As it happens I think this isn't a big issue because the cache size is being driven by config constants not by init (so using the wrong shape is actually ok!)

Anyways, we should fix this and just run through with the `config.max_prefill_predict_length`?

rwitten (changes requested):

PR #371, opened 2024-01-24
Fix creation of "/" directory in gs bucket.

abhinavclemson:


gobbleturk (approved):
I haven't seen this echo pattern before for bash functions - glad to learn of it! I have code that does the reverse (add a trailing slash if not present) here https://source.corp.google.com/piper///depot/google3/cloud/tpu/tools/multipod/multipod_tests/unsanitize_and_move_dir.sh;l=31?q=%20unsanitize_and_move_dir

rwitten (approved):

PR #373, opened 2024-01-24
Add validation for overwritten keys

RissyRan:
# Description

Add validation for overwritten keys between keys from env/command line and model:
* Move core logic to `_update_from_env_and_command_line` helper function
* Add `validate_no_keys_overwritten_twice` helper function 
* Other small changes: delete unused vars

# Tests
If there is an overwritten keys, i.e. `attention=dot_product` in the unit test, we will get error:

```
  File "/home/ranran/maxtext/MaxText/decode.py", line 274, in <module>
    app.run(main)
  File "/home/ranran/.local/lib/python3.10/site-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/home/ranran/.local/lib/python3.10/site-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "/home/ranran/maxtext/MaxText/decode.py", line 266, in main
    pyconfig.initialize(argv)
  File "/home/ranran/maxtext/MaxText/pyconfig.py", line 300, in initialize
    _config = _HyperParameters(argv, **kwargs)
  File "/home/ranran/maxtext/MaxText/pyconfig.py", line 159, in __init__
    validate_overwritten_keys(keys_from_env_and_command_line, keys_from_model)
  File "/home/ranran/maxtext/MaxText/pyconfig.py", line 75, in validate_overwritten_keys
    raise ValueError(
ValueError: Keys ['attention'] are overwritten from both the model and the environment/command line. This isn't allowed.
```

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -31,9 +31,6 @@ import jax
 from typing import Any, Union
 
 _MAX_PREFIX = "M_"
-_DEFAULT = "default"
-_LLAMA2_7B = "llama2-7b"
-_MISTRAL_7B = "mistral-7b"
 
 def yaml_key_to_env_key(s: str) -> str:
   return _MAX_PREFIX + s.upper()
@@ -62,7 +59,6 @@ def validate_keys(keys):
   assert keys["load_parameters_path"]=="" or keys["load_full_state_path"]=="",\
     "At most one of `load_parameters_path` or `load_full_state_path` should be set"
 
-
 def validate_model_name(s: str) -> bool:
   # currently supported models
   valid_model_names= ('default', 'llama2-7b', 'mistral-7b', 'gamma-7b','gamma-2b')
@@ -70,6 +66,13 @@ def validate_model_name(s: str) -> bool:
     raise ValueError(
       "Invalid model name was passed. Valid options ", valid_model_names
     )
+
+def validate_no_keys_overwritten_twice(keys1: list[str], keys2: list[str]):
+  overwritten_keys = [k for k in keys1 if k in keys2]
+  if overwritten_keys:
+    raise ValueError(
+        f"Keys {overwritten_keys} are overwritten from both the model" 
+        " and the environment/command line. This isn't allowed.")
 
 _config = None
 config = None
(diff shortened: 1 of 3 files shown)


MaxText/pyconfig.py
```diff
@@ -71,6 +67,12 @@ def validate_model_name(s: str) -> bool:
       "Invalid model name was passed. Valid options ", valid_model_names
     )
 
+def validate_overwritten_keys(keys1: list[str], keys2: list[str]):
```
rwitten, line 70: validate_no_keys_overwritten_twice would be a crisper name.

end_to_end/llama_load_and_test.sh
```diff
@@ -13,7 +13,7 @@ export M_ASYNC_CHECKPOINTING=false
 #python3 MaxText/llama_or_mistral_ckpt.py --base-model-path /tmp/meta-ckpt --model-size llama2-7b --maxtext-model-path gs://maxtext-llama/test/${idx}/decode-ckpt-maxtext/
 
 # Load after directly from parameter checkpoint
-python3 MaxText/decode.py MaxText/configs/base.yml load_parameters_path=${base_ckpt_path} run_name=runner_direct_${idx} per_device_batch_size=1 model_name='llama2-7b' assets_path=gs://maxtext-llama/llama2-7b ici_tensor_parallelism=4 max_prefill_predict_length=4  max_target_length=16 prompt="I love to" autoregressive_decode_assert="read. I love to write. I love to share." attention=dot_product
```
rwitten, line 16: please don't change this -- actually the fix is to remove attention=dot_product from the llama model. (Decode right now wants attention=dot_product but for some reason it is being set at the llama2 level)

rwitten (changes requested):
Looks amazing. 

A little nit with naming.

And a little issue you've uncovered where we were setting a variable twice but we picked wrong on where we *should* set it from.

MaxText/pyconfig.py
```diff
@@ -71,6 +67,12 @@ def validate_model_name(s: str) -> bool:
...
+  overwritten_keys = [k for k in keys1 if k in keys2]
+  if overwritten_keys:
+    raise ValueError(
+            f"Keys {overwritten_keys} are overwritten from multiple sources. This isn't allowed.")
```
rwitten, line 74: Probably we could be clear that it is overwritten both from the model and from the env or command line?

MaxText/pyconfig.py
RissyRan: Sure, I will update that. Thank you!

end_to_end/llama_load_and_test.sh
RissyRan: Sure, I will do the change accordingly.

MaxText/pyconfig.py
RissyRan: Sure, I will update that to be more specific. At the very beginning, I was thinking this would a helper func to validate any 2 lists of keys.

RissyRan:
> Looks amazing.
> 
> A little nit with naming.
> 
> And a little issue you've uncovered where we were setting a variable twice but we picked wrong on where we _should_ set it from.

Thanks Rafi! Comments addressed.

rwitten (approved):
Thank you @RissyRan

PR #377, opened 2024-01-26
Deprecate PytreeCheckpointer

rwitten:


rwitten:
This is more of a project than I realized. Closing for now, opening another issue.

PR #380, opened 2024-01-27
Update the network setting to match internal dev settings.

tonyjohnchen:
Update the network setting to match internal dev settings(http://shortn/_hVGhEM2N8p).
Merge `rto_setup.sh` and `gke_rto_setup.sh` together. 

Tested locally on v5p-8


MaxText/configs/experimental/1024b.sh
```diff
@@ -16,13 +16,13 @@ done
 
 if [[ $PLATFORM == "gce" ]]; then
```
raymondzouu, line 17: Could you change this if-else block to just `bash preflight.sh PLATFORM=$PLATFORM`? I think I forgot to update it before

raymondzouu (commented):
Thanks Tony!

MaxText/configs/experimental/1024b.sh
tonyjohnchen: I am not sure how to change the DNS lookup for GCE and GKE next:

```
# For DNS lookup when running on large number of VMs on gce
    echo '142.250.123.95 www.googleapis.com' | sudo tee -a /etc/hosts
    echo '142.251.4.128 storage.googleapis.com' | sudo tee -a /etc/hosts
```

```
 # For DNS lookup when running on large number of VMs on gke
    echo '142.250.123.95 www.googleapis.com' | tee -a /etc/hosts
    echo '142.251.4.128 storage.googleapis.com' | tee -a /etc/hosts
```

Do you mind add  `preflight.sh` in the next PR and I can help review.

raymondzouu: You can remove those, we don't need them anymore.

tonyjohnchen: Done.

preflight.sh
```diff
@@ -23,9 +23,9 @@ fi
 if [[ $PLATFORM == "gce" ]]; then
     # Set up network for running on gce
     echo "Setting up network for GCE"
-    bash rto_setup.sh
```
rwitten, line 26: Can we just check if have sudo and use sudo if we have it?

rwitten (approved):
Approving but with an idea...

preflight.sh
tonyjohnchen: Done.

preflight.sh
```diff
@@ -20,12 +20,13 @@ if [ -z $PLATFORM ]; then
     exit 1
 fi
 
-if [[ $PLATFORM == "gce" ]]; then
-    # Set up network for running on gce
-    echo "Setting up network for GCE"
-    bash rto_setup.sh
+# Check if sudo is available
```
raymondzouu, line 23: This is a great change. Now it looks like we no longer need to pass in a `PLATFORM=...` flag anymore into this script and would no longer need this intermediate script `preflight.sh`. Instead we could just directly call `rto_setup.sh` in our scripts. We should change all of the instances where we call `preflight.sh` to just call `rto_setup.sh` instead. This is non-blocking for this PR but should be done in a follow-up PR.

raymondzouu (approved):
Left a comment about what we need to do in a follow-up PR but should be non-blocking for this PR.

preflight.sh
raymondzouu: Could be worth it to keep a `preflight.sh` actually in case we add other initial workload set ups.

tonyjohnchen: Yes, I don't want to change the structure too much. We can save this for now since there might be more differences between GCE and GKE.

PR #381, opened 2024-01-27
add llama finetuning test

A9isha:


rwitten (changes requested):
This needs to be XLML now that the XLML path is fully modernized. @RissyRan can help.

(end_to_end/llama_finetuning_test.sh may merge but not the piece about the GH runner)

rwitten (approved):

PR #382, opened 2024-01-29
enable grain in input pipeline

aireenmei:
Convergence tested: loss 12.4 -> 3 (https://pantheon.corp.google.com/kubernetes/service/us-east5/v5e-256-bodaborg/default/aireen-v5e256-0129-0747/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod)
checkpointing with data iterator will be in a separate PR

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -40,6 +40,7 @@ import optimizers
 import pyconfig
 
 from input_pipeline.input_pipeline_interface import create_data_iterator_with_tokenizer
+from multihost_dataloading import get_next_batch_sharded
 from layers import models
 
 import jax.numpy as jnp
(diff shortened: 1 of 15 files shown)


MaxText/train.py
```diff
@@ -84,13 +85,18 @@ def get_first_step(state):
...
   """Loads the next batch. Can keep reusing the same batch for performance reasons """
 
   if config.reuse_example_batch and example_batch is not None:
     return example_batch
   else:
-    return train_iter()
+    if config.dataset_type == 'c4':
+      return train_iter()
+    elif config.dataset_type == 'c4-array_record':
+      return get_next_batch_sharded(train_iter, mesh)
```
rwitten, line 97: I feel like this interface should be unified -- this code shouldn't know about dataset_type

end_to_end/test_convergence_1b_params_grain.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
rwitten, line 1: Let's avoid duplication -- the other one can do this via overrides, right?

requirements.txt
```diff
@@ -1,8 +1,11 @@
 orbax-checkpoint
 absl-py
+array-record
+argparse
```
rwitten, line 4: Are you sure you use this?

rwitten (changes requested):
Very close, nice change!

(Have you converted the full Grain dataset? How did that go?)

MaxText/train.py
aireenmei: revised as suggested.

end_to_end/test_convergence_1b_params_grain.sh
aireenmei: done

requirements.txt
aireenmei: Removed, thanks!

rwitten (approved):
Amazing!

PR #388, opened 2024-01-31
Change requirements.txt deps

khatwanimohit:
Reduced deps installation time from 4m20 sec -> ~3min 
Docker image size from 7.2GB -> 4.53GB


rwitten (approved):
Thank you Mohit!

PR #391, opened 2024-01-31
Splitting kv cache 

patemotter:
This splits the single kv cache into two caches: prefill and autoregressive. 

This code does not currently pass unit tests due to an issue with the flash attention style normalization across both the prefill and ar caches. The code will run to completion, but the attention test comparing `mha_generic_flash_output` and  `mha_generic_output` fails because the diff is too large. 

Below is an idea of what the diff looks like. The original diff values not equal to zero are much smaller than those in the new method. It does seem that there are relatively few values that change from zero to non-zero. It's mostly that the non-zero entries are much larger. 

Original Code diff between `mha_generic_output` and `mha_generic_flash_output`: 
```
 [[0 0 0 ... 0 0 0]
  [0 0 0 ... 0 0 0]
  [0 0 0 ... 0 0 0]
  ...
  [0      0.001   0.007    ...  0  0.003  0.002]
  [0.004  0.0009  0        ...  0  0.004  0.004]
  [0 0 0 ... 0 0 0]]
```

New Code diff between `mha_generic_output` and `mha_generic_flash_output`: 
```
 [[0 0 0 ... 0 0 0]
  [0 0 0 ... 0 0 0]
  [0 0 0 ... 0 0 0]
  ...
  [0.066  0.449  0.297    ... 0.969  0.609  0.049]
  [0.640  0.005  0.550    ... 0.477  0.789  0.412]
  [0 0 0 ... 0 0 0]]
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,36 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug MaxText Decode",
+      "type": "python",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "justMyCode": false,
+      "python": "python3",
+      "program": "${workspaceFolder}/MaxText/decode.py",
+      "args": ["MaxText/configs/base.yml", 
+               "run_name=runner_$(date +%Y-%m-%d-%H-%M)", 
+               "base_output_directory=gs://test-maxtext-output",
+               "dataset_path=gs://test-maxtext-dataset",
+               "steps=2",
+               "attention=dot_product",
+               "enable_checkpointing=false"]
+    },
+    {
+      "name": "Debug MaxText Train",
+      "type": "python",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "justMyCode": false,
+      "python": "python3",
+      "program": "${workspaceFolder}/MaxText/train.py",
+      "args": ["MaxText/configs/base.yml", 
+               "run_name=runner_$(date +%Y-%m-%d-%H-%M)", 
+               "base_output_directory=gs://test-maxtext-output",
+               "dataset_path=gs://test-maxtext-dataset",
+               "steps=2",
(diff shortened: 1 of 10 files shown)


patemotter:
Two numerical errors were fixed in the latest commit and should be passing the unit tests now.

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
...
+      "name": "Debug MaxText Decode",
+      "type": "python",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "justMyCode": false,
+      "python": "python3",
+      "program": "${workspaceFolder}/MaxText/decode.py",
+      "args": ["MaxText/configs/base.yml", 
+               "run_name=runner_$(date +%Y-%m-%d-%H-%M)", 
+               "base_output_directory=gs://patemotter-maxtext-out",
```
rwitten, line 14: nit: can you make this one of the shared directories? and while you're being a hero can you add a a train version?

rwitten (approved):
Tiny nit, looks amazing otherwise.

MaxText/layers/models.py
```diff
@@ -79,6 +79,7 @@ def __call__(self,
       num_kv_heads=cfg.num_kv_heads,
       head_dim=cfg.head_dim,
       max_target_length=cfg.max_target_length,
+      max_prefill_predict_length=cfg.max_prefill_predict_length,
```
rwitten, line 82: I think gpt3 needs this as well.

rwitten (changes requested):
Tiny nits

MaxText/layers/models.py
patemotter: Good catch. Added.

.vscode/launch.json
patemotter: Done

patemotter:
Fixing up commits to squash everything.

PR #392, opened 2024-02-01
Update v5p configs

priyanka-ganesha:
Updating v5p configs as 32B on 2xv5p-128s and 256B on 2xv5p-1024s fail with out of memory errors - might be due to recent compiler changes. 512B on v5p-1024s fails with mesh creation related errors.
Perf numbers will need to be updated accordingly as we get additional capacity.

b/323298666


rwitten (approved):

Obliviour (approved):

raymondzouu (approved):
Thanks Priyanka!

PR #393, opened 2024-02-01
Do not quantize logits

gobbleturk:
Quantizing logits hurts training accuracy


rwitten (approved):
Just to be clear the comment should be "we do not quantize the logit matmul"

PR #401, opened 2024-02-07
Nina loading ckpt according to config.dtype

NinaCai:
Instead of loading ckpt as float32, data type is decided by config.dype.
Add Unit tests for gpu inference.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -90,6 +90,8 @@ def load_next_batch(train_iter, example_batch, config, mesh):
 
   if config.reuse_example_batch and example_batch is not None:
     return example_batch
+  elif config.dataset_type=='synthetic':
+    return next(train_iter)
   else:
     return get_next_batch_sharded(train_iter, mesh)
 


MaxText/layers/embeddings.py
```diff
@@ -56,7 +56,7 @@ def setup(self):
         'embedding',
         with_logical_partitioning(self.embedding_init, ('vocab', 'embed')),
         (self.num_embeddings, self.features),
-        jnp.float32,
```
rwitten, line 59: I gave feedback by chat but please use a config other than config.dtype (for example config.weight_dtype would be good)

rwitten (changes requested):
Mostly looks good!

.github/workflows/UnitTests.yml
```diff
@@ -102,7 +102,11 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_generate_param_only_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M) -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4'
-  
+    - name: Test gamma
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
```
rwitten, line 107: this really should be XLML! Because it is slow. If we want to do it here, can we restrict it be just the 2.5B model?

.github/workflows/UnitTests.yml
```diff
@@ -134,3 +138,15 @@ jobs:
...
+        'python3 MaxText/decode.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset steps=2 ici_tensor_parallelism=4 attention=dot_product enable_checkpointing=false max_target_length=128 per_device_batch_size=1'
+    - name: Test decode.py with per_device_batch_size < 1
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
+        'python3 MaxText/decode.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M) base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset steps=2 ici_tensor_parallelism=4 attention=dot_product enable_checkpointing=false max_target_length=128 per_device_batch_size=.25'
+    - name: Test gamma
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
```
rwitten, line 151: again, we don't want these in the GH runner (see comment above)

MaxText/configs/base.yml
```diff
@@ -63,6 +63,7 @@ decoder_block: "default" # which style of DecoderBlock to use.
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes 
 # then you should explicitly set base_embed_dim, base_num_query_heads, base_num_kv_heads,
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
+weight_type: "float32"
```
rwitten, line 66: nit: weight_dtype

rwitten (commented):
Broadly looks good, tiny nits.

.github/workflows/UnitTests.yml
rwitten: (For the TPUs, we run this in XLML already. I'm not sure on GPUs for XLML, could be a good chat with @RissyRan )

RissyRan: Yes, currently MaxText tests are running TPU. If you are working on GPU, We have it available in XL ML via 2 ways: 1) GKE via [XPK](https://github.com/google/xpk) AND 2) GCE directly. The GCE route currently only support single slice but we are working on multislice feature.

Here attaches some examples for your reference to onboard Gamma into XL ML:
* GKE example on TPU: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/examples/xpk_example_dag.py#L34-L54. I think @michelle-yooh has landed GKE with GPU.
* GCE example on GPU: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/pytorch_xla/pytorchxla_torchbench.py#L37-L46

Happy to chat more!

NinaCai: Ok, deleted test_gamma from UnitTests.yml. 
@michelle-yooh  Could you confirm if you have already added GKE with GPU for maxtext end_to_end?
If so, I will do it using GCE directly. @RissyRan Do I need to do code change after multislice feature is enabled?

rwitten (approved):

PR #408, opened 2024-02-10
added support for cudnn flash attention

kocchop:
- Implemented cudnn flash attention with Transformer Engine
- Currently it supports head_dim till 128 and does not support GQA yet. It's an unstable API and would soon change it to a stable one with more feature support
- Added a flag `--multiprocess_gpu` to support multiprocess SLURM tasks


MaxText/configs/base.yml
```diff
@@ -81,7 +81,7 @@ logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedd
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: dot_product, flash, gpu_flash_xla, gpu_flash_triton
+attention: 'dot_product' # Supported attention: dot_product, flash, gpu_flash_xla, gpu_flash_triton, cudnn_flash_te
```
rwitten, line 84: Let's remove gpu_flash_xla and gpu_flash_triton for now dispatch to the cudnn_flash_te codepath based on hardware
https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml#L99

MaxText/configs/base.yml
```diff
@@ -225,3 +225,7 @@ decode_sampling_strategy: "greedy" # decode_sampling_strategy should be one of g
 decode_sampling_nucleus_p: -1 # set if you're doing nucleus / top-p
 decode_sampling_top_k: 0 # set if you're doing top-k
 decode_sampling_temperature: 1.
+
+# enabling slurm for multiprocess in gpus
+# It must be turned on while using SLURM
+multiprocess_gpu: False
```
rwitten, line 231: Sincerely sorry for the trouble. But let's use:
https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml#L99

requirements.txt
```diff
@@ -20,4 +20,5 @@ tensorflow-text>=2.13.0
 tensorflow>=2.13.0
 tensorflow-datasets
 tensorboardx
-tensorboard-plugin-profile
\ No newline at end of file
+tensorboard-plugin-profile
+transformer-engine
```
rwitten, line 24: hmmm not loving this for all users. Is the solutions a requirements-gpu separately that setup installs for you.

(Weirdly there is no requirements-tpu becuase we get that via jax[tpu])

MaxText/layers/attentions.py
```diff
@@ -277,6 +283,70 @@ def wrap_gpu_flash_attention(query, key, value):
...
+    self,
+    query: Array,
+    key: Array,
+    value: Array,
+  ) -> Array:
+    """
+    CUDNN Flash Attention with Transformer Engine.
+    It is an unstable API. In future release, the API can get changed
+    A stable flash attention API will be included soon. Currently,
+    1. It does not support GQA, num_query_heads == num_kv_heads
```
rwitten, line 296: nit: remove comments 1 and 2, add asserts.

MaxText/configs/base.yml
```diff
@@ -65,11 +65,11 @@ decoder_block: "default" # which style of DecoderBlock to use.
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
```
rwitten, line 65: Overall, looks very strong, thank you!.

There are nits below. But more substantively, to merge this, we need a numerical test showing GPU Flash Attention works within numerical tolerence, similar (and possibly exactly the same) as one of the Attention_test.py tests but marked to run on GPU. Let's discuss off-thread if you want to do this or you'd ask a Googler to do it for you.

rwitten (changes requested):

MaxText/configs/base.yml
kocchop: Sure, I can add appropriate args to the following flag

MaxText/layers/attentions.py
kocchop: I can add the asserts.. however, the line 318 essentially does the assert within TE.. should I also add them here?

requirements.txt
kocchop: So, the cudnn flash attention uses the transformer-engine, hence it needs to be added as a requirement. Do you recommend having a seprate requirements-gpu?

MaxText/layers/attentions.py
rwitten: Thanks, no need for double assert. Thanks for explaining.

requirements.txt
rwitten: Yes I think we need a requirements-gpu.txt that setup.sh can install additionally.

rwitten (changes requested):
Thanks for the responses

kocchop:
Hi @rwitten I have made the changes. Please let me know how it looks.
Also I have been added to the contributor's list. My current commit email is verified but my inital commit email is not. Is there any way to remove it? Just wondering what could be the easiest way to fix this. Thanks!

MaxText/configs/base.yml
kocchop: I had a small doubt- as with gpu, someone would want use mha as well, is it really necessary to change it to flash attention based on hw?

PR #415, opened 2024-02-13
Refactoring Tokenizer creation

khatwanimohit:


MaxText/tokenizer.py
```diff
@@ -42,10 +42,10 @@ def _load_sentencepiece_tokenizer(model_path: str,
       model=sp_model, add_bos=add_bos, add_eos=add_eos, reverse=reverse)
   return sp_tokenizer
 
-def load_tokenizer(vocab_path: str, add_bos=False, add_eos=True):
-  """Loads the tokenizer at `vocab_path` or trains a one from `dataset`."""
+def load_tokenizer(model_path: str, add_bos=False, add_eos=True):
```
rwitten, line 45: why model path instead of vocab path?

rwitten: how about tokenizer path?

MaxText/configs/base.yml
```diff
@@ -139,8 +139,7 @@ ici_autoregressive_parallelism: 1
 # Replace with your path given as argument in download_dataset.sh, e.g. "gs://my-maxtext-dataset/"
 dataset_path: ""
 vocab_size: 32_768 # powers of 2 for sharding
-assets_path: "assets"
-vocab_relative_path: "tokenizer"  # Assumes we're allowed
```
rwitten, line 143: have you git greped to and made sure you aren't breaking anyone? I feel like you're breaking a lot of tests.

rwitten (changes requested):

rwitten (approved):
You have my approval BUT please alert in the MaxText room that you expect it might break things and you will check the XLML tests.

PR #417, opened 2024-02-15
[Nina] load ckpt according to config weight_dtype

NinaCai:
Instead of loading ckpt as float32, data type is decided by config.dype.
Add Unit tests for gpu inference.
change s/decode_iter to ms/decode_iter
change rng

### DIFF
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -232,11 +232,12 @@ def decode_loop(config, state=None):
 
   starttime = datetime.datetime.now()
   steps = range(config.max_prefill_predict_length + 1, config.max_target_length)
+  rngs = [random.PRNGKey(i) for i in range(config.max_target_length)]
   for step in steps:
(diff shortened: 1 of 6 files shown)


MaxText/decode.py
```diff
@@ -250,12 +251,12 @@ def decode_loop(config, state=None):
     f"generated text mismatch {new_text=} {config.autoregressive_decode_assert=}"
 
   num_steps = len(steps)
-  elapsed_time = (endtime-starttime).total_seconds()
+  elapsed_time = (endtime-starttime).total_seconds() * 1000
```
rwitten, line 254: I'm actually going to delete this code soonish so you don't need to sweat it ... but this var should be named `elapsed_time_ms` now (and I should have named it `elapsed_time_seconds`)

rwitten (approved):

PR #420, opened 2024-02-15
fix gpt3 e2e test

ZhiyuLi-goog:
GPT3-52k is failing due to splash attention not working on that baby head size of 8. Disable splash attention by default.

- [x] local test passed and verified.

### DIFF
--- a/MaxText/configs/models/gpt3-52k.yml
+++ b/MaxText/configs/models/gpt3-52k.yml
@@ -37,3 +37,4 @@ adam_b1: 0.9
 adam_b2: 0.95
 adam_eps: 1.e-8
 adam_weight_decay: 0.1
+attention: "dot_product"  # head_dim 8 is too small for splash/flash attention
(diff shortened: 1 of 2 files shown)


MaxText/configs/models/gpt3-52k.yml
```diff
@@ -37,3 +37,4 @@ adam_b1: 0.9
 adam_b2: 0.95
 adam_eps: 1.e-8
 adam_weight_decay: 0.1
+attention: "dot_product"  # head_dim 8 is too small for splash/flash attention
```
RissyRan, line 40: I recall Rafi suggested us to put doc_product to the end_to_end test script, instead of model yml.

ZhiyuLi-goog: We need it in checkpoint conversion as well when we compile and create initial weights. Therefore, changing the default model setting would be the safest.

rwitten: +1 to @RissyRan and +2 to @ZhiyuLi-goog

rwitten (approved):

PR #427, opened 2024-02-16
Adding GPU multi-node support to Maxtext

parambole:
Added
- Maxtext `jax.distributed.initialize` GPU support
- Support for deploying it on GKE with GPUDirectTCPx

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -180,16 +180,55 @@ def maybe_initialize_jax_distributed_system(raw_keys):
     max_logging.log("Attempting to initialize the jax distributed system...")
     jax.distributed.initialize()
     max_logging.log("Jax distributed system initialized!")
+  elif is_gpu_backend(raw_keys):
+    max_logging.log("Attempting to initialize the jax distributed system for GPU backend...")
+    initialize_jax_for_gpu()
+    max_logging.log("Jax distributed system initialized on GPU!")
   elif is_cpu_backend(raw_keys):
     max_logging.log("Attempting to initialize the jax distributed system for CPU backend...")
     initialize_jax_for_cpu()
     max_logging.log("Jax distributed system initialized on CPUs!")
 
-
+def initialize_jax_for_gpu():
+  """Jax distributed initialize for GPUs."""
+  if os.environ.get("JAX_COORDINATOR_IP") is not None:
+    coordinator_ip = str(os.getenv("JAX_COORDINATOR_IP"))
+    coordinator_port = str(os.getenv("JAX_COORDINATOR_PORT"))
+    jax.distributed.initialize(
+        coordinator_address=f"{coordinator_ip}:{coordinator_port}",
+        num_processes=int(os.getenv("JAX_NUM_PROCESSES")),
+        process_id=int(os.getenv("PROCESS_ID")),
+        local_device_ids=int(os.getenv("LOCAL_DEVICE_ID")),
+    )
+    max_logging.log(f"JAX global devices: {jax.devices()}")
 
 def initialize_jax_for_cpu():
   """Jax distributed initialize for CPUs. Includes retries until the coordinator is ready.
   """
+  coordinator_ip_address = get_coordinator_ip_address()
+  coordinator_address = coordinator_ip_address + ":1234" # JAX coordinator port used in XPK
+  # Env variables to be set in XPK or otherwise
+  job_index = int(os.environ.get("JOB_INDEX"))
+  job_completion_index = int(os.environ.get("JOB_COMPLETION_INDEX"))
+  processes_in_job = int(os.environ.get("PROCESSES_IN_JOB"))
+  pid = job_index * processes_in_job + job_completion_index
+  max_logging.log(f" Jax process id is {pid} ")
+  # Explicit initialize is needed only for CPUs
+  jax.distributed.initialize(coordinator_address=coordinator_address,
+                            process_id=pid,
+                            num_processes=int(os.environ.get("JAX_PROCESS_COUNT")))
+
+def is_cpu_backend(raw_keys):
+  """Determine whether Maxtext is intended to run on a CPU backend."""
+  return raw_keys["hardware"] == 'cpu'
+
+def is_gpu_backend(raw_keys):
+  """Determine whether Maxtext is intended to run on a GPU backend."""
+  return raw_keys["hardware"] == 'gpu'
+
+def get_coordinator_ip_address():
+  """Get coordinator IP Address with retries"""
+  coordinator_address = ""
   if os.environ.get("JAX_COORDINATOR_ADDRESS") is not None:
     coordinator_address = os.environ.get("JAX_COORDINATOR_ADDRESS")
     coordinator_found = False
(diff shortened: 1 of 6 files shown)


MaxText/configs/base.yml
```diff
@@ -101,6 +101,10 @@ jax_cache_dir: "~/jax_cache"
 # Hardware
 hardware: 'tpu' # Supported hardware types are 'tpu', 'gpu' and 'cpu'
 
+# Jax Initilization configuration
+process_id: -1
```
rwitten, line 105: please just directly pull this from env vars

rwitten (changes requested):
tiny nit mostly looks good

maxtext_gpu_dependencies.Dockerfile
michelle-yooh, line 1: I'm working on re-writing this image based on Jax image from Nvidia. Can you leave this file unchanged?

MaxText/configs/base.yml
parambole: @rwittenI I opted not to set them as envvars because they are not global across all processes. Specifically, the values for `process_id` and `local_device_id` are specific to each process, whereas values like `COORDINATOR_PORT` and `COORDINATOR_ADDRESS` are more appropriately configured as envvars since they apply globally to all processes.

Please let me know if you have a different perspective.

start_training.sh
```diff
@@ -0,0 +1,98 @@
+#! /bin/bash
```
rwitten, line 1: please put in something like gke/gpu.

gke/gpu/maxtext_chart/templates/maxtext.yaml
```diff
@@ -0,0 +1,144 @@
+{{- $requiredVar := .Values.cluster.nNodes | required ".Values.cluster.nNodes is required" -}}
```
rwitten, line 1: all fine for now but needs to go into XPK right? can you eventually own cleanup here?

MaxText/configs/base.yml
```diff
@@ -103,6 +103,10 @@ jax_cache_dir: "~/jax_cache"
...
```
rwitten, line 107: Can't we get rid of this?

MaxText/max_utils.py
```diff
@@ -180,46 +180,74 @@ def maybe_initialize_jax_distributed_system(raw_keys):
...
+def initialize_jax_for_gpu(raw_keys):
+  """Jax distributed initialize for GPUs. Includes retries until the coordinator is ready.
+  """
+  coordinator_ip_address = get_coordinator_ip_address()
+  coordinator_port = str(os.getenv("JAX_COORDINATOR_PORT"))
+
+  jax.distributed.initialize(
+      coordinator_address=f"{coordinator_ip_address}:{coordinator_port}",
+      num_processes=int(os.getenv("JAX_NUM_PROCESSES")),
+      process_id=raw_keys["process_id"],
```
rwitten, line 201: isn't this supposed to come from an env var?

rwitten (changes requested):

MaxText/max_utils.py
parambole: Done
1. Resolving the IP and setting it as environment variable
2. Consuming the Coordinator IP envvar

gke/gpu/maxtext_chart/templates/maxtext.yaml
parambole: Ack.

start_training.sh
parambole: Done

MaxText/configs/base.yml
parambole: @rwitten I opted not to set them as `envvars`  because they are not global across all processes. Specifically, the values for process_id and local_device_id are specific to each process, whereas values like `COORDINATOR_PORT` and `COORDINATOR_IP` are more appropriately configured as `envvars` since they apply globally to all processes.

Resolving for now please reopen if you have a different perspective.

rwitten: Sorry I missed this the first time. Theres a typo in `Initilization` and let's add a comment saying that this is only environments such as GPUs where we operate one process per device so an env variable is unhelpful. (P.S. you could still use an env var and that would still be preferable! You'd just export them before launching the process. But now I understand your goal enough that I think I'm ok with this...)

gke/gpu/start_training.sh
```diff
@@ -0,0 +1,136 @@
...
+# Resolving coordinator IP
+set +e
+resolve_coordinator_ip
+set -e
+
+
+PIDS=()
+for ((LOCAL_DEVICE_ID=0; LOCAL_DEVICE_ID <= $((GPUS_PER_NODE - 1)); LOCAL_DEVICE_ID++)); do
+   PROCESS_ID=$(($GPUS_PER_NODE*$NODE_RANK + $LOCAL_DEVICE_ID))
+   python MaxText/train.py MaxText/configs/base.yml hardware=gpu run_name=${RUN_NAME}_$(date +%Y-%m-%d-%H-%M) local_device_id=$LOCAL_DEVICE_ID process_id=$PROCESS_ID &
```
rwitten, line 130: just set process ID as an env var here! and local device id!

rwitten: it should be as simple as LOCAL_DEVICE_ID=$LOCAL_DEVICE_ID PROCESS_ID=$PROCESS_ID python MaxText/train.py MaxText/configs/base.yml hardware=gpu run_name=${RUN_NAME}_$(date +%Y-%m-%d-%H-%M)  &

rwitten (approved):
Looks fine, sorry for the confusion in my comments.

I gave advice for a better, more Jax-y solution. Please try to make that work.

I have given my approval either way.

MaxText/configs/base.yml
parambole: Ack. I have passed it as `envvar` to make it more `Jax-y`

gke/gpu/start_training.sh
parambole: Ack. I have passed it as `envvar` to make it more `Jax-y`

kocchop:
I am not sure if it's exactly relevant. In the latest merged [PR 430](https://github.com/google/maxtext/pull/430), I added support for multiprocess gpu. In order to invoke multiprocess in single node or multinode, for me, just doing ```jax.distributed.initialize()``` was sufficient in SLURM. In order to enable the multitask support, I also used a flag ```gpu_multiprocess```. Please see this [line in max_utils.py](https://github.com/google/maxtext/blob/5420bc5753fec4b3a811664cdb58f3c9e98d35fb/MaxText/max_utils.py#L179) and this [line in base.yaml](https://github.com/google/maxtext/blob/5420bc5753fec4b3a811664cdb58f3c9e98d35fb/MaxText/configs/base.yml#L104). 

One thing to note that we only want to use this flag when we are launching more that one process. If anyone wants to use a single task/process for the run, this flag should be turned off.

MaxText/max_utils.py
```diff
@@ -180,46 +180,73 @@ def maybe_initialize_jax_distributed_system(raw_keys):
...
     max_logging.log("Jax distributed system initialized on CPUs!")
 
+def initialize_jax_for_gpu():
+  """Jax distributed initialize for GPUs. Includes retries until the coordinator is ready.
+  """
+  coordinator_ip = str(os.getenv("JAX_COORDINATOR_IP"))
+  coordinator_port = str(os.getenv("JAX_COORDINATOR_PORT"))
+  jax.distributed.initialize(
+      coordinator_address=f"{coordinator_ip}:{coordinator_port}",
+      num_processes=int(os.getenv("JAX_NUM_PROCESSES")),
```
rwitten, line 199: I think @kocchop's point is that most of these should now work! Without you needing to do anything? If you read the Jax code.

rwitten (approved):

MaxText/max_utils.py
parambole: @kocchop & @rwitten As per https://jax.readthedocs.io/en/latest/multi_process.html#initializing-the-cluster the initialization will work for SLURM and MPI ( what is being used by @kocchop ) without any args. But that does not currently hold true for other orchestration methods.

parambole:
> I am not sure if it's exactly relevant. In the latest merged [PR 430](https://github.com/google/maxtext/pull/430), I added support for multiprocess gpu. In order to invoke multiprocess in single node or multinode, for me, just doing `jax.distributed.initialize()` was sufficient in SLURM. In order to enable the multitask support, I also used a flag `gpu_multiprocess`. Please see this [line in max_utils.py](https://github.com/google/maxtext/blob/5420bc5753fec4b3a811664cdb58f3c9e98d35fb/MaxText/max_utils.py#L179) and this [line in base.yaml](https://github.com/google/maxtext/blob/5420bc5753fec4b3a811664cdb58f3c9e98d35fb/MaxText/configs/base.yml#L104).
> 
> One thing to note that we only want to use this flag when we are launching more that one process. If anyone wants to use a single task/process for the run, this flag should be turned off.

As per https://jax.readthedocs.io/en/latest/multi_process.html#initializing-the-cluster the initialization will work for SLURM and MPI without any args. But that does not currently hold true for other orchestration methods.

rwitten (approved):

PR #430, opened 2024-02-17
added transformer-engine cudnn flash attn

kocchop:
Created this new PR with the changes @rwitten


MaxText/configs/base.yml
```diff
@@ -84,7 +84,7 @@ logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedd
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
-attention: 'flash' # Supported attention: dot_product, flash, gpu_flash_xla, gpu_flash_triton
+attention: 'dot_product' # Supported attention: dot_product, flash, cudnn_flash_te
```
rwitten, line 87: Can we leave this as dot_product by default to avoid churn? For your tests you can override via CLI or env var.

(I know this is lame. I have a change that switches to "recommended" as the default that dispatches intelligently.)

requirements-gpu.txt
```diff
@@ -0,0 +1,25 @@
+jax>=0.4.23
```
rwitten, line 1: I don't think we want to trust folks to update the different lists. Could you do this and in setup.sh if GPU also install the GPU requirements.txt?
https://stackoverflow.com/questions/36744143/how-can-i-define-multiple-requirement-files

rwitten (changes requested):
2 tiny nits

MaxText/configs/base.yml
kocchop: Hi @rwitten, it is set as `dot_product` in the PR currently. Did you mean to keep the default as `flash`?

rwitten (approved):

rwitten (approved):
Thank you and congrats!

PR #431, opened 2024-02-18
install google cloud sdk (gsutil)

ssusie:
adding google cloud sdk; in particular gsutil to work with `--debug-dump-gcs` in xpk.

### DIFF
--- a/maxtext_dependencies.Dockerfile
+++ b/maxtext_dependencies.Dockerfile
@@ -7,8 +7,12 @@ RUN apt-get update && apt-get install -y curl gnupg git numactl
 # Install dependencies for adjusting network rto
 RUN apt-get update && apt-get install -y iproute2 ethtool lsof
 
+# Add the Google Cloud SDK package repository
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
 # Install the Google Cloud SDK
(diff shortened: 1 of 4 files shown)


maxtext_dependencies.Dockerfile
```diff
@@ -7,6 +7,13 @@ RUN apt-get update && apt-get install -y curl gnupg git numactl
...
 
+# Add the Google Cloud SDK package repository
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
+# Install the Google Cloud SDK
+RUN apt-get update && apt-get install -y google-cloud-sdk
+
 # Install the Google Cloud SDK
 RUN curl -sSL https://sdk.cloud.google.com | bash
```
Obliviour, line 18: Hmm fishy that we install it twice. Why did we change to curl (breaking gsutil initially). I wonder if we need curl for anothwr reason?

rwitten: Yes super suspicious Can we get rid of the curl?

(And I imagine you've already tested this??)

rwitten: Finally can you add a test that just runs locally and makes sure gsutil is installed in our docker image? Just a unit test that does something like `gsutil --version`. (The problem is that some important flows depends on gsutil but most don't so right now we discover it being broken super slowly.)

rwitten (changes requested):

maxtext_dependencies.Dockerfile
ssusie: Addressed comments. Can you PTAL?

rwitten (approved):

PR #435, opened 2024-02-20
Move all dependencies from `preflight.sh` to `setup.sh`

tonyjohnchen:
Move all dependencies from `preflight.sh` to `setup.sh`

Tested in both GCE and GKE.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -27,6 +27,17 @@
 # Enable "exit immediately if any command fails" option
 set -e
 
+# Check if sudo is available
+if command -v sudo >/dev/null 2>&1; then
+    # sudo is available, use it
+    # install numactl for numa binding.
+    sudo apt update && sudo apt install numactl   
+else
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -27,6 +27,17 @@
...
 
+# Check if sudo is available
+if command -v sudo >/dev/null 2>&1; then
+    # sudo is available, use it
+    # install numactl for numa binding.
+    sudo apt update && sudo apt install numactl   
+else
+    # sudo is not available, run the script without sudo
+    # install numactl for numa binding.
+    sudo apt update && sudo apt install numactl
```
rwitten, line 38: bug here I think?

rwitten (changes requested):

setup.sh
tonyjohnchen: Fixed.

PR #436, opened 2024-02-21
Pass quantization options into aot

gobbleturk:
This is a quick fix, but it may be worth more thought about how to prevent breaks like this in the future either by refactoring code or more tests b/326138012


rwitten (approved):
Can we make sure that quantization is used everywhere it should be? Maybe quantization should be required?

gobbleturk:
> Can we make sure that quantization is used everywhere it should be? Maybe quantization should be required?

Modified quantization to required, now passing everywhere

PR #437, opened 2024-02-21
Adding quantization support using AQT for inference.

singh-mitali:


MaxText/decode.py
```diff
@@ -82,79 +86,99 @@ def encode_strings(strs, max_len, tokenizer, mesh):
...
     new_token,
     token_position,
     enable_dropout=False,
     model_mode=common_types.MODEL_MODE_AUTOREGRESSIVE,
     rngs={'params': rngkey},
-    mutable=["cache"])
+    mutable=mutable)
   new_flat_cache = new_vars["cache"]
-  return token_position+1, new_flat_cache, flat_logits, new_token
+  new_vars.pop('cache')
```
rwitten, line 130: can we convert in init? Super weird to have it spread out.

rwitten (changes requested):
Let's talk live -- everything looks great except for the conversion being inline in a step AFAICT?

MaxText/decode.py
```diff
@@ -173,71 +199,113 @@ def prefill_or_load(config, model, state, rng, sp_tokenizer, mesh, state_mesh_sh
...
+  prompts = default_prompts(config)
+
+  # Compute shardings
+  replicated_sharding = jax.sharding.NamedSharding(mesh, P(None))
+  kv_cache_annotations = max_utils.get_kv_cache_annotations(model, config, rng, mesh)
+  kv_cache_mesh_shardings = jax.tree_map(
+      lambda p: jax.sharding.NamedSharding(mesh, p), kv_cache_annotations)
+
+  # Run prefill step
+  cache, logits, pos, new_vars = compute_prefill(
```
rwitten, line 215: Which variable fails to go through the code path at this point?

rwitten (changes requested):

rwitten (approved):

MaxText/decode.py
singh-mitali: Discussed offline - we use the prefill apply method to call a "dummy" prefill to initiate aqt state. Cleaned up the function a bit.

MaxText/decode.py
singh-mitali: We needed to call the "dummy" predict step after the "dummy" prefill step because it was initializing extra state for attention einsum. Discussed offline and decided to remove quantization in attention layer to keep the aqt initialization more light weight.

PR #439, opened 2024-02-21
Revised: refactor maxtext checkpointing and integrate with grain

aireenmei:


rwitten (approved):
AWESOME!

Can you say explicilty what change you made relative to the previous CR?And what actions we should change for the future? Maybe add Gemma2.5B as a pre-submit?

aireenmei:
For the previous PR, I got the suggestion from orbax team to use checkpointer.metadata() to construct restore_args for loading parameters only from a checkpoint when setting "load_parameters_from_path" (https://yaqs.corp.google.com/eng/q/1295652407543332864). But the error we got indicated the older checkpoints are missing something in _METADATA, resulting in failures. I then revised to not using checkpointer.metadata(). 

For future improvement. I think adding a test that use "load_parameters_from_path" from an checkpoint we don't want to break will help.

PR #441, opened 2024-02-21
Smoke test for int8

gobbleturk:
Add G3 smoke test for int8


rwitten (approved):
This CL is fine but it is supposed to go through in a way that doesn't end up being externally visible I thought? Not worried about it either way.

gobbleturk:
Pushed cl/610524560 internally instead of this PR

PR #448, opened 2024-02-23
update v5p-3072 topology

ZhiyuLi-goog:
v5p-3072 = 1536 chips = 8x12x16

### DIFF
--- a/MaxText/accelerator_to_spec_map.py
+++ b/MaxText/accelerator_to_spec_map.py
@@ -167,7 +167,7 @@ UserFacingNameToSystemCharacteristics = {
         'tpu', 'v5:4x4x92', 'megacore', (2, 2, 1), 1472
     ),
     'v5p-3072': SystemCharacteristics(
-        'tpu', 'v5:4x12x16', 'megacore', (2, 2, 1), 1536
+        'tpu', 'v5:8x12x16', 'megacore', (2, 2, 1), 1536
     ),
     'v5p-3200': SystemCharacteristics(
         'tpu', 'v5:4x20x20', 'megacore', (2, 2, 1), 1600


ZhiyuLi-goog:
@Obliviour could you take a look?

MaxText/accelerator_to_spec_map.py
```diff
@@ -167,7 +167,7 @@ class SystemCharacteristics:
         'tpu', 'v5:4x4x92', 'megacore', (2, 2, 1), 1472
     ),
     'v5p-3072': SystemCharacteristics(
-        'tpu', 'v5:4x12x16', 'megacore', (2, 2, 1), 1536
```
rwitten, line 170: rude don't do this. make it a new name.

rwitten: v5p-3072-special or something. If you really want we could optionally provide the topology seprately.

rwitten: This is for AOT needs right? Recall that XPK has its own list.

rwitten (changes requested):

MaxText/accelerator_to_spec_map.py
ZhiyuLi-goog: Hi, @rwitten. There was a typo in this topology originated in a g3 source. The original number of devices (1536) doesn't equal to the product of topology i.e. 4x12x16. And this PR is to fix it. 

I have 2 changes in xpk and internal g3 accordingly:
* g3 source cl/609827788
* [xpk PR](https://github.com/google/xpk/pull/76)

cc @Obliviour

Obliviour: Agreed to Zhiyu's comment, this fixes the topology that is currently invalid. 

@ZhiyuLi-goog its probably worth adding in the title / description of the intent of the PR to help avoid confusion. Like the total of "Fix invalid topology for v5p-3072"

The description is something like:

Problem:
The previous topology is invalid because...

Solution:
I changed it to be x. Also changed in xpk and ctpu code

Obliviour (approved):

MaxText/accelerator_to_spec_map.py
rwitten: Thank you both.

rwitten (approved):

ZhiyuLi-goog:
Close this duplicate PR since [the commit](https://github.com/google/maxtext/commit/0f153c8baa803ea906036622e0478e4706b199db) was auto-submitted by copybara already.

This PR was not merged since I rebased it onto new head after cl submission which created a new and unmatched commit hash to the previous one in copybara. Have learned a lesson.

PR #455, opened 2024-02-27
Build maxtext dependency images

gobbleturk:
Build and upload maxtext dependency images periodically


rwitten (approved):
amazing thanks! Can we make these images public by any chance? No worries if not.

gobbleturk:
> amazing thanks! Can we make these images public by any chance? No worries if not.

I think this would be very helpful - created b/327264821 to track

PR #460, opened 2024-02-28
Make small changes for running it via XPK.

yangyuwei:


gke/gpu/start_training.sh
```diff
@@ -126,10 +123,10 @@ set -e
 PIDS=()
 for ((LOCAL_DEVICE_ID=0; LOCAL_DEVICE_ID <= $((GPUS_PER_NODE - 1)); LOCAL_DEVICE_ID++)); do
    PROCESS_ID=$(($GPUS_PER_NODE*$NODE_RANK + $LOCAL_DEVICE_ID))
-   LOCAL_DEVICE_ID=$LOCAL_DEVICE_ID PROCESS_ID=$PROCESS_ID python MaxText/train.py MaxText/configs/base.yml hardware=gpu run_name=${RUN_NAME}_$(date +%Y-%m-%d-%H-%M) &
+   LOCAL_DEVICE_ID=$LOCAL_DEVICE_ID PROCESS_ID=$PROCESS_ID python MaxText/train.py MaxText/configs/base.yml hardware=gpu run_name=${RUN_NAME}_$(date +%Y-%m-%d-%H-%M) ${ARGS} &
```
rwitten, line 126: @yangyuwei we should talk about this design a bit more.

Like all of this logic should be in XPK, not MaxText right? It is true for all Jax programs, i.e. just running jax.devices(). And XPK runs arbitrary programs, not just MaxText...

rwitten (approved):
Approving but a little scared.

PR #466, opened 2024-02-29
Ansiha ckpt perf cpu

A9isha:


MaxText/llama_or_mistral_ckpt.py
```diff
@@ -41,6 +41,7 @@
 from train import save_checkpoint
 import torch
 import sys
+import os
```
rwitten, line 44: remove

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -394,4 +417,8 @@ def convert(base_model_path, maxtext_model_path, model_size):
 
   if args.model_size not in MODEL_PARAMS_DICT:
     raise NotImplementedError
+
+  if args.model_size == "llama2-70b":
```
rwitten, line 421: probably sharding is always welcome?

rwitten (approved):
One little nit.

But more substantively I think the two paths (one sharded, one unsharded) is bad because it increases maintenance burden. I think one path is right?

MaxText/llama_or_mistral_ckpt.py
A9isha: we need this for enabling the device simulation flag
`os.environ['XLA_FLAGS'] = f'--xla_force_host_platform_device_count={SIMULATED_CPU_DEVICES_COUNT}'`

MaxText/llama_or_mistral_ckpt.py
A9isha: Done - thanks!

PR #484, opened 2024-03-05
Remove `head` mode for `setup.sh` since we don't use this mode anymore.

tonyjohnchen:
This should also address https://github.com/google/maxtext/issues/480

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Description:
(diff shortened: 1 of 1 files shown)


rwitten (approved):

tonyjohnchen:
`head` mode was a legacy feature we had last year, but I think nobody really use it anymore since `jax/jaxlib/libtpu` are more stable now.

PR #485, opened 2024-03-05
add unit test for checkpoint compatibility between tfds and grain

aireenmei:
Add unit test to make sure checkpoints generated by tfds pipeline and grain pipeline are compatible.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -114,6 +114,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_checkpointing.sh runner gs://runner-maxtext-logs gs://maxtext-dataset False c4-array_record'
+    - name: Test checkpoint compatibility
+      run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/UnitTests.yml
```diff
@@ -114,6 +114,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_checkpointing.sh runner gs://runner-maxtext-logs gs://maxtext-dataset False c4-array_record'
+    - name: Test checkpoint compatibility
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
```
rwitten, line 119: My preference would be for this to be in XLML not GH runner because I believe it is slow relative to the coverage. Feel free to let me know if you disagree.

rwitten (approved):
Giving approval either way but I have a requested change.

.github/workflows/UnitTests.yml
aireenmei: Right now this test runs ~ 1m45s, which is comparable with some other GH unit tests. Given that we are working on some checkpointing change, I would like to add it to GH now. I'll move it to XLML later. Thanks

PR #489, opened 2024-03-06
Add new remat policies

ZhiyuLi-goog:
As titled, add new remat policies providing more granular tradeoffs between speed and HBM memory:

* [`save_dot_except_mlpwi`]: similar to [`SAVE_DOT_EXCEPT_LOGITS_FFN1` in pax](https://github.com/google/praxis/blob/77675370d1150fccda0862a0ac7d1808d4bce9bf/praxis/layers/checkpoint_policy.py#L94). The policy will remat `mlpwi` activations only. `mlpwi` activations are memory demanding, since the `intermediate_dim` is bigger than `embed_dim`.
* [`save_dot_except_mlp`]: reference to [`SAVE_DOT_FOR_MLPERF_200B` in pax](https://github.com/google/praxis/blob/77675370d1150fccda0862a0ac7d1808d4bce9bf/praxis/layers/checkpoint_policy.py#L69), which remat both `mlpwi`  and `mlpwo` activations.

Note, skipped all checkpointing strategies within `AttentionOp` like the activation of `attn_weight` and `wv_product` since the improvement from Flash Attention is dominant as pointed out by Rafi.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -82,7 +82,8 @@ logits_via_embedding: False
 normalize_embedding_logits: True  # whether to normlize pre-softmax logits if logits_via_embedding is true
 logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedding dot product for stability
 
-# proj, minimal, minimal_offloaded, full, or none
+# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'minimal_offloaded' and 'full'.
+# These options offer a trade-off between speed (fastest to slowest) and HBM usage (highest to lowest)
 remat_policy: 'full'
 scan_layers: True
 param_scan_axis: 1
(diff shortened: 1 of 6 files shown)


MaxText/pyconfig.py
```diff
@@ -51,6 +51,19 @@ def validate_attention_type(s: str) -> None:
       "Invalid attention type was passed. Valid options ", valid_attention_types
     )
 
+def maybe_warn_skip_remat_names(keys):
+  if keys['attention'] in ('flash', 'cudnn_flash_te'):
+    skip_remat_names = ()
+    if keys['remat_policy'] in ('proj', 'full'):
```
rwitten, line 57: I think FlashAttention is so dominant as a solution that I'm not interested in these inferior checkpointing strategies because they only serve to confuse. Can you provide justification for why they are necessary?

rwitten (changes requested):

MaxText/pyconfig.py
ZhiyuLi-goog: You are absolutely right. Removed all the remat within `AttentionOp` (where we use FlashAttention instead). By the way, mlperf runs align your insights here, we used FlashAttention and quantize the other matmuls.

rwitten (approved):
Thanks LGTM

PR #491, opened 2024-03-06
use apt-key for installing gcsfuse

aireenmei:
This will solve b/328241961. I tested with the branch
The change: Use "apt-key add -" in both sudo and non-sudo path

Originally, "apt-key add -" is only used in the non sudo path, which is for the docker. The docker image is based on a Debian image that supports apt-key. While the sudo path was for GCE vm, which uses newer ubuntu that is supposed to have apt-key deprecated. But looks apt-key still works in ubuntu 22.04 and won't result in the error in the bug

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -27,34 +27,18 @@
 # Enable "exit immediately if any command fails" option
 set -e
 
-# Check if sudo is available
-if command -v sudo >/dev/null 2>&1; then
-    # sudo is available, use it
-    # install numactl for numa binding.
-    sudo apt update && \
-    sudo apt install -y numactl && \
-    sudo apt install -y lsb-release && \
-    sudo apt install -y gnupg && \
-    sudo apt install -y curl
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -30,6 +30,7 @@ set -e
 # Check if sudo is available
 if command -v sudo >/dev/null 2>&1; then
```
jonb377, line 31: Not blocking this PR, but it seems like the two branches should be the same and the deviation is what broke us. It would be good if we can remove the branch, to the same effect as:

```bash
(sudo bash || bash) <<'EOF'
apt update && \
...
EOF
```

jonb377 (approved):
Thanks Aireen!

setup.sh
aireenmei: That's a good idea. I modified as suggested. Thanks

rwitten (approved):

PR #494, opened 2024-03-08
add compatibility w orbax's new API and add an opttion to restore wit…

ssusie:
Changing checkpointing to use the new API and  replacing `default` with `items`. Also added an option to restore checkpoints with SingleReplicaArrayHandler


.gitignore
```diff
@@ -9,6 +9,8 @@ __pycache__/
 # C extensions
 *.so
 
+orbax/
```
rwitten, line 12: Shouldn't merge this to main?

MaxText/train.py
```diff
@@ -113,6 +116,7 @@ def record_scalar_metrics(metrics, step_time_delta, per_device_tflops, lr):
   })
   metrics['scalar'].update({'learning/current_learning_rate': lr })
 
+
```
rwitten, line 119: nit: please avoid newline changes in random urelated files (unless unavoidable!

end_to_end/test_generate_param_only_checkpoint.sh
```diff
@@ -20,7 +20,7 @@ dry_run=false
 run_id=test_model_0b_$(date +%Y-%m-%d-%H)
 dataset_path=gs://test-maxtext-dataset
 base_output_directory=gs://test-maxtext-output
-ici_tensor_parallelism=8
+ici_tensor_parallelism=4
```
rwitten, line 23: Why did this change happen?

MaxText/checkpointing.py
```diff
@@ -58,11 +60,42 @@ def create_orbax_checkpoint_manager(
...
 def load_state_if_possible(checkpoint_manager: CheckpointManager,
                            data_iterator: Union[MultiHostDataLoadIterator, None],
                            load_parameters_from_path: str,
                            load_full_state_from_path: str,
                            abstract_unboxed_pre_state: train_state.TrainState,
+                           mesh,
+                           state_mesh_annotations,
+                           enable_single_replica_ckpt_restoring: bool,
                            dataset_type: Optional[str] = 'c4'):
   """Loads TrainState as possible from the inputs.
```
rwitten, line 100: if you're changing a signature, change the docstring.

MaxText/checkpointing.py
```diff
@@ -58,11 +60,42 @@ def create_orbax_checkpoint_manager(
...
+                           axis=replica_axis_idx)
+  return np.expand_dims(replica_result, axis=replica_axis_idx)
+
+
 def load_state_if_possible(checkpoint_manager: CheckpointManager,
                            data_iterator: Union[MultiHostDataLoadIterator, None],
                            load_parameters_from_path: str,
                            load_full_state_from_path: str,
                            abstract_unboxed_pre_state: train_state.TrainState,
+                           mesh,
```
rwitten, line 96: BTW I don't understand this change. I would have though the abstract_unboxed_pre_state had all of this information.

rwitten: Can you confirm that you need these new fields and if so I'll read more deeply? I might be missing something.

rwitten (changes requested):
High levels comments because I'm surprised by the change.

end_to_end/test_generate_param_only_checkpoint.sh
ssusie: I was testing it locally on v4-8, so it was failing with 8. Will change back to 8

MaxText/train.py
ssusie: Some indentations/spaces are corrected automatically by editor. Extra new lines after function definition are added according to pystyle guide and pep8 https://peps.python.org/pep-0008/#blank-lines.

.gitignore
ssusie: removed, thanks.

MaxText/checkpointing.py
ssusie: thanks for the note. Removed mesh and state_mesh_annotations from function arguments.

MaxText/checkpointing.py
ssusie: done, thanks.

aireenmei (approved):

MaxText/checkpointing.py
```diff
@@ -93,17 +123,57 @@ def load_state_if_possible(checkpoint_manager: CheckpointManager,
...
+
+
+      def  map_to_pspec(data):
+        pspec = data.sharding.spec
+        mesh = data.sharding.mesh
+        if not enable_single_replica_ckpt_restoring:
+          return orbax.checkpoint.type_handlers.ArrayRestoreArgs(mesh=mesh, mesh_axes=pspec)
+
+        orbax.checkpoint.type_handlers.register_type_handler(
+          jax.Array,
```
rwitten, line 135: nit: is this passing pylint? The multilines really look different from how we normally do this

MaxText/checkpointing.py
```diff
@@ -58,11 +60,40 @@ def create_orbax_checkpoint_manager(
...
+    replica_axis_idx: axis dimension along which replica is taken
+
+  Returns:
+    devices inside the replica that current host is in
+  """
+  idx = _find_idx(device_array, replica_axis_idx)
+  replica_result = np.take(device_array,
+                           idx,
+                           axis=replica_axis_idx)
+  return np.expand_dims(replica_result, axis=replica_axis_idx)
```
rwitten, line 88: Should there be a safety here to detect if this host is part of different replicas? Seems good to detect it here.

MaxText/configs/base.yml
```diff
@@ -82,8 +84,7 @@ logits_via_embedding: False
 normalize_embedding_logits: True  # whether to normlize pre-softmax logits if logits_via_embedding is true
 logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedding dot product for stability
 
-# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'minimal_offloaded' and 'full'.
-# These options offer a trade-off between speed (fastest to slowest) and HBM usage (highest to lowest)
+# proj, minimal, minimal_offloaded, full, or none
```
rwitten: Somehting seems wrong that there is a diff here.

MaxText/train.py
rwitten: The issue is if there start to be other folks who have competing linters running locally. (Like if you really want to run an autocorrector it should run in Cloud -- you don't want to be accidentally messing with whitespace in your CRs, especially not if it competes with someone else who has a competing linter.)

rwitten (approved):
Looks great! One slightly substantive question but giving approval.

Please fix your linting -- if you want us to lint better, that is great but it needs to happen in the cloud. You cannot just run an autolinter locally and I will block future CRs with meaningless whitespace changes.

MaxText/checkpointing.py
ssusie: I didn't see any issues there with pylint but removed the line. There was extra line before def  map_to_pspec, which I removed as well.

MaxText/checkpointing.py
ssusie: Since the check is happening in orbax, I will skip it herehttps://github.com/google/orbax/blob/4d7702f8966a97e6ad0e2a44afe19e95e8df76b4/checkpoint/orbax/checkpoint/type_handlers.py#L1591

we can revisit if we find it useful.

MaxText/train.py
ssusie: OK, sounds good. We can discuss offline how to make the linters better.

gobbleturk (approved):
LGTM but you will need to modify this test too https://github.com/google/maxtext/pull/461 -- probably need to copy a version of the checkpoint to the new directory structure using `gsutil cp` in addition to changing the test's checkpoint path. You may want to reach out in chat about this change to check if there are other cases we are missing

PR #498, opened 2024-03-11
Adding FP8 support through Native XLA Support

abhinavgoel95:
@rwitten, following the discussion on https://github.com/google/maxtext/pull/468, I have made changes to call FP8 GeMMs in MaxText through the native XLA support.

I have tested the performance, and looks good on NVIDIA H100.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -98,7 +98,11 @@ jobs:
     - name: Test int8_training
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
-        'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset quantization=int8 steps=2 enable_checkpointing=false'
+        'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset quantization=int8 steps=2 enable_checkpointing=false' 
+    - name: Test fp8_training
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
+        'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset quantization=fp8 steps=2 enable_checkpointing=false'
     - name: Test generate_param_only_checkpoint
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -59,7 +59,7 @@ save_config_to_gcs: False
 
 # Activation dtypes.
 dtype: "bfloat16"
-quantization: "" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8'
+quantization: "" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8' or use fp8 to run with 8-bit floating-point GeMMs on NVIDIA GPUs.
 quantize_kvcache: False
 
 # Shard the range finding operation for quantization. By default this is set to number of slices.
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -282,7 +282,8 @@ class MoeBlock(nn.Module):
             dtype=self.dtype,
             kernel_init=self.kernel_init,
             kernel_axes=self.kernel_axes,
-            name='gate')(inputs)
+            name='gate',
+            quant=self.quant,)(inputs)
 
     weights, selected_experts = lax.top_k(gate_logits, self.num_experts_per_tok)
     weights = jax.nn.softmax(weights.astype(jnp.float32), axis=-1)
--- a/MaxText/layers/quantizations.py
+++ b/MaxText/layers/quantizations.py
@@ -20,11 +20,21 @@ from aqt.jax.v2 import config as aqt_config
 from aqt.jax.v2.flax import aqt_flax
 from common_types import Array, Config
 from dataclasses import dataclass
+import flax.linen as nn
 import jax
 import jax.numpy as jnp
 from jax.tree_util import tree_flatten_with_path, tree_unflatten
 
 MAX_INT8 = 127.5
+
+@dataclass
+class Quantization:
+    """Base class for quantization configurations"""
+
+    def dot_general_cls(self):
+        """ Placeholder for dot_general implementation in subclasses. """
+        pass
+
 
 @dataclass
 class AqtQuantization:
@@ -50,6 +60,15 @@ class AqtQuantization:
     )
     return aqt_einsum
 
+@dataclass
+class Fp8Quantization(Quantization):
+  """ Configures Fp8 quantization for NVIDIA GPUs"""
+  quant_mode = "train"
+
+  def dot_general_cls(self):
+    """ Returns dot_general configured with aqt params. """
+    return nn.Fp8DotGeneralOp
+
 def _get_quant_config(config):
   """Set quantization params based on user configuration."""
   if not config.quantization or config.quantization == '':
@@ -74,6 +93,8 @@ def _get_quant_config(config):
       dlhs_accumulator_dtype=jnp.int32,
       drhs_accumulator_dtype=drhs_accumulator_dtype,
     )
+  elif config.quantization == "fp8":
+    return "fp8"
   else:
     raise ValueError(f'Invalid value configured for quantization {config.quantization}.')
 
@@ -99,6 +120,8 @@ def configure_quantization(config: Config, quant_mode_str: str = 'train'):
   """ Configure quantization based on user config and quant mode."""
   quant_cfg = _get_quant_config(config)
   if quant_cfg:
+    if quant_cfg == "fp8":
+        return Fp8Quantization()
     quant_mode = get_quant_mode(quant_mode_str)
     return AqtQuantization(quant_dg=quant_cfg, quant_mode=quant_mode)
   return None
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -35,7 +35,7 @@ def main(config):
   tokens, true_length = token_utils.tokenize_and_pad(text, vocab, is_bos=True,
                                                      prefill_lengths=[config.max_prefill_predict_length])
   assert tokens.size <= config.max_prefill_predict_length, "can't take too many tokens"
-
+  assert config.quantization != "fp8", "fp8 on NVIDIA GPUs is not supported in decode.py yet"
   prefill_result = engine.prefill(
       params=params, padded_tokens=tokens, true_length=true_length
   )
--- a/MaxText/layers/models.py
+++ b/MaxText/layers/models.py
@@ -258,6 +258,7 @@ class Decoder(nn.Module):
               'cache': cache_spec,
               'intermediates': 0,
               'aqt':0,
+              '_overwrite_with_gradient': 0,
           },
           split_rngs={
               'params': True,


MaxText/configs/base.yml
```diff
@@ -60,6 +60,7 @@ dtype: "bfloat16"
 quantization: ""
 # Shard the range finding operation for quantization. By default this is set to number of slices.
 quantization_local_shard_count: -1
+use_fp8: False
```
rwitten, line 63: nit: Seems like this should be quantization=fp8 two lines above? Feel free to push back.

rwitten (commented):
Looks really promising. I left one nit.

I want Anfal on Google's side to take the first look, I will make sure he responds super prompotly.

MaxText/layers/linears.py
```diff
@@ -105,6 +106,9 @@ def compute_dot_general(inputs, kernel, axis, contract_ind):
       if self.quant:
         dot_general_cls = self.quant.dot_general_cls()
         dot_general = dot_general_cls()
+      if self.use_fp8:
```
anfals, line 109: Functionally, this looks great!

But, I think it may be cleaner if we added some conditional logic here instead `https://github.com/google/maxtext/blob/main/MaxText/layers/quantizations.py#L95`. We could introduce a general Quantization class (and existing AqtQuantization could subclass). We could then have an FP8 implementation of that class that returns nn.Fp8DotGeneralOp when dot_general_cls() is called. That is a little bit more work but feels a bit better to me. But open to hear other thoughts.

MaxText/max_utils.py
```diff
@@ -348,7 +347,10 @@ def init_initial_state(model, tx, config, is_training, key):
                           jnp.ones(input_shape, dtype=jnp.int32),
                           jnp.ones(input_shape, dtype=jnp.int32))
   if is_training:
-    return init_training_state(model.apply, model_vars['params'], tx)
+    if config.use_fp8:
+      return init_training_state(model.apply, model_vars, tx)
+    else:
+      return init_training_state(model.apply, model_vars['params'], tx)
```
anfals, line 353: Ah is this because the flax linen impl adds those variables in the `overwrite_with_gradients` namespace?

MaxText/train.py
```diff
@@ -216,7 +216,15 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
     for k, v in data.items():
       data[k] = v[:config.global_batch_size_to_train_on,:]
 
-  logits, intermediate_outputs = model.apply({'params': params},
+  if config.use_fp8:
+    logits, intermediate_outputs = model.apply(params,
```
anfals, line 220: same question as above :)

MaxText/max_utils.py
abhinavgoel95: Yes.

MaxText/train.py
abhinavgoel95: Yes, same answer.

MaxText/configs/base.yml
abhinavgoel95: Okay, I can dix that.

MaxText/layers/linears.py
abhinavgoel95: Made some changes. I didn't know what the general Quantization class should look like. So I've omitted it for now. Lmk if you'd still like me to introduce it.

MaxText/configs/base.yml
abhinavgoel95: Done this. @rwitten @anfals

MaxText/layers/quantizations.py
```diff
@@ -47,6 +48,16 @@ def einsum(self):
       )
     )
     return aqt_einsum
+  
+@dataclass
```
anfals, line 52: This is great! Maybe we could have some class that looks like as the parent for the two? More of a nit
```
@dataclass
class Quantization:
    """Base class for quantization configurations"""

    def dot_general_cls(self):
        """ Placeholder for dot_general implementation in subclasses. """
        pass 
```

anfals (commented):
Looks great! Thanks for this! I left one minor nit comment

MaxText/layers/linears.py
anfals: Awesome! Thanks for this. I've left a comment about what I think it could look like but let me know your thoughts!

MaxText/layers/quantizations.py
abhinavgoel95: Cool, Ill do that right away

abhinavgoel95: Done, added what you suggested.

anfals (approved):

anfals (approved):

MaxText/train.py
```diff
@@ -211,7 +211,15 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
     for k, v in data.items():
       data[k] = v[:config.global_batch_size_to_train_on,:]
 
-  logits, intermediate_outputs = model.apply({'params': params},
+  if config.quantization=="fp8":
```
rwitten, line 214: I don't understand this forking. It is because of the difference between passing in `params` and {'params' : params}? Let's find a solution without that forking

MaxText/max_utils.py
```diff
@@ -348,7 +348,10 @@ def init_initial_state(model, tx, config, is_training, key):
                           jnp.ones(input_shape, dtype=jnp.int32),
                           jnp.ones(input_shape, dtype=jnp.int32))
   if is_training:
-    return init_training_state(model.apply, model_vars['params'], tx)
+    if config.quantization=="fp8":
+      return init_training_state(model.apply, model_vars, tx)
```
rwitten, line 352: This is very suspicious to me. Why is it necessary, why can't we find a more universal solution?

rwitten (changes requested):
Some code smell, happy to help folks understand it.

MaxText/max_utils.py
anfals: > Ah is this because the flax linen impl adds those variables in the overwrite_with_gradients namespace?

@rwitten See the resolved conversation below. `fp8ops` has the "hack" to have some variables keyed under `overwrite_with_gradients` so they can be manually overwritten, which leads to this weirdness. But agree - its odd. Maybe there is a better way to do this...

anfals: @abhinavgoel95 Hmm what happens if we just pass in `model_vars` always here? Looks like before in MaxText we were extracting the params directly, but Flax should be able to just infer the direct parameter collection as being equivalent to `{'params': params}`

abhinavgoel95: This is needed by the Fp8 implementation in Flax.

MaxText/train.py
abhinavgoel95: cc @kaixih, do you know if there is a way around this?

MaxText/max_utils.py
abhinavgoel95: cc @kaixih

MaxText/train.py
anfals: @abhinavgoel95 I'll also poke around with this a bit more today. I think there should be a way to just make this consistent throughout maxtext such that we just pass `params` around directly (so it'll always be {'params': params} by default). But that might take a little bit of extra work in this PR. I'll report back

kaixih: The current behavior of TrainState is such that the pytree structure of optimizer states depends on what the user provides:

1. If the input is {'params': {'w': xxx, 'b': xxx}, 'other_params': yyy}, the optimizer state will be {'params': {'w': xxx, 'b': xxx}, 'other_params': yyy}.
2. If the input is {'w': xxx, 'b': xxx}, the optimizer state will be {'w': xxx, 'b': xxx}.
3. If the input is {'params': {'w': xxx, 'b': xxx}, 'other_params': yyy, 'overwrite_with_gradient': zzz}, the optimizer state will be {'w': xxx, 'b': xxx}.

The behavior of (1) and (2) represents the optimizer states when we pass `model_var` and `model_var['params']` into TrainState. However, (3) passes the full model variables while the optimizer state only focuses on 'params' and disregards 'other_params', which might need some improvement if we do have any real-world use case that we need optimizer states for non 'params' variables.

Back to our scenario here, we are employing method (2) by originally passing `model_var['params']`. The proposed change entails utilizing (3) for fp8 configuration by passing the entire `model_vars`.

Considering this, if there are no real instances where `other_params` necessitates optimizer states, I suggest modifying the TrainState behavior in Flax for (1) as follows:

(1) {'params': {'w': xxx, 'b': xxx}, 'other_params': yyy} -> opt {'w': xxx, 'b': xxx}

With this adjustment, the TrainState behavior remains consistent: it will generate optimizer states only for 'params' if they exist (cases 1 and 3); otherwise, it will generate states for all items (case 2). Consequently, there's no need for branching here, and we can simply pass in the entire `model_vars`:
```python
init_training_state(model.apply, model_vars, tx)

```

anfals: @kaixih Thank you for the thorough explanation! The behavior in (3) is strange to me - I don't understand why it differs from (1). I can raise this internally with the Flax team. That said, I think I might have a workaround for us here: 

@abhinavgoel95 For this PR, I did a little bit of debugging of MaxText (on a TPU, but I think this should generalize). Within `init_initial_state`, `model_vars` looks like the following: 

`{'params': {'token_embedder': {...}, 'decoder': {...}}}`

So we can just pass it directly to `init_training_state` directly. And then within `loss_fn` in `train.py`, we can just pass `params` along directly. With just those changes, I was able to get training running. And I believe it should generalize to your changes (I'll try to verify that today). 

That said, I think MaxText has been a little inconsistent with this params versus params[params], so I think we might get some test errors, so it might be take a bit of work to get this done properly.

anfals: One other thing that would be good to have in this PR is some kind of additional smoke test for the fp8 case. Along with the `params` testing, that is increasing the scope of the PR, so it might be good to separate the two out into separate PRs (once we verify that the change I proposed actually helps) - 1 to switch just using params everywhere instead of params[parmas]) and then the fp8 work + test. If you have bandwidth/want to do both, great! If not, I'm happy to do the first PR and then you can incorporate the changes here.

hxssgaa: @abhinavgoel95 I ran the code on H100 with FP8 quantization, but faced following issue.

```
flax.errors.ScopeCollectionNotFound: Tried to access "input_amax_history" from collection "_overwrite_with_gradient" in "/decoder/remat(scan(layers))/self_attention/query/Fp8DotGeneralOp_0" but the collection is empty. (https://flax.readthedocs.io/en/latest/api_reference/flax.errors.html#flax.errors.ScopeCollectionNotFound)
```

Is this error due to the discussion above?

anfals: @hxssgaa That looks to be related to some errors related to scan layers. That will be fixed before this PR is merged. You can try setting `scan_layers` to False in the meantime

abhinavgoel95:
@anfals @rwitten I have rebased the code on top of the latest changes. Looks cleaner now. Lmk what you think. Also works with scan = True.

anfals (changes requested):
This is great! Thank you for all the work on this so far!

I left some comments with a few changes I recommend. In addition to those, I'd suggest the following: 

1) Decoding doesn't work yet with 'fp8' so I think we could put an assert somewhere like `decode.py` that prevents decoding when fp8 is enabled (which we can remove in the follow-up PR with decode support)
2.) We should include a test with this PR. Maybe the equivalent of `train_int8_smoke_test` for fp8?

Beyond that, I verified the latest changes in main gave us checkpointing support for the `_overwrite_with_gradients` collection for free for training. It's breaking in decoding for some reason, but there is other work that needs to be done there, so we can kick the can on that until the next PR.

anfals:
> This is great! Thank you for all the work on this so far!
> 
> I left some comments with a few changes I recommend. In addition to those, I'd suggest the following:
> 
> 1. Decoding doesn't work yet with 'fp8' so I think we could put an assert somewhere like `decode.py` that prevents decoding when fp8 is enabled (which we can remove in the follow-up PR with decode support)
>    2.) We should include a test with this PR. Maybe the equivalent of `train_int8_smoke_test` for fp8?
> 
> Beyond that, I verified the latest changes in main gave us checkpointing support for the `_overwrite_with_gradients` collection for free for training. It's breaking in decoding for some reason, but there is other work that needs to be done there, so we can kick the can on that until the next PR.

Update on this: checkpointing 'works' with decoding, but the post-processing of the params in `load_params` within MaxEngine is very AQT-specific for quantization. We'll need to make some changes there to support the `_overwrite_with_gradients` collection with this FP8 quantization in the follow-up PR

MaxText/configs/base.yml
```diff
@@ -59,7 +59,7 @@ save_config_to_gcs: False
 
 # Activation dtypes.
 dtype: "bfloat16"
-quantization: "" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8'
+quantization: "fp8" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8' or use fp8 to run with 8-bit floating-point GeMMs on NVIDIA GPUs.
```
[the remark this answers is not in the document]
abhinavgoel95: Done.

MaxText/train.py
abhinavgoel95: Should be fixed now. Both with and without scan works fine.

anfals (approved):
LGTM, thanks!

.github/workflows/UnitTests.yml
```diff
@@ -99,6 +99,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset quantization=int8 steps=2 enable_checkpointing=false'
+    - name: Test fp8_training
```
anfals, line 102: We might wanna add this to the GPU tests below, but we can do that in the follow-up decode PR :). This should do a good job covering the functionality added!

anfals (approved):
LGTM, pending the tests passing!

rwitten (approved):

anfals (approved):

anfals (approved):

PR #502, opened 2024-03-12
Mitali's KV Cache Quantization

rwitten:
Hurts perf on v4-8, but I expect it to help on v5e-8.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -60,6 +60,8 @@ save_config_to_gcs: False
 # Activation dtypes.
 dtype: "bfloat16"
 quantization: "" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8'
+quantize_kvcache: False
+
 # Shard the range finding operation for quantization. By default this is set to number of slices.
 quantization_local_shard_count: -1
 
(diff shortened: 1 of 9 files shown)


MaxText/configs/base.yml
```diff
@@ -58,6 +58,8 @@ save_config_to_gcs: False
 # Activation dtypes.
 dtype: "bfloat16"
 quantization: "" #defaults to no quantization, i.e. bf16. possible alternative setting is 'int8'
+quantize_kvcache: False
```
vipannalla, line 61: If quantize_kvcache is set to True, we will use the type from "quantization" setting as the target precision? Is my understanding correct?

singh-mitali: The two are independent at present. Quantize_kvcache only quantizes to int8 and is independent of the basic quantization config.

MaxText/layers/quantizations.py
```diff
@@ -126,3 +129,17 @@ def remove_quantized_params(params, aqt_vars):
       v = {}
     tree_flat[i] = v
   return tree_unflatten(tree_struct, tree_flat)
+
+def configure_kv_quantization(config: Config):
```
singh-mitali, line 133: We can remove this as its not used anymore.

singh-mitali (approved):

PR #503, opened 2024-03-12
Fix training command of Gemma in Readme

khatwanimohit:


rwitten (approved):
Approved but we need to have a much more holistic chat to get these to the quality we expect.

PR #504, opened 2024-03-12
Pin pytype version

khatwanimohit:


rwitten (approved):
Approved! Please makea. ticket

PR #506, opened 2024-03-12
Fix gemma test

jonb377:
A typo in https://github.com/google/maxtext/pull/488 broke a gemma test.


jonb377:
Fixed in https://github.com/google/maxtext/pull/494

PR #510, opened 2024-03-13
Lock dependencies with poetry

chajath:
The point of locking is so that we can reproduce builds at any given commit hash. Will follow up with docker to build images out of poetry project definition.


chajath:
Closing in favor of #547

PR #512, opened 2024-03-13
Modify convergence test to read in run_name from env

raymondzouu:
Modifying the script so that the corresponding XLML test can generate the run_name and pass it as an env variable. This is needed for the XLML test to collect the outputted metrics.


gobbleturk:
Looks fine but I'm confused why this is necessary - the only difference is the run_name is also exported as M_RUN_NAME - is that what the outputted metrics need, the M_RUN_NAME env var?

gobbleturk (approved):

raymondzouu:
> Looks fine but I'm confused why this is necessary - the only difference is the run_name is also exported as M_RUN_NAME - is that what the outputted metrics need, the M_RUN_NAME env var?

@gobbleturk This is because the XLML test will generate the run_name and pass it in through M_RUN_NAME so that it knows where to find the tensorboard file that contains the metrics. This change is so that run_name isn't passed through both the env and the command line which would cause an error.

gobbleturk:
> > Looks fine but I'm confused why this is necessary - the only difference is the run_name is also exported as M_RUN_NAME - is that what the outputted metrics need, the M_RUN_NAME env var?
> 
> @gobbleturk This is because the XLML test will generate the run_name and pass it in through M_RUN_NAME so that it knows where to find the tensorboard file that contains the metrics. This change is so that run_name isn't passed through both the env and the command line which would cause an error.

Ah I see, thank you for explaining

rwitten (approved):

PR #518, opened 2024-03-14
Constraint dependency versions

chajath:
Workaround of https://github.com/google/maxtext/issues/516

Also pin other dependencies for mostly reproducible container build


chajath:
Closing in favor of #547

PR #519, opened 2024-03-14
Change gpu docker images project_id

NinaCai:
Change gpu docker images location to `supercomputer-testing`


gobbleturk:
Should we push to both projects? Is there any benefit to having the tpu-prod-env-multipod one?

rwitten (changes requested):
Sorry we shouldn't be moving this around -- eventually folks will come to depend on it being in one place.

Why can't supercomputer-testing pull from our project? Happy to give them the right permissions.

PR #524, opened 2024-03-15
Add Llama2 13b

morgandu:


end_to_end/test_llama2_13b.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
```
rwitten, line 2: Why two scripts? They seem duplicative?

rwitten (changes requested):
Little nit about two scripts.

Mostly, please work with Ran to figure out how to package multiple scripts that need to run on multiple devices in XLML. It is important that our tests serve as documentation for our users so we'd need to mark them as tied together, etc.

end_to_end/test_llama2_13b.sh
morgandu: the *_chkpt.sh does checkpoint conversions only, one from meta to maxtext, and one from scanned to unscanned. 

The generate_param_only_checkpoint takes a diff arg when the checkpoint is from pretrained/finetuned than its converted directly from meta. This was the error both 13b and 70b encountered during unroll conversion. Also, after the unroll conversion, added another check for golden output.

Happy to remove the *_chkpt.sh altogether, or add the unroll conversion for meta converted chkpt and decode output validation in the main script.

rwitten: I think this flow would be better if we documented the full user journey for llama-13b in one place. Happy to troubleshoot the bug.

(Also, can you think about what user facing changes we'd want to make this usable? For example, I think you're targeting a v4-16. Can you document that?)

rwitten (changes requested):

end_to_end/test_llama2_13b.sh
morgandu: Updated the testing script with more comment, can also provide a readme for the 13b

morgandu: Updated the testing script with more comment, I can also write a 13b readme, once confirm the issue when using v4-16 vs v4-128

PR #539, opened 2024-03-20
Consistent Usage of Params Within MaxText

anfals:
#498 Showed that we have inconsistent use of `params` vs `{'params': params}` within MaxText, which causes problems when we have changes that require multiple collections of variables (like https://github.com/google/flax/blob/main/flax/linen/fp8_ops.py). 

These changes streamline this as much as possible to use `params` directly, except for places that break out of standard Jax/Flax and have more complexity.


MaxText/maxengine.py
```diff
@@ -115,7 +111,7 @@ def model_apply(_p, _rng):
       params = {}
       params['aqt'] = new_vars['aqt']
       # Remove param values which have corresponding qtensors in aqt to save memory.
-      params['params'] = quantizations.remove_quantized_params(state.params, new_vars['aqt'])
+      params['params'] = quantizations.remove_quantized_params(state.params['params'], new_vars['aqt'])
```
anfals, line 114: Have to do state.params['params'] here because we are breaking away from the flax world that deals with {"params: stuff} nicely and doing our own thing in quantizations. _sigh_

.github/workflows/UnitTests.yml
```diff
@@ -110,6 +110,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_generate_param_only_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4'
+    - name: Test generate_param_only_checkpoint with int8 quantization
```
anfals, line 113: There were some code paths that our test suites missed that broke with these changes. This should cover them

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "python.testing.pytestArgs": [],
```
anfals, line 2: We already have a shared launch config and it took me a bit of time to figure out how to get the unit tests to work in VSCode as a Jetbrains user. so this might be helpful for others doing development.

PR #540, opened 2024-03-20
Flip env sensing so we can launch gpu with checkpointing

chajath:
A better fix would be to fully embrace `ClusterEnv`'s visitor pattern and write a subclass that could work with ENVs that we set from the launch script, and it is orthogonal to whether we use GPU or not. That can be done in a follow-up PR.


rwitten (commented):
I'm asked Nvidia to confirm this doesn't break them.

chajath:
Closing in favor of https://github.com/google/maxtext/pull/697

PR #543, opened 2024-03-22
Combine prefill/generate profiles; allow passing in profile name

morgandu:


MaxText/inference_microbenchmark.py
```diff
@@ -176,6 +169,9 @@ def main(config):
   num_model_params, model_size, _ = summarize_pytree_data(params, name="Model")
 
   benchmark_results = {"Prefill": {}}
+
+  max_utils.activate_profiler(config, config.profile_name)
```
patemotter, line 173: Moving the profiling into `main` may not be ideal. 
1. It captures additional setup steps that are done as part of `ar_benchmark` and `prefill_benchmark` before running the corresponding benchmark loops. 
2. It puts all the prefill and ar kernels and setup calls into one single profile, which may or may not be a bad thing. To me it introduces more noise and makes it a little harder to clearly see the pieces we care about. But I can also see the appeal of having a single profile per run.

MaxText/configs/base.yml
```diff
@@ -194,6 +194,8 @@ prompt: "I love to" # Prompt for language model sampling.
 load_from_prefill_dir: False # If true, decode.py doesn't "prefill" but just reads from directory
 prefill_cache_dir: "" # If set and load_from_prefill_dir, decode.py reads from directory. If set, decode.py writes to directory
 autoregressive_decode_assert: ""
+microbenchmark_prefill_lengths: "64,128,256,512,1024" # The prefill lengths used for microbenchmark
+profile_name: "" # The profile prefix name for microbenchmark
```
patemotter, line 198: If you are adding this to the config then I don't think we can assume that it would only ever be used by the microbenchmark even though that's the current situation.

MaxText/configs/base.yml
```diff
@@ -194,6 +194,8 @@ prompt: "I love to" # Prompt for language model sampling.
 load_from_prefill_dir: False # If true, decode.py doesn't "prefill" but just reads from directory
 prefill_cache_dir: "" # If set and load_from_prefill_dir, decode.py reads from directory. If set, decode.py writes to directory
 autoregressive_decode_assert: ""
+microbenchmark_prefill_lengths: "64,128,256,512,1024" # The prefill lengths used for microbenchmark
```
patemotter, line 197: Can we not we just use regular array notation in the config and load it as a list of ints?
`microbenchmark_prefill_lengths: [64, 128, 256, 512, 1024]`

patemotter:
Your pylint is failing, which you can run locally beforehand via `pylint MaxText/`.

MaxText/configs/base.yml
morgandu: I ran into below error when passing list of ints
```
ValueError: For key 'microbenchmark_prefill_lengths', type <class 'list'> not in dict_keys([<class 'str'>, <class 'int'>, <class 'float'>, <class 'bool'>]), can't pass at the CLI or ENV
```

MaxText/configs/base.yml
rwitten, line 198: probably we don't want this? agreed with Pate.

MaxText/inference_microbenchmark.py
rwitten: I think on the balance it would be better if it was done tastefully (a.k.a. we avoid the compiles in the profile so the profiles were lean and mean).

MaxText/configs/base.yml
rwitten: Stupid maxtext thing. You can see how we do this for the sharding rules but actually the comma separated list is much better at the command line and I think it is fine. (I think this is an improvement either way.)

rwitten (changes requested):
Gave some feedback. I think we should look for consensus here and only go forward with the one profile if users on the balance like the idea.

PR #547, opened 2024-03-25
Pinned build mode for GPU, with prebuilt Transformer Engine bdist

chajath:
Also add constraints to GPU stable build which are cherry-picked from #522 and #525 

To update the wheel, take a look at maxtext_transformerengine_builder.Docker. Once the image is built, you can extract the wheel file under `/root/TransformerEngine/dist` of the image. Unfortunately, there's no one-liner to copy a file from offline docker image, but there's no shortage of workarounds: https://stackoverflow.com/q/25292198

To upload the wheel, builder who is authorized to upload to the artifact registry can do:

```
pip install twine keyrings.google-artifactregistry-auth wheel
twine upload --repository-url=https://us-python.pkg.dev/gce-ai-infra/maxtext-build-support-packages/ ./transformer_engine-1.4.0+$GITHASH-cp310-cp310-linux_x86_64.whl
```

Where `$GITHASH` is the commit hash version of the Transformer Engine

For nightly builds, we don't apply constraints so the image can be built with the latest and greatest packages.

We also copy setup.sh, requirements.txt and constraints.txt first before executing setup.sh, to avoid running redundant setup script when there is no dependency change. We also mount pip cache from host to speed up the build process.


constraints.txt
```diff
@@ -0,0 +1,135 @@
+absl-py==1.4.0
```
rwitten, line 1: Hey -- we need to have a bit more of a holistic conversation here, probably a document describing the user journeys would be best. I ABSOLUTELY think pinning things absolutely has a place! But we need to figure out the place.

Zooming out, we've offered two modes of maxtext historically, MaxText on JaxNightly and JaxStable. (And we've tested both etc.)

I think @chajath's observation is we woudl like to offer a third mode, MaxText on Pinned. For MaxText on pinned, we would have everything pinned. And there would need to be a periodic process of turning a "golden" JaxStable into a Pinned version.

(Separately, we might also experiment with having MaxText releases that freeze a MaxText commit. But either way a MaxText commit + MODE=pinned == rock solid point in time).

But I'm very hesitant to make changes without a bit of a clearer plan. (Who updrades the pinned closure? When? What is the playbook?) And fundamental engineering support (it seems to me like pinned is an alternative to stable/nightly so it should be a supported MODE?).

And I've been getting pings on this from @chajath and Aman Gupta. I'll intro you two via google chat.

rwitten (changes requested):

.github/workflows/UploadDockerImages.yml
```diff
@@ -50,4 +50,7 @@ jobs:
         bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_jax_stable MODE=stable DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_local_jax_stable
     - name: build jax nightly image
       run : |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_jax_nightly MODE=nightly DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_local_jax_nightly
\ No newline at end of file
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_jax_nightly MODE=nightly DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_local_jax_nightly
+    - name: build jax pinned image
```
NinaCai, line 54: Does it need to be built daily? In that case we don't save any time, right?

chajath:
TE version has been updated in the anticipation of #555  landing soon. Once it lands, I will rebase and make sure to do some sanity check on all builds.

.github/workflows/UploadDockerImages.yml
chajath: Daily build would be useful in catching any dependency leakage issues. Pinning process as presented here is best-effort and isn't totally fool proof, so I would rather have some checks on pinning builds not getting affected by version drifts.

chajath: As per offline discussion: let's build all three build modes daily.

maxtext_transformerenginer_builder.Dockerfile
yangyuwei, line 1: nit: Rename as maxtext_transformerengine_builder.Dockerfile?

setup.sh
yangyuwei, line 1: May I know what's "--extra-index-url https://us-python.pkg.dev/gce-ai-infra/maxtext-build-support-packages/simple/" used for?

chajath: This is so that pip will know where to find prebuilt wheel, which is currently stored in the referenced artifact registry path

yangyuwei: Got it. Is it accessible publicly?

constraints.txt
chajath: Following up on the offline discussion - let's punt renaming build modes as a follow-up task, and for now let's settle on `nightly`, `stable`, and `pinned` modes. Could you take another look, please? Thanks!

setup.sh
chajath: Yes - any developer will be able to build it. Although uploading to the registry will require credential.

yangyuwei (approved):
Thanks In-Ho! This is really helpful!

PR #554, opened 2024-03-26
Consistent Usage of Params Within MaxText

anfals:
https://github.com/google/maxtext/pull/498 Showed that we have inconsistent use of params vs {'params': params} within MaxText, which causes problems when we have changes that require multiple collections of variables (like https://github.com/google/flax/blob/main/flax/linen/fp8_ops.py).

These changes streamline this as much as possible to use params directly, except for places that break out of standard Jax/Flax and have more complexity.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -110,6 +110,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_generate_param_only_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4'
+    - name: Test generate_param_only_checkpoint with int8 quantization
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
+        'bash end_to_end/test_generate_param_only_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4 -q int8'
     - name: Test grain checkpoint determinism
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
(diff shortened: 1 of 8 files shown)


.github/workflows/UnitTests.yml
```diff
@@ -110,6 +110,10 @@ jobs:
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
         'bash end_to_end/test_generate_param_only_checkpoint.sh -r runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} -o gs://runner-maxtext-logs -d gs://maxtext-dataset -i 4'
+    - name: Test generate_param_only_checkpoint with int8 quantization
```
anfals, line 113: There were some code paths that our test suites missed that broke with these changes. This should cover them

MaxText/maxengine.py
```diff
@@ -115,7 +111,7 @@ def model_apply(_p, _rng):
       params = {}
       params['aqt'] = new_vars['aqt']
       # Remove param values which have corresponding qtensors in aqt to save memory.
-      params['params'] = quantizations.remove_quantized_params(state.params, new_vars['aqt'])
+      params['params'] = quantizations.remove_quantized_params(state.params['params'], new_vars['aqt'])
```
anfals, line 114: Alas, we need to be more explicit here now because we are specifically pruning one collection (params) with another (aqt); before `state.params` was just gonna be the parameters directly and the variable keys lined up, so we need to index into params to get that same consistency.

.github/workflows/UnitTests.yml
rwitten: Oh interesting, we haven't really considered generating quantized checkpoints -- we assumed quantization would be in the node after loading. How would we load a quantized checkpoint?

rwitten (approved):

.github/workflows/UnitTests.yml
anfals: This question of any special considerations around loading a quantized checkpoint is on my mind for the fp8 work especially. With AQT/int8 today, I believe we are just loading params as is and then doing a dummy inference to quantize the weights (and then saving those quantized weights). This test covers that code path (see https://github.com/google/maxtext/pull/554/files#diff-f521b06e43d6c7cab9e2f65a5eb4fa64cd129e63f1796154b31b42cf4481af68R99). 

It'll get more complex with FP8, where we'll have histories and scales that are variables of their own. I'll need to do some digging.

PR #556, opened 2024-03-26
add validation for shardings

ssusie:
By default the models will run a quick check to make sure whether more than 99% of weights are sharded across 'fsdp', 'fsdp_transpose','sequence', 'tensor' axes

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -61,6 +61,19 @@ def calculate_num_params_from_pytree(params):
   assert total_parameters >= 0
   return total_parameters
 
+
+def calculate_total_params_per_chip(params):
+  def calculate_leaf_params_per_chip(arr):
+    shard = arr.addressable_shards[0]
+    return np.prod(shard.data.shape)
+
+  params_sizes_per_chip = jax.tree_util.tree_map(
+    calculate_leaf_params_per_chip, params)
+  total_parameters_per_chip = jax.tree_util.tree_reduce(
+    lambda x, y: x + y, params_sizes_per_chip)
+  return total_parameters_per_chip
+
+
 def calculate_bytes_from_pytree(params):
   params_bytes = jax.tree_util.tree_map(lambda x : x.nbytes, params)
   total_bytes = jax.tree_util.tree_reduce(lambda x, y: x + y, params_bytes)
--- a/MaxText/maxtext_utils.py
+++ b/MaxText/maxtext_utils.py
@@ -18,6 +18,7 @@
 """Utils that are only interesting to MaxText. """
 
 import jax
+import max_utils
 from jax.sharding import PartitionSpec as P
 from jax.experimental.serialize_executable import deserialize_and_load
 
@@ -124,3 +125,41 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
           f'\t\tCausal attention TFLOPs: {causal_attention_tflops} ',
           f'({100 * causal_attention_tflops/total_tflops:.2f})% of Total')
   return total_tflops, learnable_weight_tflops, causal_attention_tflops
+
+
+def assert_params_sufficiently_sharded(params, mesh, tolerance=0.01):
+  """Checks whether most params are sharded across sharding axis.
+
+  This function determines whether the majority of parameters  are distributed
+  across a specified sharding axes with an acceptable tolerance. It compares the
+  current distribution to a scenario where all parameters are fully sharded
+  across the 'fsdp', 'fsdp_transpose', 'sequence', and 'tensor' axes.
+
+  Args:
+    params: params of the model state
+    mesh: mesh constructed from config
+    tolerance: float between 0.0 and 1.0 representing the allowed percentage of
+    non-sharded parameters.
+  Returns:
+    bool: True if the majority of parameters are sufficiently sharded
+  """
+  total_num_params = max_utils.calculate_num_params_from_pytree(params)
+  product_num_devices_for_weight_sharding  = 1
+  for axis in ['fsdp', 'fsdp_transpose', 'sequence', 'tensor']:
+    product_num_devices_for_weight_sharding *= mesh.shape[axis]
+  total_num_params_per_chip = (
+    max_utils.calculate_total_params_per_chip(
+      params)
+  )
+  perfectly_sharded_params_per_chip = (
+    total_num_params / product_num_devices_for_weight_sharding
+  )
+  assert total_num_params_per_chip >= perfectly_sharded_params_per_chip, (
+    'Number of parameters per chip must not be less than in the ideal sharded '
+    'scenario accross `fsdp`, `fsdp_transpose`,`sequence`, `tensor` axes.'
+    )
+  assert (
+    total_num_params_per_chip/perfectly_sharded_params_per_chip - 1 < tolerance
+    ), (f'Number of unsharded parameters exceeds tolerance {tolerance * 100}% '
+      'of total parameters.')
+
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -360,6 +360,8 @@ def setup_train_loop(config):
   state, state_mesh_annotations, data_iterator = max_utils.setup_training_state(model, data_iterator,
           tx, config, init_rng, mesh, checkpoint_manager)
 
+  maxtext_utils.assert_params_sufficiently_sharded(state.params, mesh)
+
   return ( init_rng, writer, checkpoint_manager, state_mesh_annotations, model,
           mesh, learning_rate_schedule, data_iterator, eval_data_iterator, state)
 


MaxText/maxtext_utils.py
```diff
@@ -124,3 +126,53 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+  new_num_p = np.prod(shard.data.shape)
+
+  if new_num_p == expected_per_device_num_param:
+    print(f'Input is sharded over {num_devices_sharded} devices as expected!')
+  else:
+    print(f'Expected {expected_per_device_num_param} params but got {new_num_p}')
+    print(f'Off by a factor of {new_num_p // expected_per_device_num_param}')
+    print('key path', key)
+    print('pspec', pspec)
+    if jax.process_index() == 0:
```
rwitten, line 147: I don't understand why there should be any process_index stuff in this code? This is universal -- we can check this for all devices?

MaxText/maxtext_utils.py
```diff
@@ -124,3 +126,53 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+                      pspec,
+                      num_devices_sharded
+                      ):
+  """Returns the number of params that aren't sharded."""
+  new_num_p = np.prod(shard.data.shape)
+
+  if new_num_p == expected_per_device_num_param:
+    print(f'Input is sharded over {num_devices_sharded} devices as expected!')
+  else:
+    print(f'Expected {expected_per_device_num_param} params but got {new_num_p}')
```
rwitten, line 143: Let's match style as use max_logging.

rwitten: But as I say above this script should be totally silent unless you assert.

MaxText/train.py
```diff
@@ -340,6 +340,9 @@ def setup_train_loop(config):
   state, state_mesh_annotations, data_iterator = max_utils.setup_training_state(model, data_iterator,
           tx, config, init_rng, mesh, checkpoint_manager)
 
+  max_logging.log('Validating shardings of layers')
+  assert maxtext_utils.is_sharded_correctly(state.params)
```
rwitten, line 344: Probably put the assert into the check? But just stylistic.

MaxText/maxtext_utils.py
```diff
@@ -124,3 +126,53 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+    for axis in ['fsdp', 'fsdp_transpose','sequence', 'tensor']:
+      num_devices_sharded *= pytree.sharding.mesh.shape[axis]
+
+    num_params = max_utils.calculate_num_params_from_pytree(pytree)
+    expected_num_p = num_params // num_devices_sharded
+    for shard in pytree.addressable_shards:
+      num_not_sharded_params[0] += calc_not_sharded_params(
+        shard, expected_num_p, key, pspec, num_devices_sharded) // jax.local_device_count()
+
+  jax.tree_util.tree_map_with_path(
```
rwitten, line 174: I'm not really following this CR?

Can't we just:
```
total_num_params = max_utils.calculate_num_params_from_pytree(params)
total_num_params_per_chip = max_utils.calculate_num_params_per_chip_from_pytree(params) ###you can implement
product_num_mesh_axes_for_weight_sharded = ### you can implement but it is the number of devices in fsdp, fsdp_transpose, sequence and tensor -- you shouldn't have to look at any specific jax Arrays.
assert total_num_params_per_chip < 1.01*total_num_params/product_num_mesh_axes_for_weight_sharded
```

MaxText/maxtext_utils.py
```diff
@@ -124,3 +126,53 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+    print(f'Expected {expected_per_device_num_param} params but got {new_num_p}')
+    print(f'Off by a factor of {new_num_p // expected_per_device_num_param}')
+    print('key path', key)
+    print('pspec', pspec)
+    if jax.process_index() == 0:
+      return new_num_p
+  return 0
+
+def is_sharded_correctly(params):
+  """Checks whether most params are sharded accros sharding axis.
```
rwitten, line 152: nit: across

MaxText/maxtext_utils.py
```diff
@@ -124,3 +126,53 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+  else:
+    print(f'Expected {expected_per_device_num_param} params but got {new_num_p}')
+    print(f'Off by a factor of {new_num_p // expected_per_device_num_param}')
+    print('key path', key)
+    print('pspec', pspec)
+    if jax.process_index() == 0:
+      return new_num_p
+  return 0
+
+def is_sharded_correctly(params):
```
rwitten, line 151: Let's just call this whatever we end up wanting more descriptively (instead of `is_sharded_correctly`)

Probably assert_params_sufficiently_sharded(params, mesh, tol=1.01)

rwitten (changes requested):
Lots of nits, but mostly not sure I follow the overall structure.

MaxText/maxtext_utils.py
ssusie: The check happens on all devices. When it is not sharded (i.e. replicated on all) I return the number only for process 0. Otherwise the number of params will be a multiple of all devices. I also later divide by 4 because process 0 has 4 devices. The idea is that is something isn't sharded I want to count it only once.

MaxText/maxtext_utils.py
ssusie: I thought of using max_logging, but kept is consistent with "print" statements in the maxtext_utils.py. I will remove printing from here and will switch other prints to max_logging then.

MaxText/maxtext_utils.py
ssusie: I am not computing all the params per chip, but only not sharded params. E.g. if there is a weight of size 2048 that needs to be sharded on 4 devices and the number of params per device isn't 512, I am adding 2048 to the total number of not sharded devices. And I add 2048 only once because only process 0 does that. I can change the logic to what you have of course. That's another approach, but both work.

MaxText/maxtext_utils.py
ssusie: fixed

MaxText/maxtext_utils.py
ssusie: Rewrote the logic, this doesn't apply anymore. Closing for now.

MaxText/maxtext_utils.py
ssusie: Reimplemented the validation to match the logic you mentioned above. Checked the two are giving same results. Will close the comment for now. Please reopen if needed, thanks.

MaxText/maxtext_utils.py
ssusie: thanks, changed naming.

MaxText/train.py
ssusie: Done. thanks.

MaxText/max_utils.py
```diff
@@ -61,6 +61,18 @@ def calculate_num_params_from_pytree(params):
   assert total_parameters >= 0
   return total_parameters
 
+
+def calculate_total_params_per_chip(params):
+  def calculate_params_per_chip(pytree):
```
rwitten, line 66: `calculate_params_per_chip_leaf`

MaxText/maxtext_utils.py
```diff
@@ -124,3 +125,39 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+
+  Args:
+    params: params of the model state
+    mesh: mesh constructed from config
+    tolerance: float between 0.0 and 1.0 representing the allowed percentage of
+    non-sharded parameters.
+  Returns:
+    bool: True if the majority of parameters are sufficiently sharded
+  """
+  total_num_params = max_utils.calculate_num_params_from_pytree(params) * 1e9
```
rwitten, line 146: why is there a 1e9?

rwitten: (seems hacky!)

MaxText/maxtext_utils.py
```diff
@@ -124,3 +125,39 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+  for axis in ['fsdp', 'fsdp_transpose','sequence', 'tensor']:
+    product_num_devices_for_weight_sharding *= mesh.shape[axis]
+  total_num_params_per_chip = (
+    max_utils.calculate_total_params_per_chip(
+      params)
+  )
+  perfectly_sharded_params_per_chip = (
+    total_num_params / product_num_devices_for_weight_sharding
+  )
+  assert (
```
rwitten, line 157: also `assert total_num_params_per_chip >= perfectly_sharded_params_per_chip`

MaxText/maxtext_utils.py
```diff
@@ -124,3 +125,39 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+    mesh: mesh constructed from config
+    tolerance: float between 0.0 and 1.0 representing the allowed percentage of
+    non-sharded parameters.
+  Returns:
+    bool: True if the majority of parameters are sufficiently sharded
+  """
+  total_num_params = max_utils.calculate_num_params_from_pytree(params) * 1e9
+  product_num_devices_for_weight_sharding  = 1
+  for axis in ['fsdp', 'fsdp_transpose','sequence', 'tensor']:
+    product_num_devices_for_weight_sharding *= mesh.shape[axis]
```
rwitten, line 149: how does this work? Isn't mesh an array? Is `mesh.shape` actually a dictionary? Lots I don't understand here.

MaxText/train.py
```diff
@@ -342,6 +342,9 @@ def setup_train_loop(config):
   state, state_mesh_annotations, data_iterator = max_utils.setup_training_state(model, data_iterator,
           tx, config, init_rng, mesh, checkpoint_manager)
 
+  max_logging.log('Validating shardings of layers')
```
rwitten, line 345: Please remove this comment -- it is superfluous?

MaxText/maxtext_utils.py
rwitten: And for both of these you need to add a comment such as `assert BOOL, "comment explaining problem and how to resolve"`

rwitten (changes requested):
Structure looks great! Just some q's

MaxText/maxtext_utils.py
ssusie: Done.

MaxText/maxtext_utils.py
ssusie: You are right.

MaxText/maxtext_utils.py
ssusie: mesh.shape is an OrderedDict. E.g. 
```
(Pdb) mesh.shape
OrderedDict([('data', 1), ('fsdp', 1), ('fsdp_transpose', 1), ('sequence', 1), ('tensor', 4), ('autoregressive', 1)])
```

ssusie: Mesh is `jax.sharding.Mesh` which has a method shape which returns an OrderedDict

MaxText/max_utils.py
rwitten: `calculate_params_per_chip` is per pytree? That is for leaves I thought?

MaxText/maxtext_utils.py
rwitten: amazing, I had no idea.

MaxText/maxtext_utils.py
```diff
@@ -124,3 +125,41 @@ def calculate_tflops_prefill(num_model_parameters, prefill_length, config, log=T
...
+    params: params of the model state
+    mesh: mesh constructed from config
+    tolerance: float between 0.0 and 1.0 representing the allowed percentage of
+    non-sharded parameters.
+  Returns:
+    bool: True if the majority of parameters are sufficiently sharded
+  """
+  total_num_params = max_utils.calculate_num_params_from_pytree(params)
+  product_num_devices_for_weight_sharding  = 1
+  for axis in ['fsdp', 'fsdp_transpose','sequence', 'tensor']:
```
rwitten, line 148: nit: consistent spacing

MaxText/max_utils.py
rwitten: agreed that `calculate_total_params_per_chip` is a solid name

rwitten (changes requested):
little nits

MaxText/maxtext_utils.py
ssusie: Thanks! Looking forward to integrating pylint to maxtext!

MaxText/max_utils.py
rwitten, line 66: Why the argument to `calculate_params_per_chip` called `pytree`? Should be `calculate_params_per_chip_leaf` and the argument should be `arr` or `leaf` or something.

rwitten (approved):
One nit, please fix it.

MaxText/max_utils.py
ssusie: Done, thanks.

PR #557, opened 2024-03-26
Pin nvidia-cudnn-cu12==8.9.7.29

michelle-yooh:
cudnn 9 is not uploaded to pypi due to pypi restrictions


rwitten (changes requested):
Really this should be in GPU requirements not requirements right?

PR #564, opened 2024-03-28
Add docker support to maxtext base image

tonyjohnchen:
For AOT+Hybridsim integration, we need to pull Hybridsim docker image, so we need to install docker for maxtext base image (https://screenshot.googleplex.com/BD3SMwL57tP5cQY).

Tested AOT+Hybridsim e2e on a GKE node: https://screenshot.googleplex.com/5LQwBNBp4p6iyLq

This should be a combo PR of Maxtext and XPK:
https://github.com/google/maxtext/pull/564
https://github.com/google/xpk/pull/102


rwitten (changes requested):
I need to understand more here.

PR #569, opened 2024-03-29
Change docker upload schedule

khatwanimohit:
Airflow looks for MaxText docker image based on datetime and cannot find it since we upload them at 3:00 AM UTC. 
This effects users testing their new Airflow tests locally.

### DIFF
--- a/.github/workflows/UploadDockerImages.yml
+++ b/.github/workflows/UploadDockerImages.yml
@@ -19,8 +19,8 @@ name: Build Images
 
 on:
   schedule:
-    # Run the job daily at 7PM PST (3AM UTC)
-    - cron:  '0 3 * * *'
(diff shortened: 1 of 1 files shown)


gobbleturk (approved):

.github/workflows/UploadDockerImages.yml
```diff
@@ -19,8 +19,8 @@ name: Build Images
 
 on:
   schedule:
-    # Run the job daily at 7PM PST (3AM UTC)
-    - cron:  '0 3 * * *'
+    # Run the job daily at 4PM PST (12AM UTC)
```
rwitten, line 22: nit: 12AM UTC probably isn't always 4PM PST because of daylight savings time.

rwitten (approved):

PR #570, opened 2024-03-29
Add functionality to automatically upload logs to Vertex Tensorboard

SurbhiJainUSC:
- Add functionality to upload data in `config.tensorboard_dir` to Tensorboard in Vertex AI
- XPK users won't have to create a Tensorboard in Vertex AI. XPK will automatically handle that.
- For non-XPK cases, users can choose to manually create Tensorboard instance in Vertex AI cloud console or set the configuration to create instance in MaxText

Note: Uploader to upload logs to Vertex Tensorboard is only supported for `Tensorflow < 2.15.0`. Vertex AI team is working on fixing the issue to support the latest TF versions.


MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
 
 eval_interval: -1  # the specific number of train step between eval_step
 target_eval_loss: 0.  # early stop once reaching target eval_loss
+
+# Vertex AI Tensorboard Configurations
+# Set to True to create Vertex Tensorboard instance in GCP
+# Set to False if running via XPK as Vertex Tensorboard and Experiment will be created by XPK
```
rwitten, line 253: Can you explain in what settings you'll want the different values here? Do we always set this to true for local runs to create the Vertex Tensorboard? Or?

MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
...
+create_vertex_tensorboard: False
+# Project to create Vertex Tensorboard in if create_vertex_tensorboard=True
+# Project where Vertex Tensorboard to upload data exists if upload_data_to_tensorboard=True
+# Set to blank if running via XPK as XPK will set this as an environment variable
+vertex_tensorboard_project: ""
```
rwitten, line 258: Shouldn't this typically not be necessary -- it will inherit from the current project?

rwitten: (If not local?)

MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
...
+# Location to create Vertex Tensorboard in if create_vertex_tensorboard=True
+# Location where Vertex Tensorboard exists if upload_data_to_tensorboard=True
+# Set to blank if running via XPK as XPK will set this as an environment variable
+# Vertex AI supported locations: https://cloud.google.com/vertex-ai/docs/general/locations#available-regions
+vertex_tensorboard_location: ""
```
rwitten, line 263: Do you think we could call this vertex_tensorboard_region assuming the value would be a region?

MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
...
+# Name of Vertex Tensorboard instance to create if create_vertex_tensorboard=True
+# Name of Vertex Tensorboard to use to upload data if upload_data_to_tensorboard=True
+# Set to blank if running via XPK as XPK will set this as an environment variable 
+vertex_tensorboard_name: ""
```
rwitten, line 267: Shouldn't this be run_name regardless of XPK?

MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
...
+# Name of Vertex Experiment to create if create_vertex_tensorboard=True
+# Name of Vertex Experiment to use to upload data if upload_data_to_tensorboard=True
+# Set to blank if running via XPK as XPK will set this as an environment variable
+vertex_experiment_name: ""
```
rwitten, line 271: I don't understand the Vertex Experiment terminology (and realistically neither will users). We need to explain it! (Like when do multiple runs go into the same experiment vs. different experiments.)

MaxText/configs/base.yml
```diff
@@ -247,3 +247,28 @@ decode_sampling_temperature: 1.
...
+# Set to True to upload data in 'config.tensorboard_dir' like profiles/scalar metrics to Vertex Tensorboard
+# Set to False if running via XPK as XPK will set this as an environment variable
+upload_data_to_tensorboard: False
```
rwitten, line 274: Shouldn't this be `upload_tensorboard_data_to_vertex`?

MaxText/train.py
```diff
@@ -480,14 +482,28 @@ def main(argv: Sequence[str]) -> None:
   config = pyconfig.config
   validate_train_config(config)
   os.environ["TFDS_DATA_DIR"] = config.dataset_path
+
+  if config.create_vertex_tensorboard and jax.process_index()==0:
```
rwitten, line 486: Let's encapsulate this in a function for readability. Happy for that function to go in train.py or elsewhere.

MaxText/train.py
```diff
@@ -480,14 +482,28 @@ def main(argv: Sequence[str]) -> None:
...
       stack_trace_to_cloud = config.stack_trace_to_cloud,
       stack_trace_interval_seconds = config.stack_trace_interval_seconds))
   diagnostic_config = diagnostic_configuration.DiagnosticConfig(debug_config)
   with diagnostic.diagnose(diagnostic_config):
-    train_loop(config)
+    try:
+      train_loop(config)
+    finally:
+      if vertex_uploader_flag:
+        uploader.stop_upload_to_tensorboard()
```
rwitten, line 506: Shouldn't this be handle with a destructor on uploader?

README.md
```diff
@@ -219,3 +219,53 @@ MaxText supports training and inference of various open models. Follow user guid
 
 * [Gemma](https://ai.google.dev/gemma): a family of open-weights Large Language Model (LLM) by [Google DeepMind](https://deepmind.google/), based on Gemini research and technology. \
 You can run decode and finetuning using instructions mentioned [here](https://github.com/google/maxtext/blob/main/getting_started/Run_Gemma.md).
+
+## Automatically Upload Logs to Vertex Tensorboard
```
rwitten, line 223: Let's move this in a standalone file. I'm trying to simplify the README. We will link to the standalone file in the primary README.

https://github.com/google/maxtext/tree/main/getting_started -- place for various explainers including this one.

MaxText/configs/base.yml
rwitten: This comment and some others might be handled in the README! But then let's link to the README and try to get one source of truth.

README.md
```diff
@@ -219,3 +219,53 @@ MaxText supports training and inference of various open models. Follow user guid
...
+create_vertex_tensorboard: False
+vertex_tensorboard_project: "test-project"
+vertex_tensorboard_location: "us-central1"
+vertex_tensorboard_name: "test-tb-instance"
+vertex_experiment_name: "test-experiment"
+upload_data_to_tensorboard: True
+```
+The above configuration will upload logs in `config.tensorboard_dir` to a manually created Vertex Tensorboard: `test-tb-instance` and Experiment: `test-experiment` in `us-central1` in `test-project`.
+
+**Scenario 3: Create Vertex Tensorboard & Upload Logs in MaxText**
```
rwitten, line 258: Generally what is the use case for this scenario? (Is Vertex TB something we foresee the user setting up many times? Or is it a once per project kind of thing?)

rwitten (changes requested):
Lots of questions! But looks amazing!

MaxText/configs/base.yml
SurbhiJainUSC: Or  should we call it `upload_data_to_vertex_tensorboard`?

MaxText/train.py
SurbhiJainUSC: Do you mean using `__del__` in python? I believe it is not guaranteed that the object will be destroyed: https://docs.python.org/3/reference/datamodel.html#object.__del__. Therefore, I have explicitly added a method to destroy the thread. Also, this design was followed to align with the Vertex AI APIs: https://cloud.google.com/vertex-ai/docs/experiments/tensorboard-upload-existing-logs. 

What are your thoughts?

README.md
SurbhiJainUSC: Generally, users should set up Vertex TB once per project. But, one TB instance can only support around 10K Vertex Experiments. So, if for for some reason you have more than 10K Experiments attached to a Vertex TB, then you would have to create a new TB instance. This scenario takes into consideration this particular use case.

Scenario 2 is different from this one in the sense that you can also manually create the TB though Google Cloud console.

MaxText/configs/base.yml
SurbhiJainUSC: Sure, will update this.

MaxText/configs/base.yml
SurbhiJainUSC: Yes, added this to cater local development as well.

MaxText/configs/base.yml
SurbhiJainUSC: Hmm, let me simplify the configurations more. I think it causing a lot of confusion with multiple configs. Let me think about a good way to manage these.

README.md
SurbhiJainUSC: Done!

SurbhiJainUSC:
> Lots of questions! But looks amazing!

Thank you Rafi for your suggestions!
I have simplified the process of creating Vertex AI Tensorboard for testing MaxText on GCE. I have also updated the README to explain different scenarios to upload logs to Vertex Tensorboard. Let me know if this looks good.

MaxText/configs/base.yml
SurbhiJainUSC: Updated README with more details.

MaxText/configs/base.yml
SurbhiJainUSC: Added details about different scenarios in README.

MaxText/configs/base.yml
rwitten: But isn't local development in a project? Just trying to understand.

MaxText/train.py
rwitten: I don't understand. isn't the Python3 doc saying that in a teardown event we might not call your destructor? But in that event do you want to teardown this? (Recall in general we can segfault so there is no promise.)

What is this `stop_upload_to_tensorboard` doing exactly?

rwitten (changes requested):
The explicit teardown is still a bit weird but looks good otherwise!

MaxText/train.py
SurbhiJainUSC: Added a new `VertexTensorboardManager` class that will take care of destroying the thread when the program ends.

MaxText/configs/base.yml
rwitten: feature sounds great but now the comment is wrong right?

MaxText/train.py
```diff
@@ -166,6 +167,34 @@ def save_checkpoint(checkpoint_manager, step, state, dataset_type='c4', data_ite
         items=orbax.checkpoint.args.PyTreeSave(item=state)
         ))
 
+def configure_upload_to_vertex_tensorboard(config, vertex_tensorboard_manager):
```
rwitten, line 170: does this feel like it should be in vertex_tensorboard.py?

rwitten: just a feeling

README.md
```diff
@@ -219,3 +219,6 @@ MaxText supports training and inference of various open models. Follow user guid
 
 * [Gemma](https://ai.google.dev/gemma): a family of open-weights Large Language Model (LLM) by [Google DeepMind](https://deepmind.google/), based on Gemini research and technology. \
 You can run decode and finetuning using instructions mentioned [here](https://github.com/google/maxtext/blob/main/getting_started/Run_Gemma.md).
+
+## Automatically Upload Logs to Vertex Tensorboard
+MaxText supports automatic upload of logs collected in a directory to a Tensorboard instance in Vertex AI. Follow [user guide](https://github.com/google/maxtext/tree/main/getting_started/Use_Vertex_AI_Tensorboard.md) to know more.
```
rwitten, line 224: nit: should be `[user guide](getting_started/Use_Vertex_AI_Tensorboard.md)` or similar. (That way when folks make a change their thing updates. (I'm trying to move other absolute links to be relative links.)

rwitten (approved):
Approved (but there are some nits I want you to cleanup)

MaxText/configs/base.yml
SurbhiJainUSC: Updated the comment.

MaxText/train.py
SurbhiJainUSC: Moved to VertexTensorboardManager class.

PR #574, opened 2024-04-03
define default mode quantize_kvcache parameter

kocchop:
The default decoder layer does not have the `quantize_kvcache` parameter defined. It is causing the default decoder layer run to fail. This PR changes the `self.quantize_kvcache` to `cfg.quantize_kvcache`.


kocchop:
@rwitten @michelle-yooh just made a small correction. The default decoder layer run was failing. Let me know what do you think.

rwitten (approved):
Wow there are apparently no tests for this anymore on our side -- maybe we should deprecate it? Wild.

PR #578, opened 2024-04-08
separate tpu and gpu end-to-end scripts

NinaCai:
Add gpu llama_7b_test.sh 
Copy all end-to-end bash files to tpu folder. Will delete the scripts under end-to-end folder after XL ML also uses the scripts under either tpu/gpu folder.


yangyuwei (approved):

end_to_end/gpu/test_llama2_7b.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
rwitten, line 1: Is this file A3 specific or not? (Imagine we support other Clouds or other NV GPUs.)

end_to_end/tpu/eval_assert.py
```diff
@@ -0,0 +1,139 @@
+"""
```
rwitten, line 1: Let us stage this as a `move` that you do simultaneously in the two repos. The problem with this is that your'e creating duplication in this repo so while the repo is in the duplicated state it isn't clear how other MaxText devs should behave.

rwitten (changes requested):
Love the idea but let's talk about the execution.

end_to_end/tpu/eval_assert.py
NinaCai: The ultimate goal is to remove the scripts under `end-to-end` so there will only be two folders `tpu` and `gpu`. I am keeping the original scripts because XLML would break, I have another PR(https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/243/files) in XLML to read the new folders. After that PR is merged, I can delete the scripts under `end-to-end`.

end_to_end/gpu/test_llama2_7b.sh
NinaCai: > Is this file A3 specific or not? (Imagine we support other Clouds or other NV GPUs.)

I believe it works for A3 specific, I copied the training command from Michelle's PR (https://github.com/google/maxtext/pull/576). To make it work more generally, is it about tuning XLA params?

end_to_end/tpu/eval_assert.py
yangyuwei: Right. Nina and I discussed. Here is the migration plan:

We'd move all the test_*.sh files into maxtext/end_to_end/tpu/. To make sure we don't break existing XLML tests, we shall:

1. Have a PR to add test_*.sh files to maxtext/end_to_end/tpu/
2. Update XLML tests to read from the new path maxtext/end_to_end/tpu/
3. Have another PR to remove test_*.sh files from maxtext/end_to_end/

end_to_end/gpu/test_llama2_7b.sh
yangyuwei: Thanks Rafi for the great point. The same config didn't work for A3+ at a certain scale, it ran into OOM. Pavel suggested it might be related to the network. We will verify after the new network release. So we may or may not share the same config for A3 and A3+. For NV GPUs, the XLA flag values may be different. How about introducing sub directories under /end_to_end/gpu/? e.g. /end_to_end/gpu/a3/, /end_to_end/gpu/a3_plus/, ....

rwitten (approved):

PR #586, opened 2024-04-11
Fix subset of hosts dataloading for TPU v4

khatwanimohit:


MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -188,7 +186,7 @@ def make_mixed_train_iterator_and_tokenizer(config, mesh, add_bos, add_eos):
       print("Overwrite both add_bos and add_eos to False")
       return make_c4_mlperf_train_iterator_and_tokenizer(config, mesh, add_bos=False, add_eos=False, process_indices = process_indices)
   else:
-    return BadSyntheticDataIterator(config, mesh), None, get_tokenizer(config.tokenizer_path, add_bos, add_eos)
+    return BadSyntheticDataIterator(config, mesh).get_iterator(), None, get_tokenizer(config.tokenizer_path, add_bos, add_eos)
```
rwitten, line 189: `BadSyntheticDataIterator` shouldn't have a get_iterator? the names don't make sense. I think `BadSyntheticDataIterator` should return an iterator or if you must you could have a `BadSyntheticDataIteratorFactory` (but why?)

rwitten (changes requested):
I feel like part of this CR is missing?

MaxText/train.py
```diff
@@ -450,6 +451,13 @@ def train_loop(config, state=None):
 
     with jax.profiler.StepTraceAnnotation("train", step_num=step):
       example_batch = load_next_batch(data_iterator, example_batch, config)
+      if config.max_checkify:
```
rwitten, line 454: This really really shouldn't be in train.py -- can we put it in the step function?

We're trying to keep train.py lean and mean.

MaxText/input_pipeline/input_pipeline_interface.py
```diff
@@ -176,7 +176,7 @@ def get_process_loading_real_data(config, mesh):
 
 def make_mixed_train_iterator_and_tokenizer(config, mesh, add_bos, add_eos):
   process_indices = get_process_loading_real_data(config, mesh)
-  print(len(process_indices),"hosts out of",jax.process_count(),"are loading real data")
+  print(len(process_indices),"hosts out of",jax.process_count(),"are loading real data", flush=True)
```
rwitten, line 179: no flush, no print in maxtext mainline

rwitten (approved):
Giving approval but I have some important feedback

MaxText/train.py
```diff
@@ -310,6 +311,14 @@ def record_goodput(recorder, config, step=None, job_start=False, job_end=False):
     if step is not None:
       recorder.record_step_start_time(step)
 
+def check_example_batch(config, example_batch):
+  if config.max_checkify:
+    jittable_f = checkify.checkify(
+        lambda x: checkify.check(jnp.any(x > -1), "Batch contains bad synthetic data!")
```
rwitten, line 317: sorry I'm not following this -- won't there be bad synthetic data but it will be in positions that should be clipped out in MaxText downstream?

rwitten (changes requested):

MaxText/train.py
khatwanimohit: Fixed it

rwitten (approved):

PR #587, opened 2024-04-11
Unify WORKDIR to /deps

michelle-yooh:


rwitten (approved):
Please wait for In-Ho

chajath (approved):

PR #591, opened 2024-04-11
adding script to fix the style and adding modified/fixed files

ssusie:


.github/workflows/PyLintFixOnPush.yaml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint pyink
+    - name: Analysing the code with pylint in Maxtext
+      run: |
+         pylint MaxText/ --recursive=y \
```
khatwanimohit, line 28: Can we move these pylint runs under the cpu: section of `.github/workflows/UnitTests.yml`, that way we can avoid having a seperate yml file for this.

.github/workflows/UnitTests.yml
```diff
@@ -49,8 +49,10 @@ jobs:
         pytype --jobs auto --disable import-error MaxText/
     - name: Analysing the code with pylint
       run: |
-        pylint MaxText/
-
+        pylint MaxText/ pedagogical_examples/ \
+         --fail-under=9 \
```
khatwanimohit, line 53: Instead of using command line options here, please move the config variables inside pylintrc file. 
recursive can go under [MASTER] and check-quote-consistency can go under the [FORMAT] section

khatwanimohit (changes requested):

.github/workflows/PyLintFixOnPush.yaml
ssusie: I create this separate file because the UnitTest runs on 
```
on:
  pull_request:
  push:
    branches: [ "main" ]
```
and I want the pylint check to run on every push on every brnach. I think we don't want UnitTest.yaml to run on all pushes and all branches. Happy to change if that's the case

khatwanimohit: Then I think we can remove the cpu tests from that file and also have pytype run on all branches in this workflow.

ssusie: After offline discussion moved all cpu tests to run on each push to remove duplicate tests.

.github/workflows/UnitTests.yml
ssusie: Sounds good, done.

khatwanimohit (approved):
LGTM

rwitten (commented):
@gobbleturk can you own this one?

MaxText/generate_param_only_checkpoint.py
```diff
@@ -40,35 +40,56 @@
...
+  return modify training_state and train_state_annotations to be "unrolled".
+  Otherwise do nothing."""
   if not config.scan_layers or not config.force_unroll:
     return
 
-  training_state_layers = training_state.params['params']['decoder']['layers']
-  training_state_annotations_layers = training_state_annotations.params['params']['decoder']['layers']
+  training_state_layers = training_state.params["params"]["decoder"]["layers"]
+  training_state_annotations_layers = training_state_annotations.params[
+      "params"
```
gobbleturk, line 55: This is a surprising pylint choice to me, I guess due to max line cutoff

pylintrc
```diff
@@ -249,7 +254,10 @@ generated-members=
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=125
+max-line-length=80
```
gobbleturk, line 257: I guess this is fine, although a significant change. I am very appreciative that the provided script should wrap the lines for me. @rwitten do you have a preference here?

code_style.sh
```diff
@@ -0,0 +1,32 @@
...
+# Clean up Python codes using Pylint & Pyink
+# Googlers: please run `sudo apt install pipx; pipx install pylint --force; pipx install pyink==23.10.0` in advance
+
+set -e
+
+FOLDERS_TO_FORMAT=("MaxText" "pedagogical_examples")
+
+for folder in "${FOLDERS_TO_FORMAT[@]}"
+do
+  pyink "$folder" --pyink-indentation=2 --line-length=80
```
gobbleturk, line 24: I guess we need to duplicate the line-length here - its hard to pull the max-line-length from the pylintrc file? (Also other options)

gobbleturk:
> @gobbleturk can you own this one?

Reviewed and LGTM, but I think you should confirm the line length change from 125 -> 80

pylintrc
ssusie: Thanks for the quick review!

I chose line length=80 because that is the google standard. I can ping the reference offline. On the other hand PEP8 standard is 79 https://peps.python.org/pep-0008/#maximum-line-length

MaxText/generate_param_only_checkpoint.py
ssusie: I am surprised too. Let me fix this manually.

code_style.sh
ssusie: changed it. thanks.

ssusie:
Rafi wants 125 as the max line length and files here are fixed with 80. I am closing this PR in favor of https://github.com/google/maxtext/pull/596 .

PR #596, opened 2024-04-16
adding script to fix the style and adding modified/fixed files with l…

ssusie:
added code_stype.sh that fixes style issues with pyink. Moved pylint and pytype cpu tests to run on every push. Fixed style issues for the files in MaxText.


gobbleturk (approved):
Thanks Susie!

PR #597, opened 2024-04-16
Fix Microbenchmark Profiling Memory Issues

morgandu:
Below are changes in this PR

  - allow microbenchmark with prefill lengths
  - allow microbenchmark with loop iters
  - allow benchmark prefill / generate stage only
  - delete prefill_results to that limited the batch size
  - added / move pytree util funcs

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -264,3 +264,8 @@ vertex_tensorboard_region: ""
 
 # If set to True, MaxText will perform extra checks using jax.checkify. Note that this will effect performance. 
 max_checkify: False
+
+# Inference
+inference_microbenchmark_prefill_lengths: "64,128,256,512,1024"
+inference_microbenchmark_stages: "prefill,generate"
+inference_microbenchmark_loop_iters: 10
--- a/MaxText/inference_microbenchmark.py
+++ b/MaxText/inference_microbenchmark.py
@@ -21,24 +21,11 @@ import json
 import sys
 
 from jetstream.engine import token_utils
+
 import max_utils
 import maxengine
 import maxtext_utils
 import pyconfig
-
-
-def summarize_pytree_data(params, name="Params"):
-  """Generate basic metrics of a given Pytree."""
-  num_params, total_param_size, avg_param_size = max_utils.summarize_size_from_pytree(params)
-  num_params_in_billions = num_params / 1e9
-  total_param_size_in_gb = total_param_size / 1e9
-  print(
-      f"{name} stats: \n"
-      f"\tTotal number of params: {num_params_in_billions:.3f} billion \n"
-      f"\tTotal memory usage: {total_param_size_in_gb:.3f} GB \n"
-      f"\tAvg size: {avg_param_size:.3f} bytes\n"
-  )
-  return num_params, total_param_size, avg_param_size
 
 
 def prefill_benchmark_loop(config, engine, decode_state, params, tokens, true_length, iters, profile_name=""):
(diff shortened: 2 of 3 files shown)


MaxText/inference_microbenchmark.py
```diff
@@ -154,36 +148,45 @@ def write_results(results, filename=""):
...
+  if "AutoRegressive" in results:
+    print(f"SYSTEM_TIME_PER_DECODE_TOKEN_MS = {results['AutoRegressive']['ar_step_in_ms_per_seq']}")
 
 
 def main(config):
   engine = maxengine.MaxEngine(config)
   params = engine.load_params()
-  prefill_lengths = [64, 128, 256, 512, 1024]
-  benchmark_loop_iters = 10
+  prefill_lengths = [int(l) for l in config.inference_microbenchmark_prefill_lengths.split(",")]
```
patemotter, line 165: Does this work if you pass in a command line param like `--inference-microbenchmark-prefill-lengths="512,1024"` or something similar?

MaxText/inference_utils.py
patemotter, line 1: I don't think we need to make a file solely for inference_utils at this time. Especially since these functions are not unique to inference. I would add them to `max_utils.py` since they cover fairly generic usecases.

patemotter (commented):
Overall LGTM, but I don't think we need the inference_utils.py file.

morgandu:
> Overall LGTM, but I don't think we need the inference_utils.py file.

I am writing a batch inference, which need some common utility functions

MaxText/inference_microbenchmark.py
morgandu: Yeah, example commands:
to run a single prefill length: 
```
  inference_microbenchmark_prefill_lengths=1024
```
to run a single stage: 
```
  inference_microbenchmark_stages=generate
```

MaxText/inference_utils.py
morgandu: inference_utils.py is an existing file, but I can move these function to max_utils.py

MaxText/inference_microbenchmark.py
```diff
@@ -154,36 +148,45 @@ def write_results(results, filename=""):
...
+    for prefill_length in prefill_lengths:
+      tokens, true_length = token_utils.tokenize_and_pad(
+        text, vocab, is_bos=True, prefill_lengths=[prefill_length])
+      benchmark_results["Prefill"][prefill_length], decode_state = prefill_benchmark(
+        config, engine, params, decode_state, tokens, true_length,
+        iters=benchmark_loop_iters, num_model_params=num_model_params)
+
+  if "generate" in stages_to_benchmark:
+    benchmark_results["AutoRegressive"], decode_state = ar_benchmark(
+      config, engine, params, decode_state, iters=benchmark_loop_iters, cache_size=cache_size, model_size=model_size)
```
vipannalla, line 189: For running just generate benchmark, you still need to populate kv cache to produce proper perf numbers?

MaxText/configs/base.yml
```diff
@@ -261,3 +261,8 @@ vertex_tensorboard_project: ""
 # Region to create Vertex AI Tensorboard in for GCE, blank if running via XPK
 # Vertex AI supported regions: https://cloud.google.com/vertex-ai/docs/general/locations#available-regions
 vertex_tensorboard_region: ""
+
```
vipannalla, line 264: Not in this PR, but I see a need for separate inference specific config files in future -- both base.yml and model specific config.

rwitten (approved):
looks good

MaxText/inference_microbenchmark.py
morgandu: You will still need to initialize a decode_state for generate step calculation

rwitten (approved):
Please remember to squash your commits.

PR #602, opened 2024-04-18
Run `codespell` using pre-commit

khatwanimohit:


gobbleturk (approved):
More proffesional every day. Soon maybe even professional.

rwitten (approved):

maxtext_gpu_dependencies.Dockerfile
```diff
@@ -38,13 +38,12 @@ RUN mkdir -p /deps
 # Set the working directory in the container
 WORKDIR /deps
 
-# Copy necessary build files to docker container
-COPY setup.sh requirements.txt constraints_gpu.txt /deps/
```
chajath, line 42: This is done so that when we are making changes independent of dependencies, we won't do the installation step again, which can take >2m. Is there a specific concern to overrule it? @khatwanimohit

setup.sh
```diff
@@ -195,7 +195,8 @@ fi
 # Install dependencies from requirements.txt
 cd $run_name_folder_path && pip install --upgrade pip
 if [[ "$MODE" == "pinned" ]]; then
-    pip3 install -r requirements.txt -c constraints_gpu.txt
```
chajath, line 198: I think adding `-U` here goes against the idea of pinned build. Previously, the versions installed will be the function of the base image (which is pinned) and reqs and constraints. Now it could potentially be time-dependent if we happen to miss to capture the constraints in the constraints file. WDYT? @khatwanimohit

maxtext_gpu_dependencies.Dockerfile
khatwanimohit: setup.sh now includes running the `pre-commit install` command which installs the pre-commit hooks in the user's git repo. That command needs access to `.pre-commit-config.yaml` file

setup.sh
khatwanimohit: I understand, we can remove it.

PR #610, opened 2024-04-20
Allow inference microbenchmark to time prefill only

morgandu:
inference_microbenchmark.py
  - time prefill only
  - benchmark prefill and insert


rwitten (approved):
Giving approval but please get an approval from inference as well.

patemotter (approved):

PR #629, opened 2024-04-30
Fix test_tokenize unit test

khatwanimohit:


MaxText/tests/tokenizer_test.py
```diff
@@ -21,6 +21,7 @@
 import train_tokenizer
 import tokenizer
 import unittest
+import pytest
```
gobbleturk, line 24: Is this import necessary?

gobbleturk (approved):

PR #630, opened 2024-04-30
Add tests to GPU runner

michelle-yooh:


rwitten (approved):
But should we have a section that runs on both TPU and GPU? Duplication seems painful.

michelle-yooh:
> But should we have a section that runs on both TPU and GPU? Duplication seems painful.

Yes, I agree. But the environment for TPU and GPU are quite different; different images, different container resource options (e.g. container runtime), different arguments to the test scripts (e.g. attention). We could certainly unify those two but the logic would be a little more complicated.

PR #640, opened 2024-05-06
Eval the command string from XPK for GPU script

michelle-yooh:


end_to_end/test_generate_param_only_checkpoint.sh
```diff
@@ -112,7 +112,6 @@ load_parameters_path=${base_output_directory}/${decode_ckpt_run_id}/checkpoints/
 attention=dot_product ici_tensor_parallelism=${ici_tensor_parallelism} steps=50 \
 metrics_file=/tmp/${run_id}_metrics.txt async_checkpointing=false max_target_length=128 per_device_batch_size=1 \
 quantization=${quantization} \
-${model_params} \
```
michelle-yooh, line 115: The test fails with an error message: `jaxlib.xla_extension.XlaRuntimeError: INTERNAL: the requested functionality is not supported` on A3 nodes. 

Defaulting to base params somehow fixes the issue, and it doesn't harm the test.

end_to_end/gpu/a3/test_oss_gpu_tests.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+set -uex
+
+# This script reflects the list of GPU tests in .github/workflows/UnitTests.yml
```
rwitten, line 5: Should we be merging this? I don't understand. Don't we want to model our tests as one test per step? (That makes them more isolated and reproducible.)

end_to_end/test_generate_param_only_checkpoint.sh
rwitten: Did you cut a ticket, etc for this with HLO? The purpose of this exercise is to be squashing bugs...

rwitten (changes requested):

end_to_end/gpu/a3/test_oss_gpu_tests.sh
michelle-yooh: Yes, I agree. I first thought that it'd be too tedious and verbose to add these tests to XLML as separate tests since I would have to write a script for each test. But I think I found a more concise way to do it. Please take a look the new change and this PR https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/283

end_to_end/test_generate_param_only_checkpoint.sh
michelle-yooh: Yes, b/339473974

gpu_multi_process_run.sh
```diff
@@ -145,7 +145,7 @@ resolve_coordinator_ip
 set -e
 
 PIDS=()
-${COMMAND} &
```
rwitten, line 148: this seems right

end_to_end/gpu/test_decode_gpu.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# This script provides a convenient set of default configs used for testing decode.py on GPU
+
```
rwitten, line 4: How will `test_decode_gpu.sh` and `test_train_gpu.sh` be used? They seem kind of weird but I can see why they'd be useful as well!

gpu_multi_process_run.sh
michelle-yooh: Yes, this makes it possible to chain commands as usual when we pass command string to XPK

end_to_end/gpu/test_decode_gpu.sh
michelle-yooh: They make the test commands more concise by providing common default values for the testing.

They will be used like in [this line](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/283/files#diff-495d4687b5bff3f6b5a39a3b16c59222e8f638729e635569a8e7d8ea93d351f9R48)
I think they can be also used in our unit tests.

gobbleturk: Agreed the files seem kind of weird at first, but your explanation makes sense. If you want to re-use the same set of config options I guess the convenience is worth it. I like the name base_test_* but this is fine too

michelle-yooh: I also like the names you suggested. I renamed the scripts.

michelle-yooh: I actually figured out another solution, which is to have these base scripts as [a string template in XLML repo](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/283/files#diff-495d4687b5bff3f6b5a39a3b16c59222e8f638729e635569a8e7d8ea93d351f9R45). If it looks good, I will remove the scripts from this PR. The change in `gpu_multi_process_run.sh` is still required.

gobbleturk: Ya I like this template in XLML, defined as close as possible to where it is used

gobbleturk (approved):

rwitten (approved):

yangyuwei (approved):

PR #641, opened 2024-05-07
fix sharding for generate cache in prefill results

jwyang-google:


MaxText/maxengine.py
```diff
@@ -171,6 +171,19 @@ def prefill(
         flat_logits, (0, true_length - 1, 0), (flat_logits.shape[0], 1, flat_logits.shape[2])
     )
     selected_logits = jax.lax.with_sharding_constraint(selected_logits, self.replicated_sharding)
+
+    # No need to return generate cache: "cached_ar_key" and "cached_ar_value"
+    # Delete the "cached_ar_key" and "cached_ar_value" keys would cause error in insert() function
+    # Set cached_ar_key/cached_ar_value to dummy shape (1,) value to save memory
+    for i in range(self.config.num_decoder_layers):
```
rwitten, line 178: I don't think this production quality as a solution.

(1) it doesn't work on scan_layers=true models at all.
(2) it replicates the full structure of the path (which is brittle).

(Also the memory savings are pretty small right? Just halving the memory on the prefill -- which should just be a 2 or 3 examples?)

Happy to brainstorm to a better solution because agreed on the spirit.

rwitten (changes requested):

MaxText/maxengine.py
jwyang-google: Thanks Rafi for the review! The memory savings is large as:
1. We found that "cached_ar_key" and "cached_ar_value" is not sharded properly, it is replicated on every device, therefore it occupies much memory.
2. When running with Jetstream, Jetstream will reserve a queue of prefill results, which makes it even worse. 

ideally, i think we do not need to return generate cache as it is not used in the insert() function. But I didn't find a better way to remove these, thus making this hacky solution here. 

probably we can first make the sharding correct so that it will not occupy that much memory. But if you have better idea how we can resolve this, let me know!

morgandu: Just realized my previous comment is disappeared. 

Some extra context:

This is the whole prefill result: https://paste.googleplex.com/4930764978520064
The vars took bigger memory footprint are:
- cached_prefill_key, and cached_prefill_value, which are correctly sharded, because during key valuecalculation, there were sharding enforced, https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L542
- cached_ar_key, cached_ar_value are only initialized, and not used nor there is any sharding constraints applied on, which lead to unwanted number of chips of copies. https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L531

Two solutions based on above observations:
- comment out the initialization or ar, which I am not sure how it will affect other models (scan_layers, or training side) https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L531
 
- enforce sharding for generate cache like what https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L587
 
I think what @jwyang-google has in this PR has the similar goal with solution 1, can you @rwitten help us understand better the reason having the generate cache initialized in prefill_result?

If these vars is non removable on maxtext side, we may be able to seek some deletion on JetStream after getting the prefill result, though this is non ideal solution.

vipannalla: Agree on (2), this hardcodes to the internal model structure. However, we won't need scan_layers=true support for inference. Maxengine.py is only used by Jetstream and training shouldn't be invoking it?

As per Junwei/Morgan, we are able to double our batch_size by deleting the extra copy. Can you guys share some numbers?

jwyang-google: instead of removing the generate cache, fix the generate cache sharding which makes it occupy less memory.

MaxText/layers/attentions.py
```diff
@@ -528,7 +528,25 @@ def kv_cache_prefill(
     cached_prefill_key_var, cached_prefill_value_var, cached_prefill_segment_id = self._get_prefill_cache(
         batch, heads, kv_head_size, self.quantize_kvcache
     )
-    self._get_ar_cache(batch, heads, kv_head_size, self.quantize_kvcache)  # initialize it now
+    cached_ar_key_var, cached_ar_value_var, _, _ = self._get_ar_cache(batch, heads, kv_head_size, self.quantize_kvcache)  # initialize it now
```
rwitten, line 531: This looks good spiritually but why is it it necessary?

Shouldn't we be doing it here:
https://github.com/google/maxtext/blob/d590328e4fd74e128cf2678d51b848c148fdcc3e/MaxText/layers/attentions.py#L460-L473

(Have you confirmed this change fixes the problem? Could there be a bug in the cache initialization code.)

rwitten (changes requested):

MaxText/layers/attentions.py
jwyang-google: yes, this fixes sharding. Seems we need to call with_logical_constraint to enforce the sharding. Same thing as we did for the ar_cache here:
https://github.com/google/maxtext/blob/bdb1b3e1fca59b632ee5f43d6432d9ce8a15ded3/MaxText/layers/attentions.py#L585-L608

vipannalla: `[nn.with_logical_partitioning()](https://github.com/google/flax/blob/main/flax/linen/spmd.py#L331)` only adds partitioning metadata to the param, need to call `nn.with_logical_constraint()` to enforce the sharding.

vipannalla (commented):
LGTM

MaxText/layers/attentions.py
```diff
@@ -528,7 +528,25 @@ def kv_cache_prefill(
...
+    cached_ar_key_var[0].value = nn.with_logical_constraint(
```
rwitten, line 532: please put this into the `self._get_ar_cache` but looks ok otherwise as a short term fix.

rwitten (approved):
gave an approval but a tiny nit to make the mess smaller. Also can you put a todo with a link to an internal bug explaining the issue.

MaxText/layers/attentions.py
jwyang-google: done

PR #646, opened 2024-05-14
Support data input from HuggingFace

aireenmei:
An updated implementation of HF input pipeline, key designs:
Use HuggingFace API for loading dataset and tokenization. Then use grain for downstreaming processing to leverage grain's features such as packing, prefetch, use child process to fetch data.

Performance:
Standalone dataloader measured 2048 (seq_len) * 8 (per_device_batch) ~ 50s for 1000 batches. Increase host number from 1 to 64 does not change the speed.
No step time regression compared with TFDS/Grain, with small model (1B) on large slice (v5e-256). Step time the same (~1.008) for:
[HF pipeline](https://pantheon.corp.google.com/kubernetes/service/us-west4/v5e-256-bodaborg-us-west4/default/aireen-20240515-00-48/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod)
[TFDS](https://pantheon.corp.google.com/kubernetes/service/us-west4/v5e-256-bodaborg-us-west4/default/aireen-20240514-23-19/logs?e=13802955&mods=allow_workbench_image_override&project=tpu-prod-env-multipod)
[Grain](https://console.cloud.google.com/kubernetes/service/us-west4/v5e-256-bodaborg-us-west4/default/aireen-20240514-23-26/details?project=tpu-prod-env-multipod)

[Convergence test](https://console.cloud.google.com/kubernetes/service/us-west4/v5e-256-bodaborg-us-west4/default/aireen-20240511-20-01/details?project=tpu-prod-env-multipod) pass

Clarifications:
1. HuggingFace's tokenizer library is for generating/converting to Rust based tokenizer, and does not contain modules for using Rust based tokenizer. We still need to depend on transformers for using HuggingFace tokenizer
2. For unit test I use only one shard of the parquet dataset, because resolving the full dataset causes some overheads (~30s).


MaxText/configs/base.yml
```diff
@@ -168,7 +168,12 @@ per_device_batch_size: 12.0
 expansion_factor_real_data: -1 # if -1 then all hosts will load real data, else total_hosts//expansion_factor_real_data will pull data from GCS.
 eval_per_device_batch_size: 0
 max_corpus_chars: 10_000_000
-dataset_type: c4 # must be c4 or synthetic
+dataset_type: c4 # must be c4, hf or synthetic
+# for HuggingFace input pipeline (dataset_type=hf)
+hf_path: ''
```
gobbleturk, line 173: Could we add some hints here for what this should be set to in comments / maybe a pointer to the nice documentation you wrote in Data Input Pipeline

MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -0,0 +1,124 @@
...
+"""Input pipeline using Huggingface datasets."""
+
+import ml_collections
+import jax
+from datasets import load_dataset
+from transformers import AutoTokenizer
+import grain.python as grain
+
+from input_pipeline import _hf_operations
+from input_pipeline import _grain_operations
```
gobbleturk, line 26: Seems odd that hf depends on _grain_operations? Are these common operations that apply to both? If so could we move (or rename existing file) them to a file called something like "input_pipeline_utils" or "_common_operations".

This could be done in a followup PR

MaxText/input_pipeline/_hf_operations.py
gobbleturk, line 1: I'm not sure if any of these methods are hf specific or would also be better in a _common file

end_to_end/tpu/test_convergence_1b_params.sh
```diff
@@ -37,6 +37,12 @@ then
     CMD_DATA=" dataset_type=c4-array_record dataset_name=array-record/c4/en/3.0.1 eval_dataset_name=array-record/c4/en/3.0.1"
 fi
 
+if [ "$DATASET_TYPE" == "hf" ]
+then
+    gsutil cp -r gs://maxtext-dataset/hf/llama2-tokenizer assets
```
gobbleturk, line 42: Is there a difference between the hf llama2-tokenizer and the existing llama2-tokenizer? Is the HF tokenizer necessary for the HF dataset format? Generally different tokenizers can make a large difference in the trained model / loss profile, so this should be the same tokenizer as the other convergence test if possible.

Also we should avoid depending on bucket objects like this when possible. If this tokenizer is necessary could we instead add the hf/llama2-tokenizer into the assets of maxtext main (e.g. add the tokenizer in this PR?).

gobbleturk (changes requested):

getting_started/Data_Input_Pipeline.md
```diff
@@ -15,7 +15,32 @@
 -->
 ## Data Input Pipeline
 
-Currently MaxText supports two data input pipelines: the tfds (tensorflow_datasets) based pipeline as default, and the Grain pipeline for determinism. 
+Currently MaxText supports three types of data input for training: HuggingFace datasets, Tensorflow Datasets (TFRecord files) through the tf.data based pipeline, and ArrayRecord files through the Grain pipeline for determinism.
```
rwitten, line 18: Can you give a little more flavor? Something like

```
Currently MaxText supports three types of data input for training: HuggingFace datasets, Tensorflow Datasets (TFRecord files) through the tf.data based pipeline, and ArrayRecord files through the Grain pipeline.

HuggingFace datasets tend to be convenient because many popular datasets are available, but tend to be the least performant/scalable and lack deterministic data replay. TensorFlow datasets are highly performant/scalable but similarly lack deterministic data replay. Finally, Grain tends to be adequately performant/scalable and provides deterministic data replay but is the least convenient
```

P.S. -- even after writing this I feel like we should actually delete the paragraph I just wrote and make a table of the three datasets on the axes of Performance/Scalability, Deterministic Replayability and Convenience where:

Huggingface = Low Perf/Scalability, Not Deterministically Replayable and High Convenience
TFDS = High Perf/Scalability, Not Deterministically Replayable and Medium Convenience
Grain = Medium Perf/Scalability, Deterministically Replayable and Low Convenience

.github/workflows/UnitTests.yml
```diff
@@ -50,10 +50,14 @@ jobs:
...
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/deps --rm --privileged maxtext_base_image bash -c 'cd MaxText;python3 -m pytest'
-    - name: Test train.py with c4
+    - name: Test train.py with tf.data c4
       run: |
         docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/deps --rm --privileged maxtext_base_image bash -c \
         'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs dataset_path=gs://maxtext-dataset steps=2 enable_checkpointing=false'
+    - name: Test train.py with HF c4
+      run: |
+        docker run -v /home/runner/actions-runner/_work/maxtext/maxtext:/app --rm --privileged maxtext_base_image bash -c \
+        'python3 MaxText/train.py MaxText/configs/base.yml run_name=runner_$(date +%Y-%m-%d-%H-%M)-${RANDOM} base_output_directory=gs://runner-maxtext-logs hf_data_files=gs://maxtext-dataset/hf/c4/c4-train-00000-of-01637.parquet hf_path=parquet dataset_type=hf steps=2 tokenizer_path=google-t5/t5-large enable_checkpointing=false'
```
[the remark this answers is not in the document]
aireenmei: The concern here is that when giving a lot of files to HF's data loader, it will do a initialization to take a look at all the files. c4-train-*.parquet contains 1637 files, so it takes ~30s, which feels like an overhead for short unit tests, so I only use one file here.

MaxText/configs/base.yml
aireenmei: I added more comments and link to the doc

MaxText/input_pipeline/_hf_data_processing.py
aireenmei: They are now in _input_pipeline_utils.py

end_to_end/tpu/test_convergence_1b_params.sh
aireenmei: This tokenizer is from here: https://huggingface.co/meta-llama/Llama-2-7b-hf, which I believe is the same tokenizer as the existing llama2-tokenizer but in a different format (Rust based) that is more performant when used with HuggingFace dataset. Since it's from a gated model that requires user consent, I don't think we should add it to our repo. User can use it by providing "meta-llama/Llama-2-7b-hf" as tokenizer_path and set hf_access_token to theirs. I put this copy in our internal gcs for ease of testing so nobody's token is exposed. I added some comments to the file

getting_started/Data_Input_Pipeline.md
aireenmei: Thanks for the suggestion! I added a table. I do plan to do more improvements on the doc in future PRs

end_to_end/tpu/test_convergence_1b_params.sh
gobbleturk: Thank you for the explanation and the added comments!

getting_started/Data_Input_Pipeline.md
```diff
@@ -15,10 +15,48 @@
...
+```
+#### Limitations & Recommendations
+1. Streaming data directly from HuggingFace Hub may be impacted by the traffic of the server. During peak hours you may encounter "504 Server Error: Gateway Time-out". It's recommended to download the HuggingFace dataset to a GCS buckt or disk for the most stable experience.
+2. Streaming data directly from HuggingFace Hub works in multihost settings with a samll number of hosts. We have encountered "read time out" error with host number > 16.
+3. Multihost data input is achieved through HuggingFace's `datasets.distributed.split_dataset_by_node` API, which is more performant when (data shards/files) % (number of host reading data) == 0. To give MaxText users a performant experience, we follow this rule and will use data shards up to the number that's divisible by number of host reading data. For instance if your dataset has 100 shards, and you use 8 hosts, at most 96/100 shards of data will be used since 96%8==0. You can reshard or combine shards if you want to utilize the whole dataset.
+
+### Grain pipeline - for determinism
+
+<!-- TODO (aireenmei): add more details/examples on why determinism is important -->
```
gobbleturk, line 59: One example where deterministic replay is crucial is sensitive convergence experiments such as testing quantization techniques. When comparing a quantized vs unquantized run we want everything else to be identical, in particular the data in each batch. Such experiments may be long and require saving/resuming, and we want the data order to be replyable even if the paired runs save/resume at different steps. Another use case is to debug training spikes/anomalies - if we can replay the exact data it can help debug a bad data batch versus a hardware/SDC issue.

gobbleturk (approved):

getting_started/Data_Input_Pipeline.md
aireenmei: Thanks for the examples! I added a few more paragraphs there and incoporated your examples, with Gemimi's help :)

getting_started/Data_Input_Pipeline.md
```diff
@@ -15,12 +15,61 @@
...
+When a preempted training run resumes, the data input pipeline is re-initialized. If the same shuffle seed is used, the pipeline restarts from the beginning, potentially re-training the model on initial data. Conversely, a new seed produces a different batch sequence, making it difficult to track which data has been seen and how often each example is used for training. This lack of control can impact model performance and reproducibility.
+
+#### Cases where determinism is crucial
+* **Model sensitive to repetition.** When models are sensitive to the frequency with which they encounter specific examples, precise control over the order and repetition of data during training is essential.
+* **Convergence comparison.** In sensitive convergence experiments like testing quantization techniques, maintaining identical data batches between runs (e.g., quantized vs. unquantized) is essential for comparison. Determinism ensures consistency even the runs are long and undergo saving/resuming at different step.
```
gobbleturk, line 65: nit: consistency even **when** the runs are long

gobbleturk (approved):

gobbleturk (approved):

getting_started/Data_Input_Pipeline.md
```diff
@@ -15,12 +15,61 @@
...
+* **Debug training anomalies.** When troubleshooting training spikes or anomalies, the ability to replay the exact data sequence helps distinguish between bad data batches and underlying hardware or software issues.
+
+#### How does Grain achieve determinism
```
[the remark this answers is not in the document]
aireenmei: done

getting_started/Data_Input_Pipeline.md
aireenmei: Thanks for catching, fixed now.

getting_started/Data_Input_Pipeline.md
lhoestq: Hi ! I'm Quentin from HF :) let me know what performance aspects you'd like to see improved

Also regarding "deterministic data replay", I added it this week actually, see https://huggingface.co/docs/datasets/main/en/stream#save-a-dataset-checkpoint-and-resume-iteration

aireenmei: Hi Quentin, thanks for the pointer, this is fantastic! I'm also working on perf analysis, will get in touch with you, thanks!

PR #648, opened 2024-05-14
[Not for Merge] [POC] Goodput async monitoring and upload to Tensorboard POC

dipannita08:
This changes adds the following:
- Allows creating on a monitor object that spins up a secondary "monitor & upload" thread to query Goodput of the job using the ml-goodput-measurement pip package and and write a scalar metric to TB every interval period.

Tested:
 - [x] Example run on v4-8 w/ ~180 steps [here](https://screenshot.googleplex.com/82o5gNmW6ihb5Eq)

Note: This  is a POC and this change is intended to be moved to the cloud-accelerator-doagnostics and goodput package eventually.

### DIFF
--- a/MaxText/monitoring.py
+++ b/MaxText/monitoring.py
@@ -1,0 +1,108 @@
+"""Utilities to monitor and upload goodput data to TensorBoard asynchronously."""
+
+import datetime
+import logging
+import threading
+import time
+
+import jax
+from ml_goodput_measurement import goodput
+from tensorboardX import writer
+
+
+_exit_flag = threading.Event()
+_exit_flag.clear()
+_goodput_upload_thread = None
+logger = logging.getLogger(__name__)
+
+
+class GoodputMonitor:
+  """Queries and uploads goodput data to TensorBoard at a regular interval."""
+
+  def __init__(
+      self,
+      job_name: str,
+      logger_name: str,
+      tensorboard_dir: str,
+      upload_interval: int,
+  ):
+    """Initializes the GoodputMonitor.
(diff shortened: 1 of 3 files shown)


MaxText/monitoring.py
```diff
@@ -0,0 +1,108 @@
...
+  def _write_to_tensorboard(self, job_goodput: float, step):
+    if self._writer is not None:
+      timestamp = datetime.datetime.now().timestamp()
+      self._writer.add_scalar('goodput', job_goodput, step)
+      logger.info('Job goodput uploaded to Tensorboard.')
+      self._writer.flush()
+
+  def _query_and_upload_goodput(self):
+    """Queries and uploads goodput data to TensorBoard."""
+    time.sleep(10)
```
gobbleturk, line 65: Is this supposed to be goodput_upload_interval_seconds instead of hardcoded to 10? I don't see where goodput_upload_interval_seconds is used

gobbleturk (changes requested):

MaxText/monitoring.py
dipannita08: Ln65 can be removed, its just something I was using in the initial PoC. `goodput_upload_interval_seconds` (stored as self._upload_interval) is being used in Ln 68.

I apologize for the confusion, this PR is not meant to be merged it was just meant to be a PoC, I'll update the title - eventually this implementation will go into the Goodput package so that MaxText/anyone can import the module from the package and just instantiate the GoodputCalculator with `monitor_goodput=True`

PR #656, opened 2024-05-20
Pass metrics client config through to Jetstream

Bslabe123:


JoeZijunZhou (approved):

MaxText/configs/base.yml
```diff
@@ -281,6 +281,10 @@ inference_microbenchmark_stages: "prefill,generate"
 inference_microbenchmark_loop_iters: 10
 inference_microbenchmark_log_file_path: ""
 
+# Metrics
```
gobbleturk, line 284: There are other metric related fields already define here, could we change the comment header to Maxengine metrics?

gobbleturk (approved):

PR #666, opened 2024-05-22
Add flag for single controller workloads to avoid `jax.distributed.initialize()`

sadikneipp:
# Problem

Single controller workloads need to pass in `enable_checkpointing=true` and `async_checkpointing=true` to trigger async distributed checkpointing but will fail the `jax.distributed.initialize()` call.

# Change
Introduce a flag for single controller workloads which, if set to true, will block `jax.distributed.initialize()`


gobbleturk (approved):

sadikneipp:
#669

PR #680, opened 2024-06-03
refactor data input pipeline and add perf data

aireenmei:
This PR:
* change the flags related to data input to better align with the three existing data pipelines
* add checks & error messages for data input parameters
* add preliminary perf data of the 3 data input pipelines

### DIFF
--- a/MaxText/tests/hf_data_processing_test.py
+++ b/MaxText/tests/hf_data_processing_test.py
@@ -95,3 +95,18 @@ class HfDataProcessingTest(unittest.TestCase):
     self.assertTrue((batch1["targets_segmentation"] == batch2["targets_segmentation"]).all())
     self.assertTrue((batch1["inputs_position"] == batch2["inputs_position"]).all())
     self.assertTrue((batch1["targets_position"] == batch2["targets_position"]).all())
+
+  def test_for_loop_repeatable(self):
+    def get_first_batch(iterator):
+      batch = None
+      for batch in iterator:
+        break
+      return batch
+
+    train_batch1 = get_first_batch(self.train_iter)
+    train_batch2 = get_first_batch(self.train_iter)
+    self.assertTrue((train_batch1["inputs"] == train_batch2["inputs"]).all())
+    self.assertTrue((train_batch1["targets"] == train_batch2["targets"]).all())
+
+if __name__ == "__main__":
+  unittest.main()
--- a/getting_started/Data_Input_Perf.md
+++ b/getting_started/Data_Input_Perf.md
@@ -1,0 +1,35 @@
+## Performance of Data Input Pipeline
+* Overview of supported data input pipelines: https://github.com/google/maxtext/blob/main/getting_started/Data_Input_Pipeline.md
+* Perf data intepretation: for all three data pipelines, there are data prefetch running in parallel with computation. The goal is to hide data loading behind computation. As long as data loading step time < training computation step time, the data pipeline perf is considered sufficient. 
+
+### Methods
+* The following results are measured by [standalone_dataloader.py](https://github.com/google/maxtext/blob/main/MaxText/standalone_dataloader.py), which performs data loading without computation.
+* c4 data of different formats in GCS bucket are used. For Grain pipeline only, the GCS bucket is mounted to a local path via GCSFUSE ([script](https://github.com/google/maxtext/blob/main/setup_gcsfuse.sh))
+* The GCS bucket is multi-region (US) and the VMs that read data can be in different regions in the US.
+
+### HuggingFace pipeline
+The following data are collected using c4 data in Parquet format.
+| Pipeline    | seq_len | VM type    | per_host_batch    | # of host | # of batch | first step (s) | total time (s) |
+| ----------- | ------- | ---------- | ----------------- | --------- | ---------- | -------------  | -------------- |
+| HuggingFace | 2048    | TPU v4-8   | 32 (per_device=8) | 1         | 1000       | 6              | 72             |
+| HuggingFace | 2048    | TPU v4-128 | 32 (per_device=8) | 16        | 1000       | 6              | 72             |
+
+### Grain pipeline
+The following data are collected using c4 data in ArrayRecord format.
+| Pipeline    | seq_len | VM type    | per_host_batch    | # of host | # of batch | worker | first step (s) | total time (s) |
+| ----------- | ------- | ---------- | ----------------- | --------- | ---------- | -----  | -------------- | --------------- |
+| Grain       | 2048    | TPU v4-8   | 32 (per_device=8) | 1         | 1000       | 1      | 7              | 1200            |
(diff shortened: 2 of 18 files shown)


MaxText/tests/hf_data_processing_test.py
```diff
@@ -95,3 +95,18 @@ def test_batch_determinism(self):
     self.assertTrue((batch1["targets_segmentation"] == batch2["targets_segmentation"]).all())
     self.assertTrue((batch1["inputs_position"] == batch2["inputs_position"]).all())
     self.assertTrue((batch1["targets_position"] == batch2["targets_position"]).all())
+
+  def test_for_loop_repeatable(self):
```
gobbleturk, line 99: This method looks strange - I don't see it called anywhere and it looks like it either returns None or iterator[0] depending on if iterator is empty or not?

MaxText/tests/hf_data_processing_test.py
```diff
@@ -95,3 +95,18 @@ def test_batch_determinism(self):
...
+    def get_first_batch(iterator):
```
gobbleturk, line 100: I'm not sure I understand "get_first_batch" -- This just returns iterator[0]? What is the point of the for loop with a break / what is the "test_for_loop_repeatable" meant to test?

getting_started/Data_Input_Perf.md
```diff
@@ -0,0 +1,24 @@
...
+| Pipeline    | seq_len | per_host_batch | # of host | # of batch | worker | first step (s) | total time (s) |
+| ----------- | ------- | -------------- | --------- | ---------- | -----  | -------------- | -------------- |
+| Grain       | 2048    | 8              | 1         | 1000       | 1      | 7              | 479            |
+| Grain       | 2048    | 8              | 64        | 1000       | 1      | 6              | 354            |
+
+### TFDS pipeline
+The following data are collected using c4 data in TFRecord format.
+| Pipeline    | seq_len | per_host_batch | # of host | # of batch | first step (s) | total time (s) |
+| ----------- | ------- | -------------- | --------- | ---------- | -------------  | -------------- |
+| HuggingFace | 2048    | 8              | 1         | 1000       | 28             | 141            |
```
gobbleturk, line 23: Should the Pipeline here by TFDS instead of HuggingFace?

getting_started/Data_Input_Perf.md
```diff
@@ -0,0 +1,24 @@
+## Performance of Data Input Pipeline
+Overview of supported data input pipelines: https://github.com/google/maxtext/blob/main/getting_started/Data_Input_Pipeline.md
+
```
gobbleturk, line 3: Can you also include some information on the test setup? Mention that the datasets are in a GCS bucket (within the same region)? And the hosts are TPU VMs?

getting_started/Data_Input_Perf.md
```diff
@@ -0,0 +1,24 @@
...
+| HuggingFace | 2048    | 8              | 1         | 1000       | 6              | 52             |
+| HuggingFace | 2048    | 8              | 64        | 1000       | 5              | 49             |
+| HuggingFace | 2048    | 32             | 1         | 1000       | 7              | 191            |
+
+### Grain pipeline
+The following data are collected using c4 data in ArrayRecord format.
+| Pipeline    | seq_len | per_host_batch | # of host | # of batch | worker | first step (s) | total time (s) |
+| ----------- | ------- | -------------- | --------- | ---------- | -----  | -------------- | -------------- |
+| Grain       | 2048    | 8              | 1         | 1000       | 1      | 7              | 479            |
+| Grain       | 2048    | 8              | 64        | 1000       | 1      | 6              | 354            |
```
gobbleturk, line 17: Is this expected that more hosts is actually faster? This is pretty surprising. Are you averaging the results over many experiments?

getting_started/Data_Input_Perf.md
```diff
@@ -0,0 +1,24 @@
+## Performance of Data Input Pipeline
```
gobbleturk, line 1: Thank you for including this!!

gobbleturk (changes requested):
Looks great! Just a few minor requests

MaxText/tests/hf_data_processing_test.py
aireenmei: It will return the first batch of data. This test was added to [tfds/grain]_data_processing_test.py by this PR: https://github.com/google/maxtext/pull/363. My understanding is to make sure the for loop does not change the state of the iterator, so a new for loop always start with the same first batch. @ZhiyuLi-goog for comments

getting_started/Data_Input_Perf.md
aireenmei: Added test methods in the next paragraph

getting_started/Data_Input_Perf.md
aireenmei: Thanks for catching this. Fixed.

getting_started/Data_Input_Perf.md
aireenmei: I've consistently observed that grain performs better with more hosts. While tfds is the opposite. I guess it could be some caching mechanism at the GCS/GCSFUSE level that benefits grain? Need to dig deeper to understand it. I didn't average the results. I did run multiple times, the results are pretty consistent so I picked a representative one.

MaxText/tests/hf_data_processing_test.py
gobbleturk: I'm not sure the test is meaningful, would it fail for any kind of iterator? The same iterator is passed for train_batch1 and train_batch2, and iter[0] is returned both times?

gobbleturk (approved):

MaxText/tests/hf_data_processing_test.py
ZhiyuLi-goog: It is original used to verify eval iterator trigger in a constant interval during the training, ensuring it consistently produced the same order of data.
As for test, instead of verifying the full loop, the test focuses on validating the first batch of data.

```
class XXXIterator:

  # reset state when entering for loop
  def __iter__(self):
    self.reset()
    return self
```

ZhiyuLi-goog: > would it fail/succeed for any kind of iterator

The test would fail without manually reset, since we are wrap another abstraction layer out of `tf.data.Dataset` and `grain.DataLoader`, we should reset the state ourselves.

We can comment out [this line](https://github.com/google/maxtext/blob/c0ad11878065b4d8323512ad1ff6a3329d4571bc/MaxText/multihost_dataloading.py#L115) for testing.

ZhiyuLi-goog: Another example
https://screenshot.googleplex.com/8TJacdwAvsqkKFN

RoshaniN (approved):
Thank you, Aireen!
Changes from standalone dataloader and checkpointer perspective look good!

PR #687, opened 2024-06-05
Add gpt3 175b on v5e config

raymondzouu:
Adding GPT-3 175B training config on v5e.

### DIFF
--- a/MaxText/configs/v5e/gpt3_175b.sh
+++ b/MaxText/configs/v5e/gpt3_175b.sh
@@ -1,0 +1,48 @@
+# GPT-3 175B model.
+# This config will work out of the box for any number of v5e-256 slices.
+#
+# Command Flags:
+# OUTPUT_PATH (Required, unless base_output_directory is already set in base.yml)
+# DATASET_PATH (Required, unless dataset_path is already set in base.yml)
+# RUN_NAME (Required, unless run_name is already set in base.yml or running with XPK/GKE)
+# PLATFORM (Optional, can be "gke" or "gce", default is "gce")
+#
+# Example to invoke this script:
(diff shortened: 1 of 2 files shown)


MaxText/configs/v5e/gpt3_175b.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ -n "$RUN_NAME" ];
+then
+    export M_RUN_NAME=$RUN_NAME
+fi
+
+# Set up network
+bash preflight.sh PLATFORM=$PLATFORM
+
+# Train
+export LIBTPU_INIT_ARGS="--xla_tpu_enable_data_parallel_all_reduce_opt=true --xla_tpu_data_parallel_opt_different_sized_ops=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true --xla_tpu_overlap_compute_collective_tc=true --xla_enable_async_all_gather=true --xla_tpu_spmd_rng_bit_generator_unsafe=true"
```
ZhiyuLi-goog, line 41: Hi, @raymondzouu, `--xla_tpu_spmd_rng_bit_generator_unsafe=true` is necessary for int8 quantization only. We can remove it without int8 quantization. Then the `LIBTPU_INIT_ARGS` is the same as the other recipe in under v5e folder.

ZhiyuLi-goog:
Pretty cool! Thank you @raymondzouu ! LGTM except the nit.

ZhiyuLi-goog (approved):

MaxText/configs/v5e/gpt3_175b.sh
raymondzouu: I see, thanks for the heads up. I am planning to test with and without int8 quantization in the XLML test by setting the env variable `export M_QUANTIZATION=int8`. Can we still leave in this flag when running without int8 quantization?

ZhiyuLi-goog: Agreed!

gobbleturk (approved):

PR #688, opened 2024-06-05
Use run_id instead of sha for docker tag

jonb377:
There have been collisions between the scheduled UnitTests execution and those triggered by merge to main.

This change replaces the github SHA with the run_id to ensure a distinct image tag across executions.

Example scheduled execution failure: https://github.com/google/maxtext/actions/runs/9387491342/job/25850532082

Due to a concurrent execution from merge: https://github.com/google/maxtext/actions/runs/9387152549/job/25849389719

Context on run_id from https://docs.github.com/en/actions/learn-github-actions/contexts#github-context:

github.run_id | string | A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run.
-- | -- | --


gobbleturk (approved):
Thank you for finding this and explaining it!

PR #690, opened 2024-06-06
Reshape Q

morgandu:
Reshape Q to make the dot general convolution

### Checklist
- [x] Reshape Q
- [x] Attention Unit Tests
- [x] Microbenchmark - Performance
- [x] E2E Serving - Accuracy and Performance

##### Performance
- w-i8-kv-i8: Offline throughput improved **44.59%** when reshaping q comparing to the baseline, **7.94%**
 comparing to the original q on optimal layout. 
    - The improvement is from putting operation to convolution loop on MXU.
- w-bf16-kv-bf16: No improvement and regression 

##### Accuracy
- No regression in Rouge scores 

### Setup

- v5e4 for microbenchmark
- v5e8 for serving
- llama2-7b
- quant_mode: w-i8-kv-i8

##### Optimization_mode

| optimization_mode | optimal layout |
| -- | -- |
| baseline | 1203-1203 | 
| original_q | 0231-0231 |
| reshaping_q | 0231-0213 |

##### Accuracy

- model_mode: chat

| optimization mode | total input tokens | total output tokens | rouge 1  | rouge2 | rougeL  | rougeL sum | gen_len | gen_num |
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
| baseline | 217011.0 | 290128.0 | 41.9605 | 19.7131 | 26.7619 | 39.5253 | 1148690.0 | 995.0 |
| original_q | 217011.0 | 286422.0 | 42.1012 | 19.7373 | 26.8493 | 39.6566 | 1133161.0 | 995.0 |
| reshape_q | 217011.0 | 285745.0 | 42.1972 | 19.7037 | 26.7986 | 39.7534 | 1132540.0 | 995.0 |

##### Performance

| model mode | baseline thrpt | original_q thrpt | reshape_q thrpt | thrpt_delta from baseline | thrpt_delta from original_q |
| -- | -- | -- | -- | -- | -- |
| base | 2488.51 | 3333.38 | 3598.09 | 44.59% | 7.94% | 
| chat | 1646.43 | 2176.26 | 2256.12 | 43.16% | 3.67% |

##### original_q

https://xprof.corp.google.com/overview_page/morgandu-16956545554715004400

```
AutoRegressive results:
        AR step average time: 60.245 ms
        AR step average time per seq: 0.314 ms
        AR global batch size: 192
        AR throughput: 3186.960 tokens/second
        AR memory bandwidth per device: 231.851 GB/s
```

##### reshape_q

https://xprof.corp.google.com/overview_page/morgandu-3725675942258523013

```
AutoRegressive results:
        AR step average time: 47.868 ms
        AR step average time per seq: 0.249 ms
        AR global batch size: 192
        AR throughput: 4011.005 tokens/second
        AR memory bandwidth per device: 291.800 GB/s
```

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -113,6 +113,7 @@ class AttentionOp(nn.Module):
   prefill_value_axis_order: AxisIdxes = (1, 2, 0, 3)
   ar_key_axis_order: AxisIdxes = (1, 2, 0, 3)
   ar_value_axis_order: AxisIdxes = (1, 2, 0, 3)
+  reshape_q: bool = False
   dropout_rate: float = 0.0
   dtype: DType = jnp.float32
   quant: Optional[Quant] = None
@@ -277,7 +278,7 @@ class AttentionOp(nn.Module):
     )
     return dpa_layer(query, key, value, mask=attn_mask)
 
-  def compute_local_attention(self, attn_weights: Array, value: Array) -> tuple[Array, Array, Array]:
+  def compute_local_attention(self, attn_weights: Array, value: Array, q_seq_len: int) -> tuple[Array, Array, Array]:
     """Computes the attention of a local subset of the kv cache.
     Local attention results will need to be combined with any other local attentions and normalized
     Based on https://github.com/google-research/google-research/blob/master/scaling_transformer_inference_efficiency/attention.py
@@ -303,7 +304,12 @@ class AttentionOp(nn.Module):
     local_max = jnp.reshape(local_max, (local_max.shape[0], local_max.shape[1], local_max.shape[2] * local_max.shape[3], 1))
     local_sum = jnp.reshape(local_sum, (local_sum.shape[0], local_sum.shape[1], local_sum.shape[2] * local_sum.shape[3], 1))
 
-    local_out = self.wv_product(local_exps, value)
+    local_out = self.wv_product(local_exps, value, q_seq_len)
+
+    if self.reshape_q and q_seq_len == 1:
+      local_max = local_max[:,0:1,:,:]
+      local_sum = local_sum[:,0:1,:,:]
+
     return local_out, local_max, local_sum
 
   def apply_attention_dot(
@@ -320,7 +326,8 @@ class AttentionOp(nn.Module):
       query = query.astype(jnp.float32)
       key = key.astype(jnp.float32)
 
-    attn_weights = self.qk_product(query, key)
+    q_seq_len = query.shape[1]
+    attn_weights = self.qk_product(query, key, q_seq_len)
 
     # Casting softmaxt computation for float32 for model stability.
     if self.float32_logits:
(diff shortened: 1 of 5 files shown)


MaxText/layers/attentions.py
```diff
@@ -303,7 +304,12 @@ def compute_local_attention(self, attn_weights: Array, value: Array) -> tuple[Ar
     local_max = jnp.reshape(local_max, (local_max.shape[0], local_max.shape[1], local_max.shape[2] * local_max.shape[3], 1))
     local_sum = jnp.reshape(local_sum, (local_sum.shape[0], local_sum.shape[1], local_sum.shape[2] * local_sum.shape[3], 1))
 
-    local_out = self.wv_product(local_exps, value)
+    local_out = self.wv_product(local_exps, value, q_seq_len)
+
+    if self.reshape_q and q_seq_len == 1:
```
gobbleturk, line 309: Are there any cases where q_seq_len == 1 and we don't want to perform this reshape? If not we could remove the config option and only check for q_seq_len == 1?

gobbleturk (approved):
LGTM but do wait for the other reviewers. It looks to me like this is a strictly positive change for when q_seq_len==1, in which case we don't need to hide it behind an additional config option

MaxText/layers/attentions.py
morgandu: I haven not verified MQA/GQA so I am inclined to play safe here.

vipannalla: +1, not I'm curious how the broadcasted dimensions work if n_kv != n and t != s.

vipannalla (approved):
LGTM too

MaxText/layers/attentions.py
vipannalla: If your goal is to apply this to decode-phase only, should we use that condition instead of `q_seq_len == 1`?

morgandu: If the prefill phase has q_seq_len == 1 in dot_general, it will work for it too

MaxText/layers/attentions.py
```diff
@@ -353,10 +360,12 @@ def qk_product(self, query: Array, key: Array) -> Array:
     n_kv = key.shape[-2]
     assert n_kv == self.num_kv_heads
     query = jnp.reshape(query, (b, t, n_kv, n // n_kv, d))
+    if self.reshape_q and q_seq_len == 1:
+      query = jnp.broadcast_to(query, (b, 2, n_kv, n // n_kv, d))
```
zhihaoshan-google, line 364: Just out of my personal curiosity. It seems we are trying to broadcast the t axis (always be 1 for decode phase) to 2 to force the MXU to performance the matmul instead of on VPU.

Is this also doable by removing the t axis by `jnp.squeeze` which might be better than broadcast to save the computation and memory cost?

PR #701, opened 2024-06-12
Circular Pipelining 

gobbleturk:
Circular part of #677, followup to linear support in #691 

Copying the description from #677:

This implementation very closely follows the pax implementation [here](https://github.com/google/praxis/blob/69444e84acd0e543fb3a71ca015e55b72f30a80c/praxis/layers/pipeline.py#L4), collaborated and obtained permission from pax authors

This adds support for both circular and regular pipelining, but not loop fission. In addition this supports multiple layers per stage.

Currently there is no overlap between the main pipeline communication (activation forwarding between stages via collective permute) and the compute, this is a WIP with the compiler team. Over ICI this is not such a big hit, but over DCN we cannot obtain great performance until we develop a process to overlap. This implementation does roughly get the expected performance - e.g. the two major detrimental factors of bubble_size and exposed comms account for the realized step time.

Example runs over DCN (pipeline parallelism over DCN, fsdp within ICI):

4x v4-16

32B model, 32 layers = 4 stages * 2 layers per stage * 2 repeats. 4 microbatches
Model AI = 3 (since swiglu) * MLP(28k) * num_layers_per_stage (2) = 170k (approx)
This is actually pessimistic since ignoring attention
Hardware AI = (275 TFLOPs/s) / (1.6 GB/s DCN) also approx 170k
However this is not overlapped, so we expect steps take 50% time compute, 50% communication
Extra Bubble = (stages - 1) / (num_micro * num_repeat) = 3/(4*2) = 3/8
So we expect (1 + 3/8) extra steps * 2x longer each step = 2.75x longer
In reality we see only roughly 2x longer, MFU of ~30% [trace](https://xprof.corp.google.com/trace_viewer/mattdavidow-10770417440137691469?hosts=gke-tpu-b5c3bff2-zv6s)
Whereas a similar workload run on only 1 slice with pure FSDP the MFU is around 60% [trace](https://xprof.corp.google.com/overview_page/mattdavidow-4020802206111610241)
3x v4-128
"Cheated" the sizes to get best MFU by maxing out num layers and MLP,

emb=2x base, mlp = 5x base, layers = 9x base
144 layers = 3 stages * 6 layers per stage * 8 repeats. 3 microbatches
Model AI = 3 * ff_dim * num_layers_per_stage = 630k
Extra Bubble = (stages - 1) / (num_micro * num_repeat) = 2/(3 * 8) = 2/24
Expected extra step time= (1 + 2/24) * [(170k + 630k)/630k] = 1.37x
Actually see smaller extra step time, MFU of ~40% [XPROF](https://xprof.corp.google.com/overview_page/mattdavidow-4021953763511079633)
In addition to the correctness tests provided in this PR, I also ran a sanity check convergence test on a v4-128 with ici_pipeline_parallelism=4 and got the same results as the convergence test [XPK logs](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22tpu-prod-env-multipod%22%0Aresource.labels.location%3D%22us-central2%22%0Aresource.labels.cluster_name%3D%22v4-bodaborg%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Fjobset_sigs_k8s_io%2Fjobset-name%3D%22mattdavidow-pp-converge-a4%22%20severity%3E%3DDEFAULT%0Aresource.labels.pod_name%3D%22mattdavidow-pp-converge-a4-slice-job-0-11-g4b7h%22%0A%22completed%20step:%2010199,%22;cursorTimestamp=2024-05-25T21:57:11.392210071Z;startTime=2024-05-25T21:56:09.494Z;endTime=2024-05-26T00:56:09.494939Z?project=tpu-prod-env-multipod)


MaxText/layers/pipeline.py
```diff
@@ -113,12 +146,39 @@ def shard_dim_by_stages(self, x, dim: int):
...
+        have this dimension.
+      stages_dim_in_weights: The dimension in weights that represents parallel stages.
+    Returns:
+      The per-stage gathered values. The shape is weights.shape but with repeat_dim_in_weights
+        removed.
+    """
+    def _gather_one(x, repeat_id):
+      dim = repeat_dim_in_weights
+      if stages_dim_in_weights < dim:
+        dim -= 1
```
jonb377, line 173: Is this needed? It seems `repeat_dim_in_weights` will always be 0 and `stages_dim_in_weights` 1

MaxText/layers/pipeline.py
```diff
@@ -217,18 +298,46 @@ def run_one_iteration(self, loop_state, positions, segment_ids, deterministic, m
...
+    def prepare_vars_for_main_vmap(weights):
+      def gather_weights_for_stages_in(weights):
+        return jax.tree.map(
+            functools.partial(
+                self.vmap_parallel_gather, repeat_ids=repeat_ids, repeat_dim_in_weights=0, stages_dim_in_weights=1),
+            weights)
+      circular_metadata_params={
+        nn.PARTITION_NAME: "circular_repeats",
+        'sub_weight_split_dims_mapping': (None,),
+        "is_initializing": True,
```
jonb377, line 326: Should `is_initializing` always be True here?

MaxText/layers/pipeline.py
```diff
@@ -162,6 +226,21 @@ def _rotate_right(output_in):
...
 
+    if self.use_circ_storage:
+      # Insert the circ_storage_mover into new_circ_storage at a microbatch-rotating index.
+      # circ_storage_mover still points to the output of PREVIOUS iteration, which should aid in allowing overlapped compute/async transfers
+      def _rotate_right_and_update(circ_storage_mover_in, circ_storage_in):
+          rotated = _rotate_right(circ_storage_mover_in)
+          rotated = jnp.expand_dims(rotated, 1)
+          # The offset is the previous iterations microbatch ID of the last stage. 
+          offset = (loop_iteration - (self.num_stages - 1) - 1) % self.config.num_pipeline_microbatches # Note extra -1 b/c grabbing from the previous output - using circ_storage_mover before it is updated
+          return jax.lax.dynamic_update_slice_in_dim(circ_storage_in, rotated, offset, axis=1)
```
jonb377, line 237: Just checking my understanding: since the last stage is writing dummy data during the first `num_stages` iterations, this offset ensures that stage 0 reads real output when it starts reading from circ_storage.

jonb377 (approved):
Great stuff! LGTM

MaxText/layers/pipeline.py
```diff
@@ -73,22 +76,52 @@ def init_states(self, inputs):
     # We shard the microbatch_size axis by data/fsdp, not num_microbatches since those are looped over.
     state_io = nn.with_logical_constraint(state_io, ("activation_stage", None, "activation_batch", "activation_length", "activation_embed"),rules=self.config.logical_axis_rules, mesh=self.mesh) 
 
+    # circ_storage is used to hold the final pipeline stage outputs before it is used for the next repeat. It is only needed
+    # when num_microbatches > num_stages, else instead the final stage will immediately pass to the first without additional storage.
+    # circ_storage has shape [num_stages, microbatches, micro_size, sequence, embed].
+    if self.use_circ_storage:
+        circ_storage = jnp.zeros((self.num_stages,) + inputs.shape , dtype=inputs.dtype)
```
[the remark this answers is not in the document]
gobbleturk: Created b/347603101 to note this/list ideas and left a verbose comment describing this with the bug linked. Thanks for pointing this out!

MaxText/layers/pipeline.py
```diff
@@ -73,22 +76,52 @@ def init_states(self, inputs):
...
+        # Setup potential input from circ_storage, which also has a rotating index for microbatch, size of num_microbatches
+        circ_storage_batch_idx = loop_iteration % self.config.num_pipeline_microbatches
+        circular_stage_in = circ_storage[:,circ_storage_batch_idx]
+    else:
+        # The last stage immediately flows into the first stage, use this rotated shift instead of circular storage
+        circular_stage_in = shift
+ 
+    # For early loop iterations we grab a new input for stage 0 from the state_io. Once each microbatch has left state_io
+    # we instead grab from the last stage's output (possibly buffered when num_microbatches > num_stages, e.g. from circ_storage).
+    first_stage_in = jnp.where(loop_iteration < self.config.num_pipeline_microbatches, state_io_slice, circular_stage_in)
```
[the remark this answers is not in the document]
gobbleturk: This is an interesting idea, I think it may be complicated / require some re-sharding due to the `permute_output_micro_per_stage_dim`-like problems, but might be possible (haven't thought about it too much)

Its included in the same b/347603101 and in the comment above circ storage in reply to your earlier comment about circ_storage being larger than necessary

MaxText/layers/pipeline.py
gobbleturk: Good point, removed check

MaxText/layers/pipeline.py
gobbleturk: Hmm this is a good catch, replaced it with self.is_initializing(). Not entirely sure what behavior that changes other than denoting that the parameters can be modified

MaxText/layers/pipeline.py
gobbleturk: Thats basically right.

We will grab an input from circ_storage  on the num_microbatches index with index given by 
```
circ_storage_batch_idx = loop_iteration % self.config.num_pipeline_microbatches
```

When the first repeat happens (loop_iteration = num_microbatches), this should grab microbatch 0 (and then should grab microbatch 1 in the next loop_iteration, etc). In order to achieve this microbatch 0 should be in index 0 of the num_microbatches dimension of circ_storage. Microbatch 0 will enter the circular_storage when it is finished being processed by the last stage - which happens at iteration #(num_stages - 1), however actually it will enter 1 iteration later than that due to the iteration delay of circ_storage_mover.

PR #705, opened 2024-06-14
Add FSDP + Megablox

RissyRan:
# Description

3x perf improvement with FSDP sharding 

# Test

* Locally run: [link](https://paste.googleplex.com/6440880156442624)
* Internal benchmarks

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -49,6 +49,7 @@ bias_init = initializers.default_bias_init
 RMSNorm = normalizations.RMSNorm
 Quant = quantizations.AqtQuantization
 
+BATCH = "activation_batch"
 
 def _convert_to_activation_function(fn_or_string: Union[str, Callable[..., Any]]) -> Callable[..., Any]:
   """Convert a string to an activation function."""
@@ -250,7 +251,7 @@ class MlpBlock(nn.Module):
     x = nn.Dropout(rate=self.intermediate_dropout_rate, broadcast_dims=(-2,))(
         x, deterministic=deterministic
     )  # Broadcast along length.
-    x = nn.with_logical_constraint(x, ("activation_batch", "activation_length", "activation_mlp"))
+    x = nn.with_logical_constraint(x, (BATCH, "activation_length", "activation_mlp"))
     output = DenseGeneral(
         inputs.shape[-1],
         dtype=self.dtype,
@@ -289,14 +290,15 @@ class MoeBlock(nn.Module):
   dtype: DType = jnp.float32
 
   def generate_kernels(self, num_experts, emb_dim, mlp_dim):
-    
+
     kernel_in_axis = np.arange(1)
     kernel_out_axis = np.arange(1, 2)
     kernel_init = nd_dense_init(1.0, 'fan_in', 'truncated_normal')
 
-    kernel_axes = ('exp', 'embed', 'mlp')
-    wo_kernel_axes = ('exp', 'mlp', 'embed')
-    
+    # The first axes is expert
+    kernel_axes = (None, 'embed', 'mlp')
+    wo_kernel_axes = (None, 'mlp', 'embed')
+
     w0_kernel = self.param(
         'wi_0',
         nn.with_logical_partitioning(kernel_init, kernel_axes),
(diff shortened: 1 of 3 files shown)


MaxText/layers/linears.py
```diff
@@ -294,8 +296,8 @@ def generate_kernels(self, num_experts, emb_dim, mlp_dim):
     kernel_out_axis = np.arange(1, 2)
     kernel_init = nd_dense_init(1.0, 'fan_in', 'truncated_normal')
 
-    kernel_axes = ('exp', 'embed', 'mlp')
-    wo_kernel_axes = ('exp', 'mlp', 'embed')
+    kernel_axes = (None, 'embed', 'mlp')
```
gobbleturk, line 299: Nit: Add a comment here that the first axes corresponds to experts

MaxText/layers/linears.py
```diff
@@ -49,6 +50,7 @@
 RMSNorm = normalizations.RMSNorm
 Quant = quantizations.AqtQuantization
 
+MESH_FSDP_AXIS = "fsdp"
```
gobbleturk, line 53: I believe this should correspond to all batch dimensions, not just fsdp, e.g. should be the list [dp, fsdp, fsdp_transpose] give [here](https://github.com/google/maxtext/blob/f50f5ed2673d0c5bdf10ff06ed6a71d39a629080/MaxText/configs/base.yml#L125)

You should use use nn_logical_to_mesh((BATCH, None, None)) as done for flash attention
https://github.com/google/maxtext/blob/f50f5ed2673d0c5bdf10ff06ed6a71d39a629080/MaxText/layers/attentions.py#L198

Can you also leave a comment here that Model parallelism and sequence parallelism are not currently supported with this megablox kernel? Additionally we should assert this somehwere (I think in pyconfig) - if a user requests megablox with either TP or sequence parallelism we should error out saying this is currently not supported

gobbleturk (changes requested):
Great job figuring out this subtle issuing of wrapping these together under the same shard map. Mainly looks good, I would just like some changes regarding to how we specify shardings and assert that TP/sequence parallelism isn't used with megablox

MaxText/layers/linears.py
```diff
@@ -376,22 +369,42 @@ def gmm(inputs, kernel, group_sizes):
...
+        shard_map.shard_map,
+        mesh=self.mesh,
+        in_specs=(
+              PartitionSpec(MESH_FSDP_AXIS, None, None),
+              PartitionSpec(MESH_FSDP_AXIS, None, None),
+              PartitionSpec(None, None, None),
+              PartitionSpec(None, None, None),
+              PartitionSpec(None, None, None),
+        ),
+        out_specs=PartitionSpec(MESH_FSDP_AXIS, None, None),
```
gobbleturk, line 392: This should also be all batch dimensions (DP, FSDP, FSDP_TRANSPOSE), nothing special about the FSDP one I believe - you should be able to use the same nn.logical_to_physical API

MaxText/layers/linears.py
RissyRan: Sg! Will add that.

MaxText/layers/linears.py
RissyRan: Sg. Let me have a try and add assertion.

RissyRan: Updated. Thanks!

MaxText/layers/linears.py
RissyRan: Updated. Thanks!

MaxText/layers/linears.py
```diff
@@ -376,22 +367,43 @@ def gmm(inputs, kernel, group_sizes):
...
-    layer_act = _convert_to_activation_function(mlp_activation)(layer_w0)
-    intermediate_layer = jnp.multiply(layer_act, layer_w1)
-    output = gmm(intermediate_layer, wo_kernel, group_sizes)
-    return output
+    # Currently, we only support data parallelism with Megablox (sharding on batch dimensions)
+    @functools.partial(
+        shard_map.shard_map,
+        mesh=self.mesh,
+        in_specs=(
+              (nn.logical_to_mesh_axes(("activation_batch", None, None))),
```
gobbleturk, line 385: Nit: prefer using the common types https://github.com/google/maxtext/blob/fe4bfdcbff98877dfd611c8f802ebf94d6a9a06e/MaxText/common_types.py#L36 alias to avoid specifying strings (which could have silent typo errors)

gobbleturk (approved):

MaxText/layers/linears.py
RissyRan: sg, I will update that.

PR #706, opened 2024-06-14
Sharding the llama2 70b on v5e-16 more efficiently.

zhihaoshan-google:
https://arxiv.org/pdf/2211.05102
https://arxiv.org/pdf/1909.08053


MaxText/configs/v5e/inference/llama2_70b_v5e-16.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# WIP
+# tensor = 8, autoregressive=2
+# per_device_batch_size=6
+# weight bf16, kv cache bf16
+# around 29ms per decode step
+
+model_name: "llama2-70b"
+sharding_strategy: "experimental"
+attention: 'dot_product' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
```
morgandu, line 11: NIT: cudnn_flash_te does not work for inference, and tpu flash is broken on 70b, let's just leave supported attention as dot_product until verified.

MaxText/layers/attentions.py
```diff
@@ -759,9 +762,6 @@ def kv_cache_autoregressive(
     assert cached_ar_key_var[0].value.shape == self.cached_kv_shape((batch, self.max_target_length - self.max_prefill_predict_length, heads, kv_head_size), self.ar_key_axis_order)
     assert cached_ar_value_var[0].value.shape == self.cached_kv_shape((batch, self.max_target_length - self.max_prefill_predict_length, heads, kv_head_size), self.ar_value_axis_order)
 
-    key = nn.with_logical_constraint(key, (BATCH, LENGTH, HEAD, D_KV))
-    value = nn.with_logical_constraint(value, (BATCH, LENGTH, HEAD, D_KV))
-
```
morgandu, line 764: Does removing the logical constraint have any impact? I know in some cases, compiler is not able to propagate the sharding, thus hard coded nn.with_logical_constraint is in places in some cases.

MaxText/layers/llama2.py
```diff
@@ -80,6 +90,15 @@ def __call__(
 
     lnx = nn.with_logical_constraint(lnx, ("activation_batch", "activation_length", "activation_embed"))
 
+    if cfg.model_name == "llama2-70b" and cfg.sharding_strategy == "experimental":
+        query_axis_names: AxisNames = (KV_BATCH, LENGTH, KV_HEAD, KV_HEAD_DIM)
```
morgandu, line 94: It makes sense key and value have a different AxisNames and sharding than query, even though the different axis are length and head only; however, it's better not to mix them altogether, I'd say keep the query_axis_names as is, and change query's sharding rules to align with what works for 70B, instead of using kv rules on query.

MaxText/layers/attentions.py
zhihaoshan-google: There are already annotated in https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L1057-L1062.

MaxText/layers/llama2.py
zhihaoshan-google: Thanks, Morgan.

The main purpose here is to make the query activation sharded as the same as the key and value to decrease the cross chip communication. How about I add a comment here to explain? Otherwise, I need to crease a new axis name for query activation because the sharding of query activation is not the same as the query proj weight and currently they are sharing the same axis name.

MaxText/configs/v5e/inference/llama2_70b_v5e-16.yml
zhihaoshan-google: Thanks for this! Forget to check and remove the comment copied from base yaml.

zhihaoshan-google: done.

MaxText/layers/llama2.py
zhihaoshan-google: added comment.

morgandu: I'd prefer we do it properly even it's experimental config, we already have quite a confusion and assumption throughout the codebase and outdated comments. Let's not introduce any new ambiguity and potential issue for later.

MaxText/layers/attentions.py
```diff
@@ -961,11 +961,15 @@ def kv_projection(self, inputs_kv: Array, proj_name: str) -> Array:
     if self.num_query_heads % self.num_kv_heads != 0:
       raise ValueError("Invalid num_kv_heads for GQA.")
 
+    kernel_axes = ("embed", "heads", "kv")
+    if self.config.model_name == "llama2-70b" and self.config.sharding_strategy == "experimental":
```
gobbleturk, line 965: Can we use the more specific (embed, kv_heads, kv_head_dim) for all models, and instead its up to the config to decide how to shard this specific dim (e.g. add the default sharding in base.yml for kv_heads that makes this a no-op), but you can have wahtever you want in llama2-70b. I'd prefer this over the "experimental" or model specific if statements - ideally everything can be specified in the config and use the same generic code

MaxText/layers/llama2.py
gobbleturk: Agreed with morgan, its fine to add more axis names, which will allow us to be more specific/have more flexibility moving forward in addition to clearer code

MaxText/max_utils.py
gobbleturk, line 1: Prefer not to have random pylint fixes in untouched files

MaxText/layers/llama2.py
zhihaoshan-google: Could we do the sharding name refactoring in a separate task, as you said the sharding name is already confused and my cl doesn't introduce any degradation? Also if the query activation sharding is the same as the key value sharding, using the same axis name should be fine.

I added a TODO here.

getting_started/Data_Input_Pipeline.md
```diff
@@ -96,7 +96,7 @@ bash download_dataset.sh {GCS_PROJECT} {GCS_BUCKET_NAME}
 ```
 2. Use the following config:
 ```
-dataset_type: c4
+dataset_type: tfds
```
gobbleturk: I think this means you need to sync to main?

MaxText/layers/attentions.py
gobbleturk: Instead of removing them could you replace them with the more specific axis names?

MaxText/layers/llama2.py
gobbleturk: I would also like to avoid this model specific axis names - can we instead use the more specific axis names for all models, and its up the config to decide how to shard?

gobbleturk (changes requested):

MaxText/layers/attentions.py
zhihaoshan-google: If they are already annotated in the https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L1057-L1062, I don't think we need to annotate it again. (just like we don't annotate it in the kv_cache_prefill, I think we don't need to annotate it in the kv_cache_autoregressive as well)

MaxText/layers/llama2.py
zhihaoshan-google: I also agree with that we need to have a better sharding name for all models. But given the current timeline of the MLPerf, let's have a short term solution to unblock the optimization first and then apply a more general solution. I also added a TODO here.

MaxText/layers/attentions.py
zhihaoshan-google: Same as short term vs long term.

MaxText/max_utils.py
zhihaoshan-google: file was touched.

getting_started/Data_Input_Pipeline.md
zhihaoshan-google: dropped the tfds related commit. Will figure our how to sync.

MaxText/layers/attentions.py
morgandu: I think these two lines are safe to be deleted given https://github.com/google/maxtext/blob/main/MaxText/layers/attentions.py#L1057-L1062

MaxText/layers/attentions.py
vipannalla: +1, can you add  `kv_proj_kernel_axes` to config and specialize in llama-70-v5e-16 config? Adding model specific logic in the core attention logic is the the preferred way.

MaxText/layers/llama2.py
vipannalla: +1 to move these to config.

MaxText/max_utils.py
```diff
@@ -333,6 +336,10 @@ def create_device_mesh(config, devices=None):
   if multi_slice_env:
     dcn_parallelism = fill_unspecified_mesh_axes(dcn_parallelism, num_slices, "DCN")
     mesh = mesh_utils.create_hybrid_device_mesh(ici_parallelism, dcn_parallelism, devices)
+  elif config.model_name == "llama2-70b" and config.sharding_strategy == "experimental":
```
vipannalla, line 339: Same here, I thought this looks like llama2-70b specific the code is for multi-host inference. We should just add a `config.multi_host_inference` key.

vipannalla (commented):
Looks good, I've similar concerns as Morgan and Matthew regarding configs vs. model specific code.

MaxText/max_utils.py
zhihaoshan-google: Could we do it later? Even with `kv_proj_kernel_axes `, we still need to use `if` `else`, unless we make it general to all models which need further testing. 

Simply moving it to config doesn't work.  I only want to focus on the llama2 70B optimization in this PR.

vipannalla:
Talked off line, George is OOO soon and doesn't time right now. He can make these changes once he is back in 2 weeks. I'm ok with merging this as a short-term fix, will let @gobbleturk decide.

MaxText/max_utils.py
gobbleturk: Could we add a boolean "allow_split_physical_axes" to the base.yml config instead of using

```
elif config.model_name == "llama2-70b" and config.sharding_strategy == "experimental":
```

zhihaoshan-google: done.

gobbleturk (approved):
Thanks George!

zhihaoshan-google:
Thanks for the review, Matt, Vipan and Morgan!

PR #708, opened 2024-06-17
Allow Quantize KV Cache over Multiple Dimensions

morgandu:
# Checklist
This PR allows quantize kv cache over specified axis or axes
- [x] Microbenchmark - Performance
- [x] E2E Serving - Accuracy and Performance

# Setup

- v5e8
- llama2-7b
- quant_mode
    - for w-i8-kv-i8
- model_mode
    - base for performance
    - chat for accuracy

# Results and Analysis

### Annotation
- b: batch
- t: query_length
- h: query_heads
- d: kv_dimension
- s: kv_length
- k: kv_heads

### Layout
- 0123 | bthd | bskd
- 0213 | bhtd | bksd
- 0231 | bhdt | bkds
- 1203 | thbd | skbd

### KV Cache Quantization Axis/Axes
- d: existing quantization over kv_dimenison
- hd: new quantization over kv_heads and kv_dimension

## Summary

- Expected Acc drop, about 1.21% due to quantization axis change, from 42.4362 to 41.9234
- Perf improved 2x plus from 2488.511837 tokens/s to 5829.69 tokens/s due to composite factors including:
  - compute layout change, PR #709
  - cache layout tuning, PR #667
  - reshape_q, PR #690
  - kv cache quantization change, this PR

#### Exiting quantization over kv_dimenison, i.e. --kv_quant_axis=d

```
{'rouge1': 42.4362, 'rouge2': 20.1052, 'rougeL': 27.2529, 'rougeLsum': 39.9515, 'gen_len': 1133702, 'gen_num': 995}
```

#### New quantization over kv_heads and kv_dimenison, i.e. --kv_quant_axis=hd

```
{'rouge1': 41.9234, 'rouge2': 19.601, 'rougeL': 26.6612, 'rougeLsum': 39.5181, 'gen_len': 1156496, 'gen_num': 995}
```

###### Cache layout 0213-0213

With default cache layout, i.e. same layout as compute

```
Successful requests: 995
Benchmark duration: 179.016534 s
Total input tokens: 217011
Total generated tokens: 979107
Request throughput: 5.56 requests/s
Input token throughput: 1212.24 tokens/s
Output token throughput: 5469.37 tokens/s
```
xprof: https://xprof.corp.google.com/overview_page/morgandu-15535915910542734436

###### Cache layout 0231-0213

The cache layout tuning gave us top cache layout 0231-0213, where the default cache layout 0213-0213 was ranked the 2nd by ar throughut.

```
Successful requests: 995
[...]

PR #709, opened 2024-06-17
Allow Different Compute Layout for Attention

morgandu:
# Checklist
This PR introduced compute layout control to allowed a different compute layout for attention
- [x] Attention Unit Tests for different compute layout
- [x] Microbenchmark - Performance
- [x] E2E Serving - Accuracy and Performance

# Setup

- v5e8
- llama2-7b
- quant_mode
    - w-bf16-kv-bf16
    - for w-i8-kv-i8 see #708
- model_mode
    - base for performance
    - chat for accuracy

# Results and Analysis

The goal of introducing the new compute layout is to potentially avoid cache layout tuning, though we still can tune the cache layout to seek and verify for the best performance. 

### Annotation
- b: batch
- t: query_length
- h: query_heads
- d: kv_dimension
- s: kv_length
- k: kv_heads

### Layout
- 0123 | bthd | bskd
- 0213 | bhtd | bksd
- 1203 | thbd | skbd

## Summary

Existing attention compute layout is 0123, and we introduced a different compute layout 0213, which is of a layout that's TPU friendly. 

We introduced 0213 compute layout to verify:
    - if and how much 0213 has direct impact on performance on the default cache layout, i.e. same layout as compute layout
    - if and how much different compute layout and different cache layouts have a composite impact on performance

##### Cache layout 1203-1203
With the existing cache layout was 1203-1203, with throughput 2591.642232 tokens/s, this was improved from the default cache layout 0123-0123 about 3x. 

##### Cache layout 2013-2013
After layout tuning, we got optimal prefill-ar cache layout as 2013-2013, with throughput 3347.180221 tokens/s, which was 29% improvement.

##### Cache layout 0213-0213
With the two cache in the same layout as compute, i.e. 0213-0213 (xprof: https://xprof.corp.google.com/overview_page/morgandu-12159058496322304249), we got 3273.96 tokens/s, this is of the top performance after we verified with layout tuning.

##### Cache layout 0213-0132
The tuned cache layout that give us the best throughput 3329.45 tokens/s is 0213-0132 (xprof: https://xprof.corp.google.com/overview_page/morgandu-5743582688063478644)

### Accuracy

No regression on Rouge scores between 0123 and 0213

```
{'rouge1': 42.1738, 'rouge2': 19.6973, 'rougeL': 26.9088, 'rougeLsum': 39.6794, 'gen_len': 1144204, 'gen_num': 995}
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -31,6 +31,40 @@
                "dataset_path=gs://test-maxtext-dataset",
                "steps=2",
                "enable_checkpointing=false"]
-    }
+    },
+    {
+      "name": "Debug MaxText Inference Microbenchmark",
+      "type": "python",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "justMyCode": false,
+      "python": "python3",
+      "program": "${workspaceFolder}/MaxText/inference_microbenchmark.py",
+      "args": [
+        "MaxText/configs/base.yml",
+        "model_name=llama2-7b",
+        "tokenizer_path=assets/tokenizer.llama2",
+        "weight_dtype=bfloat16",
+        "scan_layers=false",
+        "attention=dot_product",
+        "max_prefill_predict_length=1024",
+        "max_target_length=2048",
+        "ici_fsdp_parallelism=1",
+        "ici_tensor_parallelism=-1",
+        "ici_autoregressive_parallelism=1",
+        "inference_microbenchmark_prefill_lengths=32,64,128,256,512,1024",
+        "inference_microbenchmark_stages=generate",
+        "inference_microbenchmark_loop_iters=1",
+        "run_name=runner_$(date +%Y-%m-%d-%H-%M)", 
+        "base_output_directory=gs://test-maxtext-output",
+        "prefill_cache_axis_order=0,2,1,3",
+        "ar_cache_axis_order=0,2,1,3",
+        "compute_axis_order=0,2,1,3",
+        "reshape_q=true",
+        "per_device_batch_size=24",
+        "quantization=int8",
+        "quantize_kvcache=True",
+      ]
+    },
   ]
 }
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -317,10 +317,14 @@ inference_microbenchmark_log_file_path: ""
 # KV Cache layout control
 # Logical layout: 0,1,2,3 ; CACHE_BATCH, CACHE_SEQUENCE, CACHE_HEADS, CACHE_KV
 # Default layout: 1,2,0,3 ; CACHE_SEQUENCE, CACHE_HEADS, CACHE_BATCH, CACHE_KV
-prefill_key_axis_order: "1,2,0,3"
-prefill_value_axis_order: "1,2,0,3"
-ar_key_axis_order: "1,2,0,3"
-ar_value_axis_order: "1,2,0,3"
+prefill_cache_axis_order: "1,2,0,3"
+ar_cache_axis_order: "1,2,0,3"
+
+# Compute layout control
+# Default layout: 0,1,2,3 ; BATCH, LENGTH, HEAD, D_KV
+# Currently only support compute layout: 0,1,2,3 and 0,2,1,3
+compute_axis_order: "0,1,2,3"
+
 reshape_q: False
 
 # Maxengine Metrics
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -71,6 +71,12 @@ dynamic_vector_slice_in_dim = jax.vmap(lax.dynamic_slice_in_dim, in_axes=(None, 
 # pytype: disable=attribute-error
 
 
+def validate_compute_axis_order(s: AxisIdxes) -> None:
+  valid_compute_axis_order = ((0,1,2,3), (0,2,1,3))
+  if s not in valid_compute_axis_order:  # currently supported compute_axis_order
+    raise ValueError("Invalid compute_axis_order was passed. Valid options ", valid_compute_axis_order)
+
+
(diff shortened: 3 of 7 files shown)


.vscode/launch.json
```diff
@@ -31,6 +31,40 @@
...
+        "base_output_directory=gs://test-maxtext-output",
+        "prefill_cache_axis_order=0,2,1,3",
+        "ar_cache_axis_order=0,2,1,3",
+        "compute_axis_order=0,2,1,3",
+        "reshape_q=true",
+        "per_device_batch_size=24",
+        "quantization=int8",
+        "quantize_kvcache=True",
+      ]
+    },
```
vipannalla, line 68: n00b question -- Is this to auto run for every local commit/amend on vscode? or just for convinience?

MaxText/pyconfig.py
```diff
@@ -52,6 +52,11 @@ def string_to_bool(s: str) -> bool:
 _yaml_types_to_parser = {str: str, int: int, float: float, bool: string_to_bool}
 
 
+def validate_compute_axis_order(s: str) -> None:
+  valid_compute_axis_order = ("0,1,2,3", "0,2,1,3")
```
vipannalla, line 56: Why not allow other layouts? Does the code break with others or just run slow? I'd allow other layouts similar to prefill_cache_axis_order/ar_cache_axis_order. 
Maybe remove the exception and just print a warning that others are untested?

MaxText/tests/attention_test.py
```diff
@@ -257,71 +258,73 @@ def tpu_kernel_attention_helper(self, num_kv_heads):
     )
 
   @pytest.mark.tpu
-  def test_dot_product_1203_1203(self):
-    self.dot_product_attention_helper(
-      prefill_cache_axis_order=(1,2,0,3),
-      ar_cache_axis_order=(1,2,0,3)
-    )
+  def test_dot_product_cache_axis_order(self):
```
vipannalla, line 261: Thanks for adding unitests!

MaxText/layers/attentions.py
```diff
@@ -68,6 +68,12 @@
 # pytype: disable=attribute-error
 
 
+def validate_compute_axis_order(s: str) -> None:
```
vipannalla, line 71: duplicate, remove. I see this method in pyconfig.py as well, which is right place to validate at the beginning of the run.

MaxText/pyconfig.py
vipannalla: nvm, I noticed in attention.py you specifically look for those two.

MaxText/layers/attentions.py
```diff
@@ -359,13 +366,21 @@ def qk_product(self, query: Array, key: Array, q_seq_len: int) -> Array:
...
+      if self.reshape_q and q_seq_len == 1:
+        query = jnp.broadcast_to(query, (b, 2, n_kv, n // n_kv, d))
+      result = jnp.einsum("btkgd,bskd->bkgts", query, key)
+    elif self.compute_axis_order == (0,2,1,3):
+      query = jnp.transpose(query, axes=self.compute_axis_order)
+      key = jnp.transpose(key, axes=self.compute_axis_order)
+      query = jnp.reshape(query, (b, n_kv, n // n_kv, t, d))
+      if self.reshape_q and q_seq_len == 1:
+        query = jnp.broadcast_to(query, (b, n_kv, n // n_kv, 2, d))
+      result = jnp.einsum("bkgtd,bksd->bkgts", query, key)
```
vipannalla, line 380: This lgtm for now. Since we already know 0 = b, 1 = t, 2= n etc.., can this code be made generic to support all layouts?

vipannalla (commented):
Looks good to me

.vscode/launch.json
morgandu: no, I added this when I was debugging on my local vscode, which is helpful for other engineers, but you may need to change the flags depends on your run.

MaxText/tests/attention_test.py
morgandu: of course!

MaxText/layers/attentions.py
morgandu: Actually this is a different validate. Since the base.yml only allows string, so the pyconfig validates the flag's string value. In attention module, the axis_orders are really used as tuple, and it can be hard code / overwrite in different model class initiation, i.e. https://github.com/google/maxtext/blob/main/MaxText/layers/gpt3.py#L229

I don't want to risk with possible hard coded axis_orders passing from somewhere other than yaml flags.

MaxText/layers/attentions.py
morgandu: A lot of other places needs to be changed to support more layouts, I have experienced some loud and silent bug when I was trying to add one more layout till I gave up.

Also, I don't think it's necessary to support all layouts. Since it was recommended that bh** is one of the friendly layout. I'd say if we end up really need to include new layouts then let's revisit it!

MaxText/configs/base.yml
```diff
@@ -312,10 +312,13 @@ inference_microbenchmark_log_file_path: ""
...
-prefill_key_axis_order: "1,2,0,3"
-prefill_value_axis_order: "1,2,0,3"
-ar_key_axis_order: "1,2,0,3"
-ar_value_axis_order: "1,2,0,3"
+prefill_cache_axis_order: "1,2,0,3"
+ar_cache_axis_order: "1,2,0,3"
+
+# Compute layout control
+# Default layout: 0,1,2,3 ; BATCH, LENGTH, HEAD, D_KV
+compute_axis_order: "0,1,2,3"
```
gobbleturk, line 320: Can we add a comment here that the only supported ones are (0,1,23) and (0,2,1,3)?

gobbleturk (approved):

PR #720, opened 2024-06-21
Enable saving using Orbax's emergency checkpoint manager

xuefgu:
This change incorporates Orbax's emergency checkpoint manager into MaxText. It's tested via XPK and locally saved checkpoint files are verified. The restore will be implemented in a follow-up change.

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -18,17 +18,24 @@ limitations under the License.
 
 from typing import Optional, Union
 from etils import epath
-
-from orbax.checkpoint.checkpoint_manager import CheckpointManager, CheckpointManagerOptions
+import orbax.checkpoint
 from orbax.checkpoint.logging import abstract_logger, cloud_logger, standard_logger, composite_logger
+from orbax.checkpoint import pytree_checkpoint_handler, type_handlers
+from orbax.checkpoint.checkpoint_manager import CheckpointManager, CheckpointManagerOptions, PyTree
+import orbax.checkpoint.experimental.emergency.checkpoint_manager as emergency_checkpoint_manager
 import jax
 import numpy as np
-import orbax.checkpoint
 import grain.python as grain
 
 import max_logging
 from multihost_dataloading import MultiHostDataLoadIterator
 from flax.training import train_state
+
+PyTreeCheckpointHandler = pytree_checkpoint_handler.PyTreeCheckpointHandler
+LocalCheckpointOptions = emergency_checkpoint_manager.LocalCheckpointOptions
+PersistentCheckpointOptions = (
+    emergency_checkpoint_manager.PersistentCheckpointOptions
+)
 
 
 def create_orbax_checkpoint_manager(
@@ -63,6 +70,53 @@ def create_orbax_checkpoint_manager(
   )
   max_logging.log("Checkpoint manager created!")
   return mngr
+
+
+def create_orbax_emergency_checkpoint_manager(
+    local_checkpoint_dir: str,
+    persistent_checkpoint_dir: str,
+    global_mesh: jax.sharding.Mesh,
+    abstract_state: PyTree,
+    local_save_interval_steps: int,
+    persistent_save_interval_steps: int,
+):
+  """Returns an emergency checkpoint."""
+  max_logging.log("Creating emergency checkpoint manager...")
+
+  local_registry = type_handlers.create_type_handler_registry(
+      (
+          jax.Array,
+          type_handlers.ArrayHandler(primary_host=None, replica_id=None),
+      ),
+  )
+
+  local_checkpoint_handler = PyTreeCheckpointHandler(
+      use_ocdbt=True,
+      use_zarr3=True,
+      primary_host=None,
+      type_handler_registry=local_registry,
+  )
+
+  options = emergency_checkpoint_manager.CheckpointManagerOptions(
+      local=LocalCheckpointOptions(
+          save_interval_steps=local_save_interval_steps
+      ),
+      persistent=PersistentCheckpointOptions(
+          save_interval_steps=persistent_save_interval_steps
+      ),
+  )
+
+  emergency_mngr = emergency_checkpoint_manager.CheckpointManager(
+      local_checkpoint_dir,
+      epath.Path(persistent_checkpoint_dir),
+      global_mesh=global_mesh,
+      abstract_state=abstract_state,
+      options=options,
+      local_state_handler=local_checkpoint_handler,
+  )
+
+  max_logging.log("Emergency checkpoint manager created!")
+  return emergency_mngr
 
 
 def _find_idx(array: np.ndarray, replica_axis_idx: int):
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -128,6 +128,18 @@ record_internal_nn_metrics: 0
 # Create a GCS bucket, e.g. my-maxtext-outputs and set this to "gs://my-maxtext-outputs/"
 base_output_directory: ""
 
+# Whether or not to enable emergency checkpoint. If True, `local_checkpoint_directory` and a non-zero `local_checkpoint_period` must also be specified.
+# Emergency checkpoint is an experimental Orbax feature that: periodically saves to persistent storage and, with a larger invertal, saves to a local directory.
+# During restore, if a local copy is available in any slice, it will be broadcast to other slices without having to fetch from persistent storage.
+# See more details on https://github.com/google/orbax/tree/main/checkpoint/orbax/checkpoint/experimental/emergency.
+enable_emergency_checkpoint: False
+
+# It should be specified when and only when `enable_emergency_checkpoint` is True.
+local_checkpoint_directory: ""
+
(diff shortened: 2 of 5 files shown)


MaxText/configs/base.yml
```diff
@@ -121,6 +121,15 @@ record_internal_nn_metrics: 0
 # Create a GCS bucket, e.g. my-maxtext-outputs and set this to "gs://my-maxtext-outputs/"
 base_output_directory: ""
 
+# Whether or not to enable emergency checkpoint. If True, `enable_emergency_checkpoint` must also be specified.
```
gobbleturk, line 124: I think what is meant is If True, "local_checkpoint_directory" must also be specified

MaxText/checkpointing.py
```diff
@@ -64,6 +72,53 @@ def create_orbax_checkpoint_manager(
   return mngr
 
 
+def create_orbax_emergency_checkpoint_manager(
```
gobbleturk, line 75: Is It necessary to have two separate creations for emergency and non-emergency? E.g. we can't just add the emergency checkpoint options (when requested) to the existing manager?

MaxText/configs/base.yml
```diff
@@ -121,6 +121,15 @@ record_internal_nn_metrics: 0
...
+enable_emergency_checkpoint: False
+
+# It should be speficied when and only when `enable_emergency_checkpoint` is True.
+local_checkpoint_directory: ""
+
+# It should be non-zero when and only when `enable_emergency_checkpoint` is True.
+local_checkpoint_period: 0
```
gobbleturk, line 131: Can we explain what an emergency checkpoint is (maybe in the top comment for enable_emergency_checkpoint)

I am surprised to see a field called "local_checkpoint_period" for emergency checkpointing - I thought the emergency checkpoint would only be saved once. Is this local_checkpoint_period additionally saving it every `period` steps because we are also removing the regular checkpoint manager?

MaxText/train.py
```diff
@@ -154,6 +155,11 @@ def write_metrics_to_tensorboard(writer, metrics, step, config):
 
 def save_checkpoint(checkpoint_manager, step, state, dataset_type="c4", data_iterator=None):
   """Wrapper for saving checkpoint"""
+  if isinstance(checkpoint_manager, emergency_checkpoint_manager.CheckpointManager):
```
gobbleturk, line 158: Can we not use emergency checkpointing with regular checkpointing? It looks to me like the emergency checkpoint manager is missing some of the key features of the regular one - (persistenet storage/GCS for the `period` saves, async option), but am still not sure what an emergency checkpoint is

MaxText/checkpointing.py
xuefgu: The creation signatures of the emergency vs. regular checkpoint managers are very different, as evidenced by the parameter list. I am trying to keep the logic different instead of mixing the two.

MaxText/configs/base.yml
xuefgu: Will add more comments. You are right, "emergency checkpointing" is just how Orbax calls the "local checkpointing" feature. We could make it for emergency only, but right now it saves to local disk or RAM Disk periodically.

We are not removing the regular checkpoint manager - it will be part of the ECM and still save to GCS periodically.

MaxText/checkpointing.py
```diff
@@ -64,6 +72,53 @@ def create_orbax_checkpoint_manager(
...
+      primary_host=None,
+      type_handler_registry=local_registry,
+  )
+
+  options = emergency_checkpoint_manager.CheckpointManagerOptions(
+      local=LocalCheckpointOptions(
+          save_interval_steps=local_save_interval_steps
+      ),
+      persistent=PersistentCheckpointOptions(
+          save_interval_steps=persistent_save_interval_steps
```
gobbleturk, line 105: should we have 
`enable_async_checkpointing=use_async`
and
`create=True` just like the regular checkpoint manager?

https://github.com/google/maxtext/blob/9df576142bba46ea539c6ff6fe63e743d7b7c12c/MaxText/checkpointing.py#L55C29-L56C1

MaxText/train.py
xuefgu: The saving is by default async. And as explained above, ECM has the regular checkpoint manager as one of its components.

MaxText/checkpointing.py
xuefgu: Regarding `enable_async_checkpointing=use_async`, for any workload large enough to need this feature, they wouldn't want to use sync checkpointing, so we are making the decision for them and simplifying their experience.

On `create=True`, great question! Due to some Orbax limitation, the persistent directory has to be created by the user first.

MaxText/train.py
xuefgu: Also, ECM.save does not support composite args for now, so treating the case separately upfront.

MaxText/configs/base.yml
xuefgu: Correct, also added more explanation of the feature.

cpgaffney1: Update comment?

MaxText/configs/base.yml
```diff
@@ -121,6 +121,15 @@ record_internal_nn_metrics: 0
 # Create a GCS bucket, e.g. my-maxtext-outputs and set this to "gs://my-maxtext-outputs/"
 base_output_directory: ""
 
+# Whether or not to enable emergency checkpoint. If True, `enable_emergency_checkpoint` must also be specified.
+enable_emergency_checkpoint: False
+
+# It should be speficied when and only when `enable_emergency_checkpoint` is True.
```
cpgaffney1, line 130: nit typo

MaxText/configs/base.yml
cpgaffney1, line 134: I think it's better if options like this default to None to indicate disabled, but up to you

MaxText/max_utils.py
```diff
@@ -216,6 +217,7 @@ def maybe_initialize_jax_distributed_system(raw_keys):
   ) or raw_keys["hardware"] == "gpu_multiprocess":
     max_logging.log("Attempting to initialize the jax distributed system...")
     jax.distributed.initialize()
+    ocp.multihost.utils.initialize_runtime_to_distributed_ids()
```
cpgaffney1, line 220: Don't you also need https://github.com/google/orbax/blob/main/checkpoint/orbax/checkpoint/multihost/utils.py#L37 flag set? I don't see that anywhere.

MaxText/checkpointing.py
```diff
@@ -64,6 +72,53 @@ def create_orbax_checkpoint_manager(
...
+      ),
+  )
+
+  emergency_mngr = emergency_checkpoint_manager.CheckpointManager(
+      local_checkpoint_dir,
+      epath.Path(persistent_checkpoint_dir),
+      global_mesh=global_mesh,
```
cpgaffney1, line 112: Probably a good idea to include some assertions about the mesh configuration passed to CheckpointManager - make sure the 0th mesh_axes represents the slice.

xuefgu: I would let that error out in Orbax, especially when we are considering supporting non-0th axis for `data`.

MaxText/configs/base.yml
xuefgu: Comment from an outdated file. Resolving.

MaxText/configs/base.yml
xuefgu: Done.

MaxText/max_utils.py
xuefgu: It is not needed for saving. Even for restoring, I am not sure it's required since the runtime-to-distributed id mapping is saved, and initialize_runtime_to_distributed_ids() doesn't depend on that flag. In fact, my testing without that flag didn't reveal any issues.

MaxText/configs/base.yml
xuefgu: I tried but prefer it this way to avoid unnecessarily convoluted arg parsing.

MaxText/max_utils.py
xuefgu: I submitted https://github.com/google/maxtext/pull/724 to use the latest Orbax (0.5.20) and JAX (0.4.30) and made sure that these versions were indeed being used in the actual run.

As expected, without setting the flag, no logs in https://github.com/google/orbax/blob/4cae60f87b1abfe09bfbdfea03718d4f93d5b755/checkpoint/orbax/checkpoint/multihost/utils.py#L251 are present, so `jax.process_index()` is used as multihost `process_index()`.

Still, saving never fails in 10+ tries. At this point, I suggest that we revisit the Orbax code carefully and understand this behavior before moving forward. With the many hacks, it's also a good opportunity for us to have a careful review of the e2e code path.

xuefgu: To summarize the internal investigation: setting this flag is indeed not necessary for saving. Will see its relevance in the restore piece.

cpgaffney1 (commented):
Thanks for addressing earlier comments, LGTM pending addition of the experimental_orbax_use_distributed_process_id flag, which we don't necessarily need for saving, but maybe need for restoring.

gobbleturk (approved):

PR #721, opened 2024-06-22
Use dedicated requirements for GPUs

xuefgu:
This change is necessary to separate GPU dependencies from the one-size-fits-all requirements.


requirements.txt
gobbleturk, line 1: We can discuss this more but often when we update requirements.txt its because the MaxText code now has real new dependencies.

E.g. in your PR I think these import statements will fail for old versions of jax/orbax https://screenshot.googleplex.com/4mQmQzK77RijUbF - thus the GPU pin needs to be moved up as well to work with maxtead code

xuefgu: I see two options:

1. Maintain one requirements.txt and update `constraints_gpu.txt` whenever there is a change in requirements.
2. Maintaine two requirements.txt and allow them to diverge.

I am open to #1. In fact, I tried it but found way too many changes required (> half of the packages). Is there a way to ensure that those dependency changes are not drastic enough to break anything?

For #2, you are right - that particular import will break for earlier versions for Orbax but there are no other imports that could break with older versions. We could:
1) Keep the Orbax versions the same (e.g. `orbax-checkpoint==0.5.20`) in both `requirements_gpu.txt` and `constraints_gpu.txt` but don't change anything related to JAX or jaxlib. Use newer versions of JAX in the original `requirements.txt`. OR:
2) Move the import into the conditional block when emergency checkpoint is being used, and explicitly say that it doesn't work on GPUs at least for now.

#2 does feel more convoluted and error-prone for maintenance. So it comes back to the question of how to ensure that a drastically changed `constraints_gpu.txt` due to new JAX and Orbax doesn't break anything?

PR #727, opened 2024-06-25
Prefill return first token

jwyang-google:
Modify Jetstream to make prefill return first token. Need to check in together with https://github.com/google/JetStream/pull/105

### DIFF
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -49,6 +49,7 @@ class DecodeState:
   generate_cache: jax.Array
   generate_cache_index: int
   generate_lengths: jax.Array
+  generated_token: jax.Array
 
 
 class MaxEngine(engine_api.Engine):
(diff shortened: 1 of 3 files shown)


JoeZijunZhou (approved):
LGTM!

MaxText/maxengine.py
```diff
@@ -49,6 +49,7 @@ class DecodeState:
   generate_cache: jax.Array
   generate_cache_index: int
   generate_lengths: jax.Array
+  generated_token: jax.Array
```
JoeZijunZhou, line 52: is it `generated_tokens`? BTW seems the fields of DecodeState doesn't match in the code?

jwyang-google: actually in the code, the "generated_tokens" are generated_lengths here, "tokens" are corresponding to generated_token here.

JoeZijunZhou: Should the field name match to each other? Otherwise, it's not using the pre-defined fields?

gobbleturk (approved):

PR #730, opened 2024-06-27
Support partial overrides for logical_axis_rules.

golechwierowicz:
Sometimes we want to have different shardings for particular models.
They change quite frequently in `base.yml` file and require updating in concrete files.
This diff adds support for partial overrides e.g. overriding only logical axis of interest, not the entire rule set.
The usage is best demonstrated in thy `yml` files included in the diff.

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -381,6 +381,30 @@ def validate_megablox_parallelism(raw_keys):
                                using_pipeline_parallelism(raw_keys)):
     raise ValueError("Currently we only support Megablox with data parallelism.")
 
+def create_new_logical_axis_rules(old_logical_axis_rules, new_logical_axis_rules):
+  new_logical_axis = set()
+  replacements = []
+  for logical_axis, mesh_axes in new_logical_axis_rules:
+    logical_axis_exists = any(rule for rule in old_logical_axis_rules if rule[0] == logical_axis)
+    if not logical_axis_exists:
+      continue
+    replacements.append((logical_axis, mesh_axes))
+    new_logical_axis.add(logical_axis)
+  old_logical_rules_filtered = [(old_logical_axis, old_mesh_axes) for old_logical_axis, old_mesh_axes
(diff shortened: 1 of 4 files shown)


MaxText/tests/pyconfig_test.py
gobbleturk, line 1: Wow Awesome, thank you for including this!

MaxText/pyconfig.py
```diff
@@ -376,6 +376,27 @@ def validate_megablox_parallelism(raw_keys):
                                using_pipeline_parallelism(raw_keys)):
     raise ValueError("Currently we only support Megablox with data parallelism.")
 
+def update_key(raw_keys, model_keys, key):
```
gobbleturk, line 379: Can we rename this update_model_keys?

MaxText/pyconfig.py
```diff
@@ -376,6 +376,27 @@ def validate_megablox_parallelism(raw_keys):
...
+  """Update `key` value in `raw_keys` from the value in `model_keys`. """
+  assert key in model_keys and key in raw_keys
+  if key == 'logical_axis_rules':
```
gobbleturk, line 382: Nit: Can we wrap the code handling logical_axis_rules into its own submethod?

gobbleturk (approved):

MaxText/tests/pyconfig_test.py
golechwierowicz: :)

MaxText/pyconfig.py
golechwierowicz: Done.

MaxText/pyconfig.py
golechwierowicz: Done.

golechwierowicz:
Changes applied, PTL. I'm not sure where does copybara error comes from though.

gobbleturk:
It looks like you are missing the copyright header on the new files. https://screenshot.googleplex.com/5SsDGDimrKsSUHf

E.g. you need [this](https://screenshot.googleplex.com/5HyjoykkwZeHHQ5) in the new files you are adding

PR #731, opened 2024-06-27
Load/Save Aqt quantized checkpoint.

singh-mitali:
Add scripts to load/save aqt quantized checkpoint
- Add sharding info to aqt dot general for loading checkpoints according to configured sharding of weights.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ jaxlib>=0.4.30
 orbax-checkpoint>=0.5.20
 absl-py
 array-record
-aqtp
+aqtp==0.7.4
 cloud-accelerator-diagnostics
 cloud-tpu-diagnostics
 datasets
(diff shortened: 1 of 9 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ jaxlib>=0.4.30
 orbax-checkpoint>=0.5.20
 absl-py
 array-record
-aqtp
+aqtp==0.7.4
```
gobbleturk, line 6: Is this necessary? It wans't in the previous PR? At the very least can we use >=0.7.4 instead of equals?

gobbleturk (approved):

requirements.txt
singh-mitali: Needed as some recent recent aqt changes are breaking the code. Resolving this with aqt team and will remove this requirement thereafter.

vipannalla (approved):
LGTM

PR #737, opened 2024-06-29
Adding option for int4 quantization to kvcache.

singh-mitali:


gobbleturk (approved):
LGTM.

Are there any flags or jax config necessary so that int4 actually uses less memory than int8? Have you observed less memory usage with int4 from this PR?

vipannalla (approved):
LGTM, thanks for adding this config. can you add a link to initial results doc (with access to reviewers) for reference?

mitalisi:
> LGTM.
> 
> Are there any flags or jax config necessary so that int4 actually uses less memory than int8? Have you observed less memory usage with int4 from this PR?

It will require layout tuning specific to the platform. Will share numbers as we evaluate/optimize for llama70b.

PR #744, opened 2024-07-02
[DON'T MERGE] GCS Distributed Training Benchmark Infra + File-parallelism + Range-read Parquet files

bernardhan33:
# This is created as a draft PR for GCS internal members to comment. This will not be merged to main.

## File-parallelism + Range-read Parquet files
This PR supports [Benchmark-7: File-Parallelism + Sequential-Read] of [Obidos Storage Benchmarks](https://docs.google.com/document/d/12sCP3EipoBJB10mlx67uiF27YYT_n29HRnOt2K6I1ok/edit?tab=t.0) and it builds up the infra for adding benchmark-8, benchmark-9 and potential future GCS PyTorch Connector's benchmarks that use PyTorch Data loading.

### Feature Set
- configurable `epochs`, `local_batch_size`, `prefetch_factor`, `data_loader_num_workers`, `per_step_computation_time` (won't block data prefetching) through arguments.
- a sample yaml file that denotes the workload spec. Added here to assist with code review. This YAML file will finally be checked into google3.
- File-parallelism among running pods. Range reading Parquet files. Padding parquet file lists to make sure all pods are having the same number of steps to avoid barrier deadlocking.
- Metrics aggregation: per-epoch training time, per-step data loading time and per-step total time. The metrics are gathered by variables and uploaded to the GCS bucket in CSV format after the run completes. We are providing a Python program to further aggregate the individual metrics files into a combined CSV file and optionally uploading to BigQuery for efficient and reliable log metric analysis.

### TODOs (yet to create tickets, will do)
- currently it bypasses the list storm by directly constructing the file names for each pod to read from. We can optionally turn on this option if user wants to test this behavior, and collects the listing metrics.
- supports shuffling between epochs.
- to build a docker image based off this PR so users would just need to use that image unless they'd want to read/modify the code. The built docker image will be used in the YAML file to be added in the g3doc.

### Tests
All metrics have been uploaded to the [distributed-training-metrics bucket](https://pantheon.corp.google.com/storage/browser/distributed-training-metrics/xpk-test-workload?e=-13802955&mods=monitoring_api_prod&project=gcs-tess&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))) and the BigQuery dataset is created [here](https://pantheon.corp.google.com/bigquery?e=-13802955&mods=monitoring_api_prod&project=gcs-tess&ws=!1m4!1m3!3m2!1sgcs-tess!2sxpk_test_workload). Verified that the BigQuery tables are containing the correct amount of entries and format.


MaxText/configs/base.yml
```diff
@@ -353,3 +353,17 @@ enable_single_controller: False
 
 # Split physical axes for https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.mesh_utils.create_device_mesh.html
 allow_split_physical_axes: False
+
+# Attributes needed for the distributed training loop using standalone_dataloader.
```
MattIrv, line 357: These seem a little bit separate from the rest of the configs in this file. Is there some way to create a separate set of configs specific to the standalone dataloader? Or is this where the MaxText team suggested to put them?

MaxText/configs/base.yml
```diff
@@ -353,3 +353,17 @@ enable_single_controller: False
...
+# The number of samples fetched by a batch per node. 
+local_batch_size: 32
+# Number of batch sizes that each data loader worker attempts to prefetch.
+prefetch_factor: 2
+# Number of data loader workers (per pod).
+data_loader_num_workers: 2
+# Number of seconds to sleep in between steps. This will not block prefetching.
+per_step_computation_time: 0.0
+# The GCS bucket that holds the dataset.
+dataset_bucket: xai-hf-dataset-parquet
```
MattIrv, line 369: Should this still be the default here?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
+apiVersion: jobset.x-k8s.io/v1alpha2
```
MattIrv, line 1: Does this need a license at the top and/or any sort of description?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool # 1:1 job replica to node pool assignment
+spec:
+  failurePolicy:
+    maxRestarts: 0
+  replicatedJobs:
+    - name: slice-job
+      replicas: 1
+      template:
+        spec:
+          parallelism: 1024    # Equal to the number of VMs per slice
```
MattIrv, line 18: Should the default here be something that is going to be more realistic for quick experimental runs?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+                        values:
+                        - default-pool
+
+              nodeSelector:
+                cloud.google.com/gke-ephemeral-storage-local-ssd: "true"
+              
+              # Needed for GCSFuse.
+              initContainers:
+              - name: gke-gcsfuse-sidecar
+                image: gcr.io/gcs-tess/ayushsethi-custom-csi/gcs-fuse-csi-driver-sidecar-mounter:v1.4.0-36-gb83e1b73-dirty
```
MattIrv, line 50: Do we want to use the real CSI driver by default for the benchmark?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+                restartPolicy: Always
+                env:
+                - name: NATIVE_SIDECAR
+                  value: "TRUE"
+                # resources:
```
MattIrv, line 55: Why are these commented out? Can you either add a comment explaining it or remove them if they're not needed?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+                - mountPath: /gcsfuse-buffer
+                  name: gke-gcsfuse-buffer
+                - mountPath: /gcsfuse-cache
+                  name: gke-gcsfuse-cache
+                
+              priorityClassName: medium
+              hostNetwork: true
+              dnsPolicy: ClusterFirstWithHostNet
+              terminationGracePeriodSeconds: 30
+              serviceAccountName: bernardhan-benchmark
```
MattIrv, line 83: Should this be updated for the checked-in version?

MaxText/deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+              containers:
+              - name: jax-cpu
+                image: gcr.io/gcs-tess/bernardhan_runner_test
```
MattIrv, line 86: Same as above, should this be updated for the checked-in version?

MaxText/standalone_dataloader.py
```diff
@@ -24,52 +24,183 @@
...
+        while len(each) < max_length:
+            each.append(random.choice(lst))
+
+    return sublists
+
+
+def parquet_data_loader(config):
+    batch_size = config.local_batch_size
+
+    # [Short-term]: we know what the mount path is, and we know what the dataset is, e.g., xai-hf-dataset-parquet.
```
MattIrv, line 78: Remove bucket name

MaxText/standalone_dataloader.py
```diff
@@ -24,52 +24,183 @@
...
+        )
+        max_logging.log(
+            f"STANDALONE DATALOADER Metrics: Per-step data loading times for epoch {i} on host {jax.process_index()}: {data_loading_time[i]}."
+        )
+        max_logging.log(
+            f"STANDALONE DATALOADER Metrics: Per-step times for epoch {i} on host {jax.process_index()}: {step_time[i]}."
+        )
 
 
 def main(argv: Sequence[str]) -> None:
```
MattIrv, line 191: Why the reformatting below here? Is there a specific formatter that MaxText is using to make it consistent with the rest of this codebase? Same for other files in this PR

MaxText/train.py
```diff
@@ -41,9 +41,11 @@
 import optimizers
```
MattIrv, line 41: It's hard to tell what the actual change in this file is other than formatting. If it does need to be reformatted, can you check that in with a separate PR first?

maxtext_dependencies.Dockerfile
```diff
@@ -12,6 +12,13 @@ RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyri
 # Install the Google Cloud SDK
 RUN apt-get update && apt-get install -y google-cloud-sdk
 
+# Install kubectl
```
MattIrv, line 15: Are you sure this is desired? Maybe it should be in a separate docker file specific to the benchmark?

MaxText/configs/base.yml
bernardhan33: Yeah this is the place for hosting all the config parameters as the command line argument looks like `python3 MaxText/standalone_dataloader.py MaxText/configs/base.yml`. It currently is hosting parameters that the normal `standalone_dataloader` would use. I'm just creating a separate section to group the commands that we support.

We can also implement additional command line arguments parsing directly in `standalone_dataloader` but I would think that introduced discrepancies.

MaxText/configs/base.yml
bernardhan33: Yeah, "xai-hf-dataset-parquet" default for the small Parquet dataset where files can fit into memory. ""xai-hf-dataset-parquet-10g" for the larger Parquet dataset where files cannot fit into memory.

MaxText/deployment.yaml
bernardhan33: I didn't add a license because this YAML file will only be checked in to google3 where we don't need license. For MaxText, they have been recommending users to use `xpk workload create` so they probably don't want the manual yaml files to be checked in anyways.

MaxText/deployment.yaml
bernardhan33: Yes, I'm going to provide instructions for modifying the YAML file in the google3 CL where we check in the documentation and the YAML file.

MaxText/deployment.yaml
bernardhan33: This is the specific GCSFuse CSI Driver. PD or Parallelstore will need to specify their own CSI Driver definition. I will remove this when we check in the YAML file to google3.

MaxText/deployment.yaml
bernardhan33: I commented this out because I copied this from Ayush's sample command and forgot to delete it. These lines will prevent pods from being scheduled as the requested resource can't be fulfilled.

We will not need to specify the initContainer with the default GKE managed driver and as I mentioned in the comment above, this entire subsection will get deleted when we check in the YAML file into google3.

MaxText/deployment.yaml
bernardhan33: Definitely. I will add instructions in the google3 CL.

MaxText/deployment.yaml
bernardhan33: Yep,  I will add instructions in the google3 CL. And as the description of this CL indicates, I will build a docker image based off this PR so users would just need to use that image unless they'd want to read/modify the code. The built docker image will be used in the YAML file to be added in the g3doc.

MaxText/standalone_dataloader.py
bernardhan33: Ack. The default Python formatter we use gives 4 spaces for tabs and we need to use 2 spaces. Also [recommended](go/pystyle#indentation) by internal Python style guide.

MaxText/train.py
bernardhan33: Yeah I accidentally "save" this file so it's reformatted with 4 spaces for indentation. Will fix.

maxtext_dependencies.Dockerfile
bernardhan33: Needed to work past the STS issue. Without `kubectl` and explicitly waiting for the jobset to be ready, JAX opration will time out after 5 mins when not all pods have joined the group. This is likely caused by the STS issue which hopefully should get resolved after we use the GKE managed driver version.

 I will keep re-evaluating the need of this and remove this once we can see all pods initialized in parallel.

MaxText/standalone_dataloader.py
bernardhan33: done.

MaxText/standalone_dataloader.py
bernardhan33: Modified the format to align with 2 spaces for indentation.

MaxText/train.py
bernardhan33: Not showing diff for this file now as the styles are consistent. Resolving this.

maxtext_dependencies.Dockerfile
bernardhan33: I removed this section as I confirmed that there was an issue with the GCSFuse sidecar we used. This will likely not happen for the other CSI drivers of other storage solutions.

MaxText/deployment.yaml
```diff
@@ -0,0 +1,186 @@
+apiVersion: v1
```
bernardhan33, line 3: @MattIrv @divrawal I updated the yaml file to better match what we will provide in the google3 doc and removed a couple of hacks as we are having the correct GCSFuse CSI Driver version now. Please take a look

PR #746, opened 2024-07-02
Add convergence tests on A3 GPU

michelle-yooh:


end_to_end/gpu/a3/test_convergence_1b_params.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+set -ex
+
+echo "Running test_convergence_125m_params.sh"
```
anfals, line 4:
```suggestion
echo "Running test_convergence_1b_params.sh"
```

end_to_end/gpu/a3/test_convergence_1b_params.sh
```diff
@@ -0,0 +1,64 @@
...
+# Run this on NVIDIA A3 to achieve a loss value of ~4.0 after 2500 steps
```
anfals, line 5: I assume this is also a copy paste error, but what loss value will you be targeting? The same as the TPU equivalent test?

anfals (approved):
Looks good, other than some minor comments about some print statements!

gobbleturk (approved):

PR #747, opened 2024-07-03
Update tile size

RissyRan:
# Description

Update custom tile size to a fix one, and get a better MFU and throughput: ~0.6% increase

# Test
* Test 8x7b [link](https://pantheon.corp.google.com/kubernetes/service/us-east5/v5p-128-bodaborg-us-east5-a/default/ran-xla-flag-17/logs?e=13803378&mods=allow_workbench_image_override&project=cloud-tpu-multipod-dev)

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -354,10 +354,7 @@ class MoeBlock(nn.Module):
     return output.reshape(int(self.config.per_device_batch_size), -1, self.config.emb_dim).astype(self.dtype)
 
   def megablox(self, inputs, gate_logits, config, w0_kernel, w1_kernel, wo_kernel):
-    # TODO(ranran): need to changes in JAX repo to enable optimized tile_size
-    #               instead of the static default tile_size (512, 512, 512)
-    tile_size = (512, 512, 512)
-
+    tile_size = (512, 1024, 1024)
     def gmm(inputs, kernel, group_sizes):
       hs_shape = inputs.shape
       # pad length is the 1st dimension of tiling size in gmm call


MaxText/configs/models/mixtral-8x1b.yml
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# model config for mixtral-8x1b
+
+base_emb_dim: 4096
+base_num_query_heads: 32
+base_num_kv_heads: 8
+base_mlp_dim: 14336
+base_num_decoder_layers: 5
```
gobbleturk, line 21: How did you choose these sizes? Generally we scale all "base" dimensions roughly proportionally, it looks like you only changed the layers here? Looks fine otherwise

RissyRan: Yes, I only changed the number of layers to scale down to 8x1b.

gobbleturk: I suspect this doesn't converge as well as scaling each proportionally (would have to run experiments). Could we include a comment at the top of this file saying it is a custom scaling of mixtral-8x7b where we only change the layers for smaller scale performance testing, and we haven't run convergence experiments?

RissyRan: Thanks Matt! After a second thought, I removed this small model config, and plan to add it into a branch.

gobbleturk (approved):

MaxText/configs/models/mixtral-8x1b.yml
gobbleturk: Sounds good, thank you!

PR #750, opened 2024-07-03
Fix validation error for other models

RissyRan:
# Description

Nightly gpt3 175b tests met megablox parallelism validatoin error `ValueError: Currently we only support Megablox with data parallelism` ([here](https://screenshot.googleplex.com/9ePHtqEPxwRtdsR)). This is because we set megablox flag as True when cleaning up brute force implementation ([here](https://github.com/google/maxtext/blob/89edace7ecf346381857c6d5c3ae0a432f02653e/MaxText/configs/base.yml#L94)).

The solution is to turn on this validation only for MoE model.

# Test

Running train_compile.py script (non-MoE): [link](https://paste.googleplex.com/6188773140594688)

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -396,8 +396,9 @@ def validate_and_update_keys(raw_keys, model_keys, config_name: str):
   """Validate and update model specific config keys"""
   max_logging.log("Updating following parameters in config\n")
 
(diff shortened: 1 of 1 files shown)


raymondzouu (approved):
Thank you for the quick fix @RissyRan!

gobbleturk (approved):

PR #757, opened 2024-07-08
Enable quantization for MoE Gating

RissyRan:
# Description
* Enable quantization for MoE Gating (will work on kernel quantization as next step)
* Observed perf regression when quantizing einsum(), so not include in this PR. Need to dive deep for matmul implementation later.

# Test
Adding `quantization=int8` flag:
* Megablox without kernel quantization ([test](https://pantheon.corp.google.com/kubernetes/service/us-east5/v5p-128-bodaborg-us-east5-a/default/ran-mega-quant2/logs?e=13803378&mods=allow_workbench_image_override&project=cloud-tpu-multipod-dev), mixed precision in [xprof](https://screenshot.googleplex.com/8ZMi3e63EqHYGBN) in gating)

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -278,6 +278,7 @@ class MoeBlock(nn.Module):
     kernel_axes: Tuple with axes to apply kernel function.
     weight_dtype: Type for the weights.
     dtype: Type for the dense layer.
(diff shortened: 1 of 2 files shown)


gobbleturk (approved):
Please include in the PR description that you are only quantizing the MoE expert gating, not the main FF/expert computations

PR #763, opened 2024-07-11
Add enable_model_warmup flag for AOT compilation at model server start

vivianrwu:
Add the `enable_model_warmup` flag at model server start
Associated PR: https://github.com/google/JetStream/pull/92

```
 - model_name=gemma-7b
  - tokenizer_path=assets/tokenizer.gemma
  - per_device_batch_size=1
  - max_prefill_predict_length=1024
  - max_target_length=2048
  - async_checkpointing=false
  - ici_fsdp_parallelism=1
  - ici_autoregressive_parallelism=-1
  - ici_tensor_parallelism=1
  - scan_layers=false
  - weight_dtype=bfloat16
  - load_parameters_path=<ckpt_path>
  - enable_model_warmup=true

curl --request POST --header "Content-type: application/json" -s localhost:8000/generate --data '{
    "prompt": "What are the top 5 programming languages",
    "max_tokens": 200
}'
{
    "response": " for data science in 2023?\n\n1. Python\n2. R\n3. SQL\n4. Java\n5. Scala\n\n**Note:** The order is based on popularity and demand in the data science industry in 2023."
}
```


MaxText/maxengine_server.py
```diff
@@ -54,7 +54,8 @@ def main(config):
       devices=devices,
       metrics_server_config=metrics_server_config,
       enable_jax_profiler=config.enable_jax_profiler if config.enable_jax_profiler else False,
-      jax_profiler_port=config.jax_profiler_port if config.jax_profiler_port else 9999
+      jax_profiler_port=config.jax_profiler_port if config.jax_profiler_port else 9999,
+      enable_model_warmup=config.enable_model_warmup if config.enable_model_warmup else False
```
gobbleturk, line 58: I think this logic was to try to be safe if the config variable was set to None, but actually it will never be None. If a config is missing it then this would return a "no key" error. You can just use the simpler enable_model_warmup=config.enable_model_warmup

gobbleturk (approved):

JoeZijunZhou (approved):

MaxText/maxengine_server.py
JoeZijunZhou: That's correct. Just in case if user used a config without setting a default value, we want to make sure it doesn't break the JetStream MaxText server.

vivianrwu: +1 to @JoeZijunZhou's comment, if the config used does not have a default value, having the `else False` will prevent the model warmup logic from running regardless. I think it is safer to keep as is. Let me know if you think otherwise @gobbleturk

PR #767, opened 2024-07-12
Integrate emergency checkpointer into standalone_checkpointer for CPUs.

RoshaniN:
1. When local checkpoints are available for restore, alter mesh setup as follows.

- Ignore the JAX coordinator provided by XPK and override the JAX coordinator to be the pod containing process_id 0.
- Persist the IP address of this pod using socket APIs to GCS.
- Let other processes retrieve the address of this coordinator, similar to TPUs.

2. Restore state and not state["items"] if emergency checkpoint manager is enabled.

3. Some changes in Jax/Orbax are needed to get this working end-to-end on CPUs. Other changes are addressed in this PR.

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -15,7 +15,6 @@ limitations under the License.
 """
 
 """ Common Max Utils needed by multiple modules"""
-import shutil
 import checkpointing
 import common_types
 import functools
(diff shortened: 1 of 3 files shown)


xuefgu:
Can you please resolve the conflicts first? I see that the last update was from 2 weeks ago.

PR #774, opened 2024-07-16
Fix AOT for int8 and add HLO identical test

gobbleturk:
* Add the int8 rng flag to train_compile
* Add test that train_compile produces identical HLO to train.py


tonyjohnchen (approved):

raymondzouu (approved):

RissyRan (commented):
Thanks Matt for the fix! I guess it matches on v4. Do you have a chance to test on v5? Thanks!

PR #778, opened 2024-07-16
Incorporate GKE's stopgap solution in the absence of in-place pod restart

xuefgu:


MaxText/max_utils.py
```diff
@@ -272,65 +272,36 @@ def initialize_jax_for_cpu():
...
-
-def _retrieve_coordinator_address(coordinator_address_file):
-  for _ in range(30):
-    if coordinator_address_file.exists():
-      return coordinator_address_file.read_text()
+def _retrieve_jax_init_info(raw_keys):
+  """Retrieve JAX init info from a local file."""
+  JAX_INIT_INFO_FILE = "jax-init-info.txt"
+  local_jax_init_info_file = epath.Path(raw_keys["local_checkpoint_directory"]) / JAX_INIT_INFO_FILE
+  for i in range(900):
```
gobbleturk, line 297: I'm confused by this - isn't the file local? Why does it take time to find it? Can you add a comment here?

gobbleturk (approved):

MaxText/max_utils.py
xuefgu: Added the comment to explain.

PR #781, opened 2024-07-17
Add Support for Parquet, File-Parallelism + Random-Read

JGoodlad:


MaxText/configs/base.yml
```diff
@@ -371,4 +371,6 @@ dataset_directory: "/mnt/gcsfuse"
 gcs_metrics_bucket: "distributed-training-metrics"
 ### Parameters used for GCSFuse
 # The GCS bucket that holds the dataset.
-dataset_bucket: xai-hf-dataset-parquet
\ No newline at end of file
+dataset_bucket: xai-hf-dataset-parquet
+# The strategy used for reading the dataset.
```
bernardhan33, line 375: nit: point to the location that lists the enums of these strategies.

MaxText/torch_datasets/parquet.py
```diff
@@ -14,49 +14,113 @@
...
+      - Parquet metadata for all row groups is at the beginning of the file which
+        makes getting the metadata for row groups cheap.
+
+    Some implementation details:
+      - Row groups that are not a multiple of the batch size will return partial batches
+        on the last read of the row group.
+      - The reading of the row groups within a parquet file is random.
+    """
+    for single_parquest_file in assigned_parquet_files:
+      parquet_file = pq.ParquetFile(single_parquest_file, buffer_size = 0, pre_buffer = 0)
```
bernardhan33, line 94: Would be great to add a comment (link source, if available) on why these buffer_size , pre_buffer are 0.

MaxText/torch_datasets/parquet.py
```diff
@@ -14,49 +14,113 @@
...
+      # Randomize the row groups
+      randomized_row_groups = self._random_iterator(range(parquet_file.num_row_groups))
+      for current_row_group in randomized_row_groups:
+        max_logging.log(f"row_group {current_row_group} has {parquet_file.metadata.row_group(current_row_group).num_rows} rows")
+        # TODO: Read batch size until all row groups have been read.
```
bernardhan33, line 99: Can we remove the TODO now? If this needs to be there, can we file a ticket to track it?

bernardhan33 (approved):
This is absolutely awesome work! Nice refactoring and code written for the enum, etc.

MaxText/configs/base.yml
JGoodlad: Done.

MaxText/torch_datasets/parquet.py
JGoodlad: Sure! Removed.

MaxText/torch_datasets/parquet.py
JGoodlad: I added the default parameters to make it clear that buffering would not be used (it is also not used by default but a question came up about buffering while I was working on this).

Added a comment to reflect this motivation.

MaxText/torch_datasets/parquet.py
```diff
@@ -14,17 +14,86 @@
...
+    # batch_size is basically chunk size
+    self.allocated_parquet_files = allocated_parquet_files
+    self.columns = columns
+    self.batch_size = batch_size
+
+  def __iter__(self):
+    """File Parallel Random Read iterator.
+
+    Some facts:
+      - Parquet metadata for all row groups is at the beginning of the file which
```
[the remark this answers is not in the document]
JGoodlad: You are right. Fixed.

PR #782, opened 2024-07-17
[DON'T MERGE] GCS Checkpointing Testing Workload modification

bernardhan33:
# This is created as a draft PR for GCS internal members to comment. This will not be merged to main.

## Checkpointing a 64B model through MaxText
- Read and Write times to be collected and sent to GCS buckets before a separate Python program aggregates and uploads to BQ. I've created b/353631904 to track the improvement of letting each pod to write directly to BQ, which is currently blocked by needed nodepool recreation.
- A sample YAML file is provided for code review purposes.


MaxText/configs/base.yml
```diff
@@ -390,3 +390,6 @@ enable_single_controller: False
 
 # Split physical axes for https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.mesh_utils.create_device_mesh.html
 allow_split_physical_axes: False
+
+# Attributes needed for the MaxText checkpointing loop using standalone_checkpointer.
+gcs_metrics_bucket: distributed-checkpointing-metrics
```
MattIrv, line 395: Should there be / is there any ability to provide a folder or path instead of just the bucket?

PR #784, opened 2024-07-17
Temp fix for c4_mlperf dataset pipeline

ZhiyuLi-goog:
# Desc.
Got pipeline hanging with `per_device_batch_size<1`.

Root cause of the issue:
1. train and eval may have multiple `per_device_batch_size` and we need independent `process_indices` and `expansion_factor_real_data` for each. Currently, the eval_iterator uses the same `process_indices` as the training phase, regardless of whether `eval_per_device_batch_size` matches or not. 
2. with `per_device_batch_size=0.5` in eval_iterator, where only partial processes are activated, this following line (to pad uneven num of shardings with remainders) would be problematic (hanging). It can be walked around by gathering activated processes or calculating the results with the size of eval example.
https://github.com/google/maxtext/blob/77f079f845084ba853453dc0755deb0daf312e26/MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py#L184

# Current solution
- [x] split train/eval iterator and `process_indices` in c4_mlperf dataset 
- [x] disable `per_device_batch_size < 1` for eval iterator
- [x] `per_device_batch_size < 1` in training iterator is as usual 
- [x] additionally, add a read_in_shard option for large-scale 6k runs.

Tested in experiments.

# Long term solution
Need independent `expansion_factor_real_data` for eval iterator. Refactor and split train and eval iterators for all type of datasets with independent `process_indices`.  Currently, only c4_mlperf dataset use eval iterator actively.

### DIFF
--- a/MaxText/configs/models/gpt3-175b.yml
+++ b/MaxText/configs/models/gpt3-175b.yml
@@ -42,3 +42,4 @@ adam_eps: 1.e-8
 adam_weight_decay: 0.1
 checkpoint_period: 10_000
 target_eval_loss: 2.69
+eval_per_device_batch_size: 1
--- a/MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
+++ b/MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
@@ -33,6 +33,7 @@ import tokenizer
 import multihost_dataloading
 import sequence_packing
 from input_pipeline._input_pipeline_utils import get_tokenizer
+import max_logging
 
 AUTOTUNE = tf.data.experimental.AUTOTUNE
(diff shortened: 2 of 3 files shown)


khatwanimohit (approved):
Thank you Zhiyu!

gobbleturk (approved):
Thank you for fixing this! cc @aireenmei, would request review also from Aireen but is OOO for a bit

ZhiyuLi-goog:
Thank you both.

I will merge it first and sync up with @aireenmei next week. I do not expect it will affect other pipelines.

aireenmei (commented):
LGTM, thanks Zhiyu!

PR #787, opened 2024-07-18
gradient accumulation

tonyjohnchen:
Adding new feature `gradient accumulation` to only update weight for every x steps.

Example command without using `gradient accumulation`: 
```
python3 MaxText/train.py MaxText/configs/base.yml   base_output_directory=${MAXTEXT_OUTPUT_PATH} run_name=${RUN_NAME}    enable_checkpointing=false async_checkpointing=false    per_device_batch_size=1    skip_first_n_steps_for_profiler=5 steps=30    dataset_type=synthetic    profiler=xplane
```

Example command with using `gradient accumulation`: 
```
python3 MaxText/train.py MaxText/configs/base.yml   base_output_directory=${MAXTEXT_OUTPUT_PATH} run_name=${RUN_NAME}    enable_checkpointing=false async_checkpointing=false    per_device_batch_size=1    skip_first_n_steps_for_profiler=5 steps=30    dataset_type=synthetic    profiler=xplane gradient_accumulation_steps=10
```

[Result1](https://paste.googleplex.com/6037166750957568)
[Result2](https://paste.googleplex.com/6279556128571392)

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -55,6 +55,7 @@ import jax.numpy as jnp
 from jax import random
 from jax.sharding import Mesh
 from jax.experimental import checkify
+from jax.tree_util import tree_map
 
 from cloud_tpu_diagnostics import diagnostic
 from cloud_tpu_diagnostics.configuration import debug_configuration
@@ -79,7 +80,10 @@ def validate_train_config(config):
   if not config.base_output_directory.startswith("gs://"):
     max_logging.log("WARNING: 'base_output_directory' might be pointing your local file system")
   assert config.steps > 0, "You must set steps or learning_rate_schedule_steps to a positive integer."
-
+  
+  if config.quantization=='fp8':
+    # pylint: disable=line-too-long
+    assert config.gradient_accumulation_steps == 1, "fp8 can't be used with gradient_accumulation_steps right now. Please use other quantization or set gradient_accumulation_steps to 1"
 
 def get_first_step(state):
   with jax.spmd_mode("allow_all"):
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -267,21 +272,33 @@ def train_step(model, config, state, data, dropout_rng):
...
 
-  if config.gradient_clipping_threshold > 0:
-    grads = maxtext_utils.apply_gradient_clipping(raw_grads, state, config.gradient_clipping_threshold)
-  else:
-    grads = raw_grads
-  new_state = state.apply_gradients(grads=grads)
+    grads_accumlation = tree_map(lambda x, y: x + y, grads_accumlation, grads)
+    raw_grads_accumlation = tree_map(lambda x, y: x + y, raw_grads_accumlation, raw_grads)
+
+  final_grads = tree_map(lambda x: x / config.gradient_accumulation_steps, grads_accumlation)
```
anfals, line 293: For 'FP8' we have these special 'overwrite_with_gradient' collections with quantization scales/histories. We shouldn't sum those 'gradients' - just using the latest one should be the right thing I think

anfals (commented):
Double

MaxText/train.py
tonyjohnchen: Do you mean `FP8` for GPU?

So maybe we should skip this `gradient_accumulation_steps` feature when `dtype` is `fp8`?

cc: @michelle-yooh

anfals: Yes I mean when `quantization=fp8`. And thinking more about this again, I don't think using the last one like I mentioned above would work super great - lot of history being lost. 

I think for now lets just skip/assert both can't be used. The new version of FP8 I'm working on has a different approach for updating these scales/history that I think will work better with this, so we can revisit then.

RissyRan: Wondering if not just 'FP8' to use the last grad. Why not all use the last grad?

MaxText/train.py
```diff
@@ -146,13 +147,18 @@ def write_metrics_to_tensorboard(writer, metrics, step, config):
...
-        f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
-        f"loss: {metrics['scalar']['learning/loss']:.3f}"
+    log_message = (
+      f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
+      f"TFLOP/s/device: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
+      f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
+      f"loss: {metrics['scalar']['learning/loss']:.3f}"
     )
 
+    if config.gradient_accumulation_steps > 1:
```
RissyRan, line 157: Should we just print it one time in the config, as it won't change in this run. Right?

MaxText/train.py
```diff
@@ -146,13 +147,18 @@ def write_metrics_to_tensorboard(writer, metrics, step, config):
 
     full_log = step % config.log_period == 0
 
-    max_logging.log(
-        f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
-        f"TFLOP/s/device: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
-        f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
-        f"loss: {metrics['scalar']['learning/loss']:.3f}"
+    log_message = (
+      f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
```
gobbleturk, line 151: This may be confusing to users - if each substeptakes 1 second and they have 10 gradient accumulation steps then 10 real seconds went by, but this will only print 1 second - I was thinking to change the message to "seconds per gradient accumulation step" or similar when gradient_accumulation_steps > 1

gobbleturk (changes requested):

MaxText/train.py
tonyjohnchen: Done

MaxText/train.py
tonyjohnchen: I want to print it for every step in case users missed it if we only print it one time.

MaxText/train.py
tonyjohnchen: Added the assert to make sure `quantization=fp8` can't be used with `gradient_accumulation_steps`

MaxText/train.py
```diff
@@ -267,21 +272,33 @@ def train_step(model, config, state, data, dropout_rng):
...
+  raw_grads_accumlation = tree_map(jnp.zeros_like, state.params)
+
+  for grad_accumulate_substep in range(config.gradient_accumulation_steps):
+    train_loss_fn = functools.partial(loss_fn, model, config, data[grad_accumulate_substep], dropout_rng, is_train=True)
+    grad_fn = jax.value_and_grad(train_loss_fn, has_aux=True)
+
+    (loss, aux), raw_grads = grad_fn(state.params)
+    intermediate_outputs = aux["intermediate_outputs"]
+    if config.gradient_clipping_threshold > 0:
+      grads = maxtext_utils.apply_gradient_clipping(raw_grads, state, config.gradient_clipping_threshold)
```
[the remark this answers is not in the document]
tonyjohnchen: Done.

MaxText/train.py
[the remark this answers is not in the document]
RissyRan: Agree with above test to verify the result.

Also trying to learn more here. After some research, it seems we should do:

1) Accumulated Gradients += Gradients (from current accumulation step)
2) Averaged Gradients = Accumulated Gradients / Number of accumulation steps (instead of doing average all the way, or calculate gradients at the last microbatch only)

PR #788, opened 2024-07-18
Update tp sharding annotation for MoE block

RissyRan:
# Description

Update tp sharding annotation for shard_map to match same sharding strategy in MaxText

# Test

* unit test passes
* xprof - [link](https://screenshot.googleplex.com/5BkmFGyVsxrH9MD)

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -49,9 +49,6 @@ bias_init = initializers.default_bias_init
 RMSNorm = normalizations.RMSNorm
 Quant = quantizations.AqtQuantization
 
-BATCH = common_types.BATCH
-EMBED = common_types.EMBED
-
 def _convert_to_activation_function(fn_or_string: Union[str, Callable[..., Any]]) -> Callable[..., Any]:
   """Convert a string to an activation function."""
   if fn_or_string == "linear":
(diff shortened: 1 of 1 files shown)


MaxText/layers/linears.py
```diff
@@ -221,7 +221,7 @@ def __call__(self, inputs, decode: bool = False, deterministic: bool = False):
           dtype=self.dtype,
           weight_dtype=self.weight_dtype,
           kernel_init=self.kernel_init,
-          kernel_axes=("embed", "num_activations", "mlp"),
```
gobbleturk, line 224: nit: I would leave these as mlp just b/c the mix of strings and common_types looks strange, but either is fine...

MaxText/layers/linears.py
```diff
@@ -385,27 +385,25 @@ def gmm(inputs, kernel, group_sizes):
         shard_map.shard_map,
         mesh=self.mesh,
         in_specs=(
-              (nn.logical_to_mesh_axes((BATCH, None, EMBED))),
               (nn.logical_to_mesh_axes((BATCH, None, None))),
```
gobbleturk, line 388: Can we add a comment here (or at top of annotations) saying we purposefully AG the input activations over TP to follow a specific sharding strategy outlined in https://parsa.epfl.ch/course-info/cs723/papers/Megatron.pdf 

(we can include this link the comment)

gobbleturk (approved):

MaxText/layers/linears.py
RissyRan: lol, np, I am fine with either way. I could change it back without using common_type.

MaxText/layers/linears.py
RissyRan: Sure thing, added

PR #793, opened 2024-07-22
support pre-tokenized dataset and fix hf issues

aireenmei:
Support pre-tokenized dataset as we use it in mlperf, also is a common use case.
fix hf issues:
1. https://github.com/google/maxtext/issues/734
2. https://github.com/google/maxtext/issues/571#issuecomment-2225965382
3. b/350068432


MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -44,22 +44,26 @@ def preprocessing_pipeline(
...
     shuffle: bool = False,
     data_shuffle_seed=0,
+    tokenize=True,
     add_bos=True,
     add_eos=True,
     num_epochs=1,
     packing=True,
     shift=True,
-    drop_remainder=True,
+    drop_remainder=False,
```
gobbleturk, line 56: What does drop_remainder do? Could we include a comment/documentation somewhere?

gobbleturk (approved):

MaxText/input_pipeline/_grain_data_processing.py
aireenmei: Sure, I'm still working on features related to drop_remainder. Will add comment/doc.

PR #795, opened 2024-07-22
mixed precision configs with higher tokens/sample

singh-mitali:
Adding mixed precision configs which qualify tokens/sample constraint for mlperf.

### DIFF
--- a/MaxText/configs/quantization/mp_scale_1.json
+++ b/MaxText/configs/quantization/mp_scale_1.json
@@ -1,0 +1,8 @@
+{
+  ".*/query": {"bits": 4, "scale": 0.8},
+  ".*/key": {"bits": 4, "scale": 0.9},
+  ".*/value": {"bits": 8},
+  ".*/out": {"bits": 4},
+  ".*/wi_0": {"bits": 4},
+  ".*/wo": {"bits": 8}
+}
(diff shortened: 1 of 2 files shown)


vipannalla (approved):
I know you can't add a comment in json, but can you add a README in `maxtext/MaxText/configs/quantization/` with a sentence or two about each of the configs in that dir?

mitalisi:
> I know you can't add a comment in json, but can you add a README in `maxtext/MaxText/configs/quantization/` with a sentence or two about each of the configs in that dir?

Sure. Will do this next week as part of b/352355780.

gobbleturk:
> > I know you can't add a comment in json, but can you add a README in `maxtext/MaxText/configs/quantization/` with a sentence or two about each of the configs in that dir?
> 
> Sure. Will do this next week as part of b/352355780.

Can we do it in this PR? At the very least for the two files that this PR adds? I didn't realize json files don't support comments. You can add comments for the other json files like those added in https://github.com/google/maxtext/pull/762 in future PRs, but we should try to have PRs stay self contained

gobbleturk (changes requested):
Can we add documentation for these as Vipan suggests?

mitalisi:
Did not use this config

PR #797, opened 2024-07-23
fix convert gemma link

Mddct:


gobbleturk:
Thank you for finding this!

PR #806, opened 2024-07-31
Support ahead-of-time compilation on A3-generation GPU VMs

jonb377:
This change allows precompiling for A3 GPU clusters. AoT compilation for GPUs works slightly differently from TPU, since the compiler needs access to a physical device for compilation.

### DIFF
--- a/MaxText/accelerator_to_spec_map.py
+++ b/MaxText/accelerator_to_spec_map.py
@@ -21,16 +21,17 @@ IF YOU MODIFY THIS FILE YOU SHOULD ALSO ADD CORRESPONDING MODIFICATIONS TO
 UserFacingNameToSystemCharacteristics in xpk/xpk.py !!!!! """
 
 from dataclasses import dataclass
+from typing import Optional
 
 
 @dataclass
 class SystemCharacteristics:
   platform: str
-  topology_name: str
-  chip_config_name: str  # 'megacore' or 'default'
-  chips_per_host_bounds: tuple
+  topology_name: Optional[str]
+  chip_config_name: Optional[str]  # 'megacore' or 'default'
+  chips_per_host_bounds: Optional[tuple]
   devices_per_slice: int
-  wrap: tuple
+  wrap: Optional[tuple]
 
 
 UserFacingNameToSystemCharacteristics = {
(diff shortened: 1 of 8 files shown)


MaxText/accelerator_to_spec_map.py
```diff
@@ -150,6 +151,8 @@ class SystemCharacteristics:
     "v5p-12288": SystemCharacteristics("tpu", "v5:16x16x24", "megacore", (2, 2, 1), 6144, (True, True, True)),
     "v5p-13824": SystemCharacteristics("tpu", "v5:12x24x24", "megacore", (2, 2, 1), 6912, (True, True, True)),
     "v5p-17920": SystemCharacteristics("tpu", "v5:16x20x28", "megacore", (2, 2, 1), 8960, (True, True, True)),
+    # A3 - H100
+    "a3": SystemCharacteristics("gpu", "a3:8", None, None, 8, None)
```
yangyuwei, line 155: Is `topology_name` field needed for A3? I didn't see it being used in the implementation. How about marking it as Optional and setting it as None by default? I'm just afraid that setting it as "a3:8" by default will provide an impression that only 8 A3 GPUs are being supported. 

Besides, it'd be great if you could add a comment about setting `compile_topology_num_slices` if users want to run for multi hosts (I see you added it in the doc below, but it doesn't hurt to also add in the code:)).

jonb377: Sure, great callouts! I'll add a comment to the code as well.

Since A3's topology only supports a single configuration, this field also isn't necessary. I'll make the field optional.

README.md
```diff
@@ -105,7 +105,7 @@ jsonPayload.verb="stacktraceanalyzer"
 
 Here is the related PyPI package: https://pypi.org/project/cloud-tpu-diagnostics.
 
-## Ahead of Time Compilation (AOT, tpu-only)
+## Ahead of Time Compilation (AOT)
```
yangyuwei, line 108: How about refactoring the documentation for AOT in the following way?

1. Add a sub section `TPU Support` and move all the TPU related instructions and existing examples for AOT under this section.
2. For the newly added section `GPU Support`, add examples for GPUs (similar to the ones for TPUs, but with different flags).

yangyuwei (commented):
Thanks Jon for adding this feature for GPUs. Super helpful!

Overall, looks great. I just left two comments. PTAL.

yangyuwei (approved):

README.md
```diff
@@ -151,5 +155,40 @@ base_output_directory=gs://my-output-bucket dataset_path=gs://my-dataset-bucket
...
 
+### GPU Support
```
gobbleturk, line 158: Awesome, thank you for including documentation!

gobbleturk (approved):

PR #812, opened 2024-08-02
Gradient Accumulation

gobbleturk:
Add a gradient accumulation metric to MaxText

This option allows users to effectively use a larger global batch even if it doesn't fit into memory by averaging the gradients over a set of sub-batches before updating the weights.

We choose to report metics on the global batch scale instead as opposed to the sub-batch scale - users of this feature would probably want to set profiler_steps to just 1 or 2 since they are now global steps

Thanks @tonyjohnchen for initial work on this in this PR: https://github.com/google/maxtext/pull/787/files


tonyjohnchen (approved):
LGTM.

Thanks Matt!

Mddct:
I have a question: 

Why not use optax.MultiStep ? It seems that there is no need to modify the training code?

gobbleturk:
> optax.MultiStep

Thank you for the suggestion! This looks like a reasonable alternative, although the code to accumulate the gradient is pretty straightforward to implement ourselves as done in this PR. Either way we must take care to differentiate between the microbatch size and full batch size, and to report metrics consistently. Doing it manually ourselves allows us to keep train_step = global step, whereas using optax.MultiStep train_step would refer to a microbatch step

MaxText/tests/gradient_accumulation_test.py
```diff
@@ -0,0 +1,72 @@
...
+          "gradient_accumulation_steps=1",       
+    ])
+
+    with open(run_accumulate_metrics_file, 'r', encoding='utf8') as accum_run,\
+          open(run_regular_metrics_file, 'r', encoding='utf8') as regular_run:
+        accum_run_loss = json.loads(accum_run.readlines()[-1])["learning/loss"]
+        regular_run_loss = json.loads(regular_run.readlines()[-1])["learning/loss"]
+        print(f"[Gradient Accumulation Test] Loss with gradient accumulation: {accum_run_loss}", flush=True)
+        print(f"[Gradient Accumulation Test] Loss without gradient accumulation: {regular_run_loss}", flush=True)
+        np.testing.assert_allclose(accum_run_loss, regular_run_loss, rtol=0.05)
```
RissyRan, line 70: Wondering if the flops from `calculate_tflops_training_per_device` can be captured somewhere?

MaxText/train.py
```diff
@@ -267,9 +278,33 @@ def train_step(model, config, state, data, dropout_rng):
...
   """
-  train_loss_fn = functools.partial(loss_fn, model, config, data, dropout_rng, is_train=True)
-  grad_fn = jax.value_and_grad(train_loss_fn, has_aux=True)
-  (loss, aux), raw_grads = grad_fn(state.params)
+  if config.gradient_accumulation_steps > 1:
+    def accumulate_gradient(acc_grad_and_loss, data):
+      grad_func = jax.value_and_grad(loss_fn, argnums=4, has_aux=True)
+      (cur_batch_loss, aux), cur_batch_gradient = grad_func(model, config, data, dropout_rng, state.params, is_train=True)
+      accumulated_loss = cur_batch_loss + acc_grad_and_loss['loss']
+      accumulated_gradient = jax.tree_util.tree_map(lambda x, y: x + y, cur_batch_gradient, acc_grad_and_loss['grad'])
```
RissyRan, line 286: Trying to understand more the logic here. Will we do the average after the sum? Or I missed it somewhere? Thanks!

gobbleturk: Instead of averaging gradients we scale the loss by loss = loss /accum_steps. This is equivalent to averaging the gradients

MaxText/tests/gradient_accumulation_test.py
gobbleturk: Added an assert for this!

MaxText/train.py
RissyRan: Which step is this `loss /accum_steps`? others look good. Thanks!

gobbleturk: it happens in a different place - loss_fn: https://github.com/google/maxtext/pull/812/files#diff-45acd14896a500d4b0919c3e844732d7401e45bdf822f6d7c89a52d735bc9267R255

gobbleturk: Actually updated so it all happens in the same place, and weights by unpadded tokens across microbatches

RissyRan (approved):
LGTM! Thanks!

PR #814, opened 2024-08-06
add Gemma2 support to MaxText 

ZhaoyueCheng:
add Gemma2 support to MaxText on top of [PR](https://github.com/google/maxtext/pull/804)

+ add `gemma` Decoder block which merges [local_sliding_attention, global_attention] combination into one Decoder layer with `post_attn_norm`, `post_ffw_norm` support
+ add `convert_gemma2_chkpt.py` to convert checkpoint from Gemma2 architecture
   + merges one [local_sliding_attention, global_attention] combination into one Decoder layer
   + adds `post_attn_norm`, `post_ffw_norm`, `transpose_gating_einsum` and `query_pre_attn_scalar` support which is new from Gemma1
+ enabling `post_attn_norm`, `post_ffw_norm` in the Gemma2 Decoder architecture
+ add `final_logits_soft_cap` support
+ add `gemma2-2b` `gemma2-9b` `gemma2-27b` config yaml file and 2b end_to_end test script and test golden logits dumped from flax


MaxText/configs/base.yml
```diff
@@ -137,11 +137,13 @@ param_scan_axis: 1
 
 # The attention parameter dictates the specific algorithm/methodology used to compute the attention scores
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
-
-# The attention_type parameter determines the variants of attention, e.g. global or local_sliding
```
gagika, line 141: Could you keep those comments?

MaxText/configs/models/gemma2-9b.yml
```diff
@@ -12,21 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# model config for gemma2-9B
```
gagika, line 15: Please update gemma2-9B

end_to_end/tpu/gemma2/9b/1_test_gemma.sh
```diff
@@ -1,46 +0,0 @@
-#!/bin/bash
```
gagika: Why was it deleted?

end_to_end/tpu/gemma2/9b/2_test_gemma.sh
```diff
@@ -1,43 +0,0 @@
-#!/bin/bash
```
gagika: Why was this file deleted?

MaxText/pyconfig.py
```diff
@@ -147,6 +147,9 @@ def validate_model_name(s: str) -> bool:
       "llama3-70b",
       "mistral-7b",
       "mixtral-8x7b",
+      "gemma2-2b",
+      "gemma2-9b",
+      "gemma2-27b"
```
gagika, line 152: Can you move it below gemma-* ?
Also there is gemma2-9B in the list.

MaxText/scratch_code/golden_gemma2-2b_export-flax.ipynb
```diff
@@ -0,0 +1,331 @@
+{
```
gagika, line 1: Do we want to open source all the notebooks?

gagika (commented):
Some quick comments

end_to_end/tpu/gemma2/9b/1_test_gemma.sh
ZhaoyueCheng: updated, I think it got deleted somehow during the rebase

end_to_end/tpu/gemma2/9b/2_test_gemma.sh
ZhaoyueCheng: updated, I think it got deleted somehow during the rebase

MaxText/configs/models/gemma2-9b.yml
ZhaoyueCheng: updated

MaxText/pyconfig.py
ZhaoyueCheng: updated

MaxText/configs/base.yml
ZhaoyueCheng: updated

MaxText/scratch_code/golden_gemma2-2b_export-flax.ipynb
ZhaoyueCheng: I think we had the original [gemma export notebook](https://github.com/google/maxtext/blob/main/MaxText/scratch_code/golden_gemma-2b_export.ipynb) and other notebooks (llama, mixtral, etc) open sourced so added the gemma2 notebook to scratch_code folder as well with other open source notebooks

salrowili:
Please note that if we merge local_sliding_attention, global_attention and assigned half the decoder layer to "base_num_decoder_layers", the TFLOPS calculation will report half the actual TFLOPS . This is due to how maxtext_utils calcuate the TFLOPS
https://github.com/google/maxtext/blob/644eb87ae90dd8b210ce17f1c16ca7a54e80fceb/MaxText/maxtext_utils.py#L139

gobbleturk:
> Please note that if we merge local_sliding_attention, global_attention and assigned half the decoder layer to "base_num_decoder_layers", the TFLOPS calculation will report half the actual TFLOPS . This is due to how maxtext_utils calcuate the TFLOPS
> 
> https://github.com/google/maxtext/blob/644eb87ae90dd8b210ce17f1c16ca7a54e80fceb/MaxText/maxtext_utils.py#L139

Good catch! Thank you for noting this

ZhaoyueCheng:
> Please note that if we merge local_sliding_attention, global_attention and assigned half the decoder layer to "base_num_decoder_layers", the TFLOPS calculation will report half the actual TFLOPS . This is due to how maxtext_utils calcuate the TFLOPS
> 
> https://github.com/google/maxtext/blob/644eb87ae90dd8b210ce17f1c16ca7a54e80fceb/MaxText/maxtext_utils.py#L139

updated, thanks for the comment!

MaxText/layers/gemma2.py
```diff
@@ -0,0 +1,246 @@
+"""
+Copyright 2023 Google LLC
```
khatwanimohit, line 2: nit: Update the copyright to 2024

MaxText/pyconfig.py
```diff
@@ -149,7 +149,9 @@ def validate_model_name(s: str) -> bool:
       "mixtral-8x7b",
       "gemma-7b",
       "gemma-2b",
+      "gemma2-2b",
       "gemma2-9b",
+      "gemma2-27b",
```
khatwanimohit, line 154: nit: Should we remove this key for now until we add the gemma2-27b config yml file ? Just don't want people to assume gemma-27b is supported now.

khatwanimohit (approved):
Thanks for adding this! Left a few nits. 

QQ: Are we going to add logit checker for Gemma9B in a separate PR? Since you have the code for golden logits here, can you also add golden logits jsonl file for Gemma2-9B in this PR

MaxText/pyconfig.py
ZhaoyueCheng: added gemma2-27b config yaml file, thanks for the review!

MaxText/layers/gemma2.py
ZhaoyueCheng: updated, thanks for the review!

MaxText/maxtext_utils.py
```diff
@@ -139,6 +139,11 @@ def calculate_tflops_training_per_device(config, log=True):
       attention_flops * config.num_decoder_layers * 3 / 10**12
   )
 
+  # in gemma2 we combine [local attention, global attention] into one decoder block, so need to multiply the
```
gobbleturk, line 142: Don't we also need to multiply the FF by 2 as well since there are 2 FF per gemma2 layer? I'm not sure the best way to handle this, I would rather not have decoder-block specific logic, however here it is necessary. Can we move it to its own subfunction, e.g. "if config.decoer_Block == gemma2: return calculate_gemma2_flops(config)
else:
calculate standard_flops(config)"

If you can think of a better solution let me know

MaxText/pyconfig.py
gobbleturk: Have we tested golden logits for the gemma2-27b as well?

gobbleturk (changes requested):
Overall looks very good, thank you for adding this! Just want to clean up the flops calculation.

MaxText/maxtext_utils.py
gobbleturk: I think we should have calculate_gemma2_flops defined inside of the gemma2 decoder block definition actually, and call it from max_utils. Ideally some inheritance with a default method inherited that works for most models, but can be overriden, we max_utils will call decoder_block.calculate_tflops()...

We can do in the inheritance later, but for this PR we should at least calculate the flops accurately ideally in its own method in either max_utils or the gemma2 decoder block

MaxText/pyconfig.py
ZhaoyueCheng: added the 2b and 9b, for 9b I tested it on `JAX_PLATFORMS=cpu` with `dtype=float32` locally and it works, for 27b it OOMs even on CPU with scanned checkpoint so removing it from the `pyconfig` file for now

MaxText/maxtext_utils.py
ZhaoyueCheng: Thanks for the catch! while refactoring I found since most of the code in the new `calculate_gemma2_flops` are the same I use a local variable `num_decoder_layers` and multiply it by 2 when using the `gemma2` decoder block, please let me know if this works, just updated the branch on this as well, thanks again!

gobbleturk: doesn't local/sliding attention use less flops than global attention? I think thats the main reason gemma2 uses sliding attention?

ZhaoyueCheng: yeah haven't factored that in since we are using `dot_product` which still computes the full attention, will update to  `calculate_gemma2_flops` to include the sliding window attention flops calculation, thanks for the comment!

gobbleturk: calculate_flops is meant to calculate the necessary flops for the model to train, not the amount of flops it actually uses. This allows it to be used to measure MFU

ZhaoyueCheng: Thanks again for the details! updated the calculation to include sliding window attention for gemma2 and factor in the 2 stacked decoder layer in ffn layers

gobbleturk (approved):
Awesome, thank you for fixing the tflops calculation!

ZhaoyueCheng:
> Awesome, thank you for fixing the tflops calculation!

Thanks for the detailed review and suggestions!!

salrowili:
@ZhaoyueCheng I would like to thank you for adding Gemma 2 support which really is appreciated. 

I have pre-trained Gemma 2 using maxtext but now i stuck since i could not convert MaxText to hugging face format to do the SFT stage. Gemma2, unlike Llama, Gemma and Mistral, uses local and global attention . Is there any thing you can do to "MaxText/llama_or_mistral_ckpt.py" script to include Gemma 2 checkpoint conversion to HF format? The global and local attention used in the following weights: 
```
mlp_global
mlp_local
post_ffw_norm_global
post_ffw_norm_local
post_self_attention_norm_global
post_self_attention_norm_local
pre_ffw_norm_global
pre_ffw_norm_local
pre_self_attention_norm_global
pre_self_attention_norm_local
self_attention_global
```

This related to https://github.com/google/maxtext/issues/829

PR #817, opened 2024-08-08
documenting XLA flags used by MaxText

nhira:
Added the XLA Flags that MaxText uses to the README


tonyjohnchen (approved):
Thanks for adding the XLA flags info!

gobbleturk (approved):
Thank you for adding this documentation!

I'm not sure how we can maintain this list as we add more configs and potentially update flags on old ones, but I think this beats no ducmentation

nhira:
Sorry folks -- I messed this PR up with manual merges.  I'll resubmit a new one.

PR #821, opened 2024-08-12
new features with distributed training framework

bernardhan33:
## Features

1. Support of modeling `per_step_interval`.
2. Support of modeling `max_steps`.
3. Support of listing the dataset directory instead of hard-coding the bucket name from the run flags.

(1) and (2) are coming from the [design doc](https://docs.google.com/document/d/1YC9YuHaFpeTQ_5NXCfRfbr3j9y9w-JKcDiHQBSccKKI/edit?tab=t.0#heading=h.m86zlajb3moj). (3) is a bug identified from a conversation with the HNS team that the current hard-coded values prevent the benchmark from easily run against different bucket names.

Internal CL to update the README and the yaml file: cl/662264972.

## Tested by

1. Setting `per_step_interval` to 1 second and confirming that the per step time is roughly 1 second with the exception of those steps whose data loading time takes longer.
2. Setting `max_steps` to make sure that the training can stop gracefully after the global step is met, and that the metrics are recorded correctly.

## Next steps

1. Once this PR and the CL are merged, I'll build the image and upload to gcr.io/gcs-tess/distributed_pytorch_training_benchmark.
2. Once we start to move this to the new simpler framework, I'll add unit tests to cover the various features.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -363,15 +363,15 @@ local_batch_size: 32
 prefetch_factor: 2
 # Number of data loader workers (per pod).
 data_loader_num_workers: 2
-# Number of seconds to sleep in between steps. This will not block prefetching.
-per_step_computation_time: 0.0
+# Number of seconds that represents a step interval. Default to 1 second.
+per_step_interval: 1.0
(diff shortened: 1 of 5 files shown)


MattIrv (commented):
looks good to me

PR #825, opened 2024-08-13
Removing the resgistration of the proxy backend used by Pathways.

lukebaumann:
If needed, this is handled by a separate, Pathways specific utilities package.


JoeZijunZhou (approved):

gobbleturk (approved):
Thank you!

PR #830, opened 2024-08-19
Add support for local sliding window attention in TPU splash_attention

gagika:
Added support for sliding window attention masking in TPU splash_attention


khatwanimohit (approved):

gobbleturk (approved):
Awesome, thank you!

PR #833, opened 2024-08-20
Add dropping strategy

RissyRan:
# Description

Enable token dropping for matmul implementation

Next steps:
* work on dispatch & combine mask - make expert parallelism easy
* add load balance loss

# Test

Have a unite test to check a single layer output (become dropless if capacity factor is large enough):

* capacity factor = 2 as default (it should be 1 ideally with evenly distributed) - [link](https://paste.googleplex.com/5898720970539008#l=12) - pass
* capacity factor = 0.1 (extreme case) - [link](https://paste.googleplex.com/6179650651029504) - fail

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -103,15 +103,17 @@ base_num_kv_heads: 16
 base_mlp_dim: 7168
 base_num_decoder_layers: 16
 head_dim: 128
-# mixture of experts (moe)
-num_experts: 1
-num_experts_per_tok: 1
-megablox: True
 mlp_activations: ["silu", "linear"]
 dropout_rate: 0
 logits_via_embedding: False
 normalize_embedding_logits: True  # whether to normlize pre-softmax logits if logits_via_embedding is true
 logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedding dot product for stability
+
+# mixture of experts (moe)
+num_experts: 1
+num_experts_per_tok: 1
+megablox: True
+capacity_factor: -1 # a factor to decide expert capacity for token dropping, and no dropping by default
 
 # pipeline parallelism
 # The number of decoder layers is equal to the product of num_stages, num_layers_per_pipeline_stage and num_pipeline_repeats.
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -335,7 +335,7 @@ class MoeBlock(nn.Module):
     inputs_shape = inputs.shape
     inputs_2d = jnp.reshape(inputs, (inputs_shape[0] * inputs_shape[1], inputs_shape[2]))
     weights, selected_experts = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
-    weights = jax.nn.softmax(weights.astype(self.weight_dtype), axis=-1).astype(self.dtype)
+    weights = jax.nn.softmax(weights.astype(jnp.float32), axis=-1).astype(self.dtype)
     flatten_selected_experts = jnp.ravel(selected_experts)
     sorted_selected_experts = jnp.argsort(flatten_selected_experts)
     sorted_indices = sorted_selected_experts // self.num_experts_per_tok
@@ -409,12 +409,40 @@ class MoeBlock(nn.Module):
       return output
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
-  def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
-    top_k_weights, top_k_indices = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
-    flattened_top_k_weights = top_k_weights.reshape(-1, self.num_experts_per_tok)
+  def generate_masks(self, top_k_indices):
+      # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
+      batch_size, seq_len, _ = top_k_indices.shape
+      tokens_per_batch = batch_size * seq_len * self.num_experts_per_tok
+      expert_capacity = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+      max_logging.log(f"Applying potential token dropping with an expert_capacity of {expert_capacity}")
 
-    softmax_probs = jax.nn.softmax(flattened_top_k_weights.astype(jnp.float32), axis=-1).astype(self.weight_dtype)
-    softmax_probs = softmax_probs.reshape(gate_logits.shape[:-1] + (self.num_experts_per_tok,))
+      # calculate expert mask and drop tokens if needed
+      # shape of output expert mask: (batch, sequence, num_experts_per_tok)
+      # 
+      # A small example:
+      # give num_experts=4 & num_experts_per_tok=2, and two tokens are routed to expert [0, 1] & [1, 3],
+      # then expert_mask becomes [[[[1, 0, 0, 0],[0, 1, 0, 0]], [[0, 1, 0, 0],[0, 0, 0, 1]]]],
+      # after cumsum, expert_token_count becomes [[[[1, 0, 0, 0],[1, 1, 0, 0]], [[1, 2, 0, 0],[1, 2, 0, 1]]]],
+      # if we set expert_capacity=1, 
+      # trunc_expert_mask becomes [[[[1, 0, 0, 0],[0, 1, 0, 0]], [[0, 0, 0, 0],[0, 0, 0, 1]]]],
+      # so the 2nd token for expert #1 ([0, 1] & [1, 3]) is dropped, output of updated_expert_mask is [[[1, 1],[0, 1]]].
+      expert_mask = jax.nn.one_hot(top_k_indices, num_classes=self.num_experts, dtype=jnp.int32)
+      reshaped_expert_mask = jnp.reshape(expert_mask, (batch_size, seq_len * self.num_experts_per_tok, self.num_experts))
+      expert_token_count = jnp.cumsum(reshaped_expert_mask, axis=1)
+      expert_token_count = jnp.reshape(expert_token_count, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
+      trunc_expert_mask = expert_mask * jnp.less_equal(expert_token_count, expert_capacity)
+      updated_expert_mask = jnp.sum(trunc_expert_mask, axis=3)
+      return updated_expert_mask
+
+  def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
+    # shape of top_k_weights & top_k_indices: (batch, sequence, num_experts_per_tok)
+    top_k_weights, top_k_indices = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
+    softmax_probs = jax.nn.softmax(top_k_weights.astype(jnp.float32), axis=-1).astype(self.dtype)
+
+    # token dropping if needed
+    if self.config.capacity_factor > 0:
+      expert_mask = self.generate_masks(top_k_indices)
+      softmax_probs *= expert_mask
 
     weights = jnp.zeros_like(gate_logits)
     index_update = (jnp.arange(gate_logits.shape[0])[:, None, None], jnp.arange(gate_logits.shape[1])[:, None], top_k_indices)


ZhiyuLi-goog:
Thank you @RissyRan for adding dropping strategy!!
I just added some nit.

tonyjohnchen (approved):
Thanks for adding this!

RissyRan:
> Thank you @RissyRan for adding dropping strategy!! I just added some nit.

Thanks Zhiyu! Have you published comments?

ZhiyuLi-goog:
> > Thank you @RissyRan for adding dropping strategy!! I just added some nit.
> 
> Thanks Zhiyu! Have you published comments?

Just published.

MaxText/configs/base.yml
```diff
@@ -103,16 +103,18 @@ base_num_kv_heads: 16
...
 mlp_activations: ["silu", "linear"]
 dropout_rate: 0
 logits_via_embedding: False
 normalize_embedding_logits: True  # whether to normlize pre-softmax logits if logits_via_embedding is true
 logits_dot_in_fp32: True  # whether to use fp32 in logits_dense or shared_embedding dot product for stability
 
+# mixture of experts (moe)
+num_experts: 1
+num_experts_per_tok: 1
+megablox: True
```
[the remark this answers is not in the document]
RissyRan: I don't think it's conflict. But the merit of megablox is dropless. So in this case, I plan to have 2 implementations:

1) dropless through megablox;
2) dropping through matmul (dropless if capacity factor is large enough)

MaxText/layers/linears.py
```diff
@@ -410,12 +410,26 @@ def wrapper(x, logits, w0, w1, wo):
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
   def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
+    # shape of top_k_weights & top_k_indices: (batch, sequence, exp_per_token)
     top_k_weights, top_k_indices = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
-    flattened_top_k_weights = top_k_weights.reshape(-1, self.num_experts_per_tok)
```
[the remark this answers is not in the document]
RissyRan: What's the benefit of dropless in matmul when we have megablox implementation?

MaxText/layers/linears.py
```diff
@@ -410,12 +410,26 @@ def wrapper(x, logits, w0, w1, wo):
...
+    # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
+    batch_size, seq_len, _ = inputs.shape
+    tokens_per_batch = batch_size * seq_len * self.num_experts_per_tok
+    expert_capacity = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+    max_logging.log(f"expert_capacity is {expert_capacity}")
+
+    # calculate expert mask and drop tokens if needed
+    # shape of expert_mask: (batch, sequence, exp_per_token, num_experts)
+    expert_mask = jax.nn.one_hot(top_k_indices, num_classes=self.num_experts)
+    cumsum_expert_mask = jnp.cumsum(jnp.cumsum(expert_mask, axis=1), axis=2)
```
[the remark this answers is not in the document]
RissyRan: This is a good question. The order should be kept. `jnp.cumsum` will not change the array shape. Right after this line we have `cumsum_expert_mask = cumsum_expert_mask * expert_mask`, that could map back the original one-hot. I could print an example if that helps.

MaxText/layers/linears.py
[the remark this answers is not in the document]
ZhiyuLi-goog: Thank you @RissyRan. I think your dropless in matmul is helpful in mlperf. Additionally, dense_matmul droppless implementation is equivalent to mixtral reference implementation and huggingface's. We can use this as a baseline for performance and numerical stability. What do you think?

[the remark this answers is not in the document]
RissyRan: >dense_matmul droppless implementation is equivalent to mixtral reference implementation and huggingface's

OK, I see. I recall both of them are using for loop implementation. If we want to keep it as a reference, I can add a if/else somewhere with indicator, i.e. capacity_factor, or bool value.

MaxText/layers/linears.py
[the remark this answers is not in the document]
RissyRan: Order will not be updated, we only mask weights when setting `softmax_probs`

```
    weights = jnp.zeros_like(gate_logits)
    index_update = (jnp.arange(gate_logits.shape[0])[:, None, None], jnp.arange(gate_logits.shape[1])[:, None], top_k_indices)
    weights = weights.at[index_update].set(softmax_probs)
```


No dropping example:

```
# select expert 3 & 4
top_k_indices: [[[3 4]]]

# convert to one-hot
expert_mask: [[[[0. 0. 0. 1. 0. 0. 0. 0.]
   [0. 0. 0. 0. 1. 0. 0. 0.]]]]

# those two are selected
flat_expert_mask: [[[1. 1.]]]

# softmax_probs
softmax_probs: [[[0.5019531  0.49804693]]]
```

Dropping example:

```
top_k_indices: [[[3 4]
  [4 5]
  [4 6]
  [6 4]]]

expert_mask: [[[[0. 0. 0. 1. 0. 0. 0. 0.]
   [0. 0. 0. 0. 1. 0. 0. 0.]]

  [[0. 0. 0. 0. 1. 0. 0. 0.]
   [0. 0. 0. 0. 0. 1. 0. 0.]]

  [[0. 0. 0. 0. 1. 0. 0. 0.]
   [0. 0. 0. 0. 0. 0. 1. 0.]]

  [[0. 0. 0. 0. 0. 0. 1. 0.]
   [0. 0. 0. 0. 1. 0. 0. 0.]]]]

flat_expert_mask: [[[1. 1.]
  [1. 1.]
  [1. 1.]
  [1. 0.]]]

softmax_probs: [[[0.5019531  0.49804693]
  [0.5602527  0.43974727]
  [0.5893618  0.4106382 ]
  [0.58699644 0.        ]]]

```

MaxText/layers/linears.py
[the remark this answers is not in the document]
RissyRan: Updated, and no dropping by default.

MaxText/configs/base.yml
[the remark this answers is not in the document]
gobbleturk: an intermediate step might be expert parallelism + dropping + megablox, although an ideal solution could be dropless. E.g. if there are 64 experts and we set expert parallelism = 16 there are still 4 experts per shard, so the megablox kernel can still be useful. It seems easier to perform the expert parallelism a2a with dropping

MaxText/layers/linears.py
```diff
@@ -410,11 +410,28 @@ def wrapper(x, logits, w0, w1, wo):
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
   def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
+    # shape of top_k_weights & top_k_indices: (batch, sequence, exp_per_token)
     top_k_weights, top_k_indices = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
-    flattened_top_k_weights = top_k_weights.reshape(-1, self.num_experts_per_tok)
-
-    softmax_probs = jax.nn.softmax(flattened_top_k_weights.astype(jnp.float32), axis=-1).astype(self.weight_dtype)
-    softmax_probs = softmax_probs.reshape(gate_logits.shape[:-1] + (self.num_experts_per_tok,))
+    softmax_probs = jax.nn.softmax(top_k_weights.astype(jnp.float32), axis=-1).astype(self.weight_dtype)
```
gobbleturk, line 415: Instead of weight_dtype shouldn't this be activation dtype?

MaxText/layers/linears.py
```diff
@@ -410,11 +410,28 @@ def wrapper(x, logits, w0, w1, wo):
...
+
+    # token dropping
+    if self.config.capacity_factor > 0:
+      # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
+      batch_size, seq_len, _ = inputs.shape
+      tokens_per_batch = batch_size * seq_len * self.num_experts_per_tok
+      expert_capacity = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+      max_logging.log(f"Token dropping if needed, and expert_capacity is {expert_capacity}")
```
gobbleturk, line 423: nit: "Applying potential token dropping with an expert_capacity of {expert_capacity}"

MaxText/layers/linears.py
```diff
@@ -410,11 +410,28 @@ def wrapper(x, logits, w0, w1, wo):
...
+    # shape of top_k_weights & top_k_indices: (batch, sequence, exp_per_token)
     top_k_weights, top_k_indices = jax.lax.top_k(gate_logits, self.num_experts_per_tok)
-    flattened_top_k_weights = top_k_weights.reshape(-1, self.num_experts_per_tok)
-
-    softmax_probs = jax.nn.softmax(flattened_top_k_weights.astype(jnp.float32), axis=-1).astype(self.weight_dtype)
-    softmax_probs = softmax_probs.reshape(gate_logits.shape[:-1] + (self.num_experts_per_tok,))
+    softmax_probs = jax.nn.softmax(top_k_weights.astype(jnp.float32), axis=-1).astype(self.weight_dtype)
+
+    # token dropping
+    if self.config.capacity_factor > 0:
```
gobbleturk, line 418: Can we wrap the body of this in its own method - for now the method just needs to return the softmax_probs, perhaps the mask as well once we integrate this will Expert parallelism

MaxText/layers/linears.py
```diff
@@ -410,11 +410,28 @@ def wrapper(x, logits, w0, w1, wo):
...
+      # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
+      batch_size, seq_len, _ = inputs.shape
+      tokens_per_batch = batch_size * seq_len * self.num_experts_per_tok
+      expert_capacity = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+      max_logging.log(f"Token dropping if needed, and expert_capacity is {expert_capacity}")
+
+      # calculate expert mask and drop tokens if needed
+      # shape of expert_mask: (batch, sequence, exp_per_token, num_experts)
+      expert_mask = jax.nn.one_hot(top_k_indices, num_classes=self.num_experts)
+      cumsum_expert_mask = jnp.cumsum(jnp.cumsum(expert_mask, axis=1), axis=2)
```
gobbleturk, line 428: Comments here would be helpful although it is tricky how to best explain this concisely (The code is more concise than words). I prefer the name "expert_cumusm_token_count" instead of cumsum_expert_mask for line 428, and maybe expert_token_count for line 429 (open to other ideas though). Perhaps a small example - "If expert 0 is chosen by tokens [1,3] the expert_token_count will have entries [0,1,0,2], which associates each chosen token with an increasing count"

MaxText/layers/linears.py
gobbleturk: oh I see you are not changing the dtype in this PR. But I think the dtype manipulation here and also here https://github.com/google/maxtext/blob/748305d5bd60d326c98d00b2274515f20b5968d9/MaxText/layers/linears.py#L338 should be changed to cast to float32 before softmax, then back to activation dtype. Perhaps the float32 cast should be configurable

MaxText/layers/linears.py
RissyRan: SG, will update.

MaxText/layers/linears.py
RissyRan: SG! Will update the variable name and make a comment there.

MaxText/layers/linears.py
RissyRan: Sg!

RissyRan: Moved to generate_masks() helper method.

MaxText/layers/linears.py
RissyRan: Updated the variables, and have a small example there. Please let me know if it's unclear.

ZhiyuLi-goog (approved):
Thank you @RissyRan LGTM!

gobbleturk (approved):
Thanks Ran!

PR #834, opened 2024-08-20
Integrate Badput monitoring with MaxText

dipannita08:
Integrate Badput Breakdown with MaxText

This change adds recording of additional Badput information and integrating with GoodputMonitor's Badput breakdown support.

Tested:
  - [x] Example run on v4-8 w/ ~500 steps [here](https://screenshot.googleplex.com/6FL2A8patsi5TQm)

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -416,14 +416,15 @@ def create_goodput_recorder(config):
   return None
 
 
-def record_goodput(recorder, config, step=None, job_start=False, job_end=False):
+def record_goodput(
+      recorder,
+      config,
+      record_func,
+      *args,
+  ):
+  """Record data for Goodput and Badput computation."""
   if recorder and config.enable_goodput_recording:
-    if job_start and step is None:
-      recorder.record_job_start_time()
-    if job_end and step is None:
-      recorder.record_job_end_time()
-    if step is not None:
-      recorder.record_step_start_time(step)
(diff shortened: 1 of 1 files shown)


MaxText/train.py
```diff
@@ -417,14 +417,39 @@ def create_goodput_recorder(config):
   return None
 
 
-def record_goodput(recorder, config, step=None, job_start=False, job_end=False):
+def record_goodput(
```
gobbleturk, line 420: this API looks pretty funky to me. It seemed okay with only two options, start and end, but with so many options it doesn't make sense to me to have 10 boleans where exactly one is true. I couldn't think of a great alternative due to None checking...  How about

 
```
 def record_goodput(recorder, config, record_func, *args):
     if recorder and config.enable_goodput_recording:
               record_func(*args)
               
# Example call
 record_goodput(recorder,config, recorder.tpu_init_end if recorder else None)
```

Alternatively can avoid None checking with get_attr, which is probably less preferred
```
def record_goodput(recorder, config, record_method, *args):
     if recorder and config.enable_goodput_recording:
               getattr(recorder, record_method)(*args)

 # Example call             
 record_goodput(recorder,config,"tpu_init_end")
```

dipannita08: Thanks, done.

gobbleturk (approved):

PR #835, opened 2024-08-20
Adds ragged attention.

patemotter:
Adds Ragged Attention Pallas kernel as an option when performing autoregressive attention. By default this is disabled and does not interfere with the existing AR attention. It can be enabled by the CLI argument `use_ragged_attention=true`. 

Improvements are most noticeable when 
* `quantize_kvcache=false`
* `ar_cache_axis_order = prefill_cache_axis_order = "0,2,1,3"`
* Performance improves as `max_prefill_predict_length` and `max_target_length` increase.

### DIFF
--- a/MaxText/kernels/ragged_attention.py
+++ b/MaxText/kernels/ragged_attention.py
@@ -1,0 +1,437 @@
+"""
+Copyright 2024 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Kernels for ragged attention."""
+
+import functools
+
+import jax
+from jax import lax
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+import jax.numpy as jnp
+import numpy as np
+import common_types
+
+from jax.experimental import shard_map
+
+
+BATCH = common_types.BATCH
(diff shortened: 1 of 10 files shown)


MaxText/kernels/ragged_attention.py
```diff
@@ -0,0 +1,302 @@
...
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+import jax.numpy as jnp
+import numpy as np
+import common_types
+
+from jax.experimental import shard_map
+
+
+DEFAULT_MASK_VALUE = -0.7 * float(np.finfo(np.dtype("float32")).max)
```
gobbleturk, line 16: Include comment here why large negative value is the appropriate filler as opposed to zero (value corresponds to logit, will be softmaxed)

MaxText/kernels/ragged_attention.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+@functools.partial(
+  jax.jit,
+  static_argnames=[
+    "block_size",
+    "mask_value",
+  ],
+)
+def ragged_mha(
```
gobbleturk, line 244: I only see mqa and mha, is ragged gqa also supported?

MaxText/tests/kernels_test.py
gobbleturk, line 1: Awesome, thank you for including this test!

gobbleturk (approved):
Looks great to me, this is very cool!

MaxText/kernels/ragged_attention.py
patemotter: This is already used in attentions.py and mentioned elsewhere in the codebase. So I was just reusing what was already established. But we can add some info to it either here or move it to some common place so it's not redefined.

MaxText/kernels/ragged_attention.py
patemotter: GQA was not in the original PR but is now included.

MaxText/kernels/ragged_attention.py
patemotter: Latest revision moves `DEFAULT_MASK_VALUE` to be defined in `common_types.py` and I added a small comment about why it's used.

PR #838, opened 2024-08-21
Llama3.1 (8B,70B,405B) 🦙

khatwanimohit:
Along with adding Llama3.1 configs, this PR has following changes:
1) remove dependency of train state from forward_pass_logit_checker. This helps in running forward Llama3.1 405B on v4-128 for internal testing
2) Introduce sgd as optimizer
3) Fix llama_or_mistral_ckpt.py to accomodate Llama3.1 Rope changes.


end_to_end/tpu/llama3.1/70b/2_test_llama3.1_70b.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# This file is both an integration test that runs once a day on a v4-16 and documentation for how to get started with Llama3.1-70b.
```
gobbleturk, line 3: I don't think llama 70b fits on a v4-16, should this be v4-128?

gobbleturk (approved):
Awesome! Left one request for fixing a comment about more chips needed llama3-70b

peregilk:
Wouldnt also Rope Theta scaling need to be implemented for Llama 3.1 to work correctly? Like done in HF: https://huggingface.co/meta-llama/Meta-Llama-3.1-8B/blob/48d6d0fc4e02fb1269b36940650a1b7233035cbb/config.json#L21.

Or am I missing something here?

peregilk:
@khatwanimohit I am testing the script for converting the Meta-checkpoints. Everything looks fine. However, for some reason the file `scanned_chkpt/0/items/checkpoint` is not written. This seems just to be a file for the state. The model weights seems to be stored in the bucket.

UPDATE: This seems to be just a status file, and since this is at checkpoint 0, it does not seem to matter. I can manually copy the file from Llama3, to fix this.

kocchop:
Hi, what's the eta of the PR? Wanted to test the models on MaxText

peregilk:
@khatwanimohit Any chance to get in support for the Llama 3.2 1B and 3B as well? From what I have read there are not really any major architectural changes to the 3.2 1B and 3B. Shouldnt this be just updating the config files like this? 

Llama 3.2 1B:
```
base_emb_dim: 2048
base_num_query_heads: 32
base_num_kv_heads: 8
base_num_decoder_layers: 16
base_mlp_dim: 8192
head_dim: 64
vocab_size: 128256
enable_dropout: False
logits_via_embedding: True
normalization_layer_epsilon: 1.0e-5
rope_max_timescale: 500_000
decoder_block: "llama2" # Uses the same decoder block as llama2
````

Llama 3.2 3B:
```
base_emb_dim: 3072
base_num_query_heads: 24
base_num_kv_heads: 8
base_num_decoder_layers: 28
base_mlp_dim: 8192
head_dim: 128
vocab_size: 128256
enable_dropout: False
logits_via_embedding: True
normalization_layer_epsilon: 1.0e-5
rope_max_timescale: 500_000
decoder_block: "llama2" # Uses the same decoder block as llama2
```

gagika (approved):
Thanks Mohit!

khatwanimohit:
> @khatwanimohit Any chance to get in support for the Llama 3.2 1B and 3B as well? From what I have read there are not really any major architectural changes to the 3.2 1B and 3B. Shouldnt this be just updating the config files like this?
 Yes, we still have to do some testing just to make sure the logit matches. Once testing is done for Llama3.2 then I will create a follow-up PR for adding the config.

peregilk:
@khatwanimohit Do you know the status for the official llama 3.2 code? I tried upgrading with as described above (including a few other references). It seemed OK, and the 3B model is training. Though I have not verified the result yet.

I did however have an issue with the 1B model. I got a "not implemented"-error with flash attention, saying it required layers divisible by 128 (and the head_dim = 64 in the 1B model). I have not looked into how hard it is to fix this. Do you know if there are plans to do so?

jiagaoxiang:
Any update on supporting Llama3.2 vision models as well?

PR #839, opened 2024-08-23
[mlperf/4.1] enable shard_in_read for large scaling training

ZhiyuLi-goog:


ZhiyuLi-goog:
@anfals could you take a look. We need shard_in_read for v5p-12288 run.

anfals (approved):

PR #840, opened 2024-08-23
Mlperf/4.1 grain

aireenmei:
* Use a custom grain wheel (Need to rebuild dependency image, see changes in requirements.txt and setup.sh)
* Add grain support for per_device_batch_size<1 while eval_per_device_batch_size>1
* Run on v5p-128 using xpk (per_device_batch_size=0.5, eval_per_device_batch_size=1), using the script in this PR. command: bash MaxText/configs/v5e/mlperf-grain-tpu.sh PER_DEVICE_BATCH_SIZE=0.5 GRAIN_WORKER_COUNT=1 ICI_TENSOR=16 PLATFORM=gke 
Results: https://cloudlogging.app.goo.gl/5BpmD9iCEm7uVGNx5 
* The padding batch for eval doesn't work well. Suggest to first test with eval off (eval_interval=-1), then try eval on. Eval may not terminate but hang on some slice sizes. I'll continue working on that.

### DIFF
--- a/MaxText/configs/v5e/mlperf-grain-tpu.sh
+++ b/MaxText/configs/v5e/mlperf-grain-tpu.sh
@@ -1,0 +1,157 @@
+echo "Running mlperf-grain-tpu.sh"
+# 16B parameter model.
+# This config will work out of the box for any number of v5e-256 slices.
+#
+# Command Flags:
+# OUTPUT_PATH (Required, unless base_output_directory is already set in base.yml)
+# DATASET_PATH (Required, unless dataset_path is already set in base.yml)
+# RUN_NAME (Required, unless run_name is already set in base.yml or running with XPK/GKE)
+# PLATFORM (Optional, can be "gke" or "gce", default is "gce")
+#
+# Example to invoke this script:
+# bash MaxText/configs/v5e/16b.sh RUN_NAME="<your_run_name>" OUTPUT_PATH="gs://<your_output_path>" DATASET_PATH="gs://<your_dataset_path>" PLATFORM="gke"
+#
+# Example to AOT compile:
+# bash MaxText/configs/v5e/16b.sh EXECUTABLE=train_compile.py M_COMPILE_TOPOLOGY=v5e-256 M_COMPILE_TOPOLOGY_NUM_SLICES=2
+
+
+# Stop execution if any command exits with error
+set -e -x
+
+export PLATFORM="gce"
+export EXECUTABLE="train.py" # or train_compile.py
+
+# Set environment variables
+for ARGUMENT in "$@"; do
+    IFS='=' read -r KEY VALUE <<< "$ARGUMENT"
+    export "$KEY"="$VALUE"
+done
+
+# The setup accommodates two cases:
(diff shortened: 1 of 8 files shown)


aireenmei:
The padding batch issue for eval is fixed now. To use this feature, set eval_steps to the just enough to finish the eval data (the data iter will generate empty examples until eval_steps met). For mlperf eval data, set eval_steps = math.ceil(5700/global_batch_size) and check if the output total_weights==11590004. Add more eval_steps if needed for total_weights to reach that value.

MaxText/configs/v5e/mlperf-grain-tpu.sh
gobbleturk, line 1: What is the purpose of this file? Does it serve as documentation for running mlperf + grain? Does it serve as a test + documentation? If so it should probalby live under end_to_end instead of MaxText/configs

MaxText/pyconfig.py
```diff
@@ -375,8 +375,8 @@ def configure_gpt3_task(raw_keys):
       if raw_keys['overwrite_ckpt_step'] == -1:
         raw_keys['overwrite_ckpt_step'] = math.ceil(4000.0 * 1536 / global_batch_size)
     global_batch_size_to_train_on = calculate_global_batch_sizes(raw_keys)[1]
-    if raw_keys["dataset_type"] != "synthetic":
-      raw_keys["eval_interval"] = math.ceil(24567 / global_batch_size_to_train_on)
+    # if raw_keys["dataset_type"] != "synthetic":
```
gobbleturk, line 378: Delete instead of comment out?

requirements.txt
```diff
@@ -9,7 +9,7 @@ cloud-tpu-diagnostics
 datasets
 gcsfs
 google-cloud-storage
-grain-nightly
+#grain-nightly
```
gobbleturk, line 12: Delete instead of comment

setup.sh
```diff
@@ -190,4 +190,6 @@ if [[ "$MODE" == "pinned" ]]; then
     pip3 install -U -r requirements.txt -c constraints_gpu.txt
 else
     pip3 install -U -r requirements.txt
+    gsutil cp gs://mlperf-exp-us-east1-cp0/grain-wheel/grain-0.2.2-cp310-cp310-linux_x86_64.whl ./
```
gobbleturk, line 193: This concerns me a bit since we won't be up to date with grain without manual effort. How come this is needed - maybe include a comment?

gobbleturk (approved):
Looks great aireen, thanks! One major Q is with the new config, if its location is intentional or should be moved / replaced by other documentation

setup.sh
aireenmei: Sorry for the lack of context. For grain to be used for mlperf, there's a feature in grain that's still pending CL review and not yet merged. So I built a custom wheel for mlperf use only. But looks like the perf team may not have capacity to test/use grain for this submission (@anfals will make the call). Let me convert this PR to draft for now. I'll make another PR for main branch once the grain feature is officially released

PR #841, opened 2024-08-24
Add dispatch and combine masks for dropping

RissyRan:
# Description

* Add dispatch and combine masks for dropping & expert parallelism.
* Update subscript of einsum() in non-dropping to be consistent
* Next step - add load balance loss for dropping to balance the token distribution.

# Test

A single moe layer unit test to compare before & after reshape on different capacity factors - [one example](https://paste.googleplex.com/6114986465427456).

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -409,12 +409,20 @@ class MoeBlock(nn.Module):
       return output
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
-  def generate_masks(self, top_k_indices):
+  def reshape_and_update_weights(self, weights, indices):
+    # input of weights & indices: (batch_size, seq_len, num_experts_per_tok)
+    # output of updated weights: (batch_size, seq_len, num_experts)
+    update_weights = jnp.zeros((weights.shape[0], weights.shape[1], self.num_experts), dtype=self.dtype)
+    index_update = (jnp.arange(weights.shape[0])[:, None, None], jnp.arange(weights.shape[1])[:, None], indices)
+    update_weights = update_weights.at[index_update].set(weights)
+    return update_weights
+
+  def generate_masks(self, top_k_indices, softmax_probs):
       # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
       batch_size, seq_len, _ = top_k_indices.shape
-      tokens_per_batch = batch_size * seq_len * self.num_experts_per_tok
-      expert_capacity = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
-      max_logging.log(f"Applying potential token dropping with an expert_capacity of {expert_capacity}")
+      tokens_per_batch = seq_len * self.num_experts_per_tok
+      expert_capacity_per_batch = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+      max_logging.log(f"Applying potential token dropping with a batch expert_capacity of {expert_capacity_per_batch}")
 
       # calculate expert mask and drop tokens if needed
       # shape of output expert mask: (batch, sequence, num_experts_per_tok)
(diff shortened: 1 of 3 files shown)


MaxText/layers/linears.py
```diff
@@ -429,37 +436,64 @@ def generate_masks(self, top_k_indices):
       expert_mask = jax.nn.one_hot(top_k_indices, num_classes=self.num_experts, dtype=jnp.int32)
       reshaped_expert_mask = jnp.reshape(expert_mask, (batch_size, seq_len * self.num_experts_per_tok, self.num_experts))
       expert_token_count = jnp.cumsum(reshaped_expert_mask, axis=1)
-      expert_token_count = jnp.reshape(expert_token_count, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
-      trunc_expert_mask = expert_mask * jnp.less_equal(expert_token_count, expert_capacity)
+      reshaped_expert_token_count = jnp.reshape(expert_token_count, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
```
gobbleturk, line 439: would be helpful to have consistent naming - above we have `reshaped_expert_mask` which has seq_len * experts_per_tok, but here we have `reshaped_expert_token_count, which is the opposite (seq_len, experts_per_tok). Maybe expert_token_count_seq_fused, where fused refers to seq_len * experts_per_tok (incluce comment for first *fused variable name)

MaxText/layers/linears.py
```diff
@@ -429,37 +436,64 @@ def generate_masks(self, top_k_indices):
...
-      return updated_expert_mask
+
+      # update weights for all experts
+      softmax_probs *= updated_expert_mask
+      weights = self.update_weights(softmax_probs, top_k_indices)
+
+      # shape of expert_token_position is (batch_size, seq_len * num_experts_per_tok, num_experts)
+      # calculate token position in expert capacity dimension
+      expert_token_position = reshaped_expert_mask * expert_token_count
+      reshaped_expert_token_position = jnp.reshape(expert_token_position, (batch_size, seq_len, self.num_experts_per_tok, self.num_experts))
```
gobbleturk, line 450: Same comment as above for reshaped* named - prefer consistentcy so I can tell if (seq_len * exp/tok) or (seq_len, exp/tok) just from the name - I like using *_seq_fused to refer to the former,e.g. expert_token_position_seq_fused

MaxText/layers/linears.py
```diff
@@ -429,37 +436,64 @@ def generate_masks(self, top_k_indices):
...
-      expert_token_count = jnp.reshape(expert_token_count, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
-      trunc_expert_mask = expert_mask * jnp.less_equal(expert_token_count, expert_capacity)
+      reshaped_expert_token_count = jnp.reshape(expert_token_count, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
+      trunc_expert_mask = expert_mask * jnp.less_equal(reshaped_expert_token_count, expert_capacity_per_batch)
       updated_expert_mask = jnp.sum(trunc_expert_mask, axis=3)
-      return updated_expert_mask
+
+      # update weights for all experts
+      softmax_probs *= updated_expert_mask
+      weights = self.update_weights(softmax_probs, top_k_indices)
```
gobbleturk, line 445: Is this necessary here? Isn't softmax_probs already 0 in every entry not correpsonding to a top_k_index because we just multiplied it with updated_expert_mask?

gobbleturk (changes requested):
Looks great to me, well done! There is a lot of logic here that can be easily messed up - have you been able to test this in some way? Ideally we would set up a unit test but I don't see a candidate API (stable,exposed) to test except the MoE Block with a capacity_factor. Perhaps generate_masks, but that may be more of an internal implementation API that is subject to change

ZhiyuLi-goog (approved):

MaxText/layers/linears.py
```diff
@@ -409,12 +409,19 @@ def wrapper(x, logits, w0, w1, wo):
       return output
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
-  def generate_masks(self, top_k_indices):
+  def update_weights(self, weights, indices):
```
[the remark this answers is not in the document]
RissyRan: Hmm..I may need to have a better name here. Actually what it does is to update weights (i.e. based on dropping result), and shape of weights may change. Let me add some comments here.

MaxText/layers/linears.py
RissyRan: Thanks! Good point! Let me update those variables and try to keep them consistent.

MaxText/layers/linears.py
RissyRan: Yes, this is needed. The update_weights will do 2 things:

1) update weights due to the dropping;
2) reshape the weights from (batch, seq, topk_exp) to (batch, seq, num_exp).

MaxText/layers/linears.py
RissyRan: Thanks! I like this `fused` suffix. Let me do update variable names and try to keep them consistent.

RissyRan:
> Looks great to me, well done! There is a lot of logic here that can be easily messed up - have you been able to test this in some way? Ideally we would set up a unit test but I don't see a candidate API (stable,exposed) to test except the MoE Block with a capacity_factor. Perhaps generate_masks, but that may be more of an internal implementation API that is subject to change

Yes, I tested it before/after the change in one layer, however this test will invalid once the change is merged. Yeah, let me probably add a unit test for generate masks function on a small example.

Update: add a test, and it [passed](https://paste.googleplex.com/5445497583304704).

MaxText/layers/linears.py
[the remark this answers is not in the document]
RissyRan: Will update to something like below:

```
  def reshape_and_update_weights(self, weights, indices):
    # input of weights & indices: (batch_size, seq_len, num_experts_per_tok)
    # output of updated weights: (batch_size, seq_len, num_experts)
...
```
Let me know how do you think :)

MaxText/layers/linears.py
RissyRan: Updated variables with `fused` suffix if the shape like (batch, seq * num_experts_per_tok, ...).

gobbleturk (approved):

MaxText/layers/linears.py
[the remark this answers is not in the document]
gobbleturk: Its better! "Update" is a pretty generic word - how about including as a comment what you told me in a separate convo below. Maybe "apply_drop_mask_and_exp_reshape" for a better name

1. update weights due to the dropping;
2. reshape the weights from (batch, seq, topk_exp) to (batch, seq, num_exp).

PR #845, opened 2024-08-27
Adding Mixtral-8x22b

rdyro:
- adding mixtral-8x22b config (to pyconfig as well)
- improving the llama and mistral conversion script - in-place weight writing to reduce total RAM usage - better progress tracking


end_to_end/tpu/mixtral/8x22b/2_test_mixtral.sh
```diff
@@ -0,0 +1,107 @@
...
+# Test whether the forward pass logits match the golden logits - matmul implementation
+python3 MaxText/tests/forward_pass_logit_checker.py MaxText/configs/base.yml \
+  base_output_directory=${BASE_OUTPUT_PATH} \
+  load_parameters_path=${SCANNED_CHECKPOINT} run_name=matmul_forward_pass_test \
+  per_device_batch_size=1 model_name=mixtral-8x22b \
+  tokenizer_path=${TOKENIZER_PATH} ici_tensor_parallelism=1 \
+  ici_fsdp_parallelism=-1 max_prefill_predict_length=$PREDICT_LEN max_target_length=$PREDICT_LEN \
+  dataset_type=synthetic dtype=bfloat16 weight_dtype=float16 megablox=False \
+  --atol=1e9 --rtol=1e9 --token_size=$PREDICT_LEN
+  # TODO(rdyro): figure out the reason for numerical mismatch for some tokens
```
RissyRan, line 56: let's set up reasonable tolerance? otherwise, the automation tests will fail after you add them [here](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_end_to_end.py).

assets/tokenizer.mixtral-8x22b
ZhiyuLi-goog, line 1: Can we change the name to `mixtral-8x22b-instruct` just to differentiate the tokenizer used in  `Mixtral-8x22B-v0.1`

ZhiyuLi-goog (approved):
Thank you @rdyro, awesome job!
Just have a nit.
LGTM!

assets/tokenizer.mixtral-8x22b
rdyro: Yes, that's a very good idea!

rdyro:
End-to-end tests on [ml-auto-solutions](https://github.com/GoogleCloudPlatform/ml-auto-solutions) are [a PR here](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/368)

assets/tokenizer.mixtral-8x22b
RissyRan: Good catch! Should we actually reorganize those a little bit? Currently, both mistral and mixtral8x7b-instruct use tokenizer.mistral. Should we do something like (could be a separate PR):

* tokenizer.mistral - mistral v0.1 & mixtral 8x7b v0.1 instruct
* tokenizer.mixtral8x22b - mixtral 8x22b v0.1

@ZhiyuLi-goog thoughts?

rdyro: I think mistral made the situation a little confusing (new tokenizer is 32768 tokens, the previous one is 3200), from HF configs on the models:

| Model        | Tokenizer Size (32000 = old, 32768 = new) |
|--------------|----------------|
| 7b v0.1      | 32,000         |
| 7b v0.3      | 32,768         |
| 7b v0.1 instruct | 32,000         |
| 8x7b v0.1    | 32,000         |
| 8x7b v0.1 instruct| 32,000         |
| 8x22b v0.1   | 32,000         |
| 8x22b v0.1 instruct| 32,768         |

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -129,29 +192,25 @@ def convert(base_model_path, maxtext_model_path, model_size):
...
   ckpt_paths = sorted(pathlib.Path(base_model_path).glob("[!.]*.pth"))
   pytorch_vars = {}
   for i, ckpt_path in enumerate(ckpt_paths):
-    print(f"Loading checkpoint {i+1} of {len(ckpt_paths)} ...")
+    max_logging.log(f"Loading checkpoint {i+1} of {len(ckpt_paths)} ...")
     checkpoint = torch.load(ckpt_path, map_location="cpu")
     pytorch_vars[int(ckpt_path.name.split(".", maxsplit=2)[1])] = checkpoint
   pytorch_vars = [pytorch_vars[i] for i in sorted(list(pytorch_vars.keys()))]
+  # map weight names if they use HuggingFace instead of PyTorch convention
+  pytorch_vars = [_HFNamespaceMapper(var) for var in pytorch_vars]
```
RissyRan, line 208: Nit: could we update this variable (with after usage) to something like `ckpt_vars` or similar to avoid confusion that vars from pytorch only?

rdyro: Good idea, I renamed them!

assets/tokenizer.mixtral-8x22b
ZhiyuLi-goog: A short answer is yes.

Specifically, you can see how tokenizer match each model in [here](https://github.com/mistralai/mistral-common?tab=readme-ov-file#what-is-it).

ZhiyuLi-goog: Maybe we can follow the same practice like v1, v2, & v3.
tokenizer.mistral-${model} is always easier to follow as well, if we don't care to much about the duplication tokenizer files in maxtext.

rdyro: perhaps `tokenizer.mistral-v{1, 2, 3}`? we could put the link you posted in the model config in the comments? I'm not sure what the best way is to point people to that tokenizer matrix

RissyRan: I prefer `tokenizer.mistral-v{1, 2, 3}` a little bit more, so that both mistral-7b & 8x7b will use v1, and 8x22b will use v3 (with comments). And, this `mistral` aligns with decode block name in configs as well.

gobbleturk (approved):

RissyRan (approved):

peregilk:
@rdyro  I encountered OOM (Out of Memory) errors when loading a Llama 8B model on a v4-8 after this commit.  It appears to be related to ```llama_or_mistral_chkpt.py```. While I haven’t pinpointed the exact cause, reverting to the version of MaxText/llama_or_mistral_ckpt.py from commit ```aef1bb0b60c89b6c9876e89ce0b0c35b759235d7``` resolves the issue.

I am able to reproduce the error by using the not-yet-merged script from @A9isha: [llama_or_mistral_orbax_to_huggingface.py](https://github.com/google/maxtext/pull/581/files#diff-515ed590d3a9c0d73d7c5b12cd42cd08f646dd65e9feed3c8af551f93ad6bd1b), where failure occurs at line 97. It’s likely that similar loading processes might trigger the same error.

PR #846, opened 2024-08-27
Add Llama2 config for v5p

raymondzouu:


MaxText/configs/v5p/llama2_7b.sh
```diff
@@ -0,0 +1,50 @@
...
+#
+# Example to AOT compile:
+# bash MaxText/configs/v5p/llama2_7b.sh EXECUTABLE=train_compile.py M_COMPILE_TOPOLOGY=v5p-512 M_COMPILE_TOPOLOGY_NUM_SLICES=1
+
+
+# Stop execution if any command exits with error
+set -e
+
+export EXECUTABLE="train.py" # or train_compile.py
+export DATASET_TYPE="tfds"
```
tonyjohnchen, line 22: If `DATASET_PATH` is passed, can we set DATASET_PATH='' and DATASET_TYPE='synthetic'?

tonyjohnchen (changes requested):
Thanks for adding this Raymond!

MaxText/configs/v5p/llama2_7b.sh
raymondzouu: Set dataset type to synthetic and dataset path shouldn't be needed by default

MaxText/configs/v5p/llama2_7b.sh
```diff
@@ -0,0 +1,50 @@
+# Llama2 7B model.
+# The batch size is set to achieve 4M token global batch size on 1 x v5p-512.
+# If running on a different slice size, modify the per_device_batch_size field
+# accordingly to achieve desired batch size.
+#
+# Command Flags:
+# OUTPUT_PATH (Required, unless base_output_directory is already set in base.yml)
```
tonyjohnchen, line 7: Can you update the command flags requirement as well?

Thanks!

tonyjohnchen (approved):

MaxText/configs/v5p/llama2_7b.sh
raymondzouu: Done

gobbleturk (approved):

PR #857, opened 2024-08-30
Delay Activation Forwarding

gobbleturk:
Add option to delay activation forwarding by one iteration so the communication can easily be overlapped


MaxText/train.py
```diff
@@ -460,7 +460,7 @@ def setup_train_loop(config):
 
   if config.using_pipeline_parallelism:
     # The vocab tensor(s) of shape [vocab, embed] (and transpose) are not sharded by stage
-    params_sharded_tolerance=0.1
+    params_sharded_tolerance=0.5
```
jonb377, line 463: Since the embedding table is unsharded, it's a fixed overhead regardless of the node count, so with enough nodes even this tolerance will be too low. Could we remove this check for now?

MaxText/layers/pipeline.py
```diff
@@ -210,32 +231,38 @@ def get_new_loop_state(self,output, loop_state):
...
+    old_prev_outputs = loop_state["prev_outputs"]
     # Shift becomes a rotated-right version of the previous output
     def _rotate_right(output_in):
       # Use lax.slice to avoid generating a gather.
       last = jax.lax.slice_in_dim(output_in, self.num_stages - 1, self.num_stages, axis=0)
       except_last = jax.lax.slice_in_dim(output_in, 0, self.num_stages - 1, axis=0)
       return jnp.concatenate([last, except_last], axis=0)
-    new_shift = _rotate_right(output)
+    if self.config.pipeline_delay_activation_forwarding:
+      new_shift = _rotate_right(old_prev_outputs)
```
jonb377, line 252: Might be missing something, but where do the inputs get rotated into `prev_outputs`?

MaxText/layers/pipeline.py
```diff
@@ -70,6 +83,13 @@ def init_states(self, inputs):
     shift = jnp.zeros((self.num_stages,) + inputs.shape[1:], dtype=inputs.dtype)
     shift = nn.with_logical_constraint(shift, ("activation_stage", "activation_batch", "activation_length", "activation_embed"),rules=self.config.logical_axis_rules,mesh=self.mesh)
 
+    # Prev outputs has the same shape of the output (and shift)
+    if self.config.pipeline_delay_activation_forwarding:
+      prev_outputs = jnp.zeros((self.num_stages,) + inputs.shape[1:], dtype=inputs.dtype)
```
jonb377, line 88: Disclaimer: I have not thought hard about this, but I wonder if we can consolidate some of these buffers. e.g. combine `shift` and `circ_storage` for the intermediate activations, move the inputs from `state_io` directly into the intermediate buffer so the first stage reads from the same place as the others, and just place the inputs at an offset to achieve delayed forwarding. Then we'd just have buffers for `stage_inputs` and `outputs`.

jonb377 (commented):
Very cool! I'm curious, does this change breaking the data dependency alone enable overlapped comms, or do we need further support from XLA?

MaxText/train.py
gobbleturk: Hmm I have to think about this formula more, the vocab tensor is still sharded by FSDP and TP so there shouldn't be so many more params per chip with mix of PP + FSDP vs pure FSDP. I like the idea of removing this check completely though for now.

MaxText/layers/pipeline.py
gobbleturk: inputs -> outputs -> new_prev_outputs

MaxText/layers/pipeline.py
gobbleturk: This is a good idea! I would have to think about this more. I have thought about similar solutions and it is tricky to remove circular storage - currently circular storage lives all on stage 0 - whereas your solution will keep it sharded, we would need to insert the circular output at the correct stage so it lands on stage 0 when it is needed. We also want this to be non-blocking though, so it should be on stage 0 even before it is needed - this is tricky

gobbleturk:
> Very cool! I'm curious, does this change breaking the data dependency alone enable overlapped comms, or do we need further support from XLA?

From initial experiments I have seen near perfect overlap over ICI, and partial/improved overlap over DCN. I have seen inconsistent results with the DCN experiments - it may be that the message sizes have to be a certain size for stability.  Always the exposed DCN comms reduce with this option, but from AI analysis they should be removed completely

MaxText/layers/pipeline.py
jonb377: Ah, good point! Keeping circ_storage sharded could also unnecessarily increase comms.

MaxText/layers/pipeline.py
jonb377: Synced offline - we're using `prev_outputs` to delay the next stage's compute on this iteration's output, so `prev_outputs` is just an extra intermediary between `outputs` and `shift`.

jonb377 (approved):
LGTM, thanks Matt!

PR #860, opened 2024-08-31
Add load balance loss

RissyRan:
# Description

* add load balance loss to MoE dropping strategy.
* set default `load_balance_loss_weight` as 0.01 (from [paper](https://arxiv.org/pdf/2101.03961) that this value loads balance quickly without interfering the training loss)

# Test
local test on a small model size:
* training 100 steps without lb_loss (`capacity_factor=-1`): [baseline](https://screenshot.googleplex.com/7ScSLCMGZ3W9Kvi) - lb_loss=0
* training 100 steps with lb_loss (`capacity_factor=1`): [test](https://screenshot.googleplex.com/AFEHtmpoT3gT8oB)

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -406,7 +406,7 @@ class MoeBlock(nn.Module):
       output = self.unpermute(intermediate_output,
                               sorted_selected_experts,
                               weights)
-      return output
+      return output, None
     return wrapper(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
 
   def reshape_and_update_weights(self, weights, indices):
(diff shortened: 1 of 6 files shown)


gobbleturk (approved):
Very cool, thank you!

MaxText/layers/linears.py
```diff
@@ -459,14 +459,26 @@ def generate_masks(self, top_k_indices, softmax_probs):
...
+    summed_expert_mask = jnp.sum(expert_mask, axis=2)
+    # Get fraction of tokens dispatched to each expert
+    density = jnp.mean(summed_expert_mask, axis=1)
+    # get fraction of probability allocated to each expert
+    density_prob = jnp.mean(logits, axis=1)
+    loss = jnp.mean(density * density_prob) * (self.num_experts ** 2) * self.config.load_balance_loss_weight
+    return loss
+
   def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
+    softmax_probs = jax.nn.softmax(gate_logits.astype(jnp.float32), axis=-1).astype(self.dtype)
```
rodrigo-f-nogueira, line 474: Hi @RissyRan, it seems that a bug was introduced here? Before we had top_k followed by softmax, but after this PR we have softmax followed by top_k.
I believe the former is the correct implementation for mixtral?

MaxText/train.py
```diff
@@ -607,9 +624,10 @@ def train_loop(config, state=None):
           eval_metrics = p_eval_step(state, eval_batch, nextrng)
         cumulative_eval_metrics["total_loss"] += float(eval_metrics["scalar"]["evaluation/total_loss"])
         cumulative_eval_metrics["total_weights"] += float(eval_metrics["scalar"]["evaluation/total_weights"])
+        cumulative_eval_metrics["moe_lb_loss"] += float(eval_metrics["scalar"]["evaluation/moe_lb_loss"])
         max_logging.log(f"Completed eval step {eval_step_count}")
         eval_step_count += 1
-      eval_loss = cumulative_eval_metrics["total_loss"] / (cumulative_eval_metrics["total_weights"] + EPS)
+      eval_loss = cumulative_eval_metrics["total_loss"] / (cumulative_eval_metrics["total_weights"] + EPS) + cumulative_eval_metrics["moe_lb_loss"] / eval_step_count
```
ZhiyuLi-goog, line 630: @RissyRan I am not sure if we should add aux loss in eval here. 

In genenal, we hope aux_loss affect backprops in training while don't want to mess up the eval value.

RissyRan: Hmm.. this is interesting. I see aux loss is added into the total loss, from [paper](https://arxiv.org/pdf/2101.03961) with a hyper-parameter to interrupt the training loss. Do you have a reference for eval loss? Thanks!

MaxText/layers/linears.py
RissyRan: Hi @rodrigo-f-nogueira, thanks for reaching out! You raised a good point. I recall the change is mainly due to the loss calculation in dropping. We add another normalization back to match the loggits from reference in dropless implementation ([here](https://github.com/AI-Hypercomputer/maxtext/blob/f88db08d7f61d38e806477eac0ea49e6055c9cdf/MaxText/layers/linears.py#L670)).

PR #862, opened 2024-09-02
Improve tfds perf in multihost env

aireenmei:
Have each host read a subset of data file in TFDS (already used by mlperf on mlperf/4.1 branch when dataset_type=c4_mlperf, credits to @ZhiyuLi-goog ) and document data input best practice


ZhiyuLi-goog (approved):
Awesome @aireenmei !
Looks great to me!

gobbleturk (approved):

PR #869, opened 2024-09-06
Enable expert parallelism for dropping strategy

RissyRan:
# Description

Add expert parallelism for dropping strategy

# Test

End-to-end run: [link](https://paste.googleplex.com/6224886885515264)


ZhiyuLi-goog (approved):
Thank you Ran. Look great to me!

MaxText/layers/linears.py
```diff
@@ -436,25 +436,25 @@ def generate_masks(self, top_k_indices, softmax_probs):
...
       expert_token_count_fused = jnp.cumsum(expert_mask_fused, axis=1)
       expert_token_count = jnp.reshape(expert_token_count_fused, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)))
+      expert_token_count = nn.with_logical_constraint(expert_token_count, ("activation_batch", "activation_length", None, None))
       trunc_expert_mask = expert_mask * jnp.less_equal(expert_token_count, expert_capacity_per_batch)
-      updated_expert_mask = jnp.sum(trunc_expert_mask, axis=3)
+      combined_expert_mask = jnp.sum(trunc_expert_mask, axis=2)
 
       # reshape & update weights
-      softmax_probs *= updated_expert_mask
-      weights = self.reshape_and_update_weights(softmax_probs, top_k_indices)
```
gobbleturk, line 446: We may have already discussed this - are we able to drop this here since its mostly a no-op (previous convo https://screenshot.googleplex.com/6mFBfbdjeXGCfSy?)

MaxText/layers/linears.py
```diff
@@ -471,24 +471,35 @@ def load_balance_loss(self, top_k_indices, logits):
     return loss
 
   def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
+    gate_logits = nn.with_logical_constraint(gate_logits, ("activation_batch", None, None))
```
gobbleturk, line 474: Nit: "activation_batch", "activation_length", "activation_embed"

gobbleturk: Also include "activation_embed" below where possible, I think some of None correspond to capacity axis which is fine to leave none, but for embed prefer to use explicit "activation_embed" over None

gobbleturk (approved):
Awesome, thanks Ran!

MaxText/layers/linears.py
RissyRan: Yes, it's safe to drop here with changing 

`updated_expert_mask = jnp.sum(trunc_expert_mask, axis=3)` to `combined_expert_mask = jnp.sum(trunc_expert_mask, axis=2)` at line #442. I also tested in the moe unit test, the output is the same. However, this reshape_and_update_weights function is still needed for non-dropping implementation

MaxText/layers/linears.py
RissyRan: sounds good! Let me cross check after adding them.

tonyjohnchen (approved):

PR #871, opened 2024-09-09
Add block_until_ready operation before checkpoint saving operation.

abhinavclemson:
Add block_until_ready operation before checkpoint saving operation in order to capature accurate checkpoint blocking time on orbax side.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -25,8 +25,9 @@ import os
 import sys
 from etils import epath
 import functools
+import time
 
-from typing import Sequence
+from typing import Sequence, Optional
 from absl import app
 from flax import linen as nn
 from flax.linen import partitioning as nn_partitioning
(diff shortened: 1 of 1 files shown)


MaxText/train.py
```diff
@@ -166,8 +167,28 @@ def clear_buffered_metrics():
...
+    config: Optional[pyconfig.config] = None,
+) -> bool:
+  """Wrapper for saving checkpoint."""
+  if config and config.enable_checkpointing:
+    if (step % config.checkpoint_period == 0) or (
+        config.enable_emergency_checkpoint
+        and step % config.local_checkpoint_period == 0
+    ):
+      blocking_until_ready_start = time.time()
+      max_logging.log("Blocking until ready...")
```
gobbleturk, line 185: Can we change this to be more specific "Waiting for step to finish before checkpoint..."

MaxText/train.py
```diff
@@ -166,8 +167,28 @@ def clear_buffered_metrics():
...
+      jax.block_until_ready(state)
+      max_logging.log(
+          "Blocking until ready took"
```
gobbleturk, line 188: Also here "Waited {x} seconds for step to finish before starting checkpointing"

MaxText/train.py
```diff
@@ -166,8 +167,28 @@ def clear_buffered_metrics():
   _buffered_step = None
   _buffered_metrics = None
 
-def save_checkpoint(checkpoint_manager, step, state, dataset_type="c4", data_iterator=None):
-  """Wrapper for saving checkpoint"""
+def save_checkpoint(
```
gobbleturk, line 170: Hmm not specific to this PR but this method should really be defined inside of checkpointing instead of train...

MaxText/train.py
```diff
@@ -166,8 +167,28 @@ def clear_buffered_metrics():
...
+) -> bool:
+  """Wrapper for saving checkpoint."""
+  if config and config.enable_checkpointing:
+    if (step % config.checkpoint_period == 0) or (
+        config.enable_emergency_checkpoint
+        and step % config.local_checkpoint_period == 0
+    ):
+      blocking_until_ready_start = time.time()
+      max_logging.log("Blocking until ready...")
+      jax.block_until_ready(state)
```
gobbleturk, line 186: Can we add a comment here that we are blocking only for the purpose of accurate timing metrics for checkpointing -
"We block here on the step finishing so that our checkpointing metrics measure only checkpointing time, not training time
"

gobbleturk (approved):

PR #872, opened 2024-09-09
Fix mixtral8x7b model decoder references

kyle-google:


RissyRan:
This is not needed for tokenizer versions.

PR #889, opened 2024-09-13
Add GPT-3 175B v5p MLPerf 4.0 scripts

anfals:
Adding this for easier reproducibility by customer engineers

### DIFF
--- a/MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
+++ b/MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+# GPT-3 175B Model.
+# This script abstracts away all the shared config between Google's various MLPerf 4.0 submissions, which 
+# achieved high performance on the Training benchmark. See the various different TPU type configs in this 
+# directory to actually run the training.
+
+set -euox pipefail
+
+bash preflight.sh PLATFORM=gke
+sleep 60
+
+# flags set as default
+
+# hlo dump
+export XLA_FLAGS="--xla_dump_to=/tmp/xla_dump_file"
+
+# debug
+export TPU_STDERR_LOG_LEVEL=0
+export TF_CPP_MIN_LOG_LEVEL=0
+export TPU_MIN_LOG_LEVEL=0
+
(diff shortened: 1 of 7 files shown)


MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+# GPT-3 175B Model.
+# This script abstracts away all the shared config between Google's various MLPerf 4.0 submissions, which 
+# achieved high performance on the Training benchmark. See the various different TPU type configs in this 
+# directory to actually run the training.
+
+set -euox pipefail
+
+bash preflight.sh PLATFORM=gke
+sleep 60
```
gobbleturk, line 10: Why is this sleep needed? We don't run sleep in other configs?

MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
```diff
@@ -0,0 +1,42 @@
...
+# Read arguments or use defaults from environment variables
+PER_DEVICE_BATCH_SIZE=${1:-${PER_DEVICE_BATCH_SIZE:-4}}
+REMAT_POLICY=${2:-${REMAT_POLICY:-full}}
+ICI_DATA_PARALLELISM=${3:-${ICI_DATA_PARALLELISM:-1}}
+ICI_FSDP_PARALLELISM=${4:-${ICI_FSDP_PARALLELISM:-64}}
+ICI_TENSOR_PARALLELISM=${5:-${ICI_TENSOR_PARALLELISM:-8}}
+RUNNAME=${6:-${RUNNAME:-convergence_test_0}}
+BASE_OUTPUT_DIRECTORY=${7:-${BASE_OUTPUT_DIRECTORY:-gs://some-bucket}}
+
+python3 MaxText/train.py MaxText/configs/base.yml run_name="${RUNNAME}" model_name=gpt3-175b\
```
gobbleturk, line 33: Can we allow either train or train_compile here like in our other config files, e.g. https://github.com/google/maxtext/blob/a2553fa809592a55251fc6f89ef033c264eda72f/MaxText/configs/v5p/128b.sh#L44

gobbleturk (approved):
Awesome, thank you for adding these!

MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
anfals: I based these on the MLPerf scripts we have internally and they had this. I can test without it Monday and see if it works and then update

MaxText/configs/v5p/gpt3_175b/gpt3_175b_base.sh
anfals: Ah that'll be useful! I'd want to test the script with that addition before merging and these scripts need to urgently get in for a Training on Monday, so I'll test that addition on Monday :)

PR #890, opened 2024-09-13
Docker prune in all github actions

khatwanimohit:


gobbleturk:
This looks very necessary!

PR #891, opened 2024-09-14
Add precision option

RissyRan:
# Description

* Cast to f32 before element wise option 
* Add precision option to matmul 
   * JAX [precision](https://kolonist26-jax-kr.readthedocs.io/en/latest/jax.lax.html#jax.lax.Precision), and we currently used default
   * Instead of adding precision to each dot_general or einsum(), we could use [default_matmul_precision](https://kolonist26-jax-kr.readthedocs.io/en/latest/_autosummary/jax.default_matmul_precision.html), which has similar 3 options. However I met issues during flash attention kernel call ([here](https://screenshot.googleplex.com/5AQCzosPWabC48R)). So fall back to add precision to each matmul.

# Test
* base Mixtral 8x7b run (without changes, dtype=bfloat16, weight=f32): [link](https://screenshot.googleplex.com/vXCZbSe4keki2cb) - 2637.763 tokens/s/chip
* base Mixtral 8x7b run (with changes, dtype=bfloat16, weight=f32): [link](https://screenshot.googleplex.com/7joF6i2rJsJESG6) - 2625.901 tokens/s/chip (0.5% regression)
* matmul_precision=float32 run (with changes, dtype=bfloat16, weight=f32): [link](https://screenshot.googleplex.com/9Tbkr9AiZxzzvQe) - 2588.229  tokens/s/chip (1.88% regression)
* matmul_precision=float32 run (with changes, dtype=f32, weight=f32): OOM 14GB [link](https://screenshot.googleplex.com/BjxbExFeAHu2k2J) - makes sense

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -97,6 +97,7 @@ class DenseGeneral(nn.Module):
   kernel_axes: Tuple[str, ...] = ()
   quant: Optional[Quant] = None
   use_bias: bool = False
+  matmul_precision: str = "default"
 
   @nn.compact
   def __call__(self, inputs: Array) -> Array:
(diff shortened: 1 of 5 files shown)


MaxText/layers/linears.py
```diff
@@ -352,7 +358,7 @@ def unpermute(self, intermediate, sorted_selected_experts, weights):
     tensor_parallelism = self.config.ici_tensor_parallelism * self.config.dcn_tensor_parallelism
     reshaped_intermediate = jnp.reshape(unsort_intermediate, (-1, self.num_experts_per_tok, self.config.emb_dim // tensor_parallelism))
     with jax.named_scope("weight_sum"):
-      output = jnp.einsum("BKE,BK -> BE", reshaped_intermediate, reshaped_weights)
```
gobbleturk, line 355: this is still an einsum (matmul), so the float32 doesn't take full affect without setting additional flags. I think we could instead broadcast the weights to BKE, elementwise matmul (bf16 is prob fine for elementwise matmul), then sum in float32.

Alternatively since this is a matmul you can pass in the matmul precision arg here, although I think this op should be set to f32 independent of the matmul precision

gobbleturk (approved):
Thanks Ran!

MaxText/layers/linears.py
RissyRan: ok, I will directly add precision arg there at this moment then. Thanks Matt!

PR #895, opened 2024-09-16
Initialize jax distributed when checkpointing is enabled

jonb377:
Nightly tests are failing due to jax.distributed not being initialized in the synchronous checkpointing case.

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -229,8 +229,6 @@ def maybe_initialize_jax_distributed_system(raw_keys):
     max_logging.log("Jax distributed system initialized on CPUs!")
   elif (
       raw_keys["enable_checkpointing"]
-      and raw_keys["async_checkpointing"]
-      and raw_keys["compile_topology_num_slices"] == -1
       and not raw_keys["enable_single_controller"]
   ) or raw_keys["hardware"] == "gpu_multiprocess":
     max_logging.log("Attempting to initialize the jax distributed system...")


MaxText/max_utils.py
```diff
@@ -229,8 +229,6 @@ def maybe_initialize_jax_distributed_system(raw_keys):
     max_logging.log("Jax distributed system initialized on CPUs!")
   elif (
       raw_keys["enable_checkpointing"]
-      and raw_keys["async_checkpointing"]
-      and raw_keys["compile_topology_num_slices"] == -1
```
gobbleturk, line 233: Shouldn't we keep and raw_keys["compile_topology_num_slices"] == -1? For xAOT we don't need to initialize the jax distributed system - we can compile e.g. on a CPU

jonb377: Ah I actually added this a while ago with GPU AoT: https://github.com/google/maxtext/blob/116c3f2a918a78f037b17657b75c8ed5268a408e/MaxText/max_utils.py#L215-L217

So it's a redundant check now.

gobbleturk (approved):
Thanks Jon!

jonb377:
@gobbleturk Is there a way to detect a multihost environment? I see GPU and inference tests failing due to this change.

gobbleturk:
> @gobbleturk Is there a way to detect a multihost environment? I see GPU and inference tests failing due to this change.

We have to initialize the jax distributed system before the runtime backend, so it doesn't seem this can be inferred. However we can create an additional option to MaxText - either something like "is_single_host" or "should_initialize_jax_distributed_system" in the config as a boolean option, wdyt?

gobbleturk:
> > @gobbleturk Is there a way to detect a multihost environment? I see GPU and inference tests failing due to this change.
> 
> We have to initialize the jax distributed system before the runtime backend, so it doesn't seem this can be inferred. However we can create an additional option to MaxText - either something like "is_single_host" or "should_initialize_jax_distributed_system" in the config as a boolean option, wdyt?

Do we still need to initialize the JDI even for single host? We may want to support supplying correct args in this case - process_id = 0, num_processes=1 coordinator_ip = get_own_ip()

jonb377:
For the TPU tests, perhaps I should try to fix the backend being initialized - the error is in pyconfig, so I don't expect the backend to be up.

The GPU failures can be fixed by setting the coordinator address in the environment. But that does seem overkill for single-host...

PR #901, opened 2024-09-19
Refactoring Maxtext build process with stable stack

parambole:
- Added Mode=stable_stack  `docker_build_dependency_image.sh` which enables Maxtext to be built on top of jax stable stack to maintain consistency
- Removed the old flow


MaxText/requirements_with_jax_ss.txt
gobbleturk, line 1: It looks like this file is necessary? It is being called from UploadDockerImages? Probably I am missing something

parambole: This file was used in an older flow and should have been removed when that flow was removed. https://github.com/google/maxtext/blob/main/requirements_with_jax_stable_stack.txt is the file which is used in the existing flow and replaces the file that is being removed.

gobbleturk: ah sorry didn't catch the slight filename difference, thanks!

gobbleturk (approved):

PR #903, opened 2024-09-19
Adds a new end-to-end test for Mistral 7b

shralex:
The changes include:
- A notebook in scratch_pad that runs the model from HF, for sample inputs and generates logits, which it stores in a file.
- A test that downloads the HF checkpoint, converts it to MaxText-compatible format, and uses it to do one forward pass on a sample input.
- Compares the logits obtained in both steps above to make sure they're equal. 

The old test, which asserted on the generated text being equal, rather than logits, is removed.


end_to_end/tpu/mistral/7b/test_mistral-7b.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# This file runs on daily basis and demonstrates:
```
khatwanimohit, line 3: nit: mention where we run this test. v4-8 in this case.

khatwanimohit (approved):
LGTM!

end_to_end/tpu/mistral/7b/test_mistral-7b.sh
gobbleturk, line 1: I guess this file should be split into 2 parts like the rest, but that is unrelated to fix here so okay to leave for now

gobbleturk (approved):

end_to_end/tpu/mistral/7b/test_mistral-7b.sh
shralex: It was previously split, and merged into one by Mohit's request, since this is a small model.

gobbleturk: ah makes sense, sounds good!

end_to_end/tpu/mistral/7b/test_mistral-7b.sh
shralex: Done

PR #905, opened 2024-09-20
Making sure we run pylint only once, and run pyink in the same way.

shralex:


MaxText/maxengine_server.py
```diff
@@ -45,6 +45,7 @@ def main(config):
   # We separate credential from run so that we can unit test it with
   # local credentials.
   # TODO: Add grpc credentials for OSS.
+  # pylint: disable=unexpected-keyword-arg
```
gobbleturk, line 48: This one is potentially real - will follow up with jetstream folks

gobbleturk (approved):

PR #906, opened 2024-09-21
Give user option for activation type precision

gobbleturk:
Give user option for activation type precision

With this option on, we see a slight degradation in performance ~2%


tonyjohnchen (approved):
LGTM!

MaxText/configs/base.yml
```diff
@@ -70,6 +70,7 @@ quantization: ""
 # Choose one of default, high, and highest.
 # https://kolonist26-jax-kr.readthedocs.io/en/latest/jax.lax.html#jax.lax.Precision
 matmul_precision: "default"
+activations_in_float32: False # Sets activations to float32 before nonlinearity it true, else dtype
```
tonyjohnchen, line 73: We have `dtype` and `weight_dtype` already, why don't we have `activation_dtype` and set it as bf16 by default?

PR #909, opened 2024-09-23
move gsutil copy to condtional to avoid breakages

kocchop:
gsutil is not in the requirements list, hence not installed by default. The copy command is breaking for the containers which don't have it by default. Moved it under a conditonal

### DIFF
--- a/MaxText/profiler.py
+++ b/MaxText/profiler.py
@@ -18,6 +18,7 @@ limitations under the License.
 from ctypes import cdll
 import os
 import subprocess
+import shutil
 
 import max_logging
 
@@ -61,6 +62,9 @@ class Profiler:
         max_logging.log("WARNING: library for nsys was not loaded \n" "profiler has no effect")
         return
       # Popen() instead of run() for non-blocking behavior
-      subprocess.Popen(["gsutil", "cp", "*nsys-rep", self.output_path])  # pylint: disable=consider-using-with
+      if shutil.which('gsutil') is not None:
+        subprocess.Popen(["gsutil", "cp", "*nsys-rep", self.output_path])  # pylint: disable=consider-using-with
+      else:
+        max_logging.log("WARNING: gsutil is not installed or not found in the system's PATH. Skipping upload...")
     elif self.mode == "xplane":
       jax.profiler.stop_trace()


jonb377 (approved):
Thanks @kocchop!

gobbleturk:
Isn't it better to error out when a user requests profiling but the profiling fails? User can set profiler='' (default, no profiling) to avoid profiling instead of this check?

This error points to a real bug in maxtext - we should either ensure gsutil is installed or perhaps can use a different gcs API (upload_blob)

kocchop:
Hi @gobbleturk the profiling works fine as is. This line just copies the generated `.nsys-rep` file to some directory. Without `gsutil` this copy fails and the run breaks. If for some reason the profiling breaks, it should break even with this PR. In sum, this PR does not ignore profiling failures, rather it avoids the breakages caused by gsutil file copy.

gobbleturk:
> Hi @gobbleturk the profiling works fine as is. This line just copies the generated `.nsys-rep` file to some directory. Without `gsutil` this copy fails and the run breaks. If for some reason the profiling breaks, it should break even with this PR. In sum, this PR does not ignore profiling failures, rather it avoids the breakages caused by gsutil file copy.

Don't we need to move the profile to GCS? How are we able to use the profile otherwise?

kocchop:
@gobbleturk yes, on clusters using gcs, it is needed indeed. Hence, the nightly maxtext container made through this [script](https://github.com/AI-Hypercomputer/maxtext/blob/0a919c19911ea2d99445e72a59e838f466b962c6/maxtext_gpu_dependencies.Dockerfile#L19) installs `gsutil` etc. However, for other clusters it may not be necessary as there could be different file system mounts. Without any checks, it leads to breakages.

Another way could be to enforce the `gsutil` to be on the requirements list. Even with that, we might need to change the behavior for non-gcs scenarios.

gobbleturk:
> script

Ya this makes sense. For TPUs the profile will be sent to `base_output_directory` which is configurable (can be GCS or local), I think we should change the GPU path to also be configurable, as it won't always be GCS.

gobbleturk (approved):
Will approve for now to unblock nsys profiling use cases, as we work on longer term fix

gobbleturk:
> Will approve for now to unblock nsys profiling use cases, as we work on longer term fix

Created https://github.com/AI-Hypercomputer/maxtext/issues/911 to track longer term fix

Issue #878, opened 2024-09-10
Mask is being ignored when cudnn_flash_attention is used


finbarrtimbers:
From the [TransformerEngine docs](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/jax.html#transformer_engine.jax.flax.DotProductAttention):

> mask in [__call__](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/jax.html#transformer_engine.jax.flax.DotProductAttention.__call__) is ignored for ‘no_mask’ and ‘causal’.

I think that `attn_mask_type` should be set to `causal_padding`, or else it'll ignore the mask being passed on line 370.

bvandermoon:
Thanks Finbarr. Created a [PR to address this](https://github.com/AI-Hypercomputer/maxtext/pull/913)

bvandermoon:
Resolving this issue since the PR has been merged. Please reopen or let us know if you have any further issues

PR #913, opened 2024-09-24
Use attn_mask_type of causal_padding for cudnn_flash_attention

bvandermoon:
### Notes

Update attn_mask_type so that it is not ignored. Related to issue [878](https://github.com/AI-Hypercomputer/maxtext/issues/878).

Note that this change currently only impacts workloads using `cudnn_flash_te` attention.

### Testing

* Trained 10 steps on GPUs with `cudnn_flash_te` enabled. Saw the same high-level output (step time, loss, etc.) before and after this change
* Ran the logit checker with llama2-7b logits and the test succeeded:

Screenshot 2024-10-01 at 2 06 07 PM

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -383,7 +383,7 @@ class AttentionOp(nn.Module):
         head_dim=head_dim,
         num_attention_heads=self.num_query_heads,
         num_gqa_groups=self.num_kv_heads,
-        attn_mask_type="causal",  # 'causal' or 'padding'
+        attn_mask_type="padding_causal",  # 'no_mask', 'padding', 'causal', or 'padding_causal'
         attn_bias_type="NO_BIAS",  # 'no_bias', 'pre_scale_bias' or 'post_scale_bias'
         attention_dropout=self.dropout_rate,
         dropout_rng_name="aqt",


gobbleturk:
> ### Notes
> Update attn_mask_type so that it is not ignored. Related to issue [878](https://github.com/AI-Hypercomputer/maxtext/issues/878).
> 
> Note that this change currently only impacts workloads using `cudnn_flash_te` attention.
> 
> ### Testing
> Trained 10 steps on GPUs with `cudnn_flash_te` enabled. Saw the same high-level output (step time, loss, etc.) before and after this change.
> 
> Note: The loss was suspicious since it dropped from 10.307 to 0 after the third step. But this happened before and after this change so it appears to be caused by something else

I assume the loss is dropping so quickly because you are using synthetic data (`dataset_type=synthetic`), you can use real data to measure the loss. Ideally we would test correctness with something like our [golden logits test](https://github.com/AI-Hypercomputer/maxtext/blob/5ae1b711bc94bd0003d1d0ad76a220f3a76fbe3b/MaxText/tests/forward_pass_logit_checker.py#L97)

gobbleturk (approved):
Confirmed golden test results offline!

bvandermoon:
> > ### Notes
> > Update attn_mask_type so that it is not ignored. Related to issue [878](https://github.com/AI-Hypercomputer/maxtext/issues/878).
> > Note that this change currently only impacts workloads using `cudnn_flash_te` attention.
> > ### Testing
> > Trained 10 steps on GPUs with `cudnn_flash_te` enabled. Saw the same high-level output (step time, loss, etc.) before and after this change.
> > Note: The loss was suspicious since it dropped from 10.307 to 0 after the third step. But this happened before and after this change so it appears to be caused by something else
> 
> I assume the loss is dropping so quickly because you are using synthetic data (`dataset_type=synthetic`), you can use real data to measure the loss. Ideally we would test correctness with something like our [golden logits test](https://github.com/AI-Hypercomputer/maxtext/blob/5ae1b711bc94bd0003d1d0ad76a220f3a76fbe3b/MaxText/tests/forward_pass_logit_checker.py#L97)

Thanks Matt. That was the issue with the loss. Also ran the golden logits test and it passed. Updated the description with the test

PR #921, opened 2024-09-25
Adds pathwaysutils as a dependency

lukebaumann:
This is a no-opt aside from some log messages if Pathways-on-Cloud is not being used.

### DIFF
--- a/MaxText/maxengine_server.py
+++ b/MaxText/maxengine_server.py
@@ -17,6 +17,7 @@
 import jax
 import os
 import sys
+import pathwaysutils  # pylint: disable=unused-import
 import pyconfig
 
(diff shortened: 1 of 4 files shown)


gobbleturk (approved):

MaxText/maxengine_server.py
```diff
@@ -17,6 +17,7 @@
 import jax
 import os
 import sys
+import pathwaysutils  # pylint: disable=unused-import
```
bvandermoon, line 20: @lukebaumann FYI these imports are being removed in [PR 1482](https://github.com/AI-Hypercomputer/maxtext/pull/1482). Let us know if there are any issues with that

PR #925, opened 2024-09-30
[MoE] fix aqt einum api call and optimize int8 all-gather performance

ZhiyuLi-goog:
* fix the bug that precision args is not supported in aqt `einsum`, simply print warning and skip the args
* the following lines will trigger all-gather kernel weights in its original `weight_dtype` which is problematic if we can't overlap the communication. Bypassing these lines allows us to [all-gather kernel weights with int8](https://screenshot.googleplex.com/7ZbPGNhqcaGQyuG), which would improve performance esp. with all-gather communication exposed. (e.g. 16.8% -> 23.5% in v6e)
```
# To skip in int8 since we don't want to all gather kernel weights in bf16
kernel_axes = ("exp", None, None)
kernel = nn.with_logical_constraint(kernel, kernel_axes)
```

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -495,10 +495,27 @@ class MoeBlock(nn.Module):
 
   def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = ()):
     if self.quant:
-      einsum_op = self.quant.einsum(rhs_mesh_axes)
+
+      def aqt_einsum(*args, **kwargs):
+        # simply skip kwargs, since aqt einsum doesn't support any kwargs like precision
+        return self.quant.einsum(rhs_mesh_axes)(*args)
+
+      einsum_op = aqt_einsum
     else:
       einsum_op = jnp.einsum
     return einsum_op
+
+  def is_expert_parallelism_enabled(self):
+    return self.config.ici_expert_parallelism > 1 or self.config.dcn_expert_parallelism > 1
+
+  def maybe_all_gather_kernel_weight_in_expert_parallelism(self, kernel, kernel_axes):
+    if self.is_expert_parallelism_enabled():
+      # This will trigger all-gather using weight_dtype
+      # relax it unless really necessary in expert parallelism only
+      # Otherwise compiler will handle communication automatically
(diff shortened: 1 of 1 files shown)


MaxText/layers/linears.py
```diff
@@ -495,11 +495,29 @@ def load_balance_loss(self, top_k_indices, logits):
 
   def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = ()):
     if self.quant:
-      einsum_op = self.quant.einsum(rhs_mesh_axes)
+
+      def aqt_einsum(*args, **kwargs):
+        # aqt einsum doesn't support kwargs like precision
+        max_logging.log(f"[WARNING] {kwargs=} argument is not supported and will be ignored in aqt_einsum")
```
gobbleturk, line 501: I would remove this warning, it is just spammy IMO

MaxText/layers/linears.py
```diff
@@ -495,11 +495,29 @@ def load_balance_loss(self, top_k_indices, logits):
...
+      einsum_op = aqt_einsum
     else:
       einsum_op = jnp.einsum
     return einsum_op
 
+  def is_expert_parallelism_enabled(self):
+    return self.config.ici_expert_parallelism > 1 or self.config.dcn_expert_parallelism > 1
+
+  def maybe_all_gather_kernel_weight_in_expert_parallelism(self, kernel, kernel_axes):
+    if self.is_expert_parallelism_enabled():
```
gobbleturk, line 513: Is this necessary even with expert parallelism? Does the compiler not find the right solution without this sharding constraint?

gobbleturk (approved):

MaxText/layers/linears.py
ZhiyuLi-goog: Yeap, I have tested and we need this line to let compiler understand the sharding.

MaxText/layers/linears.py
ZhiyuLi-goog: Sure!

ZhiyuLi-goog: Removed.

PR #926, opened 2024-09-30
modified the dockerfile for better caching

ultrons:
Split the copy operation into minimal setup and requirement.txt file and defer the full repo copy at the end.
This prevent full dependency rebuild every time a user makes local changes and wants to quickly build an re-run on a remote cluster.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -86,6 +86,14 @@ fi
 # Save the script folder path of maxtext
 run_name_folder_path=$(pwd)
 
+# Install dependencies from requirements.txt
+cd $run_name_folder_path && pip install --upgrade pip
+if [[ "$MODE" == "pinned" ]]; then
+    pip3 install -U -r requirements.txt -c constraints_gpu.txt
+else
+    pip3 install -U -r requirements.txt
+fi
+
 # Uninstall existing jax, jaxlib and  libtpu-nightly
 pip3 show jax && pip3 uninstall -y jax
 pip3 show jaxlib && pip3 uninstall -y jaxlib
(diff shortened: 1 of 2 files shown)


gobbleturk (approved):
Thanks Vaibhav!

jonb377:
cc @tonyjohnchen, I believe this is the use case of `docker_upload_runner.sh`

tonyjohnchen (approved):
LGTM

jonb377 (approved):

ultrons:
Added one more fix. 
Jax jax lib installation should be done last in the setup.sh
Since we allow user to control it via the build args. 
The current order, install dependencies from requirements.txt last.
If any library in this list causes jax/jax lib version overwrite, the input from the the users will not be remain in the final build.
This leads to painful jax version debug.

To repro, you can build with JAX_VERSION=0.4.30 and MODE=stable
The final build will have jax/jaxlib and 0.4.33, which is not expected.
The reason is that jax version gets overwritten during the dependency installation via requirements.txt.

tonyjohnchen (approved):
lgtm

setup.sh
```diff
@@ -86,6 +86,14 @@ fi
 # Save the script folder path of maxtext
 run_name_folder_path=$(pwd)
 
+# Install dependencies from requirements.txt
+cd $run_name_folder_path && pip install --upgrade pip
+if [[ "$MODE" == "pinned" ]]; then
+    pip3 install -U -r requirements.txt -c constraints_gpu.txt
+else
+    pip3 install -U -r requirements.txt
+fi
```
jonb377, line 95: Great catch!

jonb377 (approved):

PR #927, opened 2024-09-30
change type handlers to be registered once per host

ssusie:
tested [https://cloudloggi](https://cloudlogging.app.goo.gl/ufwez6VZWN2tuy4Y8)


gobbleturk (approved):
Thanks Susie!

PR #928, opened 2024-10-01
To shift or to rotate, that is the question

gobbleturk:
We "Shift" the shift buffer (meaning everything moves down one, we pad the new hole at position 0 and the last inputs are dropped), or we "rotate" (meaning everthing moves down one, the hole at position 0 is filled by the last input) depending on if the last stage should immediately send to first or not

### DIFF
--- a/MaxText/layers/pipeline.py
+++ b/MaxText/layers/pipeline.py
@@ -272,18 +272,32 @@ class Pipeline(nn.Module):
     loop_iteration = loop_state["loop_iteration"]
     old_prev_outputs = loop_state["prev_outputs"]
 
-    # Shift becomes a rotated right version of the previous output
-    def _rotate_right(output_in):
+    def _rotate_right(arr):
       # Use lax.slice to avoid generating a gather.
-      last = jax.lax.slice_in_dim(output_in, self.num_stages - 1, self.num_stages, axis=0)
-      except_last = jax.lax.slice_in_dim(output_in, 0, self.num_stages - 1, axis=0)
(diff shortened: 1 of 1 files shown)


MaxText/layers/pipeline.py
```diff
@@ -272,18 +272,29 @@ def get_new_loop_state(self, output, loop_state):
...
+      last = jax.lax.slice_in_dim(arr, self.num_stages - 1, self.num_stages, axis=0)
+      except_last = jax.lax.slice_in_dim(arr, 0, self.num_stages - 1, axis=0)
       return jnp.concatenate([last, except_last], axis=0)
 
+    def _shift_right(arr):
+      padding = [[1, 0]] + [[0, 0]] * (arr.ndim - 1)
+      # Use lax.slice to guarantee the gradient is a pad.
+      return jax.lax.slice(jnp.pad(arr, padding), [0] * arr.ndim, arr.shape)
+
+    # Shift either rotates or shifts depending on if the last stage immediately must send to first or not
```
jonb377, line 286: Can you explain the cases a bit more in the comment? My impression:
- No repeats => last stage output is headed to state_io (L303 in the new diff)
- use_circular_storage => last stage output is headed to circ_storage (L322 in the new diff)
- Otherwise, stage 0 immediately needs the last stage's output for its next repeat.

jonb377 (approved):
Thanks Matt!

MaxText/layers/pipeline.py
gobbleturk: Yes thats correct, definitely will add more of a comment here. Note last stage always heads to state_io regardless of config, but its often dummy unless its on the last repeat

PR #932, opened 2024-10-02
Add more MaxText reviewers

gobbleturk:
Add more MaxText reviewers


gobbleturk:
apparently my main is dirty =(

Issue #595, opened 2024-04-16
Cannot do inference in float32


borisdayma:
If we try to perform inference in float32, we get the error:
```
AssertionError: Key and Value Dtypes should match
```

This error comes from [this line](https://github.com/google/maxtext/blob/ebd39aa64d670fa13a313b6f776e01ad9e450321/MaxText/layers/attentions.py#L513).

The origin of the error is that the cache `dtype` is set to `jnp.int8 if quantize_kvcache else jnp.bfloat16` but never to `jnp.float32`.

rwitten:
What are you setting that triggets this? (Activations to float32?)

borisdayma:
Yes it's the dtype: https://github.com/google/maxtext/blob/f52e6f7b49277688f68a58284d1ff7873122ca41/MaxText/configs/base.yml#L61

bvandermoon:
Thanks @borisdayma for pointing this out. Created a [PR to address this](https://github.com/AI-Hypercomputer/maxtext/pull/936)

bvandermoon:
Resolving this issue since the PR has been merged. Please reopen or let us know if you have any further issues

borisdayma:
Thanks a lot for doing this @bvandermoon 🔥

PR #936, opened 2024-10-03
Pass config dtype into RotaryEmbedding

bvandermoon:
### Notes
Use the same dtype before and after calling RotaryEmbedding. Related to [issue 595](https://github.com/AI-Hypercomputer/maxtext/issues/595).

Not passing dtype into RotaryEmbedding was causing the keys to be cast to bfloat16. This caused the linked issue where running decode with float32 was throwing this exception:
```
AssertionError: Key and Value Dtypes should match
```

Note that I applied the same change for the query since this was also being cast to bfloat16. Let me know if I should keep the query casting the same as it was.

### Testing
* Successfully ran decode with bfloat16/float32. I was able to see key/value (and query) dtype matching for both runs. I also saw them not matching before this change
```
key: Traced<ShapedArray(float32[4,1,16,128])>with<DynamicJaxprTrace(level=3/0)>
value: Traced<ShapedArray(float32[4,1,16,128])>with<DynamicJaxprTrace(level=3/0)>
```
* Successfully ran training with bfloat16/float32

### DIFF
--- a/MaxText/inference_microbenchmark_sweep.py
+++ b/MaxText/inference_microbenchmark_sweep.py
@@ -25,10 +25,11 @@ import inference_microbenchmark
 import pyconfig
 
 try:
-    JaxRuntimeError = jax.errors.JaxRuntimeError  # added in JAX 0.4.34
+  JaxRuntimeError = jax.errors.JaxRuntimeError  # added in JAX 0.4.34
 except AttributeError:
-    from jax._src.lib import xla_extension
-    JaxRuntimeError = xla_extension.XlaRuntimeError
+  from jax._src.lib import xla_extension
+
+  JaxRuntimeError = xla_extension.XlaRuntimeError
 
 
 def main():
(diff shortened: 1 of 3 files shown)


gobbleturk (approved):
Thanks for the fix!

rwitten:
Add a test!

bvandermoon:
> Add a test!

Done, thanks @rwitten. Great to hear from you and hope things are going well

MaxText/inference_microbenchmark_sweep.py
```diff
@@ -25,10 +25,11 @@
 import pyconfig
 
 try:
-    JaxRuntimeError = jax.errors.JaxRuntimeError  # added in JAX 0.4.34
```
bvandermoon, line 28: Note that the code_style.sh was failing due to this change from an automated code test. Changes in this file aren't related other than to get the code style check to pass

PR #939, opened 2024-10-03
Adding options to enable profiler, improving logging and some minor code fixes/cleanup.

singh-mitali:


MaxText/inference_mlperf/offline_mode.py
```diff
@@ -31,12 +31,23 @@
...
+import warnings
+
+warnings.simplefilter("ignore", category=FutureWarning)
+
+import inspect
+
+current_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parent_dir = os.path.dirname(current_dir)
+sys.path.insert(0, parent_dir)
+
```
vipannalla, line 44: Is this to suppress the JAX warnings or something else?

MaxText/maxengine.py
```diff
@@ -484,27 +484,31 @@ def create_engine_from_config_flags(batch_size, max_prefill_predict_length, max_
...
-  args_str = "MaxText/maxengine_server.py configs/base.yml " + args_str
-  cmd_args = [b for b in args_str.split(" ") if len(b) > 0]
-  # Add at the end to cmd args override the default values set above
-  args.extend(cmd_args)
-
-  pyconfig.initialize(args)
+  args["max_prefill_predict_length"] = f"{max_prefill_predict_length}"
+  args["max_target_length"] = f"{max_target_length}"
+  args["per_device_batch_size"] = f"{batch_size}"
+  print(args_str)
```
vipannalla, line 500: nit: add a short phrase to indicate what is printed, how about "Default args: "?

vipannalla (approved):
Looks good, thanks for the cleanup.

MaxText/inference_mlperf/offline_mode.py
singh-mitali: 37 is to suppress jax future warnings. Others are to include python modules from parent dir

MaxText/maxengine.py
singh-mitali: Done

PR #945, opened 2024-10-04
Adding support for configuring single and multiple batches.

singh-mitali:


MaxText/inference_mlperf/offline_mode.py
```diff
@@ -344,22 +378,18 @@ def make_response(id_, response_token_ids):
   return query_sample_response
 
 
-def _count_by_bucket(dataset):
-
+def _estimated_counts_by_bucket(dataset):
```
vipannalla, line 381: I know this was a miss in original code, but can we use lengths from `FLAGS.prefill_lengths_and_batch_sizes` to generate the counts instead of 256, 512, and 1024?

vipannalla (approved):
LGTM

PR #948, opened 2024-10-05
Align profiler start and end times with train_step via block_until_ready

gobbleturk:
Align profiler start and end times with train_step via block_until_ready

Example profile with this change (with default skip_first_n_steps_for_profiler=1, profiler_steps=5) 
(https://xprof.corp.google.com/trace_viewer/mattdavidow-7890031890416702717?hosts=t1v-n-c3cb1aa0-w-0&host_index=0&view_start=-1109.907&view_end=8543.576)

PR #949, opened 2024-10-05
Fix pylint from internal PR

gobbleturk:
Fix pylint from internal cl/682425232

We should really enforce the pylint from internal CLs as well (ideally also our entire test suite)

### DIFF
--- a/MaxText/inference_microbenchmark_sweep.py
+++ b/MaxText/inference_microbenchmark_sweep.py
@@ -25,10 +25,11 @@ import inference_microbenchmark
 import pyconfig
 
 try:
-    JaxRuntimeError = jax.errors.JaxRuntimeError  # added in JAX 0.4.34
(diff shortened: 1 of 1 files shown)


jonb377 (approved):
Oh interesting, so this broke main?

khatwanimohit (approved):

gobbleturk:
> Oh interesting, so that broke main?

It looks like the pylint only runs on branch pushes (e.g I am seeing in lint failure in unrelated #948), but not part of our recurrent test suite. The lint is configured here https://github.com/AI-Hypercomputer/maxtext/blob/d847f6cb6bb0b8127f9463adb2e54ef58efcb970/.github/workflows/CPUTests.yml#L1-L6, perhaps should be moved to our unit tests which also run periodically here https://github.com/AI-Hypercomputer/maxtext/blob/d847f6cb6bb0b8127f9463adb2e54ef58efcb970/.github/workflows/UnitTests.yml#L25-L27

PR #952, opened 2024-10-07
Reduce disk usage for mixtral tests

rdyro:
This PR modifies the parameter conversion mixtral tests to go through `gcsfuse` instead of disk for lower VM disk usage

### DIFF
--- a/end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
+++ b/end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
@@ -14,21 +14,30 @@ MODEL_VARIATION='8x22b'
 
 if [ -z "${BASE_OUTPUT_PATH}" ]; then
     # Non-Googlers please remember to point BASE_OUTPUT_PATH to GCS buckets that you own, this script uses internal buckets for testing.
-    export BASE_OUTPUT_PATH=gs://runner-maxtext-logs/$(date +%Y-%m-%d)
+    export BASE_OUTPUT_PATH=gs://runner-maxtext-logs/$(date +%Y-%m-%d)/
     echo "BASE_OUTPUT_PATH is not set, using BASE_OUTPUT_PATH = ${BASE_OUTPUT_PATH}"
 fi
+BASE_OUTPUT_PATH=gs://rdyro-mixtral-storage/logs/megamem-logs
 
 # Download checkpoint
 pip3 install torch
-MODEL_NAME=Mixtral-8x22B-Instruct-v0.1
-gcloud storage cp -r "gs://maxtext-external/$MODEL_NAME" /tmp
-                     
+MODEL_NAME="Mixtral-8x22B-Instruct-v0.1"
+
+PARAM_DIR="$HOME/tempdisk"
+mkdir -p "$PARAM_DIR"
+[[ ! -z $(ls "$PARAM_DIR") ]] && fusermount -u "$PARAM_DIR"
+gcsfuse --implicit-dirs maxtext-external "$PARAM_DIR"
+# alternatively: $ gcloud storage cp -r "gs://maxtext-external/$MODEL_NAME" $PARAM_DIR
+
 # Convert it to MaxText(orbax) format - scanned ckpt
-JAX_PLATFORMS=cpu python3 MaxText/llama_or_mistral_ckpt.py --base-model-path="/tmp/$MODEL_NAME" --model-size=mixtral-8x22b --maxtext-model-path=${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt/
+JAX_PLATFORMS=cpu python3 MaxText/llama_or_mistral_ckpt.py --base-model-path="$PARAM_DIR/$MODEL_NAME" --model-size=mixtral-8x22b --maxtext-model-path=${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt/
 echo "Wrote MaxText compatible scanned checkpoint to ${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt"
+
+# unmount the gcsfuse directory
+fusermount -u "$PARAM_DIR"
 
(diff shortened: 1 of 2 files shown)


end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
```diff
@@ -14,21 +14,30 @@ MODEL_VARIATION='8x22b'
...
+gcsfuse --implicit-dirs maxtext-external "$PARAM_DIR"
+# alternatively: $ gcloud storage cp -r "gs://maxtext-external/$MODEL_NAME" $PARAM_DIR
+
 # Convert it to MaxText(orbax) format - scanned ckpt
-JAX_PLATFORMS=cpu python3 MaxText/llama_or_mistral_ckpt.py --base-model-path="/tmp/$MODEL_NAME" --model-size=mixtral-8x22b --maxtext-model-path=${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt/
+JAX_PLATFORMS=cpu python3 MaxText/llama_or_mistral_ckpt.py --base-model-path="$PARAM_DIR/$MODEL_NAME" --model-size=mixtral-8x22b --maxtext-model-path=${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt/
 echo "Wrote MaxText compatible scanned checkpoint to ${BASE_OUTPUT_PATH}/${MODEL_VARIATION}/scanned_ckpt"
 
+# unmount the gcsfuse directory
+fusermount -u "$PARAM_DIR"
```
jonb377, line 37: Excellent, thanks for scoping the mount!

end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
```diff
@@ -14,21 +14,30 @@ MODEL_VARIATION='8x22b'
 
 if [ -z "${BASE_OUTPUT_PATH}" ]; then
     # Non-Googlers please remember to point BASE_OUTPUT_PATH to GCS buckets that you own, this script uses internal buckets for testing.
-    export BASE_OUTPUT_PATH=gs://runner-maxtext-logs/$(date +%Y-%m-%d)
+    export BASE_OUTPUT_PATH=gs://runner-maxtext-logs/$(date +%Y-%m-%d-%H-%M)/
```
jonb377, line 17: Do we need to update this on the other test as well?

jonb377 (approved):
Thanks Robert! LGTM

jonb377:
Also just a high-level question - @rdyro did you notice any slowdown using gcsfuse compared to a local copy of the weights?

rdyro:
> Also just a high-level question - @rdyro did you notice any slowdown using gcsfuse compared to a local copy of the weights?

I did not check quantitatively, but qualitatively, there wasn't much of a difference. The loading of the existing checkpoint to be converted takes less than 10% of the total run, even after moving to gcsfuse.

If it's important we measure this (e.g., for elsewhere), let me know

end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
rdyro: it should work without this, but it's probably a good idea to unify the script, I updated the file

RissyRan: Probably you want to pull the latest change for this line? I think we run it at most daily, so `$(date +%Y-%m-%d)` should work?

RissyRan (commented):
Thank you!

rdyro:
> > Also just a high-level question - @rdyro did you notice any slowdown using gcsfuse compared to a local copy of the weights?
> 
> I did not check quantitatively, but qualitatively, there wasn't much of a difference. The loading of the existing checkpoint to be converted takes less than 10% of the total run, even after moving to gcsfuse.
> 
> If it's important we measure this (e.g., for elsewhere), let me know

Correction: the new strategy via `gcsfuse` is equivalent to single-threaded download which on a high-end VM caps at around 80-100 MB/s on average, implying 45 min - 60 min download for a 280 GB checkpoint (8x22b)

end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
jonb377: Ah interesting point @RissyRan, the `1_test...` and `2_test...` scripts are chained, so the bucket name needs to match between the two right? 

I would prefer more granular paths (ie including H and M and even S), since we've hit race conditions previously due to lower granularity, but that seems incompatible with the way these tests were designed.

Should we even set a default value here or require it to be user-specified?

jonb377: @rdyro Let's keep the `BASE_OUTPUT_PATH` consistent between the `1_test...` and `2_test...` files, and Ran and I can follow up offline on the right long-term approach. Feel free to merge with that modification. Thanks!

rdyro:
This test was also sometimes (?) running out of RAM without `weight_dtype=bfloat16`, I'm adding that change to this PR

end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
RissyRan: Yes, that's right. Those 2 tests need to use the same gcs location (like [this](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_end_to_end.py#L149)).

Good point for the output path! I see it has been updated to only year-month-day in this [PR](https://github.com/AI-Hypercomputer/maxtext/commit/60371239596cae3aef637b741531911b1542d9d5#diff-8e059a6911f0bb1737ca354b40f63cdef7cd0d268646cd361402d1b49c063943). Yeah, we should add those back for rest of end-to-end tests.

Default value should be fine here. I think it may be better to add model name as subfolder before the time/date. So we could easily find outputs.

PR #954, opened 2024-10-07
Checkpoint  tfds data iterator

mattdonati:
Enable deterministic training with preemption when using tfds pipeline by checkpointing data iterator. 

Creates a checkpoint handler for data iterator that implements orbax.checkpoint.CheckpointHandler, similar to https://github.com/google/grain/blob/main/grain/_src/python/checkpoint_handlers.py. Handler utilizes tf.train.Checkpoint to save and restore iterator.  

Makes checkpointing the data iterator optional, since this method will save large checkpoints. Adds a bool flag to base.yml 

Async checkpointing is handled at the level of the orbax checkpoint manager. 

Updates input pipeline description to reflect option to checkpoint tfds iterator.

### DIFF
--- a/getting_started/Data_Input_Pipeline.md
+++ b/getting_started/Data_Input_Pipeline.md
@@ -21,7 +21,7 @@ Currently MaxText has three data input pipelines:
 | -------- | --------------- | -------- | ----------- |
 | HuggingFace | datasets in HuggingFace Hub<br>local/Cloud Storage datasets in json, parquet, arrow, csv, txt | convenience<br>multiple formats | limit scalability using HuggingFace Hub<br>non-deterministic with preemption<br>(deterministic without preemption) |
 | Grain | ArrayRecord, available through Tensorflow Datasets | fully deterministic, regardless of preemption | only supports random access datasets |
-| TFDS | TFRecord, available through Tensorflow Datasets |  | only supports TFRecords<br>non-deterministic with preemption<br>(deterministic without preemption) |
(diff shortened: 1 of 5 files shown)


getting_started/Data_Input_Pipeline.md
```diff
@@ -135,3 +134,8 @@ eval_split: 'validation'
 # TFDS input pipeline only supports tokenizer in spm format
 tokenizer_path: "assets/tokenizer.llama2"
 ```
+3. Enable determinism (optional)
+```
+tfds_iter_checkpointing: True
+```
+Note that deteriminism with preemption requires checkpointing the data iterator, and the checkpoints will be larger in size.
```
aireenmei, line 141: Could you add more details on what contributes to the size of the checkpoint? Would be good to provide a way to estimate the size to help the users make plans

aireenmei (changes requested):
Thanks for using and contributing to our repo!
I would like to understand the use case of this feature. You must have checked out our doc and know about the Grain pipeline which is optimized for ckpt support of data iterators. It's efficient and will save very small data iterator ckpt containing only indices, therefore it's the recommended way for use cases that need check-pointing data iterators. We would like to hear if there's any difficulty in adopting Grain for your use case?

And I tested the branch with this command  on v4-8 `python3 MaxText/train.py MaxText/configs/base.yml steps=20 per_device_batch_size=8.0 learning_rate=3e-4 enable_checkpointing=true base_output_directory=gs://aireenmei-multipod/tfds_ckpt dataset_path=gs://maxtext_dataset tfds_iter_checkpointing=True run_name=$(date +%m%d-$H%M) checkpoint_period=10` and got this error (no error if tfds_iter_checkpointing=False): https://gist.github.com/aireenmei/42a4c4e0dd8caed0b7ce8182f5ca8292

PR #957, opened 2024-10-08
Fix JetStream and MaxText integration

lukebaumann:
Makes an rng an optional argument for several MaxEngine functions that are required and used by JetStream.
Defaults to a rng with key 0


singh-mitali (commented):
Looks good to me. Would be great if we could add a test as well.

gagika (approved):
Thanks

gobbleturk (approved):

FanhaiLu1 (commented):
Thanks for making this fix! Looks good from jetstream side.

PR #958, opened 2024-10-08
Support `allow_split_physical_axes` for multislice hybrid meshes such as FSDP on ICI and DCN.

Obliviour:
allow_split_physical_axes is only supported for device meshes atm but we also should support this for hybrid meshes. This is useful when we want to use FSDP across DCN and ICI + split the physical axis.


khatwanimohit (approved):
Thanks Victor!

gobbleturk (approved):

PR #959, opened 2024-10-08
Set additional configs for A3+

michelle-yooh:
* set NCCL_SHIMNET_GUEST_CONFIG_CHECKER_CONFIG_FILE
 * added the path of cuda compat binary to LD_LIBRARY_PATH

### DIFF
--- a/gpu_multi_process_run.sh
+++ b/gpu_multi_process_run.sh
@@ -32,10 +32,15 @@ set_nccl_gpudirect_tcpx_specific_configuration() {
     export TF_CPP_MAX_LOG_LEVEL=100
     export TF_CPP_VMODULE=profile_guided_latency_estimator=10
     export XLA_PYTHON_CLIENT_MEM_FRACTION=0.85
+    shopt -s globstar nullglob
+    IFS=:$IFS
+    set -- /usr/local/cuda-*/compat
+    export LD_LIBRARY_PATH="${1+:"$*"}:${LD_LIBRARY_PATH}:/usr/local/tcpx/lib64"
+    IFS=${IFS#?}
+    shopt -u globstar nullglob
 
     if [[ "$USE_GPUDIRECT" == "tcpx" ]]; then
       echo "Using GPUDirect-TCPX"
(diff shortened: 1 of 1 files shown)


khatwanimohit (approved):
LGTM

gpu_multi_process_run.sh
```diff
@@ -32,10 +32,15 @@ set_nccl_gpudirect_tcpx_specific_configuration() {
     export TF_CPP_MAX_LOG_LEVEL=100
     export TF_CPP_VMODULE=profile_guided_latency_estimator=10
     export XLA_PYTHON_CLIENT_MEM_FRACTION=0.85
+    shopt -s globstar nullglob
+    IFS=:$IFS
+    set -- /usr/local/cuda-12.*/compat
```
bvandermoon, line 37: Should this be `cuda-*` instead of tying to `12`?

bvandermoon (commented):
Thanks for this fix Michelle. Were you able to see this working with a real workload?

bvandermoon (approved):

gpu_multi_process_run.sh
```diff
@@ -32,10 +32,15 @@ set_nccl_gpudirect_tcpx_specific_configuration() {
...
+    export LD_LIBRARY_PATH="${1+:"$*"}:${LD_LIBRARY_PATH}:/usr/local/tcpx/lib64"
```
jonb377, line 38: We shouldn't take a dependency on the compat binaries long-term, these don't have LTS guarantees and can break with cuda version updates. 

Unfortunately JAX has pulled ahead to 12.6 and GKE has 12.4, so I don't see a way around this at the moment without changing our build infra to pin on 12.4.

jonb377 (approved):
Thanks Michelle!

michelle-yooh:
> Thanks for this fix Michelle. Were you able to see this working with a real workload?

@bvandermoon Yes, please take a look at the bug tracking this issue.

gpu_multi_process_run.sh
michelle-yooh: I could do that. It should still work with cuda-12.* as well

PR #963, opened 2024-10-11
Fix various airflow tests

khatwanimohit:
* add kl-div to gemma tests
* remove asserts from legacy decode tests
* add base_output_directory to multihost decode tests in llama2-70b


gobbleturk:
You are the hero we need but don't deserve, TY!!

gobbleturk (approved):

PR #964, opened 2024-10-11
Add memory stats

zhenying-liu:
Log memory statistics, including host memory usage. When activation offloading is specified via remat_policy, display host memory usage for better insight into resource utilization.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -175,8 +175,6 @@ def write_metrics_to_tensorboard(writer, metrics, step, config, is_training=True
           f"total_weights: {metrics['scalar']['learning/total_weights']}, "
           f"loss: {metrics['scalar']['learning/loss']:.3f}"
       )
-      memory_stats = jax.local_devices()[0].memory_stats()
-      max_logging.log(f"Jax memory stats: {memory_stats}")
 
       if full_log and jax.process_index() == 0:
         max_logging.log(f"To see full metrics 'tensorboard --logdir={config.tensorboard_dir}'")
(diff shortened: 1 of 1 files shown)


MaxText/train.py
```diff
@@ -733,6 +733,14 @@ def train_loop(config, state=None):
   max_utils.close_summary_writer(writer)
   record_goodput(recorder, config, recorder.record_job_end_time if recorder else None)
   clear_buffered_metrics()
+  compiled = p_train_step.lower(state, example_batch, nextrng).compile()
+  compiled_stats = compiled.memory_analysis()
+  max_logging.log(
+      f"Output size: {compiled_stats.output_size_in_bytes}, "
+      f"temp size: {compiled_stats.temp_size_in_bytes}, "
+      f"argument size: {compiled_stats.argument_size_in_bytes}, "
+      f"host temp size: {compiled_stats.host_temp_size_in_bytes}, in bytes."
```
nouiz, line 742: Maybe change the order of lines to:
argument
output
temp
host temp.


?

zhenying-liu:
The two errors in the tests seem not related with my changes:

1. The required checklist:

`Error: No task list was present and requireChecklist is turned on`

2. The unit test failure is due to initializing the container: 

```
Error response from daemon: network github_network_7b9f4ac9971345e2a03b2a77b6a95d97 not found
Error: failed to start containers: 84530b682854e23e71abe192721a9328f836346036dda7f446ce1f6f7e61c217
```

gobbleturk:
> The two errors in the tests seem not related with my changes:
> 
> 1. The required checklist:
> 
> `Error: No task list was present and requireChecklist is turned on`
> 
> 2. The unit test failure is due to initializing the container:
> 
> ```
> Error response from daemon: network github_network_7b9f4ac9971345e2a03b2a77b6a95d97 not found
> Error: failed to start containers: 84530b682854e23e71abe192721a9328f836346036dda7f446ce1f6f7e61c217
> ```

The required checklist was added recently https://github.com/AI-Hypercomputer/maxtext/pull/1060, I assume if you sync to head you will get the checklist in the PR. However it seems that some train.py tests passed fine, so we can get this merged as is as long as there aren't conflicts (I will work on merging, which requires a codeowner and internal processes)

MaxText/train.py
```diff
@@ -750,6 +752,14 @@ def train_loop(config, state=None):
   max_utils.close_summary_writer(writer)
   record_goodput(recorder, config, recorder.record_job_end_time if recorder else None)
   clear_buffered_metrics()
+  compiled = p_train_step.lower(state, example_batch, nextrng).compile()
```
gobbleturk, line 755: One of our internal tests actually hit a compile time OOM with this - I realize you need the wrap this with a mesh and logical axis rules context manager like we do the real p_train_step, e.g. here https://github.com/AI-Hypercomputer/maxtext/blob/27f8a81620d633ab2a7c67e6bab063d3964df474/MaxText/train.py#L683

Also do we need to print this every step - shouldn't this be roughly constant across iterations (maybe slightly different with real data due to token masks)?

zhenying-liu: Thanks for the feedback. Addressed these two issues in the latest commit. Add the wrap and don't print every iteration.

MaxText/train.py
```diff
@@ -952,6 +952,17 @@ def train_loop(config, state=None):
   max_utils.close_summary_writer(writer)
   record_goodput(recorder, config, recorder.record_job_end_time if recorder else None)
   clear_buffered_metrics()
+  memory_stats = jax.local_devices()[0].memory_stats()
```
gobbleturk, line 955: Can you sync with main and remove this line? We now print this information as of https://github.com/AI-Hypercomputer/maxtext/pull/1110

gobbleturk (changes requested):

MaxText/train.py
zhenying-liu: Done. Sync-ed with main. Removed this line and the next line of its logging.

PR #965, opened 2024-10-11
Add qkv_proj to the saved name list

zhenying-liu:
Update QKV offloading and rename remat policy.

- Add qkv_proj to offloadable names due to query/key/value fusion
- Rename remat_policy from qkv_proj_offloaded to save_qkv_proj_offloaded

### DIFF
--- a/MaxText/layers/models.py
+++ b/MaxText/layers/models.py
@@ -310,10 +310,10 @@ class Decoder(nn.Module):
             "key_proj",
             "qkv_proj",
         )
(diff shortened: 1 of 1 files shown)


zhenying-liu:
For those models like "default" model don't have qkv_proj, it won't be recognized and there will be an error regarding it. So this CL needs to  be further polished.

zhenying-liu:
Will find a better way to specify the desired activation offloading.

PR #966, opened 2024-10-14
Add v6e to accelerator_to_spec_map for AOT support.

shralex:
Adding support for v6e, for train_compile. 

cl/685054499 and cl/686282966 added downstream support for v6e.

### DIFF
--- a/MaxText/accelerator_to_spec_map.py
+++ b/MaxText/accelerator_to_spec_map.py
@@ -29,19 +29,28 @@ class SystemCharacteristics:
   platform: str
   topology_name: Optional[str]
   chip_config_name: Optional[str]  # 'megacore' or 'default'
-  chips_per_host_bounds: Optional[tuple]
+  chips_per_host_bounds: Optional[tuple]  # number of chips on each host in each dimension.
   devices_per_slice: int
-  wrap: Optional[tuple]
+  wrap: Optional[tuple]  # wrap around for each dimension (i.e., locus type)
 
 
 UserFacingNameToSystemCharacteristics = {
-    # v5e
+    # v6e: one core per chip with 32 GB HBM
+    "v6e-1": SystemCharacteristics("tpu", "v6e:1x1", "default", (1, 1, 1), 1, (False, False, False)),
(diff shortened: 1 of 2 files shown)


MaxText/accelerator_to_spec_map.py
```diff
@@ -29,19 +29,28 @@ class SystemCharacteristics:
   platform: str
   topology_name: Optional[str]
   chip_config_name: Optional[str]  # 'megacore' or 'default'
-  chips_per_host_bounds: Optional[tuple]
+  chips_per_host_bounds: Optional[tuple]  # number of chips on each host in each dimension.
```
Obliviour, line 32: I think this is actually the chips_per_vm? For example a v6e host has two vms, each with 4 chips in a 2, 2, 1 arrangement.

MaxText/accelerator_to_spec_map.py
```diff
@@ -29,19 +29,28 @@ class SystemCharacteristics:
...
   devices_per_slice: int
-  wrap: Optional[tuple]
+  wrap: Optional[tuple]  # wrap around for each dimension (i.e., locus type)
 
 
 UserFacingNameToSystemCharacteristics = {
-    # v5e
+    # v6e: one core per chip with 32 GB HBM
+    "v6e-1": SystemCharacteristics("tpu", "v6e:1x1", "default", (2, 2, 1), 1, (False, False, False)),
```
Obliviour, line 39: v6e-1 has one chip so does it use the same (2,2,1) per host/vm `chips_per_host_bounds`.

Obliviour (approved):

MaxText/accelerator_to_spec_map.py
shralex: Your comment makes sense, but "chips_per_host_bounds" is something used internally, and my comment comes from there.

MaxText/accelerator_to_spec_map.py
shralex: Thanks, you're right. Fixed.

shralex (commented):
Thanks for the review

PR #970, opened 2024-10-18
Quantization fix for AQTEinsum 

mailvijayasingh:
3 fixes - 
i) rhs_quant_mode instead of lhs_quant_mode
ii) Ignoring 'AqtEinsum' in keys
iii) using jnp.einsum instead of AQTEinsum for elementwise operations.

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -597,9 +597,9 @@ class MoeBlock(nn.Module):
         )
       with jax.named_scope("w_sum"):
         weights_axis = ("activation_batch", "activation_length", "activation_exp")
-        output = self.get_einsum(rhs_mesh_axes=weights_axis)(
-            "BSEM,BSE -> BSM", intermediate_layer.astype(jnp.float32), weights.astype(jnp.float32)
-        ).astype(self.dtype)
+        output = jnp.einsum("BSEM,BSE -> BSM", intermediate_layer.astype(jnp.float32), weights.astype(jnp.float32)).astype(
+            self.dtype
+        )
       return output, None
 
   @nn.compact
--- a/MaxText/layers/quantizations.py
+++ b/MaxText/layers/quantizations.py
@@ -142,7 +142,7 @@ class AqtQuantization:
     aqt_einsum = functools.partial(
         aqt_flax.AqtEinsum(
             cfg=self.quant_dg,
-            lhs_quant_mode=self.quant_mode,
+            rhs_quant_mode=self.quant_mode,
             lhs_freeze_mode=aqt_flax.FreezerMode.NONE,
             rhs_freeze_mode=aqt_flax.FreezerMode.CALIBRATION_AND_VALUE,
             rhs_axis_metadata_wrapper=rhs_axis_metadata_wrapper,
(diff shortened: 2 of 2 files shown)


MaxText/layers/linears.py
```diff
@@ -493,7 +493,7 @@ def load_balance_loss(self, top_k_indices, logits):
     loss = jnp.mean(density * density_prob) * (self.num_experts**2) * self.config.load_balance_loss_weight
     return loss
 
-  def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = ()):
+  def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = (),):
```
RissyRan, line 496: Is this comma on purpose? similar comments for another line.

MaxText/layers/quantizations.py
```diff
@@ -136,13 +136,13 @@ def dot_general_cls(self, mesh_axes: Tuple[str, ...] = ()):
...
 
-  def einsum(self, mesh_axes: Tuple[str, ...] = ()):
+  def einsum(self, mesh_axes: Tuple[str, ...] = (),):
     """Returns einsum configured with aqt params."""
     rhs_axis_metadata_wrapper = self._get_rhs_axis_metadata_wrapper(mesh_axes)
     aqt_einsum = functools.partial(
         aqt_flax.AqtEinsum(
             cfg=self.quant_dg,
-            lhs_quant_mode=self.quant_mode,
+            rhs_quant_mode=self.quant_mode,
```
RissyRan, line 145: Do we need this change? I thought we just replace line related `jnp.einsum`?

MaxText/layers/quantizations.py
```diff
@@ -265,6 +264,8 @@ def _get_aqt_key_paths(aqt_vars):
       if "AqtDotGeneral" in d.key:
         pruned_keys.append(jax.tree_util.DictKey(key="kernel"))
         break
+      elif "AqtEinsum" in d.key:
```
RissyRan, line 267: Still need this? i.e. for inference only?

RissyRan (commented):
Thanks!

MaxText/layers/quantizations.py
vipannalla: This code is used by MaxEngine (inferenece only) to convert checkpoint from bf16 to int8. Still needed.

MaxText/layers/quantizations.py
vipannalla: From the code, we are passing weights as rhs, which are quantized. Can we do training run to confirm its not broken?

mailvijayasingh: talked to Zhiyu offline, looks like this was a bug as well. Can we do anything else to check?

MaxText/layers/linears.py
mailvijayasingh: fixed

vipannalla (approved):

MaxText/layers/quantizations.py
RissyRan: As long as end-to-end training works should be fine. Have you double checked?

RissyRan (approved):
LGTM! Ensure we have an end-to-end passing training test with `megablox=False quantization=int8`, thank you!

PR #972, opened 2024-10-21
custom device mesh for hybrid mesh ONLY

Obliviour:


MaxText/max_utils.py
```diff
@@ -360,6 +365,65 @@ def fill_unspecified_mesh_axes(parallelism_vals, target_product, parallelism_typ
...
+    b = []
+    for i in range(8):
+      b.append([])
+      for j in range(8):
+        a_i = i*2
+        a_j = j*2
+        # forms a ring of size 4
+        b[i].append([a[a_i,a_j], a[a_i,a_j+1], a[a_i+1,a_j+1], a[a_i+1,a_j]])
+    b = np.array(b)
+    b = np.reshape(b, (64,4))
```
Obliviour, line 413: I wish we could apply this to other mesh shapes besides 64,4 easily but I don't know the underlying rules enough to make this modular.

MaxText/max_utils.py
```diff
@@ -360,6 +365,65 @@ def fill_unspecified_mesh_axes(parallelism_vals, target_product, parallelism_typ
   return parallelism_vals
 
 
+def create_custom_64x4_device_mesh(
```
Obliviour, line 368: Ideally jax should support some type of logic for this, for background this function is copied + modified from the jax internal create_hybrid_device_mesh function.

MaxText/max_utils.py
```diff
@@ -398,12 +462,20 @@ def create_device_mesh(config, devices=None):
 
   if multi_slice_env:
     dcn_parallelism = fill_unspecified_mesh_axes(dcn_parallelism, num_slices, "DCN")
-    mesh = mesh_utils.create_hybrid_device_mesh(
+    if config.custom_mesh == "hybrid_ring_64x4":
+      assert config.ici_tensor_parallelism == 4, f"Invalid custom mesh chosen for mesh shape {ici_parallelism}"
```
gobbleturk, line 466: Can we make this more general - I think we might want this 64 x 4 mesh with expert parallelism = 4 (e.g. a lot of mixtral models only have 8 experts so we cant use the full 16 axis)

MaxText/max_utils.py
gobbleturk: Ya I think we should chat with compiler folks to understand this ourselves better. Are we able to create an efficient 8x32 mesh as well?

Obliviour (commented):
lgtm, but we need to be aware that this is hacky and not a long term solution.

MaxText/max_utils.py
khatwanimohit: I saw Tynan's CL where he created this change. The logic is a little different, no rings are created for 32x8 mesh shape.

MaxText/max_utils.py
khatwanimohit: asserting on fsdp now

gobbleturk (approved):
Thanks Mohit and Victor for this! Also Tynan/ other XLA folks to help us create this mesh

MaxText/max_utils.py
tengyifei: Wondering did folks already file a feature request to JAX and/or if there's already work ongoing to generate more efficient meshes by default?

PR #974, opened 2024-10-22
Add Train Compile Option to GPT3-175B v5p Scripts

anfals:


gobbleturk (approved):
LGTM! Was this config run with a custom mesh? I guess TP=8 is achievable with default 8x8x8 mesh

anfals:
> LGTM! Was this config run with a custom mesh? I guess TP=8 is achievable with default 8x8x8 mesh

Just a default 8x8x8 mesh

PR #975, opened 2024-10-22
Fix - Faster mixtral 8x22b conversion

rdyro:


jonb377 (approved):
LGTM, but I'm not very familiar with these tests. @RissyRan @A9isha could you also take a look?

rdyro:
> LGTM, but I'm not very familiar with these tests. @RissyRan @A9isha could you also take a look?

TY, @RissyRan and I are discussing this fix right now

end_to_end/tpu/mixtral/8x22b/1_test_mixtral.sh
```diff
@@ -37,7 +36,8 @@ echo "Wrote MaxText compatible scanned checkpoint to ${BASE_OUTPUT_PATH}/${MODEL
 fusermount -u "$PARAM_DIR"
 
 # Generate unscanned ckpt for efficient decoding test
```
RissyRan, line 38: Could you remove cmds from line#38, and leave a comment something like below?

Temporarily disable the generation of unscanned checkpoints to save time, as this process takes a couple of hours and is not utilized in subsequent tests.  The unscanned test is already running for 8x7b, so this removal should not impact overall testing.

RissyRan (commented):
Thanks for the fix!

RissyRan (approved):

PR #976, opened 2024-10-22
add SWA support to cudnn flash attn API

kocchop:
Exposed the sliding window attention to the cudnn flash attn API.
1. Needed to support GEMMA2 models
2. The causal masking is taken care of by TE
3. `head_dim = 256` support should be available in the next cudnn release, no code change on MaxText would be required

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -369,13 +369,15 @@ class AttentionOp(nn.Module):
   ) -> Array:
     """CUDNN Flash Attention with Transformer Engine.
     1. Stable API, supports GQA, supports SWA
-    2. Supports head_dim till 128; head_dim=256 will be supported from cudnn 9.6
-    3. Supports causal masking, TE will generate the needed mask for SWA, no need for additional mask generation
+    2. head_dim=256 is now supported from TE 1.2 with cudnn 9.5.0
     """
     # These imports are only meant to work in a GPU build.
     from transformer_engine.jax.flax.transformer import DotProductAttention  # pytype: disable=import-error
 
     _, _, _, head_dim = query.shape  # pylint: disable=unused-variable
+
+    # generate attn_mask
+    attn_mask = self.generate_attention_mask(query, key, decoder_segment_ids, model_mode)
 
     sliding_window_size = self.sliding_window_size
     if self.attention_type == AttentionType.LOCAL_SLIDING:
@@ -385,7 +387,7 @@ class AttentionOp(nn.Module):
         head_dim=head_dim,
         num_attention_heads=self.num_query_heads,
         num_gqa_groups=self.num_kv_heads,
-        attn_mask_type="causal",  # 'no_mask', 'padding', 'causal', or 'padding_causal'
+        attn_mask_type="padding_causal",  # 'no_mask', 'padding', 'causal', or 'padding_causal'
         attn_bias_type="no_bias",  # 'no_bias', 'pre_scale_bias' or 'post_scale_bias'
         attention_dropout=self.dropout_rate,
         dropout_rng_name="aqt",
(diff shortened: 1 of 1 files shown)


MaxText/layers/attentions.py
```diff
@@ -368,32 +368,35 @@ def cudnn_flash_attention(
...
+    if self.attention_type == AttentionType.LOCAL_SLIDING:
+      sliding_window_size = [self.sliding_window_size, 0]
 
     dpa_layer = DotProductAttention(
         head_dim=head_dim,
         num_attention_heads=self.num_query_heads,
         num_gqa_groups=self.num_kv_heads,
-        attn_mask_type="padding_causal",  # 'no_mask', 'padding', 'causal', or 'padding_causal'
-        attn_bias_type="NO_BIAS",  # 'no_bias', 'pre_scale_bias' or 'post_scale_bias'
+        attn_mask_type="causal",  # 'no_mask', 'padding', 'causal', or 'padding_causal'
```
bvandermoon, line 388: Why causal? This was recently set to padding_causal due to issue [878](https://github.com/AI-Hypercomputer/maxtext/issues/878)

kocchop: This was in place because we were generating the mask and passing it to the cudnn API. However, we no longer need to generate the `causal` mask as it is internally handled by TE for with and w/o SWA. Since on MaxText we are only using causal masking, we can just pass the `attn_mask_type` and it is sufficient.

gobbleturk: Isn't there a decoder_segment_ids mask component in self.generate_attention_mask that is no longer being passed here?

gobbleturk (commented):
Awesome, thank you for this change! I have a question regarding the mask, and we can have @ZhaoyueCheng review as well when he gets back (OOO today)

MaxText/layers/attentions.py
kocchop: Hi @gobbleturk, TE generates the `causal` mask internally even when we're using `SWA`, hence we do not need to explicitly generate the mask with `decoder_segement_ids`. However, if someone wants to use anything other than `causal` or `none` as mask, then they need to explicitly create the mask and pass it to the `DPA` api. As with MaxText, we mainly focus on `causal` masking, hence I removed the redundant mask generation which can be costly for large `seq_length`. Please let me know if it makes sense.

kocchop: I can also remove the passed `decoder_segment_ids` since its no longer used. Let me know if that is preferable

gobbleturk: > Decoder_segment_ids represents sequence packing - e.g. if we have a sequence length of 8 and a training examples of 3 tokens and 5 tokens so `decoder_segment_ids=[1,1,1,2,2,2,2,2]`. The decoder_segment_ids part of the mask represents the fact that tokens '1' do not attend to tokens '2' and vice-versa (Additionally we also apply a causal mask so that only earlier tokens attend to later ones)
> 
> Its on us that we don't have a good test for this currently...

kocchop: Hi @gobbleturk right now, we are working on adding the sequence packing support to the `DotProductAttention` api. When that arrives, we can pass the `decoder_segment_ids` directly to the api and it will take care of packing and mask generation. While its on its way, could we merge the PR? I was working on another PR to support `context parallelism`. Planning to submit that once this is through. 

Apart from the `decoder_segment_ids`, if you have any other comments, please let me know, thanks!

gobbleturk (approved):

MaxText/layers/attentions.py
gobbleturk: Okay, we will get this one merged

gobbleturk: After discussing this with some folks, we don't want to break any existing functionality. Can we wrap this new SWA support behind a config option until also supports a mask?

E.g. add a new attention_type=sliding_gpu [here](https://github.com/AI-Hypercomputer/maxtext/blob/54135f84741728e05fa69950cb70d3d2b1496950/MaxText/configs/base.yml#L173) and modify code like so:

if config.attention_type == sliding_gpu:
   use your new code
else:
   use existing code

kocchop: Hi @gobbleturk I made the necessary changes, we can still use the generated mask with SWA but would need the `if..else..` when we add `context parallelism`

gobbleturk: Awesome, thank you! We will merge this in now

PR #979, opened 2024-10-23
Add benchmark folders, include benchmark_runner, xla_flags_libs and model libs for trillium.

suexu1025:
* Added maxstar model libs. xla flag libs.
* Added example llama benchmark test code.

### DIFF
--- a/benchmarks/benchmark_runner.py
+++ b/benchmarks/benchmark_runner.py
@@ -1,0 +1,162 @@
+"""
+ Copyright 2024 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+""" Script to run a benchmark/benchmarks on exsiting xpk or QR nodes (to be implmented)
+                          ***** IMPORTANT *****
+This script will run specific tuned workload on specified hardwear and software enviroments
+Example usages:
+  python3 benchmark_runner.py  --project=<my-project> --zone=<zone> \
+    --cluster_name=<xpk_cluster_name> --base_output_directory=<output_gcloud_bucket> --device_type=v6e-256 --num_slices=1 --model_name="llama2_70b_4096" --libtpu_version=20241009 --base_docker_image=maxtext_base_image
+"""
+import argparse
+import importlib
+
+import maxtext_trillium_model_configs
+from maxtext_xpk_runner import BenchmarkRunner
+from maxtext_xpk_runner import HWConfig
+from maxtext_xpk_runner import SWconfig
+from maxtext_xpk_runner import xpk_benchmark_runner
+from maxtext_xpk_runner import XpkConfig
+
+
+def add_shared_arguments(custom_parser: argparse.ArgumentParser):
+  """Add shared arguments to the parser.
+
+  Args:
+    custom_parser: parser to add shared arguments to.
+  """
+  custom_parser.add_argument(
+      '--project',
+      type=str,
+      default=None, required=True,
+      help='GCE project name, defaults to "gcloud config project."',
+  )
+  custom_parser.add_argument(
+      '--zone',
+      type=str,
+      default=None, required=True,
+      help=(
+          'GCE zone, e.g. us-central2-b, defaults to "gcloud config '
+          'compute/zone." Only one of --zone or --region is allowed in a '
+          'command.'
+      ),
+  )
+  custom_parser.add_argument(
+      '--cluster_name',
+      type=str,
+      default=None, required=True,
+      help='cluster name The name of the cluster to run the job on. command.',
+  )
+  custom_parser.add_argument(
+      '--base_output_directory',
+      type=str,
+      default=None, required=True,
+      help='gcloud bucket to store arfifacts.',
+  )
+  custom_parser.add_argument(
+      '--device_type',
+      type=str,
+      default=None, required=True,
+      help='tpu device type command.',
+  )
+  custom_parser.add_argument(
+      '--num_slices',
+      type=int,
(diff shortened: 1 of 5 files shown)


benchmarks/llama3.1_v6e-256_benchmarks.py
```diff
@@ -0,0 +1,44 @@
...
+from maxtext_xpk_runner import HWConfig
+from maxtext_xpk_runner import SWconfig
+from maxtext_xpk_runner import xpk_benchmark_runner
+from maxtext_xpk_runner import XpkConfig
+
+
+DATE = '20241009'
+BASE_DOCKER_IMAGE = 'maxtext_base_image'
+
+ZONE = 'europe-west4'
```
Obliviour, line 12: Do you mind creating an argparser for the common arguments here: ZONE, PROJECT, CLUSTER_NAME, DEVICE_TYPE, NUM_SLICES so the user can plumb those in when running the script?

Obliviour: https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/parser/common.py#L26 for reference

benchmarks/llama3.1_v6e-256_benchmarks.py
```diff
@@ -0,0 +1,44 @@
...
+DEVICE_TYPE = 'v6e-256'
+NUM_SLICES = 1
+
+v6e_env_configs = SWconfig(
+    base_docker_image=BASE_DOCKER_IMAGE, libtpu_version=DATE
+)
+v6e_256_configs = HWConfig(num_slices=NUM_SLICES, device_type=DEVICE_TYPE)
+
+## todo: replace with llama3.1 model tests
+llama2_70b_4096 = BenchmarkRunner(
```
Obliviour, line 24: Nice, we can even use argparser to take in the model name argument. The benefit of this is we can have one common benchmark_runner.py file. Curious what you think of this approach?

Obliviour (commented):
Thank you Qinwen! Let's work over chat to figure which models we want to put here first. I have some comments about how to make it easier to script the benchmark running, the format overall looks great!

benchmarks/xla_flags_library.py
```diff
@@ -0,0 +1,89 @@
+"""XLA Flags."""
```
Obliviour, line 1: We should go through this file and clean it up, I will help here

benchmarks/llama3.1_v6e-256_benchmarks.py
suexu1025: Thanks for the suggestion, I think it's a good idea, pls find the updated code.

Obliviour (approved):
This generally LGTM using the custom arguments + one file for running this config is awesome. @gobbleturk can you please take a deeper look?

benchmarks/benchmark_runner.py
```diff
@@ -0,0 +1,127 @@
+import argparse
```
gobbleturk, line 1: Can we include comments at top of this file for what it is meant to do, example invocations, documentation for arguments?

benchmarks/maxtext_xpk_runner.py
gobbleturk, line 1: Can we also include comments at the top here - what is this file, example usage, arguments

benchmarks/maxtext_xpk_runner.py
```diff
@@ -0,0 +1,412 @@
...
+    cluster_config: XpkConfig,
+    buffer_size: int,
+):
+  config_tuning_params = ''
+  for key, value in model.tuning_params.items():
+    config_tuning_params += f'{key}={value} '
+
+  install_libtpu_cmd = ''
+  if libtpu_type == LibTpuType.V6E_NIGHTLY_WHEEL:
+    install_libtpu_cmd += f' pip install *${libtpu_date}*.whl &&'
```
gobbleturk, line 256: Is this actually specific to v6e? Can this option be renamed to NIGHTLY_WHEEL if its not?

benchmarks/maxstar_model_configs.py
```diff
@@ -0,0 +1,459 @@
+"""Shared Benchmark config for v6e orchestrations."""
```
gobbleturk, line 1: Can we have a more descriptive comment at the top.

If this file is meant to be v6e specific it should be renamed to maxtext_v6e_configs or similar. If not v6e specific we should change the comment to reflect this

gobbleturk: Generally I would prefer maxtext over maxstar, we can always change the name later if at some point add non-maxtext configs

gobbleturk (commented):
This looks very cool, thank you! Ideally we should include an e2e test which exercises maxtext_xpk_runner, at least confirming it runs

benchmarks/benchmark_runner.py
suexu1025: Added

benchmarks/maxtext_xpk_runner.py
suexu1025: Added top usage comments

benchmarks/maxtext_xpk_runner.py
suexu1025: I think we can safely remove it, as now v6e libtpu can use nightly version using jax-release channel. @Obliviour to confirm?

benchmarks/maxstar_model_configs.py
suexu1025: I think this is v6e specific tuned model configs. I will rename it ``` maxtext_v6e_model_configs ```. What do you think? @Obliviour

benchmarks/maxtext_xpk_runner.py
```diff
@@ -0,0 +1,456 @@
...
+    num_slices=options.num_slices, device_type=options.device_type
+)
+
+model_runner = BenchmarkRunner(
+    model_name=benchmark_model,
+    software_config=v6e_env_configs,
+    hardware_config=v6e_256_configs,
+)
+
+xpk_benchmark_runner(cluster_config, [model_runner])
```
gobbleturk, line 44: Ah I understand files much better now thank you. Can you add that all of these steps are run for you in benchmarks/benchmark_runner - in fact you can just point to the documentation there
"This file contains data classes and runner logic to execute the XPK runs triggered by benchmarks/benchmark_runner.py"

gobbleturk (approved):
Thank you for this Qinwen and Victor, it looks very cool!

PR #981, opened 2024-10-24
Adding support for building Maxtext with JAX Stable Stack GPU Training Image

parambole:
This PR adds support for building Maxtext Docker images with `Mode=stable_stack`, which consume JAX Stable Stack for GCP GPUs as the base image.

Tested:

Ran Maxtext XLML tests with the image created from this flow.

image


gobbleturk (commented):
Is this PR meant to replace pinned completely?

Should we replace pinned -> stable stack everywhere, e.g. in setup.sh and unit tests - or is this only meant to add support for building image?

If the latter we should keep both pinned and add new stable stack option to build_dependency_image so that setup.sh and build_dependency_image have the same set, but I would prefer the former and replace pinned by stable stack everywhere.

gobbleturk (approved):

bvandermoon (approved):
LGTM pending the failing unit tests

PR #986, opened 2024-10-25
Fix dropout and add a test

gobbleturk:
This PR does 3 things:
* Fixes a bug where static arnums not specified correctly for determinism
* Adds a test for dropout
* Changes the datatype of dropout in base.yml to a float instead of an int so it can be overriden in CLI with a float easily.


khatwanimohit (approved):

mattjj:
Thanks for this!

PR #987, opened 2024-10-25
Add a customizable remat policy for offloading tensors

khatwanimohit:


MaxText/layers/linears.py
```diff
@@ -247,12 +247,12 @@ def __call__(self, inputs, decode: bool = False, deterministic: bool = False):
         )(inputs)
         if cfg.activations_in_float32:
           x = x.astype(jnp.float32)
+        x = checkpoint_name(x, "mlp" + dense_name)
```
gobbleturk, line 250: I think we want this before the float32 cast to save 2x memory in the float32 case

gobbleturk: For some reason I can't comment on lines above - but we also should add a checkpoint_name after line 230 in the fused case (can be mlpwi or mlpwi_fused)

gobbleturk: Its somewhat ugly to have mlpwi_0 and mlpwi_1, we could re-use the same name which makes for prettier code but I can see some cases where we might actually want to do different things with the two of them since they are  so huge...

MaxText/configs/base.yml
```diff
@@ -149,9 +149,13 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
-# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'qkv_proj_offloaded', 'minimal_offloaded', 'save_out_proj' and 'full'.
+# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'qkv_proj_offloaded', 'custom_save_offload' 'minimal_offloaded', 'save_out_proj' and 'full'.
 # These options offer a trade-off between speed (fastest to slowest) and HBM usage (highest to lowest)
 remat_policy: 'full'
+# If custom_save_offload remat_policy is chosen, you can select tensors from the following list to offload or keep on HBM.
+# Tensors: [decoder_layer_input,mlpwi_0,mlpwi_1,mlpwo,query_proj,key_proj,value_proj,out_proj]
+tensors_to_save: ''
```
gobbleturk, line 157: I'm not a huge fan of this API - I would still prefer

decoder_layer_input: '' #device, host, or remat
mlpwi_0: '' # same
mlpwi_1: '' #same

Pros: no comma separated list, easy to define sweeping runs through config combinations, less magic strings user types which could lead to errors
Cons: more config fields, requires some parsing (probably need new function in maxtext_utils that takes as input the remat config options and outputs the remat policy)

Actually having some validator that the strings user types for either approach matches valid options
E.g. in your approach validate that "decoder_layer_input,mlpwi_0" is valid vs "decder_lyr_input,mlpwif" is a typo
or in my suggested appraoch that "device", "host", and "remat" are no typoed

MaxText/pyconfig.py
```diff
@@ -187,6 +187,8 @@ def validate_no_keys_overwritten_twice(keys1: list[str], keys2: list[str]):
 def _lists_to_tuples(l: list[Any]) -> Union[tuple[Any], list[Any]]:
   return tuple(_lists_to_tuples(x) for x in l) if isinstance(l, list) else l
 
+def _str_to_list(s: str) -> list[Any]:
```
gobbleturk, line 190: my recommendation of using "device" "host" "remat" would require slightly more parsing, looping thorugh the config fields

gobbleturk (commented):
Awesome, thank you removing the layer inputs! I still prefer the API we discussed earlier but am open minded if this comma separated list has benefits I haven't thought about

MaxText/configs/base.yml
khatwanimohit: Thanks for pointing out the pros and cons. Changed the API

MaxText/layers/linears.py
khatwanimohit: moved and assigned a checkpoint name for fused mlp

MaxText/configs/base.yml
```diff
@@ -149,9 +149,21 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
-# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'qkv_proj_offloaded', 'minimal_offloaded', 'save_out_proj' and 'full'.
+# Choose 'remat_policy' between 'minimal', 'save_dot_except_mlpwi', 'save_dot_except_mlp', 'save_qkv_proj', 'qkv_proj_offloaded', 'custom_save_offload' 'minimal_offloaded', 'save_out_proj' and 'full'.
```
gobbleturk, line 152: Can we rename to just "custom"? Some of our custom policies won't offload at all, just optimizing remat vs device

MaxText/configs/base.yml
```diff
@@ -149,9 +149,21 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
...
 # These options offer a trade-off between speed (fastest to slowest) and HBM usage (highest to lowest)
 remat_policy: 'full'
+# If custom_save_offload remat_policy is chosen, you can select tensors from the following list to offload on host memory, rematerialize or save on device memory.
+# Pick one of these options for following tensors: ['remat','save','offload']
```
gobbleturk, line 156: Prefer "device" to "save" -- since "offloading" is also saving the tensor, but just to a different location

MaxText/configs/base.yml
```diff
@@ -149,9 +149,21 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
...
+decoder_layer_input: 'save' # this tensor cannot be rematerialized due to an optimization in XLA
```
gobbleturk, line 157: Prefer: "this tensor cannot be rematerialized - it serves as periodic checkpoints that act as the remat start points"

gobbleturk (approved):
LGTM, thank you!

I left some suggestions for different variable names

PR #990, opened 2024-10-28
quantization bug fixes for MOE

mailvijayasingh:
i) Prune quantized params correctly
ii) Initialize kernels in serve mode appropriately
iii) Bug fix in constructing abstract parameters


vipannalla:
looks goo, can you run LLama2-7B loading bf16 ckpt and saving a int8 ckpt and then running benchmark? Just to make sure this doesn't break existing flow.

vipannalla (approved):
LGTM

PR #995, opened 2024-10-30
Better error message.

nouiz:


gobbleturk:
Thank you for this improvement! Have you run configurations that ran into this error despite being nearly fully sharded (only the norm tensor is replicated)? We are considering changing this to allow either user to specify tolerance or to calculate unsharded differently

nouiz:
We hit that error some time ago. But I forgot the exact case. So I can't help more for now about this.

PR #991, opened 2024-10-28
fix c4_mlperf

aireenmei:

PR #999, opened 2024-10-31
Fix c4_mlperf data

suexu1025:
Previously missing required argument, fixed by adding 
`add_bos=False, add_eos=False`


aireenmei:
Hi Qinwen, I have this https://github.com/AI-Hypercomputer/maxtext/pull/991 to fix the same issue

PR #1003, opened 2024-10-31
Fix multihost_runner.py kill script for v6e TPUs

ianmcampbell:
The script that searches for existing processes is hard coded for v5 TPUs. This tiny edit allows the runner to kill existing processes on v6e TPUs.


gobbleturk (approved):
Thanks!

PR #1006, opened 2024-11-01
optimizations for offline mlperf inference

sixiang-google:
-Offload detokenization to another thread
-Bulk decode steps


sixiang-google:
R:@vipannalla @singh-mitali

vipannalla (approved):

singh-mitali (changes requested):
I preferred to add the flags in the model/hw specific script as these may be different. We dont need the change to offline script.

singh-mitali (approved):

PR #1009, opened 2024-11-02
[WIP] Allow PP + megablox

gobbleturk:
Will add some tests here, perhaps even AOT PP over DCN before reivew. As of now just tested on real v4-8 that PP + Megablox does fine, which is expected since PP is just vmapping over stages so isn't fundamentally changing the input shardings to megablox

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -482,7 +482,7 @@ class _HyperParameters:
 
 def validate_megablox_parallelism(raw_keys):
   if raw_keys["megablox"] and (
-      using_sequence_parallelism(raw_keys) or using_pipeline_parallelism(raw_keys) or using_expert_parallelism(raw_keys)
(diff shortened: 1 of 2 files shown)


MaxText/pyconfig.py
```diff
@@ -482,7 +482,7 @@ def update_model_vars(base_config_path, raw_keys, config_name: str):
 
 def validate_megablox_parallelism(raw_keys):
   if raw_keys["megablox"] and (
-      using_sequence_parallelism(raw_keys) or using_pipeline_parallelism(raw_keys) or using_expert_parallelism(raw_keys)
+      using_sequence_parallelism(raw_keys) or using_expert_parallelism(raw_keys)
   ):
     raise ValueError("Currently we only support Megablox with data and tensor parallelism.")
```
RissyRan, line 487: You may want to update the message in ValueError. Looking forward to see some baseline metrics! Thanks Matt!

PR #1010, opened 2024-11-02
Name the initialize state module

gobbleturk:
Currently this module is named something like "unnamed_jit_function" when HLO dumped, this will rename it to "jit_initialize_state"


khatwanimohit (approved):
Thank you Matt :)

PR #1011, opened 2024-11-02
fix grain ckpt with new orbax

aireenmei:
b/373411373


gobbleturk (approved):
Awesome, thanks Aireen!

PR #1012, opened 2024-11-04
Correct vocab size for 8x22b

RissyRan:
# Description

Correct vocab size from 32000 to 32768, which aligned with right model [config](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/models/mixtral-8x22b.yml)


mailvijayasingh (approved):

vipannalla (approved):
LGTM, ty

PR #1013, opened 2024-11-04
fix data input on single CPU host

aireenmei:
b/374212354#comment15


RoshaniN:
Thanks Aireen, this helps unblock single-host dataloading on CPUs.

khatwanimohit (approved):

PR #1018, opened 2024-11-07
[WIP] Support DiLoCo training

jonb377:
DiLoCo training paper: https://arxiv.org/pdf/2311.08105

Adds preliminary support for DiLoCo training through DrJax.

### DIFF
--- a/MaxText/diloco.py
+++ b/MaxText/diloco.py
@@ -1,0 +1,131 @@
+import jax
+import jax.numpy as jnp
+import optax
+import pyconfig
+
+from flax import struct
+from typing import Any, Callable, Tuple
+
+
+Batch = Any
+State = Any
+Params = Any
+Loss = jnp.float32
+OptState = optax.OptState
+InnerOptStates = optax.OptState
+PRNGKey = jax.Array
+Step = jnp.int32
+
+
+class DiLoCoTrainState(struct.PyTreeNode):
+  inner_state: State
+  outer_params: Params
+  outer_opt_state: OptState
+  step: Step
+
+
+def build_diloco_train_step(
+    config: pyconfig.HyperParameters,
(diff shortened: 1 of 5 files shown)


MaxText/diloco.py
```diff
@@ -0,0 +1,131 @@
...
+    avg = jax.tree.map(lambda x: (x / config.num_diloco_replicas).astype(x.dtype), total)
+    return avg
+
+  @drjax.program(placements={'diloco': config.num_diloco_replicas})
+  def diloco_train_step(state, batch, prng):
+    # Batch shape is [pdbs, S]. To use in DiLoCo, we need to add a leading axis
+    # to vmap over.
+    per_replica_batch = config.global_batch_size_to_train_on // config.num_diloco_replicas
+    batch_shape = (config.num_diloco_replicas, per_replica_batch, -1)
+    batch = jax.tree.map(lambda x: x.reshape(batch_shape), batch)
```
zcharles8, line 103: In some versions I've written, I explicitly add this extra axis before passing to the train step, so that I can ensure that it is sharded over the diloco axis. I'm not certain if that's necessary though, but wanted to flag it for posterity.

MaxText/diloco.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # Broadcast the RNG across replicas
+    broadcast_rng = drjax.broadcast(prng)
+
+    inner_state, metrics = drjax.map_fn(
+      train_step,
+      (state.inner_state, batch, broadcast_rng)
+    )
+    avg_metrics = typed_reduce_mean(metrics)
```
zcharles8, line 112: One thing to call out is that differentiating between metrics that need to be summed and metrics that need to be averaged is kind of a pain.

E.g. if you're keeping track of total number of tokens seen, then a reduce_sum here is probably desired. I don't have any great solutions though without understanding better the structure of metrics.

PR #1026, opened 2024-11-12
[MoE] notebook for numerical verification

ZhiyuLi-goog:
# Description

In addition to the typo, looks like we are missing this [normalization](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mixtral/modeling_mixtral.py#L643) in [maxtext](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/layers/linears.py#L546), which is to re-scale top_k_weights to a normalized one where their sum equals 1. I was able to match layer outputs after adding the normalization. This normalization won't affect training/inferencing since this is a constant term (per each token), which won't change softmax probability. But we can still add it for better alignment.

Update: 
fix typo and add normalization for top_k_weights was covered in these PRs #1100 & #1064
Just add numerical verification sheet in this PR.

# Tests

Tested locally.

# Checklist

### DIFF
--- a/MaxText/scratch_code/mixtral-numerical-verification.ipynb
+++ b/MaxText/scratch_code/mixtral-numerical-verification.ipynb
@@ -1,0 +1,249 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bce1951a-8eef-4842-a70f-987b85a3240f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# installation\n",
+    "!pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu\n",
+    "!pip3 install tokenizers -U\n",
+    "!pip3 install transformers -U"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9769e847-d838-473d-8d32-1061b3e0f1c8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# go to maxtext/MaxText for library import\n",
+    "\n",
+    "current_dir = %pwd\n",
+    "working_dir = current_dir.replace(\"scratch_code\", \"\") \n",
+    "%cd $working_dir"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f1c108fc-d739-471d-9c64-c08151845f06",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


gobbleturk:
Thanks for the fix Zhiyu. I'll leave for @RissyRan to review

RissyRan:
Thanks Zhiyu for the fix! I am fine to add this normalization if this makes us convenient to compare weights. Could you add this `top_k_weights /= top_k_weights.sum(-1, keepdims=True)` before this line `weights = self.reshape_and_update_weights(top_k_weights, top_k_indices)`?  It was not used anywhere before that.

RissyRan:
I think we also need to add this normalization to somewhere [here]https://github.com/AI-Hypercomputer/maxtext/blob/061abd8a375b6b279eacf0ae867e05e4e08cc360/MaxText/layers/linears.py#L488) for dropping. `softmax_probs *= combined_expert_mask`

Have you checked the benchmark scores with and without this normalization? using correct ckpt on 8x22b or 8x7b.

ZhiyuLi-goog:
> Thanks Zhiyu for the fix! I am fine to add this normalization if this makes us convenient to compare weights. Could you add this `top_k_weights /= top_k_weights.sum(-1, keepdims=True)` before this line `weights = self.reshape_and_update_weights(top_k_weights, top_k_indices)`? It was not used anywhere before that.

Done.

ZhiyuLi-goog:
> I think we also need to add this normalization to somewhere [here]
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/061abd8a375b6b279eacf0ae867e05e4e08cc360/MaxText/layers/linears.py#L488
> 
> ) for dropping. `softmax_probs *= combined_expert_mask`
> Have you checked the benchmark scores with and without this normalization? using correct ckpt on 8x22b or 8x7b.

Have checked the benchmark scores the results are the same with and without this normalization.

Additionally,added a numerical verification notebook.
Currently able get ~0.1 tolerance in logits for both megablox=True or False.

RissyRan (approved):
Thank you!

ZhiyuLi-goog:
@gobbleturk could you review it at your convenience?

ZhiyuLi-goog:
> @gobbleturk could you review it at your convenience?

Hi, @gobbleturk I need code owner's review as the final step of this PR, thank you!

gobbleturk (approved):

PR #1035, opened 2024-11-13
Update sharding annotation for dropping

RissyRan:
# Description

Add missing TP sharding for dropping weights (after initialization).


gobbleturk (approved):
Thanks Ran!

PR #1036, opened 2024-11-14
Remove aqt einsum for dropping

mailvijayasingh:
i) Remove AQTEinsum for 1st step where inputs are masked with dispatch mask. This will lead to failure in loading weights that are pre-quantized without capacity_factor > 0. 
which should not be the case. As capacity factor is only forward pass, we should be able to load weights. 

ii) Remove AQTEinsum from last step where C axis is contracted, that is element wise op and AQT does not support the same. 

iii) Add math.ceil, so expert capacity per batch is never zero

iv) Add model_call_model in base.yml to selectively choose between training and inference

The correctness tests done - 
when given capacity_factor = 2 (which is almost dropless). The rouge scores for 8x22B are

{'rouge1': 44.5123, 'rouge2': 24.5477, 'rougeL': 31.3922, 'rougeLsum': 42.0639, 'gen_len': 1269606, 'gen_num': 1200}

as compared to capacity_factor = -1

{'rouge1': 44.1463, 'rouge2': 24.2765, 'rougeL': 31.2481, 'rougeLsum': 41.6197, 'gen_len': 1229222, 'gen_num': 1200} which makes sense.

tested 8x22B with capacity factor 1.5

{'rouge1': 43.9813, 'rouge2': 24.1528, 'rougeL': 30.8425, 'rougeLsum': 41.5053, 'gen_len': 1312280, 'gen_num': 1200}

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -30,6 +30,7 @@ import numpy as np
 from jax.ad_checkpoint import checkpoint_name
 from jax.experimental import shard_map
 import max_logging
+import math
 
 try:
   from jax.experimental.pallas.ops.tpu import megablox as mblx
@@ -48,6 +49,16 @@ bias_init = initializers.default_bias_init
 
 RMSNorm = normalizations.RMSNorm
 Quant = quantizations.AqtQuantization
+
+DISPATCH = "dispatch"
+COMBINE = "combine"
+
+
+def _get_model_call_mode(config):
+  if config.model_cal_mode == "inference":
+    return "inference"
+  else:
+    return None
 
 
 def _convert_to_activation_function(fn_or_string: Union[str, Callable[..., Any]]) -> Callable[..., Any]:
@@ -149,7 +160,10 @@ class DenseGeneral(nn.Module):
     output = compute_dot_general(inputs, kernel, axis, contract_ind)
 
     if self.use_bias:
-      bias_axes, bias_shape = self.kernel_axes[-len(features) :], kernel_shape[-len(features) :]
+      bias_axes, bias_shape = (
+          self.kernel_axes[-len(features) :],
+          kernel_shape[-len(features) :],
+      )
       bias = self.param(
           "bias",
           nn.with_logical_partitioning(bias_init, bias_axes),
@@ -379,7 +393,8 @@ class MoeBlock(nn.Module):
     reshaped_weights = jnp.reshape(weights, (-1, self.num_experts_per_tok))
     tensor_parallelism = self.config.ici_tensor_parallelism * self.config.dcn_tensor_parallelism
     reshaped_intermediate = jnp.reshape(
-        unsort_intermediate, (-1, self.num_experts_per_tok, self.config.emb_dim // tensor_parallelism)
+        unsort_intermediate,
+        (-1, self.num_experts_per_tok, self.config.emb_dim // tensor_parallelism),
     )
     with jax.named_scope("weight_sum"):
       matmul_precision = lax.Precision(self.config.matmul_precision)
@@ -405,7 +420,11 @@ class MoeBlock(nn.Module):
       inputs = inputs.astype(self.dtype)
       kernel = kernel.astype(self.dtype)
       output = mblx.gmm(
-          lhs=inputs, rhs=kernel, group_sizes=group_sizes, preferred_element_type=jnp.bfloat16, tiling=tile_size
+          lhs=inputs,
+          rhs=kernel,
+          group_sizes=group_sizes,
+          preferred_element_type=jnp.bfloat16,
+          tiling=tile_size,
       )
 
       if hs_shape[0] % pad_length:
@@ -450,7 +469,11 @@ class MoeBlock(nn.Module):
     # input of weights & indices: (batch_size, seq_len, num_experts_per_tok)
     # output of updated weights: (batch_size, seq_len, num_experts)
     update_weights = jnp.zeros((weights.shape[0], weights.shape[1], self.num_experts), dtype=self.dtype)
-    index_update = (jnp.arange(weights.shape[0])[:, None, None], jnp.arange(weights.shape[1])[:, None], indices)
+    index_update = (
+        jnp.arange(weights.shape[0])[:, None, None],
+        jnp.arange(weights.shape[1])[:, None],
+        indices,
+    )
     update_weights = update_weights.at[index_update].set(weights)
     return update_weights
 
@@ -458,7 +481,13 @@ class MoeBlock(nn.Module):
     # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
     batch_size, seq_len, _ = top_k_indices.shape
     tokens_per_batch = seq_len * self.num_experts_per_tok
-    expert_capacity_per_batch = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+    # this is to avoid expert_capacity_per_batch = 0
+    expert_capacity_per_batch = int(
+        max(
+            math.ceil(tokens_per_batch / self.num_experts) * self.config.capacity_factor,
+            self.config.capacity_factor,
+        )
+    )
     max_logging.log(f"Applying potential token dropping with a batch expert_capacity of {expert_capacity_per_batch}")
 
     # calculate expert mask and drop tokens if needed
@@ -476,7 +505,8 @@ class MoeBlock(nn.Module):
     expert_mask_fused = nn.with_logical_constraint(expert_mask_fused, ("activation_batch", None, None))
     expert_token_count_fused = jnp.cumsum(expert_mask_fused, axis=1)
     expert_token_count = jnp.reshape(
-        expert_token_count_fused, ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts))
+        expert_token_count_fused,
+        ((batch_size, seq_len, self.num_experts_per_tok, self.num_experts)),
     )
     expert_token_count = nn.with_logical_constraint(
         expert_token_count, ("activation_batch", "activation_length", None, None)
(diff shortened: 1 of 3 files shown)


vipannalla (approved):
Thanks for chasing this bug and finding a fix.

MaxText/layers/linears.py
```diff
@@ -458,7 +459,9 @@ def generate_masks(self, top_k_indices, softmax_probs):
     # calculate expert_capacity = (tokens_per_batch / num_experts) * capacity_factor
     batch_size, seq_len, _ = top_k_indices.shape
     tokens_per_batch = seq_len * self.num_experts_per_tok
-    expert_capacity_per_batch = int((tokens_per_batch / self.num_experts) * self.config.capacity_factor)
+    expert_capacity_per_batch = int(
```
RissyRan, line 462: Could you add a comment for future readers?

MaxText/layers/linears.py
```diff
@@ -548,16 +551,15 @@ def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
...
       mask_axes = ("activation_batch", "activation_length", None, None)
       dispatch_mask = nn.with_logical_constraint(dispatch_mask, mask_axes)
       combine_mask = nn.with_logical_constraint(combine_mask, mask_axes)
       loss = self.load_balance_loss(top_k_indices, softmax_probs)
       inputs = nn.with_logical_constraint(inputs, ("activation_batch", "activation_length", "activation_embed"))
       with jax.named_scope("dispatch"):
-        dispatch = self.get_einsum(rhs_mesh_axes=mask_axes)(
-            "BSM,BSEC -> EBCM", inputs, dispatch_mask, precision=matmul_precision
-        )
+        dispatch = jnp.einsum("BSM,BSEC -> EBCM", inputs, dispatch_mask, precision=matmul_precision)
```
RissyRan, line 562: @ZhiyuLi-goog could you help check if this impact the training performance?

MaxText/layers/linears.py
```diff
@@ -548,16 +551,15 @@ def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
 
     if self.config.capacity_factor > 0:
       # token dropping if needed
-      dispatch_mask, combine_mask = self.generate_masks(top_k_indices, softmax_probs)
+      weights = self.reshape_and_update_weights(top_k_weights, top_k_indices)
```
RissyRan, line 554: Why do you need this?

RissyRan: `self.reshape_and_update_weights` will update weights from shape `(batch_size, seq_len, num_experts_per_tok)` to `(batch_size, seq_len, num_experts)`.  And this `update_weights = update_weights.at[index_update].set(weights)` will heavily impact the performance.

vipannalla: +1 on performance impact. I just ran benchmark with this change and see perf slightly worse than the else block at CF=2. I believe this was mainly added for correctness.

vipannalla:
Removing "pull ready" and requesting review to avoid merging.

MaxText/layers/linears.py
RissyRan: It's a quite big drop in training (please see [here](https://screenshot.googleplex.com/8cUszXRtgCHPbvz)). I am trying to understand why `softmax_probs *= combined_expert_mask` does not work as expected. As reshape_and_update_weights() did is to update the weights based on topk. And `combined_expert_mask` comes from top_k_indices. We have a unit test (here) for dropping, and it seems we haven't caught this bug. Taking a look.

mailvijayasingh: If we dont apply this line we get garbage outputs. 
{'rouge1': 0.0447, 'rouge2': 0.0, 'rougeL': 0.0447, 'rougeLsum': 0.0447, 'gen_len': 114987, 'gen_num': 100}

and all outputs being almost same. Which looks like correct weights are not selected. And yes the CL is for correctness and we still have to look into preformance

MaxText/layers/linears.py
ZhiyuLi-goog: regression from 55.0% mfu to 54.5% after changing the 2 ops in to non quantized version:
* per_layer dispatch: from 2.455ms to 3.739ms
* per_layer combine: from 2.086ms to 3.213ms
The regression should be deterministic. The % of tflops from the 2 ops is small.

Recommend to add one more condition for this 2 ops:
```
if self.model_mode == common_types.MODEL_MODE_TRAIN
```

MaxText/layers/linears.py
RissyRan: I pulled the head, and add needed changes as discussed [here](https://github.com/AI-Hypercomputer/maxtext/pull/new/fix_weights). Without changing the weights, I was able to get

```
Successful requests: 100
Benchmark duration: 701.665692 s
Total input tokens: 21249
Total generated tokens: 29693
Request throughput: 0.14 requests/s
Input token throughput: 30.28 tokens/s
Output token throughput: 42.32 tokens/s
Mean TTFT: 439501.60 ms
Median TTFT: 460449.96 ms
P99 TTFT: 575143.83 ms
Mean TPOT: 4333.02 ms
Median TPOT: 3540.90 ms
P99 TPOT: 18755.05 ms
[nltk_data] Downloading package punkt to /home/ranran/nltk_data...
[nltk_data]   Package punkt is already up-to-date!

Results

{'rouge1': 44.5787, 'rouge2': 24.1149, 'rougeL': 31.4474, 'rougeLsum': 41.9571, 'gen_len': 110198, 'gen_num': 100}
```

Cmd:

```
python MaxText/maxengine_server.py MaxText/configs/base.yml load_parameters_path=gs://singhvijaya-maxtext-logs/8-22B-correct  max_prefill_predict_length=1024 max_target_length=2048 model_name=mixtral-8x22b ici_fsdp_parallelism=1 ici_autoregressive_parallelism=1 ici_tensor_parallelism=-1 scan_layers=false per_device_batch_size=20 attention=dot_product megablox=False quantization=int8 checkpoint_is_quantized=True quantize_kvcache=True capacity_factor=2


python JetStream/benchmarks/benchmark_serving.py   --tokenizer ~/maxtext/assets/tokenizer.mistral-v3 --save-result   --save-request-outputs   --request-outputs-file-path outputs.json   --num-prompts 100   --max-output-length 1024   --dataset openorca --run-eval True
```

Comparing the topk_weights & softmax, and I could see same values, and paste a few examples:

```
...
topk_weights: [[0 0 0 0 0 0.0966797 0.625 0]]
softmax_probs: [[0 0 0 0 0 0.0966797 0.625 0]]
topk_weights: [[0 0 0 0.291016 0 0 0 0.253906]]
softmax_probs: [[0 0 0 0.291016 0 0 0 0.253906]]
topk_weights: [[0 0 0.804688 0 0 0.0371094 0 0]]
softmax_probs: [[0 0 0.804688 0 0 0.0371094 0 0]]
topk_weights: [[0 0 0 0 0.172852 0 0.384766 0]]
softmax_probs: [[0 0 0 0 0.172852 0 0.384766 0]]
topk_weights: [[0 0 0 0 0 0.578125 0.152344 0]]
softmax_probs: [[0 0 0 0 0 0.578125 0.152344 0]]
topk_weights: [[0 0 0 0 0.202148 0.349609 0 0]]
softmax_probs: [[0 0 0 0 0.202148 0.349609 0 0]]
topk_weights: [[0 0 0 0.11084 0 0 0.617188 0]]
softmax_probs: [[0 0 0 0.11084 0 0 0.617188 0]]
topk_weights: [[0.417969 0 0 0 0.539062 0 0 0]]
softmax_probs: [[0.417969 0 0 0 0.539062 0 0 0]]
```

mailvijayasingh: I replicated Ran's setup on v6e-8 and still get 0 rouge on 20 samples
{'rouge1': 0.0, 'rouge2': 0.0, 'rougeL': 0.0, 'rougeLsum': 0.0, 'gen_len': 81347, 'gen_num': 20}

RissyRan: What's the difference between two weights? Could you pull that branch, and run again (instead of update one line)? I am also curious what's the delta here. With that debug info, we would know the difference between two weights in a single run (if cannot reproduce).

MaxText/layers/linears.py
```diff
@@ -50,6 +51,13 @@
 Quant = quantizations.AqtQuantization
 
 
+def _get_model_call_mode(config):
```
RissyRan, line 54: We probably could put this util [file](https://github.com/AI-Hypercomputer/maxtext/blob/6279bf1b0b211a28cb82e3d5442bdbeace023866/MaxText/max_utils.py#L4) that can be reused in other classes. Also we could have assertion to check if any other invalid value. Currently only accept `inference` I think.

MaxText/layers/linears.py
RissyRan: Is there an extra int? `int(int(....))`?

MaxText/layers/linears.py
```diff
@@ -515,7 +546,13 @@ def load_balance_loss(self, top_k_indices, logits):
     loss = jnp.mean(density * density_prob) * (self.num_experts**2) * self.config.load_balance_loss_weight
     return loss
 
-  def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = ()):
+  def get_einsum(self, rhs_mesh_axes: Tuple[Optional[str], ...] = (), einsum_name=None):
+
+    # the check is to prevent aqteinsum as einsum op for dispatch and combine einsums in ase when capacity_factor > 0
+    # this is necessary to load pre-quantized weights in case of inference
+    if self.config.model_call_mode == "inference" and (einsum_name == "dispatch" or einsum_name == "combine"):
```
RissyRan, line 553: Nit: defined `dispatch` and `combine` as variables to re-use in this file to avoid potential typo.

RissyRan (commented):
Other LGTM! Thanks!

MaxText/layers/linears.py
RissyRan: We filed a bug to compiler team.

MaxText/layers/linears.py
mailvijayasingh: Yes, there was an extra int, removed it.

MaxText/layers/linears.py
mailvijayasingh: Removed this as this is not needed. The compiler team is investigating the behaviour.

MaxText/layers/linears.py
mailvijayasingh: Discussed offline and added assertions in pyconfig.

MaxText/layers/linears.py
mailvijayasingh: done

MaxText/pyconfig.py
```diff
@@ -85,12 +85,19 @@ def validate_profiler_type(s: str) -> None:
     raise ValueError("Invalid profiler type was passed. Valid options ", valid_profiler_types)
 
 
+def validate_model_call_mode(s: str) -> None:
+  valid_model_call_modes = ("", "inference")
+  if s not in valid_model_call_modes:  # currently supported attention
+    raise ValueError("Invalid model call mode. Valid options ", valid_model_call_modes)
```
RissyRan, line 91: Nit:
raise ValueError(f"Invalid model call mode {s}. Valid options {valid_model_call_modes}")

It will help in debugging.

RissyRan (commented):
LGTM!

RissyRan (approved):

MaxText/layers/linears.py
vipannalla: Can you reuse the the existing model_mode thats gets passed at the top level -- https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/layers/mistral.py#L68 ?

PR #1037, opened 2024-11-14
Add custom 4x64 mesh for single slice

raymondzouu:


gobbleturk (commented):
Looks great, thanks Raymond! Is it possible to add an AOT test for this - either v6e if v6e AOT is working or v5e?

MaxText/max_utils.py
```diff
@@ -362,6 +362,21 @@ def fill_unspecified_mesh_axes(parallelism_vals, target_product, parallelism_typ
   return parallelism_vals
 
 
+def reshape_mesh_to_rings(a):
```
RissyRan, line 365: Thanks Raymond! I think once you merged the code, I will modify this function a little bit to be general to fit in 32x8 split ([this](b/379317193) for your reference).

gobbleturk (approved):
Thanks Raymond!

PR #1039, opened 2024-11-15
Scan and remat only the outer pipeline iteration loop over microbatches

gobbleturk:
We have observed extra remat when scanning + rematting both the outer pipeline loop and inner loop over stages, and strange undesirable behavior when scanning + rematting only the inner loop. We have found scanning + rematting the outer loop + saving the decoder layer inputs performs well, and is a common pattern among other pipeline implementations in other codebases.

Note this replaces https://github.com/AI-Hypercomputer/maxtext/pull/971

See b/373949783 for more.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -144,10 +144,22 @@ num_pipeline_repeats: -1
 # num_pipeline_microbatches must be a multiple of the number of pipeline stages. By default it is set to the number of stages.
 # Note the microbatch_size is given by global_batch_size / num_pipeline_microbatches, where global_batch_size = per_device_batch_size * num_devices
 num_pipeline_microbatches: -1
-scan_pipeline_iterations: True # This can be set independently of scan_layers, which is relevant when num_layers_per_pipeline_stage > 1.
 pipeline_delay_activation_forwarding: False # This delays the activation forwarding one loop iteration simplifying XLA's task of overlapping since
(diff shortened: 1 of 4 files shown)


MaxText/configs/base.yml
```diff
@@ -144,11 +144,23 @@ num_pipeline_repeats: -1
...
 
+# There are two loops for PP: 
+#  1)  Outer loop over microbatches (pipeline iterations)
+#  2)  Inner loop over layers (layers per stage)
+# We have observed extra remat when a remat policy and scanning is performed on both, and recommend the default
+# settings below of scanning and setting a remat policy only over the pipeline iterations.
+# It may be useful to do the reverse when the layers_per_stage is very large.
+# The below settings only have effect when using pipeline parallelism.
+scan_pipeline_iterations: True
+scan_layers_per_stage: False # The value of scan_layers_per_stage will override the value of scan_layers.
```
jonb377, line 159: It might be confusing to have two flags for scan_layers, especially since this one overrides the other. Is this the only way to have different default values with/without pipelining?

Otherwise maybe we can print a warning when the `scan_layers` value is overridden.

jonb377 (approved):
LGTM, thanks Matt!

MaxText/configs/base.yml
gobbleturk: Removed scan_layers_per_stage, it is up to the user to set scan_layers=False for PP

PR #1048, opened 2024-11-18
Add checkpoint topology discovery for the replicator service

xuefgu:
Add checkpoint topology discovery for the replicator service

This PR generates a yaml file to be consumed by the replicator service, when it is used. Follow-up PRs will include incorporating Orbax emergency checkpoint manager with replicator service and make the solution work E2E with the new checkpoint manager and the replicator service.


gobbleturk (approved):
Thanks Xuefeng!

PR #1051, opened 2024-11-20
Adding support for mixed precision drq.

singh-mitali:


MaxText/configs/quantization/intmp_subchannel.json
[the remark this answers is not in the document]
singh-mitali: These are just some samples illustrating how you can configure mixed precision. We will need model/accelerator specific optimized configs to be checked in - in relevant directories.

MaxText/layers/quantizations.py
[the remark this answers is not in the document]
singh-mitali: Sure - please add any changes required in subsequent Cl as you work with MOE.

MaxText/configs/quantization/README.md
```diff
@@ -0,0 +1,32 @@
...
+For example the following config defines 8-bit weight only quantization for all layers.
+{
+  "__default__": {"w_bits": 8}
+}
+
+# To configure mixed precision quantization, define the following.
+    1. A json file (e.g. example.json) in this directory with desired config
+    2. Set the following parameters defined in base.yml
+        quantization="intmp"
+        quant_cfg_path="<path_to_this_dir>/example.json"
```
vipannalla, line 32: Thank you for the documentation, much appreciated!

MaxText/configs/quantization/intmp_subchannel.json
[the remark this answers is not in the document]
vipannalla: Does it makes sense to prefix these with model_name (`llama`?) or `dense_llm`?

MaxText/tests/quantizations_test.py
```diff
@@ -94,6 +100,38 @@ def test_aqt_quantization(self):
     self.assertTrue(jnp.greater(jnp.max(inputs), jnp.max(res_dg[0][0])))
     # self.assertEqual(res_dg.dtype, np.dtype(np.float32))
 
+  def test_mixed_precision_config_int8w(self):
```
vipannalla, line 103: Thanks for adding tests.

MaxText/layers/quantizations.py
```diff
@@ -222,13 +244,11 @@ def _get_quant_config(config):
     return None
   if config.quantization == "int8":
     return _get_int8_quant_config(config)
-  if config.quantization == "int8w":
-    return _get_weight_only_quant_config(lhs_bits=None, rhs_bits=8)
-  if config.quantization == "int4w":
-    return _get_weight_only_quant_config(lhs_bits=None, rhs_bits=4)
   if config.quantization == "intmp":
```
vipannalla, line 247: Why are you removing support for int/8int4 weights only?

vipannalla: Ooh wait, are you expecting these to be added through MP configs? In that case can you also add a config for int4 weights only similar to int8 weight only config?

vipannalla: Also need to update the comments here -- https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L74-L75

PR #1054, opened 2024-11-21
Add 32x8 custom split

RissyRan:
# Description

Add 32x8 custom split:
* observed perf gain compared to hard split: b/379317193

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -456,7 +456,7 @@ enable_checkpoint_cloud_logger: True
 # Single-controller
 enable_single_controller: False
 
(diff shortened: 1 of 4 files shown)


gobbleturk (approved):
Awesome Ran, thanks for the improved refactor and additional tests!

raymondzouu (approved):
Thanks Ran, LGTM! I also saw a nice speedup in my workload with this custom mesh. It looks like the only thing is the new aot compile test is failing

RissyRan:
> Thanks Ran, LGTM! I also saw a nice speedup in my workload with this custom mesh. It looks like the only thing is the new aot compile test is failing

Yes, thanks Raymond for the info! I just fixed it, and trying to merge now.

PR #1056, opened 2024-11-21
More flexible offline benchmarking.

patemotter:
Changes hardcoded values to only be used if there are no existing environment variables.

### DIFF
--- a/MaxText/inference_mlperf/trillium/benchmarks_llama2-70b-trillium_2x4.sh
+++ b/MaxText/inference_mlperf/trillium/benchmarks_llama2-70b-trillium_2x4.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 # Run command:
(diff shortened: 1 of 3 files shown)


gobbleturk (approved):

vipannalla (approved):
LGTM

MaxText/inference_mlperf/trillium/benchmarks_llama2-70b-trillium_2x4.sh
```diff
@@ -58,8 +58,14 @@ fi
 
 export TOK_OUTLEN_MULTIPLIER=${token_multiplier}
```
singh-mitali, line 59: We also want to enable accuracy runs - so either just uncomment in this script or add an option for it.

singh-mitali (approved):
LGTM - some suggestions.

PR #1062, opened 2024-11-25
Quantize megablox

lenscloth:
# Description

Support quantization on megablox

- Current implementation can accelerate training but not serving, serving acceleration will be done in another PR.
- Tested correctness after apply quantization on megablox.

# Tests

End to end tests on mixtral-8x22b model with max engine.

# Checklist


RissyRan:
We shouldn't add `megablox` as a copy from [JAX repo](https://github.com/jax-ml/jax/tree/ef7df1ae7c258e402e7d379816b4d019a1643d44/jax/experimental/pallas/ops/tpu/megablox) into MaxText. We are good to work in a dev branch, but in a long term, we need to merge related quantization changes into JAX repo. So MaxText could directly import it.

RissyRan:
> We shouldn't add `megablox` as a copy from [JAX repo](https://github.com/jax-ml/jax/tree/ef7df1ae7c258e402e7d379816b4d019a1643d44/jax/experimental/pallas/ops/tpu/megablox) into MaxText. We are good to work in a dev branch, but in a long term, we need to merge related quantization changes into JAX repo. So MaxText could directly import it.

Discussed with JAX team, and they prefer we fork the megablox kernel to avoid dependency issue to use standalone install of JAX. So we will have a "copy" of gmm implementation in MaxText. Thanks Wonpyo!

lenscloth:
@RissyRan This branch now supports setting different precision for lhs and rhs. Now it support
- Int8
- Int8w
- Int4w

MaxText/kernels/megablox/common.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2024 The JAX Authors.
```
RissyRan, line 1: Nit: update to `Copyright 2024 Google LLC`

MaxText/kernels/megablox/__init__.py
```diff
@@ -0,0 +1 @@
+from kernels.megablox.ops import gmm
```
RissyRan, line 1: Nit: copyright

MaxText/kernels/megablox/gmm.py
```diff
@@ -0,0 +1,819 @@
+# Copyright 2024 The JAX Authors.
```
RissyRan, line 1: Same

MaxText/layers/linears.py
```diff
@@ -30,9 +31,13 @@
 from jax.ad_checkpoint import checkpoint_name
 from jax.experimental import shard_map
 import max_logging
+import max_utils
+from aqt.jax.v2 import aqt_tensor
 
 try:
-  from jax.experimental.pallas.ops.tpu import megablox as mblx
+  # from jax.experimental.pallas.ops.tpu import megablox as mblx
```
RissyRan, line 38: We previously add this comment as it throws error when running on GPU. Now we moved this into MaxText, the ImportError will not be triggered anymore. We could directly put `from kernels import megablox as mblx`.

MaxText/layers/linears.py
```diff
@@ -389,7 +395,14 @@ def unpermute(self, intermediate, sorted_selected_experts, weights):
           reshaped_weights.astype(jnp.float32),
           precision=matmul_precision,
       )
-    return output.reshape(-1, self.config.max_target_length, self.config.emb_dim // tensor_parallelism).astype(self.dtype)
+    updated_batch = int(self.config.per_device_batch_size * jax.device_count() // self.config.ici_fsdp_parallelism)
```
RissyRan, line 398: An assertion may be needed if the value is indivisible cc @mailvijayasingh @ZhiyuLi-goog

MaxText/layers/linears.py
```diff
@@ -404,27 +417,56 @@ def gmm(inputs, kernel, group_sizes):
 
       inputs = inputs.astype(self.dtype)
       kernel = kernel.astype(self.dtype)
+
+      quantization_config: str = self.config.quantization
```
RissyRan, line 421: Will this look a little bit better?

```
quantization_config = self.config.quantization

quantization_types = {
    "int8": (jnp.int8, jnp.int8),
    "int8w": (None, jnp.int8),
    "int4w": (None, jnp.int4),
}

lhs_quantize_dtype, rhs_quantize_dtype = quantization_types.get(
    quantization_config, (None, None)
)
```

Also could you help add a coment for int8w and int4w [here](https://github.com/AI-Hypercomputer/maxtext/blob/58048291bb380474111b18166d54f413071fc7bd/MaxText/configs/base.yml#L73-L75)? Thank you!

MaxText/layers/linears.py
```diff
@@ -648,6 +690,21 @@ def __call__(self, inputs):
 
     if cfg.megablox:
       max_logging.log("Running MoE megablox implementation.")
+      # This is called only during tracing. This is to invoke creation of quantized tensor.
```
RissyRan, line 693: Could we wrap line# 693 to 707 into a helper function? i.e. something like process_weights_for_quantization() or others?

MaxText/kernels/megablox/gmm.py
```diff
@@ -0,0 +1,819 @@
...
+    del k_i, group_offsets, group_ids, group_offset
+    return m_tile_ids[grid_id], n_i
+
+  out_block_spec = pl.BlockSpec((tm, tn), out_transform_indices)
+  if existing_out is None:
+    in_out_block_spec: Any = None
+    input_output_aliases = {}
+  else:
+    in_out_block_spec = out_block_spec
+    num_inputs = 6
```
RissyRan, line 522: What's this hardcode 6 here?

RissyRan (commented):
Thank you Wonpyo! 

Created a diff gmm [file](https://diff.googleplex.com/#key=zPXKwKpLxuhn) for future reference.

RissyRan (commented):
Others LGTM! @ZhiyuLi-goog could you also help take a look once you are back?

MaxText/kernels/megablox/common.py
lenscloth: Resolved thanks Ran!

MaxText/kernels/megablox/__init__.py
lenscloth: Done

MaxText/layers/linears.py
lenscloth: Sounds better, I've added more logic to avoid conflict with unsupported quantization configurations.

MaxText/kernels/megablox/gmm.py
lenscloth: In Pallas, you can write back to the given input kernel using its reference.
In order to do so, you need to pass the mapping between the input and output so that you can write back.
If given `existing_out` becomes a output reference to write back, and the argument index of `existing_out` is six to the kernel. (Note that, group_metadata contains three tensor internally).

The  `6` was hardcoded in the original kernel implementation as well.

lenscloth:
Resolved some of comments @RissyRan would you do review one more time?

ZhiyuLi-goog:
Thank you @lenscloth for the awesome PR. LGTM!

ZhiyuLi-goog (approved):

MaxText/layers/linears.py
```diff
@@ -629,6 +666,33 @@ def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
         )
       return output, None
 
+  def retrieve_quantized_weight(self, inputs, gate_logits) -> tuple[QTensor, QTensor, QTensor]:
+    # This is called only during tracing. This is to invoke creation of quantized tensor inside AqtEinsum.
+    # After jit, this will become no-op and will not affect performance.
+    _ = self.capped_dense(inputs, gate_logits)
```
RissyRan, line 672: I am not quite understand this logic, but it seems we have to do so :)

MaxText/layers/linears.py
```diff
@@ -643,12 +707,8 @@ def __call__(self, inputs):
         name="gate",
         matmul_precision=self.config.matmul_precision,
     )(inputs)
-
-    w0_kernel, w1_kernel, wo_kernel = self.generate_kernels(cfg.num_experts, cfg.emb_dim, cfg.mlp_dim)
-
+    max_logging.log("Running MoE megablox implementation.")
```
RissyRan, line 710: Thank you for the change! Hmm.. but the logic seems not right now.

Could we revert it back and change it to something below?

```
    if cfg.megablox:
      max_logging.log("Running MoE megablox implementation.")
      A helper function to do whatever we like to process quantized weights. 
      return self.megablox(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
    else:
      max_logging.log("Running MoE matmul implementation.")
      w0_kernel, w1_kernel, wo_kernel = self.generate_kernels(cfg.num_experts, cfg.emb_dim, cfg.mlp_dim)
      return self.dense_matmul(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
```
I like your previous change, but just wrap those lines into a helper function. Happy to discuss offline :)

MaxText/kernels/megablox/gmm.py
RissyRan: I see. thanks for the explanation! Let's add `The argument index of existing_out is six to the kernel. Note that, group_metadata contains three tensor internally which makes its index six.` as a comment there?

RissyRan (commented):
Thank you! For code style check, you could workaround using [this](https://screenshot.googleplex.com/AJfDSmUMKJ7RTnD).

MaxText/layers/linears.py
lenscloth: revert it back to the previous code, and added helper function.

MaxText/layers/linears.py
lenscloth: AqtEinsum is a flax compact class which initializes variable only when einsum is called. 
So to invoke creation of quantized variable in `aqt` collection, we manually invoke AqtEinsum and read parameters using `aqt` collection. This is kind of unaovidable with current deign of AqtEinsum unfortunately.

MaxText/layers/linears.py
RissyRan: Thanks! After moving the logs, it WILL LGTM!

RissyRan (approved):
lgtm!

MaxText/kernels/__init__.py
gobbleturk, line 1: Is this file necessary?

MaxText/layers/linears.py
```diff
@@ -404,7 +403,14 @@ def unpermute(self, intermediate, sorted_selected_experts, weights):
           reshaped_weights.astype(jnp.float32),
           precision=matmul_precision,
       )
-    return output.reshape(-1, self.config.max_target_length, self.config.emb_dim // tensor_parallelism).astype(self.dtype)
+    updated_batch = int(self.config.per_device_batch_size * jax.device_count() // self.config.ici_fsdp_parallelism)
```
gobbleturk, line 406: Does this cover all cases - what if we use data_parallelism instead of FSDP (or a mix of them) or even fsdp_transpose?

gobbleturk: Can we get the relevant sizes from the input shape instead of based on the config?

gobbleturk (changes requested):
Looks great! Only request to get sizes based on inputs instead of config (ideally we can remove the original reference to tensor_parallelism as well)

MaxText/layers/linears.py
lenscloth: @ZhiyuLi-goog made change there. Would you take a look on this?

MaxText/kernels/__init__.py
lenscloth: It is not necessary, but it's more like legacy from forking.

Removing this code will require us to change few import statements.
If this violate MaxText policy I will delete the file. Please let me know.

@gobbleturk

gobbleturk: No this is fine, this makes sense

MaxText/layers/linears.py
ZhiyuLi-goog: Thank you for the comment @gobbleturk 

> Does this cover all cases - what if we use data_parallelism instead of FSDP (or a mix of them) or even fsdp_transpose?

Updated to cover data and fsdp (or a mix of them). Not yet fsdp_transpose since it is simply not used in MoE models.

> Can we get the relevant sizes from the input shape instead of based on the config?

It seems not easy to have a relevant sizes solution covering both prefill and decoding situations and the prefill length might be variant.  And this is the trick working pretty well in our inferencing experiments.

@RissyRan @mailvijayasingh feel free to chime in if you have new ideas.

RissyRan: > Can we get the relevant sizes from the input shape instead of based on the config?

Yeah, this is the difference between training and inference. The batch size will be different in prefill and generate stage (even we specify per_device_batch_size as a fixed value). In megablox, we need extra permute and unpermute operations, and those need to be calculated manually to map back the right shape. 

@gobbleturk what's your suggestion on this? basically we need an indicator to see if current is prefill or generate stage in inference, and then reshape the right shape, based on different batch size.

lenscloth: I have updated this to inference output shape based on input tensor.
But I am not fully familiar with this code, and afraid I might made a wrong implementation here.

@ZhiyuLi-goog Can you take a look on this.
Also @gobbleturk Can you review this branch? It would be great to merge this branch before holiday.

cc. @RissyRan

gobbleturk: I was thinking something like sharded_batch_shape = inputs.shape[0]. I think prefill and generate are two separate jits (unsure) so this sharded_batch_shape just takes exactly the value it needs

gobbleturk (approved):
LGTM, Thanks Wonpyo!

PR #1066, opened 2024-11-26
Add Pallas GPU decode attention in Maxtext inference

tohaowu:
# Description

This PR enables the use of Pallas GPU attention from `jax.experimental.pallas.ops.gpu` in Maxtext GPU inference.

Previously, Maxtext GPU decode did not support flash attention, explicitly raising a ValueError when `attention_kernel` was set to `cudnn_flash_te` in the decoder. This PR adds `pallas_gpu`, allowing the use of Pallas GPU attention for improved efficiency and speed.

This change is crucial for leveraging the performance benefits of Pallas GPU attention in Maxtext GPU inference. It allows for faster and more efficient decoding.

**Specific implementation details:**

- Imported the necessary modules from `jax.experimental.pallas.ops.gpu`.
- Replaced the existing attention calculation in decode with calls `pallas_decode_attention.gqa`
- Adjusted parameters within these functions to optimize performance.
- Ensured compatibility with the decoder's attention calculations.

**Shortcomings and future improvements:**

- Further performance tuning might be required for specific use cases.
- Explore the possibility of integrating other optimized attention kernels from `jax.experimental.pallas.ops.gpu`.

FIXES: b/366477266, b/375269239

# Tests

**Instructions to reproduce:**

- Run the decoder with the modified code using the following command:
`XLA_FLAGS="--xla_dump_to=/llama-2-1vm-2024-11-14-05-28/HLO_dumps/ --xla_gpu_enable_latency_hiding_scheduler=true --xla_gpu_enable_triton_gemm=true --xla_gpu_graph_level=0 --xla_gpu_enable_highest_priority_async_stream=true --xla_gpu_all_reduce_combine_threshold_bytes=134217728 --xla_gpu_all_gather_combine_threshold_bytes=134217728 --xla_gpu_reduce_scatter_combine_threshold_bytes=67108864 --xla_gpu_enable_pipelined_all_gather=true --xla_gpu_enable_pipelined_reduce_scatter=true --xla_gpu_enable_pipelined_all_reduce=true --xla_gpu_enable_while_loop_double_buffering=true --xla_gpu_enable_triton_softmax_fusion=false --xla_gpu_enable_all_gather_combine_by_dim=false --xla_gpu_enable_reduce_scatter_combine_by_dim=false --xla_disable_hlo_passes=rematerialization" XLA_PYTHON_CLIENT_MEM_FRACTION=0.9 TF_FORCE_GPU_ALLOW_GROWTH=true CUDA_VISIBLE_DEVICES=0 python3 MaxText/decode.py MaxText/configs/base.yml load_parameters_path=gs://tohaowu/maxtext/direct_generate_param_only_checkpoint_2024-11-14-05-21/checkpoints/0/items run_name=runner_decode_finetuned_2024-11-14-05-21 base_output_directory=gs://tohaowu/maxtext per_device_batch_size=1 model_name=llama2-7b ici_autoregressive_parallelism=1 max_prefill_predict_length=1024 max_target_length=2048 attention=pallas_gpu scan_layers=false hardware=gpu async_checkpointing=false`

# Checklist


MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
...
+    assert q_heads % kv_heads == 0
+
+    q_reshaped = q.reshape(batch_size, q_length, kv_heads, q_heads // kv_heads, head_dim)
+    k_transposed = jnp.swapaxes(k, 1, 2)
+
+    # Re-compute attention weights to get local_max and local_sum
+    attn_weights = (q_reshaped @ k_transposed.transpose(0, 1, 3, 2)) * sm_scale
+    local_max = jnp.max(attn_weights, axis=-1, keepdims=True)
+    local_exps = jnp.exp(attn_weights - local_max)
+    local_sum = jnp.sum(local_exps, axis=-1, keepdims=True)
```
Rifur13, line 295: Calculating `local_sum` and `local_max` like this defeats the whole purpose of using a kernel. A better idea is to have the kernel output these values. I'll add it to `jax.experimental.pallas.ops.gpu.pallas_decode_attention`, and then you can remove these lines.

Or better yet, do we absolutely need them? `pallas_attention_kernel` below returns `out, None, None`

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
...
+    # Reshape gqa's output to include q_length
+    local_out = local_out_gqa.reshape(batch_size, q_length, q_heads, head_dim)
+
+    # Reshape local_max and local_sum to match Maxtext requirements
+    local_max = local_max.reshape(batch_size, q_length, q_heads, 1)
+    local_sum = local_sum.reshape(batch_size, q_length, q_heads, 1)
+
+    return local_out, local_max, local_sum
+
+  def pallas_attention_kernel(self, query, key, value, decoder_segment_ids, model_mode):
```
Rifur13, line 315: A bit of nit but I would just call this directly instead of wrapping it in a function.

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
...
+    k_transposed = jnp.swapaxes(k, 1, 2)
+
+    # Re-compute attention weights to get local_max and local_sum
+    attn_weights = (q_reshaped @ k_transposed.transpose(0, 1, 3, 2)) * sm_scale
+    local_max = jnp.max(attn_weights, axis=-1, keepdims=True)
+    local_exps = jnp.exp(attn_weights - local_max)
+    local_sum = jnp.sum(local_exps, axis=-1, keepdims=True)
+
+    # Reshape q to match gqa's expected shape
+    q_for_gqa = q.reshape(batch_size * q_length, q_heads, head_dim)
```
Rifur13, line 298: I think you're doing too many reshapes above.

The kernel expects `q` to be of size `[batch_size, num_q_heads, head_dim]`, because `q_seq_len` is 1 during decoding. And you also don't need to split the q heads, it's already done inside the kernel [1]. 

[1] https://github.com/jax-ml/jax/blob/03b6945ee7ddeef4d9e2f24aad926e9b57904814/jax/experimental/pallas/ops/gpu/decode_attention.py#L316

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
...
+    # Reshape local_max and local_sum to match Maxtext requirements
+    local_max = local_max.reshape(batch_size, q_length, q_heads, 1)
+    local_sum = local_sum.reshape(batch_size, q_length, q_heads, 1)
+
+    return local_out, local_max, local_sum
+
+  def pallas_attention_kernel(self, query, key, value, decoder_segment_ids, model_mode):
+    """Pallas MHA kernel for prefill stage."""
+    sm_scale = 1.0 / math.sqrt(query.shape[-1])
+    out = pallas_attention.mha(query, key, value, decoder_segment_ids, sm_scale=sm_scale, causal=True if model_mode == common_types.MODEL_MODE_AUTOREGRESSIVE else False) # Set causal=True for autoregressive mode
```
Rifur13, line 318: You can set it to False directly because you already check if `model_mode == common_types.MODEL_MODE_AUTOREGRESSIVE` on line 261.

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
                            Use `dot_product` instead."""
         )
       return self.cudnn_flash_attention(query, key, value, decoder_segment_ids, model_mode), None, None
+    elif self.attention_kernel == "pallas_gpu":
```
Rifur13, line 260: I think it’s worth rethinking the kernel name, the fact that it uses pallas is an implementation detail. To me, it’s really a `flash` kernel, and then we can pick **which** one based on the config. For ex: ragged or not, autoregressive or not, tpu or gpu.

I propose something like:

```
if use_ragged_attention and model_mode == common_types.MODEL_MODE_AUTOREGRESSIVE:
      if lengths is None:
        lengths = jnp.sum(decoder_segment_ids, axis=-1)

      impl = self.tpu_ragged_attention if config.device_name = ‘tpu’ else self.gpu_ragged_attention

      return impl(query, key, value, lengths, self.ragged_block_size)
```

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
...
+      start_idx=None, kv_seq_len=None,
+      sm_scale: float | None = None,
+      block_h: int = 16,
+      block_k: int = 128,
+      k_splits: int = 16,
+      num_warps: int | None = None,
+      num_stages: int = 2,
+      grid: tuple[int, ...] | None = None,
+      interpret: bool = False,
+      debug: bool = False,
```
Rifur13, line 281: I would remove the params if you're only using the default values anyways. The default values are already assigned in the kernel, so you're essentially overwriting them.

MaxText/layers/attentions.py
```diff
@@ -255,9 +257,67 @@ def apply_attention(
                            Use `dot_product` instead."""
         )
       return self.cudnn_flash_attention(query, key, value, decoder_segment_ids, model_mode), None, None
+    elif self.attention_kernel == "pallas_gpu":
+      if model_mode == common_types.MODEL_MODE_AUTOREGRESSIVE:
+        return self.gqa_wrapped_maxtext_compatible(query, key, value)
```
Rifur13, line 262: Just a fyi, you're not passing in `kv_seq_len`, which is equivalent to `lengths` in the tpu ragged attention kernel, so you're not taking advantage of the `ragged` part of the kernel.

MaxText/layers/attentions.py
tohaowu: Gleb, thanks for your input!

Could you please incorporate local_max and local_sum functions into jax.experimental.pallas.ops.gpu.pallas_decode_attention, similar to how it's implemented in this code : 

https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/kernels/ragged_attention.py#L348

MaxText/layers/attentions.py
```diff
@@ -255,10 +270,57 @@ def apply_attention(
...
```
gabeweisz, line 275: self.gqa_wrapped_maxtext_compatible does not appear to be defined anywhere

MaxText/layers/attentions.py
```diff
@@ -229,7 +231,11 @@ def apply_attention(
       if lengths is None:
         lengths = jnp.sum(decoder_segment_ids, axis=-1)
 
-      return self.ragged_attention(query, key, value, lengths, self.ragged_block_size)
+      if jax.devices()[0].platform == "tpu":
+        impl = self.tpu_ragged_attention
+      if jax.devices()[0].platform == "gpu":
```
gangji, line 236: elif

gangji (commented):
LGTM. Thanks!

vipannalla (approved):
looks good.

jyj0w0 (approved):

jyj0w0 (approved):

jyj0w0 (approved):

jyj0w0 (approved):

jyj0w0 (approved):

jyj0w0 (approved):

PR #1073, opened 2024-12-02
disable tf GPU visibility

keshavb96:
Maxtext crashes when running with TFDS data with a GPU OOM. This is traced back to Tensorflow spawning processes that allocate memory on GPUs. This patch is a current work around to hide GPUs from TF.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -44,6 +44,7 @@ import optimizers
 import profiler
 import pyconfig
 import pathwaysutils  # pylint: disable=unused-import
+import tensorflow as tf
 
 from vertex_tensorboard import VertexTensorboardManager
 # Placeholder: internal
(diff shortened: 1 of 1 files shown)


wang2yn84 (approved):
Thank you! Just to keep the record, it aligns with https://www.tensorflow.org/guide/gpu#limiting_gpu_memory_growth.

MaxText/train.py
```diff
@@ -755,6 +756,9 @@ def train_loop(config, state=None):
 
 def main(argv: Sequence[str]) -> None:
   jax.config.update("jax_default_prng_impl", "unsafe_rbg")
+  # TF allocates extraneous GPU memory when using TFDS data
+  # this leads to CUDA OOMs. WAR for now is to hide GPUs from TF
```
gobbleturk, line 760: What does WAR here mean? I assume its an acronym?

gobbleturk (approved):

MaxText/train.py
wang2yn84: Work around I guess?

nouiz: Exact.

PR #1081, opened 2024-12-04
Support JAX_VERSION for nightly mode on GPU

bvandermoon:
# Description

Allow users to pass in `JAX_VERSION` for nightly mode. This lets us select a specific nightly build for JAX, e.g. `JAX_VERSION=0.4.36.dev20241109` will use nightly versions of `jax` and `jaxlib` from 11/9/2024.

This currently only works for GPUs since the way we are installing nightly jaxlib for TPUs is a little different (and possibly outdated). But the same general pattern should also work for TPUs in the future.

# Tests

Successfully built a Docker image using `MODE=nightly` with `JAX_VERSION=0.4.36.dev20241109` and with `JAX_VERSION` unset

# Checklist

### DIFF
--- a/docker_build_dependency_image.sh
+++ b/docker_build_dependency_image.sh
@@ -19,6 +19,8 @@
 # bash docker_build_dependency_image.sh DEVICE={{gpu|tpu}} MODE=stable_stack BASEIMAGE={{JAX_STABLE_STACK BASEIMAGE FROM ARTIFACT REGISTRY}}
 # bash docker_build_dependency_image.sh MODE=nightly
 # bash docker_build_dependency_image.sh MODE=stable JAX_VERSION=0.4.13
+# Nightly build with JAX_VERSION for GPUs. Available versions listed at https://storage.googleapis.com/jax-releases/jax_nightly_releases.html:
+# bash docker_build_dependency_image.sh DEVICE=gpu MODE=nightly JAX_VERSION=0.4.36.dev20241109 # Note: this sets both jax-nightly and jaxlib-nightly 
 
 # Enable "exit immediately if any command fails" option
 set -e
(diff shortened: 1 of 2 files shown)


docker_build_dependency_image.sh
```diff
@@ -19,6 +19,8 @@
 # bash docker_build_dependency_image.sh DEVICE={{gpu|tpu}} MODE=stable_stack BASEIMAGE={{JAX_STABLE_STACK BASEIMAGE FROM ARTIFACT REGISTRY}}
 # bash docker_build_dependency_image.sh MODE=nightly
 # bash docker_build_dependency_image.sh MODE=stable JAX_VERSION=0.4.13
+# Nightly build with JAX_VERSION for GPUs. Available versions listed at https://storage.googleapis.com/jax-releases/jax_nightly_releases.html:
```
gobbleturk, line 22: Is there a reason this is GPU specific? I can imagine use cases where this is useful for TPUs as well

gobbleturk (commented):
I think we can have this be GPU/TPU agnostic? I have sometimes wanted a nightly verison for some TPU runs to repro an earlier nightly run

docker_build_dependency_image.sh
bvandermoon: Yes, the TPU path is installing nightly `jaxlib` using a different command:

```
pip3 install --pre -U jaxlib -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
```

It looks like this command is outdated and versions newer than 0.4.30 don't appear to be available: https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html. We can make the change for TPU also, but we need to change the installation pattern like here: https://jax.readthedocs.io/en/latest/installation.html#jax-nightly-installation. Left this out of this PR to keep the scope down. And also because I needed this on GPUs already since there seemed to be an issue with a nightly build I was using last week

gobbleturk: Thanks for the extra info, this makes sense!

gobbleturk (approved):

PR #1085, opened 2024-12-06
Modify axes orders for optimal pipeline parallelism performance

gobbleturk:
# Description

Modify axes orders for optimal pipeline parallelism performance

# Tests

DP + TP over ICI, PP + DP over DCN [XPROF](https://xprof.corp.google.com/trace_viewer/mattdavidow-9102876018318431015?hosts=gke-tpu-ba604869-1j7l&host_index=0) - swapping the order is necessary to
1) Avoid an XLA bug where DCN axis must be first b/339009148
2) Stage should be first due to reshape of microbatch for the "activation_and_embed" batch

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -231,9 +231,6 @@ mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'tensor', 'ex
 logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_batch_no_exp', ['data', 'fsdp', 'fsdp_transpose']],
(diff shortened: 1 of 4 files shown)


RissyRan (approved):
Thanks Matt for the fix!

Is this a right summary:

* Without PP over DCN, the data parallelism config needs to be the 1st one to yield better perf;
* With PP over DCN, the stage config needs to be the 1st (before data) to yield better perf, so we need a helper function to swap.

gobbleturk:
> Thanks Matt for the fix!
> 
> Is this a right summary:
> 
> * Without PP over DCN, the data parallelism config needs to be the 1st one to yield better perf;
> * With PP over DCN, the stage config needs to be the 1st (before data) to yield better perf, so we need a helper function to swap.

Yes! For some configs it is perf we are chasing but also "core halted" errors when the DCN sharded axes is not first in the list

PR #1086, opened 2024-12-06
add Sliding Window Attention support to cudnn flash attn api

kocchop:
# Description

Start with a short description of what the PR does and how this is a change from the past.

This PR enables usage of Sliding Window Attention with cudnn flash attn api. Sliding window attention only works for causal masking for now. Hence, inserted necessary checks to choose correct mask type. TE only accepts generated mask when `mask_type="padding_causal"` but this is not supported with SWA yet. With the PR, we can now run the class of Gemma models with proper GPU flash attention now.

# Tests

I ran gemma2-9b model which uses sliding window attn and head_dim=256; also tested llama2-7b model to make sure the existing path doesn't break

# Checklist


MaxText/layers/attentions.py
```diff
@@ -393,32 +393,42 @@ def cudnn_flash_attention(
...
+    2. Head_dim = 256 is also supported from TE-1.12 stable release
     """
     # These imports are only meant to work in a GPU build.
     from transformer_engine.jax.flax.transformer import DotProductAttention  # pytype: disable=import-error
 
     _, _, _, head_dim = query.shape  # pylint: disable=unused-variable
 
-    # generate attn_mask
-    attn_mask = self.generate_attention_mask(query, key, decoder_segment_ids, model_mode)
+    sliding_window_size = self.sliding_window_size
```
wang2yn84, line 404: Nit ,better assign sliding_window_size to None when for non sliding window case to better align with the requirement of the API, wdyt?

MaxText/layers/attentions.py
```diff
@@ -393,32 +393,42 @@ def cudnn_flash_attention(
...
         dropout_rng_name="aqt",
         dtype=self.dtype,
         float32_logits=self.float32_logits,
         qkv_layout="BSHD_BSHD_BSHD",  # 'BS3HD', 'BSHD_BS2HD' or 'BSHD_BSHD_BSHD'
         scale_factor=1.0 / math.sqrt(head_dim),
         transpose_batch_sequence=False,
+        window_size=sliding_window_size,
     )
-    return dpa_layer(query, key, value, mask=attn_mask)
+    if self.attention_type == AttentionType.LOCAL_SLIDING:
```
wang2yn84, line 428: nit: just to keep the code clean if we set mask to either None or attn_mask, wdyt?

MaxText/layers/attentions.py
kocchop: sure, will change to None

kocchop:
@wang2yn84 PTAL, I made the suggested change

wang2yn84:
> @wang2yn84 PTAL, I made the suggested change

Great! Thank you! Lgtm.

wang2yn84 (approved):

gobbleturk (approved):

PR #1094, opened 2024-12-10
Add Pathways Support to Benchmark Runner

SujeethJinesh:
# Description

Add Pathways support to Benchmark Runner. Makes running benchmarks easier for Pathways scale and long running testing. Also added pathways specific configs for long running tests for now.

# Tests

Ran at scale and in long running tests

# Checklist

### DIFF
--- a/benchmarks/benchmark_runner.py
+++ b/benchmarks/benchmark_runner.py
@@ -28,6 +28,7 @@ import maxtext_trillium_model_configs
 from maxtext_xpk_runner import BenchmarkRunner
 from maxtext_xpk_runner import HWConfig
 from maxtext_xpk_runner import SWconfig
+from maxtext_xpk_runner import PathwaysConfig
 from maxtext_xpk_runner import xpk_benchmark_runner
 from maxtext_xpk_runner import XpkConfig
 
@@ -86,6 +87,11 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
           'llama2_7b_4096',
           'llama2_70b_4096',
           'llama2_70b_4096_real_data',
+          'llama2_70b_4096_pw_long_run',
+          'llama2_70b_4096_real_data_pw_long_run',
+          'llama2_70b_4096_pw_rd_tfds',
+          'llama2_70b_4096_synthetic_pw_lr',
+          'llama2_70b_4096_synthetic',
           'llama3_70b_8192',
           'llama3_1_405b_8192_fsdp_dcn',
           'mixtral_8x7b_dropped',
@@ -103,6 +109,11 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
           'llama2_7b_4096 '
           'llama2_70b_4096 '
           'llama2_70b_4096_real_data '
+          'llama2_70b_4096_pw_long_run '
+          'llama2_70b_4096_real_data_pw_long_run '
+          'llama2_70b_4096_pw_rd_tfds '
+          'llama2_70b_4096_synthetic_pw_lr '
+          'llama2_70b_4096_synthetic '
           'llama3_1_405b_8192_fsdp_dcn '
           'mixtral_8x7b_dropped '
           'mixtral_8x7b_dropped_int8 '
@@ -124,6 +135,51 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
       default='maxtext_base_image',
       help='version of base docker image to be benchmarked command.',
   )
+  custom_parser.add_argument(
+      '--pathways_server_image',
+      type=str,
+      default=(
+          'us-docker.pkg.dev/cloud-tpu-v2-images-dev/pathways/server:latest'
+      ),
+      help='version of pathways server image to be benchmarked command.',
+  )
+  custom_parser.add_argument(
+      '--pathways_proxy_image',
+      type=str,
+      default='us-docker.pkg.dev/cloud-tpu-v2-images-dev/pathways/proxy_server:latest',
+      help='version of pathways proxy image to be benchmarked command.',
+  )
+  custom_parser.add_argument(
+      '--pathways_runner_image',
+      type=str,
+      default='us-docker.pkg.dev/cloud-tpu-v2-images-dev/pathways/maxtext_jax_stable:latest',
+      help='version of pathways runner image to be benchmarked command.',
+  )
+  custom_parser.add_argument(
+      '--use_pathways',
+      type=bool,
+      default=False,
+      help='whether to use pathways or not.',
+  )
+  custom_parser.add_argument(
+      '--xpk_path',
+      type=str,
+      default='~/xpk',
+      help='path to xpk dir.',
+  )
+  custom_parser.add_argument(
+      '--priority',
+      type=str,
+      default='medium',
+      help='Priority the XPK workload should run with.',
+  )
+  custom_parser.add_argument(
+      '--max_restarts',
+      type=int,
+      default=0,
+      help='Number of restarts to attempt.',
+  )
+
 
 def main() -> None:
   parser = argparse.ArgumentParser(
@@ -139,11 +195,19 @@ def main() -> None:
       num_slices=options.num_slices,
       device_type=options.device_type,
       base_output_directory=options.base_output_directory,
+      priority=options.priority,
+      max_restarts=options.max_restarts,
   )
 
   v6e_env_configs = SWconfig(
       base_docker_image=options.base_docker_image,
       libtpu_version=options.libtpu_version,
+      pathways_config=PathwaysConfig(
+          use_pathways=options.use_pathways,
+          server_image=options.pathways_server_image,
+          proxy_image=options.pathways_proxy_image,
+          runner_image=options.pathways_runner_image,
+      ),
   )
 
   v6e_256_configs = HWConfig(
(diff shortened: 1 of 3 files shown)


benchmarks/benchmark_runner.py
```diff
@@ -122,6 +127,57 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
...
+      default=False,
+      help='whether to use pathways or not.',
+  )
+  custom_parser.add_argument(
+      '--xpk_path',
+      type=str,
+      default='~/xpk',
+      help='path to xpk dir.',
+  )
+  custom_parser.add_argument(
```
suexu1025, line 162: num_steps is defined in model configs. Adding extra parser for steps will caused non-intentional replacement.

benchmarks/maxtext_xpk_runner.py
```diff
@@ -61,13 +71,15 @@ class HWConfig:
...
   base_docker_image: str
+  pathways_config: PathwaysConfig
 
 
 @dataclasses.dataclass
 class BenchmarkRunner:
   model_name: str
   hardware_config: HWConfig
   software_config: SWconfig
+  num_steps: int
```
suexu1025, line 82: same for `num_steps`

benchmarks/maxtext_xpk_runner.py
```diff
@@ -257,21 +269,25 @@ def run_command_with_updates(command, task, verbose=True) -> int:
 
 
 def build_user_command(
```
suexu1025, line 271: currently it's adding extra pathway configs by default, we can either add maxtext_xpk_pw_runner.py or another build_user_commands function. what do you think? @SujeethJinesh

SujeethJinesh: The `PathwaysConfig` is only used if `--use_pathways` is passed in. This is intentional because it would make it easier for us to have similar configs and code changes as McJAX but just make drop in replacements to support pathways. I think having a new function or file would bifurcate the code too much and make it harder to share configs if needed over time.

benchmarks/benchmark_runner.py
SujeethJinesh: Sorry I don't see it in the model configs, I'm looking in this file, and they don't appear to be defined? Are you referring to the base.yml?

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -320,6 +359,45 @@ class MaxTextModel:
...
+        "use_iota_embed": True,
+        "dataset_path": "gs://max-datasets-rogue",
+        "dataset_type": "synthetic",
+        "reuse_example_batch": 1,
+        "profiler": "xplane",
+        "sa_block_q": 1024,
+        "sa_block_q_dkv": 2048,
+        "sa_block_q_dq": 2048,
+
+        # Additional tuning params for pathways long running test.
```
suexu1025, line 383: can we move pathways long running test related configurations like ``` "metrics_file": "metrics.txt",
        "goodput_upload_interval_seconds": 30,
        "enable_pathways_goodput": True,
        "enable_checkpoint_cloud_logger": True,
        "enable_single_controller": True, ``` in maxtext_xpk_runner? like new function `xpk-command-for-pathway`, [maxtext_trillium_model_configs.py](https://github.com/AI-Hypercomputer/maxtext/pull/1094/files#diff-5c68d18bef03e76cee74107f324af52adf487414f181baf577cdc1aad0618e8f) is supposed to have different benchmarks. In that case, it will be easy for pathway to run different benchmarks. What do you think?

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -320,6 +359,45 @@ class MaxTextModel:
     ),
 )
 
+llama2_70b_4096_pw_long_run = MaxTextModel(
+    model_name="llama2-70b-4096-pw-lr",
```
suexu1025, line 363: benchmark name can be `llama2-70b-4096-checkpoint`?

benchmarks/benchmark_runner.py
```diff
@@ -155,9 +219,10 @@ def main() -> None:
       model_name=benchmark_model,
       software_config=v6e_env_configs,
       hardware_config=v6e_256_configs,
+      num_steps=options.num_steps,
   )
 
-  xpk_benchmark_runner(cluster_config, [model_runner])
+  xpk_benchmark_runner(cluster_config, [model_runner], options.xpk_path)
```
suexu1025, line 225: will need to add default value for xpk path for `xpk_benchmark_runner `

benchmarks/maxtext_xpk_runner.py
```diff
@@ -406,7 +443,7 @@ def run_xpk_workload(
   return run_command_with_updates(command, 'Run XPK workload', cluster_config)
 
 
-def xpk_benchmark_runner(cluster_config: XpkConfig, benchmarks: list[BenchmarkRunner]):
+def xpk_benchmark_runner(cluster_config: XpkConfig, benchmarks: list[BenchmarkRunner], xpk_path: str):
```
suexu1025, line 446: default value for xpk_path, make it optional.

benchmarks/maxtext_xpk_runner.py
```diff
@@ -327,11 +338,12 @@ def generate_xpk_workload_cmd(
...
     base_output_directory: str,
     buffer_size: int,
+    num_steps: int = 100,
+    xpk_path: str = '~/xpk',
+    pathways_config: PathwaysConfig = None,
 ):
   """Generates a command to run a maxstar model on XPK."""
-  num_steps = 20
   time.localtime()
-  test_purpose_name = f'maxstar-benchmarks-{model.model_name}-{libtpu_version}'
```
suexu1025, line 334: can we add back  
```
 if "steps" not in model.tuning_params:
    num_steps = model.tuning_params["steps"]
```
so that workload specified steps will not be overwritten.

suexu1025 (commented):
a few comments

benchmarks/maxtext_trillium_model_configs.py
SujeethJinesh: These configs are actually for the specific test we're running (long running tests). I think keeping it as this config would make more sense because we may have different tests that are just for perf, or non goodput related, etc. I wouldn't want to bifurcate the code too much, it's much easier for us to just add a specialized config for Pathways and flip on the specific flags for that test I think.

benchmarks/maxtext_trillium_model_configs.py
SujeethJinesh: We don't want the name to be too long because it may overrun the 40 character limit for XPK :(

benchmarks/benchmark_runner.py
SujeethJinesh: Done

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: Done

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: I'm not quite sure this does what's intended. I believe this code is better. I've added it in.

```
  if not num_steps and 'steps' in model.tuning_params:
    num_steps = model.tuning_params['steps']
  elif not num_steps:
    num_steps = 20
```

benchmarks/benchmark_runner.py
SujeethJinesh: From the other comments, we can leave this here so the user can override it if they want, but if they don't specify it, then fallback to what's in tuning_params.

SujeethJinesh:
Thank you @suexu1025, resolved your comments

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: (Minor Update, ignore the comment above). Made the change when we're doing the tuning in particular.

benchmarks/benchmark_runner.py
SujeethJinesh: Also in the other configs, none seem to have num_steps defined, so people are just using the default at the moment. I've edited the code to make it the default still.

suexu1025: We don't want to make it `num_steps` mandatory as it is already defined in some of the workload. can we make it optional ? Only used if needed by user otherwise, it will use either  tuning_params or default 20 num of steps.

benchmarks/benchmark_runner.py
```diff
@@ -155,9 +222,10 @@ def main() -> None:
       model_name=benchmark_model,
       software_config=v6e_env_configs,
       hardware_config=v6e_256_configs,
+      num_steps=options.num_steps,
```
suexu1025, line 225: num_steps is benchmark specific, can we change it to `xpk_benchmark_runner` optional argment instead of HW config.

benchmarks/maxtext_xpk_runner.py
suexu1025: same,  it's better in ``` xpk_benchmark_runner(cluster_config: XpkConfig, benchmarks: list[BenchmarkRunner], num_steps =0) ```

benchmarks/benchmark_runner.py
SujeethJinesh: Removed num_steps

suexu1025 (approved):
LGTM, thanks for addressing all the comments.

PR #1102, opened 2024-12-17
Add options to reduce manual work in e2e JetStream benchmark tests.

fenghuizhang:
# Description

Add the `inference_server` and `inference_benchmark_test` configs so folks don't have manually modify those files for benchmark testing.

The default behavior shouldn't change. For testing, folks can override the config values to control the behavior of the server.

# Tests

TBD

# Checklist

- [ x] I have performed a self-review of my code.
- [ x] I have necessary comments in my code, particularly in hard-to-understand areas.

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -219,6 +219,9 @@ def maybe_initialize_jax_distributed_system(raw_keys):
 
   For CPUs, we call jax.distributed.initialize() explicitly, with the specified arguments.
   """
+  if raw_keys["inference_benchmark_test"]:
+    # Disable initialization for inference benmark test.
+    return
   if raw_keys["compile_topology"]:
     # Don't initialize jax distributed with AOT compilation
     return
@@ -531,7 +534,7 @@ def create_device_mesh(config, devices=None):
   if devices is None:
     devices = jax.devices()
   num_devices = len(devices)
-  num_slices = config.num_slices
+  num_slices = 1 if config.inference_benchmark_test else config.num_slices
   num_devices_per_slice = num_devices // num_slices
 
   multi_slice_env = num_slices > 1
--- a/MaxText/maxengine_server.py
+++ b/MaxText/maxengine_server.py
@@ -37,7 +37,7 @@ from jetstream.core import server_lib, config_lib
 def main(config):
   # No devices for local cpu test. A None for prefill and a None for generate.
   devices = server_lib.get_devices()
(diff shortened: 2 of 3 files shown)


MaxText/max_utils.py
```diff
@@ -525,7 +528,7 @@ def create_device_mesh(config, devices=None):
   if devices is None:
     devices = jax.devices()
   num_devices = len(devices)
-  num_slices = config.num_slices
+  num_slices = 1 if config.inference_benchmark_test else config.num_slices
```
vipannalla, line 531: We will add multi-slice support for inference soon (~Q1) and this code will be confusing when one runs inference benchmark on that setup :). 
 Is this mainly to reduce the time take for the benchmark? Or is this causing some issues with the test?
 
 We also have Jetstream specific config for inference, you can set `num_slices = 1` here and use it for testing -- https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/inference_jetstream.yml

MaxText/maxengine_server.py
```diff
@@ -37,7 +37,7 @@
 def main(config):
   # No devices for local cpu test. A None for prefill and a None for generate.
   devices = server_lib.get_devices()
-  server_config = maxengine_config.get_server_config("MaxtextInterleavedServer", config)
+  server_config = maxengine_config.get_server_config(config.inference_server, config)
```
vipannalla, line 40: +1

MaxText/max_utils.py
zhihaoshan-google: afaik, the `num_slices ` get overwritten as the number of the slices at some point in the code, even thought we pass `1` from the config file.

gobbleturk: Yes num_slices is meant to be automatically determined by jax.devices except in the AOT case

config comment: https://github.com/AI-Hypercomputer/maxtext/blob/4651cb3c73de92e528a98281be29c05b27aaf501/MaxText/configs/base.yml#L296-L299
pyconfig overwriting num_slices: https://github.com/AI-Hypercomputer/maxtext/blob/4651cb3c73de92e528a98281be29c05b27aaf501/MaxText/pyconfig.py#L428
jax.devices logic in max_utils setting num_slices: https://github.com/AI-Hypercomputer/maxtext/blob/4651cb3c73de92e528a98281be29c05b27aaf501/MaxText/max_utils.py#L359

fenghuizhang:
Tested manually in cluster to have verified the change. Can someone approve the PR? Thanks.

vipannalla (approved):

PR #1103, opened 2024-12-18
Use jax.lax.dot_general when both lhs and rhs are not quantized

lenscloth:
# Description

Aqt's dot_general did not set `preferred_element_type` properly when both lhs and rhs are not quantized.
A quick workaround is to use `jax.lax.dot_general` when quantization is not applied.

# Tests

Run end-to-end training test.

# Checklist


RissyRan (approved):
Thanks!

gobbleturk (approved):

PR #1110, opened 2024-12-20
Print memory stats after start_step in train

hengtaoguo:
# Description

Display the memory statistics (including percentages) for all utilized devices. In `max_utils.py`, replace `print` statements with `max_logging` to maintain consistency.

FIXES: b/385169408

# Tests

Ran tests on local v4-8 devbox and xpk workload on v5e-512 cluster.
image
memstats

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -935,6 +935,9 @@ def train_loop(config, state=None):
         jax.block_until_ready(state)  # Block until current state finishes to end profile cleanly
       prof.deactivate()
 
+    if step == start_step:
+      max_utils.print_mem_stats("After params initialized")
+
   if checkpoint_manager is not None:
     checkpoint_manager.wait_until_finished()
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -935,6 +935,9 @@ def train_loop(config, state=None):
         jax.block_until_ready(state)  # Block until current state finishes to end profile cleanly
       prof.deactivate()
 
+    if step == start_step:
+      max_utils.print_mem_stats(label=config.run_name)
```
gobbleturk, line 939: Can we change label here to "After params initalized" as well like https://github.com/AI-Hypercomputer/maxtext/blob/6ec3368af31fff6e6d735ac9d5fb77f91fc0f784/MaxText/maxengine.py#L156?

For reference this is only showing the HBM usage of the optimizer state. This can help us verify that host offloading is working as expected

hengtaoguo: Thanks for the suggestion and done!

gobbleturk (approved):
Thanks hengtao!

PR #1118, opened 2024-12-21
Recommend using v5e instead of v4

tornikeo:
# Description

I'm following the quickstart tutorial and the `v4-8` accelerator seems to be unavailable in most regions. I switched to `v5litepod-1` after a while. Now it works.

# Tests

I managed to complete the tutorial using v5litepod-1, 4 and 8, while v4 doesn't seem to work

# Checklist

### DIFF
--- a/getting_started/First_run.md
+++ b/getting_started/First_run.md
@@ -13,7 +13,7 @@ We recommend starting with a single host first and then moving to multihost.
 Local development is a convenient way to run MaxText on a single host. It doesn't scale to
 multiple hosts.
 
-1. [Create and SSH to the single-host VM of your choice.](https://cloud.google.com/tpu/docs/users-guide-tpu-vm#creating_a_cloud_tpu_vm_with_gcloud) We recommend a `v4-8`.
+1. [Create and SSH to the single-host VM of your choice.](https://cloud.google.com/tpu/docs/users-guide-tpu-vm#creating_a_cloud_tpu_vm_with_gcloud) We recommend a `v5litepod-8`.
 2. Clone MaxText onto that TPUVM.
 3. Within the root directory of that `git` repo, install dependencies and pre-commit hook by running:
 ```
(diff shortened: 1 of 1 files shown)


getting_started/First_run.md
```diff
@@ -13,7 +13,7 @@ We recommend starting with a single host first and then moving to multihost.
 Local development is a convenient way to run MaxText on a single host. It doesn't scale to
 multiple hosts.
 
-1. [Create and SSH to the single-host VM of your choice.](https://cloud.google.com/tpu/docs/users-guide-tpu-vm#creating_a_cloud_tpu_vm_with_gcloud) We recommend a `v4-8`.
+1. [Create and SSH to the single-host VM of your choice.](https://cloud.google.com/tpu/docs/users-guide-tpu-vm#creating_a_cloud_tpu_vm_with_gcloud) We recommend a `v5litepod-8`.
```
gobbleturk, line 16: We have done the most extensive testing on v4-8 and our unit tests run on v4-8, so I prefer to keep v4-8 as recommended. However everything should work on a single host v5litepod-8 as well - notably the default config in base.yml uses a small enough batch size on v5litepod so step 4 below runs fine.

How about instead "You can use whatever single host capacity is available to you, such as a v4-8, v5litepod-8 or v5p-8. Our unit tests run against v4-8 but any single host VM should work."

gobbleturk (commented):
Thank you for this contribution! I expect available capacity will shift over time and vary based on user /  reservation / zone etc, so perhaps we can drop the emphasis on v4-8 and instead list all of the options

getting_started/First_run.md
tornikeo: Yeah, that's much better I think. The root issue was the (surprising) unavailability of v4-8's, so I'll rephrase that.

gobbleturk (approved):

PR #1119, opened 2024-12-21
Add smaler v5e topologies to train_compile

gobbleturk:
# Description

* Add smaller v5e-1 through v5e-8 topologies to train_compile
* Add nicer error message when user enters an invalid compile_topology string option

# Tests

Ran train_compile with compile_topology=v5e-8 (success) and v5-7 (expected error message)

# Checklist

### DIFF
--- a/MaxText/accelerator_to_spec_map.py
+++ b/MaxText/accelerator_to_spec_map.py
@@ -45,6 +45,9 @@ UserFacingNameToSystemCharacteristics = {
     "v6e-128": SystemCharacteristics("tpu", "v6e:8x16", "default", (2, 2, 1), 128, (False, True, False)),
     "v6e-256": SystemCharacteristics("tpu", "v6e:16x16", "default", (2, 2, 1), 256, (True, True, False)),
     # v5e: one core per chip with 16 GB HBM
+    "v5e-1": SystemCharacteristics("tpu", "v5e:1x1", "default", (1, 1, 1), 1, (False, False, False)),
+    "v5e-4": SystemCharacteristics("tpu", "v5e:2x2", "default", (2, 2, 1), 4, (False, False, False)),
+    "v5e-8": SystemCharacteristics("tpu", "v5e:2x4", "default", (2, 2, 1), 8, (False, False, False)),
     "v5e-16": SystemCharacteristics("tpu", "v5e:4x4", "default", (2, 2, 1), 16, (False, False, False)),
     "v5e-32": SystemCharacteristics("tpu", "v5e:4x8", "default", (2, 2, 1), 32, (False, False, False)),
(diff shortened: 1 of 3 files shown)


MaxText/accelerator_to_spec_map.py
```diff
@@ -45,6 +45,9 @@ class SystemCharacteristics:
     "v6e-128": SystemCharacteristics("tpu", "v6e:8x16", "default", (2, 2, 1), 128, (False, True, False)),
     "v6e-256": SystemCharacteristics("tpu", "v6e:16x16", "default", (2, 2, 1), 256, (True, True, False)),
     # v5e: one core per chip with 16 GB HBM
+    "v5e-1": SystemCharacteristics("tpu", "v5e:1x1", "default", (2, 2, 1), 1, (False, False, False)),
```
shralex, line 48: v5e-1 has one chip so does it use the same (2,2,1) per host/vm chips_per_host_bounds ? Please see v6e-1 above.

gobbleturk: Great catch!

MaxText/pyconfig.py
```diff
@@ -717,8 +717,8 @@ def calculate_global_batch_sizes(
 
 
 def get_num_target_devices(raw_keys):
-  compile_topology = accelerator_to_spec_map.get_system_characteristics(raw_keys.get("compile_topology", ""))
-  if compile_topology is not None:
+  if raw_keys["compile_topology"]:
```
shralex, line 720: if its not in the map you'll get a KeyError exception rather than False. 
if "compile_topology" in raw_keys:

shralex: Previously, len(jax.devices()) would be returned here in two cases -- "compile_topology" is not in raw_keys (in which case "" will be passed to get_system_characteristics) and compile_topology is in raw_keys but the value for it is not in our map of supported things.

Do you want to print the new error in both cases or just in one of them ?

gobbleturk: Whenever compile_topology is set (e.g. is not ""), we should error when its not a valid topology. Also we should only call get_system_characteristics when compile_toplogy is set

With AOT we set compile topology - we will get an error only when the compile_topology is not a valid a topology

Without AOT we don't set compile topology - we should return jax.devices()

shralex: Thanks for explaining! In that case, what do you think about adding a comment above the if about AOT and doing either "if "compile_topology" in raw_keys:" or ".get" instead of [].

gobbleturk: Added a comment! Changed to using .get instead of []. Note that "compile_topology" in raw_keys would not work - every key such as compile_topology is part of raw_keys since they are defined in base.yml, but here we are really checking that it is not the empty string (which it is set to by default)

shralex (approved):

PR #1125, opened 2024-12-23
Add option to skip initializing the jax distributed system

gobbleturk:
# Description

Add option to skip jax.distributed.initialize() explicitly.
This is necessary for internal runs with checkpointing, where the jax distributed system is initialized for us and should not be initialized explicitly

# Tests

Ran with this new flag both true and false, saw expected results.

# Checklist

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -254,6 +254,9 @@ def maybe_initialize_jax_distributed_system(raw_keys):
 
   For CPUs, we call jax.distributed.initialize() explicitly, with the specified arguments.
   """
+  if raw_keys["skip_jax_distributed_system"]:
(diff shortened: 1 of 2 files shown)


MaxText/max_utils.py
```diff
@@ -219,6 +219,9 @@ def maybe_initialize_jax_distributed_system(raw_keys):
 
   For CPUs, we call jax.distributed.initialize() explicitly, with the specified arguments.
   """
+  if raw_keys["skip_jax_distributed_system"]:
```
RissyRan, line 222: Nit: Thanks for the log! So we know the reason to skip the initialization. Could you help add accordingly logs to following 2 cases? `inference_benchmark_test` & `compile_topology`?

RissyRan (approved):
Others LGTM!

PR #1133, opened 2025-01-01
Add Context Parallelism support to cudnn Flash Attention

kocchop:
# Description
This PR adds Context Parallelism support to GPU Flash Attention. It is necessary to support large sequence lengths in MaxText. Right now, the support is offered through Transformer-Engine and uses an All-Gather type implementation. Note that, it requires mask type to be `causal` and does not work with sliding window attention yet. Also it requires `transformer-engine==1.13` or above.

**NEW**
1. Implemented input sequence re-ordering at the data loading stage
2. Added `context_parallel_load_balance` flag in `base.yml` to turn on/off load balancing.
3. Added sequence packing flag `enable_packing` in `base.yml` and also modified the associated data processing files.
4. Added/modified unit tests for context parallelism and GPU flash attention `cudnn_flash_te`
5. Added sharding of parameters across CP rank

# Tests
Unit test is included with the PR with the base model for `4 x a100` gpus,

# Checklist


gobbleturk (changes requested):
Awesome, this looks great! Just one change to the sharding rules (with current state of this PR it looks like the old sequence parallelism is broken)

I'd appreciate it if you could also run `bash code_style.sh` to run our linter (I recommend saving the branch before running this just in case...)

MaxText/configs/base.yml
```diff
@@ -227,13 +227,14 @@ jax_cache_dir: "~/jax_cache"
...
 # Parallelism
-mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'tensor', 'expert', 'autoregressive']
+mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'tensor', 'expert', 'autoregressive']
 logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_batch_no_exp', ['data', 'fsdp', 'fsdp_transpose']],
                       ['activation_embed_and_logits_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_heads', ['tensor','sequence']],
                       ['activation_kv_heads', ['tensor','sequence']],
+                      ['activation_length', 'context'],
```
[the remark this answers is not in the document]
abhinavgoel95: Faysal is OOO. Ill make this change for him and re-request a review.

MaxText/layers/attentions.py
```diff
@@ -425,6 +432,8 @@ def cudnn_flash_attention(
         scale_factor=1.0 / math.sqrt(head_dim),
         transpose_batch_sequence=False,
         window_size=sliding_window_size,
+        context_parallel_causal_load_balanced=True,
```
mgoldfarb-nvidia, line 435: There is some additional detail to handle in the framework here for correctness. When we apply causal load balancing the attention expects that tokens are reordered. This is not done in the PR so the output results will not be correct.

Details of this reordering are the same as described in the paragraph: _Context parallelism for long sequences_ in the Llama 3 paper. Essentially the sequences is split into `2*CP` groups and reordered to gather chunks `i` and `(2 × CP − 1 − i)`

We have a helper function in TE that performs this re-ordering or it could be easily added separately here. https://github.com/NVIDIA/TransformerEngine/blob/c9ea6be92948e1ec553037f1a04900617b9f7f6b/transformer_engine/jax/cpp_extensions/attention.py#L1005

We are likely going to employ a different reordering approaches  if the user wants to employ sequence packing due the limited masking flexibility that exists in cudnn today. This PR does not intend to add support for sequence packed CP attention yet. We have several approaches but each has some nuance of how it works and may be worth discussing how we can pipe through the appropriate configurations to ensure consistency with cuDNN.

MaxText/configs/base.yml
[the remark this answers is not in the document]
kocchop: Hi @gobbleturk, for some reasons `['activation_length', ['context', 'sequence']]` were leading to some unexpected behavior like double sharding of the sequences. This was not something I could fix from the framework level, so left as 2 separates

MaxText/layers/attentions.py
kocchop: Implemented the required the re-ordering of the sequence dimension based on `cp_size`

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
```diff
@@ -259,7 +260,16 @@ def preprocess_train_dataset(
...
   train_ds = train_ds.shuffle(shuffle_buffer_size, seed=data_shuffle_seed)
-  train_ds = sequence_packing.pack_dataset(train_ds, max_target_length)
+
+  if enable_packing:
+    train_ds = sequence_packing.pack_dataset(train_ds, max_target_length)
+  else:
+    train_ds = train_ds.padded_batch(
+        train_global_batch_size_to_load // jax.process_count(),
+        padded_shapes={"targets": max_target_length},
+        padding_values={"targets": 0},
```
A9isha, line 270: We should use the tokenizer's `pad_id` instead of hard-coding to 0

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
```diff
@@ -271,10 +281,19 @@ def preprocess_eval_dataset(
...
 ) -> tf.data.Dataset:
   """Preprocess the evaluation dataset."""
-  eval_ds = sequence_packing.pack_dataset(eval_ds, max_target_length)
+  if enable_packing:
+    eval_ds = sequence_packing.pack_dataset(eval_ds, max_target_length)
+  else:
+    eval_ds = eval_ds.padded_batch(
+        eval_global_batch_size_to_load // jax.process_count(),
+        padded_shapes={"targets": max_target_length},
+        padding_values={"targets": 0},
```
A9isha, line 294: Same here

MaxText/configs/base.yml
```diff
@@ -592,3 +593,9 @@ sa_use_fused_bwd_kernel: False
 sa_q_layout: "HEAD_DIM_MINOR"
 sa_k_layout: "HEAD_DIM_MINOR"
 sa_v_layout: "HEAD_DIM_MINOR"
+
+### Determine if we want to use load balance for context parallelism
+context_parallel_load_balance: True
+
+# Whether to enable sequence packing during data processing
+enable_packing: True
```
A9isha, line 601: nit: let's have the newline at the end

MaxText/layers/attentions.py
```diff
@@ -477,8 +477,8 @@ def cudnn_flash_attention(
         scale_factor=1.0 / math.sqrt(head_dim),
         transpose_batch_sequence=False,
         window_size=sliding_window_size,
-        context_parallel_causal_load_balanced=True,
-        context_parallel_axis='context',
+        context_parallel_causal_load_balanced=self.config.context_parallel_load_balance,
```
A9isha, line 480: Just for my understanding, why do we use `DotProductAttention` inside `cudnn_flash_attention `?

A9isha (commented):
This is great! thank you so much

Curious, what kind of improvement are you observing with the loadbalancing enabled?

A9isha:
Oh also please merge all commits into one before pushing

kocchop:
> This is great! thank you so much
> 
> Curious, what kind of improvement are you observing with the loadbalancing enabled?

for llama3-8b, with load balancing and cp=2, we're able to see around ~10% perf improvement

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
kocchop: I have reverted the changes to the `c4_mlperf` dataloader and added an assertion in `input_pipeline_interface.py`. For the padded c4, I have verified that the tfds one works fine. For now leaving the c4_mlperf as is and will be used only with packing when `dataset_type=c4_mlperf`. Otherwise, it will use tfds path which would work for both packed and un-packed formats

MaxText/layers/attentions.py
kocchop: It's just the name of the API in TE

MaxText/max_utils.py
```diff
@@ -1132,3 +1133,50 @@ def unpermute_from_match_maxtext_rope(arr, model_size):
   evens = arr[..., ::2]
   odds = arr[..., 1::2]
   return jax.numpy.concatenate((evens, odds), axis=arr.ndim - 1)
+
+
+@partial(jax.jit, static_argnums=1)
+def reorder_sequence(tensor, cp_size: int):
+  """Reorders the sequence of the tensor. For example, with cp_size=2,
```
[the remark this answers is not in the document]
kocchop: Added parameter sharding across CP rank

> cp_size = 4?
@gobbleturk, what do I need to add here?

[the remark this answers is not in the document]
gobbleturk: ah sorry I didn't understand the example, I thought you were using CP=4 but really this example is CP=2

MaxText/max_utils.py
```diff
@@ -1132,3 +1133,50 @@ def unpermute_from_match_maxtext_rope(arr, model_size):
...
+
+  # Create first and second halves
+  first_half = jnp.arange(cp_size)
+  second_half = jnp.arange(2 * cp_size - 1, cp_size - 1, -1)
+
+  # Stack and reshape to interleave
+  src_indices = jnp.stack([first_half, second_half], axis=1).reshape(-1)
+
+  # Indexing the reshaped tensor using JAX take operation
+  reordered = jnp.take(reshaped, src_indices, axis=1)
```
gobbleturk, line 1159: I'm curious if a permute API instead of `jnp.take` might be more efficient (might be different on GPU vs TPU)

MaxText/max_utils.py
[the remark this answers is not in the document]
kocchop: yes, this is CP=2

MaxText/max_utils.py
gobbleturk: (no need to test for this PR)

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
kocchop: @A9isha did you get a chance to take a look? Let me know what do you think

MaxText/max_utils.py
mgoldfarb-nvidia: Happy to try some proposals if you have a suggestion, we measured this method on hopper GPU and its quite fast as is. We don't expect to do this operation frequently relative to other parts of the model but even so it should be cheap.

MaxText/configs/base.yml
```diff
@@ -313,6 +314,7 @@ dcn_data_parallelism: -1  # recommended DCN axis to be auto-sharded
 dcn_fsdp_parallelism: 1
 dcn_fsdp_transpose_parallelism: 1
 dcn_sequence_parallelism: 1  # never recommended
+dcn_context_parallelism: 1  # never recommended
```
gobbleturk, line 317: perhaps we would use seq and context over DCN on gpus? E.g. TP_NVLINK=8 CP_DCN=16 PP_DCN=16 is how llama3 was sharded with 128k seq length

perhaps remove these comments

MaxText/max_utils.py
```diff
@@ -1132,3 +1133,50 @@ def unpermute_from_match_maxtext_rope(arr, model_size):
   evens = arr[..., ::2]
   odds = arr[..., 1::2]
   return jax.numpy.concatenate((evens, odds), axis=arr.ndim - 1)
+
+
+@partial(jax.jit, static_argnums=1)
```
gobbleturk, line 1138: curious is the jax.jit needed? Isn't reorder_sequence only called within a train step which is jitted?

MaxText/tests/train_tests.py
```diff
@@ -185,10 +186,31 @@ def test_gpu_cudnn_flash_te(self):
         "steps=2",
         "enable_checkpointing=False",
         "attention=cudnn_flash_te",
+        "enable_packing=False",
         r"tokenizer_path=../assets/tokenizer.llama2",
     ]
     train_main(cudnn_flash_te)
 
+  @pytest.mark.gpu_only
```
gobbleturk, line 194: Awesome, thank you for the test!

MaxText/max_utils.py
gobbleturk: oh I see this method is only used for the initial inputs.

On the TPU side I believe Anisha has implemented we need to do the "unpermuting" back to 0,1,2,3,4,5,6,7 before calling our kernel attention - it doesn't support the unpermuted [0, 1, 6, 7, 2, 3, 4, 5]. Does your kernel support it well - do you have to unpermute internally inside the DotProductAttention of TE or your kernel supports arbitrary kv permutes well (I assume at least k and v need to permuted the same)

MaxText/configs/base.yml
kocchop: yeah makes sense

MaxText/max_utils.py
kocchop: Yes, the unpermute happens inside the DPA

MaxText/max_utils.py
kocchop: no, the reorder function is mapped to data loader and getting invoked whenever, we call [this](https://github.com/AI-Hypercomputer/maxtext/blob/c26285971d066218fc549d7a11fac65b84354c7f/MaxText/train.py#L824). The jitted train step only gets called with already reordered batch.

Initially, I did not have `jax.jit` and the jitting didn't happen for this function.

MaxText/max_utils.py
kocchop: also I tested the `jax.take` indexing upto 64K seq length with CP=8 and found consistent data loading time. But would be happy to test if you have other suggestions

kocchop:
@gobbleturk @A9isha once I get a LGTM from you, I can work on resolving the merge conflicts and also merge all the commits into one.

gobbleturk (approved):
LGTM! I would like either @khatwanimohit or @aireenmei to review for the data loading change part `enable_packing=True`

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
A9isha: Thank you!

A9isha (approved):
LGTM - thank you!

MaxText/max_utils.py
```diff
@@ -1132,3 +1133,50 @@ def unpermute_from_match_maxtext_rope(arr, model_size):
...
+  reordered = jnp.take(reshaped, src_indices, axis=1)
+
+  # Reshape back to original dimensions
+  return reordered.reshape(batch_size, seq_len)
+
+
+@partial(jax.jit, static_argnums=1)
+def reorder_causal_load_balanced(batch, config):
+  """Reorders the example batch sequences"""
+  context_parallel_size = config.ici_context_parallelism * config.dcn_context_parallelism
```
A9isha, line 1168: oh! sorry one more thing - these values could be "-1" for e.g., `config.ici_context_parallelism=-1` is possible, for my PR I am using `cp_size = mesh.shape["context"]`

Ref: https://github.com/AI-Hypercomputer/maxtext/pull/1445/

MaxText/max_utils.py
```diff
@@ -1132,3 +1133,50 @@ def unpermute_from_match_maxtext_rope(arr, model_size):
   evens = arr[..., ::2]
   odds = arr[..., 1::2]
   return jax.numpy.concatenate((evens, odds), axis=arr.ndim - 1)
+
+
+@partial(jax.jit, static_argnums=1)
+def reorder_sequence(tensor, cp_size: int):
```
A9isha, line 1139: Could you please make this function generic so that 

1. we can pass the dimension which has sequence_length instead of assuming `seq_len=tensor.shape[1]`
2. we can permute or unpermute based on a flag

Similar to `reorder_causal_load_balancing` from https://github.com/AI-Hypercomputer/maxtext/pull/1445

That way we can reuse the same function

MaxText/max_utils.py
kocchop: I can do that. In that case, we don't need to pass the config to dataloader, we can directly pass the cp_size.

MaxText/max_utils.py
kocchop: I am not sure how useful this would be as this expects the tensor to be [B,S] not [B,S,H,D] and we don't need to unpermute as its already handled by the attention kernel.

MaxText/max_utils.py
kocchop: @A9isha done, PTAL

MaxText/configs/base.yml
```diff
@@ -590,3 +593,9 @@ sa_use_fused_bwd_kernel: False
 sa_q_layout: "HEAD_DIM_MINOR"
 sa_k_layout: "HEAD_DIM_MINOR"
 sa_v_layout: "HEAD_DIM_MINOR"
+
+### Determine if we want to use load balance for context parallelism
+context_parallel_load_balance: True
+
+# Whether to enable sequence packing during data processing
+enable_packing: True
```
aireenmei, line 601: We just added a packing flag https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L380C1-L380C8. Let's use that

aireenmei (changes requested):

MaxText/configs/base.yml
kocchop: I am waiting on #1453 to get merged and then I will rebase it. At that time, I can make the change.

kocchop: done!

kocchop:
Hi @gobbleturk could you please approve so that the unit tests can run?

MaxText/configs/base.yml
```diff
@@ -266,15 +266,16 @@ jax_cache_dir: "~/jax_cache"
...
-mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context_autoregressive', 'tensor', 'tensor_transpose', 'tensor_sequence', 'expert', 'autoregressive']
+mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'context_autoregressive', 'tensor', 'tensor_transpose', 'tensor_sequence', 'expert', 'autoregressive']
 logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_batch_no_exp', ['data', 'fsdp', 'fsdp_transpose']],
                       ['activation_embed_and_logits_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_heads', ['tensor', 'tensor_transpose', 'sequence','tensor_sequence','autoregressive']],
                       ['activation_kv_heads', ['tensor', 'tensor_transpose', 'sequence','tensor_sequence']],
+                      ['activation_length', ['context']],
                       ['activation_length', ['sequence']],
```
gobbleturk, line 277: I think we want ['activation_length' ,['sequence, 'context']] and below it only ['activation_length', ['context']

kocchop: I see let me do that

kocchop: Hi @gobbleturk done! PTAL.

gobbleturk: ya sorry about this, I have a bug to make these sharding constraints easier, at some point we will have a big refactor to specify every tensor individually instead of these sharded logical rules

gobbleturk: This looks right I'll re-run tests

gobbleturk (approved):

MaxText/configs/base.yml
A9isha: Wonderful! I have similar change in the context parallelism for TPU PR :)

A9isha (approved):
Thank you!

aireenmei (approved):
LGTM for data input changes!

gobbleturk:
All tests passed, manually adding pull ready

gobbleturk:
pull ready might have added successfully automatically after an additional allowing step by me to run next steps, not sure

PR #1140, opened 2025-01-03
Router Sharding optimization

ZhiyuLi-goog:
# Description
Following the guide from the core ml performance team, add router sharding optimization for inferencing 11.20% gain in prefill (1024 length) step or 9.8% improvement throughputs.

Key intuition: change from pre matmul all-gather to a post matmul one, much smaller in size. The costs in HBM is almost negectable after replicating the small router_projection weight. 
* in router projection: router_logits = self.router_logits(router_input)
* without this change: all gather **router_input** with the communication int8[batch, seq_len, embed] or int8[1, 1024, 768]
* with this change: all gather **router_logits** instead, a much smaller tensor with the communication bf16[batch, seq_len, num_expert] or bf16[1, 1024, 8]
* Replicate router_projection weight [embed, num_expert] is trivial since the weight is usually smaller than the other mlp weight [embed, mlp] by an order of magnitude.

Performance in training:
It should improve MoE training if we use tensor parallelism as well. However, usually we use either FSDP or expert parallelism. The performance is the same, verified in real experiments.

# Tests
Tested in both inferencing and training. 

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -55,13 +55,6 @@ DISPATCH = "dispatch"
 COMBINE = "combine"
 
 
-def _get_model_call_mode(config):
-  if config.model_cal_mode == "inference":
-    return "inference"
-  else:
-    return None
-
-
 def _convert_to_activation_function(fn_or_string: Union[str, Callable[..., Any]]) -> Callable[..., Any]:
   """Convert a string to an activation function."""
   if fn_or_string == "linear":
@@ -107,7 +100,7 @@ class DenseGeneral(nn.Module):
   weight_dtype: DType = jnp.float32
   dtype: DType = jnp.float32
   kernel_init: NdInitializer = nd_dense_init(1.0, "fan_in", "truncated_normal")
-  kernel_axes: Tuple[str, ...] = ()
+  kernel_axes: Tuple[Optional[str], ...] = ()
   quant: Optional[Quant] = None
   use_bias: bool = False
   matmul_precision: str = "default"
@@ -309,7 +302,7 @@ class MoeBlock(nn.Module):
   num_experts_per_tok: int
   mesh: Mesh
   kernel_init: NdInitializer
-  kernel_axes: Tuple[str, ...]
+  kernel_axes: Tuple[Optional[str], ...]
   weight_dtype: DType = jnp.float32
   dtype: DType = jnp.float32
   quant: Optional[Quant] = None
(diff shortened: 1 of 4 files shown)


ZhiyuLi-goog:
@RissyRan @mailvijayasingh Could you review this change at your convenience?

MaxText/layers/linears.py
```diff
@@ -609,7 +603,10 @@ def dense_matmul(self, inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel):
       mask_axes = ("activation_batch", "activation_length", None, None)
       dispatch_mask = nn.with_logical_constraint(dispatch_mask, mask_axes)
       combine_mask = nn.with_logical_constraint(combine_mask, mask_axes)
-      loss = self.load_balance_loss(top_k_indices, softmax_probs)
+      if self.config.model_call_mode != "inference":
+        loss = self.load_balance_loss(top_k_indices, softmax_probs)
+      else:
+        loss = 0
```
RissyRan, line 609: Will there be an issue if we return as None? to be consistent with other [settings](https://github.com/AI-Hypercomputer/maxtext/blob/6eb375f4a1cdd206a3c2669103be52c3c54d1edb/MaxText/layers/linears.py#L705).

MaxText/layers/linears.py
```diff
@@ -730,7 +727,7 @@ def __call__(self, inputs):
         weight_dtype=self.weight_dtype,
         quant=self.quant,
         kernel_init=self.kernel_init,
-        kernel_axes=self.kernel_axes,
+        kernel_axes=("embed", None),
```
RissyRan, line 730: Shall we update it [here](https://github.com/AI-Hypercomputer/maxtext/blob/6eb375f4a1cdd206a3c2669103be52c3c54d1edb/MaxText/layers/mistral.py#L134)?

RissyRan (commented):
Thanks Zhiyu!

MaxText/layers/linears.py
ZhiyuLi-goog: Good catch!

MaxText/layers/linears.py
ZhiyuLi-goog: Good idea.

ZhiyuLi-goog: Done.

MaxText/layers/linears.py
ZhiyuLi-goog: Done

RissyRan (approved):

PR #1142, opened 2025-01-06
Add option to profile periodically

gobbleturk:
# Description

Add an option to collect a profile periodically, every N steps. This can help debug scenarios where performance is variable over time

This also cleans up the existing profiler_test file, which actually had dead code that was not being tested (the code was not functioning and did not have the @pytest labels)

FIXES: b/387539222

# Tests
Included some unit tests, and confirmed that running only a regular profile and also period profile works as expected on my single host dev machine.

# Checklist

### DIFF
--- a/MaxText/profiler.py
+++ b/MaxText/profiler.py
@@ -28,18 +28,27 @@ import jax
 class Profiler:
   """Activate/deactivate a profiler based on the 'profiler' config"""
 
-  def __init__(self, config, optional_postfix=""):
+  def __init__(self, config, offset_step=0):
     self.libcudart = None
     self.mode = config.profiler
-    self.upload_all_profiler_results = config.upload_all_profiler_results
     if self.mode != "":
-      self.output_path = os.path.join(config.tensorboard_dir, optional_postfix)
+      self.base_output_dir = config.tensorboard_dir
+    self.output_path = ""
+    self.upload_all_profiler_results = config.upload_all_profiler_results
+    self.profile_cleanly = config.profile_cleanly
+    self.profile_period = config.profile_periodically_period
+    self.start_initial_profile_step = self._set_first_profiler_step(config.skip_first_n_steps_for_profiler, offset_step)
+    self.finished_initial_profile_step = self._set_last_profiler_step(config.profiler_steps, config.steps)
 
-  def activate(self):
+  def activate(self, blocking_object=None, optional_postfix=""):
     """Start the profiler.
     nsys profiler becomes no-op when libcudart.so is not available on the system"""
+    if self.profile_cleanly and blocking_object is not None:
+      jax.block_until_ready(blocking_object)
     if not (self.upload_all_profiler_results or jax.process_index() == 0):
       return
+    if self.mode != "":
+      self.output_path = os.path.join(self.base_output_dir, optional_postfix)
     if self.mode == "nsys":
       try:
         self.libcudart = cdll.LoadLibrary("libcudart.so")
@@ -50,9 +59,11 @@ class Profiler:
     elif self.mode == "xplane":
       jax.profiler.start_trace(self.output_path)
 
-  def deactivate(self):
+  def deactivate(self, blocking_object=None):
     """End the profiler.
     The result is uploaded to the output bucket"""
+    if self.profile_cleanly and blocking_object is not None:
+      jax.block_until_ready(blocking_object)
     if not (self.upload_all_profiler_results or jax.process_index() == 0):
       return
     if self.mode == "nsys":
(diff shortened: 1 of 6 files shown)


MaxText/profiler.py
```diff
@@ -68,3 +81,24 @@ def deactivate(self):
...
+
+  def get_first_profiler_step(self, start_step=0):
+    return start_step + self.skip_first_n_steps
+
+  def get_last_profiler_step(self, start_step=0):
+    first_profiling_step = self.get_first_profiler_step(start_step=start_step)
+    return np.clip(first_profiling_step + self.profiler_steps - 1, first_profiling_step, self.total_job_steps - 1)
+
+  def should_activate_periodic_profile(self, step, start_step=0):
+    finished_reg_profile_step = self.get_last_profiler_step(start_step=start_step)
```
Obliviour, line 93: Ahh it took me a bit to understand but I believe this will returns the first profile's instance last step. It won't update for the following period profiles. I think this is intended to make sure that you don't try to profile periodically at the same time as the initial profile after skip_n_steps.

1. Can we rename finished_reg_profile_step to finished_first_profile_step or finished_initial_profile_step.
2. We can avoid recalling get_last_profiler_step() every training step and do it just once on the __init__ of the profiler.

MaxText/profiler.py
```diff
@@ -68,3 +81,24 @@ def deactivate(self):
...
+    return self.profile_period > 0 and step > finished_reg_profile_step and step % self.profile_period == 0
+
+  def should_deactivate_periodic_profile(self, step, start_step=0):
+    finished_reg_profile_step = self.get_last_profiler_step(start_step=start_step)
+    if self.profile_period > 0 and step == self.total_job_steps - 1:
```
Obliviour, line 98: nit: self.profile_period > 0 seems like an assert we should do on __init__ instead of during runtime? Its a static arg and a validation of profiler settings?

MaxText/train.py
```diff
@@ -940,10 +942,8 @@ def train_loop(config, state=None):
         prof.deactivate()
         break
 
-    if step == last_profiling_step:
-      if config.profile_cleanly:
-        jax.block_until_ready(state)  # Block until current state finishes to end profile cleanly
-      prof.deactivate()
+    if step == last_profiling_step or prof.should_deactivate_periodic_profile(step, start_step=start_step):
```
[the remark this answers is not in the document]
Obliviour: Same for the first_profiling_step

MaxText/profiler.py
gobbleturk: Done.

MaxText/profiler.py
gobbleturk: self.profile_period = -1 is the default to indicate that no periodic profile should be collected

MaxText/train.py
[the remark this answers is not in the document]
gobbleturk: We are checking if we should end either the initial profile or a periodic profile. The periodic profile check will return false on last_profiling_step since this PR writes code to avoid overlapping the initial profile with the periodic one

gobbleturk:
The resulting state for this PR is support for:

1. No profiling
2. Only an "initial profile", e.g. from steps 1-6
3. An initial profile and a periodic profile, e.g. 1-6 + 100-104 + 200-204 + ...

There isn't support for case 4: only the periodic profile (no "initial profile"), since I don't see a good motivation for it and it adds additional complexities. I believe case 3. is what we want - a profile to look at right away soon after the job starts (the initial profile) and also some profiles over time if we suspect the profile is changing over time.

Alternatively it may be simpler to remove support for 3. and support the only periodic profile (case 4) , then we don't have to check for collisions of the initial and the periodic

MaxText/profiler.py
Obliviour: Ah I see, nice nice, thanks Matt

MaxText/train.py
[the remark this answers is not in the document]
Obliviour: Gotcha makes sense, thanks Matt!

Obliviour (approved):

gobbleturk:
> The resulting state for this PR is support for:
> 
> 1. No profiling
> 2. Only an "initial profile", e.g. from steps 1-6
> 3. An initial profile and a periodic profile, e.g. 1-6 + 100-104 + 200-204 + ...
> 
> There isn't support for case 4: only the periodic profile (no "initial profile"), since I don't see a good motivation for it and it adds additional complexities. I believe case 3. is what we want - a profile to look at right away soon after the job starts (the initial profile) and also some profiles over time if we suspect the profile is changing over time.
> 
> Alternatively it may be simpler to remove support for 3. and support the only periodic profile (case 4) , then we don't have to check for collisions of the initial and the periodic

I've changed the PR so we only support 4, not 3, I think its a lot simpler. The periodic profile now uses the `skip_n_steps_for_profile` which gives users the control of when to start the initial one.

Now PR supports
1. No Profiling
2. only one profile starting at skip_n_steps_for_profile
3. periodic profile, starting at skip_n_steps_for_profile

PR #1167, opened 2025-01-14
add GCP workload observability feature

jcyang43:
# Description

Add option to enable GCP workload monitoring for MaxText workloads. 

- GCP workload monitoring sends performance metrics (heartbeat & training step times) to cloud monarch for monitoring such that if a metric hits its pre-defined threshold, oncalls will be notified to see if any actions are needed. This is ideal for critical workloads sensitive to infrastructure changes. 
- Each metric can be configured to be on or off based on configs. Examples are included in `MaxText/configs/base.yml`
- Documentation can be found at `getting_started/GCP_Workload_Monitoring.md`

# Tests

Tested on trillium TPU and confirmed metrics sent to cloud monarch successfully if configs are enabled. No metrics will be sent to cloud monarch if configs are set to False.

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -25,6 +25,7 @@ import os
 import sys
 import functools
 import time
+import queue
 
 from typing import Sequence, Optional
 from absl import app
@@ -51,6 +52,8 @@ from vertex_tensorboard import VertexTensorboardManager
 
 from input_pipeline.input_pipeline_interface import create_data_iterator
 from layers import models
+
+from gcp_workload_monitor import GCPWorkloadMonitor
 
 import jax.numpy as jnp
 from jax import random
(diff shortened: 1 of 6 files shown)


MaxText/train.py
```diff
@@ -875,11 +887,11 @@ def train_loop(config, state=None):
       with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
         state, metrics = p_train_step(state, example_batch, nextrng)
 
-    new_time = datetime.datetime.now()
-    record_scalar_metrics(
-        metrics, new_time - last_step_completion, per_device_tflops, learning_rate_schedule(step), per_device_tokens
-    )
-    last_step_completion = new_time
+    step_time_delta = datetime.datetime.now() - last_step_completion
```
gobbleturk, line 890: Changes like this need to be handled with care - jax is by default lazy e.g. functions return instantly, they do not block on finishing computation. In this case the results before and after this PR are roughly the same as confirmed by this diff https://diff.googleplex.com/#key=TLvLsJisjPpi (LHS main, RHS this PR)

The real blocking function is when an array is either checkpointed or printed - in this case this is done by write_metrics which is below (not record_scalar_metrics or p_train_step). This is fine since write_metrics is done after last_step_completion, so the step_time_delta will have to wait a real train_step worth of time

gobbleturk (approved):

PR #1170, opened 2025-01-16
Use an async checkpointer to initialize JDI in tests

gobbleturk:
# Description

According to this [logic](https://github.com/AI-Hypercomputer/maxtext/blob/d5e3deeaad8fb7162c8ae254f878a73ba718dba7/MaxText/max_utils.py#L276) we only initialize the jax distributed system for an async checkpointer, but it looks like sometime recently in orbax this is needed even for a sync checkpointer. We should modify this logic but it may break some single host or pathways use cases which don't need the distributed system.  A short term fix is to use an async checkpointer so that the jax distributed system is initialized, which is done for the end_to_end tests in this PR fixing b/389730949

# Tests

Ran gemma step 2 locally

# Checklist


khatwanimohit (approved):
Thank you Matt!

PR #1174, opened 2025-01-17
Remote Python Support for Benchmark Runner

SujeethJinesh:
# Description

Adds a recipe for testing and eventually benchmarking the Pathways Remote Python feature.

This is an initial PR that will set the precedent for future recipes. It is designed to be shareable and have common functionality (like flag parsing) that can be shared across recipes, but flexible and standard so that multiple team members can simply run the file with minimal configuration changes. It's designed to avoid superfluous flag adding and continue pythonically using configs.

This recipe in particular is for testing functionality for remote python.

# Tests

Tested this change on a test remote python sidecar server on a v6e and a v5e shared test cluster.

Sibling XPK change: https://github.com/AI-Hypercomputer/xpk/pull/326

# Checklist

### DIFF
--- a/benchmarks/maxtext_trillium_model_configs.py
+++ b/benchmarks/maxtext_trillium_model_configs.py
@@ -60,6 +60,38 @@ default_basic_1 = _add_to_model_dictionary(
 )
 
 
+default_basic_1_pw = _add_to_model_dictionary(
+  trillium_model_dict,
+  MaxTextModel(
+    model_name="default-basic-1-pw",
+    model_type="default",
+    tuning_params={
+        "per_device_batch_size": 1,
+        "remat_policy": "full",
+        "global_parameter_scale": 1,
+        "attention": "flash",
+        "dataset_path": "gs://max-datasets-rogue",
+        "dataset_type": "synthetic",
+        "reuse_example_batch": 1,
+        "enable_checkpointing": False,
+        # "profiler": "xplane",
+
+        # Additional tuning params for pathways long running test.
+        "enable_checkpointing": True,
+        "async_checkpointing": True,
+        "checkpoint_period": 100,
+        "checkpoint_storage_use_ocdbt": False,
+        "checkpoint_storage_use_zarr3": False,
+        "metrics_file": "metrics.txt",
+        "goodput_upload_interval_seconds": 30,
+        # "enable_pathways_goodput": True,
+        "enable_checkpoint_cloud_logger": True,
+        "enable_single_controller": True,
+    },
+    xla_flags="",
+  )
+)
+
 default_32 = _add_to_model_dictionary(
   trillium_model_dict,
   MaxTextModel(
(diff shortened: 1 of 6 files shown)


benchmarks/recipes/pw_remote_python_recipe.py
```diff
@@ -0,0 +1,123 @@
...
+
+import maxtext_trillium_model_configs as model_configs
+import maxtext_xpk_runner as mxr
+
+
+def main() -> int:
+  # V6e cluster config
+  cluster_config = mxr.XpkClusterConfig(
+      cluster_name="bodaborg-v6e-256-dnd-yucmhab",
+      project="tpu-prod-env-one-vm",
```
shauryagup, line 31: I don't know what's the policy here but let us remove these default values

benchmarks/recipes/pw_remote_python_recipe.py
```diff
@@ -0,0 +1,123 @@
+"""Copyright 2025 Google LLC
```
shauryagup, line 1: I am unclear on why we added this new recipe file. Why is updating the pathways yaml if sidecar is provided sufficient? This should be agnostic to the model config used to run with remote python.

benchmarks/recipes/pw_remote_python_recipe.py
SujeethJinesh: Accidentally left in some test code, thanks for the catch!

benchmarks/recipes/pw_remote_python_recipe.py
SujeethJinesh: I'm not sure I fully understand the question. There's an equivalent XPK change (https://github.com/AI-Hypercomputer/xpk/pull/326) needed for this to work end to end, and this benchmark runner uses that feature on XPK.

And yes the model config used is agnostic, since anyone can technically run remote python with the current setup. Since we're still developing, we'll likely need to make some edits to this recipe once RP is feature complete -- specifically to feature gate (e.g. using a USE_REMOTE_PYTHON option or such, similar to ENABLE_PERSISTENCE in pathwaysutils).

shauryagup: Thanks for sharing the xpk PR. So I understand that this is a temporary recipe till we gain enough confidence to merge this into the general pathways recipe. Let me know if that is not correct.

Looks good otherwise.

shauryagup (approved):

benchmarks/recipes/pw_remote_python_recipe.py
SujeethJinesh: Yup, this recipe will change once RP is complete.

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -60,6 +60,38 @@ def _add_to_model_dictionary(model_dictionary: dict[str, MaxTextModel], maxtext_
...
+        "remat_policy": "full",
+        "global_parameter_scale": 1,
+        "attention": "flash",
+        "dataset_path": "gs://max-datasets-rogue",
+        "dataset_type": "synthetic",
+        "reuse_example_batch": 1,
+        "enable_checkpointing": False,
+        # "profiler": "xplane",
+
+        # Additional tuning params for pathways long running test.
```
Obliviour, line 79: I am wondering if we can create a programmatic way of adding pathways params without needing to duplicate the config - for example default_basic_1 is sharable with mcjax and pathways and when the user has a pathways_config applied, it sets the pathways specific flags? I think the main goal to help solve is share a config between mcjax and pathways - and avoid the need to duplicate updates.

Curious what you think of this idea?

Obliviour (approved):

gobbleturk (approved):

benchmarks/maxtext_trillium_model_configs.py
SujeethJinesh: For right now there are only a few flags luckily that are pathways specific (e.g. enable_pathways_goodput), but we do often test using a lot more features enabled. A rough ideas I have is to separate the specific model tuning params from the workload params - e.g. like goodput, checkpointing, metrics, etc. That way the models can be the same, but the workload params can be suited to the specific type. Maybe worth expanding on if you think it's a good initial idea?

PR #1184, opened 2025-01-21
Maintain fp32 for optimizer state when offloading is enabled

zhenying-liu:
Address issues with optimizer state offloading and data type conversion

We identified two issues concerning the conversion from fp32 to fp16 for the optimizer state when enabling optimizer state offloading:

1. The comparison between configurations without and with optimizer state offloading was unfair because the data sizes differed, with the former using fp32 and the latter using fp16.
2. The presence of two modules with `jit_train_step` due to separate versions for fp32 and fp16 created inconsistencies.

This commit removes the fp32 to fp16 conversion, ensuring that the optimizer state retains its original data type.


khatwanimohit (approved):
Thank you!

gobbleturk (approved):

PR #1199, opened 2025-01-24
Add on-device runner and v5e benchmark model configs

raymondzouu:
# Description

Adding on-device runner to `benchmark_runner.py`. This will run the model benchmark on the current device rather than launching it as an xpk workload. The benchmark script must be ran on all workers. This will allow our XLML perf tests to run our latest benchmark model configs.

This change introduces a new command-line flag that specifies which runner to use (xpk, on-device). Usage is as follows:

xpk runner:
```
python3 benchmarks/benchmark_runner.py xpk --project=<project> --zone=<zone> --device_type=v5litepod-256 --num_slices=1  --cluster_name=<cluster> --base_output_directory=gs://<your-bucket> --model_name="default_16b_v5e_256" --base_docker_image maxtext_base_image
```

on-device runner:
```
python3 benchmarks/benchmark_runner.py on-device --base_output_directory=gs://<your-bucket> --model_name="default_16b_v5e_256" --run_name=test-run
```

# Tests

http://shortn/_grjxE2vPHD

# Checklist

### DIFF
--- a/benchmarks/benchmark_runner.py
+++ b/benchmarks/benchmark_runner.py
@@ -24,14 +24,16 @@ Example usages:
 import argparse
 
 from maxtext_trillium_model_configs import trillium_model_dict
+from maxtext_v5e_model_configs import v5e_model_dict
 from maxtext_xpk_runner import PathwaysConfig
 from maxtext_xpk_runner import WorkloadConfig
 from maxtext_xpk_runner import xpk_benchmark_runner
+from maxtext_xpk_runner import on_device_benchmark_runner
 from maxtext_xpk_runner import XpkClusterConfig
 from maxtext_xpk_runner import LibTpuType
 
-def add_shared_arguments(custom_parser: argparse.ArgumentParser):
-  """Add shared arguments to the parser.
+def add_xpk_runner_arguments(custom_parser: argparse.ArgumentParser):
+  """Add arguments to the xpk runner parser.
 
   Args:
     custom_parser: parser to add shared arguments to.
@@ -79,7 +81,7 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
   custom_parser.add_argument(
       '--model_name',
       type=str,
-      choices=list(trillium_model_dict.keys()),
+      choices=list(trillium_model_dict.keys()) + list(v5e_model_dict.keys()),
       default=list(trillium_model_dict.keys())[0],
       help=(
         f'model to be benchmarked, supported models are the command choices.'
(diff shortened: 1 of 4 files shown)


benchmarks/benchmark_runner.py
```diff
@@ -148,46 +157,120 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
...
+  add_xpk_runner_arguments(xpk_runner_parser)
+  add_on_device_runner_arguments(on_device_runner_parser)
   options = parser.parse_args()
 
-  cluster_config = XpkClusterConfig(
-        cluster_name=options.cluster_name,
-        project=options.project,
-        zone=options.zone,
-        device_type=options.device_type
+  assert trillium_model_dict.get(options.model_name) is not None or v5e_model_dict.get(options.model_name) is not None, f'Invalid model name: {options.model_name}'
```
Obliviour, line 218: I think we can remove this assert. It probably is excessive since the argument model_name is defined by these two dictionaries. Let me know if you see a good reason to keep the assert!

Obliviour (commented):
Really great improvements here! Adding v5e configs and a maxtext direct way of running these commands is awesome.

Can you add details on this way of directly running maxtext_xpk_runner.py in the readme file in the bnechmarks directory?

benchmarks/benchmark_runner.py
Obliviour: It might be good to alternatively check that the two / any dictionaries don't have duplicate entries.

benchmarks/benchmark_runner.py
```diff
@@ -148,46 +157,120 @@ def add_shared_arguments(custom_parser: argparse.ArgumentParser):
...
+  model = trillium_model_dict.get(options.model_name)
+  if model is None:
+    model = v5e_model_dict.get(options.model_name)
+
+  libtpu_type = options.libtpu_type
+  for t in LibTpuType:
```
Obliviour, line 224: I think you actually want a match case statement here o(1) vs o(n=num of libtpu types)

match x:
    case x == libtpu_type_1.value:
        print("x is 1")
    case 2:
        print("x is 2")
    case 3:
        print("x is 3")
    case _:
        error("unable to match case {x}".

benchmarks/benchmark_runner.py
raymondzouu: Done!

benchmarks/benchmark_runner.py
raymondzouu: Done!

Obliviour (approved):

gobbleturk (approved):

PR #1202, opened 2025-01-27
Prototype the MPMD GPipe-circular with single-host H100=8.

lc5211:
# Description
Single-host prototype for MPMD GPipe-circular for concept-proof purpose.
The prototype hardcoded the stages to 2, with each stage containing a set of 4 continous devices. The prototype needs to enable remat and now only takes care of one layer per stage.

# Checklist

- [ X] I have performed a self-review of my code.
- [ X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X ] I have made or will make corresponding changes to the doc if needed.

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -507,7 +507,6 @@ class _HyperParameters:
 def create_parallelisms_list(raw_keys):
   ici_parallelism = [
       raw_keys["ici_data_parallelism"],
-      raw_keys["ici_pipeline_parallelism"],
       raw_keys["ici_fsdp_parallelism"],
       raw_keys["ici_fsdp_transpose_parallelism"],
       raw_keys["ici_sequence_parallelism"],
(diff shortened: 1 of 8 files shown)


MaxText/pyconfig.py
```diff
@@ -507,7 +507,6 @@ def update_model_vars(base_config_path, raw_keys, config_name: str):
 def create_parallelisms_list(raw_keys):
   ici_parallelism = [
       raw_keys["ici_data_parallelism"],
-      raw_keys["ici_pipeline_parallelism"],
```
andig-google, line 510: With this deletions will the SPMD pipeline parallelism be removed? Can we find a way to support both?

MaxText/train.py
```diff
@@ -756,219 +757,59 @@ def setup_train_loop(config):
...
     config:
-    state:
-    ckpt_path:
+
   Returns:
+    state: the train state after training
   """
-  # Create a GoodputRecorder to log information
-  recorder = create_goodput_recorder(config)
-  record_goodput(recorder, config, recorder.record_job_start_time if recorder else None)
```
andig-google, line 769: Why do you need to remove the Goodput annotation?

PR #1211, opened 2025-01-28
[DRAFT] Initial commit Maxtext unit tests with Pathways backend.

RoshaniN:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist


docker_run_pathways_containers.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Parse input variables
+for ARGUMENT in "$@"; do
+    IFS='=' read -r KEY VALUE <<< "$ARGUMENT"
+    export "$KEY"="$VALUE"
+    echo "$KEY"="$VALUE"
+done
+
+cd utils_pathways
+MAXTEXT_IMAGE=${maxtext_image} COMMAND=${command} docker compose down
```
shauryagup, line 34: Can we do the cleanup everytime even during failures? And let us remove the first docker compose down on line 34 . So, something like:
```
{ # try
    docker compose up &&
} || { # catch
    docker compose down
}
```

gobbleturk (approved):

.github/workflows/run_pathways_tests_internal.yml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2024 Google LLC
```
gagika, line 1: Could you update the year to 2025 here and below (for new files)?

gagika (approved):

.github/workflows/run_pathways_tests_internal.yml
RoshaniN: Ohh yes, good catch! Thank you!

docker_run_pathways_containers.sh
shauryagup: My concern here was if the docker compose up command fails with an error that we are unable to catch, it won't execute the next line which does the cleanup. Thinking more, looks like the proposal I have would not solve for that either. The correct way would be
```
set -e
clean_up () {
    docker compose down
} 
trap clean_up EXIT
docker copose up
```
Ref: https://stackoverflow.com/questions/36335186/bash-exit-and-cleanup-on-error

RoshaniN: Thanks for the cleaner suggestion, implemented now!

PR #1217, opened 2025-01-30
Make synthetic data more realistic

philip-essential:
# Description

We've been using the `synthetic` dataset type for performance testing, and we noticed that the fact that it generates all zeros gives a false performance boost.  Something along the line must be faster when all the tokens are zero, though I'm not sure what that would be.  We observed a 3% discrepancy on v5p TPUs and a 10% discrepancy on MI300X GPUs.

This PR modifies the synthetic data iterator to produce tokens which are uniformly chosen among the vocabulary size.  While I doubt it will matter, I also set "positions" to more realistic values.

To ensure that randomly generating the data doesn't introduce a performance penalty, I pre-generate the data and just reference it later.

This seems to make the lazy computations be forced at different times, which makes the logged `step_time_seconds` incorrect.  It was printing 0.2 second step times every 1.8 seconds of wall clock time, which should be impossible.  I include a small fix to correctly keep track of the entire step time.

This is my first contribution to this repo, so let me know if there's any procedures I should follow.

# Tests

With only the step time fix, I ran gemma-2b on v5p and saw an average of 299 tflops/sec/device.  After this PR, that number became 290.  To verify that the performance difference was only due to the zeros, I took all of my changes except changed `randint` to `zeros`, and this indeed gave 299.

# Checklist

### DIFF
--- a/MaxText/input_pipeline/input_pipeline_interface.py
+++ b/MaxText/input_pipeline/input_pipeline_interface.py
@@ -41,31 +41,38 @@ class SyntheticDataIterator:
         SyntheticDataIterator.raw_generate_synthetic_data, out_shardings=data_pspec_shardings, static_argnums=0
     )
 
+    tokens = jax.random.randint(
+        jax.random.PRNGKey(0),
+        (config.global_batch_size_to_load, config.max_target_length + 1),
+        0,
+        config.vocab_size,
+        dtype=jnp.int32,
+    )
+
+    sequence_positions = jnp.arange(0, config.max_target_length + 1, dtype=jnp.int32).reshape(1, -1)
+    batch_positions = jnp.broadcast_to(sequence_positions, (config.global_batch_size_to_load, config.max_target_length + 1))
+    segmentation = jnp.ones((config.global_batch_size_to_load, config.max_target_length), dtype=jnp.int32)
+    self.data = (tokens, batch_positions, segmentation)
+
   def __iter__(self):
     return self
 
   def __next__(self):
     with self.mesh:
-      return self.data_generator(self.config)
+      return self.data_generator(self.config, self.data)
 
   @staticmethod
-  def raw_generate_synthetic_data(config):
+  def raw_generate_synthetic_data(config, data):
     """Generates a single batch of synthetic data"""
+    tokens, positions, segmentation = data
+
     output = {}
-    output["inputs"] = jax.numpy.zeros((config.global_batch_size_to_load, config.max_target_length), dtype=jax.numpy.int32)
-    output["inputs_position"] = jax.numpy.zeros(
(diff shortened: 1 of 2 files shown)


philip-essential:
The test failures look like unrelated errors downloading from Hugging Face, but let me know if there's anything I should change.  Our team has independently run into similar errors elsewhere today.

gobbleturk:
This looks awesome! I'm going to run profile before and after this - I'm curious where the diff is coming from (maybe due to attention mask)?

philip-essential:
Yeah, I'm also very curious about that, and I'd be interested if you find an answer.  I don't know of any explicit optimizations for zeros, but maybe some operations happen to complete faster in that case for some algorithmic reason.

gobbleturk:
In my tests run a v4-8 running default config (1B model with a `per_device_batch_size=12 max_target_len=2048`) I don't see any difference between real data, synthetic before this PR or synthetic after this PR - in all cases the data loading time is less than 14 micro seconds out of step of 1.17 seconds (according to the profile)

I like the more realistic data, I'll merge this through!

philip-essential:
Sounds good, thanks for looking into this.  For posterity, this is the config I was using on a v5p-8:

```
model_name: gemma-2b
per_device_batch_size: 8
max_target_length: 4096
enable_checkpointing: False
dataset_type: synthetic
```

Also, I notice that `zeros` and `ones` give exactly the same results, which suggests it isn't anything special about zero, just that all the tokens are the same.

I took a few quick profiles just now and I'm surprised that the attention op does not appear to cause the change.  Some numbers (all in ms):

| | random | zeros | relative difference |
| - | - | - | - |
| jit(train_step) | 1897 | 1843 | 2.9% | 
| backward pass (all layers) | 1.362 | 1.317 | 3.4% |
| per layer (forward pass) | 22.58 | 22.14 | 2.0% |
| mlp | 17.58 | 17.29 | 1.7% |
| attention op (excl projections) | 2.377 | 2.374 | 0.1% |
| attention projections | 1.59 | 1.39 | 14.4% |
| key projection | 0.311 | 0.111 | 280% |
| key projection's fusion op (matmul, i assume) | 0.094 | 0.094 | 0% |
| key projection's allgather | 0.20 | 0.008 | 2500% |

Although the key projection only accounts for around half the absolute difference in the forward pass, it's notable because it's such a big relative difference.  The strange part is that this is fsdp, so the allgather is for the weights (shape `(1,2048,1,256)` which matches gemma-2b's 2048 embedding dimension, 1 kv head, 256 head dim).

Anyway, I don't need to know exactly why it's different.  Thanks for merging!

With random data:
Screenshot 2025-01-30 at 4 25 32 PM
With all zeros (ones is virtually identical):
Screenshot 2025-01-30 at 4 26 30 PM

PR #1223, opened 2025-01-31
Maintain fp32 for optimizer state when offloading is enabled

zhenying-liu:
Address issues with optimizer state offloading and data type conversion.

We identified two issues concerning the conversion from fp32 to fp16 for the optimizer state when enabling optimizer state offloading:

The comparison between configurations without and with optimizer state offloading was unfair because the data sizes differed, with the former using fp32 and the latter using fp16.
The presence of two modules with jit_train_step due to separate versions for fp32 and fp16 created inconsistencies.

This commit removes the fp32 to fp16 conversion, ensuring that the optimizer state retains its original data type.

We observed no memory savings when switching from f16 to f32 previously. The root cause is that the GPU memory scheduler does not distinguish between CPU memory and GPU memory. [This XLA PR](https://github.com/openxla/xla/pull/21825) modifies the scheduler to exclude CPU memory and is merged so that we could reenable the CL (#1184) again.

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -65,10 +65,6 @@ def cast_dtype_from_to(nest, src, dst):
   return jax.tree_util.tree_map(lambda t: t.astype(dst) if t.dtype == src else t, nest)
 
 
-def cast_to_bf16(params):
-  return cast_dtype_from_to(params, np.float32, jnp.bfloat16)
-
-
 def find_nans_and_infs(pytree):
   def finder(x):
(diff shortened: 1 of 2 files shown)


zhenying-liu:
I verified there were GPU memory savings with this PR and the XLA fix when optimizer state offloading was turned on using llama2-7b. Also there is only one jit_train_step generated.

zhenying-liu:
Please take a look for the code review. The failure of "Error: No task list was present and requireChecklist is turned on" seems not a real failure.

khatwanimohit (approved):

PR #1228, opened 2025-02-03
PP: All gather once at start

gobbleturk:
# Description

Add an option to all gather weights over FSDP initially before pipeline begins. This is a memory/time tradeoff - with this option on we perform less all gathers and reduce scatters but need to store the weights and gradients without FSDP sharding.

Prior to this change, PP + FSDP had the behavior of all gather and reduce scattering the weights every microbatch. After this change with `pipeline_fsdp_ag_once=True`, we will initially AG all the weights once and RS them once, which should be a factor of `num_pipeline_microbatches` less communication.

I found that the pipeline module had to be a submodule of a decoder so we could reference (and manipulate) self.layers.variables inside of the pipeline - hence some refactoring of how a pipeline module is initialized. Additionally it was tricky to construct the sharding annotations necessary to AG over FSDP - the weights don't exist yet until calling `.init` on pipeline module, so we can't pass the sharding annotations as a property until after calling .init. This PR instead plumbs in an optional sharding constraint to `__call__`.

# Tests
On my v4-8 devbox saw a 20% speedup with FSDP=2, PP=2 pdb=1, seq=2k. Will test this more at scale. Also added an additional test in pipeline_parallelism_test.py, although we were already confident this should pass as this PR only modifies a sharding constraint. Potentially we would replace this implementation with a shard mapped one in which case the unit test is even more useful

# Checklist

### DIFF
--- a/MaxText/layers/pipeline.py
+++ b/MaxText/layers/pipeline.py
@@ -353,7 +353,44 @@ class Pipeline(nn.Module):
     output = output[:, permutation]
     return output
 
-  def get_main_vmap_func(self):
+  def get_current_stage_weights(self, pipeline_weights, loop_iteration):
+    """
+    Gets the current weights used for one iteration. Outputs a pytree whose arrays have leading dimension of stages, e.g.
+    {'mlp': 'wo': [stages, mlp, embed]}. Stage 0 will use the 0th index of this pytree, Stage 1 the 1st index, etc.
+    For non-circular pipelines, this simply returns all weights - every weight is used in every iteraiton. However
+    for ciruclar pipelines each stage grabs only the weights correpsonding to the current repeat.
+    """
+    if self.config.num_pipeline_repeats > 1:
+      return self.get_current_repeat_from_stages(pipeline_weights, loop_iteration)
+    else:
+      return pipeline_weights
+
+  def get_current_repeat_from_stages(self, weights, loop_iteration):
+    _, repeat_ids = self.get_microbatch_and_repeat_ids(loop_iteration)
+
+    def gather_weights_for_stages_in(weights):
+      return jax.tree.map(
+          functools.partial(
+              self.vmap_parallel_gather, repeat_ids=repeat_ids, repeat_dim_in_weights=0, stages_dim_in_weights=1
+          ),
+          weights,
+      )
+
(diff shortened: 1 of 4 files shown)


MaxText/layers/pipeline.py
```diff
@@ -452,6 +570,7 @@ def __call__(
       positions: jnp.ndarray,
       deterministic: bool,
       model_mode=common_types.MODEL_MODE_TRAIN,
+      partition_spec=None,  # Pytree of sharding specifications of the weights (aka self.layers.variables)
```
RissyRan, line 573: Oh nice. Does it works as expected?

RissyRan (commented):
LGTM! Thanks Matt! At high level, have 2 questions:

1) have you already had a test to ensure the same logits with/without `pipeline_fsdp_ag_once`? Or I missed it somewhere?
2) Do you want to attach profiles of 2 implementation and show the AG differences?

MaxText/layers/pipeline.py
gobbleturk: So far yes! If we find it doesn't we can replace the sharding constraint implementation with a shard map

gobbleturk:
> LGTM! Thanks Matt! At high level, have 2 questions:
> 
> 1. have you already had a test to ensure the same logits with/without `pipeline_fsdp_ag_once`? Or I missed it somewhere?
> 2. Do you want to attach profiles of 2 implementation and show the AG differences?

This PR includes a test in pipeline_parallelism_test which asserts that the output and gradients match a non-pipeline implementation.

I ran experiemnts on my v4-8 devbox with ici_pipeline_parallism=2 (fsdp=2) and pdb=2, seq=2k
* [trace](https://xprof.corp.google.com/trace_viewer/mattdavidow-5332985171866700224?hosts=t1v-n-5e50d777-w-0&host_index=0&trace_filter_config={}&view_start=-319.535&view_end=1422.812) with new feature false 
* [trace](https://xprof.corp.google.com/trace_viewer/mattdavidow-2261075635177290598?hosts=t1v-n-5e50d777-w-0&host_index=0&trace_filter_config={}&view_start=-171.229&view_end=1137.237) with  new feature true

PR #1265, opened 2025-02-12
Add prefix_cache module skeleton and Value inside

yuyanpeng-google:
# Description

The same as the title.

FIXES: b/389788256

# Tests

unittest

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed.

PR #1266, opened 2025-02-13
Implement PrefixCacheTrie

yuyanpeng-google:
# Description

Stack on https://github.com/AI-Hypercomputer/maxtext/pull/1265

Stores prefix tokens as a trie for fast lookup index of PrefixCache store in cache.

Insert longer Key replace shorter key to be the longest common prefix key. The shorter key will never be returned even if longer key is erased, and should got evicted in the future.

Assume Key is equal length to tokens, which can be used to slice prompt and cache Value. Should check the return key common prefix length by the caller.

If erase the Key not the leaf, nothing will happen. If erased key match at a leaf, delete the node and ancestors would be the leaf after deleted.

TESTED=unittest

FIXES: b/389788256

# Checklist

### DIFF
--- a/MaxText/prefix_cache.py
+++ b/MaxText/prefix_cache.py
@@ -57,6 +57,7 @@ if cached_prefix is None or matched_len != orig_key_len:
 """
 
 from typing import Tuple, Any, Optional
+import dataclasses
 import jax
 
 import max_logging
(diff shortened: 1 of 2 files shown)


yuyanpeng-google:
https://github.com/AI-Hypercomputer/maxtext/pull/1278
Please review in the overall PR

PR #1267, opened 2025-02-13
Add clone and eq in prefix_cache value

yuyanpeng-google:
On top of https://github.com/AI-Hypercomputer/maxtext/pull/1266

Clone to prevent use the same jax array.

TESTED=unittest

FIXES: b/389788256

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed.

PR #1268, opened 2025-02-13
Implement HBMCache

yuyanpeng-google:
On top of https://github.com/AI-Hypercomputer/maxtext/pull/1267

Value will be moved to the cache, which means cannot used the same value reference after add_to_cache.

The jax may modified the value even stored in another python reference. If the value need to be used after add_to_cache, make sure copy them before add_to_cache.

Return value copied from cache to avoid modified value in the cache, always copied the value before return.

TESTED=unittest
FIXES: b/389788256

# Checklist

### DIFF
--- a/MaxText/prefix_cache.py
+++ b/MaxText/prefix_cache.py
@@ -239,23 +239,35 @@ class HBMCache:
     Args:
       max_size_bytes: Maximum bytes of HBM to use for cache
     """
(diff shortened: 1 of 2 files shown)


yuyanpeng-google:
https://github.com/AI-Hypercomputer/maxtext/pull/1278
Please review in the overall PR

PR #1278, opened 2025-02-17
Prefix Caching with HBM and latency test

yuyanpeng-google:
# Description

Implements Prefix Caching in HBM and latency test in inference_microbenchmark.
 
Stores prefix tokens as a trie for fast lookup index of PrefixCache store in cache.

Insert longer Key replace shorter key to be the longest common prefix key.
The shorter key will never be returned even if longer key is erased, and should got evicted in the future.

Assume Key is equal length to tokens, which can be used to slice prompt and cache Value.
Should check the return key common prefix length by the caller.

If erase the Key not the leaf, nothing will happen.
If erased key match at a leaf, delete the node and ancestors would be the leaf after deleted.

Value will be moved to the cache, which means cannot used the same value reference after add_to_cache.

The jax may modified the value even stored in another python reference.
If the value need to be used after add_to_cache, make sure copy them before add_to_cache.

Return value copied from cache to avoid modified value in the cache, always copied the value before return.

Add PrefixCaching benchmark test in inference_microbenchmark.

Using half of the prefill_length as the common prefix and save 100 prefix in the cache.

Loading the cache (including jax.array.copy) appears to be independent of the prefill_length (tested with 128 and 1024), even though the saved cache sizes are different.

Using jax.profiler shows that the copy operation consumes a similar amount of time on TPU. This might be because the sizes aren't large or different enough to see a significant impact.

Part of results below
```
Prefix Cache benchmark results for prefill length 128:

PrefixCaching results:
	Per prefix size bytes: 0.124 GB
	Average save cache time: 12.142 ms
	Average fetch longest prefix time: 0.029 ms
	Average load cache time: 5.589 ms


Prefix Cache benchmark results for prefill length 1024:

PrefixCaching results:
	Per prefix size bytes: 0.220 GB
	Average save cache time: 12.987 ms
	Average fetch longest prefix time: 0.218 ms
	Average load cache time: 5.143 ms
```

FIXES: b/389788256
TESTED: unittest

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -525,6 +525,8 @@ vertex_tensorboard_region: ""
 max_checkify: False
 
 # Inference
+inference_microbenchmark_prefix_cache_entries_num: 100
+inference_microbenchmark_prefix_cache_common_prefix_proportion: 0.5
 inference_microbenchmark_prefill_lengths: "64,128,256,512,1024"
 inference_microbenchmark_stages: "prefill,generate"
 inference_microbenchmark_loop_iters: 10
--- a/MaxText/inference_microbenchmark.py
+++ b/MaxText/inference_microbenchmark.py
@@ -27,6 +27,7 @@ from jetstream.engine import token_utils
 import max_utils
 import maxengine
 import maxtext_utils
+import prefix_cache
 import profiler
 import pyconfig
 
@@ -37,6 +38,103 @@ warnings.simplefilter("ignore", category=FutureWarning)
 _WARMUP_ITERS = 2
 _FLATTEN_MICROBENCHMARK_RESULTS = False
 # pylint: disable=too-many-positional-arguments
+
+
+def prefix_cache_benchmark(
+    prefix, prefill_length: int, true_length: int, common_prefix_proportion: float, prefix_cache_entries_num: int, iters: int
+):
+  """Handles running prefix cache benchmark, and printing results.
+
+  Create different key with half of prefill_length common prefix insert into cache.
+  The value is not relevant to the cache for now. Just copy the prefix for every cache entry.
+  1. Fill the prefix cache to full capacity.
+  2. Benchmark save prefix cache with evicting time average by prefix_cache_entries_num.
+  3. Benchmark fetch_longest_common_prefix_key average by iters.
+  4. Benchmark load prefix cache time average by iters.
+
+  Args:
+    prefix: prefix return from prefill function
+    prefill_length: prefill token length after padding
+    true_length: true prefill token length
+    common_prefix_proportion: [0., 1.] common prefix proportion to the prefill_length
+    prefix_cache_entries_num: number of prefix cache entries insert into PrefixCache
+    iters: repeat time to test fetch_longest_common_prefix_key and load from cache
+  """
+
+  print(f"Prefix Cache benchmark results for prefill length {prefill_length}:\n")
+
+  value = prefix_cache.Value(
+      prefix=prefix,
+      true_length=true_length,
+      padded_length=prefill_length,
+      tokens=tuple(i for i in range(prefill_length)),
+  )
+  prefix_size_bytes_gb = value.prefix_size_bytes / 1024 / 1024 / 1024
+  prefix_cache_inst = prefix_cache.PrefixCache(prefix_cache_entries_num * value.prefix_size_bytes)
+  common_len = int(prefill_length * common_prefix_proportion)
+  remain_len = prefill_length - common_len
+  common_prefix_key = tuple(i for i in range(common_len))
+
+  # Fill the prefix caching
+  new_value_list = []
+  for c_idx in range(prefix_cache_entries_num):
+    # Add 100 to make sure filled prefix caching will not share the common_prefix_key.
+    # The later save prefix part will evict all of them.
+    key = tuple(100 + i + c_idx * prefill_length for i in range(prefill_length))
+    new_value = value.clone()
+    prefix_cache_inst.save(key, new_value)
+    new_value_list.append(new_value)
+  jax.block_until_ready(new_value_list)
+  del new_value_list
+
+  # Save prefix
+  new_value = None
+  save_sec = 0
+  for c_idx in range(iters):
+    key = common_prefix_key + tuple(i + c_idx * remain_len for i in range(remain_len))
+    # values are not relevant for caching now, just clone the same tokens and values for test
+    new_value = value.clone()
+    jax.block_until_ready(new_value)
+    start = datetime.datetime.now()
+    prefix_cache_inst.save(key, new_value)
+    end = datetime.datetime.now()
+    save_sec += (end - start).total_seconds()
+  del new_value
+  save_avg_ms = save_sec * 1000 / iters
+
+  # Fetch longest prefix key
+  key_load = common_prefix_key + tuple(i + prefix_cache_entries_num * remain_len for i in range(remain_len))
+  matched_key = None
+  fetch_sec = 0
+  for _ in range(iters):
+    start = datetime.datetime.now()
+    matched_key = prefix_cache_inst.fetch_longest_common_prefix_key(key_load)
+    end = datetime.datetime.now()
+    fetch_sec += (end - start).total_seconds()
+  fetch_avg_ms = fetch_sec * 1000 / iters
+
+  assert matched_key is not None
+
+  # Load prefix
+  load_sec = 0
+  value_load = None
+  for _ in range(iters):
+    start = datetime.datetime.now()
+    value = prefix_cache_inst.load(matched_key)
+    jax.block_until_ready(value)
+    end = datetime.datetime.now()
+    load_sec += (end - start).total_seconds()
+  del value_load
+  load_avg_ms = load_sec * 1000 / iters
+
+  print(
+      f"PrefixCaching results:\n"
+      f"\tPer prefix size bytes: {prefix_size_bytes_gb:.3f} GB\n"
+      f"\tAverage save cache time: {save_avg_ms:.3f} ms\n"
+      f"\tAverage fetch longest prefix time: {fetch_avg_ms:.3f} ms\n"
+      f"\tAverage load cache time: {load_avg_ms:.3f} ms\n\n\n"
+  )
+  del prefix_cache_inst
 
 
 def prefill_benchmark_loop(engine_prefill, params, tokens, true_length, iters):
@@ -304,6 +402,22 @@ def run_benchmarks(config):
       benchmark_results["prefill-result-sizes"][prefill_length] = summarize_prefill_result(
           prefill_executable[prefill_length], params, prefill_tokens[prefill_length], prefill_true_lengths[prefill_length]
       )
+
+    if "prefix_cache" in stages_to_benchmark:
+      for prefill_length in prefill_lengths:
+        rng_cache = jax.random.PRNGKey(1234)
+        prefill_result, _ = prefill_executable[prefill_length](
+            params, prefill_tokens[prefill_length], prefill_true_lengths[prefill_length], rng_cache
+        )
+        prefix_cache_benchmark(
+            prefill_result,
+            prefill_length,
+            prefill_true_lengths[prefill_length],
+            config.inference_microbenchmark_prefix_cache_common_prefix_proportion,
+            config.inference_microbenchmark_prefix_cache_entries_num,
+            benchmark_loop_iters,
+        )
+        del prefill_result
 
     for prefill_length in prefill_lengths:
       benchmark_results["prefill"][prefill_length] = prefill_benchmark(
--- a/MaxText/prefix_cache.py
+++ b/MaxText/prefix_cache.py
@@ -1,0 +1,411 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Implementation of PrefixCache with LRU cache and Trie based key lookup.
+
+i) Initialize cache
+
+# HBM size with Trillium 8 core use about 17 / 31.25 GB per core after loading models.
+# To utilize HBM memory to 80%, about (30 * 0.8 - 16) * 8 = 64GB
+# Prefix return by prefill function of mixtral-8x22b model with max_prefill_length=1024,
+# int8 quantize_kvcache is 235_930_060 bytes, nearly 256 MB.
+# HBM cache can store about 64GB / 256 MB = 256 prompts.
+hbm_bytes = 64 * 1024 * 1024 * 1024  # 64 GB
+prefix_cache = PrefixCache(hbm_bytes)
+
+ii) Read from Cache:
+
+# From request
+prompt: str = "blah"
+full_key: Key = tokenizer.tokenize(prompt)
+
+# Inside prefill call
+orig_key_len = len(full_key)
+
+matched_key = prefix_cache.fetch_longest_common_prefix_key(key)
+if matched_key is not None and load_cache_is_efficient_enough(matched_key):
+  cache_results = prefix_cache.load(matched_key)
+else:
+  cache_results = None
+
+if cache_results:
+   # load cache successfully
+   matched_len = calculate_matched_len(full_key, matched_key)
+   cached_prefix = cache_results.prefix
+else:
+   cached_prefix = None
+
+iii) Run prefill un-cached prompt and write to cache
+
+full_kv_cache = prefill(prompt, cached_prefix, matched_len)
+
+# if the cache didn't contain prefix or if it was a partial match
+if cached_prefix is None or matched_len != orig_key_len:
+    prefix_cache.save(full_key, Value(
+        prefix=full_kv_cache,
+        true_length=true_length,
+        padded_length=padded_length,
+        tokens=full_key,
+    ))
+
+"""
+
+from collections import OrderedDict
+from typing import Tuple, Any, Optional
+import dataclasses
+import jax
+import jax.numpy as jnp
+import logging
+import threading
+
+logger = logging.getLogger(__name__)
+
+Token = int
+# Tuple of tokens from prompt
+Key = Tuple[Token, ...]
+Prefix = Any  # KVCache for one prompt
+
+
+@jax.jit
+def tree_copy(tree):
+  return jax.tree.map(lambda x: x.copy() if isinstance(x, jax.Array) else x, tree)
+
+
+class Value:
+  """Object stored contains the actual KVcache
+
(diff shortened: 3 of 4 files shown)


MaxText/configs/base.yml
```diff
@@ -499,6 +499,7 @@ vertex_tensorboard_region: ""
 max_checkify: False
 
 # Inference
+inference_microbenchmark_cache_num: 100
```
mailvijayasingh, line 502: is this number of entries in the cache? Can we make it `num_entries_in_cache`? That way it will not be restricted to microbenchmarking

MaxText/inference_microbenchmark.py
```diff
@@ -39,6 +40,80 @@
...
+
+  value = prefix_cache.Value(
+      prefix=prefix,
+      true_length=true_length,
+      padded_length=prefill_length,
+      tokens=tuple(i for i in range(prefill_length)),
+  )
+  prefix_size_bytes_gb = value.prefix_size_bytes / 1024 / 1024 / 1024
+  prefix_cache_inst = prefix_cache.PrefixCache(cache_num * value.prefix_size_bytes)
+  common_len = prefill_length // 2
```
mailvijayasingh, line 56: can you abstract common_len factor to the config?

MaxText/tests/prefix_cache_test.py
```diff
@@ -0,0 +1,403 @@
...
+  def test_add_and_retrieve_multiple_keys(self):
+    hbm_cache = HBMCache(max_size_bytes=1_000_000)
+    value1 = create_default_value(tokens=[1])
+    hbm_cache.add_to_cache((1), value1)
+    value2 = create_default_value(tokens=[2])
+    hbm_cache.add_to_cache((2), value2)
+    assert hbm_cache.retrieve_from_cache((1)) == value1
+    assert hbm_cache.retrieve_from_cache((2)) == value2
+
+  def test_evict_cache(self):
```
mailvijayasingh, line 242: can you elaborate this test, add multiple values and then delete?

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+
+  def __init__(self, max_size_bytes: int):
+    """
+    Args:
+      max_size_bytes: Maximum bytes of HBM to use for cache
+    """
+    self._remain_size_bytes = max_size_bytes
+    self._saved_values: dict[Key, Value] = {}
+
+  def is_enough_space_remain(self, value: Value) -> bool:
```
vipannalla, line 278: nit: rename to `has_enough_space()`

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+    self._saved_values[key] = value
+    self._remain_size_bytes -= value.prefix_size_bytes
+    return True
+
+  def retrieve_from_cache(self, key: Key) -> Optional[Value]:
+    """Return value copied from cache or None if not found.
+    To avoid modified value in the cache, always copied the value before return.
+    """
+    if key in self._saved_values:
+      return self._saved_values[key].clone()
```
vipannalla, line 302: That looks inefficient, you are making copy of a large JAX array.

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+    logger.debug("fetch_longest_common_prefix_key, key=%r", key)
+    matched_key = self._trie.get_longest_common_prefix_key(key)
+    logger.debug("matched_key=%r", matched_key)
+    return matched_key
+
+  def save(self, key: Key, value: Value) -> bool:
+    """Save key/value to the cache."""
+    logger.debug("save key=%r", key)
+    if not self._hbm_cache.is_enough_space_remain(value):
+      self._evict_cache()
```
vipannalla, line 363: Pass `value.prefix_bytes_size` into `self._evict_cache()` to make check you evicted enough bytes. See below..

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+    return value
+
+  def clear(self):
+    """Clear entire cache."""
+    logger.debug("clear cache")
+    self._hbm_cache = HBMCache(max_size_bytes=self._hbm_bytes)
+    self._trie = PrefixCacheTrie()
+    self._cache_strategy = LRUStrategy()
+
+  def _evict_cache(self) -> Optional[Value]:
```
vipannalla, line 388: You need to evict until you make enough space:
```
def _evict_cache(self, bytes_needed: int) -> Optional[List[Value]]:
  evicted_bytes = 0
  evicted_values = []
  while evicted_bytes < bytes_needed:
     key = self._cache_strategy.evict()
     if key is None:
       logger.debug("no key to evict")
       return None
     logger.debug("evict key=%r", key)
     value = self._hbm_cache.evict_cache(key)
     if value is None:
       logger.warning("key=%r should exist in HBM cache.", key)
     self._trie.erase(key)
     evicted_values.append(value)
     
  return evicted_values
  ```

File: MaxText/prefix_cache.py [hunk 8]
```diff
@@ -0,0 +1,400 @@
...
+  def __init__(self, hbm_bytes: int):
+    """
+    Args:
+      hbm_bytes: Total amount of HBM to use for cache.
+    """
+    self._hbm_bytes = hbm_bytes
+    # init in clear()
+    self._hbm_cache: HBMCache = None
+    self._trie: PrefixCacheTrie = None
+    self._cache_strategy: LRUStrategy = None
```
vipannalla, line 349: You need a lock since the cache will get updated from multiple threads.
 ```
   self._lock = threading.RLock()
 ```

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+  def fetch_longest_common_prefix_key(self, key: Key) -> Optional[Key]:
+    """Returns key with longest common prefix matched or None if not found."""
+    logger.debug("fetch_longest_common_prefix_key, key=%r", key)
+    matched_key = self._trie.get_longest_common_prefix_key(key)
+    logger.debug("matched_key=%r", matched_key)
+    return matched_key
+
+  def save(self, key: Key, value: Value) -> bool:
+    """Save key/value to the cache."""
+    logger.debug("save key=%r", key)
```
vipannalla, line 361: Grab lock before saving:
```
  with self.lock:
    if not self._hbm_cache.is_enough_space_remain(value):
      self._evict_cache()
     ...
```

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+    if len(self._order) == 0:
+      return None
+    return self._order.popitem(last=False)[0]
+
+  def use(self, key: Key) -> None:
+    """Updated the usage history."""
+    if key not in self._order:
+      self._order[key] = None
+    else:
+      self._order.move_to_end(key, last=True)
```
vipannalla, line 331: +1 for abstracting into separate class. We should pass the `LRUStrategy` into `HBMCache` and let `HBMCache` evict the key based on the strategy.

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,400 @@
...
+    if not self._hbm_cache.add_to_cache(key, value):
+      logger.debug("cannot add to cache even after evict")
+      return False
+    self._trie.insert(key)
+    self._cache_strategy.use(key)
+    return True
+
+  def load(self, key: Key) -> Optional[Value]:
+    """Returns Value stored with key or None if not found."""
+    logger.debug("load key=%r", key)
```
vipannalla, line 373: Same here, grab the lock...

vipannalla (commented):
Looks good, I left few comments.

MaxText/configs/base.yml
yuyanpeng-google: The number was intended to facilitate inserting various suffixes into the cache easily, but I think that it's not suitable as a hard limitation.

We can optimize memory usage by saving identical prefix values shared across multiple keys. For instance, [1, 2, 3] and [1, 2, 3, 4] can share the same [1, 2, 3] value, so we only need to store one [1, 2, 3, 4] value for both entries. Adding a parameter with a maximum entries limitation could complicate the use case. I believe the optimal cache size should be determined through benchmarking under various scenarios rather than imposing a limit in the API.

MaxText/prefix_cache.py
yuyanpeng-google: Thank for the suggestion.

I'd like to keep the HBMCache implementation as simple as possible and not concern it with caching strategies.

I'd prefer to encapsulate the caching strategy within the outermost PrefixCache class. This approach would make it easier to maintain the strategy, especially if it depends on other cache layers or has other special considerations.

yuyanpeng-google:
Modify the recursive erase to prevent stack recursive limit in long context.
Fix the comments.
Rebase to main.

MaxText/configs/base.yml
yuyanpeng-google: e1b66416231c3a3a7a9df48dcee868c9e5da48a8 adjust the name

MaxText/inference_microbenchmark.py
yuyanpeng-google: f5f13ff55c78a81e83877cc13f863b4f9f2ae68a add into config

MaxText/tests/prefix_cache_test.py
yuyanpeng-google: 4b1dd4d3fcc01a9869747c18ec7b4331d4674bc1 add more test and description

MaxText/prefix_cache.py
yuyanpeng-google: ddee2f401ddb24fabe92b368e268c1e8a710dac8 rename it

MaxText/prefix_cache.py
yuyanpeng-google: d734a4c119194de97f558ce944c17986cfeb234c avoid copy and add caution in the docstring to prevent modify the return cache before clone it

MaxText/prefix_cache.py
yuyanpeng-google: 3449858eac783852c2d3aff1853f61edd35b3f42 use different approach, add while loop outside the _evict_cache

MaxText/prefix_cache.py
yuyanpeng-google: 754836fc948624cefefcad3613780287dcbd92de add lock

mailvijayasingh (approved):
LGTM!

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+
+"""Implementation of PrefixCache with LRU cache and Trie based key lookup.
+
+i) Initialize cache
+
+# HBM size with Trillium 8 core use about 17 / 31.25 GB per core after loading models.
+# To utilize HBM memory to 80%, about (30 * 0.8 - 16) * 8 = 64GB
+# Prefix return by prefill function of mixtral-8x22b model with max_prefill_length=1024,
+# int8 quantize_kvcache is 235_930_060 bytes, nearly 256 MB.
+# HBM cache can store about 64GB / 256 MB = 256 prompts.
```
liurupeng, line 23: qq, do we need memory in HBM for kv cache? Are we considering using DRAM for prefix store

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+   matched_len = calculate_matched_len(full_key, matched_key)
+   cached_prefix = cache_results.prefix
+else:
+   cached_prefix = None
+
+iii) Run prefill un-cached prompt and write to cache
+
+full_kv_cache = prefill(prompt, cached_prefix, matched_len)
+
+# if the cache didn't contain prefix or if it was a partial match
```
liurupeng, line 53: maybe elaborate the strategy if it's a partial match? thanks

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+  @property
+  def tokens(self) -> list[int]:
+    return self._tokens
+
+  @property
+  def prefix_size_bytes(self) -> int:
+    return self._prefix_size_bytes
+
+  def clone(self) -> "Value":
+    """Clone to prevent use the same jax array."""
```
liurupeng, line 145: using
```suggestion
    """Clone to prevent using the same jax array."""
```

MaxText/prefix_cache.py
mailvijayasingh: DRAM heirarchical cache is next step I think. This particular implementation is only for HBM cache.

MaxText/prefix_cache.py
mailvijayasingh: hmm, I think the partial part is abstracted in prefill chunking logic and we will only look for exact match for a chunk, in the larger picture?

MaxText/prefix_cache.py
liurupeng: I see, but is the calculation here accurate, it seems the "HBM consumption" = "prefix prompt cache" + "model weights"

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+    for token in key:
+      if token not in node.children:
+        break
+      node = node.children[token]
+      result_tokens.append(token)
+
+    if len(result_tokens) == 0:
+      return None
+
+    while not node.is_leaf():
```
liurupeng, line 239: could you elaborate the logic here for handling a non-child node? if we find the longest match and the current node is not a leaf, we will append the next token of its children into the result?

MaxText/prefix_cache.py
mailvijayasingh: total HBM consumption is yes - "prefix prompt cache" + "model", yes. 
I guess max_prefill_length is placeholder for prefix length that is to be stored in cache
Yuyan can pitch in.

liurupeng: yes, my point is that we should reserve HBM for decode's kv cache as well, it seems to be the prefix cache has used too much HBM

mailvijayasingh: correct, I guess the total size reserved for prefix cache can be tuned or changed.

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+  If erased key match at a leaf, delete the node and ancestors would be the leaf after deleted.
+  """
+
+  @dataclasses.dataclass
+  class Node:
+    """Trie Node."""
+
+    parent: Optional["PrefixCacheTrie.Node"] = None
+    token: Optional[Token] = None
+    children: dict[Token, "PrefixCacheTrie.Node"] = dataclasses.field(default_factory=dict)
```
liurupeng, line 202: the children's dictionary key here is the full_key? for example, if it's "the weather today" in the tri, then the tree could look like "the" -> "the weather" -> "the weather today"; so we stored the prefix consuming actually O(len^2) memory. The ideal way might be only storing the next token in the tri which is "the"->"weather"->"today", this previous approach will make you have to traverse the tri in a style of "for token in key", which key is the full prefix, and the time complexity of insert and search is O(layers) * O(sen_len) which the optimal one should be O(layers) * c (c is a constant)

MaxText/prefix_cache.py
```diff
@@ -0,0 +1,410 @@
...
+    if true_length > len(tokens):
+      logger.warning("true_length=%d should <= len(tokens)=%d.", true_length, len(tokens))
+
+    return min(true_length, len(tokens))
+
+
+class PrefixCacheTrie:
+  """Stores prefix tokens as a trie for fast lookup index of PrefixCache store in cache.
+
+  Insert longer Key replace shorter key to be the longest common prefix key.
```
liurupeng, line 186: sorry I didn't full understand the strategy here. So whenever we found a longer prefix, we will replace the shorter prefix with the longer prefix's kv cache?

liurupeng: I think this approach will make the cache hit rate not high

MaxText/prefix_cache.py
liurupeng: I see, for example, for prompt that is 20 tokens long, and 10 tokens partial match, we will use the first matched 10 tokens' cached kv with the next 10 not prefilled kv for chunked prefill?

MaxText/prefix_cache.py
vipannalla: We should update the comment mentioning this is an example to avoid confusion to the readers. In practice, we should use very small amount of HBM for caching (lets say 10%) and keep most of the cache in CPU RAM (v6e-8 has 1.7TB).
Given this is tunable, the actual usage depends on the deployment config.

MaxText/prefix_cache.py
vipannalla: Thanks

MaxText/prefix_cache.py
yuyanpeng-google: Yes. It is just an example calculation. The reserved size should be determined by benchmarking in a real scenario. Perhaps I should revise the description more precisely. Thanks for the feedback.

The max_prefill_length is for calculating the prompt padding to max_prefill_length here.

MaxText/prefix_cache.py
yuyanpeng-google: Yes, I believe incorporating the partial match prefix with chunked prefill could reduce the time spent calculating the prefix prompt. I will behind this abstract prefill call.

MaxText/prefix_cache.py
yuyanpeng-google: The longest key will overwrite shorter ones. To better utilize memory within the cache, we aim to store only the longest prefix cache, if feasible. Therefore, we only store key/value pairs from the root node to the leaf nodes in the cache. If the prompt match ends at a non-leaf node, the function will traverse to any leaf node from that point and return the longest key associated with that leaf.

The shorter key/value will not be evicted immediately while a longer one added. It will not be used anymore and be evicted after a while.

MaxText/prefix_cache.py
yuyanpeng-google: I'm not sure I completely understand what you mean.
In your example, the keys in the Trie would be 'the' -> 'weather' -> 'today'. A Token is represented by a single integer, and a Key is a Tuple[Token].
Regarding the value, for simplicity, it's not stored by every token separately at this time. We store the longest key/value pair. For instance, if 'the weather today' exists, storing it will cover the need for prefixes like 'the' and 'the weather'. If 'the weather tomorrow' is also present, then two key/value pairs will be stored. This may be optimized in the future.

liurupeng:
Synced with Yuyan offline, I think the current approach looks good

liurupeng (approved):

MaxText/prefix_cache.py
yuyanpeng-google: Synced with Rupeng offline, we can essentially assume a large amount of host DRAM is avaliable. Therefore the bottleneck is more likely to be the bandwidth transfer rather than memory usage. A shorter Prefix Cache less data to be transfer. If shorter prefixes are matched more frequently than longer ones, it could be beneficial to store and transfer the shorter prefixes instead.
This is a potential area for future optimization. Thanks for the feedback.

MaxText/prefix_cache.py
wang2yn84: Any updates on this? I have the same concern.

yuyanpeng-google: The key in trie is saved by Token, which is a single int, each layer.
The implement is iteratively traverse the Key once, not recursively traverse the whole Key each layer.

yuyanpeng-google:
Rebase and squash

vipannalla (approved):
looks good

yuyanpeng-google:
Fix pytype and modify the interface prefix_cache.Value.tokens to tuple[Token, ...]

PR #1279, opened 2025-02-18
Multi-Head Latent Attention for training

gagika:
# Description

Multi-Head Latent Attention for training for DeepSeek v3

Reference code: https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/model.py

b/394626699

# Tests
MLA attention test is included, this also will be tested in e2e DeepSeek v3 tests.

# Checklist

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -17,7 +17,7 @@
 import enum
 import functools
 import math
-from typing import Any, Optional
+from typing import Any, Optional, Tuple
 
 from flax import linen as nn
 import jax
@@ -43,6 +43,7 @@ from layers import quantizations
 class AttentionType(enum.Enum):
   GLOBAL = "global"
   LOCAL_SLIDING = "local_sliding"
+  MLA = "mla"
 
 
 Array = common_types.Array
@@ -52,7 +53,9 @@ Mesh = common_types.Mesh
 PRNGKey = common_types.PRNGKey
 
 DenseGeneral = linears.DenseGeneral
+RMSNorm = linears.RMSNorm
 RotaryEmbedding = embeddings.RotaryEmbedding
+YarnRotaryEmbedding = embeddings.YarnRotaryEmbedding
 NdInitializer = initializers.NdInitializer
 Quant = quantizations.AqtQuantization
 KVQuant = quantizations.KVQuant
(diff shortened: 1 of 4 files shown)


gobbleturk (approved):

MaxText/layers/attentions.py
```diff
@@ -1215,19 +1244,43 @@ def out_projection(self, output_dim: int, out: Array) -> Array:
...
+    Returns:
+      The input tensor with rotary embeddings applied.
+    """
+    if self.config.attention_type == AttentionType.MLA.value:
+      # For MLA attention RoPE is applied to only `self.qk_rope_head_dim` portion the heads.
+      rope_embedding_dims = self.qk_rope_head_dim
+    else:
+      rope_embedding_dims = self.head_dim
+
+    rope_type = self.config.rope_type.lower()
```
RissyRan, line 1263: Should we add a validation in [pyconfig file](https://github.com/AI-Hypercomputer/maxtext/blob/10b24ab5d56aa128d8b5a22ea185e4115b70860c/MaxText/pyconfig.py#L4)? i.e. rope type can only be one of "default", "llama3.1" or "yarn"

MaxText/layers/attentions.py
```diff
@@ -1295,36 +1348,210 @@ def __call__(
...
+    # Assert required configuration parameters for MLA attention.
+    assert self.config.attention_type == AttentionType.MLA.value, "MLA requires MLA attention type"
+    assert self.kv_lora_rank > 0, "KV LoRA rank must be > 0"
+    assert self.qk_nope_head_dim > 0, "QK NoPe head dim must be > 0"
+    assert self.qk_rope_head_dim > 0, "QK RoPE head dim must be > 0"
+    assert self.v_head_dim > 0, "V head dim must be > 0"
+    assert self.num_query_heads == self.num_kv_heads, "MLA requires equal number of query and kv heads"
+    assert not self.config.fused_qkv, "Fused QKV is not supported for MLA"
+
+    if self.q_lora_rank == 0:
```
RissyRan, line 1392: What's the use case for this branch? Inference?

MaxText/tests/attention_test.py
```diff
@@ -721,5 +722,164 @@ def test_sliding_window_attention(self):
...
+    )
+
+    return lnx, decoder_segment_ids, decoder_positions
+
+  @parameterized.named_parameters(
+      {"testcase_name": "RoPE_Yarn_Autoregression", "rope_type": "yarn"},
+      {"testcase_name": "Default_Autoregression", "rope_type": "default"},
+  )
+  @pytest.mark.tpu_only
+  def test_autoregression(self, rope_type):
```
RissyRan, line 826: Thanks for the test! Could you help me understand more about it? i.e. what could be difference between MLA full and prefill? should they be the same by calling mla.apply?

MaxText/layers/attentions.py
gagika: query lora path is not core part of MLA, as it won't effect kv cache size, that's why there are two paths, but by default it query go through low rank projection

philip-essential: Notably, deepseek released three configs for different model sizes, and in the [smallest version](https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/configs/config_16B.json) they do use `q_lora_rank == 0`.  I don't know if that means they think it's better at small scales, but my team does want to experiment with this option.

MaxText/tests/attention_test.py
gagika: prefill path will be used for inference (prefill phase) and will fill kv cache:

https://github.com/AI-Hypercomputer/maxtext/blob/10b24ab5d56aa128d8b5a22ea185e4115b70860c/MaxText/layers/attentions.py#L968

the prefill kv cache will also be used in decode `model_mode=common_types.MODEL_MODE_AUTOREGRESSIVE` one MLA kv cache update work is done for inference.

MaxText/layers/attentions.py
philip-essential: Yeah, I have not seen those weights.  I assume they're reasonable ways to scale down the model, but some of the choices are interesting (esp. `q_lora_rank == 0` and `n_shared_experts == 2`) so if anyone has seen discussion of those configs I'd be interested to see it.

MaxText/layers/attentions.py
gagika: Done

MaxText/layers/attentions.py
RissyRan: Thanks! Then keep this branch makes sense.

MaxText/tests/attention_test.py
RissyRan: OK, this is more like a test between training and inference?

gagika: yes, connecting all 3 mode of forward pass:  `MODEL_MODE_TRAIN` ,  `MODEL_MODE_PREFILL`, `MODEL_MODE_AUTOREGRESSIVE`.

PR #1280, opened 2025-02-18
Fix bugs in Pathways Path for Maxtext Benchmark Runner

SujeethJinesh:
# Description

This PR fixes a few bugs and adds additional Pathways support.

1. Fixes a regressions in https://github.com/AI-Hypercomputer/maxtext/pull/1094 that requires the use of remote python to run a pathways workload.
2. If use_vertex_tensorboard is not defined, pyconfig.py would throw an exception because it cannot parse a None value when it requires an argument with "=" in it. This is fixed in this PR.
3. Adds support for passing Server, Proxy, and Worker arguments to the pods directly (builds off of https://github.com/AI-Hypercomputer/xpk/pull/364)
4. Adds support for passing XLA flags into the proxy server for all configs by default.
5. Replace use of `--use-pathways` with `create-pathways` instead.

# Tests

Ran on v6e clusters.

# Checklist

### DIFF
--- a/benchmarks/maxtext_xpk_runner.py
+++ b/benchmarks/maxtext_xpk_runner.py
@@ -50,10 +50,13 @@ class LibTpuType(enum.Enum):
 
 @dataclasses.dataclass
 class PathwaysConfig:
-  server_image: str
-  proxy_image: str
-  runner_image: str
-  remote_python_sidecar_image: str
+  server_image: str = None
+  proxy_server_image: str = None
+  runner_image: str = None
+  remote_python_sidecar_image: str = None
+  server_flags: str = ''
+  proxy_flags: str = ''
+  worker_flags: str = ''
 
 
 # TODO(@vbarr): Split out parameters related to XPK workload and a General workload
@@ -284,7 +287,7 @@ def build_user_command(
 
   install_libtpu_cmd = ''
   jax_platforms = None
-  vertex_tensorboard = None
+  vertex_tensorboard = ''
   # TODO() support modifying nightly / stable dependencies in pathway flow
   if is_pw_enabled:
     jax_platforms = 'proxy'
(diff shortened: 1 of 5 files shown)


benchmarks/maxtext_xpk_runner.py
```diff
@@ -53,7 +53,10 @@ class PathwaysConfig:
   server_image: str
   proxy_image: str
   runner_image: str
-  remote_python_sidecar_image: str
+  remote_python_sidecar_image: str = None
```
RoshaniN, line 56: Nit: Can we keep the initializations consistent?

benchmarks/maxtext_xpk_runner.py
```diff
@@ -333,6 +336,43 @@ def build_user_command(
...
+      if pw_config.remote_python_sidecar_image is not None
+      else ''
+  )
+  server_image_flag = (
+      f' --server-image={pw_config.server_image}'
+      if pw_config.server_image is not None
+      else ''
+  )
+  proxy_image_flag = (
+      f' --proxy-server-image={pw_config.proxy_image}'
```
RoshaniN, line 355: Nit - Could we keep XPK flags and runner flags be names similar, so that there are lesser human errors while using the benchmark runner?

RoshaniN (approved):
Thanks Sujeeth

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: Yup, we tested without Remote Python on v6e. For better consistency I default the server/proxy/image to None. By default we do put values into server/proxy/runner image https://github.com/AI-Hypercomputer/maxtext/blob/8350feff91d8072c0fa9537bcf32b1d846c9a3b2/benchmarks/benchmark_runner.py#L251-L257

And

https://github.com/AI-Hypercomputer/maxtext/blob/8350feff91d8072c0fa9537bcf32b1d846c9a3b2/benchmarks/benchmark_runner.py#L111-L135

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: Fixed.

SujeethJinesh (commented):
Thanks Roshani!

shauryagup (approved):

gobbleturk (approved):

PR #1281, opened 2025-02-18
Support emergency checkpointing on GPUs

keshavb96:
This PR makes two changes to allow emergency checkpointing when running on GPUs:

1. **Create separate emergency ckpt directories for each process**

Running Maxtext with local/emergency checkpointing and 1 process per GPU leads to a duplicate directory runtime error currently. This is because the local_checkpoint_dir is only unique per node, and when running with one process per GPU, all processes on a node attempt to write to the same local checkpoint. 

2. **Enable emergency checkpointing when running on GPU hardware**

Currently, when emergency checkpointing is enabled Maxtext assumes that the hardware is tpu.

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -110,6 +110,16 @@ def create_orbax_emergency_checkpoint_manager(
   flags.FLAGS.experimental_orbax_use_distributed_process_id = True
   max_logging.log("Creating emergency checkpoint manager...")
 
+  # Only create directories if running on GPUs as the previous
+  # directory structure might be assumed by TPUs
+  if global_mesh.devices.flatten()[0].platform == 'gpu':
+    # pylint: disable=protected-access
+    local_checkpoint_dir = f"{local_checkpoint_dir}/{jax._src.distributed.global_state.process_id}"
+    local_p = epath.Path(local_checkpoint_dir)
+    persistent_p = epath.Path(persistent_checkpoint_dir)
+    local_p.mkdir(exist_ok=True, parents=True)
+    persistent_p.mkdir(exist_ok=True, parents=True)
+
   options = emergency_checkpoint_manager.CheckpointManagerOptions(
       local=LocalCheckpointOptions(save_interval_steps=local_save_interval_steps),
       persistent=PersistentCheckpointOptions(save_interval_steps=persistent_save_interval_steps),
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -159,7 +159,13 @@ def maybe_initialize_jax_distributed_system(raw_keys):
     if not raw_keys["enable_emergency_checkpoint"]:
       jax.distributed.initialize(initialization_timeout=raw_keys["jax_distributed_initialization_timeout"])
     else:
-      initialize_jax_for_tpu_with_emergency_checkpointing(raw_keys)
+      if raw_keys["hardware"] == "gpu_multiprocess":
+        max_logging.log("Initializing jax distribtued to support local checkpointing with GPUs...")
+        jax.distributed.initialize(initialization_timeout=raw_keys["jax_distributed_initialization_timeout"])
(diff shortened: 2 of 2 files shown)


MaxText/checkpointing.py
```diff
@@ -102,6 +102,12 @@ def create_orbax_emergency_checkpoint_manager(
   flags.FLAGS.experimental_orbax_use_distributed_process_id = True
   max_logging.log("Creating emergency checkpoint manager...")
 
+  local_checkpoint_dir += f"{local_checkpoint_dir}/{jax.process_index()}"
```
xuefgu, line 105: The process index, seems to me, should be `jax._src.distributed.global_state.process_id`, as the JAX distributed ID is used in this case (see line 102).

MaxText/checkpointing.py
```diff
@@ -102,6 +102,12 @@ def create_orbax_emergency_checkpoint_manager(
...
+  local_p = epath.Path(local_checkpoint_dir)
+  persistent_p = epath.Path(persistent_checkpoint_dir)
+  local_p.mkdir(exist_ok=True, parents=True)
```
xuefgu, line 108: I would limit this directory creation to the GPU case. Other functionalities of TPUs may depend the implicit assumption on this directory structure.

MaxText/max_utils.py
```diff
@@ -165,7 +165,12 @@ def maybe_initialize_jax_distributed_system(raw_keys):
     if not raw_keys["enable_emergency_checkpoint"]:
       jax.distributed.initialize(initialization_timeout=raw_keys["jax_distributed_initialization_timeout"])
     else:
-      initialize_jax_for_tpu_with_emergency_checkpointing(raw_keys)
+      if raw_keys["hardware"] == "gpu_multiprocess":
```
xuefgu, line 168: Adding a log line to distinguish between the two cases would be helpful.

xuefgu (changes requested):
Thanks for the PR. Would be great to see some test results on GPUs with this change!

keshavb96:
@xuefgu Resolved all comments and issues :)

MaxText/checkpointing.py
```diff
@@ -102,6 +102,15 @@ def create_orbax_emergency_checkpoint_manager(
   flags.FLAGS.experimental_orbax_use_distributed_process_id = True
   max_logging.log("Creating emergency checkpoint manager...")
 
+  # Only create directories if running on GPUs as the previous
+  # directory structure might be assumed by TPUs
+  if global_mesh.devices.flatten()[0].platform == 'gpu':
+    local_checkpoint_dir = f"{local_checkpoint_dir}/{jax._src.distributed.global_state.process_id}"
```
xuefgu, line 108: Do we have testing data from a reasonably large scale? I see that `jax._src.distributed.global_state.process_id` instead of `jax.process_index()` is being used now. Did it change things?

keshavb96: Seems like it doesn't change anything. Moreover, the two values are the same, jax.process_index() == jax._src.distributed.global_state.process_id in every run that I have tried so far. I changed it just for consistency and in line with your previous comment :)

xuefgu: @cpgaffney1 FYI on the observation on GPUs about `jax.process_index()` and the distributed ID.

@keshavb96 At how big a scale have you tested the new code?

xuefgu: BTW, we would be a lot more confident if both the saving and restore paths were tested and work well.

keshavb96: I haven't tested it at a large scale yet. Just upto 3 nodes so far. I will make sure the restore path works as well.

keshavb96: Hi @xuefgu, I have verified that the restore path works well in the case where all slices have their emergency checkpoints intact as well as when some have them missing. Can we get this PR merged?

xuefgu (approved):
LGTM, thank you!

PR #1285, opened 2025-02-19
Restructure Pyconfig

A9isha:
# Description

Restructure pyconfig.py such that we can create multiple configs. This becomes important if we have a need from an RL workload to have training and inference configs which differ in sharding strategies among others. 
We also now use [OmegaConfig](https://omegaconf.readthedocs.io/en/2.3_branch/) instead of `yaml.safe_load` and get its benefits

FIXES: b/396747968

# Tests

Tested this code with Llama2-7b on v4-8

# Checklist


MaxText/maxengine.py
```diff
@@ -54,7 +54,7 @@
 
 class MaxEngineConfig:
   """Engine specific config class to allow using multiple MaxEngine instances in an inference run.
-  The default pyconfig.config is a global param shared across multiple instances and doesn't
+  The default pyconfig.initialize is a global param shared across multiple instances and doesn't
```
shralex, line 57: Is this statement still correct ? initialize is a function returning a new config, right ?

end_to_end/tpu/llama2/7b/test_llama2_7b.sh
```diff
@@ -13,7 +13,7 @@ set -ex
 idx=$(date +%Y-%m-%d-%H-%M)
 
 # Non-Googlers please remember to point `BASE_OUTPUT_DIRECTORY` to a GCS bucket that you own, this bucket will store all the files generated by MaxText during a run
-export BASE_OUTPUT_DIRECTORY=gs://runner-maxtext-logs
+export BASE_OUTPUT_DIRECTORY=gs://mazumdera-test-bucket-europe-west4/pyconfig
```
shralex, line 16: should we revert this change ?

MaxText/maxengine.py
A9isha: You are right! I have added a TODO to reevaluate the need for this MaxEngineConfig in the light of the restructured pyconfig.py

end_to_end/tpu/llama2/7b/test_llama2_7b.sh
A9isha: Yes, sorry - reverted it now.

shralex:
if you'd like to add a unit test, you could for example test that invoking initialize twice creates two different configs, and that attempts to modify a config will result in an error

A9isha:
> if you'd like to add a unit test, you could for example test that invoking initialize twice creates two different configs, and that attempts to modify a config will result in an error

Done - thank you Alex!

MaxText/maxengine.py
A9isha: Right right, but I think it might be a good idea to investigate if we can get rid of `MaxEngineConfig` altogether

shralex (approved):

khatwanimohit (approved):
LGTM

bvandermoon (approved):

gobbleturk (approved):
Awesome work, thanks Anisha!

requirements.txt
```diff
@@ -37,3 +37,4 @@ mlperf-logging@git+https://github.com/mlperf/logging.git
 google-jetstream
 jsonlines
 pathwaysutils@git+https://github.com/google/pathways-utils.git
+omegaconf
```
gpupuck, line 40: @A9isha Could you make sure there's a newline at the end of the file?

sixiang-google:
Hi Anisha, with this PR I am seeing this error in maxengine.py:

`  File "/home/sixiang_google_com/maxtext/MaxText/inference_mlperf/offline_mode.py", line 467, in main
    engine = create_engine_from_config_flags(
  File "/home/sixiang_google_com/maxtext/MaxText/maxengine.py", line 970, in create_engine_from_config_flags
    cfg = MaxEngineConfig(cp.deepcopy(pyconfig._config.keys))  # pylint: disable=protected-access
AttributeError: module 'pyconfig' has no attribute '_config'`

Please take a look. Thanks

requirements.txt
A9isha: Sorry about that, fixing it here:
https://github.com/AI-Hypercomputer/maxtext/pull/1325

gpupuck: thanks!

PR #1288, opened 2025-02-19
Add MLA and YarnRope configs in DeepSeek v3

gagika:
# Description

Add MLA and YarnRope configs in DeepSeek v3

# Tests

Run at small scale with: 
python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=/tmp/ run_name=deepsee_training per_device_batch_size=4 enable_checkpointing=false model_name=deepseek3-671b ici_fsdp_parallelism=4 steps=5 async_checkpointing=false tokenizer_path=assets/tokenizer.mistral-v1 attention=dot_product dtype=bfloat16 weight_dtype=bfloat16 dataset_type=synthetic sparse_matmul=True megablox=True sharding_tolerance=0.3

# Checklist

### DIFF
--- a/MaxText/configs/models/deepseek3-671b.yml
+++ b/MaxText/configs/models/deepseek3-671b.yml
@@ -22,7 +22,6 @@ base_mlp_dim: 18432
 base_moe_mlp_dim: 2048
 base_num_decoder_layers: 61
 first_num_dense_layers: 3
-head_dim: 128
 mlp_activations: ["silu","linear"]
 vocab_size: 32000 # TODO(b/394635939): update after adding tokenizer 
 enable_dropout: False
@@ -34,5 +33,12 @@ shared_experts: 1
 routed_scaling_factor: 2.5
 routed_score_func: "sigmoid"
 routed_bias: True
-rope_max_timescale: 10_000
+# MLA
+attention_type: "mla"
+q_lora_rank: 1536
+kv_lora_rank: 512
+qk_nope_head_dim: 128
+qk_rope_head_dim: 64
+v_head_dim: 128
+rope_type: "yarn"
 decoder_block: "deepseek"
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -210,6 +210,12 @@ final_logits_soft_cap: 0.0
 use_post_attn_norm: False
 use_post_ffw_norm: False
 
+# MLA parameters
+q_lora_rank: 0
+kv_lora_rank: 512
+qk_nope_head_dim: 128
+qk_rope_head_dim: 64
+v_head_dim: 128
 
 # Combine matmuls for QKV and MLP
 fused_qkv: False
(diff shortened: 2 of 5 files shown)


MaxText/configs/models/deepseek3-671b.yml
```diff
@@ -34,5 +34,13 @@ shared_experts: 1
 routed_scaling_factor: 2.5
 routed_score_func: "sigmoid"
 routed_bias: True
-rope_max_timescale: 10_000
+max_target_length: 16384 # 4096 * 4
```
RissyRan, line 37: Do we need this in a model config? Or we could just pass in the script during the run?

RissyRan (commented):
LGTM! Just one comment about sequence length.

RissyRan (commented):
What's the `sharding_tolerance` in your cmd for the test?

MaxText/configs/models/deepseek3-671b.yml
gagika: DeepSeek provides the default in model config https://huggingface.co/deepseek-ai/DeepSeek-V3-Base/blob/main/inference/model.py#L22 but yes, we can just provide as an argument during training or inference. I will remove it.

gagika:
> What's the `sharding_tolerance` in your cmd for the test?

# sharding tolerance: float between 0.0 and 1.0 representing the allowed percentage of non-sharded parameters.

As I tested on v5p-8, I had to reduce number of layers (to 5) to fit the model, and needed to increase sharding_tolerance, it's not needed for full scale model.

[assert_params_sufficiently_sharded](https://github.com/AI-Hypercomputer/maxtext/blob/bea1cef3da74a3e67b52eadb3274c237331a314f/MaxText/train.py#L729)
 # The vocab tensor(s) of shape [vocab, embed] (and transpose) are not sharded by stage
 maxtext_utils.assert_params_sufficiently_sharded(state.params, mesh, config.sharding_tolerance)

MaxText/configs/models/deepseek3-671b.yml
RissyRan: I think those are default values. For instance, it has `vocab_size: int = 102400`, while the DeepSeek's vocab size is 129280. Yeah, let's remove this config from model config.

RissyRan:
> > What's the `sharding_tolerance` in your cmd for the test?
> 
> # sharding tolerance: float between 0.0 and 1.0 representing the allowed percentage of non-sharded parameters.
> As I tested on v5p-8, I had to reduce number of layers (to 5) to fit the model, and needed to increase sharding_tolerance, it's not needed for full scale model.
> 
> [assert_params_sufficiently_sharded](https://github.com/AI-Hypercomputer/maxtext/blob/bea1cef3da74a3e67b52eadb3274c237331a314f/MaxText/train.py#L729)
> 
> # The vocab tensor(s) of shape [vocab, embed] (and transpose) are not sharded by stage
> maxtext_utils.assert_params_sufficiently_sharded(state.params, mesh, config.sharding_tolerance)

I see. Makes sense.

RissyRan (approved):

PR #1298, opened 2025-02-22
LLama 405B Tweaks

anfals:
Just a few changes I made while converting and experimenting with Lllama 3.1 405B checkpoint

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -67,7 +67,7 @@ def create_orbax_checkpoint_manager(
   p.mkdir(exist_ok=True, parents=True)
   # we need to use ocdbt and zarr3 to control max file size in the checkpoint
   # omitting `iter` uses default handler for `iter`
-  item_handlers = {"items": PyTreeCheckpointHandler(use_ocdbt=use_ocdbt, use_zarr3=use_zarr3)}
+  item_handlers = {"items": PyTreeCheckpointHandler(save_concurrent_gb=500, use_ocdbt=use_ocdbt, use_zarr3=use_zarr3)}
   mngr = CheckpointManager(
       p,
       item_names=item_names,
@@ -221,6 +221,7 @@ def load_state_if_possible(
             single_replica_sharding=single_replica_sharding,
             global_shape=data.shape,
             dtype=data.dtype,
+            strict=False
         )
 
       if enable_single_replica_ckpt_restoring:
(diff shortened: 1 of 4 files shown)


MaxText/checkpointing.py
```diff
@@ -320,12 +321,20 @@ def load_params_from_path(load_parameters_from_path, abstract_unboxed_params):
   assert load_parameters_from_path, "load_parameters_from_path is not defined."
   max_logging.log(f"restoring params from {load_parameters_from_path}")
   ckpt = epath.Path(load_parameters_from_path)
-  ckptr = ocp.PyTreeCheckpointer()
+  # ckptr = ocp.Checkpointer(ocp.PyTreeCheckpointHandler(restore_concurrent_gb=500, save_concurrent_gb=500))
```
anfals, line 324: You might need something like this if you see an error like `ValueError: Requested more bytes than we reserved space for: 109924319232 > 96000000000`. You need to increase the restore limit. 

The sharded ckpt I linked to in our internal docs should resolve the need for this though

MaxText/checkpointing.py
```diff
@@ -320,12 +321,20 @@ def load_params_from_path(load_parameters_from_path, abstract_unboxed_params):
...
+  ckptr = ocp.Checkpointer(ocp.PyTreeCheckpointHandler())
   # This is a memory optimization. We don't want to restore the entire checkpoint - only the params.
   # Rather than pass the entire abstract state, which could unnecessarily restore opt_state and such and waste
   # memory, we instead specify here that we are just restoring the params field of the checkpoint
   # (which itself may be a dictionary containing a key named 'params').
   restore_args = ocp.checkpoint_utils.construct_restore_args(abstract_unboxed_params)
+  def update_restore_args(restore_args):
```
anfals, line 331: This is Orbax's way of allowing truncation of the parameters when restoring a ckpt. This is needed as NV used only a 32K vocab size, as opposed to the 128K. The truncation that Orbax does when strict mode is disabled is like A[:32000, :]

I made the entire PyTree not strict but you could probably just do the tokenizer embedding layer

MaxText/configs/base.yml
```diff
@@ -388,10 +388,10 @@ skip_jax_distributed_system: False # If True we will not initialize the jax dist
 # 2) Cosine decay from [learning_rate] to [learning_rate * cosine_learning_rate_final_fraction] from warmup to learning_rate_schedule_steps
 # 3) Constant learning rate of 0 from learning_rate_schedule_steps to steps.
 # The zero learning rate section can be used to more accurately measure the fully trained model's performance.
-learning_rate: 3.e-5
+learning_rate: 8.e-5
 cosine_learning_rate_final_fraction: 0.1
-warmup_steps_fraction: 0.1
-learning_rate_schedule_steps: -1 # By default the length of the schedule is set to the number of steps.
+warmup_steps_fraction: 0.0067
```
anfals, line 393: This is just jibberish.  We should compute this dynamically in pyconfig

MaxText/train.py
```diff
@@ -203,7 +203,7 @@ def save_checkpoint(
 ) -> bool:
   """Wrapper for saving checkpoint."""
   if config and config.enable_checkpointing:
-    if (step % config.checkpoint_period == 0) or (
+    if (step % config.checkpoint_period == 0 and step != 0) or (
```
anfals, line 206: We should make sure NOT to save a checkpoint during our real runs. GPT3 had some hackery in its ckpt to start at a non-zero ckpt. 

Also set the `checkpoint_period` to like 10000 or something huge as well

PR #1304, opened 2025-02-24
Add flops calculation for DeepSeek

RissyRan:
# Description

Update flops calculation to include MLA and DeepSeek structure
* Add a few helper functions
* Tested locally with 6BP rule

# Tests

Using test like below

```
  @pytest.mark.tpu_only
  def test_flops(self):

    import maxtext_utils
    total_tflops, learnable_weight_tflops, attention_tflops = maxtext_utils.calculate_tflops_training_per_device(self.cfg)
    print("after change....")
    print(f"total_tflops: {total_tflops}")
    print(f"learnable_weight_tflops: {learnable_weight_tflops}")
    print(f"attention_tflops: {attention_tflops}")
    total_tflops2, learnable_weight_tflops2, attention_tflops2 = maxtext_utils.calculate_tflops_training_original(self.cfg)
    print("before change....")
    print(f"total_tflops2: {total_tflops2}")
    print(f"learnable_weight_tflops2: {learnable_weight_tflops2}")
    print(f"attention_tflops2: {attention_tflops2}")
```

`per_device_batch_size=4 max_target_length=256`

| Model | Before | After |
| ------------- | ------------- | ------------- |
| llama2-7b | total_tflops: 41.00620025856, learnable_weight_tflops: 40.593883398144, attention_tflops: 0.412316860416 | total_tflops2: 41.00620025856, learnable_weight_tflops2: 40.593883398144, attention_tflops2: 0.412316860416 |
| gemma2-2b | total_tflops2: 15.515032485888, learnable_weight_tflops2: 15.399068368896, attention_tflops2: 0.115964116992 | total_tflops: 15.515032485888, learnable_weight_tflops: 15.399068368896, attention_tflops: 0.115964116992 | 
| mixtral 8x7b | total_tflops: 78.739635437568, learnable_weight_tflops: 78.327318577152, attention_tflops: 0.412316860416 | total_tflops2: 78.739635437568, learnable_weight_tflops2: 78.327318577152, attention_tflops2: 0.412316860416 | 
| deepseek2-16b | NA | total_tflops: 15.278272413696, learnable_weight_tflops: 15.060839694336, attention_tflops: 0.21743271936 | 
| deepseek3-671b | NA | total_tflops: 228.951418994688, learnable_weight_tflops: 225.021523918848, attention_tflops: 3.92989507584 | 

Comparing to 6BP rule:

```
1) llama2-7b, reported total 41 tflops, 6BP = 6 * 4 * 256 * 7 /10^3 = 43
2) DeepSeek v2-16b: reported total 15 tflops, 6BP = 6 * 4 * 256 * 2.4/10^3 = 14.7
3) DeepSeek v3-671b: reported total 229 tflops, 6BP = 6 * 4* 256 * 37/10^3 = 227
```

# Checklist

### DIFF
--- a/MaxText/maxtext_utils.py
+++ b/MaxText/maxtext_utils.py
@@ -124,51 +124,121 @@ def calculate_gemma2_tflops_training_per_device(config, total_ffn_flops, qkv_flo
   return attention_tflops, learnable_weight_tflops
 
 
+def calculate_mla_tflops_per_device(config):
+  """Calculate Multi-Head Latent Attention TFLOP"""
+  batch_len = config.per_device_batch_size * config.max_target_length
+  qk_head_dim_sum = config.qk_nope_head_dim + config.qk_rope_head_dim
+  # calculate mla query projection
+  if config.q_lora_rank == 0:
+    q_flops = 2 * batch_len * config.emb_dim * config.num_query_heads * qk_head_dim_sum
+  else:
+    # calculate query down and up flops
+    q_flops = (
+        2 * batch_len * (config.emb_dim * config.q_lora_rank + config.q_lora_rank * config.num_query_heads * qk_head_dim_sum)
+    )
+  # calculate mla kv projection with down and up flops
+  kv_flops = (
+      2
+      * batch_len
+      * (
+          config.emb_dim * (config.kv_lora_rank + config.qk_rope_head_dim)
+          + config.kv_lora_rank * config.num_query_heads * (config.qk_nope_head_dim + config.v_head_dim)
+      )
+  )
+  qkv_flops = q_flops + kv_flops
+
+  attention_flops = 2 * batch_len * config.max_target_length * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
+  projection_flops = 2 * batch_len * config.emb_dim * config.num_query_heads * config.v_head_dim
+  return qkv_flops, attention_flops, projection_flops
+
+
+def calculate_ffn_mamtul_tflops_per_device(config, mlp_dim):
+  """Helper function to calculate matmul TFLOP in ffn based on MLP dimension.
+
+  Applies to:
+    - Dense FFN layers (mlp_dim = config.mlp_dim).
+    - MoE FFN layers (mlp_dim = config.moe_mlp_dim),
+      need to scale by shared_experts or num_experts_per_tok.
+  """
+  ffn1_flops = (
+      2 * config.per_device_batch_size * config.max_target_length * mlp_dim * config.emb_dim * len(config.mlp_activations)
+  )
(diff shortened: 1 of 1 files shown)


MaxText/maxtext_utils.py
```diff
@@ -124,51 +124,95 @@ def calculate_gemma2_tflops_training_per_device(config, total_ffn_flops, qkv_flo
...
-    total_ffn_flops = gate_flops + config.num_experts_per_tok * total_ffn_flops
 
-  qkv_flops = (
-      2
-      * config.per_device_batch_size
-      * config.max_target_length
-      * config.emb_dim
-      * (config.num_query_heads + 2 * config.num_kv_heads)
-      * config.head_dim
+def calculate_ffn_tflops_per_device(config, mlp_dim):
```
gobbleturk, line 152: I understand the desire to define this function since its sort of shared between deepseek's dense and moe layers, but it doesn't calculate either one - the former has to be scaled by shared experts and the latter to experts_per_token - so this method does not do what its name suggests it does?

Perhaps you could instead just have two separate functions for the dense versus sparse ffn layers?

gobbleturk (approved):
The calculations are correct but it may not be clear to future users how to use the new methods are are adding, this might be hard to maintain. See comment I gave

MaxText/maxtext_utils.py
gobbleturk: oh okay I was indeed confused by this. I think future code writers might easily forget to add the extra `experts_per_token` factor when they call this in the future, but at least there is this example which shows how to do it =)

RissyRan: Yeah, it will depend on what's the calculation. You could see `calculate_base_ffn_mamtul_tflops_per_device` is used in 4 places (including deepseek and mixtral style). It could be:

1) multiplied by config.first_num_dense_layers to get dense layer flops
2) mulitplied by config.num_experts_per_tok to get all routed experts flops
3) mulitplied by config.shared_experts to get shared experts flops

RissyRan: Sounds good. I updated it to something like below:

```
def calculate_ffn_mamtul_tflops_per_device(config, mlp_dim):
  """Helper function to calculate matmul TFLOP in ffn based on MLP dimension.

  Applies to:
    - Dense FFN layers (mlp_dim = config.mlp_dim).
    - MoE FFN layers (mlp_dim = config.moe_mlp_dim),
      need to scale by shared_experts or num_experts_per_tok.
  """
```

After a second thought, I update dense `total_ffn_flops` with this helper function to avoid duplicate code blocks. We just need to pass mlp_dim as config.mlp_dim. Let me know if this is confusing :)

gagika (approved):
Thanks Ran

PR #1315, opened 2025-02-26
Refactor: Move metrics logic to MetricLogger class

SurbhiJainUSC:
# Description

This PR initiates the refactoring of `train.py`. As part of refactoring effort, we will be:
- Consolidating all the metrics-related code into a dedicated MetricLogger class ([PR](https://github.com/AI-Hypercomputer/maxtext/pull/1315))
- Moving some methods in `train.py` to a util module that can be re-used by other trainers (To be followed)

# Tests
Ran MaxText/train.py with `gcs_metrics=True` and metrics_file="/tmp/maxtext_metrics.txt" configurations to ensure that the metrics were stored in a local file, GCS and TensorBoard.

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -48,6 +48,8 @@ import pyconfig
 import pathwaysutils  # pylint: disable=unused-import
 import tensorflow as tf
 
+from metric_logger import MetricLogger
+
 from vertex_tensorboard import VertexTensorboardManager
 # Placeholder: internal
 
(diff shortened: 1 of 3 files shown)


MaxText/train.py
```diff
@@ -974,10 +900,9 @@ def train_loop(config, state=None):
 
   if checkpoint_manager is not None:
     checkpoint_manager.wait_until_finished()
-  write_metrics(writer, local_metrics_file, running_gcs_metrics, metrics, config.steps - 1, config)  # final step metrics
+  metric_logger.write_metrics(running_gcs_metrics, metrics, config.steps - 1)  # final step metrics
   max_utils.close_summary_writer(writer)
   record_goodput(recorder, config, recorder.record_job_end_time if recorder else None)
-  clear_buffered_metrics()
```
bvandermoon, line 980: Is this clear no longer needed?

SurbhiJainUSC: This method is basically setting the global variables to None. I have moved those global variables to class variables instead. AFAIK, python's garbage collector will automatically clear the memory occupied by that class object when program exits. So, this method shouldn't be required.
@gobbleturk - LMK if I am missing anything here.

gobbleturk: I have seen issues when calling the main() of train.py mutiple times within the same python invocation, e.g. if you a script like
```
  run_two_train()
  from train.py import main as train_main
  train_main()
  train_main()
```
then the second call of train_main() will inherit the global variables (the metrics) of the end of the first unless they are explicitly cleared. I don't have much understanding of python garbage collection though. If we refactor train well enough perhaps this clear method is not needed.

gobbleturk (approved):
LGTM, this is awesome thanks Surbhi!

I do think we still need the clear method as described in the comment though - but if you  can run a test with two train_main calls in the same python script and see that the second run has fresh metrics perhaps it can be removed

MaxText/train.py
SurbhiJainUSC: In this case, each invocation of train_main() will have its own metric logger object. All the metrics related data can only be accessed by the class object. So, IMO the second call of train_main() will not be able to access the metrics from the first train_main() call.

bvandermoon (approved):
LGTM pending the outcome of the clear method discussion

SurbhiJainUSC:
> LGTM pending the outcome of the clear method discussion

Verified that two sequential `train_main()` function calls within the same script produced different sets of metrics.

PR #1318, opened 2025-02-27
fixes to conversion script and llama3.1-405b checkpointing

vivianrwu:
# Description

This PR has the following changes:

**Fixes bug where pytorch checkpoints conversion to training checkpoint will shard. The jax_weights in the pytorch checkpoint are currently not returned. The scanning checkpoint completes but when unscanning, we encounter this error:**

```
  File "/maxtext/MaxText/generate_param_only_checkpoint.py", line 150, in <module>
    app.run(main)
  File "/usr/local/lib/python3.10/dist-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.10/dist-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "/maxtext/MaxText/generate_param_only_checkpoint.py", line 146, in main
    generate_decode_checkpoint(config)
  File "/maxtext/MaxText/generate_param_only_checkpoint.py", line 134, in generate_decode_checkpoint
    _possibly_unroll_params(config, training_state, training_state_annotations, mesh)
  File "/maxtext/MaxText/generate_param_only_checkpoint.py", line 68, in _possibly_unroll_params
    new_layer = jax.jit(slice_ith, out_shardings=new_per_layer_state_sharding)(training_state_layers)
TypeError: Argument 'input_layers['mlp']['wi_0']['kernel']' of shape bfloat16[4096,32,14336] of type <class 'jax._src.api.ShapeDtypeStruct'> is not a valid JAX type.
``` 

After returning the jax_weights, we do not encounter this error anymore when unscanning. 

**Additionally fixes Llama3.1-405b checkpoint conversion for decode checkpoint / unscanning.**
Since the model is very large, the default bytes supported for restoring and saving is not enough, leading to errors such as

```
packages/orbax/checkpoint/_src/serialization/serialization.py", line 130, in wait_for_bytes
    raise ValueError('Requested more bytes than we reserved space for: '
ValueError: Requested more bytes than we reserved space for: 439697276928 > 96000000000
```

We bump up the bytes for larger models such as the llama3.1-405b. 

**Fixes llama3.1-405b kv weight replication within every two files.**

FIXES: b/397970646

# Tests

Tested with an e2e job that downloads checkpoint from Meta, converts to scanned checkpoint, and unscans. Further includes logic for quantization too. 

# Checklist

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -173,6 +173,7 @@ def load_state_if_possible(
     data_iterator: Union[MultiHostDataLoadIterator, None],
     load_parameters_from_path: str,
     load_full_state_from_path: str,
+    checkpoint_storage_concurrent_gb: int,
     abstract_unboxed_pre_state: train_state.TrainState,
     enable_single_replica_ckpt_restoring: Optional[bool] = False,
     dataset_type: Optional[str] = "tfds",
@@ -191,6 +192,7 @@ def load_state_if_possible(
       matches type against.
     enable_single_replica_ckpt_restoring: bool flag for restoring checkpoitn
       with SingleReplicaArrayHandler
+    checkpoint_storage_concurrent_gb: concurrent GB for checkpoint byte I/O.
 
   Returns:
     A tuple of (train_state, train_state_params) where full_train_state captures
@@ -282,7 +284,9 @@ def load_state_if_possible(
         )
 
   if load_parameters_from_path != "":
-    restored_params = load_params_from_path(load_parameters_from_path, abstract_unboxed_pre_state.params)
+    restored_params = load_params_from_path(
+        load_parameters_from_path, abstract_unboxed_pre_state.params, checkpoint_storage_concurrent_gb
+    )
     return None, restored_params
   elif load_full_state_from_path != "":
     max_logging.log(f"restoring full state from {load_full_state_from_path=}")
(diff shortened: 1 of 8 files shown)


MaxText/llama_or_mistral_ckpt.py
vipannalla, line 1: @wang2yn84 , @RissyRan, please take a look since this script was recently updated for DeepSeek and you are most familiar now.

vivianrwu: Thanks, the main update is accommodating for Llama3.1-405b conversion where similar change was made here: https://github.com/AI-Hypercomputer/maxtext/blob/experimental_proxy/MaxText/llama_ckpt_conversion_inference_only.py#L174-L175

The fix for sharding is because looks like jax_weights is not being returned, therefore sharding never happens (this code source: https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/llama_or_mistral_ckpt.py#L759-L778)

RissyRan: Thanks! I will take a look today.

MaxText/checkpointing.py
```diff
@@ -315,12 +317,21 @@ def setup_checkpoint_logger(config) -> cloud_logger.CloudLogger | None:
...
-def load_params_from_path(load_parameters_from_path, abstract_unboxed_params):
+def load_params_from_path(load_parameters_from_path, abstract_unboxed_params, model_name):
   """Load decode params from checkpoint at specified path."""
   assert load_parameters_from_path, "load_parameters_from_path is not defined."
   max_logging.log(f"restoring params from {load_parameters_from_path}")
   ckpt = epath.Path(load_parameters_from_path)
-  ckptr = ocp.PyTreeCheckpointer()
+
+  # The model name will determine if we need to set the concurrent_gb
+  concurrent_gb = 500 if model_name == "llama3.1-405b" else 96
```
RissyRan, line 327: It seems a hardcode value here based on a specific model. Will it be better to add a config for this, such as `concurrent_gb` like other ckpt related config: https://github.com/AI-Hypercomputer/maxtext/blob/4797997fcd08ac287f26c099ede9ae6c2bce77e4/MaxText/configs/base.yml#L54 ? Is 96 as default for the API?

RissyRan (commented):
Thanks for fixing that return! Just have one comment if we should pass a specific config, or model name. I would prefer a new config as values of `500` and `96` are hardcoded.

MaxText/checkpointing.py
vivianrwu: Thanks! Yes, 96 is the default for the API. Added `checkpoint_storage_concurrent_gb` to the config to follow the naming conventions.

RissyRan (approved):
Thanks!

vipannalla (approved):
LGTM too

PR #1325, opened 2025-02-28
Add tokenizer_type tiktoken for llama3.1

A9isha:
# Description

Llama3.1 implementation in MaxText needs `tokenizer_type: "tiktoken"`, so adding that

FIXES: b/399771671

# Tests

Tested by running the code on a devbox as well as on a cluster via XPK

# Checklist


richjames0 (approved):
Maybe there should be a base llama config that includes things common like the tokenizer - your call

gobbleturk (changes requested):
The tokenizer type is not an inherent property of the model, it should be possible to run this model with different datasets and different tokenizers (which can be trained with different weights). Is there a different place you can make this change?

gobbleturk:
This PR was actually merged in a copybara squash https://github.com/AI-Hypercomputer/maxtext/commit/91cb00b01b85f6ac592684026c0a7e9f86379f54

gobbleturk:
This is roughly reverted though in https://github.com/AI-Hypercomputer/maxtext/pull/1513

PR #1327, opened 2025-02-28
Supporting LoRA in MaxText for inference only, to support multi-LoRA adapters inferencing via JetStream.

aman2930:
# Description
Supporting LoRA in MaxText for inference only, to support multi-LoRA adapters inferencing via JetStream.

# Approach and changes details
- Added some LoRA specific config variables inside base.yml.
- Implemented naive way of merging LoRA weights to Base weights, which can be used only for inference purposes.
- Implemented methods to create the abstract_state of the LoRA params using the base_mode_abstract_state and the lora_adapter_config.
- The abstract_state of the LoRA params is used to Read the HuggingFace format of the LoRA checkpoints. It is further used to convert, validate & write the LoRA params in Jax format, which is used during the inference. 
- Implemented couple of util functions in max_utils which can be reused.
- The whole workflow of LoRA conversion of weights is guarded by the config flag `lora_adapters_path`.
- For reviewers, max_utils.py and maxengine.py doesn't have any significant change in existing methods. Only new methods were added which can be triggered in LoRA workflows. So ideally, non-LoRA workflows should not have any functionality impact.

# What next:
- In coming future there could be more core changes in MaxText to support multiple LoRA adapters in same batch of decoding.

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -31,6 +31,10 @@ normalization_layer_epsilon: 1.e-05
 # Loads a just parameters from a specific directory
 # e.g. gs://my-base-output-directory/my-previous-run-name/checkpoints/items/NUMBER or NUMBER/items
 load_parameters_path: ""
+
+# LoRA adapter support configs
+lora_input_adapters_path: ""    # Input GCS path for a parent directory which has all the LoRA adapters (lora_id as subdir)
+
 # Loads a full checkpoint including optimizer state and step count from a specific directory
 # e.g. gs://my-base-output-directory/my-previous-run-name/checkpoints/items/NUMBER or NUMBER/items
 load_full_state_path: ""
@@ -547,6 +551,7 @@ inference_metadata_file: "" # path to a json file
 inference_server: "MaxtextInterleavedServer"  # inference server to start
 inference_benchmark_test: False
 enable_model_warmup: False
+enable_llm_inference_pool: False          # Bool to launch inference server for llm_inference_gateway with their specified APIs
 multi_sampling: False
 
 # Stack prefill cache across the layer to reduce the
--- a/MaxText/generate_param_only_checkpoint.py
+++ b/MaxText/generate_param_only_checkpoint.py
@@ -37,6 +37,8 @@ from jax import random
 from typing import Sequence
 from layers import models, quantizations
 from train import save_checkpoint
+from utils import gcs_utils
+from utils import lora_utils
 
 Transformer = models.Transformer
 
@@ -99,6 +101,43 @@ def _read_train_checkpoint(config, checkpoint_manager, mesh):
   return state, state_mesh_notations
 
 
+def _generate_lora_decode_checkpoints(config, mesh):
+  """Read lora checkpoints checkpoint at path defined by load_full_state_path."""
+  # Model and Optimizer definition
+  quant = quantizations.configure_quantization(config)
+  model = Transformer(config, mesh, quant)
+  rng = random.PRNGKey(0)
+  learning_rate_schedule = max_utils.create_learning_rate_schedule(config)
+  tx = optimizers.get_optimizer(config, learning_rate_schedule)
+
+  lora_adapters = gcs_utils.gcs_list_directories(config.lora_input_adapters_path)
+  for lora_id in lora_adapters:
+    # Expected lora_checkpoint_dir = <checkpoint_dir>/loras/<lora_id>
+    lora_checkpoint_dir = f"{config.checkpoint_dir}loras/{lora_id}/"
+
+    lora_adapter_path = f"{config.lora_input_adapters_path}/{lora_id}/"
+
+    # Create a checkpoint manager to save decode checkpoint at lora_checkpoint_dir
+    checkpoint_manager = checkpointing.create_orbax_checkpoint_manager(
+        lora_checkpoint_dir,
+        config.enable_checkpointing,
+        config.async_checkpointing,
+        config.checkpoint_period,
+    )
+
+    lora_config, lora_state, lora_state_annotations = lora_utils.setup_initial_lora_state(
+        model, None, tx, config, rng, mesh, checkpoint_manager, lora_adapter_path
+    )
+
+    _possibly_unroll_params(config, lora_state, lora_state_annotations, mesh)
+
+    gcs_utils.write_dict_to_gcs_json(lora_config, lora_checkpoint_dir + "adapter_config.json")
+
+    # Save decode state to config's checkpoint directory at step 0
+    _save_decode_checkpoint(config, lora_state, checkpoint_manager)
+    max_logging.log(f"Successfully saved LoRA checkpoint at: {lora_checkpoint_dir}0/items")
+
+
 def _save_decode_checkpoint(config, state, checkpoint_manager):
   """Generate checkpoint for decode from the training_state."""
   with jax.spmd_mode("allow_all"):
@@ -129,8 +168,13 @@ def generate_decode_checkpoint(config):
       path.rmtree()
 
   # Create a checkpoint manager to save decode checkpoint at config.checkpoint_dir
+  base_checkpoint_dir = config.checkpoint_dir
+
+  if config.lora_input_adapters_path:
+    base_checkpoint_dir += "base/"
+
   checkpoint_manager = checkpointing.create_orbax_checkpoint_manager(
-      config.checkpoint_dir,
+      base_checkpoint_dir,
       config.enable_checkpointing,
       config.async_checkpointing,
       config.checkpoint_period,
@@ -143,9 +187,13 @@ def generate_decode_checkpoint(config):
   _possibly_unroll_params(config, training_state, training_state_annotations, mesh)
 
   # Save decode state to config's checkpoint directory at step 0
-  max_logging.log(f"Save decode checkpoint at: {config.checkpoint_dir}")
+  max_logging.log(f"Save decode checkpoint at: {base_checkpoint_dir}")
   _save_decode_checkpoint(config, training_state, checkpoint_manager)
-  max_logging.log(f"Successfully generated decode checkpoint at: {config.checkpoint_dir}0/items")
+  max_logging.log(f"Successfully generated decode checkpoint at: {base_checkpoint_dir}0/items")
+
+  if config.lora_input_adapters_path:
+    _generate_lora_decode_checkpoints(config, mesh)
+
   return True
 
 
--- a/MaxText/llama_or_mistral_ckpt.py
+++ b/MaxText/llama_or_mistral_ckpt.py
@@ -36,6 +36,7 @@ import os
 import gc
 import re
 import logging
+import json
 from dataclasses import dataclass
 
 os.environ["JAX_PLATFORMS"] = "cpu"
@@ -49,10 +50,11 @@ import psutil
 from tqdm import tqdm
 
 import max_logging
+import max_utils
 from train import save_checkpoint
 import checkpointing
 from safetensors import safe_open
-import max_utils
+from utils import gcs_utils
 
 MODEL_PARAMS_DICT = {
     "llama2-70b": {
@@ -175,6 +177,15 @@ def _hf_mapping(layer_idx: int = -1, expert_idx: int = -1) -> dict:
       f"layers.{layer_idx}.feed_forward.w1.weight": f"model.layers.{layer_idx}.mlp.gate_proj.weight",
       f"layers.{layer_idx}.feed_forward.w2.weight": f"model.layers.{layer_idx}.mlp.down_proj.weight",
       f"layers.{layer_idx}.feed_forward.w3.weight": f"model.layers.{layer_idx}.mlp.up_proj.weight",
+      # LoRA Adapter
+      f"layers.{layer_idx}.attention.wq.lora_A.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.q_proj.lora_A.weight",
+      f"layers.{layer_idx}.attention.wq.lora_B.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.q_proj.lora_B.weight",
+      f"layers.{layer_idx}.attention.wk.lora_A.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.k_proj.lora_A.weight",
+      f"layers.{layer_idx}.attention.wk.lora_B.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.k_proj.lora_B.weight",
+      f"layers.{layer_idx}.attention.wv.lora_A.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.v_proj.lora_A.weight",
+      f"layers.{layer_idx}.attention.wv.lora_B.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.v_proj.lora_B.weight",
+      f"layers.{layer_idx}.attention.wo.lora_A.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.o_proj.lora_A.weight",
+      f"layers.{layer_idx}.attention.wo.lora_B.weights": f"base_model.model.model.layers.{layer_idx}.self_attn.o_proj.lora_B.weight",
   }
 
 
@@ -190,6 +201,7 @@ def _hf_to_maxtext_mapping(layer_idx: int = -1, expert_idx: int = -1) -> dict:
       f"model.layers.{layer_idx}.self_attn.k_proj.weight": f"layers.{layer_idx}.attention.wk.weight",
       f"model.layers.{layer_idx}.self_attn.v_proj.weight": f"layers.{layer_idx}.attention.wv.weight",
       f"model.layers.{layer_idx}.self_attn.o_proj.weight": f"layers.{layer_idx}.attention.wo.weight",
+      f"model.layers.{layer_idx}.self_attn.rotary_emb.inv_freq": f"layers.{layer_idx}.attention.rotary_emb.inv_freq",
       # MOE model
       f"model.layers.{layer_idx}.block_sparse_moe.gate.weight": f"layers.{layer_idx}.feed_forward.gate.weight",
       f"model.layers.{layer_idx}.block_sparse_moe.experts.{expert_idx}.w1.weight": f"layers.{layer_idx}.feed_forward.experts.{expert_idx}.w1.weight",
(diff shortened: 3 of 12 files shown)


MaxText/generate_param_only_checkpoint.py
```diff
@@ -90,6 +90,40 @@ def _read_train_checkpoint(config, checkpoint_manager, mesh):
...
+  # Model and Optimizer definition
+  quant = quantizations.configure_quantization(config)
+  model = Transformer(config, mesh, quant)
+  rng = random.PRNGKey(0)
+  learning_rate_schedule = max_utils.create_learning_rate_schedule(config)
+  tx = optimizers.get_optimizer(config, learning_rate_schedule)
+
+  lora_adapters = max_utils.gcs_list_directories(config.lora_adapters_path)
+  for lora_id in lora_adapters:
+    lora_checkpoint_dir = config.checkpoint_dir + config.lora_dir + lora_id + "/"
```
richjames0, line 104: nit: this would be cleaner with an f-string

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -228,6 +239,164 @@ def permute_to_match_maxtext_rope(arr):
   return np.concatenate((evens, odds), axis=arr.ndim - 1)
 
 
+def convert_lora_weights_to_jax_weights(lora_config, model_size):
```
richjames0, line 242: nit: this is a very long function. of course you're following the existing "style" but i would love it if someone took the initiative to improve things :D for a willing party, the for loop could be modularized and some of the structures could be pulled out to constants or something?

MaxText/max_utils.py
```diff
@@ -200,6 +202,104 @@ def upload_dump(local_dir, target_dir, module_name=None, delete_local_after=True
     shutil.rmtree(local_dir)
 
 
+def gcs_path_exists(file_path):
```
richjames0, line 205: thanks for adding these!

MaxText/generate_param_only_checkpoint.py
aman2930: Done.

MaxText/llama_or_mistral_ckpt.py
aman2930: I totally agree that it need to be refactored and modularized. Instead there should be a modularized library to convert any format to MaxText and vice-versa. But unfortunately it will not fit in my bandwidth/timelines.

vipannalla:
Reviewed at very high level and had some code structuring comments. Will take a deeper look soon.

MaxText/max_utils.py
aman2930: Done.

MaxText/maxengine.py
```diff
@@ -224,9 +228,35 @@ def load_params(self, *args, rng: Optional[PRNGKeyType] = None, **kwargs) -> Par
...
     return params
 
+  def load_single_adapter(self, adapter_path):
+    """
+    Load Single adapter from adapter_path.
+    Expect adapter_config.json and LoRA adapter weights at this path within subdirectory `/0/items`.
+    """
+
+    adapter_config_path = adapter_path + "/adapter_config.json"
+    adapter_weights_path = adapter_path + "/0/items"
```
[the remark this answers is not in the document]
aman2930: `os.path.join()` doesn't work for paths starting with `gs://`. Used f-string instead to make the code more readable.

> Do these names apply across any adaptor downloaded from HF or other places?
- For `adapter_config.json`, Yes, it was consistent for as many as I saw. 
- For `/0/items`, this is the path our scripts generate for saving the checkpoints in JAX format. So I would say it is also consistent for MaxText.

MaxText/max_utils.py
```diff
@@ -980,6 +1207,152 @@ def get_abstract_state(model, tx, config, rng, mesh, is_training=True):
   )
 
 
+def get_lora_abstract_state(base_abstract_params, lora_config):
```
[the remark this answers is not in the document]
aman2930: Done

MaxText/generate_param_only_checkpoint.py
```diff
@@ -90,6 +90,41 @@ def _read_train_checkpoint(config, checkpoint_manager, mesh):
   return state, state_mesh_notations
 
 
+def _read_load_save_lora_checkpoints(config, mesh):
```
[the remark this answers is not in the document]
aman2930: Updated it to more intuitive name. 
Actually this function was reading the training_specific (scanned checkpoints), then loading it based on the abstract_state of the model (to check any discrepancy) and then saving it in the decode format(unscanned checkpoints).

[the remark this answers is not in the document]
aman2930: Done.

MaxText/max_utils.py
```diff
@@ -200,6 +203,104 @@ def upload_dump(local_dir, target_dir, module_name=None, delete_local_after=True
     shutil.rmtree(local_dir)
 
 
+def gcs_path_exists(file_path):
+  """Checks if a GCS file_path exits."""
+  try:
+    storage_client = storage.Client()
+
```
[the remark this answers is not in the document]
aman2930: Done in the refactored code.

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -801,18 +988,78 @@ def checkpoint_device_put(arr):
...
+    if args.lora_adapters_path.startswith("gs://"):
+      max_logging.log("GCS Source path for the LoRA adapters is not supported as of now.")
+      raise NotImplementedError
+
+    lora_ids = list_folders_pathlib(args.lora_adapters_path)
+
+    for lora_id in lora_ids:
+      lora_path = f"{args.lora_adapters_path}/{lora_id}"
+      lora_config_path = f"{lora_path}/adapter_config.json"
+      with open(lora_config_path, "r", encoding="utf8") as file:
```
vipannalla, line 1049: Right now code  will fail if the config doesn't exist. Can we check for existence of the adaptor_config.json and add a log like `max.logging(f"ignoring {lora_id} adaptor because adaptor_config.json ...")` doesn't exist?

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -801,18 +988,78 @@ def checkpoint_device_put(arr):
...
 
   if args.model_size not in MODEL_PARAMS_DICT:
     raise NotImplementedError
 
   os.environ["XLA_FLAGS"] = f"--xla_force_host_platform_device_count={SIMULATED_CPU_DEVICES_COUNT}"
+
+  base_weights_path = args.maxtext_model_path
+
+  if args.lora_adapters_path:
+    base_weights_path += "/base"
```
vipannalla, line 1031: Changing the path will impact current checkpoint loading code. I'd keep it as-is and just add a subdirectory to store LoRA specific weights.

MaxText/utils/lora_utils.py
```diff
@@ -0,0 +1,301 @@
...
+from flax.training import train_state
+from flax.linen import partitioning as nn_partitioning
+
+import max_utils
+import max_logging
+from utils import gcs_utils
+
+
+def apply_lora_on_base_params(base_params, lora_params, lora_scale_factor=1.0):
+  """
```
vipannalla, line 33: Thanks for detailed docbook!

MaxText/maxengine_server.py
```diff
@@ -56,6 +56,7 @@ def main(config):
       enable_jax_profiler=config.enable_jax_profiler if config.enable_jax_profiler else False,
       jax_profiler_port=config.jax_profiler_port if config.jax_profiler_port else 9999,
       enable_model_warmup=config.enable_model_warmup if config.enable_model_warmup else False,
+      enable_llm_inference_pool=config.enable_llm_inference_pool if config.enable_llm_inference_pool else False,
```
vipannalla, line 59: Do you have the corresponding Jetstream PR I can look at?

MaxText/maxengine.py
```diff
@@ -224,9 +229,34 @@ def load_params(self, *args, rng: Optional[PRNGKeyType] = None, **kwargs) -> Par
       params = self.quantize_params(state, rng3)
     else:
       params = state.params
-    max_utils.print_mem_stats("After load_params")
+
+    self.print_stats("After load_params")
+
     return params
 
+  def load_single_adapter(self, adapter_path):
```
vipannalla, line 237: Where is this method used? I don't see the code in `load_params` calling this method for each lora adaptor. Is that in a followup PR?

vipannalla (commented):
Thanks for re-organizing the code, looks modular and easier to follow. I'd some more comments around checkpoint conversion and loading.

MaxText/maxengine.py
aman2930: MaxEngine's instance in the JetStream's Driver will call it during the load_adapter calls or during the request with new adapter.

MaxText/maxengine_server.py
aman2930: JetStream PR is in progress (Pending refactoring and cleanup). 
If you just want to look at the changes, you can check it out here: https://github.com/AI-Hypercomputer/JetStream/compare/main...amangu-lora

MaxText/llama_or_mistral_ckpt.py
aman2930: I don't think it will impact the current non-lora workflow. That is why the check of lora_adapters_path is added. It is required only in the case where we are converting base as well as the LoRA together to differentiate between the base model checkpoint paths and the LoRA adapter paths. LoRa specific weights are already getting saved in the subdirectory `LoRA`.

vipannalla: Okay, this may be ok. Note that just because we convert a checkpoint for base and LoRA together doesn't mean they have to be used together. The base checkpoint should be usable without LoRA enabled. Adding a end-2-end test will make it clearer to future users no how this script supposed to be used for LoRA.

aman2930: I think there is some confusion in understanding. 
> The base checkpoint should be usable without LoRA enabled.
I preferred this only. 

So the script can be used for base_checkpoints without LoRA enabled. And this workflow doesn't impact any of the checkpoint loading path or code. 

But LoRA_checkpoint conversion can not be done without base checkpoints, as that is required to get the base params structure etc. Later on we can think of removing that dependency but for now I don't think it should be the priority. When both LoRA and base checkpoints are getting converted, we need a clear way to separate those. So here is how they will get saved when LoRA conversion enabled:

- Base checkpoints: <checkpoint_dir>/base/...
- LoRA checkpoints: <checkpiont_dir>/LoRAs/<lora_id>/...

MaxText/configs/base.yml
```diff
@@ -540,6 +547,7 @@ inference_metadata_file: "" # path to a json file
 inference_server: "MaxtextInterleavedServer"  # inference server to start
 inference_benchmark_test: False
 enable_model_warmup: False
+enable_llm_inference_pool: False
```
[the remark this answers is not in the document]
aman2930: Done.
This is to run the JetStream server with APIs specific to LLM Inference Gateway requirements.

MaxText/generate_param_only_checkpoint.py
```diff
@@ -90,6 +91,41 @@ def _read_train_checkpoint(config, checkpoint_manager, mesh):
...
+
+    # Create a checkpoint manager to save decode checkpoint at lora_checkpoint_dir
+    checkpoint_manager = checkpointing.create_orbax_checkpoint_manager(
+        lora_checkpoint_dir,
+        config.enable_checkpointing,
+        config.async_checkpointing,
+        config.checkpoint_period,
+    )
+
+    lora_config, lora_state, lora_state_annotations = max_utils.setup_initial_lora_state(
```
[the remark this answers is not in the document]
aman2930: Fixed it. 
I will add the test in a followup PR.

MaxText/configs/base.yml
```diff
@@ -31,6 +31,13 @@ normalization_layer_epsilon: 1.e-05
 # Loads a just parameters from a specific directory
 # e.g. gs://my-base-output-directory/my-previous-run-name/checkpoints/items/NUMBER or NUMBER/items
 load_parameters_path: ""
+
+# LoRA adapter support configs
+lora_adapters_path: ""
+lora_dir: "LoRAs/"
+lora_config_file_name: "adapter_config.json"
```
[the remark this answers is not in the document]
aman2930: Done.
- `lora_input_adapters_path `: An input path, which has all the LoRA adapters in the directory named by the specific LoRA_ID.
- `lora_checkpoint_dir`: An output directory, which will save all the checkpoints converted LoRA adapters.

MaxText/configs/base.yml
```diff
@@ -31,6 +31,13 @@ normalization_layer_epsilon: 1.e-05
...
+base_dir: "base/"
```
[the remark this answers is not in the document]
aman2930: Added comment to make it more clear. Adding lora prefix will make it more confusing.

MaxText/llama_or_mistral_ckpt.py
aman2930: Done.

MaxText/maxengine.py
```diff
@@ -224,9 +229,34 @@ def load_params(self, *args, rng: Optional[PRNGKeyType] = None, **kwargs) -> Par
...
+
+    params, config = lora_utils.load_adapter(self.config, self.abstract_params, adapter_config_path, adapter_weights_path)
+
+    config["adapter_path"] = adapter_weights_path
+
+    self.print_stats("After load_single_adapter.")
+
+    return params, config
+
+  def apply_adapter(self, base_params, adapter_config, adapter_params):
```
mailvijayasingh, line 253: What dtype are these Lora params and how do they work with quantization?

MaxText/utils/lora_utils.py
```diff
@@ -0,0 +1,301 @@
...
+import jax.numpy as jnp
+from flax.training import train_state
+from flax.linen import partitioning as nn_partitioning
+
+import max_utils
+import max_logging
+from utils import gcs_utils
+
+
+def apply_lora_on_base_params(base_params, lora_params, lora_scale_factor=1.0):
```
mailvijayasingh, line 32: How do these work with QTensors that we have from AQT ?

MaxText/maxengine.py
aman2930: AFAIK, its should be same as that of the main base model params.

MaxText/configs/base.yml
```diff
@@ -31,6 +31,13 @@ normalization_layer_epsilon: 1.e-05
 # Loads a just parameters from a specific directory
 # e.g. gs://my-base-output-directory/my-previous-run-name/checkpoints/items/NUMBER or NUMBER/items
 load_parameters_path: ""
+
+# LoRA adapter support configs
+lora_input_adapters_path: ""    # Input GCS path for a parent directory which has all the LoRA adapters (lora_id as subdir)
+lora_checkpoint_dir: "LoRAs/"   # GCS directory for saving all the converted checkpoints of all the LoRA adapters. 
+lora_config_file_name: "adapter_config.json"    # Name of the LoRA config file expected in each LoRA directory
+base_dir: "base/"             # GCS directory for saving checkpoints of base model if exists along with LoRA checkpoints
```
vipannalla, line 39: If this config is only used for LoRA adaptor loading, we don't need this in the config (i.e, doesn't need to be configured). Please make this a constant in the module where its used.

aman2930: Done. 

`lora_input_adapters_path` is required because it was required during both the conversion scripts and the JetStream server. And JetStream doesn't have its own config.

vipannalla (approved):
LGTM

richjames0 (approved):
lgtm

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -230,6 +242,174 @@ def permute_to_match_maxtext_rope(arr):
...
+        lora_B_q = lora_chkpt_vars[f"layers.{layer_idx}.attention.wq.lora_B.weights"].type(torch.float16).numpy().transpose()
+
+        lora_B_q = np.reshape(lora_B_q, [lora_rank, base_num_query_heads, head_dim])
+
+        if self_attention_lora["query"]["lora_a.kernel"] is None:
+          self_attention_lora["query"]["lora_a.kernel"] = np.zeros(stack_shape + lora_A_q.shape, dtype=np.float16)
+          self_attention_lora["query"]["lora_b.kernel"] = np.zeros(stack_shape + lora_B_q.shape, dtype=np.float16)
+
+        self_attention_lora["query"]["lora_a.kernel"][layer_idx, ...] = lora_A_q  # pylint: disable=E1137
+        self_attention_lora["query"]["lora_b.kernel"][layer_idx, ...] = lora_B_q  # pylint: disable=E1137
```
bvandermoon, line 330: Is there any way to move this functionality to separate methods? I see they are slightly different, but there is a lot of repeated code several times in this file

MaxText/llama_or_mistral_ckpt.py
```diff
@@ -230,6 +242,174 @@ def permute_to_match_maxtext_rope(arr):
...
+        self_attention_lora["out"]["lora_a.kernel"][layer_idx, ...] = lora_A_o  # pylint: disable=E1137
+        self_attention_lora["out"]["lora_b.kernel"][layer_idx, ...] = lora_B_o  # pylint: disable=E1137# pylint: disable=E1137
+
+  if self_attention_lora["query"]["lora_a.kernel"] is not None:
+    self_attention_lora["query"]["lora_a.kernel"] = np.transpose(
+        self_attention_lora["query"]["lora_a.kernel"], axes=(1, 0, 2)
+    )
+    self_attention_lora["query"]["lora_b.kernel"] = np.transpose(
+        self_attention_lora["query"]["lora_b.kernel"], axes=(1, 0, 2, 3)
+    )
```
bvandermoon, line 379: Same with these. Is a helper method possible here?

bvandermoon (approved):
Generally LGTM. A couple comments on if we can separate these methods out.

And thanks Aman. I will reach out to you next week to discuss the training side a little more

MaxText/llama_or_mistral_ckpt.py
aman2930: Done.

MaxText/llama_or_mistral_ckpt.py
aman2930: I don't think this helps a lot. In fact it is making confusing because we need to be careful for the shapes. 
I would  prefer to delegate this refactor until this whole file is refactored.

PR #1355, opened 2025-03-06
support hf tokenizer in grain

aireenmei:
# Description
support hf tokenizer in grain

# Tests
Tested locally

# Checklist

### DIFF
--- a/MaxText/input_pipeline/_grain_data_processing.py
+++ b/MaxText/input_pipeline/_grain_data_processing.py
@@ -46,6 +46,8 @@ def preprocessing_pipeline(
     dataloading_host_index,
     dataloading_host_count,
     data_columns,
+    tokenizer_type: str = "sentencepiece",
+    hf_access_token=None,
     shuffle: bool = False,
     data_shuffle_seed=0,
     tokenize=True,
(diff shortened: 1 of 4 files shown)


khatwanimohit (approved):

MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -46,6 +46,8 @@ def preprocessing_pipeline(
     dataloading_host_index,
     dataloading_host_count,
     data_columns,
+    tokenizer_type: str = "sentencepiece",
+    hf_access_token=None,
```
RissyRan, line 50: nit: it seems the format is not aligned in those 2 lines

RissyRan (approved):

MaxText/input_pipeline/_grain_data_processing.py
aireenmei: Thanks for checking formats. Looks like for params with default value but without type hint (hf_access_token), there's no space around "=", but when there's both type hint and default value, we need space around "=". code_style.sh makes it this way, and Gemini agrees :)

PR #1360, opened 2025-03-07
Add DeepSeek ckpt conversion script

RissyRan:
# Description

Add DeepSeek ckpt conversion script for DeepSeek v2-16b and v3-671b.

# Tests

* Tested it locally with v2-16b in training to compare logits

# Checklist


MaxText/convert_deepseek_ckpt.py
```diff
@@ -0,0 +1,517 @@
...
+        "qk_nope_head_dim": 128,
+        "qk_rope_head_dim": 64,
+        "v_head_dim": 128,
+    },
+}
+
+
+# skip to convert ckpts related to fp8 and MTP for now
+SKIPPED_KEYS_SUFFIX = [
+    "_scale_inv",
```
gagika, line 72: I don't think we need this `_scale_inv` here, as this script don't work on fp8 checkpoint and for bf16 converted checkpoint `_scale_inv` .

It will be better if we removed `_scale_inv` check, so it will fail if we run on fp8 checkpoint (even better, if you could check "_scale_inv" suffix and provide an error message that the script shouldn't be run on fp8 checkpoint.

If you remove "_scale_inv" from the list, the rest refers to MTP, so perhaps you can update the name SKIPPED_KEYS_SUFFIX to MTP_KEYS_SUFFIX

gagika (approved):
LGTM, only one comment

MaxText/convert_deepseek_ckpt.py
RissyRan: Sounds good! I will make the update.

RissyRan: Updated.

khatwanimohit (approved):

PR #1367, opened 2025-03-08
Add Assert in runner to avoid invalid gs bucket

Obliviour:
# Description

Add Assert in runner to avoid invalid gs bucket

# Tests

Saw the issue

# Checklist

### DIFF
--- a/benchmarks/maxtext_xpk_runner.py
+++ b/benchmarks/maxtext_xpk_runner.py
@@ -659,7 +659,9 @@ def on_device_benchmark_runner(
 # Run maxtext_xpk_runner.py as a script for executing multiple workloads pythonically!
 def main() -> int:
   # Variables to configure:
-  output_bucket = 'gs://DIR'
+  # For example: 'gs://DIR'
+  output_bucket = None
+  assert output_bucket is not None, f"Please set your output bucket which is current {output_bucket}"
   base_docker_image = _DEFAULT_MAXTEXT_BASE_DOCKER_IMAGE_NAME
 
   # Set up the clusters to run workloads on!


gobbleturk (approved):

benchmarks/maxtext_xpk_runner.py
```diff
@@ -659,7 +659,9 @@ def on_device_benchmark_runner(
 # Run maxtext_xpk_runner.py as a script for executing multiple workloads pythonically!
 def main() -> int:
   # Variables to configure:
-  output_bucket = 'gs://DIR'
+  # For example: 'gs://DIR'
+  output_bucket = None
+  assert output_bucket is not None, f"Please set your output bucket which is current {output_bucket}"
```
bvandermoon, line 664: Just for my understanding, how is this working? I see `output_bucket` set to None, then asserting it is not None on the next line. I think I am just missing something here

gobbleturk: haha that is pretty funny looking

this code is meant to be edited, perhaps we could add a clearer comment

gobbleturk: also maxtext should be runnable without a base output bucket in the future, but currently step time is only printed with TB which requires a base output bucket... we should print step time regardless and then base_outpt_dir = None should be fine

Obliviour: > Just for my understanding, how is this working?

It is not suppose to work until someone sets the output_bucket themselves. This prevents someone from accidently not setting it.

bvandermoon: Got it, thanks both. So the user is supposed to update it in the actual code file themselves, right? Not just as an input argument?

bvandermoon (approved):

benchmarks/maxtext_xpk_runner.py
gobbleturk: Right, currently this file doesn't take any arguments, instead it is meant to be updated

SurbhiJainUSC: Is there a reason for not adding command line arguments?

PR #1371, opened 2025-03-10
Allow to dump at the specified step. This is useful for AutoPGLE

PatriosTheGreat:
# Description

When AutoPGLE is used the modules are getting recompiled after several training steps. We need to be able to dump modules after recompilation.

# Test

Locally launched llama2_7b


gobbleturk (approved):
Looks good, we will merge this on on our end!

PR #1375, opened 2025-03-11
Use ragged input paged attention kernel for paged attention

liurupeng:
# Description

This change uses the ragged input paged attention kernel for both prefill&decode stages. I deliberately kept the kernel call for both stages so in future we could potentially use different kernel for different stages.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/inference/paged_attention.py
+++ b/MaxText/inference/paged_attention.py
@@ -28,6 +28,7 @@ from jax.experimental.pallas.ops.tpu.paged_attention import paged_attention_kern
 from jax.sharding import PartitionSpec as P
 
 from inference import page_manager
+from inference import paged_attention_kernel_v2
 
 # pytype: disable=attribute-error
 
@@ -46,6 +47,7 @@ HEAD = common_types.HEAD
 D_KV = common_types.D_KV
 
 shard_map = shard_map.shard_map
+_use_kernel_v2 = False
 
 
 class PagedAttentionOp(nn.Module):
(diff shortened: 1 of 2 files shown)


MaxText/inference/paged_attention_kernel_v2.py
```diff
@@ -0,0 +1,721 @@
+# Copyright 2025 The JAX Authors.
```
xy12181, line 1: should we import this file from https://github.com/jax-ml/jax/blob/main/jax/experimental/pallas/ops/tpu/ragged_paged_attention.py?

liurupeng: The jax version that maxtext is using doesn't have the kernel yet, and migrate the jax version seems need some work

MaxText/inference/paged_attention.py
```diff
@@ -125,25 +126,69 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
+        kv_lens=page_state.sequence_lengths, #[max_num_seqs]
+        cu_q_lens=c_q_l, #[max_num_seqs+1]
+        page_indices=page_state.page_map, 
+        num_seqs=num_seqs,
+        num_kv_pages_per_block=self.pages_per_compute_block,
+    )
+    return jnp.expand_dims(result, axis=1)
 
-    d: kv_head_dim_size
+  def paged_attention_v1_decode(
```
patemotter, line 183: Is there any benefit to keeping the v1 code here?

MaxText/inference/paged_attention.py
```diff
@@ -125,26 +126,67 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
+    v_p = jnp.permute_dims(value_pages_var.value, (1, 2, 0, 3))
+    c_q_l = jnp.array([0, page_state.sequence_lengths[0]])
+    num_seqs = jnp.array([1])
+    query = query[0]  # [batch_size, max_num_tokens, num_kv_heads, head_dim] to [max_num_tokens, num_kv_heads, head_dim]
+    result = paged_attention_kernel_v2.ragged_paged_attention(
+        q=query,
+        k_pages=k_p,  # [total_num_pages, page_size, num_kv_heads, head_dim]
+        v_pages=v_p,
+        kv_lens=jnp.array([query.shape[0]]),  # max_prefill_length
+        cu_q_lens=c_q_l,
```
patemotter, line 149: What exactly are c_q_l and cu_q_lens? It's not very clear just from the variable naming.

MaxText/inference/paged_attention.py
```diff
@@ -46,6 +47,7 @@
 D_KV = common_types.D_KV
 
 shard_map = shard_map.shard_map
+use_kernel_v2 = False
```
patemotter, line 50: 1. Do we want to default to this being False? 
2. If this variable is going to be in use for a while then we should add it to the actual config in base.yml.

liurupeng: we could set it to true, my worry is that the new kernel doesn't support SPMD, folks who use > 1 host TPU slice will have issue using this. For 2, the long term goal might be having a flag to choose which kernels to use, but this flag is only temporary.

MaxText/inference/paged_attention.py
liurupeng: SG! will update this

MaxText/inference/paged_attention.py
liurupeng: From XLA team's benchmarking, this kernel has 20% performance gain than v2 kernel in "decode only" tasks. So we could keep this in case we use different kernels for different stages

patemotter: Sounds good, maybe add that as a comment that v2 may not always be the best choice.

MaxText/inference/paged_attention.py
patemotter: That's fine, let's default to it being false until we feel more comfortable with the state of things.

MaxText/inference/paged_attention.py
liurupeng: will do

MaxText/inference/paged_attention.py
```diff
@@ -125,26 +127,68 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
-    Annotations:
-    b: batch_size
-    s: sequence length
-    n: query heads
+    """Apply ragged input Paged Attention in prefill only."""
+    k_p = jnp.permute_dims(key_pages_var.value, (1, 2, 0, 3))
+    v_p = jnp.permute_dims(value_pages_var.value, (1, 2, 0, 3))
+    c_q_l = jnp.array([0, page_state.sequence_lengths[0]])  # [0, prefill_true_length]
+    num_seqs = jnp.array([1])
+    query = query[0]  # [batch_size, max_num_tokens, num_kv_heads, head_dim] to [max_num_tokens, num_kv_heads, head_dim]
```
wyzhang, line 143: nit: assert batch_size == 1?

MaxText/inference/paged_attention.py
```diff
@@ -125,26 +127,68 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
       page_state: page_manager.PageState,
   ) -> Array:
-    """Apply Paged Attention.
-
-    Annotations:
-    b: batch_size
-    s: sequence length
-    n: query heads
+    """Apply ragged input Paged Attention in prefill only."""
+    k_p = jnp.permute_dims(key_pages_var.value, (1, 2, 0, 3))
```
wyzhang, line 139: nit: Mind adding a comment for the expected shape of key_pages_var

num_kv_heads, num_pages, tokens_per_page, kv_head_dim_size

MaxText/inference/paged_attention.py
```diff
@@ -125,26 +127,68 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
+    query = query[0]  # [batch_size, max_num_tokens, num_kv_heads, head_dim] to [max_num_tokens, num_kv_heads, head_dim]
+    result = paged_attention_kernel_v2.ragged_paged_attention(
+        q=query,
+        k_pages=k_p,  # [total_num_pages, page_size, num_kv_heads, head_dim]
+        v_pages=v_p,
+        kv_lens=jnp.array([query.shape[0]]),  # max_prefill_length
+        cu_q_lens=c_q_l,  # the accumulative real lengths of requests, starting from 0
+        page_indices=page_state.page_map,
+        num_seqs=num_seqs,
+        # num_kv_pages_per_block=self.pages_per_compute_block, #I saw repeated response when enabled this
```
wyzhang, line 152: nit: Add a TODO. Move the comment above the line of the code.

MaxText/inference/paged_attention.py
```diff
@@ -125,26 +127,68 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
...
+    )
+    return jnp.expand_dims(result, axis=0)
```
wyzhang, line 154: nit: Mind adding a comment for the shape of the result before expand_dims

wyzhang (approved):

MaxText/inference/paged_attention.py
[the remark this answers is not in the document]
liurupeng: thanks! updated

MaxText/inference/paged_attention.py
liurupeng: updated!

MaxText/inference/paged_attention.py
liurupeng: udpated!

MaxText/inference/paged_attention.py
liurupeng: Sg, updated!

MaxText/inference/paged_attention.py
liurupeng: updated, thanks!

gobbleturk (approved):

vipannalla (approved):

PR #1376, opened 2025-03-11
Decouple printing logs to stdout from tensorboard

pavanholla:
Commit 7bc0f48 introduced a config to disable tensorboard. However, it unintentionally disable printing to stdout as well. This CL separates the logic that prints logs  from the logic the writes to tensorboard.

Test:
Check that enable_tensorboard=False and enable_tensorboard=True lead to the same logs on stdout.

### DIFF
--- a/MaxText/metric_logger.py
+++ b/MaxText/metric_logger.py
@@ -60,6 +60,7 @@ class MetricLogger:
           raise ValueError(f"When writing metrics, {self.buffered_step=} was none")
         metrics_to_write = self.buffered_metrics
         steps_to_write = self.buffered_step
+        self.log_metrics(metrics_to_write, steps_to_write)
       self.buffered_metrics = metrics
       self.buffered_step = step
     else:
@@ -75,6 +76,16 @@ class MetricLogger:
 
       if self.config.gcs_metrics and jax.process_index() == 0:
         running_gcs_metrics = self.write_metrics_for_gcs(metrics_to_write, steps_to_write, running_gcs_metrics, is_training)
+
+  def log_metrics(self, metrics, step):
+    """Logs metrics via max_logging"""
+    max_logging.log(
+        f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
+        f"TFLOP/s/device: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
+        f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
+        f"total_weights: {metrics['scalar']['learning/total_weights']}, "
+        f"loss: {metrics['scalar']['learning/loss']:.3f}"
+    )
 
   def write_metrics_locally(self, metrics, step):
     """Writes metrics locally for testing"""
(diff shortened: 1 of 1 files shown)


bvandermoon (approved):

MaxText/metric_logger.py
```diff
@@ -76,6 +78,22 @@ def write_metrics(self, running_gcs_metrics, metrics, step, is_training=True):
...
+        f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
+        f"total_weights: {metrics['scalar']['learning/total_weights']}, "
+        f"loss: {metrics['scalar']['learning/loss']:.3f}"
+    )
+
+  def log_metrics(self, metrics, step, is_training):
+    """Prints metrics"""
+    with jax.spmd_mode("allow_all"):
+      if is_training:
+        self._log_metrics(metrics, step)
```
bvandermoon, line 95: Is it necessary to gate the logging with `with jax.spmd_mode("allow_all")` and `if metrics to write`? I see a [duplicate PR](https://github.com/AI-Hypercomputer/maxtext/pull/1381) was created that isn't gating with those. Are the extra conditions needed here?

pavanholla: I think gating with `if metrics to write` is recommended due to [this comment](https://github.com/AI-Hypercomputer/maxtext/blob/892c6da71626130f3773be5000679a698a1e7bc7/MaxText/metric_logger.py#L51).

W.R.T. `with jax.spmd_mode("allow_all")` , I am trying it out on GPU's. I don't have a way to test on TPU's. If the reviewers are confident that it isn't required, then we can remove it.

MaxText/metric_logger.py
```diff
@@ -60,6 +60,7 @@ def write_metrics(self, running_gcs_metrics, metrics, step, is_training=True):
           raise ValueError(f"When writing metrics, {self.buffered_step=} was none")
         metrics_to_write = self.buffered_metrics
         steps_to_write = self.buffered_step
+        self.log_metrics(metrics_to_write, steps_to_write)
```
SurbhiJainUSC, line 63: I think this method should be moved out of if block. Initially, `self.buffered_metrics` is None, so the metrics for the first step will not be displayed until the next step is completed.

MaxText/metric_logger.py
SurbhiJainUSC: IIUC, we need to wrap only jax computations with `spmd_mode` context manager: [spmd_mode](https://github.com/jax-ml/jax/blob/main/jax/_src/config.py#L1013). In this case, we are just logging the step and the metrics. Better if we could verify this.

MaxText/metric_logger.py
pavanholla: Could that be the original intent? That was my read of [this comment](https://github.com/AI-Hypercomputer/maxtext/blob/892c6da71626130f3773be5000679a698a1e7bc7/MaxText/metric_logger.py#L51).

SurbhiJainUSC: Assigning this to @gobbleturk to clarify.

gobbleturk: Right the point of buffered metrics is to allow jax to queue up and run as much as possible before the metrics writing becomes blocking. I believe this PR is still in that spirit, since it is checking for buffered metrics, only printing the buffered ones, and setting the new buffered ones only after potentially printing

gobbleturk (approved):

SurbhiJainUSC (approved):

SurbhiJainUSC:
@pavanholla - when can we expect this PR to be closed?

pavanholla:
@SurbhiJainUSC I don't have permissions to proceed further. Could someone from the maxtext team help merge and close the PR?

image

pavanholla:
@bvandermoon Is it possible to approve the workflows

image

SurbhiJainUSC (approved):

SurbhiJainUSC:
> @SurbhiJainUSC I don't have permissions to proceed further. Could someone from the maxtext team help merge and close the PR?
> 
> image

I am looking into this.

bvandermoon:
> @bvandermoon Is it possible to approve the workflows
> 
> image

Looks like @SurbhiJainUSC (or someone else) was able to approve the workflows. Please squash your commits into a single commit. Looks like the cl to merge should be created after that

pavanholla:
@SurbhiJainUSC Please trigger the workflows now. FYI, the reason a second commit got added was because I clicked on "Update Branch"

image

PR #1378, opened 2025-03-12
Fix Tokenizer for Llama3 tests

SujeethJinesh:
# Description

Fixes the tokenizer for llama3.1 70b synthetic and real data runs.

# Tests

Tested on v6e clusters.

# Checklist

### DIFF
--- a/benchmarks/maxtext_trillium_model_configs.py
+++ b/benchmarks/maxtext_trillium_model_configs.py
@@ -1064,6 +1064,7 @@ llama3_1_70b_8192_iter_real_data_and_checkpointing_tfds = _add_to_model_dictiona
             "profiler": "xplane",
             "skip_first_n_steps_for_profiler": 10,
             "profiler_steps": 5,
+            "tokenizer_path": "assets/tokenizer_llama3.tiktoken",
         },
         xla_flags=(
             xla_flags_library.DENSE_VMEM_LIMIT_FLAG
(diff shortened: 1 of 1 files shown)


hengtaoguo (approved):

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -1064,6 +1064,7 @@ def _add_to_model_dictionary(
             "profiler": "xplane",
             "skip_first_n_steps_for_profiler": 10,
             "profiler_steps": 5,
+            "tokenizer_path": "assets/tokenizer_llama3.tiktoken",
```
khatwanimohit, line 1067: Did this run for you ? You'll also have to set `tokenizer_type=tiktoken`

SujeethJinesh: Yes it did run over the weekend and didn't need to set the tokenizer_type, should I set it just to be explicit?

This is the only part I wanted to check in from a  draft PR I used to run tests: https://github.com/AI-Hypercomputer/maxtext/pull/1342/files

SujeethJinesh: Chatted offline and confirmed tiktoken was being set in https://github.com/AI-Hypercomputer/maxtext/blob/4bf3beaa5e721745427bfed09938427e369c2aaf/MaxText/configs/models/llama3.1-70b.yml#L30.

bvandermoon (approved):
LGTM, do we have this issue with all LLama3.1 models in this file?

SujeethJinesh:
> LGTM, do we have this issue with all LLama3.1 models in this file?

I did not check the other Llama3.1 models as we were only blocked on these two configs. It may be likely though. I will send up a follow up PR if needed later since I will need to test llama3.1 8b in the next week anyway.

PR #1383, opened 2025-03-12
Add Disruption Manager & Elastic Training Recipe (SIGILL path)

SujeethJinesh:
# Description

This PR creates the disruption manager detailed in go/disruption-manager-for-model-benchmarking. It only creates the SIGILL path and Time step path with future iterations eventually adding the SIGTERM and Step based paths. It also includes a sample recipe for pw elastic training, with an option to also run a McJAX recipe alongside it for comparison.

# Tests

Tested on a v5e cluster. Some bugs with McJAX prevent making sure it's working fully, but the disruption manager code itself works end to end.

# Checklist

### DIFF
--- a/benchmarks/disruption_management/disruption_handler.py
+++ b/benchmarks/disruption_management/disruption_handler.py
@@ -1,0 +1,178 @@
+"""
+ Copyright 2025 Google LLC
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      https://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+import abc
+import dataclasses
+import enum
+import os
+import subprocess
+import sys
+
+parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.append(parent_dir)
+
+from xpk_configs import XpkClusterConfig
+
+
+MCJAX_STANDARD_TARGET_POD_REGEX_SUFFIX = ".*slice-job-0-0.*"
+MCJAX_STANDARD_STEP_POD_REGEX_SUFFIX = ".*slice-job-0-0.*"
+PATHWAYS_STANDARD_TARGET_POD_REGEX_SUFFIX = ".*worker-0-0.*"
+PATHWAYS_STANDARD_STEP_POD_REGEX_SUFFIX = ".*main-0-0.*"
+
+STANDARD_STEP_LOG_REGEX = "completed step: (\\d+)"
+
+PATHWAYS_WORKER_CONTAINER_NAME = "pathways-worker"
+MCJAX_WORKER_CONTAINER_NAME = "jax-tpu"
+
+
+class TriggerType(enum.Enum):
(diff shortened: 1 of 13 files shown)


benchmarks/disruption_management/disruption_handler.py
```diff
@@ -0,0 +1,175 @@
...
+@dataclasses.dataclass
+class DisruptionConfig:
+  """Configuration for workload disruptions and recovery."""
+  name: str  # Name of the disruption config
+  trigger_type: TriggerType  # Trigger for disruption (Time_Seconds or Step)
+  trigger_value: int  # Value for disruption trigger (seconds or steps)
+  disruption_method: DisruptionMethod  # Disruption method (SIGTERM, SIGILL)
+  worker_container_name: str  # Name of the container to disrupt
+
+  # Some disruption types may require a recovery mechanism.
```
shauryagup, line 61: Can you give an example of this inline?

benchmarks/disruption_management/disruption_manager.py
```diff
@@ -0,0 +1,139 @@
...
+      workload_name: The name of the workload to remove.
+    """
+    if workload_name in self.threads_to_monitor:
+      self.threads_to_monitor[workload_name].clear()
+      print(f"Stopped monitoring threads for workload '{workload_name}'.")
+    else:
+      print(f"No monitoring threads found for workload '{workload_name}'.")
+
+  def start_disruptions_and_wait_for_completion(self) -> None:
+    """Waits for all disruption monitoring threads to complete.
```
shauryagup, line 91: Mention that this starts all disruption threads as well and then waits for all of them to complete

benchmarks/maxtext_xpk_runner.py
```diff
@@ -734,12 +728,27 @@ def xpk_benchmark_runner(
...
+      )
+
   # TODO(@vbarr) Support batch workloads.
-  for xpk_workload_name, xpk_workload_cmd in zip(xpk_workload_names, xpk_workload_cmds):
+  for xpk_workload_name, xpk_workload_cmd in zip(
+      xpk_workload_names, xpk_workload_cmds
+  ):
     return_code = run_command_with_updates(xpk_workload_cmd, xpk_workload_name)
     if return_code != 0:
+      disruption_manager.remove_workload(xpk_workload_name)
```
shauryagup, line 746: Shouldn't we do this only for workloads which have a disruption_config and also irrespective of the return_code?

shauryagup (commented):
Thanks!

benchmarks/disruption_management/disruption_handler.py
SujeethJinesh: This feature isn't fully implemented yet. But added a TODO and an explanation of when this may be necessary.

benchmarks/disruption_management/disruption_manager.py
SujeethJinesh: Added.

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: The `run_command_with_updates` is a bit poorly named so added comments for clarification.

Basically, the `run_command_with_updates function` only starts the workload, but doesn't do any monitoring for its completion. If the return code is not 0, then it indicates a start failure, and so we want to not monitor that specific workload (hence removing it from the disruption manager). We still want to monitor the other workloads that have successfully started. Also we can remove any workload name because it's just a no-op if it doesn't exist -- keeps the code simpler to read.

shauryagup (approved):

gobbleturk (approved):

SujeethJinesh:
I will add the pull ready label since checks have passed, and it appears to not be done automatically.

PR #1386, opened 2025-03-13
Disable the padding causal mask and jax compilation cache.

wang2yn84:
# Description

Disable padding causal mask due to this feature has defects. Disable the jax_enable_compilation_cache due to the recent autotuner sharding issue. We need this fix to build the image for a4 NPI. Will revert this change after NV landed the proper fix in the transformer engine for the padding causal mask (should be very soon) and the autotuner sharding fix.

# Tests

commands to reproduce.

# Checklist


yangyuwei (approved):

hengtaoguo (approved):

bvandermoon (approved):
LGTM but please do give a fast follow up with reverting this change like you mentioned in the description. I believe this causes issues related to https://github.com/AI-Hypercomputer/maxtext/issues/878. Let me know if I am misunderstanding something here

MaxText/train.py
```diff
@@ -921,6 +921,7 @@ def train_loop(config, state=None):
 
 def main(argv: Sequence[str]) -> None:
   jax.config.update("jax_default_prng_impl", "unsafe_rbg")
+  jax.config.update("jax_enable_compilation_cache", False)
```
gobbleturk, line 924: Can we add this as a configurable option instead of always setting it to False? Ideally it defaults to True, generally we want the compilation cache

gobbleturk (changes requested):

MaxText/train.py
wang2yn84: Updated. Ptal

wang2yn84:
> LGTM but please do give a fast follow up with reverting this change like you mentioned in the description. I believe this causes issues related to #878. Let me know if I am misunderstanding something here

Thank you! Will revert as long as the NV fix lands. IIUC, with the

> LGTM but please do give a fast follow up with reverting this change like you mentioned in the description. I believe this causes issues related to #878. Let me know if I am misunderstanding something here

Yea that makes sense. Added flags to control the behavior.

MaxText/configs/base.yml
```diff
@@ -632,3 +632,10 @@ pagedattn_pages_per_compute_block: 8
 # Chunked Prefill Parameters
 prefill_chunk_size: 256
 use_chunked_prefill: False
+
+#######################
+### Temporary Flags ###
+#######################
```
gobbleturk, line 638: I think we can just have a single comment for this one flag - I hope we are not in a scenario where this list is populated so often that we need a dedicated section like this

e.g. insteadof ## Temporary Flags ## section
just a single comment # This is a temporary flag until we switch to an updated TE, see more in b/....

gobbleturk (approved):

bvandermoon (approved):

MaxText/configs/base.yml
wang2yn84: Make sense. Updated!

lukebaumann:
> We also need to disable the Jax compilation cache due to the recent autotuner sharding issue. In this PR, we defaults these 2 to True so it won't affect other workloads

Pathwaysutils set the JAX compilation cache to False and this overwrites that to True. Can you submit a follow up PR with something like

```
if "JAX_ENABLE_COMPILATION_CACHE" in os.environ:
  jax.config.update("jax_enable_compilation_cache", os.environ["JAX_ENABLE_COMPILATION_CACHE"])
```

PR #1391, opened 2025-03-14
[MultiLora for inference] Unapply adapter and safe-checks

aman2930:
# Description
- Adding safechecks to raise errors.
- Added unapply_adapter, which performs action reverse of the apply adapter. Previously I was creating a deep copy of the base params to create the merged params and deleting the merged params after prefill. But this operation is more expensive and not scalable for inference of large models. Instead we should support unapply of lora params from base.

# Checklist

### DIFF
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -244,6 +244,12 @@ class MaxEngine(engine_api.Engine):
 
     params, config = lora_utils.load_adapter(self.config, self.abstract_params, adapter_config_path, adapter_weights_path)
 
+    if config is None:
+      raise ValueError(f"Failed to read lora_config from {adapter_config_path}")
+
+    if params is None:
+      raise ValueError(f"Failed to read lora_config from {adapter_config_path}")
+
     config["adapter_path"] = adapter_weights_path
 
     self.print_stats("After load_single_adapter.")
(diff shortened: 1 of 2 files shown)


vipannalla (commented):
Looks great, can you add a unitest where you use random weights and assert `unapply_lora_from_base_params(lora_from_base_params(params) == params`?

vipannalla (approved):
lgtm

MaxText/utils/lora_utils.py
```diff
@@ -68,6 +68,45 @@ def apply_lora_recursively(base_params, lora_params, module_name):
...
+  module for lora in its config.
+  """
+
+  def lora_update_or_base(base_weight, lora_a, lora_b):
+    if lora_a is not None and lora_b is not None:
+      return base_weight - jnp.einsum("br,rnd->bnd", lora_b, lora_a) * lora_scale_factor
+    else:
+      return base_weight  # Keep the base weight if no Lora update
+
+  def unapply_lora_recursively(base_params, lora_params, module_name):
```
bvandermoon, line 91: Why does this need to be recursive?

MaxText/maxengine.py
```diff
@@ -257,6 +263,13 @@ def apply_adapter(self, base_params, adapter_config, adapter_params):
     lora_scale_factor = float(adapter_config["lora_alpha"]) / lora_rank
     lora_utils.apply_lora_on_base_params(base_params, adapter_params, lora_scale_factor)
 
+  def unapply_adapter(self, base_params, adapter_config, adapter_params):
```
bvandermoon, line 266: Is this supposed to be called somewhere?

bvandermoon (approved):
LGTM, just a couple questions. Thanks Aman

MaxText/maxengine.py
aman2930: Yes. It will be called from the JetStream Engine.

PR #1395, opened 2025-03-14
Removed jax pinned version for TPU on stable stack images

Rohan-Bierneni:
# Description

Due to the new stable stack image being released with JAX version 0.5.2, no longer need to pin the libtpu version in stable stack dockerfile if device is TPU. However, to ensure backwards compatability with older stable stack versions, changed the dockerfile to include a conditional that upgrades the libtpu version if using an old stable stack version(0.4.35 AND 0.4.37).

FIXES: b/403608588 

# Tests

Tested the new stable stack image by building maxtext on the new JStS candidate image 0.5.2-rev1 and verifying that they pass DAG's. 

Link to candidate image: https://pantheon.corp.google.com/artifacts/docker/tpu-prod-env-multipod/us/jax-stable-stack/candidate%2Fgpu?e=13802955&inv=1&invt=AbsCTg&mods=allow_workbench_image_override&project=tpu-prod-env-multipod 

The conditional was tested by building docker images using 0.4.37 and 0.5.2 as base images. Then verified in logs that the right libtpu version is being installed.

# Checklist

### DIFF
--- a/maxtext_jax_stable_stack.Dockerfile
+++ b/maxtext_jax_stable_stack.Dockerfile
@@ -2,6 +2,7 @@ ARG JAX_STABLE_STACK_BASEIMAGE
 
 # JAX Stable Stack Base Image
 FROM $JAX_STABLE_STACK_BASEIMAGE
+ARG JAX_STABLE_STACK_BASEIMAGE
 
 ARG COMMIT_HASH
 
@@ -16,18 +17,19 @@ WORKDIR /deps
 COPY . .
 RUN ls .
 
-# b/399968784 - Temporary fix till we install JStS[tpu]>=0.5.1
+
+# For stable stack tpu training images 0.4.37 AND 0.4.35
 # Orbax checkpoint installs the latest version of JAX,
-# but the libtpu version in this base image is older.
+# but the libtpu version in the base image is older.
 # This version mismatch can cause compatibility issues
 # and break MaxText.
+# Upgrade libtpu version if using either of the old stable images
 
 ARG DEVICE
 ENV DEVICE=$DEVICE
 
-RUN if [ $DEVICE = "tpu" ]; then \
-        pip install --no-cache-dir jax[tpu]==0.5.1; \
-    fi
+RUN if [ "$DEVICE" = "tpu" ] && ([ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.4.37-rev1" ] || [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.4.35-rev1" ]); then \
+        pip install --no-cache-dir --upgrade jax[tpu]; fi
 
 # Install Maxtext requirements with Jax Stable Stack
 RUN pip install -r /deps/requirements_with_jax_stable_stack.txt


parambole (approved):
LGTM

maxtext_jax_stable_stack.Dockerfile
```diff
@@ -25,10 +25,6 @@ RUN ls .
 ARG DEVICE
 ENV DEVICE=$DEVICE
 
-RUN if [ $DEVICE = "tpu" ]; then \
-        pip install --no-cache-dir jax[tpu]==0.5.1; \
-    fi
-
```
bvandermoon, line 31: Won't this break MaxText for people using older stable stack versions?

Rohan-Bierneni: due to orbax-checkpoint maxtext requires jax to be on the latest versions and overrides the jax version specified on the older stable stack image. Thus this change should not have an effect and is taken care of in the new tpu training image release.

bvandermoon: Yes, agreed on the new stable stack image release. But what about if someone tries to use the JAX 0.4.37 stable stack image? If this is removed, orbax-checkpoint will bump up the JAX version without making the corresponding libtpu-upgrade, and the workload will fail

Rohan-Bierneni: I understand. However we don't want this restriction to apply to jsts image releases from now on. But in order to be backwards compatible, we would need to keep this restriction for older images. 

From my understanding there are 2 stable stack tpu images: 0.4.35 and 0.4.37. Should I change the code such that the restriction only applies if an image is tagged with either one of these versions. And all other image release from now onwards will be compatible out the gate.

bvandermoon: Yes, good idea. Please also leave a code comment so we remember why it's there for those two versions. Thanks @Rohan-Bierneni

Rohan-Bierneni: Resolved.

maxtext_jax_stable_stack.Dockerfile
```diff
@@ -16,18 +17,19 @@ WORKDIR /deps
...
+# Upgrade libtpu version if using either of the old stable images
 
 ARG DEVICE
 ENV DEVICE=$DEVICE
 
-RUN if [ $DEVICE = "tpu" ]; then \
-        pip install --no-cache-dir jax[tpu]==0.5.1; \
-    fi
+RUN if [ "$DEVICE" = "tpu" ] && ([ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.4.37-rev1" ] || [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.4.35-rev1" ]); then \
+        pip install --no-cache-dir jax[tpu]==0.5.1; fi
```
bvandermoon, line 32: @parambole should this be hard-coded as 0.5.1? Or should it just be set to latest JAX version?

parambole: @bvandermoon - We should set it to the latest JAX version now ( given that we have seen a case with orbax requiring the latest JAX version )

cc: @Rohan-Bierneni

Rohan-Bierneni: Resolved.

bvandermoon (approved):

parambole (approved):
LGTM

RissyRan (approved):

PR #1400, opened 2025-03-15
Fix regex in pytest.ini for running train and decode tests

khatwanimohit:
# Description

We were not running training/decode tests in pytest due to the mismatched regex in pytest.ini

# Tests

Unit tests

# Checklist

### DIFF
--- a/MaxText/pytest.ini
+++ b/MaxText/pytest.ini
@@ -2,7 +2,7 @@
 [pytest]
 testpaths =
     tests
-python_files = *_test.py
+python_files = *_test.py *_tests.py
 addopts =
     -rf --import-mode=importlib --strict-markers
     --ignore=tests/profiler_test.py
--- a/MaxText/tests/decode_tests.py
+++ b/MaxText/tests/decode_tests.py
@@ -77,6 +77,7 @@ class DecodeTests(unittest.TestCase):
   def test_gpu_base(self):
     decode_main(DecodeTests.CONFIGS["base"] + ["attention=dot_product"])
 
+  @pytest.mark.skip(reason="until b/400476456 is fixed")
   @pytest.mark.tpu_only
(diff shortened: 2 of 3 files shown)


MaxText/pytest.ini
```diff
@@ -2,7 +2,7 @@
 [pytest]
 testpaths =
     tests
-python_files = *_test.py
+python_files = *_test.py *_tests.py
```
khatwanimohit, line 5: I realized we had written the unit test to catch this problem but were not running decode_tests.py and train_tests.py until now because of our regex. Scary!

wang2yn84: Sad to hear. Thank you for catching this issue!

bvandermoon, line 5: Thanks Mohit for fixing this. A couple thoughts since this is a concerning issue like you mentioned.
1. I think we should move all tests to a separate /tests directory. This is a pretty standard practice. Then we can just set `python_files = tests/*` or something similar. I think that would avoid this type of issue
2. Tests should generally end with `_test`, not `_tests`. Agree it makes sense to check and run both like you have here. Not sure how we can catch that going forward

bvandermoon (approved):
Approving to get the fix out. But I think we should go back and move all tests to a /tests folder

MaxText/pytest.ini
khatwanimohit: All the tests are inside tests folder which is indicated by testpaths option here. I like your idea, we can change the regex to tests/*.py which will avoid such confusion

gobbleturk (approved):

MaxText/pytest.ini
bvandermoon: Got it, thanks Mohit. And great find on this issue

PR #1401, opened 2025-03-15
support SFT for conversational dataset

SurbhiJainUSC:
# Description
This PR adds a support to run supervised fine-tuning on a conversational dataset.

# Tests
Unit testing, integration testing with HuggingFace dataset
Correctness test with SFTTrainer in TRL

# Checklist


MaxText/configs/base.yml
```diff
@@ -373,19 +373,25 @@ eval_per_device_batch_size: 0.0
...
+packing: True
 
 # direct preference optimization (DPO)
 use_dpo: False
 dpo_label_smoothing: 0.0
 dpo_beta: 0.1
 
+# Supervised Fine-Tuning (SFT)
+use_sft: False
+sft_train_on_completion: False
```
gobbleturk, line 385: this name surprised me at first - I think either
sft_train_on_prompt or sft_train_only_on_completion make more sense? We always train on the completion, the question is whether or not we also train on the prompt?

MaxText/sft_trainer.py
```diff
@@ -0,0 +1,357 @@
...
+    record_goodput,
+    record_scalar_metrics,
+    save_checkpoint,
+    setup_mesh_and_model,
+    train_step,
+    validate_train_config,
+)
+from ml_goodput_measurement import monitoring
+
+EPS = 1e-8
```
gobbleturk, line 54: Can you include a comment what this is used for?

MaxText/train.py
```diff
@@ -126,13 +126,14 @@ def save_checkpoint(
     checkpoint_manager,
     step,
     state,
-    dataset_type="c4",
+    dataset_type="tfds",
```
gobbleturk, line 129: Curious why is this changed?

MaxText/train.py
```diff
@@ -161,6 +162,7 @@ def save_checkpoint(
     return checkpoint_manager.save(
         step,
         args=orbax.checkpoint.args.PyTreeSave(item=state, save_args=save_args, ocdbt_target_data_file_size=chunk_byte_size),
+        force=force,
```
gobbleturk, line 165: Curious why add force in this PR?

gobbleturk (approved):
Awesome, this looks great!!

I love the sft_trainer.py structure and the unit test for correctness =)

Just a nit or two with var names

MaxText/train.py
SurbhiJainUSC: If `checkpoint_period=100` and we are out of data in `steps=20`, then checkpoint manager won't save a checkpoint till 20 steps if we do not set `force=True` for that case.

MaxText/train.py
SurbhiJainUSC: `c4` is not a dataset type `tfds` is.

aireenmei: This is legacy code. 'c4' is the old name of the current tfds pipeline

bvandermoon (approved):

bvandermoon (commented):
Generally LGTM. But looks like the new config values should also be used in a few other places

MaxText/sft_trainer.py
```diff
@@ -0,0 +1,356 @@
...
+      model,
+      mesh,
+      learning_rate_schedule,
+      data_iterator,
+      eval_data_iterator,
+      state,
+  )
+
+
+def train_loop(config, state=None):
```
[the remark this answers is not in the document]
SurbhiJainUSC: The plan is not to merge this with `train.py` because we want to avoid adding unnecessary `if/else` blocks and make `train.py` unmanageable and difficult to read. We will refactor the code in `train.py` and move the common functionalities into a util and then `sft_trainer.py`, `train.py` and other training modules will import those.

MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -165,8 +203,11 @@ def make_hf_train_iterator(
       data_shuffle_seed=config.data_shuffle_seed,
       add_bos=config.add_bos,
       add_eos=config.add_eos,
-      generate_padding_example=True,
+      packing=config.packing,
+      generate_padding_example=False,
       use_dpo=config.use_dpo,
+      use_sft=config.use_sft,
+      sft_train_on_completion_only=config.sft_train_on_completion_only,
```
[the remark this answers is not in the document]
SurbhiJainUSC: We are adding support for SFT for just HF data pipeline for now. We will update tfds and grain pipeline later.
I have alrady updated this change in `eval_iter` for HF pipeline.

MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -146,7 +184,7 @@ def make_hf_train_iterator(
       config.hf_path,
       data_dir=config.hf_data_dir,
       data_files=config.hf_train_files,
-      split="train",
+      split=config.train_split,
```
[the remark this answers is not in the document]
SurbhiJainUSC: Similar to above comment, we are only supporting HF pipeline for now. We are not making any change to other pipelines for now.

MaxText/configs/base.yml
```diff
@@ -369,19 +369,25 @@ eval_per_device_batch_size: 0.0
...
+packing: True
 
 # direct preference optimization (DPO)
 use_dpo: False
 dpo_label_smoothing: 0.0
 dpo_beta: 0.1
 
+# Supervised Fine-Tuning (SFT)
+use_sft: False
+sft_train_on_completion_only: False
```
gobbleturk, line 381: nit: sorry I should have commented in my first review as well - can you have a one line description for "sft_train_on_completion_only" - e.g.

"If False trains on both the prompt and completion tokens, else trains only on completion tokens"

feel free to push back if you think this is obvious, but I think new users of SFT may be confused by this setting

gobbleturk (approved):

MaxText/input_pipeline/_hf_data_processing.py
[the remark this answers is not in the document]
bvandermoon: Thanks Surbhi. So there is a plan to update those locations after this, right? Agreed that this PR only needs the HF pipeline, my concern is that it would be confusing for someone trying to use the values with the other pipeline and finding they have no impact. Sounds like you already have a plan for that though

MaxText/sft_trainer.py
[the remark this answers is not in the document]
bvandermoon: Sounds good, thanks Surbhi

MaxText/input_pipeline/_hf_data_processing.py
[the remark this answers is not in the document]
SurbhiJainUSC: Yes, we will update other pipelines in the future.

MaxText/configs/sft.yml
```diff
@@ -0,0 +1,29 @@
...
+
+base_config: "base.yml"
+
+use_sft: True
+# sft_train_on_completion_only=False trains on both prompt and completion tokens; trains only on completion tokens otherwise
+sft_train_on_completion_only: True
+packing: True
+learning_rate: 2.e-5
+
+# -------------- HF pipeline --------------
```
gagika, line 23: Do we plan to support HF pipeline only?
Can it later be extended to other datasets, e.g. TFDS?

MaxText/configs/sft.yml
```diff
@@ -0,0 +1,29 @@
...
+dataset_type: hf
+hf_path: 'HuggingFaceH4/ultrachat_200k'
```
gagika, line 25: any reason to have hf dataset default in base sft.yml instead of having pass as an argument?

also cloud you name it `hf_dataset` or `hf_dataset_path`?

A9isha: I think having it in sft.yml is equivalent to passing this as default argument - correct?

MaxText/tests/sft_trainer_correctness.py
```diff
@@ -0,0 +1,259 @@
...
+        trl_logits,
+        rtol=1e-05,
+        atol=0.06,
+        equal_nan=False,
+    )
+
+    kl_div = self.get_kl_div(maxtext_logits, trl_logits)
+    assert jax.numpy.all(kl_div < 7e-5)
+
+  def test_maxtext_without_sft_in_trl(self):
```
A9isha, line 234: it seems to be same as comparing with HF model which we do today like https://github.com/AI-Hypercomputer/maxtext/blob/main/end_to_end/tpu/llama2/7b/test_llama2_7b.sh#L80

end_to_end/tpu/test_sft_trainer.sh
```diff
@@ -0,0 +1,51 @@
...
+FINE_TUNED_MODEL_CKPT_PATH=${CHECKPOINTS_PATH}/${largest_dir}/items
+
+# Decode
+python MaxText/decode.py MaxText/configs/sft.yml \
+    run_name=${RUN_NAME}-hf-decode \
+    model_name=${PRE_TRAINED_MODEL} tokenizer_path=assets/tokenizer.llama2 \
+    load_parameters_path=${FINE_TUNED_MODEL_CKPT_PATH} \
+    per_device_batch_size=${PER_DEVICE_BATCH_SIZE} max_target_length=128 max_prefill_predict_length=64 \
+    attention=dot_product decode_sampling_strategy=greedy prompt="<user>Suggest some famous landmarks in London.</user> <assistant>" \
+    autoregressive_decode_assert="1. Tower of London"
```
A9isha, line 51: nit: please add a newline :)

A9isha (commented):
Excellent work Surbhi!

MaxText/configs/sft.yml
SurbhiJainUSC: Aireen will work on extending the support for other datasets.

MaxText/configs/sft.yml
SurbhiJainUSC: Yes, you can pass it as an argument as well. This is just for testing purposes.

MaxText/configs/sft.yml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+base_config: "base.yml"
+
+use_sft: True
+sft_train_on_completion_only: True
```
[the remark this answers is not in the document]
SurbhiJainUSC: Currently, TRL also supports training only on prompts. We have added this configuration just to align with TRL implementation. If our customers don't find this useful, then we can definitely remove this.

MaxText/tests/sft_trainer_correctness.py
SurbhiJainUSC: Yes, I have removed this in another PR.

end_to_end/tpu/test_sft_trainer.sh
SurbhiJainUSC: Updated in another PR.

PR #1404, opened 2025-03-16
Add Step Monitor to Disruption Manager & Suspend/Resume Recipe (SIGTERM Path)

SujeethJinesh:
# Description

Adds a Step Monitor to the Disruption Manager so the user can specify they want a trigger to occur at a specific step. I also added a SIGTERM Handler that will SIGTERM the primary process on the pod, which simulates a maintenance event. To exercise these paths I also adds a Suspend/Resume Recipe for pathways.

Note that this implementation does not simulate an unschedulable pod using maintenance handler labels as there are some possible problems with that. This SIGTERM path appears to be consistently able to reproduce Suspend/Resume logic.

This PR should be checked into main after https://github.com/AI-Hypercomputer/maxtext/pull/1383 is checked in.

# Tests

Tested on a v5e cluster. b/395936838 and b/401565554

# Checklist

### DIFF
--- a/benchmarks/disruption_management/disruption_handler.py
+++ b/benchmarks/disruption_management/disruption_handler.py
@@ -15,12 +15,14 @@ import abc
 import dataclasses
 import enum
 import os
-import subprocess
 import sys
 
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 sys.path.append(parent_dir)
 
+
+from disruption_management.disruption_utils import execute_command_as_subprocess
+from disruption_management.disruption_utils import get_pod_name_from_regex
 from xpk_configs import XpkClusterConfig
 
 
@@ -28,8 +30,6 @@ MCJAX_STANDARD_TARGET_POD_REGEX_SUFFIX = ".*slice-job-0-0.*"
 MCJAX_STANDARD_STEP_POD_REGEX_SUFFIX = ".*slice-job-0-0.*"
 PATHWAYS_STANDARD_TARGET_POD_REGEX_SUFFIX = ".*worker-0-0.*"
 PATHWAYS_STANDARD_STEP_POD_REGEX_SUFFIX = ".*main-0-0.*"
-
-STANDARD_STEP_LOG_REGEX = "completed step: (\\d+)"
 
 PATHWAYS_WORKER_CONTAINER_NAME = "pathways-worker"
 MCJAX_WORKER_CONTAINER_NAME = "jax-tpu"
@@ -71,6 +71,9 @@ class DisruptionConfig:
   # Target pod regex needed for triggering disruption.
   target_pod_regex: str = MCJAX_STANDARD_TARGET_POD_REGEX_SUFFIX
 
+  # Step pod regex needed for step-based disruption.
+  step_pod_regex: str = MCJAX_STANDARD_STEP_POD_REGEX_SUFFIX
+
 
 class DisruptionHandler(abc.ABC):
   """Abstract interface for disruption handlers."""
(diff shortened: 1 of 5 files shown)


SujeethJinesh (commented):
This change must be checked into main only after https://github.com/AI-Hypercomputer/maxtext/pull/1383 is checked in!

benchmarks/disruption_management/disruption_handler.py
```diff
@@ -163,10 +164,85 @@ def trigger_disruption(
       print(f"error: {e}")
 
 
+class SIGTERMHandler(DisruptionHandler):
+  """Handles SIGTERM disruption by sending a SIGTERM signal to the pod."""
+
+  def trigger_disruption(
```
shauryagup, line 170: Can we move the common functionality like greping for the matching pod prefix between the SIGTERM and SIKILL handler into a common function?

benchmarks/disruption_management/monitor.py
```diff
@@ -54,18 +57,154 @@ def __init__(
...
     """Initializes StepMonitor."""
     super().__init__(workload_name, disruption_config)
-    self.target_pod_regex = target_pod_regex
-    if not target_pod_regex:
-      raise ValueError("Target pod regex is required for monitoring.")
+    self.step_pod_regex = step_pod_regex
+    if not step_pod_regex:
+      raise ValueError("Step pod regex is required for monitoring.")
+
+  def _wait_for_pod_to_start(self) -> str | None:
```
shauryagup, line 68: Should we cache the value of this and just return it if set?

benchmarks/disruption_management/monitor.py
```diff
@@ -54,18 +57,154 @@ def __init__(
...
+        f"Workload '{self.workload_name}': Timed out waiting for step pod"
+        f" matching '{self.step_pod_regex}' to reach 'Running' state."
+    )
+    return None
 
-  # TODO(sujinesh): Implement step monitor.
   def monitor_and_detect_trigger(self):
-    """Monitors logs and detects step trigger."""
-    raise NotImplementedError("Step monitor not implemented yet.")
+    """Monitors logs in real-time and detects step trigger by reading chunks of logs."""
```
shauryagup, line 143: Can we clarify that this waits for the target pod to start and returns if the trigger condition is met at that moment in time?

benchmarks/recipes/pw_suspend_resume.py
```diff
@@ -0,0 +1,148 @@
...
+COUNTRY = "us"
+DEVICE_TYPE = "v6e-256"
+
+# Other parameters (MUST BE SET BY USER)
+XPK_PATH = "../xpk"  # We're running this script from the maxtext directory
+USER = os.environ["USER"]
+BASE_OUTPUT_DIRECTORY = (
+    f"gs://{USER}-{PROJECT}-{COUNTRY}/disruption_management/"
+)
+MAX_RESTARTS = 10
```
shauryagup, line 50: Reduce this to 0 since we should not have any job restarts in suspend/resume?

benchmarks/disruption_management/disruption_handler.py
SujeethJinesh: Done.

benchmarks/disruption_management/monitor.py
SujeethJinesh: Done.

benchmarks/recipes/pw_suspend_resume.py
SujeethJinesh: Done.

benchmarks/disruption_management/monitor.py
SujeethJinesh: I refactored some things and moved this function to a util function instead, so maybe unnecessary now.

shauryagup (approved):

gobbleturk:
I like the idea of a disruption handler, but this is very strongly coupled to our benchmarking tools. From this PR, it looks like the disruption handler is only supported with our benchmarking tools. The name "benchmarking" indicates to me a short 20-50 step run to measure MFU, perhaps we should consider either supporting these disruption tools more generally or commit to having the benchmark tools as the main way to run, which should include renaming our benchmark directory to something more general

SujeethJinesh:
> I like the idea of a disruption handler, but this is very strongly coupled to our benchmarking tools. From this PR, it looks like the disruption handler is only supported with our benchmarking tools. The name "benchmarking" indicates to me a short 20-50 step run to measure MFU, perhaps we should consider either supporting these disruption tools more generally or commit to having the benchmark tools as the main way to run, which should include renaming our benchmark directory to something more general

Talked about this offline, that we should transition the name "benchmarks" to a more appropriate name, along with decoupling disruption manager from the xpk runner in the future (already part of the consideration in go/disruption-manager-for-model-benchmarking)

benchmarks/disruption_management/disruption_handler.py
```diff
@@ -0,0 +1,155 @@
...
+        f" {target_pod_regex} 🔥🔥🔥"
+    )
+    container_name = disruption_config.worker_container_name
+
+    pod_name = get_pod_name_from_regex(workload_name, target_pod_regex)
+    if not pod_name:
+      return
+
+    kill_command = (
+        f"kubectl exec -it {pod_name} -c {container_name} -- /bin/bash -c "
```
shauryagup, line 113: This would need to be changed to `/bin/sh` after cl/741197499

benchmarks/disruption_management/disruption_handler.py
```diff
@@ -0,0 +1,155 @@
...
```
shauryagup, line 139: This would need to be changed to `/bin/sh` after cl/741197499

gobbleturk (approved):

SujeethJinesh:
Adding pull ready since tests passed

PR #1410, opened 2025-03-17
Added explicit pathwaysutils.initialize() call to MaxText

lukebaumann:
# Description

pathwaysutils will soon require a call to initialize().

# Tests

Depending on GitHub actions for tests

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -45,7 +45,7 @@ import max_logging
 import optimizers
 import profiler
 import pyconfig
-import pathwaysutils  # pylint: disable=unused-import
+import pathwaysutils
 import tensorflow as tf
 
 from metric_logger import MetricLogger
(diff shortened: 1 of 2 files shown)


MaxText/train.py
```diff
@@ -920,6 +920,7 @@ def train_loop(config, state=None):
 
 
 def main(argv: Sequence[str]) -> None:
+  pathwaysutils.initialize()
```
bvandermoon, line 923: Just for my understanding - what happens for runs that are not using pathways? Does this initialization need to be behind some type of `is_pathways_enabled` flag?

lukebaumann: Inside pathwaysutils.initialize() it checks to see if pathways is used and will be a noop if it is not used.

gobbleturk (approved):

bvandermoon (approved):

PR #1413, opened 2025-03-18
Skip checkpointing at step=0

khatwanimohit:
# Description

- Skip checkpointing at step=0
- add abs for max numerical diff log in forward_pass_checker

# Tests

Integration tests

# Checklist


MaxText/train.py
```diff
@@ -132,7 +132,7 @@ def save_checkpoint(
 ) -> bool:
   """Wrapper for saving checkpoint."""
   if config and config.enable_checkpointing:
-    if (step % config.checkpoint_period == 0) or (
+    if (step % config.checkpoint_period == 0 and step != 0) or (
```
bvandermoon, line 135: Just for my understanding - was this causing an issue? Or are you just removing it since there isn't anything to checkpoint before the training has started?

bvandermoon (approved):

MaxText/train.py
khatwanimohit: We were checkpointing at step=0 which just causes delays while there is nothing new to ckpt. There is no issue, just wasteful.

gobbleturk:
@xuefgu originally added the checkpoint at step 0, do you have strong opinions about removing it?

xuefgu:
> @xuefgu originally added the checkpoint at step 0, do you have strong opinions about removing it?

No objections. My ancient change, if memory serves, was only to avoid calling `save_checkpoint` when `step` was not a multiple of the interval.

Two nits on the PR though:

1. We could place the `step != 0` check in the front.
2. Line 136 could benefit from the same check.

PR #1415, opened 2025-03-18
Refactor page attention logic

liurupeng:
# Description

The changes is based on Pate and Wangyuan's work for refactoring page manager and integrating with JetStream.
It does below things:
1. Keep a global page_state and page_manager for each maxengine instance
2. Page_state will be the adapter to each prefill/insert/generate for sharing page information

tested with command: 
```
python MaxText/decode.py   MaxText/configs/base.yml   tokenizer_path=assets/tokenizer.llama2   load_parameters_path=gs://rupliu-llama2-7b-f32/unscanned/checkpoints/0/items/   max_prefill_predict_length=16   max_target_length=32   model_name=llama2-7b   ici_fsdp_parallelism=1   ici_autoregressive_parallelism=1   ici_tensor_parallelism=-1   scan_layers=false   weight_dtype=float32   per_device_batch_size=1   checkpoint_is_quantized=false  attention=paged  pagedattn_num_pages=64   pagedattn_tokens_per_page=8   pagedattn_pages_per_compute_block=4
```

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/inference/page_manager.py
+++ b/MaxText/inference/page_manager.py
@@ -22,6 +22,8 @@ of variable-length sequences by dividing the attention context into fixed-size p
 similar to virtual memory systems.
 """
 
+# TODO Need to update unit tests for this file under Maxtext/tests/page_manager.py.
+
 from typing import Optional, Tuple
 
 import common_types
@@ -33,6 +35,7 @@ from flax import struct
 Array = common_types.Array
 DType = common_types.DType
 AxisNames = common_types.AxisNames
+Config = common_types.Config
 
 
 @struct.dataclass
@@ -56,7 +59,22 @@ class PageState:
   current_page_position: Array
 
 
-class PageManager(nn.Module):
+def initialize_page_state(
+    num_pages: int,
+    num_slots: int,
+    max_pages_per_slot: int,
+) -> PageState:
+  return PageState(
+      page_status=jnp.zeros((num_pages,), dtype=jnp.int32),
+      page_map=jnp.zeros((num_slots, max_pages_per_slot), dtype=jnp.int32),
+      sequence_lengths=jnp.zeros((num_slots,), dtype=jnp.int32),
+      num_pages_used=jnp.zeros((num_slots,), dtype=jnp.int32),
+      current_page=jnp.zeros((num_slots,), dtype=jnp.int32),
+      current_page_position=jnp.zeros((num_slots,), dtype=jnp.int32),
+  )
+
+
+class PageManager:
   """Manages paged attention mechanism for efficient sequence processing.
 
   The PageManager implements a virtual memory-like system for attention, where the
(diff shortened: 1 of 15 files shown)


MaxText/inference/page_manager.py
```diff
@@ -22,6 +22,8 @@
 similar to virtual memory systems.
 """
 
+# TODO(rupliu) Add unit tests for this file
```
patemotter, line 25: Just adding a note that these tests largely exist already, but may need some tweaking when added.

https://github.com/AI-Hypercomputer/maxtext/blob/page_manager/MaxText/tests/page_manager_test.py.

MaxText/layers/attentions.py
```diff
@@ -1858,6 +1860,7 @@ def __call__(
       model_mode: str = common_types.MODEL_MODE_TRAIN,
       deterministic: bool = False,
       previous_chunk: Any = None,
+      slot: Optional[int] = 0,
```
patemotter, line 1863: Should we default to `None` instead of `0` in the various functions with `slot` as a parameter since `slot=0` is a valid case? It could be inadvertently hiding issues.

patemotter (commented):
Overall LGTM, just a minor comment about defaults for `slot`.

MaxText/inference/page_manager.py
liurupeng: thanks, updated!

MaxText/layers/attentions.py
liurupeng: Updated

patemotter (commented):
Overall LGTM, I think the only change would be to the defaults for `slot`. Right now we use a combination of 

`slot: Optional[int] = None`, 
`slot: Optional[int] = 0`, 

I think the best long-term plan is to use `slot: Optional[int] = None` everywhere and handle the logic accordingly. I have had some issues with the `0` default because it's 1) a valid slot and could be hiding issues, and 2) it could also be hiding issues because it's known at compile time and may be impacting jitted functions in a non-obvious way. 

These are based on my own experiences with debugging my version of the code, but it may not be as impactful in your version.

liurupeng:
> Overall LGTM, I think the only change would be to the defaults for `slot`. Right now we use a combination of
> 
> `slot: Optional[int] = None`, `slot: Optional[int] = 0`,
> 
> I think the best long-term plan is to use `slot: Optional[int] = None` everywhere and handle the logic accordingly. I have had some issues with the `0` default because it's 1) a valid slot and could be hiding issues, and 2) it could also be hiding issues because it's known at compile time and may be impacting jitted functions in a non-obvious way.
> 
> These are based on my own experiences with debugging my version of the code, but it may not be as impactful in your version.

Thanks, update all default from 0 to None

patemotter (approved):

MaxText/inference/paged_attention.py
```diff
@@ -331,7 +323,7 @@ def update_decode_step_pages(self, key_pages_var, value_pages_var, key, value, p
     kv_heads, num_pages, tokens_per_page, head_dim = key_pages.shape
 
     new_key = key.reshape(batch_size, kv_heads, head_dim)[:, :, :]
-    new_key = jnp.transpose(new_key, (1, 0, 2))  # [n_kv, b, d]
+    new_key = jnp.transpose(new_key, (1, 0, 2))  # [n_kv_heads, batch_size, head_dim]
```
wang2yn84, line 326: Nit, for comments, if you'd like to change the format from [n_kv, b, d] to [n_kv_heads, batch_size, head_dim], please keep it consistent. The comments in the following lines are still using the old format.

MaxText/layers/models.py
```diff
@@ -164,7 +165,14 @@ class SequentialBlockDecoderLayers(nn.Module):
 
   @nn.compact
   def __call__(
-      self, inputs: jnp.ndarray, decoder_segment_ids, decoder_positions, deterministic, model_mode, page_state=None
+      self,
+      inputs: jnp.ndarray,
+      decoder_segment_ids,
+      decoder_positions,
+      deterministic,
```
wang2yn84, line 172: Nit, please add types.

MaxText/layers/models.py
```diff
@@ -576,41 +588,6 @@ def setup(self):
...
-      )
-    elif model_mode == common_types.MODEL_MODE_AUTOREGRESSIVE:
-      page_state = self.page_manager(model_mode)
-    else:
-      raise ValueError(f"Unsupported model_mode {model_mode} by paged attention")
-    return page_state
-
   def __call__(
       self,
       decoder_input_tokens,
```
wang2yn84, line 593: Nit, type here.

MaxText/maxengine.py
```diff
@@ -365,7 +377,7 @@ def prefill_aot(  # pylint: disable=too-many-positional-arguments
     )
 
   @functools.partial(jax.jit, static_argnums=(0,), static_argnames=("request_id",))
-  def prefill(
+  def _prefill_jit(
```
wang2yn84, line 380: Private? __prefill_jit?

MaxText/maxengine.py
```diff
@@ -736,12 +793,14 @@ def prefill_insert(  # pylint: disable=too-many-positional-arguments
     return self.insert(prefix, decode_state, slot)
 
   @functools.partial(jax.jit, static_argnums=(0,), donate_argnums=(2,))
-  def generate(
+  def _generate_jit(
```
wang2yn84, line 796: ditto

MaxText/maxengine.py
```diff
@@ -793,6 +853,31 @@ def generate(
...
+      params: Params,
+      decode_state: DecodeState,
+      sampler: Optional[Callable[[Any], Any]] = None,  # pylint: disable=unused-argument
+      rng: Optional[PRNGKeyType] = None,
+  ) -> Tuple[DecodeState, engine_api.ResultTokens]:
+    """Public API for generate that updates page state outside JIT."""
+    # Update page state before JIT call
+
+    if self.page_manager is not None and self.page_state is not None:
+      self.page_state = self.page_manager.reserve_decode_step_pages(self.page_state)
```
wang2yn84, line 868: reserve_decode_step_pages() seems jitable? Any reason why leave it outside of the jit function?

wang2yn84 (commented):
Thank you Rupeng for the PR. Dropped a few comments.

MaxText/maxengine.py
patemotter: Just tested and in the current state there doesn't seem to be any reason why we can't jit here. In other iterations there may have been a reason for omitting it, but they don't apply currently. 

`@functools.partial(jax.jit, static_argnames=("self",))` gave the same result in decode.py as without

vipannalla (approved):
Did a quick review, looks good. Thanks for refactoring!

MaxText/inference/paged_attention.py
liurupeng: updated

MaxText/layers/models.py
liurupeng: done

MaxText/maxengine.py
wang2yn84: If you have any specific reason to leave it outside of the jittable function, please leave a comment here, otherwise reviewers might get confused. Thank you!

liurupeng:
merged in copybara

MaxText/maxengine.py
wang2yn84: Just trying to clarify, any reason for not changing it?

wyzhang (approved):

liurupeng:
merged into main

PR #1416, opened 2025-03-18
Fix padding token in data pipelines

aireenmei:
# Description

Currently the padding tokens are hard coded to 0. But for some tokenizers (llama3), 0 is a non-special token_id. This PR uses special tokens defined by tokenizer for padding.
FIXES: b/401311301

# Checklist


MaxText/tokenizer.py
```diff
@@ -193,6 +193,10 @@ def __init__(self, model_path: str, add_bos: bool, add_eos: bool):
     with tf.io.gfile.GFile(model_path, "rb") as model_fp:
       sp_model = model_fp.read()
     self.sp_tokenizer = tftxt.SentencepieceTokenizer(model=sp_model, add_bos=add_bos, add_eos=add_eos, reverse=False)
+    self.pad_id = self.sp_tokenizer.string_to_id("<pad>")
+    self.unk_id = self.sp_tokenizer.string_to_id("<unk>")
+    self.bos_id = self.sp_tokenizer.string_to_id("<s>")
```
khatwanimohit, line 198: I don't think all SentencePiece tokenizers have <s>,</s> as EOS and BOS tokens. for example, I checked for Gemma3 tokenizer. BOS token is `<bos>`.

aireenmei: Thanks for the info. I now removed self.bos_id and self.eos_id, they were not used in downstream. I agree it's not ideal to get special tokens using string_to_id(), but I couldn't find a better way since the class tftxt.SentencepieceTokenizer doesn't provide a way to get special token ids. For the definition of pad_id and unk_id, the possible misses will be that the tokenizer don't use <pad> for pad token, then string_to_id("<pad>") gives unk_token_id. And if unk token is not <unk>, string_to_id("<unk>") still gives unk_token_id. Both cases seem fine?

khatwanimohit (approved):

gobbleturk (approved):
Thanks aireen!

MaxText/input_pipeline/_tfds_data_processing_c4_mlperf.py
```diff
@@ -274,7 +278,7 @@ def preprocess_eval_dataset(
     num_examples: Optional[int] = None,
 ) -> tf.data.Dataset:
   """Preprocess the evaluation dataset."""
-  eval_ds = sequence_packing.pack_dataset(eval_ds, max_target_length)
+  eval_ds = sequence_packing.pack_dataset(eval_ds, max_target_length, pad_id=0)
```
suexu1025, line 281: default value is -1, why we need 0 here? can we use same  pad_id for both train and eval?

aireenmei: Made eval the same as train now. Please review again.

PR #1420, opened 2025-03-18
Save the last ckpt when running out of data

aireenmei:
# Description

We found this feature useful when working on SFT which uses a smaller dataset, and we want to save the last ckpt before running out of data. The same change has been added to the sft_trainer. This PR adds it to train.py

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -820,7 +820,23 @@ def train_loop(config, state=None):
 
     with jax.profiler.StepTraceAnnotation("train", step_num=step):
       record_goodput(recorder, config, recorder.record_data_loading_start_time if recorder else None)
-      example_batch = load_next_batch(data_iterator, example_batch, config)
+      try:
+        example_batch = load_next_batch(data_iterator, example_batch, config)
+      except Exception as e:  # pylint: disable=broad-except
+        max_logging.log(f"load_next_batch failed, you may have run out of data. Error message: {e}")
+        if checkpoint_manager is not None:
+          state_to_save = state if not config.use_dpo else _split_dpo_state(state)[0]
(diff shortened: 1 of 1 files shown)


MaxText/train.py
```diff
@@ -820,7 +820,23 @@ def train_loop(config, state=None):
...
+              checkpoint_manager,
+              int(state_to_save.step),
+              state_to_save,
+              config.dataset_type,
+              data_iterator,
+              config,
+              force=True,
+          ):
+            checkpointing.print_save_message(state_to_save.step, config.async_checkpointing)
+        break
```
bvandermoon, line 839: Why break here?

MaxText/train.py
```diff
@@ -820,7 +820,23 @@ def train_loop(config, state=None):
 
     with jax.profiler.StepTraceAnnotation("train", step_num=step):
       record_goodput(recorder, config, recorder.record_data_loading_start_time if recorder else None)
-      example_batch = load_next_batch(data_iterator, example_batch, config)
+      try:
+        example_batch = load_next_batch(data_iterator, example_batch, config)
+      except Exception as e:  # pylint: disable=broad-except
+        max_logging.log(f"load_next_batch failed, you may run out of data. Error message: {e}")
```
bvandermoon, line 826: Nit: "you may _have_ run out of data"

bvandermoon (approved):
Generally LGTM. Is the `break` actually needed though? Please remove it if not

MaxText/train.py
aireenmei: Revised, thanks!

SurbhiJainUSC (approved):

PR #1421, opened 2025-03-19
Adjust chunked_prefill

yuyanpeng-google:
# Description

Add chunked prefill unittest.
Adjust some call to use previous chunk.
Mask from decoder_segment_ids maybe needed, or the attention could wrong with batch or padding.
Some call with decoder_segment_ids=None, return None for some cases in chunked prefill kv cache.

Todo
* Although the kv cache without chunked return key is equal to the cache without quant, it still produce different results.

# Tests

unittest

# Checklist


RissyRan (approved):

vipannalla (approved):

yuyanpeng-google:
unittest will be in the follow PR to remove the dependency to jetstream.

PR #1423, opened 2025-03-19
Update the megablox kernel utility function to support float16

wang2yn84:
# Description

The current function only switches between bf16 and f32. It should also take into consideration of f16.

# Tests

No functional change. Reply on presubmit tests.

# Checklist

### DIFF
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -184,8 +184,8 @@ class MaxEngine(engine_api.Engine):
         .compile(),
     )
 
-  def load_params(self, *args, rng: Optional[PRNGKeyType] = None, **kwargs) -> Params:
-    """Load Parameters, typically from GCS"""
+  def load_params(self, *args, params=None, rng: Optional[PRNGKeyType] = None, **kwargs) -> Params:
+    """Load Parameters from GCS or reshard given Parameters"""
     # pylint: disable=unused-argument
 
     if rng is None:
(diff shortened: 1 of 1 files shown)


yangyuwei (approved):

bvandermoon (approved):
LGTM. Re: `No functional change, rely on presubmit test.`

Is there a presubmit test that will test what you need here? Or do we need to try running with these dtypes separately?

wang2yn84:
> LGTM. Re: `No functional change, rely on presubmit test.`
> 
> Is there a presubmit test that will test what you need here? Or do we need to try running with these dtypes separately?

That's a good question. Which script shall I run to validate the change?

bvandermoon:
> > LGTM. Re: `No functional change, rely on presubmit test.`
> > Is there a presubmit test that will test what you need here? Or do we need to try running with these dtypes separately?
> 
> That's a good question. Which script shall I run to validate the change?

@yangyuwei do you have a GPU-specific test that @wang2yn84 could use here?

MaxText/kernels/megablox/common.py
```diff
@@ -55,5 +55,7 @@ def select_input_dtype(lhs: jnp.ndarray, rhs: jnp.ndarray) -> jnp.dtype:
   # input precision, we need to convert bf16 argument to fp32 beforehand.
   if supports_bfloat16_matmul() and lhs.dtype == jnp.bfloat16 and rhs.dtype == jnp.bfloat16:
     return jnp.bfloat16
+  elif lhs.dtype == jnp.float16 and rhs.dtype == jnp.float16:
```
RissyRan, line 58: I am don't think any current test will cover newly added float16 type?

PR #1426, opened 2025-03-19
Fix the dtype when calling Moe megablox gmm kernel.

wang2yn84:
# Description

When calling the MoE megablox gmm kernle, the preferred_element_type is hardcoded to bfloat16. Replace with config controlled dtype instead of hardcoding.

# Tests

No functional change, reply on presubmit test.

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -445,7 +445,7 @@ class MoeBlock(nn.Module):
             lhs=inputs,
             rhs=kernel,
             group_sizes=group_sizes,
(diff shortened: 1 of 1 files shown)


bvandermoon (approved):
Re: `No functional change, rely on presubmit test.`

Is there a presubmit test that will test what you need here? Or do we need to try running with a couple different dtypes separately?

RissyRan (commented):
Could you help check which type is supported? I recall f32 is not supported previously. Otherwise, we may want to add some assertion to avoid error from kernel.

PR #1437, opened 2025-03-20
Add option to set mu_dtype differently than weight_dtype

gobbleturk:
# Description

This PR adds an option to set mu_dtype differently than weight_dtype.

DeepSeek uses bfloat16 for mu (first moment) and nu (second moment) and float32 for the parameters. Prior to this change we used only one setting weight_dtype for mu,nu and params. Now we are able to set mu to bfloat16 and params to float32. Unfortunately optax does not yet support setting nu differently than params, we may add this functionality to maxtext once its also in optax.
This should save 20% of memory from the optimizer state, instead of 12 bytes per param (float32 mu, float32 nu, float32 params), we can use 10 bytes per param (bf16 mu, float32 nu, float32 params)

FIXES: b/399961932

# Tests

Added unit tests for  mu dtype with and without setting new option mu_dtype

Also locally confirmed with memory traces on a v4-8 that memory usage goes down
(for a 1B model we save 2 bytes / param / 4 devices * 1B params = 0.5GB), and can confirm the mu tensors are bf16 from trace memory viewer pane.

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -115,7 +115,7 @@ decoder_block: "llama2" # which style of DecoderBlock to use.
 # Global parameter scale needs to be a power of 2. If you want finer grained control of the model sizes
 # then you should explicitly set base_embed_dim, base_num_query_heads, base_num_kv_heads,
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
-weight_dtype: float32
+weight_dtype: "float32"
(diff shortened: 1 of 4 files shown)


RissyRan (approved):
Thanks for the change! Wish we could set up 2nd moment's dtype.

bvandermoon (approved):
LGTM. Awesome to see this memory improvement

PR #1439, opened 2025-03-21
Updated get_tokenizer and build_tokenizer in maxengine to support Hug…

gpolovets1:
…gingFace tokenizer types.

# Description

This PR updates Maxtext to be able to use tokenizers from HuggingFace. To do this, it also passes adds a few more tokenizer metadata entries into base.yml and updates get/build_tokenizer to use the updated TokenizerParameters proto.

This change is needed to provide e2e testing of DeepSeek-v3 and to more generally allow using models that use HuggingFace tokenizers.

This is one of 2 PRs needed to fix b/403276371 (the second one is a JetStream PR).

# Tests
I ran the following decode.py command and confirmed the output looked sensible:
python MaxText/decode.py MaxText/configs/base.yml  run_name=test_dsv2_16b_decode max_prefill_predict_length=100 per_device_batch_size=1 model_name=deepseek2-16b async_checkpointing=false ici_tensor_parallelism=4 max_target_length=2048 ici_fsdp_parallelism=1 ici_autoregressive_parallelism=2 prompt="An attention function can be described as mapping a query and a set of key-value pairs to an output, where the query, keys, values, and outputs are all vectors. The output is " megablox=False sparse_matmul=False scan_layers=False attention=dot_product tokenizer_type=huggingface tokenizer_path=deepseek-ai/DeepSeek-V2-Lite load_parameters_path=gs://agagik-us/deepseek/maxtext_checkpoints/ds_16b_unscanned_new_3/unscanned_chkpt/checkpoints/0/items scan_layers=False

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -364,7 +364,8 @@ tokenizer_path: "assets/tokenizer.llama2"
 # tfds pipeline supports tokenizer_type: sentencepiece, huggingface, tiktoken
 # grain pipeline supports tokenizer_type: sentencepiece, huggingface
 # hf pipeline only supports huggingface type, and will ignore tokenizer_type flag
-tokenizer_type: "sentencepiece"
+tokenizer_type: "sentencepiece" # Currently supporting: "tiktoken", "sentencepiece", "huggingface"
+use_chat_template: False
 tokenize_train_data: True  # False if the dataset is pre-tokenized
 tokenize_eval_data: True  # False if the dataset is pre-tokenized
 add_bos: True
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -93,7 +93,14 @@ def main(argv: Sequence[str]) -> None:
   text = config.prompt
   metadata = engine.get_tokenizer()
   tokenizer_model = engine.build_tokenizer(metadata)
-  tokens, true_length = tokenizer_model.encode(text, is_bos=True, prefill_lengths=[config.max_prefill_predict_length])
+  try:
+    # TODO: update jetstream.engine.tokenizer_api.Tokenizer to maintain tokenizer state.
+    has_chat_template = getattr(tokenizer_model.tokenizer, "chat_template", False)  # pytype: disable=attribute-error
+  except AttributeError as _:
+    has_chat_template = False
+  tokens, true_length = tokenizer_model.encode(
+      text, is_bos=not has_chat_template, prefill_lengths=[config.max_prefill_predict_length]
+  )
   assert true_length <= config.max_prefill_predict_length, "can't take too many tokens"
   assert config.quantization != "fp8", "fp8 on NVIDIA GPUs is not supported in decode.py yet"
   assert config.quantization != "nanoo_fp8", "NANOO fp8 on AMD MI300/MI325 GPUs is not supported in decode.py yet"
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -32,7 +32,8 @@ from jax.experimental import layout as jax_layout
 import common_types
 from jetstream.core import config_lib
 from jetstream.engine import engine_api
-from jetstream.engine import tokenizer_pb2
+from jetstream.engine.tokenizer_pb2 import TokenizerParameters
+from jetstream.engine.tokenizer_pb2 import TokenizerType
 from jetstream.engine import tokenizer_api
 from jetstream.engine import token_utils
 from utils import lora_utils
@@ -40,7 +41,6 @@ from utils import lora_utils
 import max_utils
 import inference_utils
 import pyconfig
-
 import warnings
 
 warnings.simplefilter("ignore", category=FutureWarning)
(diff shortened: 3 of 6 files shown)


MaxText/configs/base.yml
```diff
@@ -364,7 +364,10 @@ tokenizer_path: "assets/tokenizer.llama2"
 # tfds pipeline supports tokenizer_type: sentencepiece, huggingface, tiktoken
 # grain pipeline supports tokenizer_type: sentencepiece, huggingface
 # hf pipeline only supports huggingface type, and will ignore tokenizer_type flag
-tokenizer_type: "sentencepiece"
+tokenizer_type: "sentencepiece" # Currently supporting: "tiktoken", "sentencepiece", "huggingface"
+use_chat_template: False
+# Specify if the tokenizer API (e.g. HuggingFace) requires an access token.
+tokenizer_access_token: ""
```
RissyRan, line 370: Can we re-use `hf_access_token`?

MaxText/configs/base.yml
```diff
@@ -364,7 +364,10 @@ tokenizer_path: "assets/tokenizer.llama2"
 # tfds pipeline supports tokenizer_type: sentencepiece, huggingface, tiktoken
 # grain pipeline supports tokenizer_type: sentencepiece, huggingface
 # hf pipeline only supports huggingface type, and will ignore tokenizer_type flag
-tokenizer_type: "sentencepiece"
+tokenizer_type: "sentencepiece" # Currently supporting: "tiktoken", "sentencepiece", "huggingface"
+use_chat_template: False
```
RissyRan, line 368: Could you add a comment that indicates use_chat_template will affect is_bos in encode?

MaxText/maxengine.py
```diff
@@ -1165,16 +1165,26 @@ def get_prefix_destination_sharding(self) -> Any:
...
+
+  def build_tokenizer(self, metadata: TokenizerParameters) -> tokenizer_api.Tokenizer:
     """Return a tokenizer"""
-    if "tiktoken" in metadata.path:
+    if metadata.tokenizer_type == TokenizerType.tiktoken:
       return token_utils.TikToken(metadata)
-    else:
+    elif metadata.tokenizer_type == TokenizerType.sentencepiece:
       return token_utils.SentencePieceTokenizer(metadata)
+    elif metadata.tokenizer_type == TokenizerType.huggingface:
```
RissyRan, line 1185: Shall we have a branch for non supported type? Or `get_tokenizer` is called here?

RissyRan (commented):
Thanks for this support!

MaxText/configs/base.yml
```diff
@@ -364,7 +364,10 @@ tokenizer_path: "assets/tokenizer.llama2"
...
+# Specify if the tokenizer API (e.g. HuggingFace) requires an access token.
```
mitalisi, line 369: Should we add a check in validate config for this?

mitalisi (approved):
Overall looks good. Do you need to make changes to JetStream first before committing this?

MaxText/configs/base.yml
gpolovets1: Oh thanks I didn't see this!

MaxText/configs/base.yml
gpolovets1: It won't really affect it - I override is_bos=False if chat_template is specified. I've added comments and prints of this happening in other sections of the code (e.g. token_utils.py).

MaxText/maxengine.py
gpolovets1: Yea get_tokenizer will already take care for this (I tested locally) so I thought this could be overkill.

MaxText/configs/base.yml
gpolovets1: I added a check in get_tokenizer - do you think I should add more in validate config?

MaxText/decode.py
```diff
@@ -93,7 +93,9 @@ def main(argv: Sequence[str]) -> None:
   text = config.prompt
   metadata = engine.get_tokenizer()
   tokenizer_model = engine.build_tokenizer(metadata)
-  tokens, true_length = tokenizer_model.encode(text, is_bos=True, prefill_lengths=[config.max_prefill_predict_length])
+  has_chat_template = getattr(tokenizer_model.tokenizer, "chat_template", False)
```
SurbhiJainUSC, line 96: What is the purpose of `chat_template`? The reason I asked is if you are applying Tokenizer's chat template somewhere in the code then it might affect my SFT workflow.

MaxText/configs/base.yml
gpolovets1: Thought about this a bit more and I felt tokenizer_accesss_token is a more general name that can be used for any repo that require an access token. It is also utilized by TokenizerParameters, which is more general beyond just HF.
On the other hand, I looked a bit more and hf_access_token seems to be used in other parts of the codebase that rely specifically on HF and this code does not seem to intersect with token_utils. hf_access_token touches a ton of code (maybe training-related?) so it may not be worth replacing it with tokenizer_access_token.

MaxText/decode.py
SurbhiJainUSC: For SFT case (instruction-fine tuning), we are applying chat template which is different from the tokenizer's template. This is by design, so want to make sure that we do not apply another template to it. Please verify that SFT unit/integration/e2e tests work.

MaxText/configs/base.yml
RissyRan: wondering what're other use cases for tokenizer_access_token other than huggingface?

MaxText/maxengine.py
RissyRan: ok, see you added `raise ValueError(f"Unsupported tokenizer type: {metadata.tokenizer_type}")` here

RissyRan (commented):
Others LGTM, just one comment for that tokenizer_access_token.

MaxText/configs/base.yml
mitalisi: Validation in pyconfig happens before the actual run. I was suggesting a check that access token is configured if tokenizer type is hugging face. Its a small change - nice to have but not a must.

MaxText/configs/base.yml
gpolovets1: I updated the code to use just hf_access_tokenizer and removed tokenizer_access_token references.

RissyRan (approved):
LGTM! Please address remaining comments from other reviewers. Thank you!

MaxText/decode.py
gpolovets1: Discussed offline and confirmed that tests ran successfully.

RissyRan:
Hi @gpolovets1, do you think we could merge this PR shortly?

The decoding from DeepSeek will depend on this [feature](https://github.com/AI-Hypercomputer/maxtext/blob/main/end_to_end/tpu/deepseek/Run_DeepSeek.md). Thank you!

MaxText/decode.py
```diff
@@ -93,7 +93,13 @@ def main(argv: Sequence[str]) -> None:
   text = config.prompt
   metadata = engine.get_tokenizer()
   tokenizer_model = engine.build_tokenizer(metadata)
-  tokens, true_length = tokenizer_model.encode(text, is_bos=True, prefill_lengths=[config.max_prefill_predict_length])
+  try:
+    has_chat_template = getattr(tokenizer_model.tokenizer, "chat_template", False)
+  except AttributeError as _:
+    has_chat_template = False
```
bvandermoon, line 99: Why do we need a try/except for this?

bvandermoon (approved):

MaxText/decode.py
gpolovets1: Not all of the tokenizer_util classes have a tokenizer instance variable.

SurbhiJainUSC (approved):

PR #1442, opened 2025-03-21
Updates the jax compilation cache setting.

wang2yn84:
# Description

In PR#1386, the JAX_ENABLE_COMPILATION_CACHE always defaults to True even if not set in the environment variable. However, pathway utils set this to False but this setting overwrites it. This PR only removes it.

# Tests

Manually run and see the JAX_ENABLE_COMPILATION_CACHE takes effect when it's set explicitly.

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -948,7 +948,6 @@ def train_loop(config, state=None):
 def main(argv: Sequence[str]) -> None:
   pathwaysutils.initialize()
   jax.config.update("jax_default_prng_impl", "unsafe_rbg")
-  jax.config.update("jax_enable_compilation_cache", os.environ.get("JAX_ENABLE_COMPILATION_CACHE", True))
   # TF allocates extraneous GPU memory when using TFDS data
   # this leads to CUDA OOMs. WAR for now is to hide GPUs from TF
   tf.config.set_visible_devices([], "GPU")


lukebaumann (approved):
Thanks!

bvandermoon (approved):

yangyuwei (approved):

MaxText/train.py
```diff
@@ -947,7 +947,8 @@ def train_loop(config, state=None):
 
 def main(argv: Sequence[str]) -> None:
   jax.config.update("jax_default_prng_impl", "unsafe_rbg")
-  jax.config.update("jax_enable_compilation_cache", os.environ.get("JAX_ENABLE_COMPILATION_CACHE", True))
+  if "JAX_ENABLE_COMPILATION_CACHE" in os.environ:
+    jax.config.update("jax_enable_compilation_cache", os.environ["JAX_ENABLE_COMPILATION_CACHE"])
```
shauryagup, line 951: Running `JAX_ENABLE_COMPILATION_CACHE=False python3 train.py...` did not disable the compilation cache for me. Setting it explicitly in code like `jax.config.update("jax_enable_compilation_cache", False)` disabled the cache. I am wondering if there is a recommended way to set the env variable `JAX_ENABLE_COMPILATION_CACHE` when one wants to disable it?

lukebaumann: I don't think this code is even needed. If you set the env variable `JAX_ENABLE_COMPILATION_CACHE`, that will set the default value for `jax_enable_compilation_cache` which will be used unless overwritten elsewhere.

wang2yn84: The reason why we keep the config here is because we'd like to default it to True.

wang2yn84: Tested local and I can set the JAX_ENABLE_COMPILATION_CACHE to False, otherwise defaults to True.

PR #1443, opened 2025-03-21
Save checkpoint for last executed step if checkpoint is not saved

SurbhiJainUSC:
# Description
This PR forces checkpoint save for the following two cases:
1. when we run out of data before config.steps
2. when config.steps-1 % config.checkpoint_period  != 0 

# Tests
Ran training for following configurations:
1. `steps=20` and `checkpoint_period=20`: checkpoint saved for step 0 (by periodic ckpt path) and step 19(by forceful save)
3. `steps=21` and `checkpoint_period=20`: checkpoint saved for step 0 (by periodic ckpt path) and step 20 (by periodic ckpt path)
4. `steps=20` and `checkpoint_period=20` and ran out of data in step 1: checkpoint saved for step 0 (by periodic ckpt path)
5. `steps=20` and `checkpoint_period=20` and ran out of data in step 5: checkpoint saved for step 0 (by periodic ckpt path) and step 5 (by forceful save)

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -824,18 +824,6 @@ def train_loop(config, state=None):
         example_batch = load_next_batch(data_iterator, example_batch, config)
       except Exception as e:  # pylint: disable=broad-except
         max_logging.log(f"load_next_batch failed, you may have run out of data. Error message: {e}")
-        if checkpoint_manager is not None:
-          state_to_save = state if not config.use_dpo else _split_dpo_state(state)[0]
-          if save_checkpoint(
-              checkpoint_manager,
-              int(state_to_save.step),
-              state_to_save,
-              config.dataset_type,
-              data_iterator,
-              config,
-              force=True,
-          ):
-            checkpointing.print_save_message(state_to_save.step, config.async_checkpointing)
         break
       record_goodput(recorder, config, recorder.record_data_loading_end_time if recorder else None)
       check_example_batch(config, example_batch=example_batch)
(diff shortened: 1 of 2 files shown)


aireenmei:
Looks like checkpoint_manager is created regardless of enable_checkpointing=False. Will it save the last ckpt when enable_checkpointing=False?

SurbhiJainUSC:
> Looks like checkpoint_manager is created regardless of enable_checkpointing=False. Will it save the last ckpt when enable_checkpointing=False?

No, it won't. `save_checkpoint()` takes care of checkpointing: https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/train.py#L135

MaxText/train.py
```diff
@@ -923,6 +911,22 @@ def train_loop(config, state=None):
       max_utils.print_mem_stats("After params initialized")
 
   if checkpoint_manager is not None:
+    if int(state.step) - 1 % config.checkpoint_period != 0:
```
bvandermoon, line 914: Just want to confirm my understanding here - is the point of the mod to avoid saving twice when the checkpoint period happens to line up with the last step?

bvandermoon: I see that is what's happening based on the commit message. 

Would you mind updating the description with a little more context on the updated criteria for checkpointing? E.g. why are we checking the step before instead of the current state? I am assuming you have it right, I think I am just missing something

gobbleturk (approved):

MaxText/train.py
lukebaumann: `(int(state.step) - 1) % config.checkpoint_period != 0`

This needs to be in parenthesis because the `%` has higher precendence than `-`.

Right now, this is executing `(int(state.step) - (1 % config.checkpoint_period)) != 0` which is the same as `(int(state.step) - 1) != 0` which is always true.

PR #1448, opened 2025-03-24
Fix chunked prefill and fix kvcache

yuyanpeng-google:
Fix conflict in kvcache.
* remove redundant define
* move newest version chunked prefill
* fix decoder_segement_id
* add unittest of chunked prefill

# Tests

unittest

# Checklist

### DIFF
--- a/MaxText/inference/kvcache.py
+++ b/MaxText/inference/kvcache.py
@@ -158,362 +158,6 @@ class KVQuant:
 
 
 class KVCache(nn.Module):
-  max_prefill_length: int
-  max_target_length: int
-  dtype: common_types.DType
-  kv_quant: Optional[KVQuant] = None
-  prefill_cache_logical_axis_names: AxisNames = (CACHE_BATCH_PREFILL, CACHE_SEQUENCE, CACHE_HEADS, CACHE_KV)
-  cache_logical_axis_names: AxisNames = (CACHE_BATCH, CACHE_SEQUENCE, CACHE_HEADS, CACHE_KV)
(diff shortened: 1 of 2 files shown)


MaxText/inference/kvcache.py
```diff
@@ -746,11 +391,16 @@ def chunked_prefill_kv_cache(self, key: Array, value: Array, decoder_segment_ids
       cached_prefill_value_vars[0].value = jax.lax.dynamic_update_slice(
           cached_value_value, value_shaped_for_cache, (next_pos, 0, 0, 0)
       )
-      cached_prefill_segment_id_var.value = decoder_segment_ids
+
+      if decoder_segment_ids is not None:
+        # decoder_segment_ids is complete prompt, while cached_prefill_segment_id is max_prefill_length
+        new_segment_id_var = jnp.zeros(cached_prefill_segment_id_var.value.shape, jnp.int32)
```
mailvijayasingh, line 397: Is this max_prefill_length or 'chunk_size'

mailvijayasingh: Is decoder_segment_ids need a fix here, it will also need a fix in prefill call in maxengine, can you take a look at that part? 

I think this will impact accuracy a little bit

yuyanpeng-google: It is max_prefill_length here.
If we do `output_mask = jnp.logical_and(mask, causal_mask)` in `generate_attention_mask`, the mask which depend on decoder_segment_ids need to be max_prefill_length to match the shape of causal_mask.

The prefill call in maxengine take the complete prompt as the decoder_segment_ids. However, the prompt could be shorter than max_prefill_length using shorter prompt bucket.

gpolovets1 (approved):
LGTM based on Mitali & Vijay's approval.

mailvijayasingh (approved):

PR #1484, opened 2025-03-26
Parameter Offloading for Single/Multi-Devices

zhenying-liu:
Enable parameter offloading for all the parameters, supporting both single and multiple device configurations, with control via a dedicated flag (disabled by default when optimizer state offloading is enabled).

Parameter Offloading Process (When Explicitly Enabled):
1. Initialization: Parameters are initially placed on the host and passed the sharding with memory kind to the jitted function. So they are on the host by default inside the jitted function.
2. Pre-Computation: all parameters: decoder layer, linear, embedding and normalization parameters are moved to the device before any computations occur. This ensures parameters are on the device for both forward and backward pass in the training loop. 
3. Post-Backward Pass: Parameters are moved back to the device before the optimizer update. Otherwise, they are on the host again and can not run on the device efficiently.

Enablement Criteria: param_scan_axis = 0 (no transpose).

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -19,16 +19,19 @@ import operator
 from typing import Any, Callable, Iterable, Optional, Sequence, Tuple, Union
 
 import numpy as np
+import jax
+import jax.numpy as jnp
 
 from jax import lax
 from jax.ad_checkpoint import checkpoint_name
-import jax.numpy as jnp
+from jax._src.sharding_impls import TransferToMemoryKind
 
 from aqt.jax.v2 import aqt_tensor
 
 import flax.linen as nn
 
 from MaxText import common_types
+from MaxText import max_logging
 from MaxText.common_types import DecoderBlockType
 from MaxText.layers import initializers, normalizations, quantizations
 
@@ -106,6 +109,7 @@ class DenseGeneral(nn.Module):
   quant: Optional[Quant] = None
   use_bias: bool = False
   matmul_precision: str = "default"
+  parameter_memory_host_offload: bool = False
 
   @nn.compact
   def __call__(self, inputs: Array) -> Array:
@@ -140,6 +144,10 @@ class DenseGeneral(nn.Module):
           kernel_in_axis,
           kernel_out_axis,
       )
+    # Move logit_dense kernel to device if parameter offloading is enabled
+    if self.parameter_memory_host_offload:
+      max_logging.log("linear.py: Moving parameter logits_dense kernel to device")
+      kernel = jax.device_put(kernel, TransferToMemoryKind("device"))
     kernel = jnp.asarray(kernel, self.dtype)
     contract_ind = tuple(range(0, len(axis)))
     output = _compute_dot_general(inputs, kernel, self.kernel_axes, axis, contract_ind, self.matmul_precision, self.quant)
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -468,8 +468,6 @@ def train_step(model, config, state_mesh_shardings, state, data, dropout_rng):
     aux = jax.tree.map(lambda x: jnp.sum(x, axis=0), aux)  # pytype: disable=module-attr
   else:
     if config.optimizer_memory_host_offload:
-      cast_params = jax.device_put(state.params, max_utils.with_memory_kind(state_mesh_shardings.params, "device"))
-      state = state.replace(params=cast_params)
       if config.use_dpo:
         reference_params = jax.device_put(reference_params, max_utils.with_memory_kind(reference_params_sharding, "device"))
         extra_dpo_args = [reference_params]
@@ -488,6 +486,18 @@ def train_step(model, config, state_mesh_shardings, state, data, dropout_rng):
         opt_state=jax.device_put(
             state.opt_state,
             jax.tree_util.tree_map(lambda x: x.with_memory_kind(kind="device"), state_mesh_shardings.opt_state),
+        )
+    )
+  # Move all parameters to device before optimizer update
+  if config.parameter_memory_host_offload:
+    max_logging.log("\nMoving all parameters to device before optimizer update")
+    def move(path, value):
+      max_logging.log(f"train.py: Moving f{path} to device")
+      return value.with_memory_kind(kind="device")
+    state = state.replace(
+        params=jax.device_put(
+            state.params,
+            jax.tree_util.tree_map_with_path(move, state_mesh_shardings.params),
         )
     )
   new_state = state.apply_gradients(grads=grads)
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -215,6 +215,7 @@ qkv_proj: 'remat'
 out_proj: 'remat'
 
 optimizer_memory_host_offload: False
+parameter_memory_host_offload: False
 scan_layers: True # We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations.
 param_scan_axis: 1
 
--- a/MaxText/layers/embeddings.py
+++ b/MaxText/layers/embeddings.py
@@ -20,8 +20,10 @@ from typing import Any, Optional
 from flax import linen as nn
 import jax
 from jax import lax
+from jax._src.sharding_impls import TransferToMemoryKind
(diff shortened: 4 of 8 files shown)


MaxText/max_utils.py
```diff
@@ -885,6 +880,58 @@ def _cross_entropy_with_logits_bwd(
...
+    'value': ['params', 'decoder', 'layers', 'self_attention', 'value', 'kernel'],
+    'wo':    ['params', 'decoder', 'layers', 'mlp', 'wo', 'kernel'],
+    'out':   ['params', 'decoder', 'layers', 'self_attention', 'out', 'kernel']
+}
+
+def should_offload_parameters(path, allowed_paths):
+  """
+  Determine if parameters should be offloaded based on the number of devices.
+
+  All self-attention and MLP layer parameters will be offloaded for a single device
```
khatwanimohit, line 912: Can you elaborate more on this, please. I didn't understand why we are not offloading outputs when working with multiple devices ?

khatwanimohit (commented):
Thank you for working on this. Can you share a profile with this working. It can be any model and even a single host tpu-vm

MaxText/max_utils.py
zhenying-liu: This was fixed in [XLA PR #24248](https://github.com/openxla/xla/pull/24248). I'm working on a more general solution without specifying the parameter paths.

zhenying-liu:
> Thank you for working on this. Can you share a profile with this working. It can be any model and even a single host tpu-vm

Screenshot 2025-04-02 at 8 32 50 PM
This is a memory profile for llama2-7b on a single GH200 with 96 GB. I enabled both parameter offloading and activation offloading; it will go OOM when either offloading technique was used individually.
The memory stats was also reported at the end of the log, for example,
`Output size: 53907325324, temp size: 33614286824, argument size: 53907406860, host temp size: 11140071424, in bytes. `
These numbers matched the above profile.

zhenying-liu:
Thanks @khatwanimohit for the earlier code review. I managed to do parameter offloading without limiting to self attention and mlp parameters. Please take another look.

MaxText/train.py
```diff
@@ -478,6 +476,18 @@ def reshape_to_microbatch_accumulations(batch_arr):
             jax.tree_util.tree_map(lambda x: x.with_memory_kind(kind="device"), state_mesh_shardings.opt_state),
         )
     )
+  # Move attention parameters to device before optimizer update
+  if config.parameter_memory_host_offload:
+    max_logging.log("\nMoving attention parameters to device before optimizer update")
+    def move(path, value):
+      max_logging.log(f"train.py: Moving f{path} to device")
+      return value.with_memory_kind(kind="device")
+    state = state.replace(
```
khatwanimohit, line 485: For my understanding, we are transferring params to device from here then why do need to have device_put calls in models.py? Can you explain when in the flow "models.py: Moving parameter {path} to device" happens. Does it happen only during initialization ?

zhenying-liu:
I drew the program execution flow to explain. Note the left branches happen before the right branches.

Screenshot 2025-04-24 at 12 24 09 PM


And below is the text version with more details (state.params stores parameters):

Screenshot 2025-04-24 at 2 04 50 PM

zhenying-liu:
From this HLO dump file [module_0041.jit_train_step.sm_9.0_gpu_after_optimizations.txt](https://github.com/user-attachments/files/19900333/module_0041.jit_train_step.sm_9.0_gpu_after_optimizations.txt) (meaning the HLO for jit_train_step function after XLA optimizations), we can see parameter decoder layer mlp's input wi_0 parameter is from the host as it has the host memory space **S(5)** at the beginning of the function main, which is also before the forward pass. It is assigned to %Arg_2.3. %Arg_2.3 is moved from the host to device via a **copy-start** operation. After the copy is done, it's assigned to **%copy-done.4.0** which is on device.
```
ENTRY %main(,.., Arg_2.3,...)
 %Arg_2.3 = f32[32,4096,11008]{2,1,0:S(5)} parameter(2), sharding={replicated}, metadata={op_name="state.params[\'params\'][\'decoder\'][\'layers\'][\'mlp\'][\'wi_0\'][\'kernel\']"}
 ...
 %copy-start.4.0 = (f32[32,4096,11008]{2,1,0}, f32[32,4096,11008]{2,1,0:S(5)}, u32[]) copy-start(%Arg_2.3)
 %copy-done.4.0 = f32[32,4096,11008]{2,1,0} copy-done(%copy-start.4.0)
 %input_add_reduce_fusion.11 = (f32[88064]{0}, f32[32,4096,11008]{2,1,0}) fusion(%copy-done.4.0, %get-tuple-element.633, %get-tuple-element.645, %get-tuple-element.294, %loop_subtract_fusion.1, /*index=5*/%loop_multiply_fusion.6), kind=kInput, calls=%fused_add_reduce.11, metadata={op_name="jit(train_step)/jit(main)/reduce_sum" source_file="/opt/maxtext/MaxText/max_utils.py" source_line=64 deduplicated_name="input_add_reduce_fusion.10"}
```
At the end of the main() function of jit_train_step, we can see some outputs have this kind of shape:
f32[32,4096,11008]{2,1,0:S(5)}
It means the output shape is on the host due to the host memory space S(5), following the out_shardings of pinned_host memory kind.

In short, this is what's happening at the beginning of the forward pass:
there is memory copy H2D for mlp wi_0 parameter and followed by a kernel called input_add_reduce_fusion.10. From the nsys profile, we could find such a pattern. Unfortunately the nsys profile doesn't show the variable name for memcpyH2D,  the size of ~5GB of memcpyH2D matches.
Screenshot 2025-04-24 at 4 22 55 PM
The memcpyH2D is the green bar and the subsequent kernel input_add_reduce_fusion.10 is highlighted.

zhenying-liu:
The error message is not related with the code in this PR:
`Error: No task list was present and requireChecklist is turned on.`

khatwanimohit (approved):

MaxText/layers/linears.py
```diff
@@ -139,6 +143,10 @@ def __call__(self, inputs: Array) -> Array:
           kernel_in_axis,
           kernel_out_axis,
       )
+    # Move logit_dense kernel to device if parameter offloading is enabled
+    if self.parameter_memory_host_offload:
+      max_logging.log(f"linear.py: Moving parameter logits_dense kernel to device")
```
gobbleturk, line 148: remove "f" from string - this might fail our linter

MaxText/train.py
```diff
@@ -489,6 +487,18 @@ def reshape_to_microbatch_accumulations(batch_arr):
             jax.tree_util.tree_map(lambda x: x.with_memory_kind(kind="device"), state_mesh_shardings.opt_state),
         )
     )
+  # Move attention parameters to device before optimizer update
```
gobbleturk, line 490: Isn't this moving all parameters? Why do you reference specifically the attention parameters?

gobbleturk (commented):
Can we add a test with parameter offloading turned on (maybe only run on gpus?) Something like these

https://github.com/AI-Hypercomputer/maxtext/blob/8c0271a10a0ae6d5d32a31aae2d5739c667c6dcc/MaxText/tests/integration_tests/train_tests.py#L175-L210

zhenying-liu:
> Can we add a test with parameter offloading turned on (maybe only run on gpus?) Something like these
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/8c0271a10a0ae6d5d32a31aae2d5739c667c6dcc/MaxText/tests/integration_tests/train_tests.py#L175-L210

Yes, added unit tests for both parameter and optimizer state offloading.

zhenying-liu:
There seems to be integration test failures for the two tests I added: 
1. optimizer state offload: failed with jax.distributed.initialize() for 2 processes. I will try to make number of processes 1 to avoid such a failure.
2. parameter offloading: Error while querying and uploading step deviation to Tensorboard. This will not impact the workload. Error: 403 Request had insufficient authentication scopes. I don't know how to resolve this.

zhenying-liu:
From the test results, it seems I didn't set num_processes properly. Will fix.
`Key num_processes was passed at the command line but isn't in config.`

gobbleturk (approved):

PR #1489, opened 2025-03-27
maxtext_xpk_runner bq upload fixes

raymondzouu:
# Description

- remove the raise error during topology parsing when running with other device types
- fix `upload_metrics_to_bq_cmd` so that it works when bq uploading is disabled
- add missing v5e/v6e slice sizes

# Tests

```
bash docker_build_dependency_image.sh MODE=stable MANTARAY=true
python benchmarks/maxtext_xpk_runner.py 
```

Be sure to modify main of maxtext_xpk_runner.py such that it is targeting the correct xpk cluster. After the model finishes running, you should be able to see DB write success logs https://screenshot.googleplex.com/3EfzU6Dh9QbGShi

# Checklist

### DIFF
--- a/benchmarks/maxtext_xpk_runner.py
+++ b/benchmarks/maxtext_xpk_runner.py
@@ -107,14 +107,21 @@ class WorkloadConfig:
   db_is_test: bool = True
 
   def __post_init__(self):
-    if self.device_type.startswith("v6e"):
+    """Initializes num_devices_per_slice and topology for recording the run into BigQuery"""
+    if self.device_type.startswith("v6e") or self.device_type.startswith("v5e") or self.device_type.startswith("v5litepod"):
       size = int(self.device_type.split("-")[-1])
       if size == 256:
         self.num_devices_per_slice = 256
         self.topology = "16x16"
+      elif size == 128:
+        self.num_devices_per_slice = 128
+        self.topology = "8x16"
       elif size == 64:
         self.num_devices_per_slice = 64
         self.topology = "8x8"
+      elif size == 32:
+        self.num_devices_per_slice = 32
+        self.topology = "4x8"
       elif size == 16:
         self.num_devices_per_slice = 16
         self.topology = "4x4"
@@ -125,9 +132,10 @@ class WorkloadConfig:
         self.num_devices_per_slice = 4
         self.topology = "2x2"
       else:
-        raise ValueError(f"Unsupported v6e size: {size}")
+        raise ValueError(f"Unsupported v5e or v6e size: {size}")
     else:
-      raise ValueError(f"topology and num_devices_per_slice must be inferred when device_type starts with v6e. device_type: {self.device_type}")
+      self.num_devices_per_slice = int(self.device_type.split("-")[1])/2
+      self.topology = ""
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


benchmarks/maxtext_xpk_runner.py
```diff
@@ -122,9 +122,10 @@ def __post_init__(self):
         self.num_devices_per_slice = 4
         self.topology = "2x2"
       else:
-        raise ValueError(f"Unsupported v6e size: {size}")
+        raise ValueError(f"Unsupported v5e or v6e size: {size}")
     else:
-      raise ValueError(f"topology and num_devices_per_slice must be inferred when device_type starts with v6e. device_type: {self.device_type}")
+      self.num_devices_per_slice = int(self.device_type.split("-")[1])/2
+      self.topology = ""
```
bvandermoon, line 128: Does this work fine `topology = ""`?

benchmarks/maxtext_xpk_runner.py
```diff
@@ -107,7 +107,7 @@ class WorkloadConfig:
   db_is_test: bool = True
 
   def __post_init__(self):
-    if self.device_type.startswith("v6e"):
+    if self.device_type.startswith("v6e") or self.device_type.startswith("v5e") or self.device_type.startswith("v5litepod"):
```
bvandermoon, line 110: Do `v5e` and `v5litepod` have different topologies available?

bvandermoon (commented):
Thank you for fixing this @raymondzouu. Do results get updated to BigQuery by default when running with `benchmark_runner.py` as well? Assuming we have built the Docker image with `mantaray=true`

benchmarks/maxtext_xpk_runner.py
raymondzouu: They are the same but xpk is still using `v5litepod` https://github.com/AI-Hypercomputer/xpk/blob/6ba0019ac14f9c0b1d5c51371bc40bc43e3b65a8/src/xpk/core/system_characteristics.py#L1144

benchmarks/maxtext_xpk_runner.py
raymondzouu: Yes topology can be empty however we should change it later so that topology can be specified by the user

benchmarks/maxtext_xpk_runner.py
```diff
@@ -107,14 +107,20 @@ class WorkloadConfig:
   db_is_test: bool = True
 
   def __post_init__(self):
```
gobbleturk, line 109: Can you add a comment why this function is added (big query annotations?)

gobbleturk (approved):

bvandermoon (approved):

benchmarks/maxtext_xpk_runner.py
raymondzouu: Done

raymondzouu:
```
Do results get updated to BigQuery by default when running with benchmark_runner.py as well? Assuming we have built the Docker image with mantaray=true
```
@bvandermoon It looks like we set `generate_metrics_and_upload_to_big_query=False` https://github.com/AI-Hypercomputer/maxtext/blob/c12f9860b8caf6173d0a6b41dad9b329ef17cc81/benchmarks/benchmark_runner.py#L283 in benchmark_runner.py. This is needed if an external user were running it since bigquery uploading is for internal use cases. If you wanted bigquery upload with benchmark_runner.py you would just need to flip that flag.

bvandermoon:
> ```
> Do results get updated to BigQuery by default when running with benchmark_runner.py as well? Assuming we have built the Docker image with mantaray=true
> ```
> 
> @bvandermoon It looks like we set `generate_metrics_and_upload_to_big_query=False`
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/c12f9860b8caf6173d0a6b41dad9b329ef17cc81/benchmarks/benchmark_runner.py#L283
> 
> in benchmark_runner.py. This is needed if an external user were running it since bigquery uploading is for internal use cases. If you wanted bigquery upload with benchmark_runner.py you would just need to flip that flag.

Great, thanks @raymondzouu

PR #1492, opened 2025-03-28
Fix DeepSeek routing logic

RissyRan:
# Description

* Add device-limit logic into DeepSeek routing
* Add a `GateLogit` layer to return pre_bias_output

**I will have another PR to move MoE related blocks out of linears.py into moe.py**

# Tests

* github runners pass
* test smaller config of v2 without sharding issue with `pre_bias_logits_pspec = None` [link](https://paste.googleplex.com/6226575585181696)
* test on v3 logits comparison, and see improvement [link](https://b.corp.google.com/issues/394669494#comment39)
* Reload the v2-lite ckpt, and no issues for decoding 

# Checklist


MaxText/layers/linears.py
```diff
@@ -988,7 +1028,7 @@ def retrieve_quantized_weight(
   def __call__(self, inputs):
     cfg = self.config
     inputs = inputs.astype(cfg.dtype)
-    gate_logits = DenseGeneral(
+    gate_logits, pre_bias_logits = DenseGeneral(
```
gagika, line 1031: instead of modifying DenseGeneral to fit DeepSeek's need, let's create some new layer (perhaps inheriting from DenseGeneral) with DeepSeek specific change, e.g. pre_bias_logits logic and score_func=self.config.routed_score_func, wdyt?

RissyRan: That's a good point. Will update the change.

RissyRan: Hi Gagik, updated to a new module named as `DenseWithNormBias` as `use_bias` is also used in mlp block and gpt3 block, so not DeepSeek specific. Feel free to let me know your thoughts! Thank you!

MaxText/layers/linears.py
```diff
@@ -93,8 +93,6 @@ class DenseGeneral(nn.Module):
     weight_dtype: the dtype of the weights (default: float32).
     dtype: the dtype of the computation (default: float32).
     kernel_init: initializer function for the weight matrix.
-    use_bias: whether to add bias in linear transformation.
-    score_func: scoring function for output normalization.
```
gagika, line 97: the common convention for Dense layer is to have linear layer and bias, 
e.g. flax.linen.Dense, keras.layers.Dense, torch.nn.Linear
Let's follow the convention for DenseGeneral too, unless we have a strong reason not too, and gpt3 can use it as is.

We can remove score_func from DenseGeneral, as it's very DeepSeek specific.

MaxText/layers/linears.py
```diff
@@ -156,11 +152,36 @@ def compute_dot_general(inputs, kernel, axis, contract_ind):
 
     contract_ind = tuple(range(0, len(axis)))
     output = compute_dot_general(inputs, kernel, axis, contract_ind)
+    return output
+
+
+class DenseWithNormBias(DenseGeneral):
```
gagika, line 158: If we are going to have a separate layer Let's make this layer very specific for DeepSeek Gating (e.g. reflecting in the name).

We can also not have a separate layer, and for DeepSeek use `DenseGeneral` for DeepSeek, then  add a bias (while keeping previous weights) and apply scaling.

Both options are good in my opinion.

gagika (commented):
a few comments about the design of the layer.

MaxText/layers/linears.py
RissyRan: Yeah, sounds good. The tricky thing is this is just a dense matmul, not like gating, which also need to combines routing logic. I think we have 2 good options, 1) have 'DenseDeepseek', which will have some duplicate logics with 'DenseGeneral', as score_func change is in the middle (before bias). 2) (this could be a later version) refactor routing logics into a new 'MoeGate', this will need ckpt structure changed (probably just an extra layer), since MoeGate will call DenseGeneral or DenseDeepSeek and then select topk. I will do a refactor option 1 first, and follow up with 2nd option.

RissyRan: Let's quickly chat in person later? and then I will do the refactor once we are aligned. Thank you!

MaxText/configs/base.yml
```diff
@@ -147,6 +147,8 @@ shared_experts: 1
 routed_scaling_factor: 1.0 # scaling factor for routing scores
 routed_score_func: "" # scoring function for routing
 routed_bias: False # a flag if a bias term is added for routing
+n_groups: 1 # number of groups for routing
```
gobbleturk, line 150: can we give this a more specific name like n_routing_groups

RissyRan: Sounds good! Will update both to `n_routing_groups` and `topk_routing_group`

MaxText/layers/linears.py
RissyRan: Discussed offline, and updated accordingly.

MaxText/layers/linears.py
```diff
@@ -106,7 +105,99 @@ class DenseGeneral(nn.Module):
...
+          bias_shape,
+          self.weight_dtype,
+      )
+      bias = jnp.asarray(bias, self.dtype)
+      output += bias
+    return output
+
+
+class GateLogit(nn.Module):
+  """A linear transformation with flexible axes.
```
gagika, line 175: could you update this description?

This layer is used gate logit calculation, allowing for the specific pre-bias logic needed for DeepSeek routing.

MaxText/layers/linears.py
```diff
@@ -106,7 +105,99 @@ class DenseGeneral(nn.Module):
...
+
+  Attributes:
+    features: tuple with numbers of output features.
+    decoder_block: which DecoderBlock to use.
+    axis: tuple with axes to apply the transformation on.
+    weight_dtype: the dtype of the weights (default: float32).
+    dtype: the dtype of the computation (default: float32).
+    kernel_init: initializer function for the weight matrix.
+    kernel_axes: tuple with axes to apply kernel function.
+    use_bias: whether to add bias in linear transformation.
```
gagika, line 185: Could you add something like this: 
whether to add bias in gate logit scores, it's not part of loss function, but used for expert load balancing (similar to DeepSeek v3)

MaxText/layers/linears.py
```diff
@@ -156,9 +247,13 @@ def compute_dot_general(inputs, kernel, axis, contract_ind):
 
     contract_ind = tuple(range(0, len(axis)))
     output = compute_dot_general(inputs, kernel, axis, contract_ind)
+    pre_bias_logits = None
 
     if self.score_func:
       output = _convert_to_activation_function(self.score_func)(output)
+      if self.decoder_block == "deepseek" and self.use_bias:
```
gagika, line 254: Could you check deepseek v3? e.g. config.model_name.startswith("deepseek3")

MaxText/layers/linears.py
```diff
@@ -382,17 +490,47 @@ def deepseek_scale_weights(self, weights):
     weights *= self.config.routed_scaling_factor
     return weights
 
-  def permute(self, inputs, gate_logits):
+  def deepseek_routing(self, gate_logits, pre_bias_logits):
+    """DeepSeek routing logit."""
```
gagika, line 494: Some comments here?

e.g. 
This involves a two-stage selection: first selecting top groups of experts,
then selecting the top experts within those chosen groups. Selection uses
post-bias/post-score_func logits, but the returned weights are based on
pre-bias/pre-score_func logits.

MaxText/layers/linears.py
```diff
@@ -382,17 +490,47 @@ def deepseek_scale_weights(self, weights):
     weights *= self.config.routed_scaling_factor
     return weights
 
-  def permute(self, inputs, gate_logits):
+  def deepseek_routing(self, gate_logits, pre_bias_logits):
```
gagika, line 493: Can this be disabled, if users want to run with different routing?

gagika (approved):
Thanks, only minor comments

MaxText/layers/linears.py
RissyRan: SG. Updated to `A layer used to compute gate logits, allowing to return the pre bias values for DeepSeek routing.`

MaxText/layers/linears.py
RissyRan: Updated to `use_bias: whether to add learnable bias in gate logit scores. When enabled, this bias aids expert load balancing (like in DeepSeek V3) and is not part of the loss calculation.

MaxText/layers/linears.py
RissyRan: sounds good, will update

MaxText/layers/linears.py
RissyRan: Add a TODO `TODO(ranran): Add an option to turn off device-limited routing

MaxText/layers/linears.py
```diff
@@ -157,8 +155,96 @@ def compute_dot_general(inputs, kernel, axis, contract_ind):
...
+  kernel_axes: Tuple[Optional[str], ...] = ()
+  use_bias: bool = False
+  score_func: str = ""
+  quant: Optional[Quant] = None
+  matmul_precision: str = "default"
+
+  @nn.compact
+  def __call__(self, inputs: Array) -> Array:
+
+    def compute_dot_general(inputs, kernel, axis, contract_ind):
```
gobbleturk, line 206: is there a way we can share code with DenseGeneral? I assume this was copied from DenseGeneral?

RissyRan: Yes, I could refactor this out as a global helper function. Will update that

MaxText/configs/base.yml
```diff
@@ -147,6 +147,8 @@ shared_experts: 1
 routed_scaling_factor: 1.0 # scaling factor for routing scores
 routed_score_func: "" # scoring function for routing
 routed_bias: False # a flag if a bias term is added for routing
+n_routing_groups: 1 # number of groups for routing
+topk_routing_group: 1 # number of top groups to route inputs
```
gobbleturk, line 151: Is there a recommendation for how this interacts with EP? My guess is that this only can increase performance when topk_routing_groups < EP

Perhaps some comment here about performance tradeoff? 
"The activations only have to be sent to at most topk_routing_group different devices for EP, which can have performance benefits"

MaxText/layers/linears.py
```diff
@@ -173,7 +259,7 @@ def compute_dot_general(inputs, kernel, axis, contract_ind):
       )
       bias = jnp.asarray(bias, self.dtype)
       output += bias
-    return output
+    return output, pre_bias_logits
 
 
 class MlpBlock(nn.Module):
```
gobbleturk, line 265: I wonder if we can have a separate dense MlpBlock versus sparse... Don't have to fix in this PR....

MaxText/layers/linears.py
```diff
@@ -382,17 +481,56 @@ def deepseek_scale_weights(self, weights):
     weights *= self.config.routed_scaling_factor
     return weights
 
-  def permute(self, inputs, gate_logits):
-    """Permute tokens to group by expert to fit gmm call."""
+  def deepseek_routing(self, gate_logits, pre_bias_logits):
+    """DeepSeek routing logit. It involves two-stage selection process:
 
+    1) Group Scoring: Experts are partitioned into n_routing_groups.
```
gobbleturk, line 487: Great comment!

gobbleturk (changes requested):
Mostly LGTM! Can we add a unit test for the new logic for the new device limited `deepseek_routing`?

MaxText/configs/models/deepseek3-671b.yml
```diff
@@ -32,6 +32,8 @@ shared_experts: 1
 routed_scaling_factor: 2.5
 routed_score_func: "sigmoid"
 routed_bias: True
+n_routing_groups: 8
```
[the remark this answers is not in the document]
RissyRan: Yes, that's right! I will add a comment `adjustable to controlling communication costs in routing` for now, and will update it according to some benchmarks later this week (i.e. values better for TPUs).

MaxText/layers/linears.py
RissyRan: What will be the delta for sparseMlpBlock?

MaxText/layers/linears.py
```diff
@@ -382,17 +481,56 @@ def deepseek_scale_weights(self, weights):
     weights *= self.config.routed_scaling_factor
     return weights
 
-  def permute(self, inputs, gate_logits):
-    """Permute tokens to group by expert to fit gmm call."""
+  def deepseek_routing(self, gate_logits, pre_bias_logits):
```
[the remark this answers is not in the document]
RissyRan: Sure, will add.

MaxText/configs/base.yml
RissyRan: Yes, topk_routing_groups < EP, and it should improve the all_to_all communication. However, how GPU structure is different to TPU, so like you also mentioned, this config could be configuable, or we could even have an option to turn it off. I am planning to have some tests later this week. 

Will have the comment you mentioned above. Thanks!

MaxText/configs/models/deepseek3-671b.yml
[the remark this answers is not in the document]
gobbleturk: Can we instead remove these from /configs/model/deepseek? Otherwise allall users of deepseek are forced to use exactly these values - we have logic that enforces config set in /configs/models are not overriden by CLI, but users may want to override these particular `n_routing_groups` choices

See our logic for not overwriting the /config/model settings:
https://github.com/AI-Hypercomputer/maxtext/blob/948fb88eceb0a35167f43fa190436649e7b2522e/MaxText/pyconfig.py#L243-L249

MaxText/layers/linears.py
gobbleturk: like 400 lines of code or so lol

gobbleturk (approved):

PR #1497, opened 2025-03-31
Fix DeepSeek checkpoint conversion script

gagika:
# Description

1. Checkpoint conversion script update to match model name and python file name (some typos there).
2. Disabling "Running MoE sparse matmul implementation." logging in each forward pass, as it spams the logs

# Tests

Tested with with those changes.

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -1003,14 +1003,12 @@ class MoeBlock(nn.Module):
 
     w0_kernel, w1_kernel, wo_kernel = self.generate_kernels(cfg.num_experts, cfg.emb_dim, self.intermediate_dim)
     if cfg.sparse_matmul:
-      max_logging.log("Running MoE sparse matmul implementation.")
       if quantizations.in_serve_mode(self.quant):
         w0_kernel, w1_kernel, wo_kernel = self.retrieve_quantized_weight(
             inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel
         )
       return self.sparse_matmul(inputs, gate_logits, w0_kernel, w1_kernel, wo_kernel)
     else:
-      max_logging.log("Running MoE dense matmul implementation.")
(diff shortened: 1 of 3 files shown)


MaxText/layers/linears.py
```diff
@@ -1003,7 +1003,6 @@ def __call__(self, inputs):
 
     w0_kernel, w1_kernel, wo_kernel = self.generate_kernels(cfg.num_experts, cfg.emb_dim, self.intermediate_dim)
     if cfg.sparse_matmul:
-      max_logging.log("Running MoE sparse matmul implementation.")
```
RissyRan, line 1006: Let's keep it?

RissyRan (commented):
Thanks for the fix! Just one minor comment for the log, I guess it's not on purpose.

MaxText/layers/linears.py
gagika: this prints at every forward pass, so it can spam the logs (for un-scanned checkpoint it logs for all the layers).

As we print model config, this can be deducted from there right?

RissyRan: This log is added on purpose to compare to `Running MoE dense matmul implementation`.  But I agree we could add it somewhere that only occur once if this is too much (i.e. for each layer). Let's chat more on Monday.

richjames0 (approved):
lgtm

MaxText/layers/linears.py
gobbleturk: I also recommend removing it. Its currently can be figured out by the config which is printed, and we can print it explicitly only once if we really want, but right now it looks like ugly spam

vipannalla: +1, this gets printed for every prefill and decode op and really spammy. Can we atleast reduce the log-level you want to keep?

RissyRan (approved):

PR #1502, opened 2025-03-31
Update mmlu path

RissyRan:
# Description

Enable to run MMLU without changing the local Python path. This will be helpful to run on other GKE cluster, like xpk.

# Tests

Tested it locally as well as triggered a xpk workload. 

Error before the change:

```
  File "/home/ranran/maxtext/benchmarks/mmlu/mmlu_eval.py", line 48, in <module>
    import max_logging
ModuleNotFoundError: No module named 'max_logging'
```

I think it's missing python path `maxtext/MaxText` to find the max_logging.

# Checklist

### DIFF
--- a/benchmarks/mmlu/mmlu_eval.py
+++ b/benchmarks/mmlu/mmlu_eval.py
@@ -41,11 +41,19 @@ prompt_template='Example 1:\nQuestion: What is the capital of France?\nChoices:\
 import collections
 import re
 import sys
+import os
 
+current_dir = os.path.dirname(os.path.abspath(__file__))
+benchmark_parent_dir = os.path.dirname(current_dir)
+maxtext_parent_dir = os.path.dirname(benchmark_parent_dir)
+
+sys.path.append(maxtext_parent_dir)
+sys.path.append(maxtext_parent_dir + "/MaxText")
+
+import max_logging
 from absl import flags
 import datasets
 import jax
-import max_logging
 import max_utils
 import maxengine
 from mmlu_categories import categories


richjames0 (approved):
This feels a bit weird but I'm guessing we don't control the code that follows, that depends on the path, in which case I imagine this is the best we can do :)

bvandermoon (approved):
What was the error here?

RissyRan:
> What was the error here?

Sorry, I forgot to add the error, it will be like this:

Error before the change:

```
(maxtext) ranran@t1v-n-cf37f197-w-0:~/maxtext$ python3 benchmarks/mmlu/mmlu_eval.py
  File "/home/ranran/maxtext/benchmarks/mmlu/mmlu_eval.py", line 48, in <module>
    import max_logging
ModuleNotFoundError: No module named 'max_logging'
```

bvandermoon:
> > What was the error here?
> 
> Sorry, I forgot to add the error, it will be like this:
> 
> Error before the change:
> 
> ```
> (maxtext) ranran@t1v-n-cf37f197-w-0:~/maxtext$ python3 benchmarks/mmlu/mmlu_eval.py
> Traceback (most recent call last):
>   File "/home/ranran/maxtext/benchmarks/mmlu/mmlu_eval.py", line 48, in <module>
>     import max_logging
> ModuleNotFoundError: No module named 'max_logging'
> ```

Thanks @RissyRan. Looks like `max_logging` was also removed in this PR. Was there still an error after that?

PR #1505, opened 2025-04-01
Refactor chunked prefill using existing prefix

yuyanpeng-google:
* Use the existing prefix for general semantic mean and prevent multiple highly relevant args.
* Concatenate prefix KV cache instead of update in max prefill length.

The dynamic length of KV cache could produce more compiled graph. It's fine for now since the KV cache length always the same to the common prefix tokens. Need more improvement after benchmark with prefix caching.

also see https://github.com/AI-Hypercomputer/JetStream/pull/232

# Tests
Test on v6e-8
Using model=mixtral-8x7b, max_prefill_predict_length=8192, prefill_chunk_size=2048,
benchmark_chunked_prefill get similar latency 182 ms

Using model=mixtral-8x7b, max_prefill_predict_length=1024, max_target_length=2048, prefill_chunk_size=256, dataset=openrca
benchmark_serving get similar result between with and without chunked prefill.
rouge1 46.2

# Checklist

### DIFF
--- a/MaxText/benchmark_chunked_prefill.py
+++ b/MaxText/benchmark_chunked_prefill.py
@@ -24,7 +24,6 @@ import pyconfig
 
 from typing import Sequence
 from absl import app
-import jax.numpy as jnp
 import datetime
 from jetstream.engine import token_utils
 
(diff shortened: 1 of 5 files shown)


mailvijayasingh (approved):
Thanks for adding these, I have separately verified accuracy and perf in vij_opt_c_p branch and looks good.

mailvijayasingh (approved):

gpolovets1 (approved):

yuyanpeng-google:
Fix use wrong common prefix tokens in benchmark chunked prefill. It return back to similar latency before refactor.

MaxText/benchmark_chunked_prefill.py
```diff
@@ -46,50 +45,54 @@ def main(argv: Sequence[str]) -> None:
...
+  # set it to max complete prompt length that has to be benchmarked with chunked prefill
+  max_prefill_length = config.max_prefill_predict_length
+  tokens, _ = tokenizer_model.encode(text, is_bos=True, prefill_lengths=[max_prefill_length])
 
-  chunk_size = 2048
+  chunk_size = config.prefill_chunk_size
   # set this to array of acceptable chunk sizes
-  prefill_lengths = [1024, 2048, 4096, 7000]
-  # prefill_lengths = [1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144]
+  prefill_lengths = [1024, 2048, 4096, 8192]
```
vipannalla, line 54: Can you use this config instead -- https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L578 instead of hardcoding?

MaxText/inference/kvcache.py
```diff
@@ -351,21 +351,17 @@ def _get_ar_cache_vars(self, batch, key_heads, value_heads, key_head_size, value
...
-    if no pervious chunk is processed,
-    function returns a cache that has non zero first chunk part of key and value
-
-    else
-    function updates current key and value at the desired location and returns entire key and value
+  def kv_cache_prefill_concatenated(self, key: Array, value: Array, decoder_segment_ids: Array, previous_chunk: Any):
+    """Return KV cache concatenating to existing prefix."""
 
-    """
+    assert not self.kv_quant, "Not support kv_quant now."
```
vipannalla, line 357: What is blocker for kv cache quantization with chunked_prefill support? Can we get this working soon?

vipannalla (approved):
looks good, one minor comment and a question.

MaxText/inference/kvcache.py
mailvijayasingh: Before it was a blocker, as Qinwen's changes were in flight for quantized KV cache, but it can be done now.
The major difference is arounf Qtensors being updated and not jax tensors. It might also change some API types in Prefix cache API and not so much the API itself.

MaxText/benchmark_chunked_prefill.py
mailvijayasingh: cool, we can do that. Shall I follow up on that? Jetstream side is merged and that renders maxtext head broken with Jetstream head

vipannalla: Sounds good, a followup PR to address this one and kv cache quantization support is fine.

PR #1514, opened 2025-04-03
fix sequence_packing

aireenmei:
# Description
Fix a bug in https://github.com/AI-Hypercomputer/maxtext/pull/1416 that uses pad_id for axis. Also fix a pylint error.

# Tests

Manually tested

# Checklist

### DIFF
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -398,7 +398,7 @@ class MaxEngine(engine_api.Engine):
 
   @functools.partial(jax.jit, static_argnums=(0,), static_argnames=("request_id",))
   def _prefill_jit(
-      self, # pytype: disable=signature-mismatch
+      self,  # pytype: disable=signature-mismatch
       *,
       params: Params,
       existing_prefix: Optional[ExistingPrefix] = None,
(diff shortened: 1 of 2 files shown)


gobbleturk (approved):

MaxText/sequence_packing.py
```diff
@@ -171,7 +171,7 @@ def true_fn():
       for k in keys:
         new_seq = one_example[k][: key2length[k]]
         new_seq_len = tf.size(new_seq)
-        new_partial[k] = tf.concat([partial[k], new_seq], pad_id)
+        new_partial[k] = tf.concat([partial[k], new_seq], 0)
```
bvandermoon, line 174: What was the error you were seeing because of this?

MaxText/maxengine.py
```diff
@@ -510,7 +510,7 @@ def _prefill_jit(
 
   # Public non-JIT prefill method that updates page state
   def prefill(
-      self, # pytype: disable=signature-mismatch
+      self,  # pytype: disable=signature-mismatch
```
bvandermoon, line 513: Just for my understanding, what was the pylint error here? Are two spaces required between code/comment?

bvandermoon (approved):

MaxText/maxengine.py
aireenmei: This was not my change, but pylint failed on this

MaxText/sequence_packing.py
aireenmei: I misunderstood this function in my previous PR. tf.concat is expecting an axis number here. So for deepseek tokenizer that defines pad_id as 100001, you got this error: "ValueError: Shape must be at least rank 100002 but is rank 1"

PR #1533, opened 2025-04-07
Fix the dataloading sharding, which causes Tensor Parallelism issue on GPU.

wang2yn84:
# Description

When we enable Tensor Parallelism on GPU normally the workload hang or corrupted with TransformerEngine errors. The root cause is the data sharding is incorrect and the decoder_segment_id will be sharded by (fsdp, tensor) on (Batch, Sequence). This PR fixed the data sharding of both synthetic data loader and the actual data loader. 

Fixes: b/384441280

# Tests

Manually tested the workload with 1 host + TP with synthetic dataloader. Relying on the presubmit test for the actual dataloader.

# Checklist

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -373,21 +373,27 @@ def get_coordinator_ip_address():
   return coordinator_ip_address
 
 
+def get_unspecified_mesh_axes_value(parallelism_vals, target_product, parallelism_type):
+  """Evaluates unspecified DCN/ICI parallelism values"""
+  assert (
+      parallelism_vals.count(-1) == 1
+  ), f"Found unspecified values (-1) for more than one {parallelism_type}\
+    parallelism axis. At most one axis can be unspecified."
+
+  determined_val = target_product / np.prod(parallelism_vals) * -1
+
+  assert (
+      determined_val >= 1 and determined_val.is_integer
+  ), f"Unspecified value unable to be determined with the given\
+    {parallelism_type} parallelism values"
+
+  return int(determined_val)
+
+
 def fill_unspecified_mesh_axes(parallelism_vals, target_product, parallelism_type):
   """Evaluates unspecified DCN/ICI parallelism values"""
   if -1 in parallelism_vals:
-    assert (
-        parallelism_vals.count(-1) == 1
-    ), f"Found unspecified values (-1) for more than one {parallelism_type}\
-      parallelism axis. At most one axis can be unspecified."
-
-    determined_val = target_product / np.prod(parallelism_vals) * -1
-
-    assert (
-        determined_val >= 1 and determined_val.is_integer
-    ), f"Unspecified value unable to be determined with the given\
-      {parallelism_type} parallelism values"
-
+    determined_val = get_unspecified_mesh_axes_value(parallelism_vals, target_product, parallelism_type)
     parallelism_vals[parallelism_vals.index(-1)] = int(determined_val)
 
   target_type = "slices" if parallelism_type == "DCN" else "devices per slice"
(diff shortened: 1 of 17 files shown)


MaxText/layers/gemma3.py
```diff
@@ -93,11 +93,7 @@ def __call__(self, inputs, train=False):
     cfg = self.config
     print(f"Gemma3VisionEncoderLayer input: {inputs.shape}")
     x = inputs.squeeze(axis=1)
-    x = nn.Conv(features=1152,
```
gobbleturk, line 96: lets try to avoid lint changes in untouched files

you probably just ran bash code_style.sh - unfortunately our linter exceptions are chaning rapidly so this probably doesn't work - touching more files than just the ones you changed

MaxText/max_utils.py
```diff
@@ -780,6 +780,16 @@ def reorder_causal_load_balanced(batch, cp_size):
   }
 
 
+def shard_reorder_causal_load_balanced(batch, cp_size):
```
gobbleturk, line 783: is this moved from somewhere?

MaxText/multihost_dataloading.py
```diff
@@ -27,46 +27,102 @@
...
+
+  return (batch_product, sequence_product)
+
+
+def _form_global_array(
+    path,
+    array: np.ndarray,
+    global_mesh: Mesh,
+    input_data_shardings: NamedSharding,
+    input_data_sharding_factor: tuple[int, ...],  # e.g., (4, 1)
```
gobbleturk, line 80: can we make this variable name plural because its a tuple, e.g. 

input_data_sharding_factors

MaxText/multihost_dataloading.py
```diff
@@ -27,46 +27,102 @@
...
+
+  global_shape = _build_global_shape(array.shape)
+
+  if len(input_data_sharding_factor) != array.ndim:
+    raise ValueError(
+        f"Sharding factor {input_data_sharding_factor} must match array rank {array.ndim} "
+        f"for array of shape {array.shape}"
+    )
+
+  def recursive_split(arr, factors, axis=0):
```
gobbleturk, line 92: why is this function needed? Can you include comments? Is it because we are now sharding on both axies [batch, sequence] whereas previously we were only sharding on batch?

gobbleturk (approved):
Generally LGTM, we should try to avoid unrelated whitespace and pylint changes (probably code_style.sh is broken right now, sorry about that)

Please do have either @khatwanimohit or @aireenmei review for the data loader change as well

MaxText/max_utils.py
wang2yn84: No, I created this one. The reason why I add this function is output sharding for reorder_causal_load_balanced() is not specified and compiler overwrite it with batch only sharding.

MaxText/multihost_dataloading.py
wang2yn84: Sure, changed to input_data_sharding_prods. Thank you!

MaxText/multihost_dataloading.py
wang2yn84: Add the comment, yes it's because we are now considering 2D sharding. Previously only sharding on batch is considered.

MaxText/multihost_dataloading.py
```diff
@@ -27,46 +27,102 @@
...
+  except AttributeError:
     raise ValueError(
-        f"Unable to put to devices shape {array.shape} with "
-        f"local device count {len(global_mesh.local_devices)} "
-        f"at {jtu.keystr(path)}"
-    ) from array_split_error
+        f"input_data_shardings does not have a .spec attribute (expected a PartitionSpec), got: {type(input_data_shardings)}"
+    )
+
+  def compute_product(axis_spec):
```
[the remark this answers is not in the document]
wang2yn84: Sounds good, moved to max_utils. Thank you!

MaxText/layers/gemma3.py
wang2yn84: Gotcha. Reversed!

MaxText/multihost_dataloading.py
aireenmei, line 1: Could you make similar changes to the RemoteIterator class? This class is used with pathway, for a distributed data loading behavior similar to McJax, instead of loading through the controller.

aireenmei (changes requested):

MaxText/multihost_dataloading.py
wang2yn84: Hi Aireen, there is no test for RemoteIterator at all, it's very risky to touch the function without the protection of unit tests. Would you like to add a test for that first?

aireenmei: OK, let me add some tests, then we can apply similar changes.

aireenmei (approved):

PR #1541, opened 2025-04-08
Change the FP8 dot from fake quantization to direct quantization op

serach24:
Currently the FP8 dot in maxtext is using `Fp8DotGeneralOp` which is a fake quantization pattern for FP8 dot product. This pattern is not trivial to capture in the XLA compiler and could silently fall back to high precision if the pattern was broken.
The preferred API for FP8 dot product is now changed to [`Fp8DirectDotGeneralOp`](https://github.com/google/flax/blob/e58aeea9286f891b2f210dc84363986e93fdb255/flax/linen/fp8_ops.py#L642) in flax, which ensures the FP8 GEMM is emitted during the XLA compilation. 

### DIFF
--- a/MaxText/layers/quantizations.py
+++ b/MaxText/layers/quantizations.py
@@ -204,7 +204,7 @@ class Fp8Quantization(Quantization):
 
   def dot_general_cls(self, mesh_axes: Tuple[str, ...] = ()):
     """Returns dot_general configured with aqt params."""
-    return nn.Fp8DotGeneralOp
+    return nn.Fp8DirectDotGeneralOp
 
   def einsum(self, dtype: DType = jnp.float32):
     return Fp8Einsum(dtype=dtype)


bvandermoon (approved):
Thanks for updating this. Is there any documentation on the difference between the two? Or is it just the code?

serach24:
For the PR to maxtext, there is just the code, as it is internal use and is not exposed to users. 
For the reference, the doc between the two is sent to flax [here](https://github.com/google/flax/pull/4686/files)
>  ## Deprecated APIs
> Previously, we provided APIs like `fp8_ops.quantize_dequantize` for current
> scaling and `fp8_ops.[in|out]_qdq` for delayed scaling. These were used with
> high precision dot operations, leveraging an XLA-FP8 feature that
> pattern-matched QDQ->dot sequences to Q->fp8_cublas_gemm. The corresponding
> high-level API was called `fp8_ops.Fp8DotGeneralOp`. However, this pattern
> matching-based solution proved brittle, as the patterns could be easily broken
> by other XLA optimizations. We recommend users migrate from these deprecated
> APIs to the newer ones described above.
> 
> For migration, users should replace:
> * `fp8_ops.quantize_dequantize -> jnp.dot` with `fp8_ops.quantize -> jnp.dot ->
>   fp8_ops.dequantize`
> * `fp8_ops.in_qdq -> jnp.dot -> fp8_ops.out_qdq` with `fp8_ops.in_q -> jnp.dot
>   -> fp8_ops.out_dq`
> * `fp8_ops.Fp8DotGeneralOp` with `fp8_ops.Fp8DotGeneral`
> 
> Additionally, we provide an einsum variant through `fp8_ops.Fp8Einsum`.

khatwanimohit (approved):

PR #1545, opened 2025-04-08
Llama 4 chunked attention

gagika:
# Description

Llama 4 has [Chunked Attention](https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/models/llama4/modeling_llama4.py#L828) which we need to support in MaxText.

create_chunked_attention_mask tokens (0, K), (K, 2K), (2K, 3K) attend to each other when doing local chunked attention.

# Tests
Added Tests for chunked attention

# Checklist

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -17,6 +17,7 @@
 import enum
 import functools
 import math
+import numpy as np
 from typing import Any, Optional, Tuple
 
 from MaxText import common_types
@@ -49,6 +50,7 @@ from MaxText.layers import quantizations
 class AttentionType(enum.Enum):
   GLOBAL = "global"
   LOCAL_SLIDING = "local_sliding"
+  CHUNK = "chunk"
   MLA = "mla"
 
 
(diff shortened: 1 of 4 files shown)


MaxText/layers/attentions.py
```diff
@@ -225,6 +296,15 @@ def generate_attention_mask(
...
       output_mask = sliding_mask * output_mask
+    elif self.attention_type == AttentionType.CHUNK and output_mask is not None:
+      if self.chunk_window_size is None:
+        raise ValueError("chunk_window_size must be set if chunk attention type")
+
+      # chunk mask calculation - similar to pytorch reference
+      block_pos = jnp.abs((row_ids // self.chunk_window_size) - (col_ids // self.chunk_window_size))
+      token_pos = row_ids - col_ids
+      chunk_mask = (block_pos == 0) & (token_pos >= 0)
+      output_mask = chunk_mask * output_mask
```
gobbleturk, line 307: For my understanding it looks like `generate_attention_mask` won't be called for flash attention on TPUs, instead it will use your ChunkedCausalMask - this class exists so that the mask can be used without ever instantiating it to avoid the seq^2 memory? If this is the case can we add a comment for this somewhere - perhaps both in the generate_attention_mask and also in ChunkedCausalMask?

MaxText/layers/attentions.py
```diff
@@ -225,6 +296,15 @@ def generate_attention_mask(
       all_ones = jnp.ones_like(output_mask)
       sliding_mask = jnp.triu(all_ones, -1 * self.sliding_window_size + 1) * jnp.tril(all_ones, self.sliding_window_size - 1)
       output_mask = sliding_mask * output_mask
+    elif self.attention_type == AttentionType.CHUNK and output_mask is not None:
```
gobbleturk, line 299: nit: possibly move this code under a method "generate_chunk_attention_mask" for clearer modularity / readability. This is a stylistic nit though, its fine as it is

MaxText/tests/attention_test.py
gobbleturk, line 1: Awesome tests!

gobbleturk (approved):
LGTM, thanks Gagik! Two small nits, feel free to ignore

MaxText/layers/attentions.py
gagika: Thanks Matt for the comment, yes your understanding is correct, I added comments in both places, please take a look.

MaxText/layers/attentions.py
gagika: Good suggestion, done, and I could also add it to my tests, please have a look.

MaxText/layers/attentions.py
```diff
@@ -141,6 +143,108 @@ def apply_mask_to_logits(logits: Array, mask: Array):
...
+
+  chunk_size: int
+
+  def __init__(
+      self,
+      shape: tuple[int, int],
+      chunk_size: int,
+      shard_count: int = 1,
+  ):
+    if chunk_size <= 0:
```
RissyRan, line 169: Let's also check it in the pyconfig.py for early detection for config setting?

RissyRan (approved):
LGTM! Do we expect other changes for inference? Or this should be sufficient for 10m context?

gagika:
> LGTM! Do we expect other changes for inference? Or this should be sufficient for 10m context?

For long context support, the inference team needs to leverage "chunk" attention to keep KV cache small, e.g  <= chunk_size for those layers. cc @vipannalla

PR #1547, opened 2025-04-08
Improved Paged Attention

patemotter:
# Description

* Improved version of PageManager and PageState
* Corresponding unit tests for PageManager
* Should not have any noticeable impact existing non-paged code.  

# Tests

page_manager_test.py
interleaved version of decode.py

# Checklist


MaxText/configs/base.yml
```diff
@@ -652,6 +652,8 @@ sa_v_layout: "HEAD_DIM_MINOR"
 pagedattn_num_pages: 64
 pagedattn_tokens_per_page: 32
 pagedattn_pages_per_compute_block: 8
+pagedattn_max_page_groups: -1  # defaults to global batch size
```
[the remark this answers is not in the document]
patemotter: It seemed like a plausible scenario where a user may not always want them exactly matching 1:1. But we are defaulting to using the global batch size since that would be the most likely use case.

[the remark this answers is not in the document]
liurupeng: I see, we should be conscious about adding a new field since in production use case, it will be hard to deprecate it if it's used by customers. Could share the exact scenario that customers would need it?

MaxText/maxengine.py
```diff
@@ -1224,6 +1231,14 @@ def copy(path, partial_cache, full_cache, annotations):
         "tokens": inserted_tokens,
     }
 
+  def release_pages(self, slot: int):
```
liurupeng, line 1234: is this used somewhere?

patemotter: Yes, in `decode_multi.py`.

MaxText/configs/base.yml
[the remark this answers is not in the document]
patemotter: Tying it directly to the batch size might make sense for the basic circumstances (which is why it's the default). But in other cases like beam search you would want `pagedattn_max_page_groups = num_beams * global_batch_size`. To me it does not seem like an unreasonable thing to add this into a config file that already has approximately 300 variables.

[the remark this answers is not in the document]
liurupeng: I see, but in the beam search case, you wouldn't need a config for `pagedattn_max_page_groups` right? num_beams would be the configurable flag. (I think we will need to clean up these variables in future if we want to make the code quality is good)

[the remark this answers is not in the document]
liurupeng: let's not add this flag for now, only if must needed

MaxText/decode_multi.py
```diff
@@ -0,0 +1,218 @@
...
+          if slot_idx < sampled_tokens.data.shape[0]:
+            token_for_slot = sampled_tokens.get_result_at_slot(slot_idx).tokens.item()
+            streams_results[slot_idx].append(token_for_slot)
+          else:
+            print(f"Warning: Tried to get token for slot {slot_idx}, but batch size seems smaller.")
+
+        # Call release_pages if finished this step
+        if finished_this_step:
+          print(f"    Calling engine to release pages for finished slot {slot_idx}...")
+          engine.release_pages(slot=slot_idx)
```
liurupeng, line 150: qq, why do we need to manually release the pages. My understanding is: In prefill stage, when we allocate this slot to for the request, we will firstly release the pages for this slot. So we don't need to worry about clean up here?

patemotter: 1. This is just a demo script showing how it _can_ be used so it's not required to be used this way. 
2. There are still valid cases for manually releasing pages via the engine api. I.e. to return pages to the pool if there is an error at the orchestration level without needing to wait for a prefill to be called.

MaxText/configs/base.yml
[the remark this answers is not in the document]
patemotter: I understand the concern about adding potentially unnecessary flags. While I still see value in decoupling this for flexibility in certain cases, I'm willing to remove it if other reviewers agree it should not be added at this time.

vipannalla:
I took a quick pass, code looks good at a high level. I don't have context into nitty-gritty of paged attention logic. I'll leave it to @wyzhang and @liurupeng to verify that.

MaxText/configs/base.yml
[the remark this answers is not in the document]
vipannalla: As much as I like more flexible code, I'd agree with Rupeng. If we don't have a use case right now for this config, lets not add it. In future, if this is asked for a customer it shouldn't be too hard to add it.

liurupeng (approved):

MaxText/configs/base.yml
```diff
@@ -323,8 +323,8 @@ logical_axis_rules: [
                       ['cache_kv', []],
                       ['cache_sequence', []],
                       ['exp', 'expert'],
-                      ['paged_kv_heads', []],
-                      ['num_pages', ['tensor']],
+                      ['paged_kv_heads', ['tensor']],
```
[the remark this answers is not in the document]
wyzhang: Noticed this. Didn't get a chance to make a PR to fix

MaxText/configs/base.yml
```diff
@@ -654,14 +654,13 @@ sa_v_layout: "HEAD_DIM_MINOR"
...
-#######################
 ### Paged Attention ###
-#######################
 # These settings take effect only when `attention=paged`.
 # They should be adjusted based on the available HBM and model config.
 pagedattn_num_pages: 64
 pagedattn_tokens_per_page: 32
-pagedattn_pages_per_compute_block: 8
+pagedattn_pages_per_compute_block: 4
+pagedattn_max_pages_per_group: -1  # defaults to (max_target_length + pa_tokens_per_page - 1) // pa_tokens_per_page
```
wyzhang, line 663: nit: An explicit comment saying default is one group for all page, this saves reader from parsing and understanding the equation. Fine to keep as is.

MaxText/decode_multi.py
```diff
@@ -0,0 +1,212 @@
+# Copyright 2025 Google LLC
```
wyzhang, line 1: Consider moving this to MaxText/inference/decode_mutil.py, since this file is inference specific. We can gradually move inference specific things to inference subdir.

patemotter: Should the existing decode.py also be moved then?

patemotter: That's what I was concerned about. So given that does it still make sense to move `decode_multi.py` to the inference folder or keep it in the same space as `decode.py`?

MaxText/configs/base.yml
wyzhang: Yes.

Correct. I meant one page group per max target seq

MaxText/decode_multi.py
wyzhang: I lean toward moving this to inference subdir. But fine if you decide to keep it here

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
-AxisNames = common_types.AxisNames
 Config = common_types.Config
 
+# Aliases using <Dims><Type><Rank>d convention
+# We use string names for dimensions as they are symbolic within the type hints.
+PagesInt1d = Integer[Array, "num_pages"]
+GroupsPagesInt2d = Integer[Array, "max_page_groups max_pages_per_group"]
+GroupsInt1d = Integer[Array, "max_page_groups"]
+GroupsBool1d = Bool[Array, "max_page_groups"]
+ScalarInt = Shaped[Integer[Array, ""], ""]
```
wyzhang, line 41: nit: This can be simply Integer[Array, ""]. Same for below

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+  active_page: GroupsInt1d
+  has_active_page: GroupsBool1d
+  active_page_position: GroupsInt1d
 
 
 def initialize_page_state(
     num_pages: int,
-    num_slots: int,
-    max_pages_per_slot: int,
+    max_page_groups: int,
```
wyzhang, line 92: Curious can we use batch_size for max_page_groups?
concurrent number of sequences or requests is essentially the generate batch size.

MaxText/decode_multi.py
wyzhang: Okay to keep as is in this PR and move in a follow up PR.

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+    max_page_groups: The maximum number of page groups (concurrent sequences/requests)
+      the system can track.
+    max_pages_per_group: The maximum number of pages that can be allocated to
+      a single page group (determines the size of the second dimension of `page_map`).
+
+  Returns:
+    An initialized `PageState` object with all values set to their defaults (zeros/False).
+  """
+  initial_page_status = jnp.zeros((num_pages,), dtype=jnp.int32)
+  initial_page_status = initial_page_status.at[0].set(1)  # Workaround page 0
```
wyzhang, line 113: Maybe add a TODO for any issue with using page 0, so we can follow up and have it fixed.

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+  Args:
+    page_status: A 1D `jnp.ndarray` representing the global status of pages
+      (0 for free, 1 for allocated). Should have shape [num_pages].
+
+  Returns:
+    A scalar `jnp.int32` array containing the index of the next free page
+    (the lowest index >= 1 where `page_status` is 0).
+    Returns -1 if no free pages (at index >= 1) are found.
+  """
+  search_status = page_status[1:]
```
wyzhang, line 140: Same here. Add a TODO gives some context on what's the issue with page 0, so we don't forget to resolve

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+  overall_free_mask = search_status == 0
+  next_free_relative = jnp.argmax(overall_free_mask)
+  next_free_overall = next_free_relative + 1
+  has_free_overall = jnp.any(overall_free_mask)
+  return jnp.where(has_free_overall, next_free_overall, -1)
+
+
+def _release_pages_for_group(
+    page_state: PageState,
+    page_group_id: int,
```
wyzhang, line 150: Could we use Integer[Array, ""] rather than int?

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
-      slot, page_map, page_status, current_page = state
-      page_idx = jnp.where((page_status[1:] == 0), size=1)[0][0] + 1
-      page_status = page_status.at[page_idx].set(1)
-      page_map = page_map.at[slot, i].set(page_idx)
-      current_page = current_page.at[slot].set(page_idx)
-      return slot, page_map, page_status, current_page
-
-    _, page_map, page_status, current_page = jax.lax.fori_loop(
-        0, prefill_slot_num_pages, _reserve_page, (slot, page_map, page_status, current_page)
+def _update_prefill_pages_for_group(
```
wyzhang, line 196: nit: consider naming it release_and_reserve_for_group to be explicit.

MaxText/decode_multi.py
patemotter: Will add a todo for it and move it in future PR.

MaxText/inference/page_manager.py
patemotter: Ah thanks, will change.

MaxText/inference/page_manager.py
patemotter: Will add.

MaxText/inference/page_manager.py
patemotter: Will add.

MaxText/inference/page_manager.py
patemotter: Probably, I will try this and check the results.

MaxText/inference/page_manager.py
patemotter: This is similar to one of Rupeng's earlier comments, which I addressed by removing the config variable `pagedattn_max_page_groups` from `base.yml`. So now inside `pyconfig.py` `max_page_groups = global_batch_size_to_load`. 

It could be made cleaner by removing the variable assignment in `pyconfig.py` and changing the line in `PageManager.__init__`

from 
`self.max_page_groups: int = config.pagedattn_max_page_groups`
to 
`self.max_page_groups: int = config.global_batch_size_to_load`.

MaxText/inference/page_manager.py
patemotter: I will rename it. 

We would still need to call a release somewhere before doing the reserve in order to ensure we are keeping the PageState accurate. Can you explain a bit more about what you have in mind for this?

wyzhang: From API design aspect, it is more natural to me to have two apis. One for reserve and the other for release. As we have a release_page API above, I expect this API to be only perform reservation. 

Alternatively we can have 3 APIs, one for reserve, one for release and another one for release_and_reserve. 

IMHO, it looks to me okay to require callsites to invoke release first and then reserve, thus no need for an API/method to do release_and_reserve. 

This is nitpicking. Feel free to leave as is if you feel strongly about doing this week

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+          has_active_page=released_state.has_active_page.at[page_group_id].set(True),
+          active_page_position=released_state.active_page_position.at[page_group_id].set(last_page_position),
+      )
+
+    if has_enough_resources:
+      return allocate_and_update_state((current_page_status, current_page_map, current_num_pages_used))
+    else:
+      return released_state
+
+  if true_length == 0:
```
wyzhang, line 291: Use python native condition when the value is known at compile time. 

In this case, true length is known only at computation time, so we may want to use jax.lax.cond

patemotter: Good catch, thought I had caught all of these. Will change.

MaxText/inference/page_manager.py
```diff
@@ -1,232 +1,582 @@
...
+    return handle_zero_length(None)
```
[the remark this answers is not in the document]
patemotter: I believe this was originally related to handling the init phase, but I will re-examine if this is still needed in the current version of the code. At a glance it looks like it might not be needed anymore, but I will double check.

MaxText/inference/page_manager.py
patemotter: I will change the name, but leave the functionality as-is for now in order to finish the PR. I will add a todo and fix this in the next round of cleanup.

wyzhang (approved):

MaxText/decode_multi.py
patemotter: Went ahead and moved it in this PR since it was simple.

MaxText/inference/page_manager.py
patemotter: I went ahead and changed this to have individual release and reserve page functions and then left the `release_and_reserve` to call both.

PR #1563, opened 2025-04-10
Update per-device batch size to 190, use inference.yml for inference by default, and use relative paths benchmarking scripts 

dli357:
Changes in this PR:
 - Change per device batch size to 190
 - Execute microbenchmark by module
 - Fix working directory module pathing for mlperf

# Description

Current H100 tests can support per-device batch size of 190, so we update the default to be 190 for current highest performance pre https://github.com/jax-ml/jax/issues/27817.

# Tests

Ran the tests and was able to confirm best performance using container from 2025-03-25.

# Checklist


wang2yn84 (approved):
Thank you!

wyzhang (approved):

PR #1571, opened 2025-04-12
Move MaxText related util functions to maxtext_utils.py

wang2yn84:
# Description

Right now there are quite a few MaxText related functions sits in max_utils.py that makes it hard to avoid the circular dependency, which is exactly what I'm running into in PR#1533. I'm sending this PR to clean up the util file first. 

What I did:
For 3P, we need to move all the functions in max_utils that referencing other MaxText functions (depend on MaxText imports) to maxtext_utils
For 1P, we need to move functions in max_utils that referencing MaxText targets to maxtext_utils

Principle:
In the future, all the functions in max_utils should not depend on any other MaxText functions or objects, with very few exceptions such as max_logging, common_types, etc.

# Tests

Rely on the presubmit tests.

# Checklist


gobbleturk (approved):
Thanks Lance!

vipannalla (approved):
looks good to me

bvandermoon (approved):

PR #1598, opened 2025-04-16
Warn when invoked as a script

richjames0:
# Description

We recently changed the MaxText API such that invoking e.g. `MaxText/train.py` will no longer work (`MaxText.train` needs to be invoked as a module. This change adds a warning when the user tries to invoke in the old manner (vs. a cryptic import error message, without this change)

# Tests

- Tested get warning: `python3 MaxText/train.py MaxText/configs/base.yml   run_name=rjx-run-v5p-8-b   base_output_directory=gs://rjx-storage   dataset_type=synthetic`
- Tested works: `python3 -m MaxText.train MaxText/configs/base.yml   run_name=rjx-run-v5p-8-b   base_output_directory=gs://rjx-storage   dataset_type=synthetic`
- Tested works (imports from train.py): `python3 -m MaxText.sft_trainer MaxText/configs/base.yml   run_name=rjx-run-v5p-8-b   base_output_directory=gs://rjx-storage   dataset_type=synthetic`

I believe this covers all cases. I might try to capture these in unit tests but want to get this change in ASAP.

# Checklist

### DIFF
--- a/MaxText/tests/module_tests.py
+++ b/MaxText/tests/module_tests.py
@@ -1,0 +1,53 @@
+"""
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Tests for train.py with various configs"""
+
+import os
+import sys
+import subprocess
+import unittest
+import pytest
+from absl.testing import absltest
+
+from MaxText.globals import PKG_DIR
+
+class ModuleTests(unittest.TestCase):
+  """Tests train.py with various invokation methods"""
+
+  def test_get_informative_error(self):
+    # cwd is MaxText so step up a level
+    os.chdir("..")
+    
+    command = [
+        sys.executable,  # use the same interpreter instance
+        "MaxText/train.py",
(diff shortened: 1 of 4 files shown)


MaxText/tests/module_tests.py
```diff
@@ -0,0 +1,50 @@
...
+        sys.executable,  # use the same interpreter instance
+        "MaxText/train.py",
+        "MaxText/configs/base.yml",
+        "run_name=maxtext-module-test",
+        "base_output_directory=gs://does-not-exist",
+        "dataset_type=synthetic"
+    ]
+
+    with self.assertRaises(subprocess.CalledProcessError) as context:
+      subprocess.run(command, check=True, text=True, capture_output=True)
```
gobbleturk, line 42: We are trying to avoid subprocess in our unit tests, they create strange race conditions with pytest and libtpu locks. Is there a way to re-write without subprocess? This one might be harder since you can't import main of train....

shralex: Yes, if this specific command may contend on TPU with other commands. Maybe for the purpose of what the test is checking you don't need to invoke train.py but can do something simple ?

shralex: Ah sorry this will just error out without using TPU! maybe its ok then ?

MaxText/tests/module_tests.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+"""Tests for train.py with various configs"""
+
+import os
+import sys
+import subprocess
+import unittest
+import pytest
+from absl.testing import absltest
```
shralex, line 24: nit: is this needed ? many tests just use unittest

shralex (approved):

MaxText/tests/module_tests.py
gobbleturk: ya I think this one is okay

gobbleturk (approved):

bvandermoon (approved):

PR #1600, opened 2025-04-16
Include venv in getting started steps

richjames0:
# Description

In some configurations installing in the default Python env fails so here we instruct users to create and work in a venv.

# Tests

Manually

# Checklist

### DIFF
--- a/getting_started/First_run.md
+++ b/getting_started/First_run.md
@@ -17,6 +17,8 @@ multiple hosts.
 2. Clone MaxText onto that TPUVM.
 3. Within the root directory of that `git` repo, install dependencies and pre-commit hook by running:
 ```
+python3 -m venv ~/venv-maxtext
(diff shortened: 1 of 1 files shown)


gobbleturk (approved):
yay thank you!

gobbleturk:
cc @bvandermoon rather important documentation change, although this is probably not included in tpu-recipes

bvandermoon (approved):

bvandermoon:
> cc @bvandermoon rather important documentation change, although this is probably not included in tpu-recipes

Thanks @gobbleturk. I put this as an optional step directly in the tpu-recipe [MAXTEXT_README](https://github.com/AI-Hypercomputer/tpu-recipes/blob/main/training/trillium/MAXTEXT_README.md)

But yeah this commit is after the release tag, so it won't be included for the recipes we are publishing unfortunately

PR #1608, opened 2025-04-18
Add multimodal_utils and support multimodal in decode.py

hengtaoguo:
# Description

Add multimodal support for decode.py. Add image preprocessing functions for Gemma3, especially used for decode.py. 

1. Add local image read function and resize/normalize for preprocessing.
2. Allow the loaded image to be passed through the Conv layer in Gemma3 vision encoder. Add conditional logic that if no images are passed, the vision_encoder branch won't be used.
3. Update prefill, insert and generate logics in maxengine.

FIXES: [b/407783855](https://b.corp.google.com/issues/407783855)

# Tests

Locally tested for both use_multimodal=True/False.

`python -m MaxText.decode MaxText/configs/base.yml model_name=gemma3-4b tokenizer_path=assets/tokenizer.gemma3 load_parameters_path=gs://hengtaoguo-maxtext-logs/checkpoints/gemma3-4b/vit/unscanned/checkpoints/0/items per_device_batch_size=1 run_name=runner_2025-04-18-18-20-33 max_prefill_predict_length=8 max_target_length=16 dataset_type=synthetic steps=1 async_checkpointing=false scan_layers=false prompt=\'I\ love\ to\' use_multimodal=True image_path=\'/home/hengtaoguo/projects/dog.jpg\'`

Compared with HF Gemma3 implementation. Loaded images and feature maps after Conv are matching:
image

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -681,3 +681,4 @@ nope_layer_interval: -1
 # Multimodal flags
 use_multimodal: False
 image_size_for_vit: 896 # Default for Gemma3, and should be overwritten by model's config
+image_path: "" # Local image path used for decoding
--- a/MaxText/layers/gemma3.py
+++ b/MaxText/layers/gemma3.py
@@ -90,14 +90,10 @@ class Gemma3VisionEncoderLayer(nn.Module):
     Returns:
       jnp.array for image embeddings, shaped [B, N, P, D], e.g. [4, 1, 256, 2560]
     """
-    cfg = self.config
-    print(f"Gemma3VisionEncoderLayer input: {inputs.shape}")
-    x = inputs.squeeze(axis=1)
-    x = nn.Conv(features=1152,
-        kernel_size=(14, 14),
-        strides=14,
-        padding="VALID",
-        name="embedding")(x)
+    b, n, h, w, c = inputs.shape
+    x = jnp.reshape(inputs, [b * n, h, w, c])
+    x = nn.Conv(features=1152, kernel_size=(14, 14), strides=14, padding="VALID", name="embedding")(x)
+    jax.debug.print("x after: {}", x.mean())
     n, h, w, c = x.shape
     x = jnp.reshape(x, [n, h * w, c])
     # TODO(hengtaoguo): finish the ViT with posemb, dropout and transformation layers.
--- a/MaxText/maxengine.py
+++ b/MaxText/maxengine.py
@@ -408,6 +408,7 @@ class MaxEngine(engine_api.Engine):
       params: Params,
       existing_prefix: Optional[ExistingPrefix] = None,
       padded_tokens: jax.Array,
+      images: Optional[jax.Array] = None,
       true_length: int,
       sampler: Optional[Callable[[Any], Any]] = None,  # pylint: disable=unused-argument
       rng: Optional[PRNGKeyType] = None,
@@ -448,6 +449,9 @@ class MaxEngine(engine_api.Engine):
     input_tokens = jnp.expand_dims(padded_tokens, 0)  # [BATCH, SEQUENCE]
     positions = jnp.expand_dims(jnp.arange(start_position, start_position + input_tokens.shape[1]), 0)
 
+    if self.config.use_multimodal:
+      input_images = images[jnp.newaxis, jnp.newaxis, ...]  # Add batch and sequence dimension [B, N, H, W, C]
+
     # sequence_indicator will be concatenated to existing_prefix decoder_segment_ids
     start_to_n = jnp.arange(start_position, start_position + input_tokens.shape[1])
     ones_to_keep = start_to_n < full_true_length
@@ -460,6 +464,7 @@ class MaxEngine(engine_api.Engine):
           input_params,
           input_tokens,
           positions,
+          encoder_images=input_images if self.config.use_multimodal else None,
           decoder_segment_ids=sequence_indicator,
           enable_dropout=False,
           model_mode=common_types.MODEL_MODE_PREFILL,
@@ -520,6 +525,7 @@ class MaxEngine(engine_api.Engine):
       params: Params,
       existing_prefix: Optional[ExistingPrefix] = None,
       padded_tokens: jax.Array,
+      images: Optional[jax.Array] = None,
       true_length: int,
       sampler: Optional[Callable[[Any], Any]] = None,  # pylint: disable=unused-argument
       rng: Optional[PRNGKeyType] = None,
@@ -540,6 +546,7 @@ class MaxEngine(engine_api.Engine):
         params=params,
         existing_prefix=existing_prefix,
         padded_tokens=padded_tokens,
+        images=images,
         sampler=sampler,
         true_length=true_length,
         page_state=self.page_state,  # Pass current page state
@@ -1284,10 +1291,20 @@ class MaxEngine(engine_api.Engine):
           (int(self.config.per_device_batch_size * jax.device_count()), 1),
           dtype=jnp.int32,
       )
+      dummy_image = jnp.ones(
+          (
+              int(self.config.per_device_batch_size * jax.device_count()),
+              maxtext_utils.NUM_IMAGES_PER_SEQUENCE,
+              self.config.image_size_for_vit,
+              self.config.image_size_for_vit,
+              maxtext_utils.NUM_IMAGE_CHANNELS,
+          ),
+      )
       _, cache = self.model.apply(
           abstract_params,
           x,
           x,
+          encoder_images=dummy_image if self.config.use_multimodal else None,
           enable_dropout=False,
           model_mode=common_types.MODEL_MODE_AUTOREGRESSIVE,
           rngs={"params": rng},
--- a/MaxText/maxtext_utils.py
+++ b/MaxText/maxtext_utils.py
@@ -45,6 +45,10 @@ from jax.experimental import mesh_utils
 
 OVERWRITE_WITH_GRADIENT = "_overwrite_with_gradient"
 
+# Multimodal constants
+NUM_IMAGES_PER_SEQUENCE = 1
+NUM_IMAGE_CHANNELS = 3
+
 
 def get_functional_train_with_signature(train_step, mesh, state_mesh_shardings, model, config):
   """Get the shardings (both state and data) for train_step"""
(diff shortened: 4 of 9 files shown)


MaxText/layers/gemma3.py
```diff
@@ -90,14 +90,16 @@ def __call__(self, inputs, train=False):
     Returns:
       jnp.array for image embeddings, shaped [B, N, P, D], e.g. [4, 1, 256, 2560]
     """
-    cfg = self.config
     print(f"Gemma3VisionEncoderLayer input: {inputs.shape}")
-    x = inputs.squeeze(axis=1)
+    jax.debug.print("GemmaViT inputs mean: {}", inputs.mean())
```
aireenmei, line 94: Is this for debugging? Should we remove?

aireenmei (approved):

MaxText/layers/gemma3.py
hengtaoguo: Yes just for my debugging purpose. Better to keep it in my local branch, I will remove them!

MaxText/multimodal_utils.py
```diff
@@ -0,0 +1,72 @@
...
+      shape=image_shape,
+      method="bilinear",
+      antialias=True,
+  )
+  image = _normalize_images(image)
+  image = jnp.clip(image, -1, 1)
+  return image
+
+
+def pre_process_image(
```
gagika, line 58: If this is only for gemma3 let's rename with to add gemma3 in the name.

MaxText/maxtext_utils.py
```diff
@@ -599,11 +599,19 @@ def init_kv_cache(model, config):
         config.global_batch_size_to_load,
         config.max_prefill_predict_length,
     )
+    image_shape = (
+        config.global_batch_size_to_load,
+        1,
+        config.image_size_for_vit,
+        config.image_size_for_vit,
+        3,
+    )
```
gagika, line 608: Could you use some constants for 1 and 3, e.g. 

NUM_IMAGES_PER_SEQUENCE and NUM_IMAGE_CHANNELS

MaxText/maxengine.py
```diff
@@ -448,6 +449,10 @@ def _prefill_jit(
     input_tokens = jnp.expand_dims(padded_tokens, 0)  # [BATCH, SEQUENCE]
     positions = jnp.expand_dims(jnp.arange(start_position, start_position + input_tokens.shape[1]), 0)
 
+    if self.config.use_multimodal:
+      input_images = jnp.expand_dims(images, 0)  # Add batch dimension [B, H, W, C]
+      input_images = jnp.expand_dims(input_images, 1)  # Add sequence dimension [B, N, H, W, C]
+
```
gagika, line 455: nit:
I think cleaner to replace those two with lines with single line using `jnp.newaxis`, e.g.

`input_images = images[jnp.newaxis, jnp.newaxis, ...]`

MaxText/configs/base.yml
```diff
@@ -681,3 +681,4 @@ nope_layer_interval: -1
 # Multimodal flags
 use_multimodal: False
 image_size_for_vit: 896 # Default for Gemma3, and should be overwritten by model's config
+image_path: "" # Local image path used for decoding
```
bvandermoon, line 684: Just for my understanding - why is this `Local image path` in the code comment? Is the user expected to upload images to the TPU VM for decoding? Or are the images read from GCS?

MaxText/multimodal_utils.py
```diff
@@ -0,0 +1,72 @@
...
+_GEMMA_DEFAULT_IMAGE_SIZE = 896
+_GEMMA_IMAGE_MEAN = (127.5,) * 3
+_GEMMA_IMAGE_STD = (127.5,) * 3
+
+
+def load_image_from_path(image_path):
+  """Loads an image from a given file path and returns a jnp.array."""
+  if not os.path.isfile(image_path):
+    print(f"Error: Image not found at path '{image_path}'")
+    return None
```
bvandermoon, line 17: Do we need a separate check for if `image_path` is not set at all? Something like 

```
Error: image_path must be set to run decode.py with use_multimodal=True?
```

MaxText/maxtext_utils.py
bvandermoon: +1 - also in `multimodal_utils.py`

requirements.txt
```diff
@@ -25,6 +25,7 @@ pytest
 pyink
 pre-commit
 pytype
+pillow>=11.1.0
```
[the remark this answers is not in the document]
hengtaoguo: This `pillow` is a standard image library that is for loading image files, and converted to array using `np.array(image)`. It is used in function `load_image_from_path()` inside multimodal_utils.py, when we try to load a local image and decode with it. Example usage:
```
from PIL import Image

image = Image.open(image_path).convert("RGB")
image_array = jnp.asarray(np.array(image))
```

MaxText/multimodal_utils.py
```diff
@@ -0,0 +1,72 @@
+import os
+import numpy as np
+import jax
+import jax.numpy as jnp
+from PIL import Image
+
+
+_GEMMA_DEFAULT_IMAGE_SIZE = 896
+_GEMMA_IMAGE_MEAN = (127.5,) * 3
+_GEMMA_IMAGE_STD = (127.5,) * 3
```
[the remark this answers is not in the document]
hengtaoguo: Yes, these values are Gemma3 specific.

Thanks for pointing this out! Now I'm thinking that `_normalize_images()` could be reusable across different VisionEncoder models in the future. So I've updated its interface to `_normalize_images(images, mean, std)`, so that we can pass in model-specific vars to do the normalization.

MaxText/multimodal_utils.py
hengtaoguo: Thanks for pointing this out! `pre_process_image` is actually a router function, and re-route to model-specific preprocessing steps based on `model_name`. For example, we have a `pre_process_gemma3_image()` function for gemma-specific steps if the model name is in gemma3 family.

I designed in this way, because different models may require different image size and intensities for the pretrained VisionEncoder. So in the future, we will make another `pre_process_llama4_image()` function and re-route it inside the `pre_process_image`. Please let me know if this sounds reasonable!

MaxText/maxtext_utils.py
hengtaoguo: Done! Added `NUM_IMAGES_PER_SEQUENCE=1` and `NUM_IMAGE_CHANNELS=3` in maxtext_utils.py. Also _NUM_IMAGE_CHANNELS in multimodal_utils.py.

MaxText/maxengine.py
hengtaoguo: Thanks for the suggestion! Changed to `input_images = images[jnp.newaxis, jnp.newaxis, ...]` for cleaner codes.

MaxText/configs/base.yml
hengtaoguo: Oh, this PR expects users to upload images to VM for decoding. So the image_path is the image's path inside the VM. For the text decoding, we only need to specify `prompt="I love to"`; but for image decoding, we definitely need a place to store the image data.

There are three ways to load an image for decoding. I plan to allow all of them in the future, but this PR only support the first one:
1. Load image from TPU VM.
2. Load image from a GCS path.
3. Load image from a URL on the internet.

Please let me know if this sounds reasonable to you!

MaxText/multimodal_utils.py
hengtaoguo: Thanks for the suggestion! Added a check to pyconfig, if use_multimodal=True but image_path is empty string or None, it will raise an AssertionError:

```
if keys["use_multimodal"]:
    validate_multimodal_model_name(keys["model_name"])
    assert keys["image_path"], "image_path must be set to run decode.py with use_multimodal=True"
```

MaxText/configs/base.yml
bvandermoon: Thanks @hengtaoguo. Makes sense to me. If there is extra work to support GCS then it's fine to add it in the near future. Just for my understanding - why is there extra work to support GCS?

hengtaoguo: I think it involves with two additional steps to support GCS:
1.  Download a image file from GCS (I guess we already have functions in maxtext to do the download)
2. Read this image using PIL and convert to numpy/jnp, just like what we did here for the local images.

Technically it is easy! We now support local image loading just for prototyping and faster iteration.

MaxText/multimodal_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    return jnp.asarray(np.array(image))
+  except (IOError, OSError) as e:
+    print(f"Error loading image from '{image_path}': {e}")
+    return None
+
+
+def _normalize_images(images, mean, std):
+  """Normalize the image to zero mean and unit variance.
+  In order to change the image mean and std, we need to change the _IMAGE_MEAN
+  and _IMAGE_STD global constants in this file.
```
gagika, line 31: _IMAGE_MEAN and _IMAGE_STD -> mean, std ?

gagika (approved):
Thanks, only one comment

gagika:
> Thanks, only one comment

Could you also run `bash code_style.sh ` to make sure new code style is good?

bvandermoon (approved):

MaxText/multimodal_utils.py
hengtaoguo: I just realized during training, we also need to set use_multimodal=True, but we don't actually need the image_path because it is only used for single case decoding.

So I changed the `load_image_from_path()` and let it catch the image loading error if the path in invalid:

```
if not os.path.isfile(image_path):
    raise FileNotFoundError(f"Image not found at path {image_path}. Please specify a valid image path")
```

MaxText/maxengine.py
```diff
@@ -1284,10 +1291,12 @@ def init(abstract_params, page_state):
           (int(self.config.per_device_batch_size * jax.device_count()), 1),
           dtype=jnp.int32,
       )
+      dummy_image = jnp.ones((int(self.config.per_device_batch_size * jax.device_count()), 1, self.config.image_size_for_vit, self.config.image_size_for_vit, 3), )
       _, cache = self.model.apply(
           abstract_params,
           x,
           x,
+          encoder_images=dummy_image if self.config.use_multimodal else None,
```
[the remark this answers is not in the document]
hengtaoguo: This is used as a placeholder to initiate the decode state, similar to what we already have during initiation. The above `x` is also a dummy_token vector and dummy_position vector, using jnp.ones:

```
x = jnp.ones(
          (int(self.config.per_device_batch_size * jax.device_count()), 1),
          dtype=jnp.int32,
      )
```

MaxText/multimodal_utils.py
hengtaoguo: Thanks for the catch! Updated the doc strings

PR #1612, opened 2025-04-21
Support zarr checkpoints for inference with Pathways.

RoshaniN:
# Description
Adding support for zarr checkpoints, compatible with Pathways. This helps with disaggregated serving using persistence APIs in Pathways.

Zarr checkpoints are generated when `use_ocdbt=False` and `use_zarr3=False` . There are config variables in Maxtext corresponding to both these flags `checkpoint_storage_use_ocdbt` and `checkpoint_storage_use_zarr3` which default to True. Through this PR, I am plumbing config overrides all the way to checkpointing functions.

Bug b/409627162

# Checklist

### DIFF
--- a/MaxText/checkpointing.py
+++ b/MaxText/checkpointing.py
@@ -184,6 +184,8 @@ def load_state_if_possible(
     enable_single_replica_ckpt_restoring: Optional[bool] = False,
     dataset_type: Optional[str] = "tfds",
     step: int = -1,  # -1 means latest
+    use_ocdbt = True,
+    use_zarr3 = True,
 ):
   """Loads TrainState as possible from the inputs.
 
@@ -291,7 +293,7 @@ def load_state_if_possible(
 
   if load_parameters_from_path != "":
     restored_params = load_params_from_path(
-        load_parameters_from_path, abstract_unboxed_pre_state.params, checkpoint_storage_concurrent_gb
+        load_parameters_from_path, abstract_unboxed_pre_state.params, checkpoint_storage_concurrent_gb, use_ocdbt=use_ocdbt, use_zarr3=use_zarr3
     )
     return None, restored_params
   elif load_full_state_from_path != "":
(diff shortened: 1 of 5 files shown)


khatwanimohit (approved):

MaxText/checkpointing.py
```diff
@@ -351,9 +353,9 @@ def load_params_from_path(load_parameters_from_path, abstract_unboxed_params, ch
   return restored["params"]
 
 
-def save_params_to_path(checkpoint_dir, params):
+def save_params_to_path(checkpoint_dir, params, use_ocdbt=True, use_zarr3=True):
   """Save decode params in checkpoint at specified path."""
   assert checkpoint_dir, "checkpoint_dir is not defined."
-  orbax_checkpointer = ocp.PyTreeCheckpointer()
+  orbax_checkpointer = ocp.PyTreeCheckpointer(use_ocdbt=use_ocdbt, use_zarr3=use_zarr3)
```
gobbleturk, line 359: I see the default (and thus the previous behavior) was use_zarr3=False

http://google3/third_party/py/orbax/checkpoint/_src/checkpointers/pytree_checkpointer.py;l=23;rcl=698173307

can we keep this default?

RoshaniN: I was trying to understand why use_zarr3 is False by default here. 

On poking into Maxtext a little more I see instances like https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/checkpointing.py#L58 and https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/configs/base.yml#L60 where it is set to be True.

Can I get some guidance on what the default should be? I think this is a case of missing plumbing.

richjames0 (commented):
Matt's comment is well noted but I think this change is ok

richjames0 (approved):
lgtm

PR #1614, opened 2025-04-22
Fix setup.sh to not block on the service restart screen

dgup-ai:
# Description

Fix the setup.sh script to not block on a service restart screen. 

Certain TPU images use a Linux version where when running setup.sh, users get a purple screen to restart services. This ends up often hanging the session and preventing set up from continuing. This changes enables the auto mode for service restart and stops this from happening. It uses sed to replace the interactive flag with the automatic flag in the /etc/needrestart/needrestart.conf file. 

This is fairly basic quick fix and more comprehensive solution would require larger changes to the set up workflow. 

# Tests

1. Launched a version of Linux where setup.sh triggers the purple screen that hangs set up.
2. Verified that with this change that screen doesn't appear and set up continues. 
3. Confirmed I could run the training loop.

# Checklist

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -31,6 +31,14 @@ export DEBIAN_FRONTEND=noninteractive
 export NEEDRESTART_SUSPEND=1
 export NEEDRESTART_MODE=l
 
+# Enable automatic restart of services without the need for prompting 
+if command -v sudo &> /dev/null
+then
+   sudo sed -i "s/#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf
+else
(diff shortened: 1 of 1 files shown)


richjames0 (approved):
To the moon!

vipannalla (approved):

gobbleturk:
I love this thank you! It looks like it is failing our GPU tests - can you wrap the change if a check that `DEVICE=tpu`?

dgup-ai:
> I love this thank you! It looks like it is failing our GPU tests - can you wrap the change if a check that `DEVICE=tpu`?

It failed because sudo was missing - i'll add a check to first look for sudo

PR #1615, opened 2025-04-23
support bidirectional attention mask for multimodal

aireenmei:
# Description

Support Gemma3 style bidirectional attention for image tokens in dot_product attention
(image from: https://huggingface.co/blog/gemma3)
image

# Checklist

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -244,6 +244,41 @@ def _generate_chunk_attention_mask(mask_shape: tuple[int, int], chunk_size: int)
   chunk_mask = same_chunk & (row_ids >= col_ids)
   return chunk_mask
 
+def _make_block_mask_indices(bidirectional_mask):
+  """Creates block mask identifying segments based on a bidirectional mask.
+
+  Args:
+    bidirectional_mask: boolean mask, e.g. [011110011010].
+
+  Returns:
+    block mask for segments, e.g. [011110022030].
+  """
+  # Left pad 0.
+  padded_mask = jnp.pad(bidirectional_mask, [(0, 0), (1, 0)], constant_values=0)
+  boundary = padded_mask[..., 1:] > padded_mask[..., :-1]
+  numbered_boundary = jnp.cumsum(boundary, axis=-1)
+  return bidirectional_mask * numbered_boundary
+
+def _make_bidirectional_block_mask(bidirectional_mask):
+  """Creates bidirectional block mask from bidirectional_mask, where True corresponds to image tokens.
+  bidirectional_mask shape: [B, L]
+  bidirectional_block_mask shape: [B, L, L]
+  Examples:
+  bidirectional_mask = [[0, 1, 1, 1, 0, 0]]
+  bidirectional_block_mask = [[
+      [False, False, False, False, False, False],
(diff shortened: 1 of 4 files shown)


MaxText/layers/gemma3.py
```diff
@@ -47,7 +47,12 @@
 LENGTH = common_types.LENGTH
 HEAD = common_types.HEAD
 D_KV = common_types.D_KV
-
+BEGIN_IMAGE_TOKEN = 255999
+END_IMAGE_TOKEN = 262144
+NEW_LINE_TOKEN = 108
+TOKEN_PLACEHOLDER = -2
+NUM_PLACEHOLDER_TOKENS_PER_IMAGE = 256
+NUM_TOKENS_PER_MEDIA = NUM_PLACEHOLDER_TOKENS_PER_IMAGE + 4
```
hengtaoguo, line 55: Thank you Aireen!

I see these variables not being used in the current PR. Should we include them in this PR, or address them in a following one? Either way works for me.

hengtaoguo (approved):
Thanks for the unit tests!

MaxText/layers/gemma3.py
aireenmei: I leave TOKEN_PLACEHOLDER which is used in this PR, removed others. Thanks for the suggestion

MaxText/tests/attention_test.py
```diff
@@ -40,6 +40,60 @@
 MLA = attentions.MLA
 
 
+class BidirectionalBlockMaskTest(unittest.TestCase):
```
gagika, line 43: could you also test when B > 1, e.g. B == 2 with each batch element has different image position;

[[0, 1, 1, 1, 0, 0],
 [1, 1, 0, 0, 0, 0]]

MaxText/layers/attentions.py
```diff
@@ -335,6 +370,10 @@ def generate_attention_mask(
       chunk_mask = _generate_chunk_attention_mask(mask_shape=(q_seq_len, kv_seq_len), chunk_size=self.chunk_attn_window_size)
       output_mask = chunk_mask * output_mask
 
+    if bidirectional_mask is not None:
+      image_mask = _make_bidirectional_block_mask(bidirectional_mask)
+      output_mask = output_mask | image_mask[:, None, None, ...]
```
gagika, line 375: Does output_mask need to be extended? 
e.g. if sequence length is 6, output_mask is 6x6 ? do you need to extend it to count for image token position?
It would be great to have some tests of combining text mask + image mask.

gagika (commented):
Thanks, I have 2 comments about testing.

MaxText/tests/attention_test.py
aireenmei: Good suggestion, I added test_batch_block_masks

MaxText/layers/attentions.py
aireenmei: output_mask is already broadcasted to (1,1,1,seq_len, seq_len). So only extend image_mask to (batch, 1, 1, seq_len, seq_len). I added a unit test: test_combine_with_causal_mask

gagika (approved):
thanks

PR #1616, opened 2025-04-23
Initial support for pipelining a subset of layers

gobbleturk:
# Description

 Support an option to pipeline a subset of layers. This is useful for our currently SPMD pipelining implementation which requires that the pipelining degree must divide the number of layers. Some OSS models don't have a friendly number of layers, e.g. 58 for deepseek.

This additionally separates the scanning pipeline layers per stage into its own option.

Full support for PP of subset on DeepSeek and Llama4 will require more modifications because they have non-homogenous layers which is handled [here](https://github.com/AI-Hypercomputer/maxtext/blob/3409fb4f9264809723a83892240b452feb5837fb/MaxText/layers/models.py#L462) 

Supporting: b/404566880

# Tests

* Added tests in this PR
* [xprof](https://xprof.corp.google.com/trace_viewer/mattdavidow-12989516961696272071?hosts=t1v-n-70a81276-w-0&host_index=0&trace_filter_config={}&view_start=-789.326&view_end=6058.069) of PP=4, subset layers=8, total layers=16 on a v6e-16

# Checklist

### DIFF
--- a/MaxText/maxtext_utils.py
+++ b/MaxText/maxtext_utils.py
@@ -689,6 +689,23 @@ def add_config_to_summary_writer(config, summary_writer):
     for key, value in config.get_keys().items():
       max_utils.add_text_to_summary_writer(key, str(value), summary_writer)
 
+def logical_axis_rules_pp_act_as_dp(logical_rules):
+  """Add stage as a physical axes before data for each rule, so stage acts just like data instead of PP.
+  This is used when we want to pipeline only a subset of layers, and leave the rest like DP.
+  """
+  new_rules = []
+  for key, physical_axes in logical_rules:
+    if isinstance(physical_axes, str):
+      physical_axes = (physical_axes,)
+    else:
+      physical_axes = tuple(physical_axes)
+    new_physical_axes = tuple(axis for axis in physical_axes if axis != "stage")
+    if "data" in new_physical_axes:
+      data_idx = new_physical_axes.index("data")
+      new_physical_axes = new_physical_axes[0:data_idx] + ("stage",) + new_physical_axes[data_idx:]
+    new_rules.append((key, new_physical_axes))
+  return tuple(new_rules)
+
 
 def create_device_mesh(config, devices=None):
   """Creates a device mesh with each slice in its own data parallel group. If there is only one slice, uses two replicas"""
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -775,18 +775,25 @@ def set_and_validate_pipeline_config(raw_keys):
     raw_keys["logical_axis_rules"] = modify_activation_embed_and_logits_batch(raw_keys["logical_axis_rules"])
     raw_keys = pipeline_first_axis(raw_keys)
     num_stages = int(raw_keys["ici_pipeline_parallelism"] * raw_keys["dcn_pipeline_parallelism"])
+    if raw_keys["pipeline_parallel_layers"] == -1:
+      raw_keys["pipeline_parallel_layers"] = raw_keys["num_decoder_layers"]
+    else:
+      assert (raw_keys["pipeline_parallel_layers"] <= raw_keys["num_decoder_layers"]), f"You can only pipeline a subset of the decoder layers, but you requested to pipeline {raw_keys['pipeline_parallel_layers']} with pipeline_parallel_layers and there are only {raw_keys['num_decoder_layers']} decoder layers."
+    assert raw_keys["scan_layers"] or raw_keys["pipeline_parallel_layers"]==raw_keys["num_decoder_layers"], "Currently we only support scan_layers=True when pipelining a subset of layers."
+    assert raw_keys["pipeline_parallel_layers"] > 0, "You must set pipeline_parallel_layers to a positive integer less than or equal to the number of layers"
+
     if raw_keys["num_pipeline_repeats"] == -1:
       num_pipeline_repeats, remainder = divmod(
-          raw_keys["num_decoder_layers"], num_stages * raw_keys["num_layers_per_pipeline_stage"]
+          raw_keys["pipeline_parallel_layers"], num_stages * raw_keys["num_layers_per_pipeline_stage"]
       )
       assert (
           not remainder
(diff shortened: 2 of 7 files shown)


MaxText/maxtext_utils.py
```diff
@@ -689,6 +689,24 @@ def add_config_to_summary_writer(config, summary_writer):
     for key, value in config.get_keys().items():
       max_utils.add_text_to_summary_writer(key, str(value), summary_writer)
 
+def logical_axis_rules_pp_act_as_dp(logical_rules):
+  """Add stage as a physical axes before data for each rule, so stage acts just like data instead of PP.
+  This is used when we want to pipeline only a subset of layers, and leave the rest like DP.
+  """
+  new_rules = []
+  for rule in logical_rules:
+    key = rule[0]
```
richjames0, line 698: nit: tuple unpacking would be cleaner here, i.e. for key, value in logical_rules

MaxText/pyconfig.py
```diff
@@ -770,18 +770,25 @@ def pipeline_first_axis(raw_keys):
     raw_keys["logical_axis_rules"] = modify_activation_embed_and_logits_batch(raw_keys["logical_axis_rules"])
     raw_keys = pipeline_first_axis(raw_keys)
     num_stages = int(raw_keys["ici_pipeline_parallelism"] * raw_keys["dcn_pipeline_parallelism"])
+    if raw_keys["pipeline_parallel_layers"] == -1:
+      raw_keys["pipeline_parallel_layers"] = raw_keys["num_decoder_layers"]
+    else:
+      assert (raw_keys["pipeline_parallel_layers"] <= raw_keys["num_decoder_layers"]), f"You can only pipeline a subset of the decoder layers, but you requested to pipeline {raw_keys['pipeline_parallel_layers']} with pipeline_parallel_layers and there are only {raw_keys['num_decoder_layers']} decoder layers."
+    if raw_keys["pipeline_parallel_layers"] < raw_keys["num_decoder_layers"]:
```
richjames0, line 777: I would personally incorporate the if into the assertion. also i would throw instead of asserting cause this isn't a coding error, it's an input error

richjames0 (commented):
a couple of nits that it'd be nice to address but if asked nicely i'd approve

MaxText/pyconfig.py
gobbleturk: I agree they should probably by throws instead of asserting but currently in all of pyconfig we use asserts, so we can en masse change all of the asserts to throws at some point in 2.0 land

MaxText/maxtext_utils.py
gobbleturk: Done

wang2yn84 (commented):
Thank you Matt! Looks good.

MaxText/pyconfig.py
```diff
@@ -770,18 +770,24 @@ def pipeline_first_axis(raw_keys):
     raw_keys["logical_axis_rules"] = modify_activation_embed_and_logits_batch(raw_keys["logical_axis_rules"])
     raw_keys = pipeline_first_axis(raw_keys)
     num_stages = int(raw_keys["ici_pipeline_parallelism"] * raw_keys["dcn_pipeline_parallelism"])
+    if raw_keys["pipeline_parallel_layers"] == -1:
+      raw_keys["pipeline_parallel_layers"] = raw_keys["num_decoder_layers"]
+    else:
+      assert (raw_keys["pipeline_parallel_layers"] <= raw_keys["num_decoder_layers"]), f"You can only pipeline a subset of the decoder layers, but you requested to pipeline {raw_keys['pipeline_parallel_layers']} with pipeline_parallel_layers and there are only {raw_keys['num_decoder_layers']} decoder layers."
```
yangyuwei, line 776: Shall we also assert raw_keys["pipeline_parallel_layers"] > 0 here?

yangyuwei (approved):
Thanks Matt for adding this feature. Very helpful!

MaxText/maxtext_utils.py
```diff
@@ -689,6 +689,23 @@ def add_config_to_summary_writer(config, summary_writer):
...
       max_utils.add_text_to_summary_writer(key, str(value), summary_writer)
 
+def logical_axis_rules_pp_act_as_dp(logical_rules):
+  """Add stage as a physical axes before data for each rule, so stage acts just like data instead of PP.
+  This is used when we want to pipeline only a subset of layers, and leave the rest like DP.
+  """
+  new_rules = []
+  for key, physical_axes in logical_rules:
+    if isinstance(physical_axes, str):
+      physical_axes = (physical_axes)
```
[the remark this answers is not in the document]
gobbleturk: Done, good catch

MaxText/pyconfig.py
```diff
@@ -775,18 +775,25 @@ def pipeline_first_axis(raw_keys):
...
+    assert raw_keys["scan_layers"] or raw_keys["pipeline_parallel_layers"]==raw_keys["num_decoder_layers"], "Currently we only support scan_layers=True when pipelining a subset of layers."
```
wang2yn84, line 782: I thought we set scan_layers=False and scan_pipeline_iterations=True for PP?

gobbleturk: ya so this is a fairly breaking change for PP, from now on we should set

scan_pipeline_iterations=True
scan_layers_per_stage=False

these are set by default so also using defaults is fine

gobbleturk: scan_layers now only refers to data parallel - for training we also recommend setting this to true, and currently when pipeline_parallel_layers< num_decoder_layers I am only supporting scan_layers=True for now

gobbleturk: to summarize there are scans

* pipeline iterations `scan_pipeline_iterations` - generally recommended to True
* layers within a single stage `scan_layers_per_stage` - generally recommended to False, however when there are many layers per stage possibly setting this to True might save memory
* Data parallel layers use `scan_layers` - generally recommended to True. With this PR when `pipeline_parallel_layers< num_decoder_layers` only true is supported

wang2yn84: Ah gotcha, the summary is helpful! Better document somewhere in the PP doc.

wang2yn84 (approved):
Thank you!

PR #1623, opened 2025-04-24
Moving the checkpointing util funtions to a spearate file to avoid circular dependency

wang2yn84:
# Description

The circular dependency is multihost_dataloading -> maxtext_utils -> checkpointing -> multihost_dataloading. Moving all the maxtext_util functions that depends on checkpointing to a separate util function solves this issue.

# Tests

Presubmit tests.

# Checklist


MaxText/maxtext_checkpointing_utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+from MaxText import max_utils
+
+import functools
+
+
+from flax.linen import partitioning as nn_partitioning
+
+from MaxText import max_logging
+from MaxText import maxtext_utils
```
bvandermoon, line 30: `from MaxText import max_utils` is repeated twice

MaxText/convert_gpt3_ckpt_from_paxml.py
```diff
@@ -52,6 +52,7 @@
 import jax
 import gc
 from MaxText import max_logging
+from MaxText import maxtext_checkpointing_utils
```
bvandermoon, line 55: Looks like `from MaxText import max_utils` is no longer needed. Can you remove it from this file and other ones where it isn't needed either?

MaxText/maxtext_checkpointing_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    # Load params from checkpoint
+    max_logging.log(f"Loading decode params from {config.load_parameters_path}")
+    unboxed_abstract_state, state_mesh_annotations, _ = maxtext_utils.get_abstract_state(model, None, config, rng, mesh, False)
+    with nn_partitioning.axis_rules(config.logical_axis_rules):
+      params = checkpointing.load_params_from_path(
+          config.load_parameters_path, unboxed_abstract_state.params, config.checkpoint_storage_concurrent_gb, config.checkpoint_storage_use_ocdbt, config.checkpoint_storage_use_zarr3
+      )
+    state = maxtext_utils.init_decode_state(None, params)
+
+  state = max_utils.unbox_logicallypartioned(state)
```
bvandermoon, line 141: Is this change trying to get rid of the max_utils dependency?

bvandermoon (commented):
Thanks @wang2yn84 for the nice refactor. Was the circular dependency causing some type of error? Or is the change for cleaner code?

wang2yn84:
> Thanks @wang2yn84 for the nice refactor. Was the circular dependency causing some type of error? Or is the change for cleaner code?

Hi Branden, in the current implementation, we have to carefully avoid depend on maxtext_utils from multihost_dataloading. There are couple of util functions in multihost_dataloading we should place in maxtext_utils but we can't.

MaxText/maxtext_checkpointing_utils.py
wang2yn84: The one in line22 is max_utils and line30 is maxtext_utils and those are 2 different util modules.

MaxText/maxtext_checkpointing_utils.py
wang2yn84: The change is to move all the util functions that depends on MaxText.checkpointing to this file so that other files can depends on maxtext_utils without circular dependency issue. maxtext_checkpointing_utils can depend on max_utils and maxtext_utils without any issue.

bvandermoon (approved):

MaxText/maxtext_checkpointing_utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+def save_quantized_checkpoint_if_configured(config, params):
+  assert config.quantization, "quantization must be configured"
+  if config.save_quantized_params_path:
+    checkpointing.save_params_to_path(config.save_quantized_params_path, params)
+  else:
+    "Skipping saving quantized checkpoint as save_quantized_params_path is null."
+
+
+def setup_training_state(model, data_iterator, tx, config, rng, mesh, checkpoint_manager):
```
gobbleturk, line 154: I don't think this method fits in this file - this define the training state, its needed even without checkpointing

MaxText/maxtext_checkpointing_utils.py
```diff
@@ -0,0 +1,166 @@
...
+      )(rng)
+      if raw_params:  # If we loaded a partial state, we need to merge it.
+        state = state.replace(params=raw_params)
+
+  state = max_utils.unbox_logicallypartioned(state)
+
+  return state, state_mesh_annotations, state_mesh_shardings, data_iterator
+
+
+def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
```
gobbleturk, line 112: This is not a checkpointing method, its needed even when checkpointing is turned off

MaxText/maxtext_checkpointing_utils.py
```diff
@@ -0,0 +1,166 @@
...
+from flax.linen import partitioning as nn_partitioning
+
+from MaxText import max_logging
+from MaxText import maxtext_utils
+from MaxText import checkpointing
+import orbax.checkpoint.experimental.emergency.checkpoint_manager as emergency_checkpoint_manager
+import orbax.checkpoint.experimental.emergency.replicator_checkpoint_manager as emergency_replicator_checkpoint_manager
+
+
+def setup_initial_state(
```
gobbleturk, line 36: this is not a checkpointing function, its needed even when checkpointing is turned off

MaxText/maxtext_checkpointing_utils.py
wang2yn84: Chatted offline, agreed to rename this new util file to maxtext_state_initializion_utils.py.

MaxText/maxtext_checkpointing_utils.py
wang2yn84: Ditto

MaxText/maxtext_checkpointing_utils.py
wang2yn84: Ditto.

MaxText/maxtext_state_initialization_utils.py
```diff
@@ -0,0 +1,166 @@
...
+          config.load_parameters_path, unboxed_abstract_state.params, config.checkpoint_storage_concurrent_gb, config.checkpoint_storage_use_ocdbt, config.checkpoint_storage_use_zarr3
+      )
+    state = maxtext_utils.init_decode_state(None, params)
+
+  state = max_utils.unbox_logicallypartioned(state)
+  return state, state_mesh_annotations
+
+
+
+def save_quantized_checkpoint_if_configured(config, params):
```
gobbleturk, line 146: Can you have some inference folks look at this change @vipannalla or @mitalisi

gobbleturk (approved):

vipannalla (approved):
Looks like a simple split, not new functionality. LGTM, did you do a end-2-end run for inference to confirm nothing is broken?

PR #1629, opened 2025-04-24
Add support to convert deepseek-r1 from fp8 to bf16 on CPU

Lumosis:
# Description

Converts weights from FP8 to BF16 for a HF model on CPU.

The script is mainly ported from https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/fp8_cast_bf16.py. The original script is GPU-based.

Example cmd:

python3 -m MaxText.fp8_to_bf16 --input-fp8-hf-path <path/to/fp8/ckpt> \
    --output-bf16-hf-path <local/path/to/save/new/bf16/ckpt>

For deepseek-R1, the conversion process takes 1~2 hrs and the memory usage is below ~25 GB.

# Tests

I compared the model files generated using my script with those generated using the original script by running `diff` command:
```
root@ranlihao-m1-ultramem-160:/mnt/disks/persist# diff models/deepseek_bf16 models/deepseek_bf16_ref/hf-r1-671b-bf16/

root@ranlihao-m1-ultramem-160:/mnt/disks/persist# 
```
There is no difference.

# Checklist

### DIFF
--- a/MaxText/deepseek_fp8_to_bf16.py
+++ b/MaxText/deepseek_fp8_to_bf16.py
@@ -1,0 +1,172 @@
+"""
+Copyright 2025 Google LLC
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     https://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+r"""Convert weights from FP8 to BF16 for a HF model.
+
+Install these dependencies before running this script:
+
+pip install torch==2.4.1 safetensors==0.4.5
+
+Example cmd:
+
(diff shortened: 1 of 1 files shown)


mitalisi (approved):

requirements.txt
```diff
@@ -40,3 +40,5 @@ google-jetstream@git+https://github.com/AI-Hypercomputer/JetStream.git
 jsonlines
 pathwaysutils@git+https://github.com/AI-Hypercomputer/pathways-utils.git
 omegaconf
+torch==2.4.1
+safetensors==0.4.5
```
gagika, line 44: This will make MaxText install size much larger.

let's not add those dependancies in `requirements.txt` we can add an instruction to install them before running your script, or we can make your script a bash that installs those.

MaxText/fp8_to_bf16.py
gagika, line 1: Could you rename it to something like, deepseek_p8_to_bf16.py?

gagika (commented):
a few comments

requirements.txt
Lumosis: Done

MaxText/fp8_to_bf16.py
Lumosis: Done

gagika (approved):
Thanks

PR #1647, opened 2025-04-28
Ignore auth errors in CI Tests

shralex:
See title. Some of these runners already have auth set-up, so a failure at this stage may not actually fail the task. 

# Checklist


gobbleturk (approved):

A9isha (approved):

khatwanimohit (approved):
Did that command fail if the auth was already setup. I would guess running that command should be a no-op

bvandermoon (approved):

PR #1652, opened 2025-04-29
gemma3 multimodal: image and text token fusion

aireenmei:
# Description
In gemma3, <start_of_image> is used in prompt to mark the place of images (for exp, prompt="Describe this image <start_of_image>"). The <start_of_image> token is then replaced by some placeholder tokens, which is later used when merging text and image embeddings.
 
Reference functions published in gemma3: https://github.com/google-deepmind/gemma/blob/main/gemma/gm/vision/_token_utils.py#L45

# Tests
Added unit tests. Also breakpoint checked when passing an image and run decode.py

# Checklist

### DIFF
--- a/MaxText/decode.py
+++ b/MaxText/decode.py
@@ -94,11 +94,10 @@ def main(argv: Sequence[str]) -> None:
   params = engine.load_params(rng_load_params)
 
   text = config.prompt
-  images = None
+  prefill_length = config.max_prefill_predict_length
   if config.use_multimodal:
-    # TODO(hengtaoguo): Support multiple images as input.
-    images = multimodal_utils.load_image_from_path(config.image_path)
-    images = multimodal_utils.pre_process_image(images, model_name=config.model_name)
+    text = multimodal_utils.reformat_prompt(text, config.model_name)
+    prefill_length -= multimodal_utils.get_image_offsets(config.model_name)
 
   metadata = engine.get_tokenizer()
   tokenizer_model = engine.build_tokenizer(metadata)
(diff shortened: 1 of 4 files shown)


MaxText/decode.py
```diff
@@ -94,11 +94,9 @@ def main(argv: Sequence[str]) -> None:
   params = engine.load_params(rng_load_params)
 
   text = config.prompt
```
hengtaoguo, line 96: Should we also add this "format" function to transform the text prompt string? (Next PR is totally fine)

From: Describe this image <start_of_image>
To: <start_of_turn>user\nDescribe image <start_of_image><end_of_turn>\n<start_of_turn>model\n


https://github.com/google-deepmind/gemma/blob/3d97415726e385c96cae52b9598444c5dfc4fd1d/gemma/gm/text/_chat_sampler.py#L182

MaxText/multimodal_utils.py
```diff
@@ -82,3 +82,244 @@ def pre_process_image(image, model_name):
     return pre_process_gemma3_image(image)
   else:
     raise ValueError(f"Model {model_name} does not support multimodal inference.")
+
+
+def get_image_offsets(model_name):
```
hengtaoguo, line 87: Could you also add some description to this function?

hengtaoguo (approved):
Thank you Aireen for the quick implementation!

MaxText/multimodal_utils.py
aireenmei: Added.

MaxText/decode.py
aireenmei: Thanks for the pointer! Added function multimodal_utils.reformat_prompt

gagika (approved):
thanks

PR #1673, opened 2025-05-02
Update Llama4 configs

RissyRan:
# Description

Update Llama4 configs:
* Add check to avoid scanned layer usage
* Remove `capacity_factor` config from model config as we already have check in pyconfig
* Remove configs to allow users to run megablox/ragged_dot, supported in [PR](https://github.com/AI-Hypercomputer/maxtext/commit/52a7fcd452e1be2a156ffa66d995d93f1812e689)

# Tests

Test locally for the assertion - behavior as expected

# Checklist

### DIFF
--- a/MaxText/pyconfig.py
+++ b/MaxText/pyconfig.py
@@ -262,6 +262,8 @@ def validate_llama4_config(keys: dict):
     raise ValueError("Llama4 decoder has not been tested with capacity_factor >= 0 -- please set that value to -1 for now!")
   if keys["num_experts_per_tok"] > 1:
     raise ValueError("Only top-1 routing is supported for Llama4 for now!")
+  if keys["scan_layers"]:
+    raise ValueError("Only unscanned layer is supported for Llama4, and please set scan_layers=False for now!")
   if keys["base_num_decoder_layers"] % keys["interleave_moe_layer_step"] != 0:
     raise ValueError(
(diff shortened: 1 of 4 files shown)


MaxText/pyconfig.py
```diff
@@ -262,6 +262,8 @@ def validate_llama4_config(keys: dict):
     raise ValueError("Llama4 decoder has not been tested with capacity_factor >= 0 -- please set that value to -1 for now!")
   if keys["num_experts_per_tok"] > 1:
     raise ValueError("Only top-1 routing is supported for Llama4 for now!")
+  if keys["scan_layers"]:
+    raise ValueError("Only unscanned layer is supported for Llama4 for now!")
```
jrplatin, line 266: nit: might be better to say `scan_layers=False` for clarity

jrplatin (approved):
LGTM, thanks!

MaxText/pyconfig.py
RissyRan: Do you actually mean to write `if keys["scan_layers"] is False`, as we raise error when scan_layers=True?

Gemini mentioned:

```
For most common scenarios where you want to check if a value is "truthy" (i.e., not False, None, empty, or zero), if keys["scan_layers"]: is the preferred and more Pythonic choice. It leads to cleaner, more readable, and conventional Python code.
```

bvandermoon (approved):
LGTM pending open conversation with @jrplatin

MaxText/pyconfig.py
RissyRan: Oh, I see. Updated it!

PR #1681, opened 2025-05-02
Prepare for running on GKE runners

shralex:
This PR is refactoring the code to be able to use GKE cloud runners. One of the current requirements of this framework is using containers.

# Checklist

### DIFF
--- a/.github/workflows/RunTests.yml
+++ b/.github/workflows/RunTests.yml
@@ -67,6 +67,7 @@ jobs:
     with:
       device_type: tpu
       device_name: v4-8
+      # cloud_runner: linux-x86-ct4p-240-4tpu
       pytest_marker: 'not gpu_only and not integration_test'
       test_directory: 'tests'
       xla_python_client_mem_fraction: 0.75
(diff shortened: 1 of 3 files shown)


gobbleturk (approved):

.github/workflows/RunTests.yml
```diff
@@ -67,6 +67,7 @@ jobs:
     with:
       device_type: tpu
       device_name: v4-8
+      # cloud_runner: linux-x86-ct4p-240-4tpu
```
gobbleturk, line 70: what are these comments for? Generally we should not comment out code, its quite confusing

hengtaoguo (approved):

.github/workflows/RunTests.yml
gobbleturk: Is this intentional - is the cloud_runner a cpu runner? why is it under the tpu_unit_tests?

shralex: There are three groups of cloud runners, each one with 1 tag. This one is the tag for TPU runners. You can see them in the runners page. This PR added a new parameter to run_tests_internal (cloud_runner). Basically, once we want to enable this we just need to uncomment these lines. The reason I'm doing this PR now is to allow Rohan to build upon it to enable CPU cloud runners first, because this part should already work.

PR #1683, opened 2025-05-02
Fix stable stack missing google-cloud-monitoring dependency

bvandermoon:
# Description

Fixed missing dependency for JSS stable stack. This was hidden in tests because we are using a hardcoded pip freeze list for JSS 0.5.2 instead of `requirements_with_jax_stable_stack.txt`. I am thinking about removing that short-term solution file and noticed the issue then. But I believe JSS 0.4.35 and JSS 0.4.37 are likely impacted by this missing dependency.

I saw this workload error when running JSS 0.5.2 without using `requirements_with_jax_stable_stack_0_5_2_pipreqs.txt` locally:

Screenshot 2025-05-02 at 4 35 34 PM

# Tests

```
python3 -m benchmarks.benchmark_runner xpk \
    --project=$PROJECT \
    --zone=$ZONE \
    --device_type=v6e-8 \
    --num_slices=1  \
    --cluster_name=${CLUSTER_NAME} \
    --base_output_directory=${OUTPUT_DIR} \
    --model_name="llama3_1_8b_8192_no_collective_matmul" \
    --base_docker_image=maxtext_base_image
```

This workload runs successfully now after building JSS 0.5.2 without using the `pip freeze` requirements file locally.

# Checklist

### DIFF
--- a/maxtext_jax_stable_stack.Dockerfile
+++ b/maxtext_jax_stable_stack.Dockerfile
@@ -34,6 +34,9 @@ RUN if [ "$DEVICE" = "tpu" ] && ([ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pk
 # Install Maxtext requirements with Jax Stable Stack
 RUN apt-get update && apt-get install --yes && apt-get install --yes dnsutils
 
+# TODO(bvandermoon, parambole): Remove this when it's added to JAX AI Image
+RUN pip install google-cloud-monitoring
+
 # Install requirements file generated with pipreqs for JSS 0.5.2. 
 # Othewise use general requirements_with_jax_stable_stack.txt
 RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.5.2-rev1" ]; then \


hengtaoguo (approved):

gobbleturk (approved):

requirements_with_jax_stable_stack.txt
```diff
@@ -2,6 +2,7 @@
 # These requirements are additional to the dependencies present in the JAX SS base image.
 datasets
 grain[parquet]>=0.2.6
+google-cloud-monitoring
```
parambole, line 5: @bvandermoon - Looks like just adding this without pinning other dependencies is causing it to go to though `pip resolver` and is increasing the build time which would break ( potentially ) unit tests due to longer build times.

bvandermoon: Thanks for noticing this @parambole. Any suggestions on what we can do here? Would pinning it to a specific version help?

parambole: @bvandermoon - Instead of putting in the requirements file which make it harder for pip can we move it to https://github.com/AI-Hypercomputer/maxtext/blob/main/maxtext_jax_stable_stack.Dockerfile

like `RUN pip install google-cloud-monitoring`

bvandermoon: Thanks @parambole. Will it not work to pin the dependency to a specific version? I am a little concerned about moving dependencies directly into the `Dockerfile` for this, but if it's the only way to speed up builds we can do it

parambole: @bvandermoon - Pinning the dependency might help, but it might not be trivial, as it is not only this but other dependencies that work well with this current package to ensure that pip does not take a lot of time resolving the other dependencies.

**However, while that is an option, I would still recommend putting it into the Dockerfile for now to get unblocked.**

Additionally, we can also take an action item to include it in the next JAX AI Image release.

bvandermoon: Makes sense, thanks @parambole. Updated and re-tested

PR #1686, opened 2025-05-05
Disable Llama4 unscanned test

RissyRan:
# Description

Disable Llama4 unscanned test for a shorter compiler test on github runners

# Tests

Expect github runners to pass

# Checklist

### DIFF
--- a/MaxText/tests/train_compile_test.py
+++ b/MaxText/tests/train_compile_test.py
@@ -559,6 +559,7 @@ class TrainCompile(unittest.TestCase):
         )
     )
 
+  @pytest.mark.skip(reason="b/415132665: Enable it once scan is supported in training for shorter compiler time")
   @pytest.mark.tpu_only
   def test_moe_llama4_17b_16e(self):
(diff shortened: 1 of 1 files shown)


shralex (approved):

MaxText/tests/train_compile_test.py
```diff
@@ -558,7 +558,8 @@ def test_moe_deepseek_without_device_limit(self):
             "topk_routing_group=-1",
         )
     )
-
+  
+  @pytest.mark.skip(reason="b/415132665: Enable it once scan is supported in training for shorter compiler time")
```
bvandermoon, line 562: Is there a testing gap without this? Or are we still covered by other tests?

bvandermoon (approved):

MaxText/tests/train_compile_test.py
RissyRan: Thanks for the question! Yes, there will be a risk that breaks Llama4 implementation with other features onboarding. I think we are prioritizing the scan version for performance purpose, and should be able to bring it up shortly.

PR #1687, opened 2025-05-05
Add support for SPMD PP for deepseek decoder block

gobbleturk:
# Description

Add support for using PP with deepseek, including with the new feature `pipeline_parallel_layers` which only pipelines a subset of layers. This change can help out with SPMD pipelining since PP must divide the number of layers. E.g. for deepseek there are 58 sparse layers, which does not have many friendly divisors. 

With this PR we can pipeline just a subset of the sparse layers, e.g. 56 of them and set PP=8. Other layers will be sharded like DP.

# Tests

Ran some locally of smaller v2-16B model  and added AOT test

* DeepSeekV2 16B local on v6e-8 with pdb=1, PP=8, pipeline_subset_layers=24 [xprof](https://xprof.corp.google.com/trace_viewer/mattdavidow-2021970820805495549?hosts=t1v-n-70a81276-w-0&host_index=0&trace_filter_config={}&view_start=-2803.551&view_end=172.797)
* With pure FSDP step time is roughly 2x [xprof](https://xprof.corp.google.com/trace_viewer/mattdavidow-463037051654911153?hosts=t1v-n-70a81276-w-0&host_index=0&trace_filter_config={}&view_start=-678.904&view_end=2265.952)

# Checklist

### DIFF
--- a/MaxText/layers/models.py
+++ b/MaxText/layers/models.py
@@ -210,7 +210,7 @@ class Decoder(nn.Module):
     self.decoder_layer = self.get_decoder_layers()
     self.norm_layer = self.get_norm_layer()
     if self.config.using_pipeline_parallelism:
-      pipeline_stage_module = self.get_pipeline_stage_module(self.decoder_layer[0])
+      pipeline_stage_module = self.get_pipeline_stage_module(self.decoder_layer)
       remat_policy = self.get_remat_policy()
       self.pipeline_module = pipeline.Pipeline(
           config=self.config, mesh=self.mesh, layers=pipeline_stage_module, remat_policy=remat_policy
@@ -421,9 +421,17 @@ class Decoder(nn.Module):
     )
     return scan_fn(config=cfg, mesh=mesh, name=metdata_axis_name, quant=self.quant, **kwargs)
 
-  def get_pipeline_stage_module(self, base_stage):
+  def get_pipeline_stage_module(self, decoder_blocks):
     """get pipeline stage module"""
+    def get_layer_to_pipeline(blocks, cfg):
+      if cfg.decoder_block == DecoderBlockType.DEEPSEEK:
+        return blocks[1] # return the sparse block
+      else:
+        return blocks[0]
+    
     cfg = self.config
+    base_stage = get_layer_to_pipeline(decoder_blocks, cfg)
+
     if cfg.set_remat_policy_on_layers_per_stage:
       policy = self.get_remat_policy()
       base_stage = self.set_remat_policy([base_stage], policy)[0]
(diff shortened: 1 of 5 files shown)


bvandermoon (commented):
Quick question to help my understanding while reviewing - why can we only pipeline 56 of the 58 sparse layers for deepseek?

MaxText/pyconfig.py
```diff
@@ -844,8 +844,8 @@ def pipeline_first_axis(raw_keys):
 
 
 def validate_deepseek_moe(raw_keys):
-  if raw_keys["decoder_block"] == "deepseek" and using_pipeline_parallelism(raw_keys):
-    raise ValueError("Currently we do not support DeepSeek MoE with pipeline parallelism.")
+  # if raw_keys["decoder_block"] == "deepseek" and using_pipeline_parallelism(raw_keys):
+  #   raise ValueError("Currently we do not support DeepSeek MoE with pipeline parallelism.")
```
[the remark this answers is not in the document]
gobbleturk: ofc deleted thanks

gobbleturk:
> Quick question to help my understanding while reviewing - why can we only pipeline 56 of the 58 sparse layers for deepseek?

we can pipeline 58 but the PP rank must divide the number of layers - 58 isn't a particularly divisor friendly number, I updated this in PR description

MaxText/pyconfig.py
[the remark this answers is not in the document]
gobbleturk: Done!

MaxText/layers/models.py
```diff
@@ -463,20 +470,46 @@ def __call__(
...
           )
+          if num_moe_layers_outside_pp > 0:
+            y, _ = self.scan_decoder_layers(cfg, moe_layer, num_moe_layers_outside_pp, "moe_layers", mesh)(
+                y,
+                decoder_segment_ids,
+                decoder_positions,
+                deterministic,
+                model_mode,
+            )
+        y = self.pipeline_module(y, decoder_segment_ids, decoder_positions, deterministic, model_mode, partition_spec=partition_spec)
```
bvandermoon, line 497: Just for my understanding, how are the 56 layers that need to be pipelined being specified here?

bvandermoon (commented):
LGTM, just one general question for my understanding

MaxText/layers/models.py
gobbleturk: 56 was an example, it can be set via `pipeline_parallel_layers` which was added in a previous PR

bvandermoon (approved):
LGTM pending the test failures

MaxText/layers/models.py
```diff
@@ -463,20 +470,46 @@ def __call__(
...
+      else:
+        y = self.pipeline_module(
+            y, decoder_segment_ids, decoder_positions, deterministic, model_mode, partition_spec=partition_spec
+        )
+        remaining_layers = self.config.num_decoder_layers - self.config.pipeline_parallel_layers
+        if remaining_layers > 0:
+          logical_axis_rules_pp_as_dp = maxtext_utils.logical_axis_rules_pp_act_as_dp(self.config.logical_axis_rules)
```
RissyRan, line 504: Do you think we could add some flexibility here? Instead of act as dp, could we act based on passed in config (not sure if doable). DP may not be helpful.

For instance, for DS-v3 (61 layers with 58 MoE layers) with pipeline_parallel_layers=56, we could run FSDP or EP from config bellow?

`pipeline_parallel_layers=56 && ici_fsdp=-1/ici_ep=-1`?

RissyRan (commented):
Thanks Matt! One more comment, can we add a unit test in moe_test.py like [this](https://github.com/AI-Hypercomputer/maxtext/blob/0add45d734a7d0bc04cec1d3793a95b4db81c18e/MaxText/tests/moe_test.py#L423, to ensure PP works as expected with assertion?

MaxText/layers/models.py
gobbleturk: with `pipeline_parallel_layers=56` there are only 2 moe layers that will be treated with PP replaced with DP. These two layers will still be sharded by the other sharding strategies - e.g. if config was EP_ICI=16 FSDP_ICI=16 PP_DCN=16, these two layers will be sharded as EP_ICI=16, FSDP_ICI=16 DP_DCN=16, e.g. still the weights are sharded 256 ways

I do like the idea of flexibility, but perhaps we can save this as potential future work?

RissyRan: > these two layers will be sharded as EP_ICI=16, FSDP_ICI=16 DP_DCN=16, e.g. still the weights are sharded 256 ways

SG! I thought it will only have DP sharding.

gobbleturk:
I ran into b/418313093 - PP + dropless EP do not work together.

Replacing this one with https://github.com/AI-Hypercomputer/maxtext/pull/1762 to initially support only PP + dropping, sorry to lose the comment history on this one

PR #1700, opened 2025-05-08
Resolve pylint e-class: E0102,E0606,E0611,E1102,E1111,E1120,E1121,E1123,E1135,E1136

SamuelMarks:
# Description

Linting is fun?

# Tests

N/A

# Checklist

WiP


MaxText/layers/moe.py
```diff
@@ -785,10 +785,9 @@ def dense_matmul(self, inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kerne
     # TODO @jacobplatin: allow Llama4 MoE to handle capacity factor > 0
     if self.config.capacity_factor > 0:
       # token dropping if needed
-      if self.config.model_call_mode != "inference":
+      if self.config.model_call_mode != "inference" and not is_llama4_decoder_layer:
```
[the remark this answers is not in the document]
SamuelMarks: `input_axis` is unused at this branch and only the `not is_llama4_decoder_layer` branch guarantees that `weights` is set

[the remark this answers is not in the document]
shralex: @jrplatin can you please review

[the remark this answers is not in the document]
jrplatin: This makes sense to me! @SamuelMarks  Can you please also remove this old comment: `# TODO @jacobplatin: allow Llama4 MoE to handle capacity factor > 0`? Thanks!

@RissyRan Might be good if you double check here as well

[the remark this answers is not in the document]
RissyRan: Yes, let's remove that TODO for now. Llama4 is using nocap/dropless implementation. We already have assertion here to avoid dropping [here](https://github.com/AI-Hypercomputer/maxtext/blob/b04c6daf531b866b3da814647fe068c2a53fb017/MaxText/pyconfig.py#L261-L262).

I would prefer not to add `not is_llama4_decoder_layer`; otherwise, it is kind of misleading users to think fall into `else` branch for llama4.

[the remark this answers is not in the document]
SamuelMarks: I suppose we could skip the extra branch and move `weights = self.reshape_and_update_weights(top_k_weights, top_k_indices)` to the only branch it's used to avoid the "Local variable 'weights' might be referenced before assignment" lint issue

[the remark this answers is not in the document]
RissyRan: LGTM! Thanks!

MaxText/layers/quantizations.py
```diff
@@ -263,13 +264,12 @@ def __call__(self, eqn, *args, **kwargs):
     k = jnp.asarray(k, comp_dtype)
     x = jnp.asarray(x, comp_dtype)
 
-    x_qdq = fp8_ops.in_qdq(comp_dtype, self.e4m3_dtype, x, self.input_scale.value, self.input_amax_history.value)
-    k_qdq = fp8_ops.in_qdq(comp_dtype, self.e4m3_dtype, k, self.kernel_scale.value, self.kernel_amax_history.value)
+    x_qdq = fp8_ops.in_qdq(self.e4m3_dtype, x, self.input_scale.value, self.input_amax_history.value)
```
RissyRan, line 267: Should we still keep the comp_dtype here? same as bellow.

RissyRan (commented):
LGTM! Just not sure if it's ok to remove `comp_dtype` from ` x_qdq = fp8_ops.in_qdq(comp_dtype, self.e4m3_dtype, x, self.input_scale.value, self.input_amax_history.value)`. Probably let's just keep as it is?

MaxText/layers/quantizations.py
SamuelMarks: Ohhhh https://github.com/google/flax/blob/main/flax/linen/fp8_ops.py#L243 has two types whereas mine only has one type. Let me check the dependencies to ensure they require this higher version then I'll revert my change here.

shralex (approved):

RissyRan (approved):

requirements.txt
```diff
@@ -14,7 +14,7 @@ google-cloud-monitoring
 google-api-core
 google-api-python-client
 grain[parquet]>=0.2.6
-flax>=0.8.0
+flax>=v0.10.6
```
bvandermoon, line 17: Is this and the corresponding `requirements_with_jax_stable_stack_0_5_2_pipreqs.txt` flax change related to linting? We have a PR including these changes already since it is needed for NNX: https://github.com/AI-Hypercomputer/maxtext/pull/1710

SamuelMarks: @bvandermoon Yes, see https://github.com/AI-Hypercomputer/maxtext/pull/1700#discussion_r2082647581 the lint picked up a change in this function's arity which happened at that version. My local dependency resolved to the lowest applicable, but it's wrong, so I've fixed it and now the lint passes.

bvandermoon: Ok makes sense, thanks @SamuelMarks. The requirements/contstraints files LGTM from the linting perspective

MaxText/layers/moe.py
```diff
@@ -910,11 +907,11 @@ def dense_matmul(self, inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kerne
       with jax.named_scope("w_sum"):
         if is_llama4_decoder_layer:
           weights = self.reshape_and_update_weights(jnp.ones_like(top_k_weights), top_k_indices)
-        output = jnp.einsum(
-            "BSEM,BSE -> BSM",
-            intermediate_layer,
-            weights,
-        ).astype(self.dtype)
+          output = jnp.einsum(
```
RissyRan, line 910: Extra space for the `output` that causes one test fails. Could you move it out of `if` statement?

MaxText/layers/moe.py
```diff
@@ -782,13 +780,12 @@ def dense_matmul(self, inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kerne
 
     cp, sub_seq = self.get_context_partition_and_sub_seq(seq_len)
 
-    # TODO @jacobplatin: allow Llama4 MoE to handle capacity factor > 0
     if self.config.capacity_factor > 0:
       # token dropping if needed
       if self.config.model_call_mode != "inference":
+        weights = self.reshape_and_update_weights(top_k_weights, top_k_indices)
```
RissyRan, line 786: We saw an exhausted resource issue in the test. If this still happens in tests, could you remove `weights` back, and then we could follow it up with a new PR. Thank you!

MaxText/layers/moe.py
SamuelMarks: That's a weird one; ok I manually adjusted spacing but it needs to be in the `if` statement otherwise `weights` isn't guaranteed to be set

PR #1701, opened 2025-05-08
Micro optimisations (e.g., if only one line is used only read one line with `jsonlines`)

SamuelMarks:
# Description

Various tiny micro optimizations.

# Tests

N/A

# Checklist


MaxText/tests/forward_pass_logit_checker.py
```diff
@@ -95,12 +95,12 @@ def main(config, test_args):  # pylint: disable=W0621
...
     state, _ = maxtext_utils.setup_decode_state(model, config, rng1, mesh, None)
 
-  if test_args.golden_logits_path == "":
+  if not test_args.golden_logits_path:
     input_golden_data_path = os.path.join(PKG_DIR, "test_assets", f"golden_data_{config.model_name}.jsonl")
   else:
     input_golden_data_path = test_args.golden_logits_path
   with jsonlines.open(input_golden_data_path, "r") as f:
-    golden_data = list(f)
+    golden_data = tuple(f)
```
shralex, line 103: @khatwanimohit is there always only one line in this file, and if so why do we need a for loop in line 105 ?

shralex (changes requested):
Hi Samuel, thank you for this PR. A few comments:
1. Come golden_data files have more than 1 line. lets revert the changes around this.
2. Lets consider closing this PR and moving the lint-related changes to one of the linting PRs. 
3. For PRs that change logic and aren't trivial, there should be a PR description explaining the motivation, what's being done, etc. Please see the PR template.

SamuelMarks:
> Hi Samuel, thank you for this PR. A few comments:
> 
>     1. Some golden_data files have more than 1 line. lets revert the changes around this.
> 
>     2. Lets consider closing this PR and moving the lint-related changes to one of the linting PRs.
> 
>     3. For PRs that change logic and aren't trivial, there should be a PR description explaining the motivation, what's being done, etc. Please see the PR template.

Ok I moved all the lint and `os.path.join` stuff out of this PR. Now it just does that one micro-optimisation.

SamuelMarks:
@shralex Ok I moved all the lint and `os.path.join` stuff out of this PR. Now it just does that one micro-optimisation.

MaxText/tests/forward_pass_logit_checker.py
shralex: @SamuelMarks lets revert this one please -- see for loop below.

shralex (changes requested):

hengtaoguo (approved):
Upon checking, golden_data_grpo/vit/sft files are all one line data for now, and the tests were using [0] to get the first element. This optimization seems OK to me, thanks for the change!

bvandermoon:
> Upon checking, golden_data_grpo/vit/sft files are all one line data for now, and the tests were using [0] to get the first element. This optimization seems OK to me, thanks for the change!

@shralex @hengtaoguo These files are always intended to a be a single line, right? I am wondering if the way Hengtao had it in the `vision_encoder_test` is actually more future-proof

hengtaoguo:
> > Upon checking, golden_data_grpo/vit/sft files are all one line data for now, and the tests were using [0] to get the first element. This optimization seems OK to me, thanks for the change!
> 
> @shralex @hengtaoguo These files are always intended to a be a single line, right? I am wondering if the way Hengtao had it in the `vision_encoder_test` is actually more future-proof

Thank you Branden! I assume these files are currently single-line, probably due to the large data point size (a single line in sft logits is more than 74MB). For Gemma3 ViT embeddings, I'm also holding off on adding more test examples because the embedding size was too big. 

And I also feel that looping through all lines would be more convenient than strictly taking the first line, and we have the flexibility to revert the change if we want to expand one specific unit test with more examples...

SamuelMarks:
@hengtaoguo Trivial to add support for multiline later but for now only one line is being used so a potentially large waste is now avoided. This can also be optimised further with [pysimdjson](https://pypi.org/project/pysimdjson) which might be especially noticeable as these files become larger.

Also probably worth looking through [grain](https://github.com/google/grain) / [array_record](https://github.com/google/array_record) code to find similar optimisation opportunities…

PR #1712, opened 2025-05-09
Inference Support in sparse_matmul

gpolovets1:
# Description

This PR provides support for inference workloads with sparse_matmul. The main difference is that Inference does not use batch sharding (not possible for Prefill for instance) while the sparse_matmul logic was originally created assuming that the batch was sharded.
Changes also include a fix for local_permute which counts just the data to be processed and fixes a bug where the last assigned expert recieves the padding assignments.
Changes also support using AQT-quantized checkpoints.
There was also a small fix to support TP + EP sharding.

- these changes are needed to explore sparse_matmul techniques such as ragged_dot and Megablox which previously OOM'd due to a bug.
- After the PR, the prefill and decode latencies drop from 6.48s/6.08s to 346ms/78ms respectively (see b/411762258#comment7).
- I am still finding that ragged_all_to_all performs rather slowly (compared to all-gather) so the investigation may continue separately.

FIXES: b/406081672

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/layers/moe.py
+++ b/MaxText/layers/moe.py
@@ -354,7 +354,10 @@ class RoutedMoE(nn.Module):
     # sort inputs for number of selected experts
     sorted_inputs = jnp.take(inputs_2d, indices=sorted_indices, axis=0).astype(self.dtype)
     group_size = jnp.bincount(flatten_selected_experts, length=self.num_experts)
-    return sorted_inputs, sorted_selected_experts, weights, group_size
+    # Return the experts for each sorted input.
+    expert_indices = jnp.arange(self.num_experts)
+    sorted_experts = jnp.repeat(expert_indices, repeats=group_size, total_repeat_length=flatten_selected_experts.shape[0])
+    return sorted_inputs, sorted_selected_experts, weights, group_size, sorted_experts
 
   def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size, sequence_length):
     """Unpermute tokens to original order and combine weights."""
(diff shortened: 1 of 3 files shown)


MaxText/layers/moe.py
```diff
@@ -357,29 +361,115 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
-      """Sort inputs by expert within each shard."""
-      local_id = jax.lax.axis_index("expert")
-      local_sizes = jax.lax.dynamic_slice_in_dim(
-          global_group_sizes, local_id * local_expert_size, local_expert_size, axis=1
-      ).reshape(-1)
+
+  @staticmethod
+  def local_permute(inputs, global_group_sizes, local_expert_size,
+                    shard_index, is_offset=False, global_sorted_experts=None):
+    """Sort inputs by expert within each shard."""
```
jrplatin, line 368: Can you add docstrings with args / return values / shaping / typing if possible too any of the major functions you are adding/updating?

MaxText/layers/moe.py
```diff
@@ -357,29 +361,115 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+    output_offsets = transform_array( 
+      all_shards_group_sizes, shard_id, TransformStrategy.OUTPUT_OFFSET, is_batch_sharded)
+    recv_sizes = transform_array(
+      all_shards_group_sizes, shard_id, TransformStrategy.RECV_SIZE, is_batch_sharded)
+    return input_offsets, send_sizes, output_offsets, recv_sizes
 
-    def gmm(inputs, kernel, group_sizes):
+  def sparse_matmul(self, inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kernel, wo_kernel):
+    def gmm(inputs, kernel, group_sizes, expert_assignments):
+      tile_size = (512, 1024, 1024)  # (m, k, n)
```
jrplatin, line 465: Is this something that can be tuned? If so, we might want to make a note of this / abstract it out to the config

MaxText/layers/moe.py
```diff
@@ -403,71 +493,54 @@ def gmm(inputs, kernel, group_sizes):
             rhs_quantize_dtype=rhs_quantize_dtype,
         )
       else:
-        if self.quant is not None:
-          raise NotImplementedError("Quantization is not yet supported with ragged_dot, please set" " megablox=True")
+        rhs_inputs = kernel
+        if isinstance(kernel, QTensor):
+          assert not (kernel.bias or kernel.sparsity_mask or len(kernel.scale) > 1), "Unsupported usecase for ragged_dot with quantized kernel."
```
jrplatin, line 498: Why is this check needed?

jrplatin (approved):
I didn't dig too, too deep into the core logic, but at a high-level, things LGTM

MaxText/layers/moe.py
```diff
@@ -357,29 +361,115 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+    all_shard_local_sizes = jax.lax.dynamic_slice_in_dim(
+        global_group_sizes, shard_index * local_expert_size, local_expert_size, axis=1
+    )
+    local_sizes = all_shard_local_sizes.reshape(-1)
+    local_group_size = jnp.sum(all_shard_local_sizes, axis=0)
+    
+    # In this case, the data that needs to be processed by the local shard
+    # does not start from row 0 but actually starts at
+    # jnp.concatenate((jnp.array([0]), jnp.cumsum(local_group_sizes[:-1]))[shard_id]
+    # This happens for instance if the data is not produced via a ragged_all_to_all.
```
RissyRan, line 378: What do you mean by `data is not produced via a ragged_all_to_all`?

MaxText/layers/moe.py
```diff
@@ -357,29 +361,115 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+    else:
       base_indices = jnp.mod(jnp.arange(local_sizes.shape[0]), local_expert_size)
       expert_indices = jnp.repeat(base_indices, local_sizes, total_repeat_length=inputs.shape[0])
-      sorted_indices = jnp.argsort(expert_indices)
-      sorted_inputs = jnp.take(inputs, indices=sorted_indices, axis=0)
-      group_size = jnp.bincount(expert_indices, length=local_expert_size)
-      return sorted_inputs, sorted_indices, group_size
+    sorted_indices = jnp.argsort(expert_indices)
+    sorted_inputs = jnp.take(inputs, indices=sorted_indices, axis=0)
+    sorted_experts_ids = expert_indices[sorted_indices]
```
RissyRan, line 391: Thanks for the change! I will pull down and see if this improves the training side.

MaxText/layers/moe.py
```diff
@@ -403,71 +493,54 @@ def gmm(inputs, kernel, group_sizes):
...
+    # Useful for Interleaved Inference where Prefill always has batch_size=1 while Decode
+    # can have batch_size > 1.
+    try:
+      is_batch_sharded_by_expert = 'expert' in tuple(filter(
+        lambda tup: tup[0] == "activation_batch_moe", self.config.logical_axis_rules)
+        )[0][1]
+    except:
+      is_batch_sharded_by_expert = False
+    if is_batch_sharded_by_expert and inputs.shape[0] > 1:
+      batch_logical_axis = "activation_batch_moe"
```
RissyRan, line 530: Shall we align with [activation_batch](https://github.com/AI-Hypercomputer/maxtext/blob/75c8ef6aa33bd5192f88b5c35786a624420866c3/MaxText/configs/base.yml#L277C1-L278C1) from base.yml?

MaxText/layers/moe.py
```diff
@@ -403,71 +493,54 @@ def gmm(inputs, kernel, group_sizes):
...
     # Currently, we only support data and tensor parallelism with Megablox.
     # We all gather the input activations over tensor parallelism to follow strategy
     # in https://parsa.epfl.ch/course-info/cs723/papers/Megatron.pdf.
-    input_partition_pspec = nn.logical_to_mesh_axes(("activation_batch", None, None))
-    gate_logits_pspec = nn.logical_to_mesh_axes(("activation_batch", None, None))
+    
+    # Check if the batch should be sharded by expert and whether the batch_size supports this.
+    # Useful for Interleaved Inference where Prefill always has batch_size=1 while Decode
+    # can have batch_size > 1.
+    try:
```
RissyRan, line 523: Thinking if we could simplify this logic a little bit. No better idea yet, let me think about it.

MaxText/layers/moe.py
```diff
@@ -484,73 +557,110 @@ def transform_array(input_array, shard_id, strategy):
...
-        )
-        intermediate_output = jax.lax.ragged_all_to_all(
-            local_output,
-            output_shape,
-            input_offsets,
-            send_sizes,
-            output_offsets,
-            recv_sizes,
-            axis_name=axis_name,
+        assert sorted_selected_experts.shape[0] == original_inputs_first_dim, "original_inputs_first_dim does not match the original tensor shape!"
```
RissyRan, line 619: Could we update to raise condition instead of assert? assertion could be disabled globally and mainly for test/debug purpose.

RissyRan (commented):
Thanks for the change and test! I have some initial comments, and would like to explore a little bit quickly locally and then take another look. Thank you!

MaxText/layers/moe.py
RissyRan: +1. Also, Let's change back to the raise condition instead of assertion for the similar reason for another assertion?

MaxText/layers/moe.py
RissyRan: Yes, this can be tuned for kernel part. Yeah, for convenience, let's add into the base.yml with something like? 

tile_batch_seq: 512
tile_activation_dim: 1024 (could be related to model or mlp dim)
tile_weight_dim: 1024 (could be related mlp or model dim)

based on: https://github.com/AI-Hypercomputer/maxtext/blob/75c8ef6aa33bd5192f88b5c35786a624420866c3/MaxText/layers/moe.py#L395C19-L395C51

MaxText/tests/moe_test.py
```diff
@@ -369,6 +369,151 @@ def test_megablox(self):
     actual_output, _ = self.get_moe_output(variables, hidden_states, cfg, mesh)
     self.assertTrue(jax.numpy.allclose(expected_output, actual_output, rtol=1e-02, atol=1e-02, equal_nan=False))
 
+  def test_local_permute_no_offset(self):
```
RissyRan, line 372: Nit: could you add those tests at the end? It seems break up the test_megablox and test_ragged_dot tests for readability purpose.

MaxText/layers/moe.py
RissyRan: +1 Thank you!

MaxText/layers/moe.py
```diff
@@ -357,29 +361,115 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+          array_with_zeros = jnp.concatenate((zero_row, input_array), axis=0)
+          cumulated_array = jnp.cumsum(array_with_zeros, axis=0, dtype=input_array.dtype)
+          return cumulated_array[shard_id]
+        elif strategy == TransformStrategy.RECV_SIZE:
+          # Received size in the traget output
+          return input_array[:, shard_id]
+        else:
+          raise ValueError(f"Unknown tranform array strategy: {strategy}")
+      
+      # If the batch is unsharded then the we send the same information to all other shards.
```
RissyRan, line 429: > then the we send the same information to all other shards

Are you sending the data/activation as duplicate to other devices? Have you tried to use `jax.device_put` instead of all_to_all?

MaxText/layers/moe.py
```diff
@@ -484,73 +557,110 @@ def transform_array(input_array, shard_id, strategy):
...
-      layer_w1 = gmm(x, w1, group_sizes)
+      layer_w1 = gmm(x, w1, group_sizes, selected_experts)
       layer_w1 = checkpoint_name(layer_w1, "mlpwi_1")
       layer_act = linears._convert_to_activation_function(self.config.mlp_activations[0])(layer_w0)
       intermediate_layer = jnp.multiply(layer_act, layer_w1)
-      intermediate_output = gmm(intermediate_layer, wo, group_sizes)
+      intermediate_output = gmm(intermediate_layer, wo, group_sizes, selected_experts)
       intermediate_output = checkpoint_name(intermediate_output, "mlpwo")
-      if self.get_tensor_parallelism_size() > 1:
+      if num_expert_parallelism > 1:
```
RissyRan, line 614: This should be related to tensor parallelism, not expert parallelism.

RissyRan (commented):
I was checking some internal values so pull the changes locally, and test this expert sharding [unit test](https://github.com/AI-Hypercomputer/maxtext/blob/75c8ef6aa33bd5192f88b5c35786a624420866c3/MaxText/tests/moe_test.py#L423. The original tolerance seems failed ([here](https://paste.googleplex.com/6117093541675008#l=506)), but outputs are not far away. Could you double check on your side? Thanks!

MaxText/layers/moe.py
gpolovets1: I think AQT may support fancier quantization schemes than just applying a scaling factor but I am calling out that this is not supported.

MaxText/layers/moe.py
gpolovets1: I am sending the same slice of data to all other devices so I think in this case device_put would not work.

MaxText/layers/moe.py
gpolovets1: Updated.

MaxText/layers/moe.py
gpolovets1: Updated in latest commit.

MaxText/layers/moe.py
gpolovets1: Updated, thanks!

MaxText/layers/moe.py
gpolovets1: Updated.

MaxText/tests/moe_test.py
gpolovets1: Updated.

MaxText/layers/moe.py
```diff
@@ -479,78 +589,123 @@ def transform_array(input_array, shard_id, strategy):
...
-        local_output = jnp.take(intermediate_output, indices=jnp.argsort(local_sorted_indices), axis=0)
+      if num_expert_parallelism > 1:
         original_inputs_first_dim = batch_size * sequence_length * self.config.num_experts_per_tok
-        output_shape = jnp.zeros((original_inputs_first_dim, self.config.emb_dim), dtype=local_output.dtype)
-        input_offsets, send_sizes, output_offsets, recv_sizes = get_all_to_all_params(
-            jnp.transpose(all_shards_group_sizes), local_expert_size, self.get_expert_parallelism_size()
+        if sorted_selected_experts.shape[0] != original_inputs_first_dim:
+          raise ValueError("original_inputs_first_dim does not match the original tensor shape!")
+        output_shape = jnp.zeros(
+            (original_inputs_first_dim, self.config.emb_dim // self.get_tensor_parallelism_size()),
```
RissyRan, line 664: `self.config.emb_dim // self.get_tensor_parallelism_size()` nice catch!

MaxText/layers/moe.py
```diff
@@ -357,29 +360,152 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+      A tuple containing:
+        sorted_inputs: Input data permuted local expert ID.
+        sorted_indices: Indices used to permute the inputs.
+        local_group_size: Number of tokens assigned to each local expert on this
+          shard.
+        sorted_experts_ids: expert ID corrsponding to each token of the permuted inputs.
+    """
+
+    # Slice the count of local expert IDs in each batch shard.
+    all_shard_local_sizes = jax.lax.dynamic_slice_in_dim(
```
RissyRan, line 395: When rebase from head, could you help keep those shape info for some variables? I think I had some updates there. Thank you!

RissyRan (commented):
Thank you for the change!

MaxText/tests/moe_test.py
```diff
@@ -429,7 +429,7 @@ def test_megablox_expert_parallelism(self):
         dtype="bfloat16",
         megablox=True,
         sparse_matmul=True,
-        per_device_batch_size=4,
+        per_device_batch_size=8,
```
RissyRan, line 432: I noticed in the test I used `per_device_batch_size` as global batch size. I will handle it after your PR. Just FYI.

RissyRan (approved):
Great work! Thank you for those tests!

MaxText/layers/moe.py
```diff
@@ -357,29 +360,152 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+
+    def gmm(inputs, kernel, group_sizes, expert_assignments):
+      tile_size = (self.config.tile_batch_seq, self.config.tile_activation_dim, self.config.tile_weight_dim)
+      PAD_LENGTH = 512
       hs_shape = inputs.shape
       # pad length is the 1st dimension of tiling size in gmm call
-      pad_length = 512
-      if hs_shape[0] % pad_length:
-        pad_length = pad_length - hs_shape[0] % pad_length
+      assert (
```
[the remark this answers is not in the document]
gpolovets1: Previously fixed - thanks!

MaxText/layers/moe.py
```diff
@@ -357,29 +360,152 @@ def unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+    output_offsets = transform_array(all_shards_group_sizes, shard_id, TransformStrategy.OUTPUT_OFFSET, is_batch_sharded)
+    recv_sizes = transform_array(all_shards_group_sizes, shard_id, TransformStrategy.RECV_SIZE, is_batch_sharded)
+    return input_offsets, send_sizes, output_offsets, recv_sizes
+
+  def sparse_matmul(self, inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kernel, wo_kernel):
+    """Perform sparse matrix multiplication of inputs and Experts."""
+
+    def gmm(inputs, kernel, group_sizes, expert_assignments):
+      tile_size = (self.config.tile_batch_seq, self.config.tile_activation_dim, self.config.tile_weight_dim)
+      PAD_LENGTH = 512
```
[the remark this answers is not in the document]
gpolovets1: Updated.

MaxText/layers/moe.py
```diff
@@ -484,73 +594,118 @@ def transform_array(input_array, shard_id, strategy):
...
+              intermediate_output,
+              output_shape,
+              input_offsets,
+              send_sizes,
+              output_offsets,
+              recv_sizes,
+              axis_name=expert_axis_name,
+          )
+
+        output = self.unpermute(
```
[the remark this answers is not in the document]
gpolovets1: Nice catch - updated!

MaxText/tests/moe_test.py
```diff
@@ -446,6 +446,173 @@ def test_megablox_expert_parallelism(self):
...
+    rng1, rng2, rng3, rng4 = jax.random.split(rng, 4)
+    bs1_group_sizes = jax.random.randint(rng1, num_expert_parallelism, 0, 5)
+    bs2_group_sizes = jax.random.randint(rng2, num_expert_parallelism, 0, 5)
+    bs3_group_sizes = jax.random.randint(rng3, num_expert_parallelism, 0, 5)
+    bs4_group_sizes = jax.random.randint(rng4, num_expert_parallelism, 0, 5)
+    all_group_sizes = jnp.reshape(
+        jnp.stack((bs1_group_sizes, bs2_group_sizes, bs3_group_sizes, bs4_group_sizes)),
+        (num_expert_parallelism, num_expert_parallelism),
+    )
+    for expert_shard_id in range(num_expert_parallelism):
```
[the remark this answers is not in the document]
gpolovets1: Good suggestion - I guess originally it was a way of future-proofing that the logic of the code is not broken. But hardcoding also helps check the original logic. Updated the test case!

MaxText/tests/moe_test.py
gpolovets1: Ah thanks - let me also revert my local change to be of size 4 again.

PR #1717, opened 2025-05-11
Add comment that docker_upload_runner script needs bash version >= 4.2

wstcliyu:
# Description

`[[ ! -v VARIABLE ]]` only works on bash of version >= 4.2

The default Bash version on older macOS systems (specifically before macOS Catalina) is often 3.2, which does not support the -v option within the [[ ]] conditional construct. We should remind users of upgrading Bash version for this script to work.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/docker_upload_runner.sh
+++ b/docker_upload_runner.sh
@@ -22,6 +22,7 @@
 
 # Example command:
 # bash docker_upload_runner.sh CLOUD_IMAGE_NAME=${USER}_runner
+# This script needs bash version >= 4.2 to execute correctly.
 
 set -e
 


bvandermoon (commented):
Thanks @wstcliyu. This LGTM. Just want to understand what error you saw/how you found this?

wstcliyu:
Thanks @bvandermoon. This was found by a xWF engineer verifying Pathways user guide. But this is actually due to a MaxText script instead of Pathways.

bvandermoon (approved):

PR #1722, opened 2025-05-12
Fix get_input_data_sharding method name in trainers

SurbhiJainUSC:
# Description
Fix `get_input_data_shardings()` to `get_input_data_sharding()` as defined in `maxtext_utils` [here](https://github.com/AI-Hypercomputer/maxtext/blob/75c8ef6aa33bd5192f88b5c35786a624420866c3/MaxText/maxtext_utils.py#L56)

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/elastic_train.py
+++ b/MaxText/elastic_train.py
@@ -267,7 +267,7 @@ def train_loop(config, elastic_manager, state=None):
       block=True,
   )
 
-  input_data_shardings = maxtext_utils.get_input_data_shardings(config, mesh)
+  input_data_shardings = maxtext_utils.get_input_data_sharding(config, mesh)
   # Using while loop instead of a for loop because with elasticity
   # the step is restored back to the latest snapshot when a slice is lost
   while step < config.steps:
--- a/MaxText/experimental/rl/grpo_trainer.py
+++ b/MaxText/experimental/rl/grpo_trainer.py
@@ -806,7 +806,7 @@ def train_loop(config, config_inference, state=None):
       gcp_workload_monitor.start_performance_reporting_thread(performance_metric_queue)
 
   metric_logger = MetricLogger(writer, config)
-  input_data_shardings = maxtext_utils.get_input_data_shardings(config, mesh)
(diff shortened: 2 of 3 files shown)


bvandermoon (commented):
Thanks for the fix @SurbhiJainUSC. Are there unit tests that can be added that could have caught this?

SurbhiJainUSC:
> Thanks for the fix @SurbhiJainUSC. Are there unit tests that can be added that could have caught this?

I am not sure about `grpo_trainer` and `elastic_train`, but we do have Airflow tests for `sft_trainer` that caught this issue.

bvandermoon:
> > Thanks for the fix @SurbhiJainUSC. Are there unit tests that can be added that could have caught this?
> 
> I am not sure about `grpo_trainer` and `elastic_train`, but we do have Airflow tests for `sft_trainer` that caught this issue.

Can you check with the @A9isha and @lukebaumann on if we can add unit tests for `grpo_trainer` and `elastic_train` that would catch this type of error? Calling a non-existent method should be catchable, unless I am missing something

Also, `sft_trainer_correctness` is just for Airflow, right? Does `train.py` have an equivalent test that can be reused for this type of issue?

bvandermoon (approved):

gobbleturk (approved):

SurbhiJainUSC:
> > > Thanks for the fix @SurbhiJainUSC. Are there unit tests that can be added that could have caught this?
> > 
> > 
> > I am not sure about `grpo_trainer` and `elastic_train`, but we do have Airflow tests for `sft_trainer` that caught this issue.
> 
> Can you check with the @A9isha and @lukebaumann on if we can add unit tests for `grpo_trainer` and `elastic_train` that would catch this type of error? Calling a non-existent method should be catchable, unless I am missing something
> 
> Also, `sft_trainer_correctness` is just for Airflow, right? Does `train.py` have an equivalent test that can be reused for this type of issue?

Anisha will work on adding Airflow tests for GRPO.

PR #1724, opened 2025-05-12
Support vision attention for Llama4

aireenmei:
# Description

Add more options to the existing Attention class to support the attention in vision encoder:
* Full attention (non-causal)
* use bias in qkvo projection
* add is_vision flag to attention module for vision specific settings
Future item: we may implement a separate VisionAttention, as the vision dimensions are different from text, which increases the complexity if reuse the current attention module. We are working on the design.

# Tests
Test by comparing outputs with Llama4 VisionAttention module
Screenshot 2025-05-14 at 3 50 43 PM
Currently dot_product attention works as expected, there are some issues with flash attention. Will enable flash attention in future PRs

# Checklist

### DIFF
--- a/MaxText/input_pipeline/_input_pipeline_utils.py
+++ b/MaxText/input_pipeline/_input_pipeline_utils.py
@@ -101,14 +101,14 @@ def is_conversational(features, data_columns):
   return False
 
 
-def apply_chat_template(example, tokenizer, data_column_name):
+def apply_chat_template(example, tokenizer_model, data_column_name):
   """Formats conversational data by applying the tokenizer's chat template
   and identifying prompt/completion segments.
 
   Args:
     example: A dictionary containing conversational data. It is expected to have a key
       specified by `data_column_name` that holds a list of messages.
-    tokenizer: The tokenizer instance associated with the language model,
+    tokenizer_model: The tokenizer instance associated with the language model,
       which contains the specific chat template.
     data_column_name: The name of the column in the `example` dictionary
       that contains the list of messages.
@@ -128,16 +128,16 @@ def apply_chat_template(example, tokenizer, data_column_name):
     for message in example[data_column_name]:
       if message["role"] == "user":
         prompt = message
-        prompt_in_chat_template = tokenizer.apply_chat_template([prompt], add_generation_prompt=False, tokenize=False)
+        prompt_in_chat_template = tokenizer_model.apply_chat_template([prompt], add_generation_prompt=False, tokenize=False)
         messages.append(prompt_in_chat_template)
         is_prompt.append(True)
       elif message["role"] == "assistant":
-        prompt_completion_tokens = tokenizer.apply_chat_template(
+        prompt_completion_tokens = tokenizer_model.apply_chat_template(
             [prompt, message], add_generation_prompt=False, tokenize=True
         )
-        prompt_tokens = tokenizer.apply_chat_template([prompt], add_generation_prompt=False, tokenize=True)
+        prompt_tokens = tokenizer_model.apply_chat_template([prompt], add_generation_prompt=False, tokenize=True)
         completion_tokens = prompt_completion_tokens[len(prompt_tokens) :]
-        completion_in_chat_template = tokenizer.decode(completion_tokens, skip_special_tokens=False)
+        completion_in_chat_template = tokenizer_model.decode(completion_tokens, skip_special_tokens=False)
         messages.append(completion_in_chat_template)
         is_prompt.append(False)
   except ValueError as e:
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -55,10 +55,11 @@ from MaxText.layers.quantizations import AqtQuantization as Quant
 
 
 class AttentionType(enum.Enum):
-  GLOBAL = "global"
+  GLOBAL = "global"  # default, with causality
   LOCAL_SLIDING = "local_sliding"
   CHUNK = "chunk"
   MLA = "mla"
+  FULL = "full"
 
 
 # Used to pass in splash attention block sizes from config.
@@ -389,7 +390,7 @@ class AttentionOp(nn.Module):
 
     causal_mask = None
     # We enforce causality except for AUTOREGRESSION
-    if model_mode != MODEL_MODE_AUTOREGRESSIVE:
+    if model_mode != MODEL_MODE_AUTOREGRESSIVE and self.attention_type != AttentionType.FULL:
       mask_shape = (q_seq_len, kv_seq_len)
       # row_ids indicates the position of query
       # col_ids indicates the position of kv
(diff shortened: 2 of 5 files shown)


MaxText/layers/attentions.py
```diff
@@ -710,7 +711,10 @@ def tpu_flash_attention(
     )
 
     mask_shape = (self.config.max_target_length, self.config.max_target_length)
-    mask = splash_attention_mask.CausalMask(shape=mask_shape)
+    if self.attention_type == AttentionType.FULL:
```
hengtaoguo, line 714: Does this mean future flash attention mask will be generated here in the branch?

hengtaoguo (approved):

MaxText/layers/attentions.py
```diff
@@ -710,7 +711,10 @@ def tpu_flash_attention(
...
+      mask = splash_attention_mask.FullMask(mask_shape)
```
ZhaoyueCheng, line 715: since we already have the corresponding full mask generation for splash attention can we try if we can directly enable splash attention for the vision transformer?

ZhaoyueCheng (approved):

MaxText/layers/attentions.py
aireenmei: We have been using this branch to generate mask for flash attention. Before this change, we by default started with a causal mask, and add modifications such as chunks on top of it. After this change, we started with a full mask when attention type is full, and likely won't add more modifications due to the conditions for modifications will be False

MaxText/layers/attentions.py
aireenmei: I added Full mask for flash attention here, but still hit other issues, mainly related to the shape not meeting the block_size requirements for flash attention. Will keep working on it.

MaxText/tests/check_llama4_layers.py
```diff
@@ -337,5 +347,222 @@ def __init__(self):
...
+      self,
+      hidden_states: torch.Tensor,
+      freqs_ci: torch.Tensor,
+      attention_mask: Optional[torch.Tensor] = None,
+      past_key_value: Optional[torch.Tensor] = None,
+      **kwargs,
+  ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
+    input_shape = hidden_states.shape[:-1]
+    hidden_shape = (*input_shape, -1, self.head_dim)
+    print(hidden_shape)
```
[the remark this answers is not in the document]
aireenmei: Good catch. Fixed

MaxText/layers/attentions.py
ZhaoyueCheng: yeah makes sense! Thanks Aireen!

MaxText/layers/attentions.py
```diff
@@ -1208,6 +1212,7 @@ class Attention(nn.Module):
   ragged_block_size: int = 256
   use_qk_norm: bool = False
   query_pre_attn_scalar: float | None = None
+  use_bias_in_projection: bool = False
```
gagika, line 1215: nit: `use_bias_in_projections` (plural) and please add comment:
looks like it's being applied to many projections (q, k, v) and output projections.

MaxText/layers/attentions.py
```diff
@@ -1394,7 +1403,15 @@ def apply_rotary_embedding(self, inputs: Array, inputs_positions: Array, name: s
 
     rope_type = self.config.rope_type.lower()
     rope_use_scale = self.config.rope_use_scale
-    if self.config.model_name.startswith("llama3.1") or rope_type.startswith("llama3.1"):
+    if self.config.model_name.startswith("llama4") and self.name == "self_attention_vision":
```
gagika, line 1406: could you replace `self.name == "self_attention_vision"` with an explicit boolean or enum parameter in the Attention?

e.g. `is_vision_attention`?

Another option is to have VisionAttention as a subclass of Attention that overrides 

class VisionAttention(Attention):
    def apply_rotary_embedding ...

MaxText/layers/attentions.py
aireenmei: fixed

MaxText/layers/attentions.py
aireenmei: I updated to use enum RopeType to replace rope_type to model config. So we can set rope_type at attention level instead of setting one at model level.

aireenmei: I see you are OOO. Could @RissyRan review this change, also changes in Llama4 and Deepseek?

MaxText/configs/models/deepseek2-16b.yml
```diff
@@ -40,8 +40,6 @@ kv_lora_rank: 512
 qk_nope_head_dim: 128
 qk_rope_head_dim: 64
 v_head_dim: 128
-# RoPE
-rope_type: "yarn"
```
RissyRan, line 44: I feel this remove the flexibility of rope in models. For instance, how we apply llama3.1 type to a new future deepseek model?

MaxText/layers/attentions.py
```diff
@@ -407,18 +416,18 @@ def generate_attention_mask(
     elif causal_mask is not None:
       output_mask = causal_mask
 
-    is_seq_len_greater_than_chunk_window = False
```
RissyRan, line 410: Be careful on this as output_mask is updated in the block `self.attention_type == AttentionType.LOCAL_SLIDING and output_mask is not None`. Worth to double check the logic here.

RissyRan (commented):
Thanks Aireen! Just minor comments.

MaxText/configs/models/deepseek2-16b.yml
aireenmei: I reverted this change as discussed offline. Thanks for your insights!

MaxText/layers/attentions.py
aireenmei: I discussed with @RissyRan offline, I ended up adding an is_vision flag to the attention module. We may add a visionAttention module in the future (see the future item in PR description). But for now we use this way to get an early version to work.

RissyRan (approved):

MaxText/input_pipeline/_input_pipeline_utils.py
aireenmei, line 1: This is for solving pylint error

MaxText/layers/embeddings.py
aireenmei, line 1: This is for solving presubmit error about API arg counts

MaxText/input_pipeline/_input_pipeline_utils.py
hengtaoguo: Thanks! There's also another PR noticed the same issue. @gobbleturk  or @SurbhiJainUSC  are working on a fix soon: https://github.com/AI-Hypercomputer/maxtext/pull/1762/files#diff-f4632952887d83be3fc9646878680a77dab4b70af4159c7218e71b860754a55e

PR #1728, opened 2025-05-13
convert RMSNorm to NNX

cgarciae:
# Description

### DIFF
--- a/MaxText/layers/initializers.py
+++ b/MaxText/layers/initializers.py
@@ -17,6 +17,7 @@
 from typing import Callable, Tuple, Union
 
 from flax import linen as nn
+from flax import nnx
 import jax
 from MaxText import common_types
 
@@ -42,3 +43,12 @@ def nd_dense_init(scale, mode, distribution):
     return fn(key, shape, dtype)
 
   return init_fn
+
+def variable_to_logically_partitioned(variable: nnx.VariableState):
+  metadata = variable.get_metadata()
+  return nn.LogicallyPartitioned(  # type: ignore[wrong-keyword-args]
+      variable.value,
+      variable.sharding,  # type: ignore[arg-type]
+      mesh=metadata.get("mesh"),
+      rules=metadata.get("rules"),
+  )
--- a/MaxText/layers/models.py
+++ b/MaxText/layers/models.py
@@ -42,6 +42,7 @@ ScanIn = common_types.ScanIn
 Embed = embeddings.Embed
 Attention = attentions.Attention
 RMSNorm = normalizations.RMSNorm
+rms_norm = normalizations.rms_norm
 PositionalEmbedding = embeddings.PositionalEmbedding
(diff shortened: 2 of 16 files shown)


bvandermoon (commented):
Do you know why the full test suite didn't run for the checks?

MaxText/layers/normalizations.py
```diff
@@ -17,35 +17,68 @@
...
   """RMS normalization."""
 
-  epsilon: float = 1e-6
-  dtype: Any = jnp.float32
-  weight_dtype: Any = jnp.float32
-  kernel_axes: Tuple[Optional[str], ...] = ()
-  scale_init: Initializer = nn.initializers.ones
+  def __init__(
+      self,
+      features: int,
```
bvandermoon: Is it possible to make this more descriptive like in https://github.com/AI-Hypercomputer/maxtext/pull/1840?

cgarciae: Changing it to `num_features`.

bvandermoon: Thanks @cgarciae. As a nit, should it be `features_shape` to match the style from #1840? Not 100% sure though since this is strictly an int

cgarciae: In this case its not a shape (tuple). I'm also not super convinced about the shape semantics in `DenseGeneral` because you can also pass and `int`.

bvandermoon: Agreed. Let me know if you have any name suggestions. Happy to update if we have something better

MaxText/layers/initializers.py
```diff
@@ -39,3 +40,12 @@ def init_fn(key, shape, dtype, in_axis, out_axis):
     return fn(key, shape, dtype)
 
   return init_fn
+
+def variable_to_logically_partitioned(variable: nnx.VariableState):
```
cgarciae, line 44: @bvandermoon I'm thinking of moving this function to `initializers.py` since its shared but maybe there is a better place?

bvandermoon: I'm ok with it here. Could also make a separate file like `variable_partitions.py` or something which would probably be better IMO

MaxText/layers/normalizations.py
bvandermoon: Quick follow up on this - think this update may have been missed. Good to see the tests are now passing though

MaxText/layers/models.py
```diff
@@ -380,7 +381,7 @@ def get_norm_layer(self):
         DecoderBlockType.SIMPLE_MLP,
         DecoderBlockType.LLAMA4,
     ):
-      return RMSNorm
+      return functools.partial(rms_norm, num_features=num_features)
```
bvandermoon, line 389: Why does this need to be changed to functools.partial?

bvandermoon (approved):

richjames0 (approved):
lgtm

PR #1731, opened 2025-05-13
Adding ml_goodput_measurement package to jax stable stack requirements

Perseus14:
# Description

Add `ml_goodput_measurement` library to jax stable stack requirements file

FIXES: b/416288633 (partially)

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/requirements_with_jax_stable_stack.txt
+++ b/requirements_with_jax_stable_stack.txt
@@ -22,3 +22,4 @@ pathwaysutils==0.1.1
 google-api-python-client
 omegaconf
 jaxtyping
+ml-goodput-measurement==0.0.10


bvandermoon (approved):
LGTM pending the Airflow test. @parambole can you please also take a look?

requirements_with_jax_stable_stack.txt
```diff
@@ -22,3 +22,4 @@ pathwaysutils==0.1.1
 google-api-python-client
 omegaconf
 jaxtyping
+ml-goodput-measurement==0.0.10
```
SurbhiJainUSC, line 25: Instead of pinning a specific version, can we just keep it `ml-goodput-measurement`, so that it always install the latest package?

bvandermoon: @dipannita08 please also let us know your thoughts on this

parambole, line 25: My concern with this is around build time with this dependency in the requirements file. Given that it forces pip to check the compatibility of all of the dependencies including the ones as part of `ml_goodput_measurement` this can cause an increase in build time.

dipannita08: Given that this package is (currently) rapidly being updated, pinning it allows us incrementally test upgrades and reduces the possibility of regressions with automatic upgrades since we don't yet have E2E MaxText tests for Goodput.

Having said that, the library is meant to be backward compatible so in ideal scenarios this situation should not occur. 

I'm not yet sure why pinning increase build times. I was under the impression that pinning the release version should help reduce the recursive searches and reduce build time.

Perseus14: I tested it by not pinning a version and pip seems to ignore it as a previous version (v0.0.5) exists. Jax stable stack base image seems to have an existing older version of `ml-goodput-measurement`

Additionally in v0.0.10 the library fixes a few bugs and supports pushing goodput to GCM compared to the older versions where it only supports tensorboard.

Perseus14: I saw a build time increase of about 100s. Is that acceptable?

parambole: @Perseus14 - Did you get this from building the image locally? The bigger concern is that the dependencies the goodput libraries have been seen to interfere with (resolving versions) of the existing libraries in the requirements file.

Can you please share more about the requirements of the `goodput` library, and if they are common to what is in the requirements file, we can potentially have them at different stages?

Perseus14: I tested them on both cloudtop and TPU VM, the increase is about 100s in both.

These are the dependencies taken from [here](https://github.com/AI-Hypercomputer/ml-goodput-measurement/blob/main/pyproject.toml). I suspect the `google-api-core` and `google-cloud-logging` might be the cause

```
  "google-api-core>=2.24.1",
  "google-cloud-logging>=3.5.0",
  "google-cloud-monitoring>=2.20.0",
  "numpy",
  "requests",
  "scipy",
  "tensorboardx",
  "urllib3",
```

PR #1736, opened 2025-05-13
Moving the GCS copy command to the script

lukebaumann:
# Description

Recently, many test_assets were moved to a GCS bucket. The Dockerfile was modified to copy the files from the bucket as part of `docker build ...` but Docker cannot copy because the request is unauthenticated. This PR moves the copy into the script `docker_upload_runner.sh` and also restores the separate layers in the Dockerfile for copying the assets and the rest of the code to ensure fast rebuild times. 

I added `MaxText/test_assets/.gitignore` to ignore the downloaded files. I do see two files are part of the repo but overwritten with different data from the bucket.

```
golden_data_grpo_default.jsonl
golden_data_llama4-17b-128e.jsonl
```

After building the runner, these two files will be modified.

# Tests

Manually created a maxtext runner image successfully via `bash docker_upload_runner.sh`

# Checklist


maxtext_runner.Dockerfile
```diff
@@ -8,10 +8,9 @@ FROM $BASEIMAGE
...
 WORKDIR /deps
 
-# Copy all files from local workspace into docker container. This includes
-# MaxText/test_assets which contain some of the reference "golden" data.
-COPY . .
-
-# Download other test assets from GCS into /deps/MaxText/test_assets
-RUN gsutil -m cp gs://maxtext-test-assets/* MaxText/test_assets
+# Copy assets separately
+COPY assets/ .
```
bvandermoon, line 12: Why does this speed up build times?

docker_upload_runner.sh
```diff
@@ -44,6 +44,9 @@ if [[ ! -v CLOUD_IMAGE_NAME ]]; then
   exit 1
 fi
 
+# Download other test assets from GCS into MaxText/test_assets
+gcloud storage cp gs://maxtext-test-assets/* MaxText/test_assets
+
```
bvandermoon, line 49: Should this be behind a flag? I think this download was only part of `maxtext_runner.Dockerfile`. But this file is also used for other images that don't need the test assets

maxtext_runner.Dockerfile
shralex, line 12: I think this should be copied under assets/ rather than under deps

maxtext_runner.Dockerfile
```diff
@@ -8,10 +8,9 @@ FROM $BASEIMAGE
...
+COPY MaxText/test_assets/ MaxText/.
```
shralex, line 13: I think this should be copied under MaxText/test_assets rather than MaxText/

shralex (changes requested):
Thanks for the changes Luke! Please note that there is a second script from which maxtext_runner.Dockerfile is used (build_and_upload_images)

maxtext_runner.Dockerfile
lukebaumann: If `COPY . .` detects any change in the current directory, that layer of the image needs to be rebuilt. `assets` and `MaxText/test_assets` are directories with large files so it takes time to create the layer and then upload that layer.

If we separate the `COPY` command to isolate the copying of `assets` and `MaxText` into their own layers, those layers will not have cache misses, will not need to be remade, and not need to be re-uploaded.

This specifically speeds up *rebuild* times. The first build will not be sped up at all. This is still important for developers because the flow for the runner images is typically build, test, change, rebuild, test, change, ...

docker_upload_runner.sh
lukebaumann: This script is only for building and uploading the maxtext runner images.

My perspective is this PR is to fix a bug that we cannot copy from the bucket during `docker build ...`. I agree that this is not the best solution. This PR does not address the tech debt that we do not have two types of runners, "development runners" and "test runners".

A future PR could introduce different runner kinds behind a flag as you suggested and the GCS copy only would be needed for the "test runner" image.

maxtext_runner.Dockerfile
lukebaumann: Thanks! Fixed and confirmed that `maxtext/assets` is in `/deps/assets`.

lukebaumann:
> Thanks for the changes Luke! Please note that there is a second script from which maxtext_runner.Dockerfile is used (build_and_upload_images)

Found it from your PR #1729

lukebaumann:
@shralex for the failed TPU test, is this because the GCS bucket's test assets are outdated for `MaxText/test_assets/golden_data_grpo_default.jsonl` and `MaxText/test_assets/golden_data_llama4-17b-128e.jsonl`?

shralex (approved):

bvandermoon (approved):

shralex:
Not sure why tests aren't running, usually this is a sign of an issue in the test yml files.

lukebaumann:
I saw other PRs are not running as well so I do not suspect it is related to my PR

PR #1737, opened 2025-05-13
Add random routing for test purpose

RissyRan:
# Description

Add random routing for test purpose

# Tests

Unit test to pass

# Checklist

### DIFF
--- a/MaxText/layers/moe.py
+++ b/MaxText/layers/moe.py
@@ -59,6 +59,31 @@ DISPATCH = "dispatch"
 COMBINE = "combine"
 
 
+def random_routing(rng_key, gate_logits, num_experts_per_tok):
+  """
+  Performs random routing of tokens to experts.
+
+  Args:
+    rng_key: A JAX PRNGKey for randomness.
+    gate_logits: A JAX array of shape (batch_size, sequence_length, num_experts)
+                 representing the logits for each expert.
+    num_experts_per_tok: The number of experts to select for each token.
+
+  Returns:
+    A tuple containing:
+      - top_k_indices: JAX array of shape (batch_size, sequence_length, num_experts_per_tok)
+                       representing the indices of the selected experts for each token.
+      - top_k_weights: JAX array of shape (batch_size, sequence_length, num_experts_per_tok)
+                       representing the weights for the selected experts.
+  """
+  bs, seq_len, num_experts = gate_logits.shape
+  indices = jnp.arange(num_experts).repeat(bs * seq_len)
+  selected_num = bs * seq_len * num_experts_per_tok
+  top_k_indices = jax.random.choice(rng_key, indices, shape=(selected_num,)).reshape(bs, seq_len, num_experts_per_tok)
+  top_k_weights = jnp.take_along_axis(gate_logits, top_k_indices, axis=-1)
+  return top_k_weights, top_k_indices
+
+
 class GateLogit(nn.Module):
   """A layer used to compute gate logits, allowing to return the pre bias values for DeepSeek routing.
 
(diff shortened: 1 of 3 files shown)


gobbleturk (approved):

MaxText/layers/moe.py
```diff
@@ -59,6 +59,31 @@
 COMBINE = "combine"
 
 
+def random_routing(rng_key, gate_logits, num_experts_per_tok):
```
gobbleturk, line 62: I was imagining a different implementation where gate_logits was randomly generated, but I guess both implementations are roughly the same

MaxText/layers/moe.py
```diff
@@ -59,6 +59,31 @@
...
+    A tuple containing:
+      - top_k_indices: JAX array of shape (batch_size, sequence_length, num_experts_per_tok)
+                       representing the indices of the selected experts for each token.
+      - top_k_weights: JAX array of shape (batch_size, sequence_length, num_experts_per_tok)
+                       representing the weights for the selected experts.
+  """
+  bs, seq_len, num_experts = gate_logits.shape
+  indices = jnp.arange(num_experts).repeat(bs * seq_len)
+  selected_num = bs * seq_len * num_experts_per_tok
+  top_k_indices = jax.random.choice(rng_key, indices, shape=(selected_num,)).reshape(bs, seq_len, num_experts_per_tok)
```
gobbleturk, line 82: should we have sharding constraints for top_k_indices and top_k_weights? I guess we don't have sharding constraints in other conditional branches and don't see performance issues so I guess they are not needed

RissyRan: SG. We have sharding constraints for logits [here](https://github.com/AI-Hypercomputer/maxtext/blob/0f387905c8fbabbe556a1a502043bc0241ef3053/MaxText/layers/moe.py#L519 (as an example before calling top_k for gate_logits). I could re-visit if needed.

bvandermoon (approved):

MaxText/layers/moe.py
```diff
@@ -239,6 +264,13 @@ def generate_kernels(self, num_experts, emb_dim, mlp_dim):
 
   def get_topk(self, gate_logits, pre_bias_logits):
     """get topk. shape of top_k_weights & top_k_indices: (batch, sequence, num_experts_per_tok)"""
+    if self.config.use_random_routing:
+      # For debug and test purpose:
+      # Ideally we should split a new key for each layer, but for simplicity re-use the key to mimic the random routing
+      rng = jax.random.PRNGKey(1234)
```
gagika, line 270: To get a real random routing, you need to change (split) the key at each generation.

RissyRan: Yeah, I made a comment that `Ideally we should split a new key for each layer, but for simplicity re-use the key to mimic the random routing`. Otherwise, we need to pass it the from previous layer into next layer for split. I think this is for test and debug purpose, probably it's ok?

gagika: you likely won't get nice distribution with reusing the key. We had a few bugs for repeated decode that producing the same token when key was reused.

Looks like you are splitting the key in the unit test.

gagika: you can try something like this:

if self.config.use_random_routing:
      rng = self.make_rng('random_routing') 
      top_k_weights, top_k_indices = random_routing(rng, gate_logits, self.num_experts_per_tok)

RissyRan: Good catch for the test!

You mean have a helper function to generate a key? Let me have a try. But will it be related to `random_routing` always?

PR #1741, opened 2025-05-14
Add deepseek c4 convergence test and v5p recipe

suexu1025:
# Description

1. to support different c4 variance version in c4_mperf datatype
2. add deepseek v3 convergence configs.
3. add deepseek v5p recipe 

# Tests
validate #2 convergence run and #3 configs

# Checklist

### DIFF
--- a/benchmarks/maxtext_v5p_model_configs.py
+++ b/benchmarks/maxtext_v5p_model_configs.py
@@ -1,0 +1,107 @@
+"""
+ Copyright 2025 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+"""Shared Benchmark config for v6e orchestrations."""
+
+import os.path
+from benchmarks import xla_flags_library
+from benchmarks.maxtext_trillium_model_configs import MaxTextModel, _add_to_model_dictionary
+
+
+v5p_model_dict = {}
+
+deepseek_v3_ep_256_v5p_512  = _add_to_model_dictionary(
+    v5p_model_dict,
+    MaxTextModel(
+        model_name="deepseek_v3_ep_256_v5p_512",
+        model_type="deepseek3-671b",
+        tuning_params={
+            "per_device_batch_size": 4,
+            "max_target_length": 8192,
+            "ici_fsdp_parallelism": 1,
+            "ici_expert_parallelism": -1,
+            "remat_policy": "custom",
+            "decoder_layer_input": "offload",
+            "gcs_metrics": True,
+            "use_iota_embed": True,
+            "dataset_path": "gs://max-datasets-rogue",
+            "dataset_type": "synthetic",
+            "reuse_example_batch": 1,
+            "enable_checkpointing": False,
+            "skip_first_n_steps_for_profiler": 5,
+            "profiler_steps": 5,
+            "profiler": "xplane",
+            "sa_block_q": 2048,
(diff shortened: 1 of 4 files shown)


benchmarks/maxtext_v5p_model_configs.py
```diff
@@ -0,0 +1,107 @@
...
+            "megablox": False,
+            "sparse_matmul": False,
+            "capacity_factor": 1.0,
+            "tokenizer_type": "huggingface",
+            "tokenizer_path": "deepseek-ai/DeepSeek-V3",
+            "dtype": "bfloat16",
+            "opt_type": "adam_pax",
+            "attention": "flash",
+        },
+        xla_flags=(
```
RissyRan, line 58: In my recent dropping experiments, I noticed those flags may be helpful (1-2% MFU): https://screenshot.googleplex.com/3zAXSRMNQN6ufoL

1) HOST_OFFLOAD_FLAGS
2) LAYOUT_FOR_ALL_REDUCE_SCATTER
3) ENABLE_SPARSECORE_OFFLOADING_FOR_RS_AG_AR (does not help)
4) also worth to try if `remat_policy=custom decoder_layer_input=offload out_proj=offload query_proj=offload key_proj=offload value_proj=offload` can fit

RissyRan (commented):
Usually we need EP sharding in dropping use case to boost MFU. Curious what's the number looks like in FSDP + EP.

benchmarks/maxtext_v5p_model_configs.py
suexu1025: Thanks @RissyRan sharing all this info. Are you seeing perf improvement on v6e? 
We will retune all those flags for v5p. this is the baseline setup.

benchmarks/maxtext_v5p_model_configs.py
```diff
@@ -0,0 +1,107 @@
...
+from benchmarks import xla_flags_library
+from benchmarks.maxtext_trillium_model_configs import MaxTextModel, _add_to_model_dictionary
+
+
+v5p_model_dict = {}
+
+deepseek_v3_fsdp = _add_to_model_dictionary(
+    v5p_model_dict,
+    MaxTextModel(
+        model_name="deepseek_v3_fsdp",
```
bvandermoon, line 28: Please include `v5p` in the `model_name` for benchmarks in this file. The benchmark runner will have issues if there are duplicate model names, including across the v6e, v5e, and v5p files. Something similar is [being done for v5e](https://github.com/AI-Hypercomputer/maxtext/blob/main/benchmarks/maxtext_v5e_model_configs.py#L26), but unfortunately not for v6e

suexu1025: good idea, I added for v5p-suffix at model name.

benchmarks/maxtext_v5p_model_configs.py
RissyRan: Yes, I did observe the perf improvement so sharing it here :) Could you just run a test with FSDP + EP, it should give a better perf. 

ici_expert_parallelism=8 or 64, and ici_fsdp_parallelism=64 and 8 respectively. Other configs could keep the same. With dropping, EP sharding will boost the MFU.

suexu1025:
will waiting for PR approval until merge, but need to tag 'pull ready' to unblock internal testing

RissyRan (approved):
Thanks!

RissyRan:
Oh, one more thing, don't forget to squash commits into 1 commit for clean history. Thank you!

benchmarks/maxtext_v5p_model_configs.py
```diff
@@ -0,0 +1,107 @@
+"""
+ Copyright 2024 Google LLC
```
gagika, line 2: 2025

benchmarks/maxtext_trillium_model_configs.py
gobbleturk, line 1: Please try to avoid whitespace changes in unrelated files

suexu1025: good catch

benchmarks/maxtext_v5p_model_configs.py
suexu1025: updated

gagika (approved):

PR #1757, opened 2025-05-19
Update naming jax ai image

Rohan-Bierneni:
# Description

Migration from JAX STABLE STACK to JAX AI Images naming convention. 

- Updated the artifact registry repos to point to the new jax-ai-image repo instead of the old jax-stable-stack repo
- Changed files in readMe and tutorials to JAX AI Image
- Updated the maxtext_jax_ai_image Dockerfile with new base image argument
- Made sure consumers of the Dockerfile and jax ai image artifact registry also got updated

FIXES: b/418009488
FIXES: b/414612860

# Tests

Made sure changed links work and new artifact registry repos are publicly accessible with all the new images added. 

Built an image locally successfully with new MODE=jax_ai_image

# Checklist

### DIFF
--- a/.github/workflows/UploadDockerImages.yml
+++ b/.github/workflows/UploadDockerImages.yml
@@ -58,15 +58,15 @@ jobs:
     - name: build jax nightly image
       run : |
         bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_nightly MODE=nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_nightly
-    - name: build jax stable stack image
+    - name: build jax AI image
       run : |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
-    - name: build image with stable stack nightly jax
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=jax_ai_nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:latest
+    - name: build image with JAX AI nightly
       run: |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_stable_stack_nightly_jax MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_tpu_jax_stable_stack_nightly BASEIMAGE=us-docker.pkg.dev/tpu-prod-env-multipod/jax-stable-stack/tpu/jax_nightly:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_stable_stack_nightly_jax MODE=jax_ai_nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_tpu_jax_stable_stack_nightly BASEIMAGE=us-docker.pkg.dev/tpu-prod-env-multipod/jax-stable-stack/tpu/jax_nightly:latest
     - name: build image with jax stable stack release candidate image
       run: |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_stable_stack_candidate MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_stable_stack_candidate BASEIMAGE=us-docker.pkg.dev/tpu-prod-env-multipod/jax-stable-stack/candidate/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_stable_stack_candidate MODE=jax_ai_nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_stable_stack_candidate BASEIMAGE=us-docker.pkg.dev/tpu-prod-env-multipod/jax-stable-stack/candidate/tpu:latest
   gpu:
     strategy:
       fail-fast: false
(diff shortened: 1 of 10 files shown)


.github/workflows/UploadDockerImages.yml
```diff
@@ -58,9 +58,9 @@ jobs:
     - name: build jax nightly image
       run : |
         bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_nightly MODE=nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_nightly
-    - name: build jax stable stack image
+    - name: build jax AI image
       run : |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
```
parambole, line 63: `mode=jax_ai_image`

@Rohan-Bierneni - We will have to update this here and in other places in the repo. This might break the existing docs, so let's create a follow-up bug for this, i.e., making changes to the mode and all of the docs (Tutorials).

I would recommend making it part of this PR by adding logic to replace stable_stack with `jax_ai_image` so that both modes are supported while the docs are being refactored.

.github/workflows/UploadDockerImages.yml
```diff
@@ -84,9 +84,9 @@ jobs:
...
       run : |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_jax_stable_stack MODE=stable_stack DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_jax_stable_stack BASEIMAGE=us-central1-docker.pkg.dev/deeplearning-images/jax-stable-stack/gpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_jax_stable_stack MODE=stable_stack DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_jax_stable_stack BASEIMAGE=us-central1-docker.pkg.dev/deeplearning-images/jax-ai-image/gpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
     - name: build image with stable stack nightly jax
       run: |
         bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_gpu_stable_stack_nightly_jax MODE=stable_stack DEVICE=gpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_gpu_jax_stable_stack_nightly BASEIMAGE=us-docker.pkg.dev/tpu-prod-env-multipod/jax-stable-stack/gpu/jax_nightly:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
```
parambole, line 92: @Rohan-Bierneni - Can you also make changes to the nightly flow here and at the other places where it's being used?

parambole, line 92: @Rohan-Bierneni Additionally, it looks like we will have to make a change to the MaxText (stable stack) images that are being created, along with the corresponding XLML changes. Please make those changes as well.

.github/workflows/UploadDockerImages.yml
Rohan-Bierneni: Added the logic to support jax_ai_image as a MODE. However, still kept support for using MODE=stable_stack. The changes now that need to be done are adding new artifact registry repos and then updating the CLOUD_IMAGE_NAME argument to match the new repos here. Also need to update the requirements_with_stable_stack.txt file to requirements_with_ai_image.txt. Did not make this change now to first make sure all the consumers of this file is accounted for. Will be tracking these updates in bug b/418854585

.github/workflows/UploadDockerImages.yml
Rohan-Bierneni: Added new mode=jax_ai_image to the nightly and candidate flows

Rohan-Bierneni: Will be adding these changes in follow up pr. Tracking this issue in b/418854585

.github/workflows/UploadDockerImages.yml
```diff
@@ -58,15 +58,15 @@ jobs:
     - name: build jax nightly image
       run : |
         bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_nightly MODE=nightly DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_nightly
-    - name: build jax stable stack image
+    - name: build jax AI image
       run : |
-        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=stable_stack DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
-    - name: build image with stable stack nightly jax
+        bash .github/workflows/build_and_upload_images.sh CLOUD_IMAGE_NAME=maxtext_jax_stable_stack MODE=jax_ai_image DEVICE=tpu PROJECT=tpu-prod-env-multipod LOCAL_IMAGE_NAME=maxtext_jax_stable_stack BASEIMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:latest MAXTEXT_REQUIREMENTS_FILE=requirements_with_jax_stable_stack.txt
```
bvandermoon, line 63: Is the old image name still going to work also? In addition to places @parambole mentioned, some [tpu-recipes](https://github.com/AI-Hypercomputer/tpu-recipes/tree/main/microbenchmarks/trillium/collectives) would also be impacted by not finding this image. Will there be two copies of the image so both names can stilL be used for backwards compatibility?

docker_build_dependency_image.sh
```diff
@@ -66,30 +66,30 @@ if [[ -z ${DEVICE} ]]; then
...
 }
 
 if [[ -z ${LIBTPU_GCS_PATH+x} ]] ; then
   export LIBTPU_GCS_PATH=NONE
   echo "Default LIBTPU_GCS_PATH=${LIBTPU_GCS_PATH}"
   if [[ ${DEVICE} == "gpu" ]]; then
-    if [[ ${MODE} == "stable_stack" ]]; then
-      build_stable_stack
+    if [[ ${MODE} == "stable_stack" || ${MODE} == "jax_ai_image" ]]; then
+      build_ai_image
```
bvandermoon, line 92: Looks like the plan is to support both of these `MODEs` for backwards compatibility, right?

.github/workflows/UploadDockerImages.yml
Rohan-Bierneni: The old images are still supported in that we have the artifact registry still up. We created a new artifact registry repo for jax-ai-image and all further updates will be added to the jax-ai-image repo. However, we do want to deprecate the jax-stable-stack repo and potentially delete it entirely. Is it possible to update the tpu-recipes to point to the new repo in AR?

docker_build_dependency_image.sh
Rohan-Bierneni: Yes, this pr just added the logic to support MODE="jax_ai_image". In a future pr, if we want to deprecate the "stable_stack" mode, then need to assess all consumers of it and change to "jax_ai_image".

.github/workflows/UploadDockerImages.yml
bvandermoon: Good question. I am not sure if that would fully work or not. Newer tpu-recipes require MaxText to be checked out at a specific commit, so those ones would always use the `MODE=stable-stack` command. It could be possible to switch over the `BASE_IMAGE` that gets passed to the command though. But based on the way older tpu-recipes have degraded over time (newer recipes are hopefully in better shape), it could be a problem for re-testing/validating the recipes. Thoughts?

Rohan-Bierneni: Then we can keep the old stable-stack repo since they only contain a few images. This shouldn't cause any breakage. However, from now on that repo will not get updated so for any new images needed please use the jax-ai-image repo.

RissyRan (approved):

Rohan-Bierneni:
Merged these changes via copybara CL: cl/763888216

PR #1761, opened 2025-05-21
Add preprocessing functions for Llama4 multimodal (tiles)

hengtaoguo:
# Description

Image preprocessing utils for Llama4 multimodal. Core workflow in `pre_process_llama4_image()`. Input example: jnp.array of (536, 640, 3); Output example: (5, 3, 336, 336).
- Find best Resolution Canvas: `get_best_resolution` determines the best canvas resolution for the input image from a list of supported resolutions.
- Padding: `pad_to_best_fit_jax` pads the image to fit the determined best resolution canvas.
- Normalization: Normalizes the pixel values of the padded image.
- Tiling: `split_to_tiles_jax` Splits the processed image into a number of tiles.
- Global Tile Addition: If the image is split into multiple tiles, an additional global tile (a resized/normalized version of the original image) is added to the set of tiles.

# Tests

Passing unit tests, processed images match with Llama4 from HuggingFace transformers.

# Checklist

### DIFF
--- a/MaxText/multimodal_utils.py
+++ b/MaxText/multimodal_utils.py
@@ -16,7 +16,8 @@ limitations under the License.
 
 """Utils needed by multimodal pipelines for image processing."""
 
-from typing import List
+from typing import List, Tuple, Union
+from collections import defaultdict
 import os
 
 import numpy as np
@@ -28,6 +29,7 @@ import jax.numpy as jnp
 
 NUM_IMAGE_CHANNELS = 3
 
+# TODO(hengtaoguo): Move following constants to a separate file
 # Constants for Gemma3-specific processing
 GEMMA_DEFAULT_IMAGE_SIZE = 896
 GEMMA_IMAGE_MEAN = (127.5,) * 3
@@ -41,6 +43,13 @@ GEMMA_NUM_PLACEHOLDER_TOKENS_PER_IMAGE = 256
 # +4 means 4 extra tokens to pad around image: \n\n, <start_of_image>, <end_of_image>, \n\n
 # One MEDIA means one image or multiple images in one video, but now we only support one image
 GEMMA_NUM_TOKENS_PER_MEDIA = GEMMA_NUM_PLACEHOLDER_TOKENS_PER_IMAGE + 4
+
+# Constants for Llama4-specific processing
+LLAMA4_TILE_SIZE = 336
+LLAMA4_TILES_NUM = 16
+LLAMA4_PIXEL_VALUE_RESCALE_FACTOR = 1.0 / 255.0
+LLAMA4_IMAGE_MEAN = (0.5,) * 3
+LLAMA4_IMAGE_STD = (0.5,) * 3
 
 
 def load_image_from_path(image_path):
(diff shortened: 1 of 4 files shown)


MaxText/tests/pyconfig_test.py
hengtaoguo, line 1: This change was generated from `bash code_style.sh`

ZhaoyueCheng (approved):

aireenmei (approved):
LGTM, thanks Hengtao!

MaxText/multimodal_utils.py
```diff
@@ -42,6 +43,13 @@
 # One MEDIA means one image or multiple images in one video, but now we only support one image
 GEMMA_NUM_TOKENS_PER_MEDIA = GEMMA_NUM_PLACEHOLDER_TOKENS_PER_IMAGE + 4
 
+# Constants for Llama4-specific processing
```
gobbleturk, line 46: perhaps a single file containing all constants (this is fine for now with only 2 models)

MaxText/multimodal_utils.py
```diff
@@ -70,6 +78,190 @@ def _normalize_images(images, mean, std):
   return images
 
 
+def get_factors(dividend: int):
```
gobbleturk, line 81: I think we can just use block_size * (x // block_size + 1) (rounded up)

MaxText/tests/multimodal_utils_test.py
```diff
@@ -64,5 +65,56 @@ def test_add_images_in_batch(self):
...
+class TestLlama4ImageProcessing(unittest.TestCase):
+  """Test Llama4 image processing"""
+
+  def setUp(self):
+    super().setUp()
+    self.LLAMA4_TILES_NUM = 16
+    self.LLAMA4_TILE_SIZE = 336
+    self.NUM_IMAGE_CHANNELS = 3
+
+  def test_get_best_resolution(self):
```
gobbleturk, line 77: possibly this can be removed with refactor

num_tiles_x_direction = math.ceil(x_size/x_block_size)

MaxText/tests/pyconfig_test.py
```diff
@@ -120,7 +120,7 @@ def test_overriding_model(self):
         skip_jax_distributed_system=True,
         model_name="gemma-7b",
         override_model_config=True,
-        base_emb_dim=1024, # Defined as 3072 in gemma-7b
+        base_emb_dim=1024,  # Defined as 3072 in gemma-7b
```
gobbleturk, line 123: mb.

gobbleturk (approved):
Awesome!!!

MaxText/multimodal_utils.py
```diff
@@ -70,6 +78,190 @@ def _normalize_images(images, mean, std):
...
+  Returns:
+      set: A set containing all factors of the number.
+  """
+  factors_set = set()
+
+  for i in range(1, int(dividend**0.5) + 1):
+    if dividend % i == 0:
+      factors_set.add(i)
+      factors_set.add(dividend // i)
+  return factors_set
```
gobbleturk, line 96: possibly there is a math built in for this

gobbleturk (approved):

MaxText/multimodal_utils.py
```diff
@@ -70,6 +78,190 @@ def _normalize_images(images, mean, std):
...
+      image_width (int): The width of the image.
+      possible_resolutions (list): A list of possible resolutions.
+      resize_to_max_canvas (bool): Whether to resize to max canvas or not.
+  Returns:
+      tuple: The best resolution for the image.
+  """
+  if resize_to_max_canvas:
+    return max(possible_resolutions, key=lambda x: x[0] * x[1])
+  else:
+    return min(possible_resolutions, key=lambda x: abs(x[0] - img_height) + abs(x[1] - image_width))
```
gobbleturk, line 132: okay I started at this code a bit more - I don't think this is the same as ceil(x/block_x) - in some cases it will round down, e.g. if the image is 201 x 201 and the block size is 100 x 100 I think it will still choose to use 2x2 blocks? Some pixels are are not in these four blocks but at least they will appear in the global image

hengtaoguo:
> Awesome!!!

Thank you so much for the thorough review!

MaxText/multimodal_utils.py
hengtaoguo: Great suggestion! I will refactor these variables to a separate file for multimodal stuff.

MaxText/multimodal_utils.py
hengtaoguo: Thank Matt for the great catch! This is line is to find the resolution closest to the original image dimensions (minimizing padding/cropping), it could be slightly smaller than the actual image size so the ceiling might not work here. Yeah let's keep it the current way to temporarily unblock downstream tasks!

PR #1765, opened 2025-05-21
Enable all previously disabled integration tests

SurbhiJainUSC:
# Description
This PR enables all integration tests that were previpusly disabled in [PR].

# Tests

commands to reproduce.

# Checklist


gobbleturk:
Are we still using subprocess for some of these tests?

aireenmei:
My understand is when "bash setup_gcsfuse.sh" is called from a test script, it will use subprocess? For our github runners, the bucket maxtext-dataset is already mounted to tmp/gcsfuse as part of the GKE setup, we no loner need to run setup_gcsfuse.sh in test script. But I don't want to just remove them since developers may manually run the test in their devbox, can we add some conditional statement so bash setup_gcsfuse.sh doesn't run on github runners? cc @shralex

SurbhiJainUSC:
> My understand is when "bash setup_gcsfuse.sh" is called from a test script, it will use subprocess? For our github runners, the bucket maxtext-dataset is already mounted to tmp/gcsfuse as part of the GKE setup, we no loner need to run setup_gcsfuse.sh in test script. But I don't want to just remove them since developers may manually run the test in their devbox, can we add some conditional statement so bash setup_gcsfuse.sh doesn't run on github runners? cc @shralex

Yes, that's why I am trying to skip gcsfuse setup only for unit tests and testing it out on github runners.

SurbhiJainUSC:
> Are we still using subprocess for some of these tests?

Yes, these tests run a bash script and that can be done by subprocess afaik.

PR #1766, opened 2025-05-21
PP support for DeepSeek

gobbleturk:
# Description

Add support for SPMD pipelining deepseek with dropping.

This is achieved in this PR by building on this [PR](https://github.com/AI-Hypercomputer/maxtext/compare/main) which added support for SPMD pipelining only subset of layers. Here we pipeline only a subset of the sparse layers (since there are more sparse layers and those are the ones which need pipelining since the sparse DP comms are expensive). The layers that are not pipelined are executed via data parallelism.

Dropless is blocked on b/418313093

This is a modification of https://github.com/AI-Hypercomputer/maxtext/pull/1687

Pipeline parallelism offers performance benefits when data parallelism comms are too large - which occurs when the batch is too small and/or the model is very sparse. The experiments below on the smaller deepseek model show that the large data parallel comms are reduced when using PP with this PR.

# Tests

Ran some locally of smaller v2-16B model and added AOT test

* DeepSeekV2 16B local on v6e-8 with pdb=1, PP=8, pipeline_subset_layers=24 [xprof](https://xprof.corp.google.com/trace_viewer/mattdavidow-2021970820805495549?hosts=t1v-n-70a81276-w-0&host_index=0&trace_filter_config=%7B%7D&view_start=-2803.551&view_end=172.797)
* With pure FSDP step time is roughly 2x [xprof](https://xprof.corp.google.com/trace_viewer/mattdavidow-463037051654911153?hosts=t1v-n-70a81276-w-0&host_index=0&trace_filter_config=%7B%7D&view_start=-678.904&view_end=2265.952)
* Prior to this PR, PP could not be run at all with DeepSeek
# Checklist

### DIFF
--- a/MaxText/tests/train_compile_test.py
+++ b/MaxText/tests/train_compile_test.py
@@ -561,6 +561,29 @@ class TrainCompile(unittest.TestCase):
     )
 
   @pytest.mark.cpu_only
+  def test_moe_deepseek_pipeline_subset(self):
+    compiled_trainstep_file = "/tmp/test_moe_deepseek_pipeline_subset.pickle"
+    train_compile_main(
+        (
+            None,
+            os.path.join(PKG_DIR, "configs", "base.yml"),
+            f"compiled_trainstep_file={compiled_trainstep_file}",
+            "compile_topology=v6e-256",
+            "compile_topology_num_slices=8",
+            "use_iota_embed=true",
+            "model_name=deepseek3-671b",
+            "megablox=False", # dropless not yet supported (b/418313093)
+            "sparse_matmul=False", 
+            "capacity_factor=1",
+            "per_device_batch_size=1",
+            "max_target_length=2048",
+            "pipeline_parallel_layers=56",
+            "ici_expert_parallelism=16",
+            "dcn_pipeline_parallelism=8"
+        )
+    )
+
+  @pytest.mark.cpu_only
   def test_moe_llama4_17b_16e(self):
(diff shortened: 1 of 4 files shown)


RissyRan (approved):

MaxText/tests/train_compile_test.py
```diff
@@ -560,6 +560,29 @@ def test_moe_deepseek_without_device_limit(self):
...
+        (
+            None,
+            os.path.join(PKG_DIR, "configs", "base.yml"),
+            f"compiled_trainstep_file={compiled_trainstep_file}",
+            "compile_topology=v6e-256",
+            "compile_topology_num_slices=8",
+            "use_iota_embed=true",
+            "model_name=deepseek3-671b",
+            "megablox=False", # dropless not yet supported (b/418313093)
+            "sparse_matmul=False",
```
bvandermoon, line 576: Is my memory right that sparse matmul can't be used with dropping? Noticed this was changed from https://github.com/AI-Hypercomputer/maxtext/pull/1687 so just wanted to double check that

bvandermoon (approved):

MaxText/tests/train_compile_test.py
gobbleturk: ya our two sparse matmul paths (megablox and ragged dot) are used with dropless approaches

there is an issue specific with PP + EP + droppless b/418313093 - thus this PR only supports dropping (capacity_factor > 0, megablox=False, sparse_matmul=False)

PR #1767, opened 2025-05-22
tighten up the buffer size for ra2a during expert sharding.

gpolovets1:
# Description

Reduces the buffer size used for ragged_all_to_all when the number of experts per shard is less than number of experts per token. This helps if you have a large amount of expert sharding or large amount of experts per token.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/layers/moe.py
+++ b/MaxText/layers/moe.py
@@ -758,16 +758,21 @@ class RoutedMoE(nn.Module):
               expert_shard_id,
               num_expert_parallelism,
           )
-          # TODO(ranran): For better performance, we could update output buffer
-          # to a smaller size to replace self.get_expert_parallelism_size() for
-          # efficiency, or we could apply capacity_factor for excessive experts.
-          # Note: Reducing buffer increase the risk of token dropping under
-          # unbalanced distribution.
+
+          # TODO(ranran): For better performance, we could update output buffer to a smaller
+          # size to replace self.get_expert_parallelism_size() for efficiency,
+          # Or we could apply capacity_factor for excessive experts.
+          # Note: Reducing buffer increase the risk of token dropping under unbalanced distribution.
+
+          # In the worst case, all of the global input data is assigned to each expert in the current shard.
+          # This would result in num_expert_shards * input_size * experts_per_shard assignments. However, if
+          # experts_per_shard > num_experts_per_tok we cannot assign more than num_experts_per_tok to all of the inputs.
+          max_local_experts_per_tok = min(local_expert_size, self.config.num_experts_per_tok)
           buffer_size = int(
               num_expert_parallelism
               * self.config.per_device_batch_size
               * self.config.max_target_length
-              * self.config.num_experts_per_tok
+              * max_local_experts_per_tok
           )
           output_shape = jnp.zeros((buffer_size, self.config.emb_dim), dtype=x.dtype)
 


RissyRan (approved):
LGTM! Just add some explanation (verifying in review) for future reference.

Activation is permuted from size `(batch_per_shard * seq, emb)` to `batch_per_shard * seq * num_experts_per_tok, emb)`. In the worst case, tokens are in one shard. Then the buffer size is num_expert_shards * batch_per_shard * seq * num_experts_per_tok.

1 - In Mixtral example, we select 2 out of 8, and we use EP=4, so each shard has 2 experts, and batch_per_shard=4, seq=10:
* for each shard it's 4 x 10 x 2 = 80
* if we assume those shards are routed to expert 0 & 1 that on device 0
* then, the buffer size needs to be 80 x 4 = 320 to hold all tokens.

2 - If we use min(local_expert_size, self.config.num_experts_per_tok), it will get buffer size = 4 x 4 x 10 x 2 = 320.

In Mixtral example, we select 2 out of 8, and we use EP=8, so each shard has 1 experts, and batch_per_shard=4, seq=10:
* for each shard it's 4 x 10 x 2 = 80
* if we assume those shards are routed to expert 0 & 1 on 2 devices
* then, the buffer size needs to be 80 x 8 / 2 = 320 to hold all tokens.

If we use `min(local_expert_size, self.config.num_experts_per_tok)`, it will get buffer size = 8 x 4 x 10 x 1 = 320.

3 - In Mixtral example, we select 2 out of 8, and we use EP=2, so each shard has 4 experts, and batch_per_shard=4, seq=10:
* for each shard it's 4 x 10 x 2 = 80
* if we assume those shards are routed to expert 0 & 1 that on 1 device
* then, the buffer size needs to be 80 x 2 = 160 to hold all tokens.

Use the formula: buffer_size = 2 x 4 x 10 x 2 = 160.

gobbleturk (approved):
Curious do you actually have a config where this is helpful - I guess DeepSeek with EP>=64?

MaxText/layers/moe.py
```diff
@@ -630,15 +630,21 @@ def wrapper(x, logits, pre_bias_logits, w0, w1, wo):
...
+          
           # TODO(ranran): For better performance, we could update output buffer to a smaller
           # size to replace self.get_expert_parallelism_size() for efficiency,
           # Or we could apply capacity_factor for excessive experts.
           # Note: Reducing buffer increase the risk of token dropping under unbalanced distribution.
+
+          # In the worst case, all of the global input data is assigned to each expert in the current shard.
+          # This would result in num_expert_shards * input_size * experts_per_shard assignments. However, if
+          # experts_per_shard > num_experts_per_tok we cannot assign more than num_experts_per_tok to all of the inputs.
+          max_experts_per_input = min(local_expert_size, self.config.num_experts_per_tok)
```
gobbleturk, line 642: nit: perhaps a better variable name here - max_local_experts_per_tok

gobbleturk:
> In the worst case, tokens are in one shard

I agree with this except for the wording "In the worst case, tokens are in one shard". This isn't true when # local experts < # exp_per_tok, which is what the this PR is taking advantage of. Its illustrated nicely by your 2nd example where you had to divide by 2 since there is only 1 expert per shard but 2 experts per token. Perhaps you meant "In the worst case the tokens are all routed to only exp_per_tok experts" or "In the worst case, tokens are in one shard" is true when # local experts >= # exp_per_tok?

PR #1802, opened 2025-06-04
Faysal/add thd support

kocchop:
# Description

This PR adds packing support to `cudnn_flash_te` api for GPU flash attention. This would require Transformer-Engine >=2.1

# Tests

Added an integration test to the `train_tests.py` 

# Checklist

### DIFF
--- a/MaxText/layers/attentions.py
+++ b/MaxText/layers/attentions.py
@@ -842,6 +842,8 @@ class AttentionOp(nn.Module):
       sliding_window_size = [self.sliding_window_size, 0]
 
     if self.config.packing and self.config.dataset_type != "synthetic":
+      if decoder_segment_ids is None:
+        decoder_segment_ids = jnp.ones(shape=query.shape[:2], dtype=jnp.int32)
       attn_mask = SequenceDescriptor.from_segment_ids_and_pos(segment_ids=decoder_segment_ids, segment_pos=None)
       qkv_layout = "THD_THD_THD"  # 'T3HD', 'THD_T2HD' or 'THD_THD_THD'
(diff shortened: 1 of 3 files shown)


wenchenvincent:
Hi @kocchop what's the status of this PR?

MaxText/layers/attentions.py
```diff
@@ -820,31 +820,37 @@ def cudnn_flash_attention(
...
       sliding_window_size = [self.sliding_window_size, 0]
 
-    if self.attention_type == AttentionType.LOCAL_SLIDING or using_context_parallelism:
-      mask_type = "causal"  # SWA and Context Parallelism only work with causal masking
+    if self.config.packing and self.config.dataset_type != "synthetic":
+      if decoder_segment_ids is None:
+        decoder_segment_ids = jnp.ones(shape=query.shape[:2], dtype=jnp.int32)
+      attn_mask = SequenceDescriptor.from_segment_ids_and_pos(segment_ids=decoder_segment_ids, segment_pos=None)
+      qkv_layout = "THD_THD_THD"  # 'T3HD', 'THD_T2HD' or 'THD_THD_THD'
+      max_segments_per_seq = 32
```
wenchenvincent, line 849: `max_segments_per_seq` needs to be consistent with the max number of segments per example resulting from the sequence packing algorithm in the data input pipeline. If `max_segments_per_seq`  is not large enough, there will be an issue. 

Instead of hardcoding this parameter, we shall add it as a configurable parameter. We would also need to change the data input pipeline to be able to take such a parameter and honor it in sequence packing.

PR #1810, opened 2025-06-06
sharding validator Improvements

wei879-100:
# Description
This change significantly enhances the sharding validation utility by providing detailed, actionable error messages. The previous version only indicates that some matrices were not fully sharded, but it failed to specify which matrices were unsharded by which axis. This forced developers into a tedious manual search to find the unsharded matrices. This improvement is able to show a more readable error message with large matrices that has not been fully sharded by which axes.

Specifically, for each replicated tensor, the error now shows:

**Which Matrix:** The full name of the parameter (e.g., layers.10.attention.query.kernel).
**Its Current (Problematic) Sharding Status:** The parameter's current PartitionSpec is displayed (e.g., PartitionSpec(None, 'expert')), making it clear why it's considered replicated on the main data-parallel axes.
**Which Axes are not sharded:** A list of available mesh axes that could have been used for sharding (e.g., could be sharded on: ['fsdp', 'tensor']), giving the developer an immediate hint for a valid solution.

FIXES: b/367055330

# Tests
This change was validated using a comprehensive unit test suite, <em>TestAssertParamsSufficientlySharded</em>, which ensures the new validation logic is both correct and robust. The tests cover a wide variety of sharding configurations to confirm that the assertions pass and fail as expected.

**Test Coverage**
The test suite validates the following key scenarios:
**Correctly Passing Scenarios:**
Fully Sharded: A tensor that is fully sharded across all available primary mesh axes passes the assertion.
Sufficiently Sharded: A tensor sharded on at least one valid target axis (e.g., fsdp) passes, even if its other dimensions are replicated.
Complex Mesh: A tensor correctly sharded on a valid axis within a complex, multi-dimensional mesh also passes.
**Correctly Failing Scenarios:**
Completely Unsharded: A fully replicated parameter with no sharding specification correctly triggers an AssertionError.
Incorrectly Sharded: A tensor that is sharded, but not along any of the required target axes (e.g., sharded on 'sequence' but not on 'fsdp'), correctly fails.
Mixed Sharding: A model containing both correctly sharded and fully replicated parameters fails when the total size of replicated parameters exceeds the specified tolerance.

Each failing test case implicitly verifies that the raised AssertionError contains the new, detailed diagnostic message, including the names of problematic tensors and the sharding axes they could be mapped to. The tests also confirm that the companion utility, get_formatted_sharding_annotations, runs without error in all configurations.

# Checklist

### DIFF
--- a/MaxText/maxtext_utils.py
+++ b/MaxText/maxtext_utils.py
@@ -27,11 +27,14 @@ from flax.training import train_state
 
 import numpy as np
 
+from collections.abc import Iterable
 from jax.experimental import mesh_utils
 from jax.experimental.serialize_executable import deserialize_and_load
 from jax.sharding import PartitionSpec as P
+
 import jax
 import jax.numpy as jnp
+import jax.tree_util as jtu
 
 import optax
 
@@ -343,25 +346,51 @@ def calculate_prefill_tflops_per_device(num_model_parameters, prefill_length, co
   return total_tflops, learnable_weight_tflops, causal_attention_tflops
 
 
-def assert_params_sufficiently_sharded(params, mesh, tolerance):
-  """Checks whether most params are sharded across sharding axis.
+def get_mesh_axes_used_by_tensor_spec(tensor_sharding_spec):
+  """
+  Extracts the set of mesh axis names that a tensor's PartitionSpec uses.
 
-  This function determines whether the majority of parameters  are distributed
-  across a specified sharding axes with an acceptable tolerance. It compares the
-  current distribution to a scenario where all parameters are fully sharded
-  across the 'fsdp', 'fsdp_transpose', 'sequence', and 'tensor' axes.
+  This function inspects a tensor's sharding specification (PartitionSpec) and
+  identifies which mesh axes are actively used for sharding. If a tensor is not
+  sharded (i.e., fully replicated), the resulting set will be empty.
 
   Args:
-    params: params of the model state
-    mesh: mesh constructed from config
-    tolerance: float between 0.0 and 1.0 representing the allowed percentage of
-    non-sharded parameters.
+    tensor_sharding_spec: The PartitionSpec of a tensor, which defines how it's partitioned across the mesh.
+    It can be None or contain strings and iterables representing the mesh axes.
+    all_mesh_axis_names: A collection of all available mesh axis names in the current device mesh.
+
   Returns:
-    bool: True if the majority of parameters are sufficiently sharded
+    A set of strings, where each string is a mesh axis name used by the
+    tensor's sharding spec. Returns an empty set for unsharded tensors.
   """
-  total_num_params = max_utils.calculate_num_params_from_pytree(params)
-  product_num_devices_for_weight_sharding = 1
-  for axis in [
+  # Flatten the sharding spec, as it can contain nested iterables (e.g., ('data', 'mdl')).
+  tensor_sharding_spec = sum(
+      [
+          [axis] if isinstance(axis, str) else list(axis) if isinstance(axis, Iterable) else []
+          for axis in tensor_sharding_spec
+      ],
+      [],
+  )
+  return tensor_sharding_spec
+
+
+def _get_nontrival_mesh_axes(mesh):
+  """
+  Returns mesh axes from config that are valid and have more than one shard.
+
+  This function identifies which of the predefined potential sharding axes are
+  actually present in the current device mesh and are configured with a size
+  greater than one (i.e., are actually sharded).
+
+  Args:
+    mesh: The device mesh object, which contains information about the mesh topology, including axis names and their sizes.
+
+  Returns:
+    A set of strings, where each string is a mesh axis name that is both
+    pre-configured as a target for sharding and has more than one shard in the mesh.
+  """
+
+  target_sharding_axes_config = [
       "fsdp",
       "fsdp_transpose",
       "sequence",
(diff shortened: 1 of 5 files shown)


MaxText/maxtext_utils.py
```diff
@@ -343,25 +346,56 @@ def calculate_prefill_tflops_per_device(num_model_parameters, prefill_length, co
   return total_tflops, learnable_weight_tflops, causal_attention_tflops
 
 
-def assert_params_sufficiently_sharded(params, mesh, tolerance):
-  """Checks whether most params are sharded across sharding axis.
+def get_mesh_axes_used_by_tensor_spec(tensor_sharding_spec, all_mesh_axis_names):
```
gobbleturk: I don't think we need to filter by all_mesh_axis_names but its also fine to filter I suppose

MaxText/maxtext_utils.py
```diff
@@ -372,19 +406,148 @@ def assert_params_sufficiently_sharded(params, mesh, tolerance):
...
+    # If not, raise an error, as all parameters are expected to have it.
+    assert has_sharding_spec, (
+        f"Parameter '{param_name_str}' is missing sharding information."
+        "All parameters must have an explicit '.sharding.spec' attribute."
+    )
+
+    current_sharding_spec = p_leaf.sharding.spec  # Extract the current tensor's sharding spec
+    # Determine which mesh axes are used by this parameter's sharding spec.
+    mesh_axes_used = get_mesh_axes_used_by_tensor_spec(current_sharding_spec, mesh.axis_names)
+    # Check if the parameter is sharded on at least one of the valid target axes.
```
gobbleturk: We actually want to assert something stronger - that its sharded on ALL of the target axes

MaxText/tests/maxtext_utils_test.py
```diff
@@ -265,5 +271,157 @@ def multiple_rules(self):
...
+      # Define a sharding spec that shards the first tensor dimension by the 'fsdp' mesh axis
+      # and the second dimension by the 'tensor' mesh axis.
+      pspec = PartitionSpec("fsdp", "tensor")
+      # Create a parameter and apply the sharding, ensuring it's distributed across all devices.
+      params = {"layer1": jax.device_put(jnp.ones((8, 8)), NamedSharding(self.mesh, pspec))}
+
+      # Assert that the parameters are sufficiently sharded; this should pass with no error.
+      assert_params_sufficiently_sharded(params, self.mesh, tolerance=0.1)
+      # Verify that the annotation function runs without error.
+      self.assertIsNotNone(get_formatted_sharding_annotations(params, self.mesh))
```
gobbleturk: I would not include get_formatted_sharding_annotations in this test - you can have just a single test for it if you want

A general principal is we prefer "narrow" tests - they test only one thing. E.g. you should only test either assert_params_sufficiently_sharded or get_formatted_sharding_annotations, but not both.

MaxText/tests/maxtext_utils_test.py
```diff
@@ -265,5 +271,157 @@ def multiple_rules(self):
...
+  def test_3d_tensor_sharded_on_fsdp_axis(self):
+    """
+    Tests that a 3D tensor sharded only on a valid target axis ('fsdp') passes.
+    """
+    with self.mesh:
+      pspec = PartitionSpec("fsdp", None, None)
+      params = {"conv3d_layer": jax.device_put(jnp.ones((8, 4, 4)), NamedSharding(self.mesh, pspec))}
+
+      # This should pass
+      assert_params_sufficiently_sharded(params, self.mesh, tolerance=0.2)
```
gobbleturk: This is a great test case, we actually probably should fail here though - the tensor is only sharded on one axis instead of both - it is not fully sharded. I'm surprised this passes with tolerance=0.2

MaxText/tests/maxtext_utils_test.py
```diff
@@ -265,5 +271,157 @@ def multiple_rules(self):
...
+      self.assertIsNotNone(get_formatted_sharding_annotations(params, self.mesh))
+
+  def test_4d_tensor_fully_sharded(self):
```
gobbleturk: these seem repetitive from the previous test cases, I would remove the 4d ones

MaxText/maxtext_utils.py
wei879-100: Sure, I removed

MaxText/maxtext_utils.py
wei879-100: Yes, I changed this part

MaxText/tests/maxtext_utils_test.py
wei879-100: Great suggestion! I separated them

MaxText/tests/maxtext_utils_test.py
wei879-100: Changed this part, and also change the code logic, now the unit test also passed

MaxText/tests/maxtext_utils_test.py
wei879-100: I just removed all the 4d ones

MaxText/maxtext_utils.py
```diff
@@ -343,25 +346,51 @@ def calculate_prefill_tflops_per_device(num_model_parameters, prefill_length, co
...
+      [
+          [axis] if isinstance(axis, str) else list(axis) if isinstance(axis, Iterable) else []
+          for axis in tensor_sharding_spec
+      ],
+      [],
+  )
+  return tensor_sharding_spec
+
+
+def _get_valid_target_axis(mesh):
```
gobbleturk: nit: I would rename to _get_nontrival_mesh_axes

MaxText/maxtext_utils.py
```diff
@@ -372,19 +401,152 @@ def assert_params_sufficiently_sharded(params, mesh, tolerance):
...
+
+    current_sharding_spec = p_leaf.sharding.spec  # Extract the current tensor's sharding spec
+    # Identify axes used for sharding
+    mesh_axes_used = get_mesh_axes_used_by_tensor_spec(current_sharding_spec)
+    # Check if the parameter is sharded on all the valid target axes.
+    is_sharded_on_all_target_axis = all(axis in mesh_axes_used for axis in valid_target_mesh_axes)
+
+    # If the parameter is not sharded on all of the target axes, it's considered "problematic."
+    if not is_sharded_on_all_target_axis:
+      unsharded_params_total_size += p_leaf.size  # Add to total unsharded parameter size
```
gobbleturk, line 456: Note this is different logic than what existed originally -- 

Before we would sum up the local shard sizes and compare to the ideal fully sharded size.

Here we are summing up all non-fully params and comparing that to the number of params

MaxText/maxtext_utils.py
wei879-100: Sure, I just changed it

MaxText/maxtext_utils.py
```diff
@@ -372,19 +401,152 @@ def assert_params_sufficiently_sharded(params, mesh, tolerance):
...
+    mesh: The device mesh object.
+    tolerance: A float representing the maximum allowed percentage of unsharded parameters.
+  """
+  # Calculate the total size of all parameters in the model.
+  total_num_params = max_utils.calculate_bytes_from_pytree(params)
+
+  # Get the set of valid mesh axes that can be used for sharding.
+  valid_target_mesh_axes = _get_valid_target_axis(mesh)
+  # If there are no valid axes to shard along, there's nothing to check, so we can exit.
+  if not valid_target_mesh_axes:
```
RissyRan, line 543: Does this mean it's not sharded? Should we assert error?

MaxText/maxtext_utils.py
```diff
@@ -372,19 +401,152 @@ def assert_params_sufficiently_sharded(params, mesh, tolerance):
...
+
+    # If the parameter is not sharded on all of the target axes, it's considered "problematic."
+    if not is_sharded_on_all_target_axis:
+      unsharded_params_total_size += p_leaf.size  # Add to total unsharded parameter size
+      unsharded_axes = set(valid_target_mesh_axes) - set(mesh_axes_used)
+      # Add detailed info to list of problematic tensors
+      problematic_tensors_details.append(
+          {
+              "name": param_name_str,  # Tensor name
+              "size": p_leaf.size,  # tensor size
```
RissyRan, line 462: Will it be more helpful to have tensor shape instead of total size?

RissyRan: Also could you attach one example of running logs and see how this looks like? Thanks!

MaxText/maxtext_utils.py
```diff
@@ -372,19 +401,152 @@ def assert_params_sufficiently_sharded(params, mesh, tolerance):
...
+    total_size: The total size of all parameters in the model.
+    tolerance: A float (e.g., 0.05 for 5%) representing the maximum allowed percentage of unsharded parameters.
+    problematic_tensors_details: A list of details about the unsharded tensors,
+    used to generate an informative error message.
+
+  Raises:
+    AssertionError: If the percentage of unsharded parameters is greater than the tolerance.
+  """
+  # Calculate the percentage of unsharded parameters.
+  unsharded_param_perc = unsharded_size / total_size if total_size > 0 else 0.0
```
RissyRan: Should we raise an error if total_size is <=0? instead of setting unsharded_param_perc=0?

RissyRan (commented):
LGTM! This is a great feature to have! 

Also, could you attach an example of logging with assertion failure of insufficient sharding in the PR?

MaxText/maxtext_utils.py
wei879-100: No, this does not mean it is not sharded. The basic logic for this is identifying that if there are no axes to shard along. If there is nothing to analyze, and it gracefully exits.

MaxText/maxtext_utils.py
wei879-100: Thanks for the suggestion!
The current code treats a model with no parameters as a trivial case where there are no sharding issues to report. It avoids a "division by zero" error and correctly determines the percentage of unsharded parameters is 0.

The total_size is calculated from max_utils.calculate_bytes_from_pytree(params). This variable represents the total size of all parameters in the model. If total_size is 0, it means the model has no parameters. While this might seem unusual for a typical large model, it is a perfectly valid scenario in several common situations such as unit tests. We might have unit tests that check the behavior of the training loop or other components with an empty model to isolate the logic being tested. In some possible valid cases, a model with no parameters has, by definition, zero unsharded parameters.

If we try to raise an error when total_size is 0, it might cause the program to crash in some legitimate scenarios (e.g., unit tests). This would make the assertion utility less robust. Since this assertion is aiming to check for inefficient sharding, not sure whether it is necessary to check the empty model parameter.

MaxText/maxtext_utils.py
wei879-100: For now, our model sharding is good enough and there is no assertion error.  Since the default configuration is set up correctly, we need to intentionally misconfigure the model's sharding to trigger the assertion.

Here is what the output would look like if a user misconfigured their sharding. The run would halt and print this detailed, actionable error message right after the model parameter count is calculated:
Name: 'decoder/embed_tokens/embedding', Size: 262144000, Spec: P(('fsdp', 'tensor'),)) is unsharded on axis: ('fsdp', 'tensor') could be sharded on: ('fsdp', 'tensor')
 - Name: 'decoder/layers_0/mlp/wi_0/kernel', Size: 234881024, Spec: P(('fsdp', 'tensor'),)) is unsharded on axis: ('tensor',) could be sharded on: ('fsdp', 'tensor')

The output log has been attached for reference
[Output Log.txt](https://github.com/user-attachments/files/20702303/Output.Log.txt)

wei879-100: > Will it be more helpful to have tensor shape instead of total size?

Yes, it's a great idea to have the tensor shape print out with the total size. I just changed this. Thanks a lot for the suggestion!

MaxText/maxtext_utils.py
RissyRan: Could you explain a little bit more? `assert_params_sufficiently_sharded` function is used for overall checking right? `valid_target_mesh_axes = _get_valid_target_axis(mesh)` is the 1st line inside of this funciton. Wondering what does this mean if `valid_target_mesh_axes` is empty/None?

MaxText/maxtext_utils.py
RissyRan: Thanks for explanation! Yeah, just curious about logging when it fails. I think shape may be more helpful here.

MaxText/maxtext_utils.py
RissyRan: > While this might seem unusual for a typical large model

Yeah, just a sanity check. Instead of silently pass `unsharded_param_perc = unsharded_size / total_size if total_size > 0 else 0.0`.

We could do:

```
if total_size <= 0:
  raise value error....

unsharded_param_perc = unsharded_size / total_size
```

> If we try to raise an error when total_size is 0, it might cause the program to crash in some legitimate scenarios (e.g., unit tests).

Do we run a model with total_size=0 in unit test?

MaxText/maxtext_utils.py
wei879-100: `assert_params_sufficiently_sharded` is for overall checking of parameter sharding. The line `valid_target_mesh_axes = _get_valid_target_axis(mesh)` identifies the actual sharding dimensions available in your JAX mesh (e.g., "data", "fsdp", "tensor"). If `valid_target_mesh_axes` is empty, it means there are no user-defined axes in your mesh that the system considers "valid targets" for sharding parameters. 

In this specific `if not valid_target_mesh_axes` block, the code is indeed designed to gracefully exit without asserting an error. This is because if there are no sharding dimensions to distribute parameters across, there's nothing to analyze or enforce sharding on. It doesn't mean the parameters aren't sharded (they might be fully replicated if the mesh has only one device, for instance), but rather that the `assert_params_sufficiently_sharded` function cannot perform its intended check for sufficient sharding because the concept of target sharding axes is absent or undefined in the current mesh configuration.

MaxText/maxtext_utils.py
wei879-100: No, actually we do not run a model with total_size = 0 in the unit test. 

Good suggestion! I will add this error assertion part instead of silently passing.

MaxText/maxtext_utils.py
gobbleturk: When we remove "tensor" sharding from vocab [here](https://github.com/AI-Hypercomputer/maxtext/blob/3296117b471c4297a8eebf86af0c3905b63df454/MaxText/configs/base.yml#L314) and run with pure TP, we will see an error message like https://paste.googleplex.com/5298003403079680

MaxText/maxtext_utils.py
```diff
@@ -848,3 +1014,67 @@ def schedule(step):
...
+      # Case 2: The parameter is explicitly marked as fully replicated.
+      elif hasattr(p_leaf.sharding, "spec") and p_leaf.sharding.spec is None:
+        sharding_desc = "Fully Replicated (spec is None)"
+      # Case 3: A generic fallback if a sharding object exists but has no recognized spec attribute.
+      else:
+        # Print the string representation of the sharding object itself.
+        sharding_desc = str(p_leaf.sharding)
+    # Case 4: The parameter has no .sharding attribute at all.
+    else:
+      sharding_desc = "No .sharding attribute found"
```
RissyRan, line 1075: nit: extra `.` here?

wei879-100: This is not extra `.`, this comes with `.sharding`.

PR #1813, opened 2025-06-10
Add total memory usage logging in GB and unit tests for activation offload

zhenying-liu:
Add total memory usage logging in GB which is more readable and useful to be tracked especially for offloading.
Also add two unit tests for activation offloading with and without scan.

### DIFF
--- a/MaxText/max_utils.py
+++ b/MaxText/max_utils.py
@@ -18,7 +18,6 @@ limitations under the License.
 All the functions include MaxText modules, such as Pyconfig, should be moved to MaxText utils file."""
 
 import functools
-import itertools
 import time
 import os
 import socket
@@ -39,14 +38,12 @@ import flax
 import psutil
 
 from etils import epath
-from flax.linen import partitioning as nn_partitioning
 
 import orbax.checkpoint as ocp
 
 from tensorboardX import writer
 
 from MaxText import max_logging
-from MaxText.train import load_next_batch
 
 
 HYBRID_RING_64X4 = "hybrid_ring_64x4"
@@ -741,15 +738,6 @@ def print_compiled_memory_stats(compiled_stats):
         f"Argument size: {compiled_stats.argument_size_in_bytes/(1024**3):.1f} GB, "
         f"Host temp size: {compiled_stats.host_temp_size_in_bytes/(1024**3):.1f} GB."
     )
-
-def compile_sample_batch(data_iterator, p_train_step, state, config, mesh, input_data_shardings):
-  copied_iter, _ = itertools.tee(data_iterator)
-  sample_batch = load_next_batch(copied_iter, None, config)
-  sample_batch = jax.lax.with_sharding_constraint(sample_batch, input_data_shardings)
-  with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
-    compiled = p_train_step.lower(state, sample_batch, jax.random.PRNGKey(0)).compile()
-    compiled_stats = compiled.memory_analysis()
-    print_compiled_memory_stats(compiled_stats)
 
 def print_system_information():
   """Print system information of the current environment.
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -837,7 +837,11 @@ def train_loop(config, recorder, state=None):
 
   metric_logger = MetricLogger(writer, config)
   input_data_shardings = maxtext_utils.get_input_data_sharding(config, mesh)
-  max_utils.compile_sample_batch(data_iterator, p_train_step, state, config, mesh, input_data_shardings)
+  with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
+    shaped_batch = maxtext_utils.get_shaped_batch(config)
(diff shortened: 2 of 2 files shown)


MaxText/train.py
```diff
@@ -963,11 +963,16 @@ def train_loop(config, recorder, state=None):
       compiled = p_train_step.lower(state, example_batch, nextrng).compile()
       compiled_stats = compiled.memory_analysis()
       if compiled_stats is not None:
+        total = (compiled_stats.output_size_in_bytes +
```
gobbleturk: I realize the existing code doesn't follow this pattern, but can we move this formula + print statement into max_utils? Other files (e.g. sft_trainer or grpo_trainer) may also want to call this function, and it cleans up train.py which we prefer to be lean

MaxText/tests/integration_tests/train_tests.py
```diff
@@ -276,6 +276,47 @@ def test_gpu_tensor_parallelism(self):
...
+        "dataset_type=synthetic",
+        "enable_checkpointing=False",
+        "enable_goodput_recording=False",
+        rf"tokenizer_path={os.path.join(os.path.dirname(PKG_DIR), 'assets', 'tokenizer.llama2')}",
+    ]
+    train_main(optimizer_offload)
+
+  @pytest.mark.integration_test
+  @pytest.mark.gpu_only
+  def test_gpu_activation_offload_without_scan(self):
```
gobbleturk, line 301: Curious why have scan=False, is this something we really care about?

MaxText/train.py
zhenying-liu: Yes, moved this formula + print statement into max_utils. Done.

MaxText/tests/integration_tests/train_tests.py
zhenying-liu: We also want to keep the explicit loop (non-scan version) working with good performance, as it may be needed for some non-LLM models. Also because they use different HLOs (copy-start/done vs dynamic-update-start/done, dynamic-slice-start/done), we want to track both.

MaxText/train.py
zhenying-liu: Also as requested in G Chat: "printing the stats at the end of the first step instead of at the end of all of the steps". This is done. It is moved to before the first step.

The current logs are in this order:
```
Memstats: After params initialized:
        Using (GB) XX / YY (ZZ%) on cuda:0
Total memory size: AA GB, Output size: BB GB, Temp size: CC GB, Argument size: DD GB, Host temp size: EE GB.
completed step: 0, seconds: xxx, TFLOP/s/device: yyy, Tokens/s/device: zzz, total_weights: www, loss: ooo
To see full metrics 'tensorboard --logdir=/tmp/tmp.Exyqpj9oUF/logdir/tensorboard/'
completed step: 1, seconds: xxx, TFLOP/s/device: yyy, Tokens/s/device: zzz, total_weights: www, loss: ooo
```

gobbleturk (approved):

MaxText/train.py
```diff
@@ -841,6 +841,10 @@ def train_loop(config, recorder, state=None):
     if step == first_profiling_step or prof.should_activate_periodic_profile(step):
       optional_postfix = f"step_{step}" if config.profile_periodically_period > 0 else ""
       prof.activate(blocking_object=state, optional_postfix=optional_postfix)
+      with mesh, nn_partitioning.axis_rules(config.logical_axis_rules):
```
gobbleturk: This fails our internal linter since nextrng may not be defined yet. Is it possible to move this before the main train for loop (before`for step in np.arange(start_step, config.steps):`)? You may have to plug in a random rngkey

zhenying-liu: Is there any quick way to fake example_batch with the same structure and types outside the loop? It is loaded and sharded inside the loop:
```
# Reshard data from loaded sharding to performant activation sharding
example_batch = jax.lax.with_sharding_constraint(example_batch, input_data_shardings)
```

zhenying-liu: Yes, moved this before the main train for loop and defined in max_utils.py to make train.py lean.
The data_iterator is copied to get the "example_batch" without disrupting the training loop. Please take a look.

MaxText/max_utils.py
```diff
@@ -38,12 +39,14 @@
...
 
 from etils import epath
+from flax.linen import partitioning as nn_partitioning
 
 import orbax.checkpoint as ocp
 
 from tensorboardX import writer
 
 from MaxText import max_logging
+from MaxText.train import load_next_batch
```
gobbleturk: We don't want max_utils to depend on train, can we instead get a shaped version of batch similar to how its done in train_compile here
https://github.com/AI-Hypercomputer/maxtext/blob/3296117b471c4297a8eebf86af0c3905b63df454/MaxText/train_compile.py#L102

zhenying-liu: It makes sense. Took this approach of maxtext_utils.get_shaped_batch(config). Done.

SurbhiJainUSC:
These changes are included in https://github.com/AI-Hypercomputer/maxtext/pull/1904 and it has been merged. Closing this PR.

PR #1809, opened 2025-06-05
Add Llama4VisionModel for multimodal decoding

hengtaoguo:
# Description

This PR implements a complete class `Llama4VisionModel`, by integrating all Llama4 basic vision components. It allows Llama4 multimodal decode by describing an input image. Joint by @hengtaoguo and @aireenmei .

- **Core change `Llama4VisionModel`**, which converts image tiles (batch_size, num_tiles, C, H, W) to feature activations (batch_size, num_tiles, num_patches, vision_output_dim_for_vit). Then `Llama4MultiModalProjector` projects it to (batch_size, num_tiles, num_patches, base_emb_dim). Example: (8, 5, 3, 336, 336) -> (8, 5, 144, 4096) -> (8, 5, 144, 5120). After that, the image tokens will be merged into text tokens by re-using `merge_mm_embeddings`. 
- Refactor by adding a `get_dummy_image_shape_for_init()` to create desired dummy images for different models, for jit init purpose.

# Tests

Tested full multimodal decode on v5p-16 cluster with this [command](https://paste.googleplex.com/6229393653104640), and [workload](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22cloud-tpu-multipod-dev%22%0Aresource.labels.location%3D%22europe-west4%22%0Aresource.labels.cluster_name%3D%22mlperf-v5p-16%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name:%22htg-llama4-17b-16e-decode-mm-full-ep8-slice-job-0-0-%22%20severity%3E%3DDEFAULT%0Atimestamp%20%3E%3D%20%222025-06-19T01:49:47.511191474Z%22%20AND%20timestamp%20%3C%3D%222025-06-19T01:56:47.705371521Z%22;storageScope=project;cursorTimestamp=2025-06-19T01:49:52.163035619Z;duration=P1D?e=13802955&mods=allow_workbench_image_override&project=cloud-tpu-multipod-dev&inv=1&invt=Ab0mDQ) with [screenshot](https://screenshot.googleplex.com/6cqjHnUupQQniXY).
image

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #1816, opened 2025-06-10
llama4 ckpt conversion

aireenmei:
# Description

Joint by @aireenmei and @hengtaoguo .
Add ckpt conversion for llama4 vision model from hf to maxtext. 

# Tests

Verified in https://github.com/AI-Hypercomputer/maxtext/pull/1809

# Checklist

### DIFF
--- a/MaxText/llama4_ckpt_unscanned.py
+++ b/MaxText/llama4_ckpt_unscanned.py
@@ -182,6 +182,16 @@ class _NamespaceMapper:
     return self.collection[new_key]
 
 
+def _pt_to_np(pt_weight, cast_dtype=None, transpose=False):
+  if cast_dtype:
+    np_weight = pt_weight.to(torch.float32).numpy().astype(cast_dtype)
+  else:
+    np_weight = pt_weight.to(torch.float32).numpy()
+  if transpose:
+    np_weight = np_weight.transpose()
+  return np_weight
+
+
 def _convert_huggingface_to_jax_weights(base_model_path: str, model_size: str, model_params: dict, mem_info: psutil.Process):
   """Convert a Huggingface Checkpoint to a dictionary of Numpy arrays representing the weights.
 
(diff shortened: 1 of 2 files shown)


hengtaoguo (approved):
Thank you Aireen for the great work!

I just checked the vision_config between [Scout](https://huggingface.co/meta-llama/Llama-4-Scout-17B-16E-Instruct/blob/main/config.json)/[Maverick](https://huggingface.co/meta-llama/Llama-4-Maverick-17B-128E-Instruct/blob/main/config.json), and they seem to be the same. It appears we can reuse the same scripts for Maverick conversion too.

MaxText/llama4_ckpt_unscanned.py
```diff
@@ -230,8 +233,190 @@ def _convert_huggingface_to_jax_weights(base_model_path: str, model_size: str, m
...
+              "Llama4VisionPixelShuffleMLP_0": {},
+          },
+          "Llama4MultiModalProjector_0": {"vit_multi_modal_projector": {"kernel": None}},
+      },
   }
 
+  # vision model ###########################################
+  max_logging.log("Processing vision model")
+  jax_weights["vision_encoder"]["Llama4VisionModel_0"]["class_embedding"] = (
+      chkpt_vars["vision_model.class_embedding"].to(torch.float32).numpy().astype(CAST_DTYPE)
```
gagika: could you extract the conversions part into a function and re-use it?

def torch_to_numpy(tensor: torch.Tensor, transpose: bool = False):
  """Converts a PyTorch tensor to a NumPy array with the target dtype."""
  result = tensor.to(torch.float32).numpy().astype(CAST_DTYPE)
  if transpose:
    return result.transpose()
  return result

aireenmei: Good point! I added function _pt_to_np and refactored

gagika (approved):
Thanks!

PR #1819, opened 2025-06-11
fix runtime errors in colocated python dataloader

sadikneipp:
### Title

Fix: Resolve runtime errors in colocated TFDS dataloader

### Description
fixes:

*   **`_tfds_data_processing.py`**:
    *   The `tfds.builder` now correctly uses the `dataset_path` variable, which is needed for tfds.
    *   The constructor for `RemoteIterator` has been updated to pass `global_shape` as argument, fixing an interface mismatch.

*   **`multihost_dataloading.py`**:
    *   The `put_to_tpu_devices` function now correctly returns the array after it has been placed on the TPU device. Otherwise the iterator would have empty data.
    *   Initialization logging for `RemoteIterator` is now more informative, including a test output to aid in debugging and validation.

### Tests

This patch was used in the last round of scale tests, validated up to 4 slice scale.

### Checklist

### DIFF
--- a/MaxText/input_pipeline/_tfds_data_processing.py
+++ b/MaxText/input_pipeline/_tfds_data_processing.py
@@ -48,9 +48,10 @@ def get_datasets(
     shuffle_seed,
     dataloading_host_index,
     dataloading_host_count,
+    dataset_path=None,
 ):
   """Load a TFDS dataset."""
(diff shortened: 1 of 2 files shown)


aireenmei (approved):
Thanks Sadi!

PR #1823, opened 2025-06-11
Bug Fixes from Scale Testing And McJAX Long Running Test

SujeethJinesh:
# Description

Fixing some benchmark recipe issues faced during scale testing. Fixes issues in orbax checkpoint logging logic, enables goodput logging by default for pathways benchmarks, and shortens the run name so "pathways-head" does not exceed the 64 character limit.

Also checks in a McJAX long running test script used for testing McJAX paths.

# Tests

Tested during scale testing.

# Checklist

### DIFF
--- a/benchmarks/maxtext_xpk_runner.py
+++ b/benchmarks/maxtext_xpk_runner.py
@@ -477,10 +477,17 @@ def _get_pathways_proxy_flags(wl_config: WorkloadConfig):
     flags_to_add = wl_config.model.pathways_xla_flag_options[
         xla_flags.ADD_PROXY
     ]
-    proxy_flags.append(flags_to_add)
+    flags_to_add_list = flags_to_add.strip().split()
+    proxy_flags += flags_to_add_list
 
   # Join the list of flags back into a single string, space-separated
   return ' '.join(proxy_flags)
+
+
+def _combine_flag_strings(base_flags: str, flags_to_add: str) -> str:
+  """Combines two flag strings and removes extraneous whitespace."""
+  all_flags = base_flags.split() + flags_to_add.split()
+  return ' '.join(all_flags)
 
 
 def _get_pathways_worker_flags(wl_config: WorkloadConfig):
@@ -499,7 +506,8 @@ def _get_pathways_worker_flags(wl_config: WorkloadConfig):
     flags_to_add = wl_config.model.pathways_xla_flag_options[
         xla_flags.ADD_WORKER
     ]
-    worker_flags += flags_to_add
+
+    worker_flags = _combine_flag_strings(worker_flags, flags_to_add)
 
   # Join the list of flags back into a single string, space-separated
   return worker_flags
@@ -521,7 +529,7 @@ def _get_pathways_server_flags(wl_config: WorkloadConfig):
     flags_to_add = wl_config.model.pathways_xla_flag_options[
         xla_flags.ADD_SERVER
     ]
-    server_flags += flags_to_add
+    server_flags = _combine_flag_strings(server_flags, flags_to_add)
 
   # Join the list of flags back into a single string, space-separated
   return server_flags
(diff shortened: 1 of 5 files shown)


benchmarks/maxtext_xpk_runner.py
```diff
@@ -499,7 +500,10 @@ def _get_pathways_worker_flags(wl_config: WorkloadConfig):
     flags_to_add = wl_config.model.pathways_xla_flag_options[
         xla_flags.ADD_WORKER
     ]
-    worker_flags += flags_to_add
+
+    worker_flags_list = worker_flags.strip().split()
```
SurbhiJainUSC: What is the point of converting a string (worker_flags) into a list (worker_flags_list) and then back to a string (worker_flags)? Can't we do like this:
```
worker_flags += ' ' + flags_to_add.strip()
```

MaxText/checkpointing.py
```diff
@@ -383,10 +383,10 @@ def setup_checkpoint_logger(config) -> Any | None:  # pytype: disable=attribute-
   max_logging.log("Setting up checkpoint logger...")
   if config.enable_checkpoint_cloud_logger:
     logger_name = f"goodput_{config.run_name}"
-    options = cloud_logger.CloudLoggerOptions(
+    options = ocp.logging.CloudLoggerOptions(
```
SurbhiJainUSC: Should we fix [this](https://github.com/AI-Hypercomputer/maxtext/blob/ff6c8aca080eb55cde3e26ccbcdba5fbe0ebe8b9/MaxText/checkpointing.py#L52) instead?

benchmarks/recipes/mcjax_long_running_recipe.py
```diff
@@ -0,0 +1,133 @@
...
+  # Handle command line arguments using args_helper
+  should_continue = helper.handle_cmd_args(
+      cluster_config, helper.DELETE, xpk_path=XPK_PATH
+  )
+
+  if not should_continue:
+    return 0
+
+  model_list = [
+      # model_configs.llama3_1_70b_8192_pw_lr_real_data,
```
SurbhiJainUSC, line 68: Do you mean to uncomment these?

MaxText/checkpointing.py
SujeethJinesh: Orbax actually removed the `ocp.logging.cloud_logger`, but that's a good point that we should get rid of that line. Added to the PR.

benchmarks/recipes/mcjax_long_running_recipe.py
SujeethJinesh: No these are kept there for ease of use next time. Easier for us to switch between the specific models. We do this in other recipes like the pathways long running one as well.

https://github.com/AI-Hypercomputer/maxtext/blob/ff6c8aca080eb55cde3e26ccbcdba5fbe0ebe8b9/benchmarks/recipes/pw_long_running_recipe.py#L69-L74

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: We did this just because there's mixing between commas and spaces sometimes, so it was easier for us to just do a split and rejoin in this fashion.

SurbhiJainUSC: The default delimiter for `split()` is whitespace. So, `flags_to_add.strip().split()` would just create a list with flags separated by spaces. So, how would your logic take into account commas?

SujeethJinesh: Ah sorry, it wasn't actually commas that were the issue. It was just that extra spaces would cause the yaml to look like this and fail the jobset. Let me see if we can avoid this with your method.

https://screenshot.googleplex.com/Arp5sEJkpcqhZS8

MaxText/checkpointing.py
```diff
@@ -383,10 +381,10 @@ def setup_checkpoint_logger(config) -> Any | None:  # pytype: disable=attribute-
...
```
SurbhiJainUSC: FYI: Matt has another PR for fixing this: https://github.com/AI-Hypercomputer/maxtext/pull/1825

SujeethJinesh: Thanks! I'll rebase with his changes once they're merged.

benchmarks/maxtext_xpk_runner.py
SujeethJinesh: Ok, so this issue arises from when either worker_flags or flags_to_add has extra spaces, hence why we need a split() for both. To make it simpler I've just made a helper function for it. Any additional whitespace causes the jobset to have null exception issues, hence why we really want to just strip() it as much as we can.

SurbhiJainUSC (approved):

benchmarks/maxtext_xpk_runner.py
SurbhiJainUSC: Thanks for fixing this.

hengtaoguo (approved):

PR #1829, opened 2025-06-13
JetStream Offline Engine 

wenxindongwork:
# JetStream Offline Engine for GRPO 

This pull request introduces the JetStream Offline Engine `offline_engine.py`, an inference engine implemented to support sampling in GRPO, but also intended to be used for generic RL sampling and batch inference. 

Example Usage: 

```
offline_engine = OfflineEngine(
    config=maxtext_config,
    params=None,
    enable_batch_prefill=True,
)

input_data = [
    jax.numpy.arange(80),
    jax.numpy.arange(90),
    jax.numpy.arange(100),
]

results = offline_engine.batch_inference(input_data)

for completion_output in results:
    text = offline_engine.tokenizer.decode(completion_output.token_ids)
    max_logging.log(f"Output: {text}")
```

This Offline Engine works on both McJAX and Pathways. Currently it supports continuous batching and batch prefill. Support for multisampling is yet to be added. 

### Logical Axis Updates for supporting DP

This PR also introduces changes to support a new logical axis for supporting sequence parallel during Prefill. This is required for supporting Offline Engine with our version of data parallel where prefill shards sequence dimension and decode shards batch dimension. 

* Added `PREFILL_LENGTH` to `MaxText/common_types.py` to define prefill activation lengths. 
* Updated logical axis rules in `MaxText/configs/base.yml` and `MaxText/configs/inference.yml` to include `prefill_activation_length` and `prefill_activation_norm_length`. 
* Modified logical constraints in `MaxText/layers/deepseek.py`, `MaxText/layers/llama2.py` , `MaxText/layers/gemma2.py` to dynamically adjust activation sharding based on `MODEL_MODE_PREFILL`.
* Modified `MaxText/layers/attentions.py` to apply logical constraints specific to prefill mode for attention weights, masks, and outputs. 
* Modified `MaxText/layers/embeddings.py` to accept `model_mode` as an argument for embedding operations 
* Modified logical constraints in `MaxText/layers/linears.py` to apply logical constraints specific to prefill mode

### RL specific changes
* Added `@jax.jit` to `log_prob_of_chosen_token` in `MaxText/inference_utils.py` for improved performance.  
* Added logic in `prefill_packing.py` to avoid pre-compiling the batch prefill function as it did not work on Pathways. 
* Added logic in `prefill_packing.py` to conditionally return prompt token log probabilities. This is needed for GRPO loss calculation. 

# Tests

Added `offline_engine_test.py`

# Checklist


khatwanimohit:
@vipannalla @mitalisi Can you help review this PR?

MaxText/configs/inference.yml
```diff
@@ -1,6 +1,8 @@
 base_config: "base.yml"
 
 logical_axis_rules: [
+                      ['prefill_activation_length', ['data']],
```
mitalisi: Same comment as before - we should leave the base config as it is and create an RL config with these changes.

MaxText/configs/base.yml
```diff
@@ -286,6 +286,8 @@ hardware: 'tpu' # Supported hardware types are 'tpu', 'gpu', 'gpu_multiprocess'
 # Parallelism
 mesh_axes: ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'context_autoregressive', 'tensor', 'tensor_transpose', 'tensor_sequence', 'expert', 'autoregressive']
 logical_axis_rules: [
+                      ['prefill_activation_length', ['data']],
```
mitalisi: This should be set same as default activation_length in base.yml. Create a separate config for RL usecase. Same for prefill_activation_norm_length. Also curious if you considered using the sequence dim instead of data which seems better aligned.

mitalisi (commented):
Overall looks good. Few comments.

MaxText/inference_utils.py
```diff
@@ -52,6 +52,7 @@ def str2bool(v: str) -> bool:
     raise ValueError(f"Invalid value '{v}'!")
 
 
+@jax.jit
```
gpolovets1, line 55: Just curious if this is needed since it's being called inside functions that are already jitted?

MaxText/configs/base.yml
mitalisi: Will this will also require changes to configs here: https://source.corp.google.com/piper///depot/google3/third_party/py/maxtext/inference/configs/multi_host/

MaxText/layers/attentions.py
```diff
@@ -985,7 +984,7 @@ def apply_attention_dot(
 
     # special sharding for decode
     q_seq_len = query.shape[1]
-    prefill_qkv_sharding = (BATCH, LENGTH, HEAD, D_KV)
+    prefill_qkv_sharding = (BATCH, PREFILL_LENGTH, HEAD, D_KV)
```
gpolovets1, line 1070: Is this used anywhere?

MaxText/layers/attentions.py
```diff
@@ -1010,7 +1009,7 @@ def apply_attention_dot(
     if self.is_partition_in_decode(q_seq_len):
       attn_weights = partitioning.with_sharding_constraint(attn_weights, (KV_LENGTH, HEAD, None, None, None))
     elif model_mode == MODEL_MODE_PREFILL:
-      attn_weights = partitioning.with_sharding_constraint(attn_weights, (BATCH, HEAD, None, LENGTH, KV_LENGTH))
+      attn_weights = partitioning.with_sharding_constraint(attn_weights, (BATCH, HEAD, None, PREFILL_LENGTH, KV_LENGTH))
```
gpolovets1, line 1095: IIRC "sequence" sharding for prefill was traditionally done on the #heads. Otherwise it becomes more difficult to perform attention if you split up the context into chunks.

MaxText/layers/attentions.py
```diff
@@ -1308,15 +1308,17 @@ class Attention(nn.Module):
...
   # TODO: Further break down the Training and Inference axes for the q, k, v.
-  prefill_query_axis_names: AxisNames = (PREFILL_KV_BATCH, LENGTH, KV_HEAD, KV_HEAD_DIM)
-  prefill_key_axis_names: AxisNames = (PREFILL_KV_BATCH, LENGTH, KV_HEAD, KV_HEAD_DIM)
-  prefill_value_axis_names: AxisNames = (PREFILL_KV_BATCH, LENGTH, KV_HEAD, KV_HEAD_DIM)
+  prefill_query_axis_names: AxisNames = (PREFILL_KV_BATCH, PREFILL_LENGTH, KV_HEAD, KV_HEAD_DIM)
+  prefill_key_axis_names: AxisNames = (PREFILL_KV_BATCH, PREFILL_LENGTH, KV_HEAD, KV_HEAD_DIM)
+  prefill_value_axis_names: AxisNames = (PREFILL_KV_BATCH, PREFILL_LENGTH, KV_HEAD, KV_HEAD_DIM)
   query_axis_names: AxisNames = (KV_BATCH, LENGTH, KV_HEAD, KV_HEAD_DIM)
   input_axis_names: AxisNames = (BATCH, LENGTH, EMBED)
+  prefill_axis_names: AxisNames = (PREFILL_KV_BATCH, PREFILL_LENGTH, EMBED)
```
gpolovets1: nit: prefill_input_axis_names is more informative

MaxText/layers/deepseek.py
```diff
@@ -144,7 +146,11 @@ def __call__(
       slot=None,
   ):
     cfg = self.config
-    inputs = nn.with_logical_constraint(inputs, ("activation_batch", "activation_norm_length", "activation_embed"))
+    if model_mode == MODEL_MODE_PREFILL:
+      axis_names = ("activation_batch", "prefill_activation_norm_length", "activation_embed")
```
gpolovets1: nit: logical_axis_names may be a bit clearer.

MaxText/layers/models.py
```diff
@@ -78,7 +81,10 @@ def __call__(
         epsilon=cfg.normalization_layer_epsilon,
         kernel_axes=("norm",),
     )(inputs)
-    lnx = nn.with_logical_constraint(lnx, ("activation_batch", "activation_length", "activation_embed"))
+    if model_mode == MODEL_MODE_PREFILL:
+      lnx = nn.with_logical_constraint(lnx, ("activation_batch", "prefill_activation_length", "activation_embed"))
```
gpolovets1: Why not create axis_names variable here, as you did before, given this reused multiple times?

MaxText/maxengine.py
```diff
@@ -399,7 +399,7 @@ def prefill_aot(  # pylint: disable=too-many-positional-arguments
         rng=rng,
     )
 
-  @functools.partial(jax.jit, static_argnums=(0,), static_argnames=("request_id",))
+  @functools.partial(jax.jit, static_argnums=(0,), static_argnames=("request_id", "return_prompt_logp"))
```
gpolovets1, line 402: Interesting, I wonder why the other parameters don't need to be added to static_argnames/nums?

gpolovets1: Wondering if request_id should be static as it will force recompilation for each request.. I also am not sure if these argument is being used in the jit functions.

MaxText/prefill_packing.py
```diff
@@ -135,6 +136,7 @@ def _process_compiled(self, params: Params, padded_length: int):
                   self.engine.decode_state_layouts,
               ),
               donate_argnames=("decode_state"),
+              static_argnames=("return_prompt_logp"),
```
gpolovets1: Wondering if instead of adding static_argnames, you should be storing the results in self.process_fn? e.g. self[process_fn][padded_length][return_prompt_lopg] = ...

MaxText/prefill_packing.py
```diff
@@ -115,13 +115,14 @@ def process(
       input_tokens_padded: jax.Array,
       input_true_length: int,
       rng: PRNGKeyType,
+      return_prompt_logp: bool = False,
   ) -> Tuple[engine_api.ResultTokens, DecodeState]:
     """Process a new input."""
 
-    process_fn = self._process_compiled(model_params, len(input_tokens_padded))
-    return process_fn(model_params, input_tokens_padded, decode_slot, input_true_length, decode_state, rng)
```
gpolovets1, line 122: Is there a reason for removing rng?

MaxText/prefill_packing.py
```diff
@@ -264,24 +290,45 @@ def zero_padded(arr: list[int], padding: int):
...
     for i in range(bucket.count):
-      prefill_result.append((first_tokens[i], bucket.slots[i]))
+      if return_prompt_logp:
+        prompt_logp = prompt_logp_numpy[:, offsets[i] : offsets[i] + lengths[i]]
+        prefill_result.append(PrefillResult(first_tokens[i], bucket.slots[i], prompt_logp))
+      else:
+        prefill_result.append(PrefillResult(first_tokens[i], bucket.slots[i], None))
     return prefill_result, decode_state
 
   def _process_batch_compiled(self, params: Params, padded_length: int, capacity: int, num_prompts: int):
```
gpolovets1: Do you need to add return_prompt_lopg arg here? And you may also need to store it in self.process_batch_func.

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+@dataclasses.dataclass
+class TokenOutput:
+  """Container for individual token generation result."""
+
+  token: np.ndarray
+  log_prob: np.ndarray
+
+
+class SafeThread(threading.Thread):
+  """Thread class with exception handling to prevent silent failures."""
```
gpolovets1, line 116: Awesome!

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+    self._type = prefill_type
+    self.engine = engine
+    self.prefill_lengths = sorted(prefill_lengths)
+    self.max_prefill_length = self.prefill_lengths[-1]
+    self.batch_prefill_max_batch_size = batch_prefill_max_batch_size
+    self.rng = jax.random.PRNGKey(0) if rng is None else rng
+    if prefill_type == PrefillType.DEFAULT:
+      self._processor = PrefillProcessor(engine)
+    elif prefill_type == PrefillType.BATCH:
+      self._batch_processor = BatchedPrefillProcessor(
```
gpolovets1, line 176: This could make more sense to me during Disagg, but does this help in the case of continual batching?

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+          max_batch_size=batch_prefill_max_batch_size,
+          auto_layout_supported=False,
+      )
+      # Keep fallback processor for edge cases
+      self._processor = PrefillProcessor(engine)
+    else:
+      raise ValueError(f"Invalid prefill type: {prefill_type}")
+
+  @functools.partial(jax.jit, static_argnums=(0), donate_argnames=("decode_state",))
+  def _jitted_single_prefill(
```
gpolovets1, line 187: Why not use _process_compiled instead?

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+        model_params: Model parameters for inference
+        decode_state: Current decode state
+        decode_slot: The decode slot index to use for this input
+        input_id: Unique identifier for this input
+        input_tokens_padded: Padded token array for the input
+        input_true_length: Actual length of the input before padding
+        prefill_done: Callback function called when prefill completes
+    """
+    padded_length = len(input_tokens_padded)
+    # Use default processor if configured or if input is already at max length
```
gpolovets1, line 230: Why can you not batch prefills that are already at max_prefill_length?

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+    self.max_decode_length = max_decode_length
+    self.eos_ids = eos_ids
+    self.tokenizer = tokenizer
+    self.batch_prefill_max_batch_size = batch_prefill_max_batch_size
+    self.min_decode_steps = min_decode_steps
+    self.mesh = mesh
+    self.rng = jax.random.PRNGKey(0) if rng is None else rng
+    self.debug = debug
+
+    # Inference state (initialized later)
```
gpolovets1, line 368: nit: it could be worth containerizing related variables into a class

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+      destination_sharding: jax.sharding.NamedSharding,
+      is_pw_reshard: bool,
+  ):
+    """Update the model parameters"""
+    if is_pw_reshard:
+      with (
+          jax.transfer_guard_device_to_host("disallow_explicit"),
+          jax.transfer_guard_host_to_device("disallow_explicit"),
+      ):
+        self.params = pathwaysutils_reshard.reshard(params, destination_sharding, cache_resharding_plans=True)
```
gpolovets1, line 447: Interesting, why is this needed? MaxText previously only used device_put and I don't think it needed a code change for Pathways.

MaxText/inference/offline_engine.py
```diff
@@ -0,0 +1,943 @@
...
+    self.running = True
+    self.true_lengths = {input.id: input.true_length for input in data}
+
+    max_logging.log("Continous batching started")
+
+    self._run_continous_batching(data)
+
+    return self._build_final_outputs(data)
+
+  def _run_continous_batching(
```
gpolovets1, line 477: You could also consider adding a transfer queue that may help for your batched prefill case, where you can add batches of prefills to the queue and pull from it as decode slots fill up.

MaxText/inference/offline_engine.py
wenxindongwork: @khatwanimohit added the code for resharding

MaxText/inference/offline_engine.py
wenxindongwork: I believe the current batch prefill logic already uses a queue. The batch prefill helper queues up incoming prefill requests and dispatches the batch once the sequence is filled up.

MaxText/layers/attentions.py
wenxindongwork: yes, but for GRPO we are using sequence parallel *with TP* for Prefill (and DP+TP for decode), so the heads will already be sharded by TP.  In decode we split up the batch dimension, and in prefill we split up the sequence dimension, amongst the model replicas. 

When performing attention during prefill, all devices participate in the all-gathers, so it might introduce some overhead. 

I should make this clearer in the PR description and in the code. Thanks George for the question, and please lmk if this sounds right to you!

MaxText/configs/base.yml
wenxindongwork: We can't directly use the sequence dimension here. It shards attention by head. It also shards the vocab embedding differently https://buganizer.corp.google.com/issues/422995498#comment2.

MaxText/inference_utils.py
wenxindongwork: It's not needed when this function is called within another jitted function (which is the case with GRPO), but I added the jit decorator so that this function could be invoked standlone as well

MaxText/layers/attentions.py
wenxindongwork: I believe it's being used to shard the qkv tensors a few lines down.

MaxText/layers/attentions.py
wenxindongwork: great suggestion, thank you!

MaxText/layers/deepseek.py
wenxindongwork: agreed, thank you!

MaxText/layers/models.py
wenxindongwork: I should have done that, thank you for catching this!

MaxText/maxengine.py
wenxindongwork: you are right! Request id should not be static, and since it is not being used in the `_prefill_jit` function, I went ahead and removed the arg.

MaxText/prefill_packing.py
wenxindongwork: This argument is not expected to change during the runtime of the RL program and it's set in the base config. I agree that it would matter if this argument changes.

wenxindongwork: I still think your idea of `self[process_fn][padded_length][return_prompt_lopg]` is safer, since perhaps one day there will be cases where this variable changes during the runtime.

MaxText/prefill_packing.py
wenxindongwork: `PrefillResult` expects None for prompt_logp if it's not specified. I realized I didn't pass the `return_prompt_logp` arg to `self._process_batch`, thank you for catching this!

MaxText/inference/offline_engine.py
wenxindongwork: I ported this over from the MLPerf offline inference script :D

MaxText/inference/offline_engine.py
wenxindongwork: hmm what do you mean? Do you mean we should always use one or the other for continuous batching?

MaxText/inference/offline_engine.py
wenxindongwork: `_process_compiled` is a AOT function but Pathways (when I tried) did not support the JetStream AOT functions. `_jitted_single_prefill` is the none AOT version of that.

MaxText/inference/offline_engine.py
wenxindongwork: ah I see your later comment, to answer the question of whether it's beneficial to use batch prefill with continuous batching: 

The cost of doing batch prefill is that some decode slots are "wasted" while the bucket is being filled up. The batch prefill bucket "hoards" the decode slot while waiting for the bucket to fill up. However, the pro is that each prefill is filled to max length and we do less prefill calls. 

The optimal scenario is where there is not much waiting for the buckets to fill up. This happens at the beginning of the batch inference. 

It's probably a case by case scenario whether it is beneficial to do continuous batching with batch prefill or non batched prefill.

MaxText/inference/offline_engine.py
khatwanimohit: device_put triggers the transfer of parameters through controller if there is change in sharding from source to destination. We use this reshard utility which creates an efficient plan for transfer of parameters

MaxText/prefill_packing.py
wenxindongwork: nope, was an oversight, thank you for catching this!

MaxText/inference/offline_engine.py
wenxindongwork: thanks for this suggestion, let me do that in a future PR since I don't have access to VMs right now and it's hard for me to develop.

MaxText/configs/inference.yml
wenxindongwork: thanks Mitali for this suggestion, I created a new rl.yml file and left the base configs unchanged

wenxindongwork:
Thank you George so much for the detailed review!!

> * How does batched prefill benefit if typicall we insert into decode one slot at a time?

Replied to your other comment above. 

> * I think traditionally "sequence" sharding during prefill has been applied to the #heads but I think you are applying it to the sequence dimension? If that is the case, my concern is that you are actually performing context parallelism which may add some complications (e.g. see design doc for context parallelism): https://docs.google.com/document/d/1lZ7YXOjWutTo2LvvTmZXpz1KIji6hkscS0MjzR5qjCM

Yes, this design is indeed doing context parallel which shards the sequence dimension. The word "sequence parallel" is overloaded here and I should have just called it context parallel instead :) @A9isha hi Anisha, since we are implementing TP + CP for Prefill here and you are the author of context parallelism, do you see any problems with the current design?

MaxText/inference/offline_engine.py
wenxindongwork: I asked Xiang who implemented batch prefill and he said that there is no need to pass in all the metadata required by batch prefill when the prefill can be invoked directly.

khatwanimohit (approved):
LGTM!

MaxText/layers/attentions.py
gpolovets1: Thanks for explaining WenXin! Yes I was imagining it could cause an all-gather but seems you are already aware =]

PR #1833, opened 2025-06-14
Refactor profiler and fix profiling activation/deactivation only when enabled

SurbhiJainUSC:
# Description
* This PR refactors profiler code that we have in all trainers.
* Additonally, activate and deactivate profiler only when profiling is enabled.

# Tests

Verified that the profile is uploaded successfully on TensorBoard: 
image

# Checklist

### DIFF
--- a/MaxText/profiler.py
+++ b/MaxText/profiler.py
@@ -40,6 +40,18 @@ class Profiler:
     self.profile_period = config.profile_periodically_period
     self.start_initial_profile_step = self._set_first_profiler_step(config.skip_first_n_steps_for_profiler, offset_step)
     self.finished_initial_profile_step = self._set_last_profiler_step(config.profiler_steps, config.steps)
+    if config.profiler != "" and self.start_initial_profile_step >= config.steps:
+      raise ValueError("Profiling requested but initial profiling step set past training final step")
+
+  def maybe_activate_profiler(self, step, state):
+    """Conditionally activates the profiler based on the current step.
+    This method checks if the current training step matches the step designated
+    for starting an initial profile, or if it meets the criteria for
+    activating a new periodic profile.
+    """
+    if self.mode != "" and (step == self.start_initial_profile_step or self.should_activate_periodic_profile(step)):
+      optional_postfix = f"step_{step}" if self.profile_period > 0 else ""
+      self.activate(blocking_object=state, optional_postfix=optional_postfix)
 
   def activate(self, blocking_object=None, optional_postfix=""):
     """Start the profiler.
@@ -59,6 +71,15 @@ class Profiler:
       self.libcudart.cudaProfilerStart()
     elif self.mode == "xplane":
       jax.profiler.start_trace(self.output_path)
+
+  def maybe_deactivate_profiler(self, step, state):
+    """Conditionally deactivates the profiler based on the current step.
+    This method checks if the current training step matches the step designated
+    for finishing the initial profile, or if it meets the criteria for
+    deactivating a periodic profile.
+    """
+    if self.mode != "" and (step == self.finished_initial_profile_step or self.should_deactivate_periodic_profile(step)):
+      self.deactivate(blocking_object=state)
 
   def deactivate(self, blocking_object=None):
     """End the profiler.
--- a/MaxText/elastic_train.py
+++ b/MaxText/elastic_train.py
@@ -236,10 +236,6 @@ def train_loop(config, elastic_manager, recorder, state=None):
 
   start_step = get_first_step(state)  # this is the start_step for training
   prof = profiler.Profiler(config, offset_step=start_step)
-  first_profiling_step = prof.start_initial_profile_step
-  if config.profiler != "" and first_profiling_step >= config.steps:
-    raise ValueError("Profiling requested but initial profiling step set past training final step")
-  last_profiling_step = prof.finished_initial_profile_step
 
   example_batch = None
   last_step_completion = datetime.datetime.now()
@@ -271,9 +267,7 @@ def train_loop(config, elastic_manager, recorder, state=None):
   # the step is restored back to the latest snapshot when a slice is lost
   while step < config.steps:
     try:
-      if step == first_profiling_step or prof.should_activate_periodic_profile(step):
-        optional_postfix = f"step_{step}" if config.profile_periodically_period > 0 else ""
-        prof.activate(blocking_object=state, optional_postfix=optional_postfix)
+      prof.maybe_activate_profiler(step, state)
 
       max_logging.log(f"{step=} {elastic_manager.elastic_down_event_count=} {elastic_manager.good_slice_count=}")
       with mesh, nn_partitioning.axis_rules(config.logical_axis_rules), jax.default_device(elastic_manager.default_device):
@@ -310,8 +304,7 @@ def train_loop(config, elastic_manager, recorder, state=None):
 
         metric_logger.write_metrics(running_gcs_metrics, metrics, step)
 
-        if step == last_profiling_step or prof.should_deactivate_periodic_profile(step):
-          prof.deactivate(blocking_object=state)
+        prof.maybe_deactivate_profiler(step, state)
 
       elastic_manager.maybe_snapshot(
           step=step,
--- a/MaxText/experimental/rl/grpo_trainer.py
+++ b/MaxText/experimental/rl/grpo_trainer.py
@@ -779,10 +779,6 @@ def train_loop(config, config_inference, recorder, state=None):
 
   start_step = get_first_step(state)  # this is the start_step for training
   prof = profiler.Profiler(config, offset_step=start_step)
-  first_profiling_step = prof.start_initial_profile_step
-  if config.profiler != "" and first_profiling_step >= config.steps:
-    raise ValueError("Profiling requested but initial profiling step set past training final step")
-  last_profiling_step = prof.finished_initial_profile_step
 
   example_batch = None
   last_step_completion = datetime.datetime.now()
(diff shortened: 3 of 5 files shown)


SurbhiJainUSC:
> @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> 
> ```
> Default value for global ARG results in an empty or invalid base image name
> 
> InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> ```

This warning appears because there is no default value defined for `JAX_AI_IMAGE_BASEIMAGE`. If the intent is to always explicitly provide the value for `JAX_AI_IMAGE_BASEIMAGE` while building the docker, then Dockerfile is technically correct and we can ignore the warning.

Rohan-Bierneni:
> @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> 
> ```
> Default value for global ARG results in an empty or invalid base image name
> 
> InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> ```

As @SurbhiJainUSC  pointed out, since the JAX AI Image Dockerfile uses a base image that is specified via a --build-arg, when this is not passed to the docker build command this error would be thrown. We could set default value for the arg JAX_AI_IMAGE_BASEIMAGE but it could become an old image for the default value if it doesn't get updated periodically.

bvandermoon:
> > @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> > ```
> > Default value for global ARG results in an empty or invalid base image name
> > 
> > InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> > More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> > ```
> 
> As @SurbhiJainUSC pointed out, since the JAX AI Image Dockerfile uses a base image that is specified via a --build-arg, when this is not passed to the docker build command this error would be thrown. We could set default value for the arg JAX_AI_IMAGE_BASEIMAGE but it could become an old image for the default value if it doesn't get updated periodically.

Thanks @Rohan-Bierneni and @SurbhiJainUSC. @Rohan-Bierneni Do you know how we can get the warning to stop showing up on unrelated PRs?

SurbhiJainUSC:
> > > @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> > > ```
> > > Default value for global ARG results in an empty or invalid base image name
> > > 
> > > InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> > > More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> > > ```
> > 
> > 
> > As @SurbhiJainUSC pointed out, since the JAX AI Image Dockerfile uses a base image that is specified via a --build-arg, when this is not passed to the docker build command this error would be thrown. We could set default value for the arg JAX_AI_IMAGE_BASEIMAGE but it could become an old image for the default value if it doesn't get updated periodically.
> 
> Thanks @Rohan-Bierneni and @SurbhiJainUSC. @Rohan-Bierneni Do you know how we can get the warning to stop showing up on unrelated PRs?

I believe the warning comes from the tests that run on every PR. This dockerfile is used to build the docker image for every PR to run the tests. So, I don't think so we can stop this warning unless we assign any default value as @Rohan-Bierneni mentioned.

Rohan-Bierneni:
> > > > @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> > > > ```
> > > > Default value for global ARG results in an empty or invalid base image name
> > > > 
> > > > InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> > > > More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> > > > ```
> > > 
> > > 
> > > As @SurbhiJainUSC pointed out, since the JAX AI Image Dockerfile uses a base image that is specified via a --build-arg, when this is not passed to the docker build command this error would be thrown. We could set default value for the arg JAX_AI_IMAGE_BASEIMAGE but it could become an old image for the default value if it doesn't get updated periodically.
> > 
> > 
> > Thanks @Rohan-Bierneni and @SurbhiJainUSC. @Rohan-Bierneni Do you know how we can get the warning to stop showing up on unrelated PRs?
> 
> I believe the warning comes from the tests that run on every PR. This dockerfile is used to build the docker image for every PR to run the tests. So, I don't think so we can stop this warning unless we assign any default value as @Rohan-Bierneni mentioned.

@SurbhiJainUSC @bvandermoon To summarize, the issue stems from the JAX AI Image Dockerfile requiring a specific docker image to be specified that gets used as a base-image. However, Github has a "beta" feature called "unchanged files with check annotations" that flags this Dockerfile since there is a --build-arg which doesn't have a default value i.e $JAX_AI_BASE_IMAGE. 

We can either do follow the Github Annotations rules and set a default value so that this Dockerfile doesn't get flagged, or we can disable the "Unchanged files with Check Annotations" feature. I found a discussion post on how to do this. It would require a change in the workflow file: https://github.com/orgs/community/discussions/8379#discussioncomment-11943317

@bvandermoon which option would be best here. And if going with option 1 what default value for the JAX AI Image would be best?

bvandermoon:
> > > > > @parambole @Rohan-Bierneni do you know what is causing this message in `maxtext_jax_ai_image.Dockerfile`? I am seeing it in other PRs also so it is not specific to this PR
> > > > > ```
> > > > > Default value for global ARG results in an empty or invalid base image name
> > > > > 
> > > > > InvalidDefaultArgInFrom: Default value for ARG $JAX_AI_IMAGE_BASEIMAGE results in empty or invalid base image name
> > > > > More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/
> > > > > ```
> > > > 
> > > > 
> > > > As @SurbhiJainUSC pointed out, since the JAX AI Image Dockerfile uses a base image that is specified via a --build-arg, when this is not passed to the docker build command this error would be thrown. We could set default value for the arg JAX_AI_IMAGE_BASEIMAGE but it could become an old image for the default value if it doesn't get updated periodically.
> > > 
> > > 
> > > Thanks @Rohan-Bierneni and @SurbhiJainUSC. @Rohan-Bierneni Do you know how we can get the warning to stop showing up on unrelated PRs?
> > 
> > 
> > I believe the warning comes from the tests that run on every PR. This dockerfile is used to build the docker image for every PR to run the tests. So, I don't think so we can stop this warning unless we assign any default value as @Rohan-Bierneni mentioned.
> 
> @SurbhiJainUSC @bvandermoon To summarize, the issue stems from the JAX AI Image Dockerfile requiring a specific docker image to be specified that gets used as a base-image. However, Github has a "beta" feature called "unchanged files with check annotations" that flags this Dockerfile since there is a --build-arg which doesn't have a default value i.e $JAX_AI_BASE_IMAGE.
> 
> We can either do follow the Github Annotations rules and set a default value so that this Dockerfile doesn't get flagged, or we can disable the "Unchanged files with Check Annotations" feature. I found a discussion post on how to do this. It would require a change in the workflow file: https://github.com/orgs/community/discussions/8379#discussioncomment-11943317
> 
> @bvandermoon which option would be best here. And if going with option 1 what default value for the JAX AI Image would be best?

Agree with you both that we don't want to include a default since it will just go out of date.

Is there a way to set an ignore flag like we do for linting? https://github.com/AI-Hypercomputer/maxtext/blob/f827a0304acbb1ad446270d5dc7e44688ce081e0/MaxText/checkpointing.py#L40

If not, I guess we can disable the feature in the workflow. Hopefully there is a way to just ignore though

hengtaoguo (approved):

SurbhiJainUSC:
> LGTM. Just one question which is outside of the scope of this PR - would it be cleaner to just use the config values directly for fields that directly set an equivalent field? Not sure how much value there is in a line like this:
> 
> ```
> self.profile_cleanly = config.profile_cleanly
> ```
> 
> Let me know if there is some convention here for not wanting to set the full config as a self variable or something

I don't think there is an issue setting `config` as a class variable too. I believe the only reason to set the profiler specific variables is just for clarity. 

For example, to access `profile_cleanly`:
`self.profile_cleanly` is more cleaner as compared to `self.config.profile_cleanly` imo. 

I am open to change it if you think otherwise. @bvandermoon

MaxText/profiler.py
```diff
@@ -40,6 +40,18 @@ def __init__(self, config, offset_step=0):
     self.profile_period = config.profile_periodically_period
     self.start_initial_profile_step = self._set_first_profiler_step(config.skip_first_n_steps_for_profiler, offset_step)
     self.finished_initial_profile_step = self._set_last_profiler_step(config.profiler_steps, config.steps)
+    if config.profiler != "" and self.start_initial_profile_step >= config.steps:
+      raise ValueError("Profiling requested but initial profiling step set past training final step")
+
+  def maybe_activate_profiler(self, step, state):
```
SurbhiJainUSC, line 46: Noticed that `maybe_activate_profiler` and `maybe_deactivate_profiler` should also have a `config.profiler!=""` check to ensure we only call `activate()` and `deactivate()` when profiling is enabled. Is that right understanding @bvandermoon ?

PR #1836, opened 2025-06-15
pyconfig → pydantic

SamuelMarks:
# Description

## Background

MaxText is currently engineered around pyconfig. Pyconfig—[https://pypi.org/project/pyconfig/](https://pypi.org/project/pyconfig/)—was last updated in 2017 and has [20k monthly downloads](https://pypistats.org/packages/pyconfig).

Pydantic—[https://pypi.org/project/pydantic/](https://pypi.org/project/pydantic/)—is constantly updated and has [hundreds of millions of monthly downloads](https://pypistats.org/packages/pydantic).

\> Pydantic is the most widely used data validation library for Python.

[https://docs.pydantic.dev](https://docs.pydantic.dev/latest/)

## Summary

The TL;DR version is:

* pydantic has become basically the standard for configuration formats, specifying the inputs and outputs (e.g., for REST APIs in FastAPI framework)  
* pydantic links in well with the Python type-checker so you can oft find errors before runtime  
* pydantic errors are clean and clear  
* with my python compiler, you can have clean CLIs with `--help` and GUIs and generate SQL models (SQLAlchemy) as desired  
* SDK documentation becomes much cleaner (which is good in preparation for the pypi release \+ hosted doc pages)

## Migration

Proposed changes to the MaxText codebase:

1. Completely remove the pyconfig dependency  
   1.  maybe not immediately so there’s a chance for people to easily migration of their existing setups, e.g., with new functions \`from\_pyconfig\_to\_pydantic\`  
2. Migrate all examples, and codebase occurrences with new pydantic types  
3. Put pydantic types on a computer+human understandable hierarchy, e.g.:  
   1. One global \`types.py\`, or  
   2. One \`types.py\` per config occurrence (e.g., one per module if each module has a different config)  
4. Create new CLI that uses common CLI syntax (e.g., this can be automatically created using my Python compiler [https://github.com/offscale/cdd-python](https://github.com/offscale/cdd-python))  
5. Migrate all shell scripts and docs to use this new CLI  
   1. TBD: remove the shell scripts in favour of Python SDK usage.

# Tests

Tested the change on your CI.

# Checklist

### DIFF
--- a/MaxText/benchmark_chunked_prefill.py
+++ b/MaxText/benchmark_chunked_prefill.py
@@ -46,7 +46,7 @@ from absl import app
 
 from MaxText import max_utils
 from MaxText import maxengine
-from MaxText import pyconfig
+import MaxText.configs.loader
 
 _WARMUP_ITERS = 2
 _BENCHMARK_ITERS = 5
@@ -373,7 +373,7 @@ def prepare_setting(argv: Sequence[str]):
   """
   jax.config.update("jax_default_prng_impl", "unsafe_rbg")
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
-  config = pyconfig.initialize(argv)
+  config = MaxText.configs.loader.initialize(argv)
   max_utils.print_system_information()
 
   prefix_caching_hbm_byte = config.prefix_caching_hbm_byte
--- a/MaxText/configs/__init__.py
+++ b/MaxText/configs/__init__.py
@@ -1,0 +1,15 @@
+"""
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
--- a/MaxText/configs/loader.py
+++ b/MaxText/configs/loader.py
@@ -1,0 +1,89 @@
+"""
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import sys
+from typing import Sequence
+
+import yaml
+
+from MaxText.configs.types import MaxTextConfig
+
+def load_config_from_yaml_and_argv(argv: Sequence[str]) -> MaxTextConfig:
+    """
+    Loads the MaxTextConfig from a YAML file specified in argv[1]
+    and applies any command line overrides of the form key=value.
+
+    Args:
+        argv: sys.argv-like list of arguments.
+            argv[1] should be path to YAML file.
+            Subsequent arguments optionally in key=value format as overrides.
+
+    Returns:
+        An instance of MaxTextConfig.
+    """
+
+    if len(argv) < 2:
+        raise ValueError("Please specify config YAML path as first argument")
+
+    config_path = argv[1]
+
+    # Load YAML config file into dictionary
+    with open(config_path, "r") as f:
+        config_dict = yaml.safe_load(f)
+
+    # Parse overrides from argv (key=value)
+    for arg in argv[2:]:
+        if "=" not in arg:
+            continue
+        key, val = arg.split("=", 1)
+        # Convert val to proper type
+        try:
+            # Use yaml to parse val to proper python types (int, float, bool, etc)
+            parsed_val = yaml.safe_load(val)
+        except yaml.YAMLError:
+            parsed_val = val
+        # Support nested keys separated by dots
+        apply_override(config_dict, key.strip(), parsed_val)
+
+    # pydantic config object with applied overrides
+    return MaxTextConfig(**config_dict)
+
+def apply_override(config_dict: dict, key: str, value):
+    """
+    Modifies config_dict in-place to apply override given a "dot" separated key.
+
+    Example:
+        key="model.base_emb_dim"
+        value=1024
+
+        Will set config_dict['model']['base_emb_dim'] = 1024
+    """
+    keys = key.split(".")
+    d = config_dict
+    for k in keys[:-1]:
+        if k not in d or not isinstance(d[k], dict):
+            d[k] = {}
+        d = d[k]
+    d[keys[-1]] = value
+
+initialize = load_config_from_yaml_and_argv
+
+if __name__ == "__main__":
+    # Simple CLI demo
+    cfg = load_config_from_yaml_and_argv(sys.argv)
+    print(cfg.model_dump_json(indent=4))
+
+__all__ = ["initialize", "load_config_from_yaml_and_argv"]
--- a/MaxText/configs/loader.py
+++ b/MaxText/configs/loader.py
@@ -1,89 +1,117 @@
 """
-Copyright 2025 Google LLC
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-     https://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+Config loader that manually builds pydantic classes from dictionaries and CLI overrides.
 """
 
-import sys
-from typing import Sequence
+import os
+from typing import Any, Dict, Optional
 
 import yaml
 
-from MaxText.configs.types import MaxTextConfig
+from MaxText.configs.types import (
+    MaxTextConfig,
+    CoreConfig,
+    ModelConfig,
+    CheckpointConfig,
+    OptimizerConfig,
+    DatasetConfig,
+    TokenizerConfig,
+    ParallelismConfig,
+    InferenceConfig,
+)
 
-def load_config_from_yaml_and_argv(argv: Sequence[str]) -> MaxTextConfig:
+
+def _merge_dicts(base: Dict[str, Any], override: Dict[str, Any]) -> Dict[str, Any]:
+    """Recursively merge two dicts, with `override` taking priority."""
+    merged = dict(base)
+    for k, v in override.items():
+        if k in merged and isinstance(merged[k], dict) and isinstance(v, dict):
+            merged[k] = _merge_dicts(merged[k], v)
+        else:
+            merged[k] = v
+    return merged
+
+
+def load_yaml(path: str) -> Dict[str, Any]:
+    with open(path, "rt", encoding="utf8") as f:
+        return yaml.safe_load(f) or {}
+
+
+def load_config(
+        config_path: str, overrides: Optional[Dict[str, Any]] = None, base_dir: Optional[str] = None
+) -> MaxTextConfig:
     """
-    Loads the MaxTextConfig from a YAML file specified in argv[1]
-    and applies any command line overrides of the form key=value.
-
-    Args:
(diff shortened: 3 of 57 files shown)


MaxText/configs/types_j.py
```diff
@@ -0,0 +1,1878 @@
...
+
+# ----------------------------------------------------------------------------
+# Pydantic models for configuration
+# ----------------------------------------------------------------------------
+
+
+class RunInfo(BaseModel):
+    """Configuration for the overall run, model identity, and logging."""
+
+    run_name: str = Field(
```
richjames0: nit: this would fit on a single line with MaxText's line limit (which I'm very much in favour of)

MaxText/configs/types_j.py
```diff
@@ -0,0 +1,1878 @@
...
+
+def _parse_cli_key_value_pairs(args: List[str]) -> Dict[str, Any]:
+    """Parses key=value pairs from a list of strings."""
+    overrides = {}
+    for arg in args:
+        if "=" in arg:
+            key, value_str = arg.split("=", 1)
+            key = key.strip()
+            value_str = value_str.strip()
+            try:
```
richjames0: nit: i'm sure i'll surprise you 0% by saying i don't love this function. a match statement would help clean it up a bit and maybe finding a way to avoid all the exception handling

richjames0 (changes requested):
LGTM pending line lengths which make it hard to read

MaxText/configs/types_j.py
richjames0: also i am not the world's biggest fan of aliases but in think case i would be tempted to alias Field as F and save a zillion characters in this file (and have to wrap fewer lines)

richjames0: also i am not the world's biggest fan of aliases but in think case i would be tempted to alias Field as F and save a zillion characters in this file (and have to wrap fewer lines)

SamuelMarks: Yeah the alias thing could work; but I do think the full name is more readable… for the hosted-web API Doc generation especially.

MaxText/configs/types_j.py
SamuelMarks: You're not wrong. But switch/case was only introduced in 3.10

Oh wait you guys only need 3.10+ sure I can switch to that new syntax

SamuelMarks: Indentation all the way isn't super readable but whatever it works and I'm confident it's efficient enough:
```py
    for arg in args:
        if "=" in arg:
            key, value_str = arg.split("=", 1)
            key = key.strip()
            value_str = value_str.strip()
            try:
                value = int(value_str)
            except ValueError:
                try:
                    value = float(value_str)
                except ValueError:
                    match value_str:
                        case s if s.lower() in ("true", "t", "1"):
                            value = True
                        case s if s.lower() in ("false", "f", "0"):
                            value = False
                        case s if "," in s and "[" not in s:
                            value = [item.strip() for item in s.split(",")]
                        case _:
                            try:
                                value = yaml.safe_load(value_str)
                            except (yaml.YAMLError, AttributeError):
                                value = value_str
```

SamuelMarks:
> What do the file name letters stand for? `types_j`, `types_g`, etc.

@bvandermoon Oh ignore that. Each is a different attempt (in lexicographical order). All will be removed with a singular `types.py` to take its place when I rebase this PR to `1 commit`.

src/MaxText/pyconfig_deprecated.py
```diff
@@ -0,0 +1,1301 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pytype: skip-file
+# pylint: disable=missing-module-docstring, bare-except, consider-using-generator, missing-function-docstring
+from collections import OrderedDict
+from typing import Any
+from math import prod
```
bvandermoon, line 19: What does this file name mean?

SamuelMarks: This is the original pyconfig file. og=original

Over time I expect this file to be deleted. It's kept in for now; need to replace all config usage across the entire codebase with pydantic types before we can remove it.

bvandermoon (commented):
Generally LGTM. Can you please include the manual tests you ran in the PR description?

src/MaxText/pyconfig_deprecated.py
bvandermoon: Thanks @SamuelMarks. Really minor but can you please rename this to something like `pyconfig_deprecated.py` or something similar?

tests/pyconfig_test.py
```diff
@@ -19,7 +19,7 @@
 
 from MaxText import pyconfig
 from MaxText.globals import MAXTEXT_PKG_DIR
-from MaxText.pyconfig import resolve_config_path
+from MaxText.pyconfig_og import resolve_config_path
```
bvandermoon: Can you please add an equivalent test with the new config?

tests/pyconfig_test.py
```diff
@@ -130,10 +129,6 @@ def test_overriding_model(self):
     self.assertEqual(config.base_emb_dim, 1024)
     self.assertEqual(config.base_mlp_dim, 24576)
 
-  def test_resolve_config_path(self):
-    self.assertEqual(resolve_config_path("foo"), os.path.join("src", "foo"))
-    self.assertEqual(resolve_config_path(__file__), __file__)
```
bvandermoon, line 135: Why is this not needed anymore?

src/MaxText/pyconfig.py
```diff
@@ -13,1289 +13,116 @@
...
-      "gpt3-22b",
-      "gpt3-6b",
-      "gpt3-52k",
-      "gpt-oss-20b",
-      "gpt-oss-120b",
-      "llama4-17b-16e",
-      "llama4-17b-128e",
-  )
-  if s not in valid_model_names:
-    raise ValueError(f"Invalid model name was passed. Got {s}, Valid options {valid_model_names}")
```
bvandermoon, line 509: Why aren't we validating model names anymore?

src/MaxText/configs/types.py
```diff
@@ -0,0 +1,1362 @@
...
+
+  NONE = ""
+  INT8 = "int8"
+  INTMP = "intmp"
+  FP8 = "fp8"
+  NANOO_FP8 = "nanoo_fp8"
+  FP8_NANO_V2 = "fp8_nanoo"
+  FP8_GPU = "fp8_gpu"
+  FP8_FULL = "fp8_full"
+
```
bvandermoon: Just for my understanding, how did you generate all the options for each type? I see some of these listed in [`base.yml` comments](https://github.com/AI-Hypercomputer/maxtext/blob/9b2f657683f411c260c76dda610715f80376f711/src/MaxText/configs/base.yml#L91-L98), others listed [elsewhere in MaxText](https://github.com/AI-Hypercomputer/maxtext/blob/9b2f657683f411c260c76dda610715f80376f711/docs/explanations/quantization.md?plain=1#L51-L56), etc.

Also, do we have a mismatch in what is [actually allowed to be called](https://github.com/AI-Hypercomputer/maxtext/blob/9b2f657683f411c260c76dda610715f80376f711/src/MaxText/pyconfig.py#L278) in MaxText vs. what is intended to be supported? Or is this method not actually getting picked up?

src/MaxText/pyconfig.py
SamuelMarks: Ignore the red, GitHub / git is confused about the file. The old file is verbatim unchanged, just `mv src/MaxText/pyconfig.py src/MaxText/pyconfig_deprecated.py`.

src/MaxText/configs/types.py
SamuelMarks: The idea is to use _exclusively_ pydantic types throughout the codebase. Whence incongruity presents: pydantic type changes. All custom input validation is to be moved up to pydantic, and then [¿slowly?] every class, function, CLI, etc. becomes completely pydantic.

tests/pyconfig_test.py
SamuelMarks: Ignore the red, git / GitHub is confused about the file. The old file is verbatim unchanged, just `mv tests/pyconfig_test.py tests/pyconfig_deprecated_test.py`.

src/MaxText/pyconfig.py
bvandermoon: Yes, agreed Github is confused about the file. What I am meaning to ask is why don't we need this validation in the new pyconfig?

SamuelMarks: All the existing validation is still happening. The configuration is still passed to the deprecated pyconfig.

bvandermoon: Isn't that just temporary though? What will happen when that is turned off?

SamuelMarks: I can do this, if acceptable:
```py
type ModelName = "default" | "llama2-7b"| "llama2-13b"| "llama2-70b"| "llama3-8b"|"llama3-70b"|"llama3.1-8b"|"llama3.1-70b"|"llama3.1-405b"| "llama3.3-70b"| "mistral-7b"| "mixtral-8x7b"| "mixtral-8x22b"| "deepseek2-16b"| "deepseek2-236b"| "deepseek3-671b"| "deepseek3-test"| "deepseek3-tiny"| "kimi-k2-1t"| "gemma-7b"| "gemma-2b"| "gemma2-2b"| "gemma2-9b"| "gemma2-27b"| "gemma3-4b"| "gemma3-12b"| "gemma3-27b"| "qwen3-0.6b"| "qwen3-4b"| "qwen3-4b-thinking-2507"| "qwen3-8b"| "qwen3-14b"| "qwen3-32b"| "qwen3-235b-a22b"| "qwen3-30b-a3b"| "qwen3-480b-a35b"| "qwen3-next-80b-a3b"| "gpt3-175b"| "gpt3-22b"| "gpt3-6b"| "gpt3-52k"| "gpt-oss-20b"| "gpt-oss-120b"| "llama4-17b-16e"| "llama4-17b-128e"

class RunInfo(BaseModel):
  """Configuration for the overall run, model identity, and logging."""

  run_name: str = Field(
      "",
      description="The name of the run. Checkpoints will be stored under this name.",
  )
  model_name: ModelName = Field("default", description="The name of the model configuration to use.")
```

bvandermoon: Not fully following this. Does this snippet validate the `model_name` comes from a list of allowed values? Or is this setting some type of default?

SamuelMarks: Pydantic uses [PEP484](https://peps.python.org/pep-0484/) syntax

dependencies/requirements/base_requirements/requirements.txt
```diff
@@ -25,6 +25,7 @@ pathwaysutils
 pillow
 pre-commit
 protobuf
+pydantic
```
bvandermoon, line 28: For this file, you will also want to create the generated-requirements with `seed-env`. Guide in progress: https://github.com/AI-Hypercomputer/maxtext/pull/2536

Also since we are not fully migrated to this setup yet, we should add this to the existing `requirements.txt` and `requirements_with_jax_ai_image.txt`

SamuelMarks: It is technically already in those files as a second-order dependency. It's just that now it's a first-order also so should be added.

bvandermoon: Thanks Samuel. So then the generated requirements should be updated with the CLI tool to reflect this change, right?

Also, I think it should be included in `requirements.txt` and `requirements_with_jax_ai_image.txt`

SamuelMarks: `pydantic>=2.12.3` or `pydantic>=2.12.0` is in both `generated_requirements/cuda12-requirements.txt` and `generated_requirements/tpu-requirements.txt`

bvandermoon: Can you add `pydantic` to this file also? https://github.com/AI-Hypercomputer/maxtext/blob/main/requirements.txt

bvandermoon (commented):
Generally LGTM, please also add the additional benchmark runner tests you mentioned offline in the PR description

dependencies/requirements/base_requirements/requirements.txt
SamuelMarks: Done

bvandermoon (commented):
Thanks @SamuelMarks for the offline testing. Discussed offline, it looks like the `base.yml` configs are being picked up, but the [model-specific configs](https://github.com/AI-Hypercomputer/maxtext/blob/d29bb51868e0ecf4964297ee06c3e0607e645281/src/MaxText/configs/models/llama3.1-70b.yml) are not. Temporarily removing the `Pull Ready` tag so we can sort this out

bvandermoon (approved):

SamuelMarks:
This successfully made it into the `main` branch in 81d653c93f6868984090e5959a8e26c972d13fab (via internal `copybara` processes; identical Python code to this PR)

PR #1837, opened 2025-06-16
Integrate Multi-Token Prediction (MTP) Training objective

parambole:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist


MaxText/layers/models.py
```diff
@@ -59,14 +62,25 @@ def setup(self):
...
-
     self.vision_encoder = VisionEncoder(config=cfg, mesh=mesh) if cfg.use_multimodal else None
     self.decoder = Decoder(config=cfg, shared_embedding=self.shared_embedding, mesh=mesh, quant=self.quant)
+    # If MTP is enabled via config, set up the MTP block.
+    if self.config.mtp_num_layers > 0:
+      # Get the list of layer blueprints for the current model.
+      layer_types = maxtext_utils.get_decoder_layers(self.config)
+      # For MTP, we use the primary (usually dense) transformer block blueprint
+      # to ensure architectural consistency. By convention, this is the first in the list.
+      mtp_layer = layer_types[0]
```
RissyRan: For DeepSeek, it's mixed layers (`[deepseek.DeepSeekDenseLayer, deepseek.DeepSeekMoELayer]`. Could you confirm if this is dense or moe layer? It is a moe layer if I recall correctly.

MaxText/layers/models.py
```diff
@@ -111,4 +125,35 @@ def __call__(
...
+
+    # The Multi-Token Prediction (MTP) block functions as a "side-car" to the main
+    # model, active only during training. It computes an auxiliary loss based on
+    # predicting multiple future tokens, as described in the DeepSeek-V3 paper.
+    # To ensure architectural consistency, it uses two key components from the parent Transformer:
+    #   1. The same `DecoderLayer` blueprint for its internal transformer blocks.
+    #   2. The `shared_embedding` for both embedding future tokens and for its final
+    #      logit projection.
+    # Its only effect is to "sow" these losses; it does not alter the primary logits output.
+    if self.config.mtp_num_layers > 0 and model_mode == MODEL_MODE_TRAIN:
```
RissyRan: Shall we add assertion in pyconfig.py that mtp_num_layers>1 for inference/serving is not supported?

MaxText/layers/models.py
gagika: should we change it to layer_types[-1]?

Also wondering what will be a better API to handle mutliple type of decoder layer support

MaxText/train.py
```diff
@@ -403,6 +409,26 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
...
+    weights_path = ("mtp_losses", "mtp_block", "weights")
+
+    mtp_losses = maxtext_utils.get_nested_value(intermediate_outputs, losses_path, default=())
+    mtp_weights = maxtext_utils.get_nested_value(intermediate_outputs, weights_path, default=())
+    if mtp_losses:  # Ensure MTP heads ran
+      sum_of_all_mtp_losses = jnp.sum(jnp.array(mtp_losses))
+      sum_of_all_mtp_weights = jnp.sum(jnp.array(mtp_weights))
+
+      avg_mtp_loss = sum_of_all_mtp_losses / (sum_of_all_mtp_weights + EPS)
+      scaled_mtp_loss = avg_mtp_loss * config.mtp_loss_scaling_factor
```
gagika: Could you extract this into a helper function, e.g.

`calculate_mtp_loss(intermediate_outputs, config)`

gagika (commented):
thanks, a few comments

MaxText/layers/models.py
parambole: > For DeepSeek, it's mixed layers ([deepseek.DeepSeekDenseLayer, deepseek.DeepSeekMoELayer]. Could you confirm if this is dense or moe layer? It is a moe layer if I recall correctly.

Hey @RissyRan you are right. I double checked with the deepseek weights description and it's MoE Layer I have made the changes. 

> should we change it to layer_types[-1]?

Hey @gagika - I have made the necessary changes.

MaxText/train.py
parambole: Hey @gagika - I have made the necessary changes.

MaxText/layers/models.py
parambole: Hey @RissyRan - Thank you for the suggestion. I have added this check.

parambole:
> Also wondering what will be a better API to handle mutliple type of decoder layer support

Hey @gagika - Over here are you suggestion that different depths of the MTP would have different decoder types ?

MaxText/layers/models.py
```diff
@@ -765,14 +63,25 @@ def setup(self):
...
     )
-
     self.vision_encoder = VisionEncoder(config=cfg, mesh=mesh) if cfg.use_multimodal else None
     self.decoder = Decoder(config=cfg, shared_embedding=self.shared_embedding, mesh=mesh, quant=self.quant)
+    # If MTP is enabled via config, set up the MTP block.
+    if self.config.mtp_num_layers > 0:
+      # Get the list of layer blueprints for the current model.
+      layer_types = maxtext_utils.get_decoder_layers(self.config)
+      # For MTP, we use the primary (usually dense) transformer block blueprint
+      # to ensure architectural consistency. By convention, this is the first in the list.
```
gagika: Does this comment need to be updated? seams not consistent with DeepSeek or the code

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +141,84 @@ def __call__(
...
     return next_hidden_state
+
+
+class MultiTokenPredictionBlock(nn.Module):
+  """Orchestrates the MTP process by running a sequence of MTP layers."""
+
+  config: Config
+  mesh: Mesh
+  transformer_layer_module: Type[DecoderLayer]
+  decoder: Type[Decoder]
```
gagika: should it be `decoder: Decoder` instead?
looks like you are passing decoder instant and not a class.

parambole:
> Overall LGTM! But I think we could load DeepSeek v3 ckpt before we claim the support.

Hey @RissyRan I have updated the conversion script to load the MTP checkpoints and I have been sucessfully able to load them and trigger a training run. However the expected bheaviour of the MTP loss after loading the checkpoints is not well documented. 

For now I have requested it in a couple of places and referenced this bug. But in general I have seen that the gradients are flowing as expected as both the MTP loss and the main model loss decrease and eventually converge and training progresses.

MaxText/convert_deepseek_ckpt.py
```diff
@@ -429,7 +477,6 @@ def _convert_huggingface_to_jax_weights(base_model_path, model_params, mem_info)
         moe["shared_experts"]["wi_1"]["kernel"][layer_idx, ...] = shared_wi_1
         moe["shared_experts"]["wo"]["kernel"][layer_idx, ...] = shared_wo
 
-      # re-order
```
gobbleturk: Why delete this?

MaxText/layers/decoders.py
```diff
@@ -232,6 +249,42 @@ def setup(self):
           config=self.config, mesh=self.mesh, layers=pipeline_stage_module, remat_policy=remat_policy
       )
 
+  def get_decoder_layers(self):
+    """Get decoder layers, one of `DecoderBlockType` discriminants or a direct `nn.Module` inheritor."""
```
gobbleturk: is discriminants here the right word? I googled it and don't see a relevant definition lol

MaxText/layers/decoders.py
```diff
@@ -232,6 +249,42 @@ def setup(self):
...
+    # This function now uses pre-imported modules for each case.
```
gobbleturk: IMO remove this comment

MaxText/layers/decoders.py
```diff
@@ -232,6 +249,42 @@ def setup(self):
...
+    match self.config.decoder_block:
+      case DecoderBlockType.DEFAULT:
+        return [DecoderLayer]  # Assumed to be defined in the current scope
```
gobbleturk: Remove comment? Or change to "The default decoder layer defined in this file" (current scope = current file?)

MaxText/layers/decoders.py
```diff
@@ -232,6 +249,42 @@ def setup(self):
...
+      case DecoderBlockType.GPT3:
+        return [gpt3.Gpt3DecoderLayer]
+      case DecoderBlockType.QWEN3:
+        return [qwen3.Qwen3DecoderLayer]
+      case DecoderBlockType.SIMPLE:
+        return [simple_layer.SimpleDecoderLayer]
+      case DecoderBlockType.SIMPLE_MLP:
+        return [simple_layer.SimpleMlpDecoderLayer]
+      case DecoderBlockType.LLAMA4:
+        # Use a conditional expression for conciseness.
```
gobbleturk: nit: remove comment

MaxText/layers/models.py
```diff
@@ -61,14 +62,25 @@ def setup(self):
...
-
     self.vision_encoder = VisionEncoder(config=cfg, mesh=mesh) if cfg.use_multimodal else None
     self.decoder = Decoder(config=cfg, shared_embedding=self.shared_embedding, mesh=mesh, quant=self.quant)
+    # If MTP is enabled via config, set up the MTP block.
+    if self.config.mtp_num_layers > 0:
+      # Get the list of layer blueprints for the current model.
+      layer_types = self.decoder.get_decoder_layers()
+      # For MTP, we use the primary (usually dense) transformer block blueprint
+      # to ensure architectural consistency. By convention, this is the first in the list.
+      mtp_layer = layer_types[-1]
```
gobbleturk, line 73: this confuses me - the previous comment says "first in the list" and the next line of the code grabs index[-1] - why are we not grabbing index[0]?

MaxText/maxtext_utils.py
```diff
@@ -1027,6 +1030,27 @@ def schedule(step):
   return optax.join_schedules(pieces, boundaries)
 
 
+def roll_and_mask(x: jnp.ndarray, shift: int = -1) -> jnp.ndarray:
```
gobbleturk: I doubt anyone will use this method except for MTP, in which case it makes more sense to put this in multi_token_prediction.py instead of maxtext_utils.

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +141,84 @@ def __call__(
...
+      )
+
+      # Project to logits using the shared embedding transpose
+      mtp_logits = self.decoder._apply_output_head(next_mtp_hidden_state, deterministic, model_mode)
+
+      # Calculate cross-entropy loss for this specific layer's prediction
+      mtp_xent, _ = max_utils.cross_entropy_with_logits(mtp_logits, jax.nn.one_hot(rolled_target_ids, cfg.vocab_size), 0.0)
+      mtp_xent_masked = mtp_xent * rolled_target_mask
+
+      # This logic doesn't run during model initialization.
```
gobbleturk: would remove comment, its not informative. Either make it more informative (why don't we want this to run during initialization) or remove

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +141,84 @@ def __call__(
...
+      if not self.is_initializing():
+        # For evaluation, save the top prediction and a valid token mask.
+        # This is only active for the target layer during an eval run.
+        if cfg.mtp_eval_target_layer == k and self.is_mutable_collection("mtp_acceptance"):
+          mtp_top_1_pred = jnp.argmax(mtp_logits, axis=-1)
+          self.sow("mtp_acceptance", "mtp_preds", mtp_top_1_pred)
+          self.sow("mtp_acceptance", "mtp_mask", rolled_target_mask)
+
+        # For training, save the loss components for this MTP head.
+        # This is only active during a training run.
```
gobbleturk, line 248: Where is the logic that implements this? How does this code differentiate between training and inference?

MaxText/maxtext_utils.py
```diff
@@ -911,6 +912,8 @@ def create_device_mesh(config, devices=None):
   """Creates a device mesh with each slice in its own data parallel group. If there is only one slice, uses two replicas"""
   if devices is None:
     devices = jax.devices()
+  if config.subslice_shape and config.enable_single_controller and config.num_slices == 1:
```
gobbleturk: Is this some merge artifact, why is this part of this PR?

MaxText/maxtext_utils.py
```diff
@@ -1091,6 +1115,53 @@ def get_formatted_sharding_annotations(params, mesh=None):
   return "\n".join(annotation_lines)
 
 
+def calculate_mtp_loss(intermediate_outputs, config):
```
gobbleturk: Generally we want methods to be defined as close as possible to where the are called (in this case train.py). However we also want to keep train.py as lean as possible and this might be used by other trainers (e.g. sft?) so consider moving this to train_utils (here is also fine though)

gobbleturk: Moving this to multi_token_prediction.py also makes sense (everything related to mtp in one file), at least I prefer this method in multi_token_prediction to maxtext_utils

MaxText/maxtext_utils.py
```diff
@@ -1091,6 +1115,53 @@ def get_formatted_sharding_annotations(params, mesh=None):
...
+
+  sum_of_all_mtp_losses = jnp.sum(jnp.array(mtp_losses))
+  sum_of_all_mtp_weights = jnp.sum(jnp.array(mtp_weights))
+
+  avg_mtp_loss = sum_of_all_mtp_losses / (sum_of_all_mtp_weights + EPS)
+  scaled_mtp_loss = avg_mtp_loss * config.mtp_loss_scaling_factor
+  return scaled_mtp_loss
+
+
+def calculate_mtp_acceptance_rate(intermediate_outputs, config):
```
gobbleturk: Similar to above I prefer this in multi_token_prediction.py

MaxText/pyconfig.py
```diff
@@ -163,6 +163,12 @@ def validate_keys(keys):
   validate_prefill_and_target_lengths(keys["max_prefill_predict_length"], keys["max_target_length"])
   validate_rope_type(keys["rope_type"])
 
+  if keys["mtp_num_layers"] > 0 and keys["model_call_mode"] == "inference":
```
gobbleturk, line 169: ah I found where the only training logic was implemented =D

MaxText/tests/maxtext_utils_test.py
gobbleturk, line 1: I recommended moving some methods from maxtext_utils to multi_token_prediction.py, in which case some of these tests can also be moved to multi_token_prediction_test.py

MaxText/tests/multi_token_prediction_test.py
```diff
@@ -112,5 +116,160 @@ def test_multi_token_prediction_layer_output(self):
...
+        name="shared_embedding",
+    )
+    self.decoder = Decoder(
+        config=self.config, mesh=self.mesh, shared_embedding=self.shared_embedding, name="decoder_for_mtp"
+    )
+    self.mtp_block = multi_token_prediction.MultiTokenPredictionBlock(
+        config=self.config,
+        mesh=self.mesh,
+        name="mtp_block",
+        transformer_layer_module=DecoderLayer,
```
gobbleturk, line 144: you can also consider using SimpleDecoderLayer in place of DecoderLayer (its simpler so slightly faster to test and debug, just a single matmul instead of attention + FF)

MaxText/tests/multi_token_prediction_test.py
```diff
@@ -112,5 +116,160 @@ def test_multi_token_prediction_layer_output(self):
...
+        self.main_hidden_state,
+        self.input_ids,
+        self.target_ids,
+        self.target_mask,
+        self.position_ids,
+        self.decoder_segment_ids,
+        deterministic=True,
+    )
+
+  def test_sow_functionality(self):
```
gobbleturk, line 199: These are great, thank you!

MaxText/convert_deepseek_ckpt.py
```diff
@@ -69,46 +71,26 @@
...
-      return False
-  return True
+  return key not in banned_keys
 
 
-def hf_to_maxtext_mapping(layer_idx, num_experts, first_num_dense_layers) -> dict:
+def hf_to_maxtext_mapping(layer_idx, num_experts, first_num_dense_layers, num_main_layers) -> dict:
   """
   Generates a mapping between Hugging Face (HF) and MaxText model weight names.
   HF MLP is using self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x)).
```
RissyRan, line 93: Why remove the args and returns in doc string?

MaxText/layers/models.py
RissyRan: Yeah, should be updated to use last in the list for DS v3, it's MoE layer. Or, we also could put it in general by passing a config here if we'd like to generalize it.

MaxText/tests/multi_token_prediction_test.py
```diff
@@ -112,5 +116,160 @@ def test_multi_token_prediction_layer_output(self):
...
+      avg_mtp_loss = sum_of_all_mtp_losses / (sum_of_all_mtp_weights + 1e-8)
+      scaled_mtp_loss = avg_mtp_loss * self.cfg.mtp_loss_scaling_factor
+
+      final_loss_for_gradient += scaled_mtp_loss
+      mtp_loss_for_logging = scaled_mtp_loss
+    # -------------------------------------------------------------
+
+    # 5. Assert that the final values are correct.
+    # The final loss should have increased from its base value.
+    self.assertGreater(final_loss_for_gradient, 100.0)
```
RissyRan, line 271: Why the loss with MTP should be larger than base value?  because of `main_loss + mtp_loss_scaling_factor * avg_mtp_loss`? Then it seems you only need to check `avg_mtp_loss` > 0 with valid `mtp_loss_scaling_factor`?

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +143,131 @@ def __call__(
...
+  sum_of_all_mtp_weights = jnp.sum(jnp.array(mtp_weights))
+
+  avg_mtp_loss = sum_of_all_mtp_losses / (sum_of_all_mtp_weights + EPS)
+  scaled_mtp_loss = avg_mtp_loss * config.mtp_loss_scaling_factor
+  return scaled_mtp_loss
+
+
+def calculate_mtp_acceptance_rate(intermediate_outputs, config):
+  """Calculates the MTP acceptance rate from intermediate outputs."""
+  if config.mtp_eval_target_layer <= 0:
```
RissyRan: Nit: should we add an assertion that eval_target_layer should be valid? i.e. >=0 in the pyconfig.py file

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +143,131 @@ def __call__(
...
+    return 0.0
+
+  sown_data = maxtext_utils.get_nested_value(intermediate_outputs, ("mtp_acceptance", "mtp_block"), {})
+  mtp_preds = maxtext_utils.get_nested_value(sown_data, ("mtp_preds",), [None])[0]
+  valid_mask = maxtext_utils.get_nested_value(sown_data, ("mtp_mask",), [None])[0]
+
+  if mtp_preds is None or valid_mask is None:
```
RissyRan, line 286: Could you make a comment which case those could be None? Just trying to avoid we silently skip some unexpected issues.

MaxText/layers/multi_token_prediction.py
```diff
@@ -138,3 +143,131 @@ def __call__(
...
+
+  # Get the main model's greedy predictions from the logits.
+  main_model_preds = jnp.argmax(intermediate_outputs["logits"], axis=-1)
+
+  # Roll the main model's predictions to align them in time with the MTP head's target.
+  rolled_main_preds = main_model_preds
+  for _ in range(config.mtp_eval_target_layer):
+    rolled_main_preds = maxtext_utils.roll_and_mask(rolled_main_preds)
+
+  # Compare the aligned predictions, applying the valid_mask to ignore junk values.
```
RissyRan: Could you explain a little bit on junk values? Thanks!

MaxText/layers/decoders.py
parambole: It was actually part of the original code https://github.com/AI-Hypercomputer/maxtext/pull/1852/files#diff-a2c720f20f263b267aac73ff3c660977394e349a1f3d9c22bec15cc7233c69faR338 not sue what it truly means :)

But I will look into rephrasing it.

MaxText/layers/decoders.py
parambole: Done.

MaxText/layers/decoders.py
parambole: Done

MaxText/layers/decoders.py
parambole: Done.

MaxText/layers/models.py
parambole: > this confuses me - the previous comment says "first in the list" and the next line of the code grabs index[-1] - why are we not grabbing index[0]?

Yes Gagik had the same comment. It had resolved it but looks like the code change didn't make it to the PR. I have updated it now.

MaxText/maxtext_utils.py
parambole: Makes sense. Moved it to multi_token_prediction

MaxText/layers/multi_token_prediction.py
parambole: Updated with "This logic doesn't run during model initialization to avoid unwated population of the mutable collections."

Thanks!

MaxText/layers/multi_token_prediction.py
parambole: The logic is in `train.py`. Also over here there are two modes training and evaluation.

MaxText/maxtext_utils.py
parambole: You are right. This looks like part of a merge. Let me remove it.

MaxText/maxtext_utils.py
parambole: Ack. I have moved this to `multi_token_prediction.py`

MaxText/tests/maxtext_utils_test.py
parambole: Done.

MaxText/maxtext_utils.py
parambole: Ack. Done

MaxText/tests/multi_token_prediction_test.py
parambole: Makes Sense. Done.

MaxText/tests/multi_token_prediction_test.py
parambole: Thank you :)

MaxText/convert_deepseek_ckpt.py
parambole: > Why remove the args and returns in doc string?

I have only added an additional arg to the function signature. Apart from that everything else remains the same.

MaxText/tests/multi_token_prediction_test.py
parambole: Hey @RissyRan yes and that is being checked by - `self.assertGreater(mtp_loss_for_logging, 0.0)`.

But since we are adding it to the main model I added an additional check to confirm that that flow is working as well through `self.assertGreater(final_loss_for_gradient, 100.0)`.

MaxText/layers/multi_token_prediction.py
parambole: Thanks for pointing this out. I have added more details. PTAL.

MaxText/layers/multi_token_prediction.py
parambole: Thanks for the feedback! That's a great point—"junk values" was too vague. I've updated the comment in the code to be more precise.

To clarify, the "junk values" are the placeholder zeros introduced at the end of a sequence by the roll_and_mask function. The mask is used to ignore them during calculations.

For example:

If an original_sequence is [10, 20, 30, 40]

And its original_mask is [1, 1, 1, 1]

After a roll_and_mask(shift=-1):

The rolled_sequence becomes [20, 30, 40, 0]  (the 0 is the placeholder value)

The rolled_mask becomes [1, 1, 1, 0]

When we multiply by the rolled_mask, we only consider the valid tokens and correctly ignore the comparison at the placeholder 0 at the end.

parambole:
> Shall we add general MTP support into [announcement](https://github.com/AI-Hypercomputer/maxtext?tab=readme-ov-file#announcements)?

That is a great suggestion. I have updated the announcement section.

MaxText/layers/multi_token_prediction.py
parambole: Makes sense. Done

README.md
```diff
@@ -31,6 +31,7 @@ Key supported features:
 * Models: Llama 2, Llama 3, Llama 4, Mistral and Mixtral family, Gemma, Gemma 2, Gemma 3, and DeepSeek family
 
 ## Announcements
+* [July, 2025] Multi-Token Prediction (MTP) training is now supported! This feature adds an auxiliary loss based on predicting multiple future tokens, inspired by the [DeepSeek-V3 paper](https://arxiv.org/html/2412.19437v1), to enhance training efficiency.
```
RissyRan: Nit: ensure to add specific date before the merge.

MaxText/layers/multi_token_prediction.py
RissyRan: Thanks for the explanation! Very helpful!

RissyRan (approved):
Thank you! LGTM

PR #1852, opened 2025-06-19
Refactor: Decouple Core Transformer Blocks

parambole:
**TL;DR**

-   **What:** This PR refactors the core `DecoderLayer` and its related components out of `layers/models.py` and into a new, foundational file: `layers/decoders.py` & `layers/encoders.py`.

-   **Why:** To improve the overall code architecture and break potential circular dependencies. This is a necessary prerequisite for adding new, complex modules that also need access to these core building blocks.

-   **How:** By creating `layers/decoders.py` & `layers/encoders.py` to house the fundamental `Decoder` and `DecoderLayer` classes. Higher-level files like `models.py` and other future modules now import these components from a single location.

### **Detailed Description**

This pull request introduces a structural refactoring to improve modularity and maintainability.

The primary change is the creation of `MaxText/layers/blocks.py`, which now serves as the source for fundamental building blocks of the Transformer architecture, such as:

-   `DecoderLayer`

-   `Decoder`

Previously, these classes were located in `layers/models.py`, which created tight coupling. As we add more features this tight coupling would lead to circular import dependencies.

By decoupling these core components, we establish a clearer hierarchy in the codebase, where high-level modules can depend on these "building blocks" without depending on each other.

# Checklist


MaxText/layers/blocks.py
```diff
@@ -0,0 +1,702 @@
...
+from MaxText.layers import quantizations
+from MaxText.layers.attentions import Attention
+from MaxText.layers.embeddings import Embed, PositionalEmbedding
+from MaxText.layers.normalizations import RMSNorm
+
+# Type alias for cleaner type hints
+Quant = quantizations.AqtQuantization
+
+
+class DecoderLayer(nn.Module):
```
bvandermoon: Is there a more descriptive file name that can be used here instead of 'blockes.py'? Maybe naming this file `decoders.py`? Would need to create a separate `encoders.py` file for the `VisionEncoder`

RissyRan: I agree to name this as `decoders.py`, and move `class VisionEncoder` to `encoders.py`.

RissyRan (commented):
Could you select one model to run a quick perf test with same configs? main branch vs. your branch. I see functional tests are passing, just ensure some breakdown does not hurt performance.

MaxText/layers/blocks.py
parambole: Hey @bvandermoon & @RissyRan thank you for the suggestion. I have renamed the files. PTAL.

parambole:
> Could you select one model to run a quick perf test with same configs? main branch vs. your branch. I see functional tests are passing, just ensure some breakdown does not hurt performance.

Hey @RissyRan - I performed `test_convergence_1b_params` with and without the changes and got comparable results.

parambole:
> Thanks for refactoring @parambole. Can you explain the circular dependency you mentioned in the description?

In the case the `multi_token_prediction.py` used parts of `models.py` ( like decoder ) but in turn models.py used `multi_token_predicition.py` ( which caused the cyclic dependency ) 

In general certain components of `models.py` could ( IMO should be decoupled and hence I split them into `decoders.py` and `encoders.py` )

parambole:
> Thanks @parambole. I think there is some overlap with @richjames0's PR here: #1852. Not sure if you were already aware of that, but the two of you can discuss if not

Hey @bvandermoon - thanks for pointing this out. I have synced with @richjames0 and incorporated the overlapping changes.

MaxText/maxtext_utils.py
```diff
@@ -1095,3 +1103,39 @@ def get_formatted_sharding_annotations(params, mesh=None):
...
+    case DecoderBlockType.QWEN3:
+      return [qwen3.Qwen3DecoderLayer]
+    case DecoderBlockType.SIMPLE:
+      return [simple_layer.SimpleDecoderLayer]
+    case DecoderBlockType.SIMPLE_MLP:
+      return [simple_layer.SimpleMlpDecoderLayer]
+    case DecoderBlockType.LLAMA4:
+      return [llama4.Llama4ScannableBlock] if config.scan_layers else [llama4.Llama4DecoderLayer]
+    case _:
+      raise ValueError(f"Incorrect decoder_block name {config.decoder_block.value=}")
```
bvandermoon: Not sure about this in maxtext_utils. Can this go somewhere else, maybe decoders.py?

MaxText/layers/models.py
```diff
@@ -16,731 +16,26 @@
...
 from MaxText.inference import page_manager
-from MaxText.layers import linears
-from MaxText.layers import quantizations
-from MaxText.layers import pipeline
-from MaxText import maxtext_utils
 from MaxText import multimodal_utils
-from MaxText.layers.attentions import Attention
-from MaxText.layers.normalizations import rms_norm
-from MaxText.layers.embeddings import positional_embedding_as_linen, Embed
+from MaxText.layers.decoders import Decoder
```
RissyRan, line 29: Shall we just import module instead of class? Also mentioned in python guide, [FYI](https://screenshot.googleplex.com/9DMo3p58g4e4g67). Also, could you double check other imports?

However, I see this import is widely used `from MaxText.common_types import ....` in the repo.

RissyRan (approved):
Overall LGTM! Just one comment about import.

MaxText/maxtext_utils.py
parambole: Hey @bvandermoon - I added it to maxtext_utils because it's being used in multiple places, such as Decoder and MTP. My reasoning was that this approach would prevent future dependency overhead. For instance, if I only need to extract DecoderLayer type, I wouldn't have to rely on classes like Decoder or even the entire file; I could simply retrieve it from a utilities file. This avoids potential issues that would arise if it were used within the Decoders file itself and in another file/class, as a neutral location prevents such conflicts.

I'm open to placing it elsewhere if you think that's a better solution. Please let me know.

bvandermoon: Could you make a new file for this in the layers directory (and also add a subdirectory like `mappings` if it makes sense)? This seems really tied to the layers directory. Thoughts?

MaxText/layers/deepseek.py
```diff
@@ -30,7 +30,8 @@
 from MaxText.layers import attentions
 from MaxText.layers import initializers
 from MaxText.layers import linears
-from MaxText.layers import models
+from MaxText.common_types import Config
```
bvandermoon, line 33: Just to confirm my understanding - this was previously coming from here? [models.py](https://github.com/AI-Hypercomputer/maxtext/blob/93930f692992519bee2bcc1f4cd18fe64e095592/MaxText/layers/models.py#L30)

MaxText/layers/decoders.py
```diff
@@ -0,0 +1,711 @@
...
+    if self.config.decoder_block in (
+        DecoderBlockType.DEFAULT,
+        DecoderBlockType.LLAMA2,
+        DecoderBlockType.MISTRAL,
+        DecoderBlockType.MIXTRAL,
+        DecoderBlockType.DEEPSEEK,
+        DecoderBlockType.GEMMA,
+        DecoderBlockType.GEMMA2,
+        DecoderBlockType.GEMMA3,
+        DecoderBlockType.SIMPLE,
```
bvandermoon: Missing the recent Qwen3 addition here. Was this removed in some type of rebase conflict? I see Qwen3 is present in other places in this file so it seems like the rebase already happened

MaxText/layers/decoders.py
```diff
@@ -0,0 +1,711 @@
...
+  ):
+    """Applies token and positional embeddings to the input tokens."""
+
+    cfg = self.config
+    # [batch, length] -> [batch, length, emb_dim]
+    y = self.shared_embedding(decoder_input_tokens.astype("int32"))
+
+    # Merge the image embeddings with the text embeddings for multimodal models
+    if image_embeddings is not None and cfg.use_multimodal:
+      if cfg.model_name in ["gemma3-4b", "gemma3-12b", "gemma3-27b"]:
```
bvandermoon: What about `llama4-17b-16e` and `llama4-17b-128e`?

MaxText/layers/decoders.py
```diff
@@ -0,0 +1,711 @@
...
+    if cfg.using_pipeline_parallelism:
+      if cfg.pipeline_fsdp_ag_once:
+        partition_spec = self.pipeline_module.get_weight_sharding(
+            y, decoder_segment_ids, decoder_positions, deterministic, model_mode
+        )
+      else:
+        partition_spec = None  # This partition spec is only used for the fsdp_ag_once feature.
+      if cfg.decoder_block == DecoderBlockType.DEEPSEEK:
+        assert len(RemattedBlockLayers) == 2, "Scanned layers must have a length of 2 using deepseek."
+        dense_layer = RemattedBlockLayers[0]
```
bvandermoon, line 540: Are these not needed before/after this anymore?

```
layer_call_kwargs = {
    "page_state": page_state,
    "previous_chunk": previous_chunk,
    "slot": slot,
}
dense_layer = RemattedBlockLayers[0]
dense_layer.__call__ = functools.partial(dense_layer.__call__, **layer_call_kwargs)
```

Also `moe_layer.__call__ = functools.partial(moe_layer.__call__, **layer_call_kwargs)`

bvandermoon (commented):
Thank you @parambole. This looks good. Just a few things I wanted to check in on that I noticed were different in the diff. Especially the Qwen3 comment

MaxText/layers/decoders.py
parambole: Hey @bvandermoon - I have made this identical to the original file and will make any MTP related refactor in the next PR.

MaxText/layers/decoders.py
parambole: Hey @bvandermoon - I have made this identical to the original file and will make any MTP related refactor in the next PR.

bvandermoon (approved):
LGTM assuming https://github.com/AI-Hypercomputer/maxtext/pull/1947 fixes the unit test issue

PR #1865, opened 2025-06-23
All Gather Once (FSDP Zero-one)

wei879-100:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -186,6 +186,9 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
+model_fsdp_ag_once: True # This controls whether the Zero-1 optimization is active.
+# This is a memory/time tradeoff - True: Use ZeroOneTransformer to gather weights once per gradient step. 
+# False: Use the standard Transformer, which gathers for each microbatch's fwd/bwd pass.
 pipeline_fsdp_ag_once: False # If set to true then all gather all of the weights over FSDP before the first pipeline iteration.
 # This is a memory/time tradeoff - we now have to store the FSDP gathered weights and gradients (typically in bf16), as opposed
 # to only one stage's worth, however we only execute one all-gather and reduce across per repeat, as opposed
--- a/MaxText/layers/models.py
+++ b/MaxText/layers/models.py
@@ -804,4 +804,175 @@ class Transformer(nn.Module):
         image_embeddings=image_embeddings,
     )
     return logits
+
+class ZeroOneTransformer(nn.Module):
+  """
+  A wrapper for the base Transformer model designed to implement the Zero-1
+  FSDP optimization.
+
+  The goal of this optimization is to reduce communication overhead. In the standard
+  FSDP implementation, an all-gather operation on the model weights is performed twice
+  for each microbatch (once for the forward pass, once for the backward pass).
+
+  This class changes that behavior. When enabled, it performs the all-gather operation
+  only *once* per full gradient accumulation step. It gathers the full weights into
+  memory, runs all the microbatch forward and backward passes, and then releases the
+  full weights. This trades higher peak memory usage for significantly reduced
+  network communication, which can improve training speed if sufficient memory is
+  available.
+  """
+  config: Config
+  mesh: Mesh
+  quant: Quant
+
+  def get_weight_sharding(self, *init_args):
+    """
+    Initializes the model to inspect and return the logical partition specs
+    for all of its weights. This is a helper function to understand the intended
+    sharding layout of the model as defined by the user.
+
+    Args:
+      *init_args: The arguments required to initialize the model (e.g., input shapes).
+
+    Returns:
+      A PyTree representing the logical PartitionSpec for the model's parameters.
+    """
+    key = jax.random.PRNGKey(0)
+    # Create dummy keys for initialization
+    keys = {"params": key, "dropout": key, "aqt": key}
+    # Temporarily initialize the model to get its complete variable structure.
+    weights = self.init(keys, *init_args)
+
+    def get_partition_spec(pytree):
+      """Inner function to traverse the PyTree of weights."""
+      # This function is designed to find all leaves that are LogicallyPartitioned
+      # and extract their partition spec.
+      raise Exception("Please check the model definition") 
   
+      def _is_leaf(x):
+        return ininstance(x,nn.LogicallyPartitioned)
+      
+      def get_partition_spec_leaf(leaf):
+        return leaf.get_partition_spec()
+      
+      partition_spec_tree = jax.tree.map(get_partition_spec_leaf, pytree, is_leaf=_is_leaf)
+      return partition_spec_tree
+    # Run the function to get the full tree of partition specs.
+    partition_spec_with_extra_layer = get_partition_spec(weights)
+    # The actual model parameters are nested under the 'layers' key. This extracts them.
+    partition_spec = {"params": partition_spec_with_extra_layer["params"]["layers"]}
+    return partition_spec
+  
+  def get_physical_spec_no_fsdp(self, full_logical):
+    """
+    Converts a logical partition spec into a physical sharding spec that
+    describes a fully replicated (unsharded) layout for the weights.
+
+    It does this by taking the original sharding rules and removing any axis
+    sharded over 'fsdp' or 'fsdp_transpose'. Replacing a sharding axis with
+    None tells JAX to replicate the data along that physical mesh dimension.
+
+    Args:
+      full_logical: The original logical partition specs for all weights.
+
+    Returns:
+      A PyTree of physical NamedSharding objects configured to represent
+      fully-gathered weights (replicated across the 'fsdp' axis).
+    """
+
+    def remove_fsdp_sharding(sharding_tree):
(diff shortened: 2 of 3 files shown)


MaxText/configs/base.yml
```diff
@@ -186,6 +186,9 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
+model_fsdp_ag_once: False # This controls whether the Zero-1 optimization is active.
```
gobbleturk, line 189: how does this interplay with pipeline_fsdp_ag_once? If you set the model one to true, what happens when we set pipeline_fsdp_ag_once to true/false?

MaxText/layers/models.py
```diff
@@ -804,4 +804,175 @@ def __call__(
...
\ No newline at end of file
+
+class ZeroOneTransformer(nn.Module):
+  """
+  A wrapper for the base Transformer model designed to implement the Zero-1
+  FSDP optimization.
+
+  The goal of this optimization is to reduce communication overhead. In the standard
+  FSDP implementation, an all-gather operation on the model weights is performed twice
+  for each microbatch (once for the forward pass, once for the backward pass).
```
gobbleturk: microbatch -> gradient accumulation microbatch

MaxText/configs/base.yml
```diff
@@ -186,6 +186,9 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
...
+# This is a memory/time tradeoff - True: Use ZeroOneTransformer to gather weights once per gradient step.
```
gobbleturk: Add - This is zero one sharding

MaxText/configs/base.yml
```diff
@@ -186,6 +186,9 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
+model_fsdp_ag_once: False # This controls whether the Zero-1 optimization is active.
+# This is a memory/time tradeoff - True: Use ZeroOneTransformer to gather weights once per gradient step. 
+# False: Use the standard Transformer, which gathers for each microbatch's fwd/bwd pass.
```
gobbleturk: Add - this is zero 3 sharding

MaxText/layers/models.py
```diff
@@ -804,4 +804,175 @@ def __call__(
...
+      
+      partition_spec_tree = jax.tree.map(get_partition_spec_leaf, pytree, is_leaf=_is_leaf)
+      return partition_spec_tree
+    # Run the function to get the full tree of partition specs.
+    partition_spec_with_extra_layer = get_partition_spec(weights)
+    # The actual model parameters are nested under the 'layers' key. This extracts them.
+    partition_spec = {"params": partition_spec_with_extra_layer["params"]["layers"]}
+    return partition_spec
+  
+  def get_physical_spec_no_fsdp(self, full_logical):
```
gobbleturk: Can this implementation be shared with the PP version as well (https://github.com/AI-Hypercomputer/maxtext/blob/f52c26c17b53b36bf1e741e19550cade98772737/MaxText/layers/pipeline.py#L561)

MaxText/layers/models.py
```diff
@@ -804,4 +804,175 @@ def __call__(
...
+        return named_sharding
+      return jax.tree.map(_remove_fsdp_from_partition_spec, sharding_tree)
+
+    # Convert the high-level logical spec to a physical one using default rules.
+    physical = nn.logical_to_mesh_sharding(full_logical, mesh=self.mesh, rules=self.config.logical_axis_rules)
+    # Apply the function to remove the FSDP sharding, defining our target layout.
+    physical_no_fsdp = remove_fsdp_sharding(physical)
+    return physical_no_fsdp
+  
+  def all_gather_over_fsdp(self,sharding_info):
```
gobbleturk: Can this function also be shared (if so can we define it in maxtext_utils)

MaxText/layers/models.py
gobbleturk: If so lets move it to maxtext_utils and have both PP and this one call the shared one in maxtext_utils

gobbleturk:
Can we also add a unit test (integration test) with this feature on in [train_tests](https://github.com/AI-Hypercomputer/maxtext/blob/f52c26c17b53b36bf1e741e19550cade98772737/MaxText/tests/integration_tests/train_tests.py#L147)

in your PR description you can link the profiles before and after (ideally with pdb=1 to make the feature more obvious)

MaxText/configs/base.yml
wei879-100: Xplane:
model_fsdp_ag_once: True; pipeline_fsdp_ag_once: True:  
https://xprof.corp.google.com/trace_viewer/fanwei-3085710629162492243?hosts=t1v-n-aade02bb-w-0&host_index=0&trace_filter_config={}

model_fsdp_ag_once: False; pipeline_fsdp_ag_once: True:  
https://xprof.corp.google.com/trace_viewer/fanwei-3085710629162494718?hosts=t1v-n-aade02bb-w-0&host_index=0&trace_filter_config={}&view_start=-343.119&view_end=2636.957

both as True takes longer training time. Although the time for the fwd/bwd pass is shorter, there is a little "tail" after the bwd pass, some extra copy did take longer. The bug has been created here: b/427789745

MaxText/configs/base.yml
```diff
@@ -186,7 +186,10 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
-pipeline_fsdp_ag_once: False # If set to true then all gather all of the weights over FSDP before the first pipeline iteration.
+model_fsdp_ag_once: False # This is ZeRO-one sharding that controls whether the Zero-one optimization is active.
+# This is a memory/time tradeoff - True: Use ZeroOneTransformer to gather weights once per gradient step. 
+# False: Use the standard Transformer, which gathers for each microbatch's fwd/bwd pass.
+pipeline_fsdp_ag_once: False # This is ZeRO-three sharding. If set to true then all gather all of the weights over FSDP before the first pipeline iteration.
```
gobbleturk: I wouldn't change this comment

MaxText/layers/models.py
```diff
@@ -804,4 +803,74 @@ def __call__(
         image_embeddings=image_embeddings,
     )
     return logits
-  
\ No newline at end of file
+
+
+class ZeroOneTransformer(nn.Module):  # maybe change ZeroOneWrapper()
```
gobbleturk: We can keep it as ZeroOneTransformer - I would remove the # maybe change ZeroOneWrapper comment

gobbleturk: Especially as implemented now its hardcoded to the Transformer, if this was a generic wrapper it should take the wrapped model (e.g. Transformer) as input

MaxText/tests/integration_tests/train_tests.py
```diff
@@ -334,6 +334,38 @@ def test_gpu_cudnn_flash_jax(self):
...
+    train_main(TrainTests.CONFIGS["base"] + ["model_fsdp_ag_once=True"])
+
+  @pytest.mark.integration_test
+  @pytest.mark.gpu_only
+  def test_gpu_synthetic_model_ag_once(self):
+    train_main(TrainTests.CONFIGS["synthetic"] + ["model_fsdp_ag_once=True"])
+
+  @pytest.mark.integration_test
+  @pytest.mark.gpu_only
+  def test_gpu_context_parallel_ag_once(self):
```
gobbleturk: I would remove this test, I don't see ithe value its adding, you are already adding a GPU test above

MaxText/configs/base.yml
wei879-100: I put the comment to the wrong line previously, should be after "False". Revised

MaxText/train.py
```diff
@@ -602,7 +602,13 @@ def setup_mesh_and_model(config, devices=None):
...
   # Model and Optimizer definition
   quant = quantizations.configure_quantization(config)
-  model = Transformer(config, mesh, quant=quant)
+  # Determine which model class to use based on the config flag.
+  if config.model_fsdp_ag_once:
+    # Instantiate the Zero-1 wrapper model if the optimization is enabled.
+    model = ZeroOneTransformer(config, mesh, quant=quant)
+  else:
+    # Otherwise, use the standard memory-efficient Transformer model.
+    model = Transformer(config, mesh, quant=quant)
```
bvandermoon: Could you move this logic to a separate method (and possibly a separate file - maybe `maxtext_utils` if there isn't a better spot)? Then this call could just be something like `model = get_transformer_model(...)`

wei879-100: Thanks for your suggestion! I put the call on the train_utils.py instead. Please check again. Thanks!

hx89:
Thanks for the PR! 

I was testing this PR on llama3 70b mode and the profile had two unexpected behavior:
1. all-gather in the fprop is in fp32 instead of bf16
2. bprop has all-reduce instead of reduce-scatter

Did you see these behavior in your test?

And could you please share the profile you got into a public repo so that I can compare? The original link is Google internal and I don't have access.

PR #1871, opened 2025-06-24
Update MaxText's requirements.txt

kanglant:
This change shouldn't significantly affect MaxText's existing environment setup, and it is crucial for consistent python environment setups.

When updating MaxText's requirements.txt file, please keep these recommendations in mind:
- For each dependency, we recommend setting a lower bound to ensure the python seed env tool works well
- All source links must be pinned to a specific hash to guarantee reproducibility

We picked the lower bounds based on the generated maxtext requirements lock files in python seed env experiments, e.g., https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/maxtext_requirements_lock_3_10.txt. Please update the lower bounds if you found different versions are needed.

# Checklist


Rohan-Bierneni:
For the maxtext_gpu lock file generated at https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/maxtext_requirements_lock_3_10.txt, I noticed that the package transformer_engine is missing. IIRC this is a hard requirement for maxtext to function. It is currently installed via setup.sh: https://github.com/AI-Hypercomputer/maxtext/blob/79e717f8983dd49ae2c18e9c548f7289a36b5148/setup.sh#L158. 

Can this dep be added directly to https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/constraints_gpu_only.txt and requirements.txt or would it cause any issues?

requirements.txt
```diff
@@ -1,43 +1,46 @@
...
+aqtp>=0.8.4
+cloud-accelerator-diagnostics>=0.1.1
+cloud-tpu-diagnostics>=0.1.5
+datasets>=3.6.0
+gcsfs>=2024.5.0
+google-cloud-aiplatform>=1.61.0
+google-cloud-storage>=2.19.0
+google-cloud-monitoring>=2.27.2
+google-api-core>=2.25.1
+google-api-python-client>=2.172.0
```
bvandermoon, line 16: Just for my understanding - why is it necessary to specify the minimum version for every dependency like this? Just wondering if this will be harder to maintain. Wouldn't the upcoming uv-generated lock file still work even without minimum versions?

kanglant:
> For the maxtext_gpu lock file generated at https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/maxtext_requirements_lock_3_10.txt, I noticed that the package transformer_engine is missing. IIRC this is a hard requirement for maxtext to function. It is currently installed via setup.sh:
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/79e717f8983dd49ae2c18e9c548f7289a36b5148/setup.sh#L158
> 
> .
> Can this dep be added directly to https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/constraints_gpu_only.txt and requirements.txt or would it cause any issues?

We should add this to the requirements.txt if this is a hard requirement for maxtext to funciton. 

Do you need to always build from the stable branch? I tried to run `python3 -m pip install git+https://github.com/NVIDIA/TransformerEngine.git@stable` but failed due to permission issues. Can we install a released version https://pypi.org/project/transformer-engine/ instead?

requirements.txt
kanglant: The python seed env tool will still function and generate similar requirement lock files even if deps in maxtext's requirements.txt don't have lower bounds. The primary reason for this change is to update existing exact pins (`==`) to lower bounds (`>=`), e.g., `google-cloud-aiplatform>=1.61.0`.

While adding lower bounds for all other dependencies isn't strictly necessary, we recommend it to prevent potential resolution errors during lock file generation (though it is rare and unlikely). For instance, we once saw `tensorflow-metadata` resolve to a very old, python 3 incompatible version (0.5.0). Adding a lower bound like `tensorflow-metadata>=1.14.0` fixed it. This was an unusual error, possibly a uv bug or dependency issue, and we'll investigate further if it reoccurs.

parambole:
> Generally LGTM. @parambole can you take a look as well?

LGTM to me as well. @kanglant could you also document in the description - how you derived these bounds i.e. the versions. 

cc: @Rohan-Bierneni PTAL.

Rohan-Bierneni (approved):

Rohan-Bierneni:
> > For the maxtext_gpu lock file generated at https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/maxtext_requirements_lock_3_10.txt, I noticed that the package transformer_engine is missing. IIRC this is a hard requirement for maxtext to function. It is currently installed via setup.sh:
> > https://github.com/AI-Hypercomputer/maxtext/blob/79e717f8983dd49ae2c18e9c548f7289a36b5148/setup.sh#L158
> > 
> > .
> > Can this dep be added directly to https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/constraints_gpu_only.txt and requirements.txt or would it cause any issues?
> 
> We should add this to the requirements.txt if this is a hard requirement for maxtext to funciton.
> 
> Do you need to always build from the stable branch? I tried to run `python3 -m pip install git+https://github.com/NVIDIA/TransformerEngine.git@stable` but failed due to permission issues. Can we install a released version https://pypi.org/project/transformer-engine/ instead?

Yes using the pypi version should also work as well. I believe we use that in JAX AI Image flow: https://source.corp.google.com/h/cloud-tpu/infra/stable-stack/jax/+/main:gpu/Dockerfile;drc=a3e742d334eed54fcd7111517bd3480022bbecd4;l=54

kanglant:
> Yes using the pypi version should also work as well. I believe we use that in JAX AI Image flow: https://source.corp.google.com/h/cloud-tpu/infra/stable-stack/jax/+/main:gpu/Dockerfile;drc=a3e742d334eed54fcd7111517bd3480022bbecd4;l=54

Cool. I added `transformer_engine[jax]` to maxtext's requirements.txt.

kanglant:
MaxText does not allow merging from a fork. Created a new PR https://github.com/AI-Hypercomputer/maxtext/pull/1971. Close this one.

PR #1902, opened 2025-06-30
Add name to linen modules for RotaryEmbedding and LLaMARotaryEmbedding

bvandermoon:
# Description

Add name to the Linen bridges for these modules. Adding these back to keep them consistent with how they were before. In general this could impact checkpoint loading, but don't think that is happening here because there are no params in these modules (our checkpoint test is still passing)

# Tests

Ran MaxText base model on a v6e-8 devbox. Also ran Llama3.1-8b and got an OOM as expected (but did not fail before the OOM):

```
python3 -m MaxText.train MaxText/configs/base.yml \
    run_name=<run_name> \
    base_output_directory=gs://<gcs_bucket> \
    dataset_type=synthetic \
    steps=10    # model_name=llama3.1-8b (gives OOM as expected)
```

# Checklist

### DIFF
--- a/MaxText/layers/embeddings.py
+++ b/MaxText/layers/embeddings.py
@@ -129,7 +129,8 @@ def rotary_embedding_as_linen(
     max_timescale: int,
     embedding_dims: int = 0,
     cast_as_fprop_dtype: bool = True,
-    fprop_dtype: DType = jnp.bfloat16
+    fprop_dtype: DType = jnp.bfloat16,
+    name: str | None = None
 ):
   """Initializes the RotaryEmbedding module and returns it as a Linen module.
 
@@ -141,6 +142,7 @@ def rotary_embedding_as_linen(
     embedding_dims: Dimension of the embedding to be generated.
     cast_as_fprop_dtype: Whether to cast the output to the fprop dtype.
     fprop_dtype: The dtype of the output.
+    name: Name of the Linen module.
   """
   return nnx.bridge.to_linen(
     RotaryEmbedding,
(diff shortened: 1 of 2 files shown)


MaxText/layers/embeddings.py
```diff
@@ -129,7 +129,8 @@ def rotary_embedding_as_linen(
     max_timescale: int,
     embedding_dims: int = 0,
     cast_as_fprop_dtype: bool = True,
-    fprop_dtype: DType = jnp.bfloat16
+    fprop_dtype: DType = jnp.bfloat16,
+    name: str = None
```
cgarciae:
```suggestion
    name: str | None = None
```

MaxText/layers/embeddings.py
```diff
@@ -244,7 +247,8 @@ def llama_rotary_embedding_as_linen(
     embedding_dims: int = 0,
     cast_as_fprop_dtype: bool = True,
     fprop_dtype: DType = jnp.bfloat16,
-    use_scale: bool = True
+    use_scale: bool = True,
+    name: str = None
```
cgarciae:
```suggestion
    name: str | None = None
```

cgarciae (approved):

MaxText/layers/embeddings.py
bvandermoon: Done, thanks Cristian

MaxText/layers/embeddings.py
bvandermoon: Done, thanks Cristian

gobbleturk (approved):

PR #1595, opened 2025-04-16
Add pipreqs requirements file for JSS 0.5.2

bvandermoon:
# Description

Use a hardcoded requirements.txt file for JSS 0.5.2. This is a short-term solution to urgently support current recipes. We have a separate track to solve MaxText versioning long-term.

This PR intends to solve the problem of Docker builds changing over time. If a user checks out this MaxText commit and builds the Docker image, it should very closely match regardless of when the image is built.

Note: I used a tool called `pipreqs` to generate this file since `pip freeze` had a bunch of conflicts. I manually loosened/changed some of dependencies to handle the conflicts. I left the original values as comments so we don't lose what was actually generated.

# Tests

Built the Docker image and ran the benchmark runner:

```
BASE_IMAGE=us-docker.pkg.dev/cloud-tpu-images/jax-stable-stack/tpu:jax0.5.2-rev1
bash docker_build_dependency_image.sh DEVICE=tpu MODE=stable_stack BASEIMAGE=${BASE_IMAGE}

python3 -m benchmarks.benchmark_runner xpk \
    --project=${PROJECT} \
    --zone=${ZONE} \
    --device_type=v6e-256 \
    --num_slices=1  \
    --cluster_name=${CLUSTER_NAME} \
    --base_output_directory=${OUTPUT_DIR} \
    --model_name="mixtral_8x7b_dropped" \
    --base_docker_image=maxtext_base_image
```

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #1905, opened 2025-06-30
Add configs for Llama 3.1 in small v6e clusters

cfbh-google:
# Description

Add missing Llama 3.1 configurations for small clusters (derived from b/410302796). The small clusters covered include Llama 3.1 8B on v6e-64, and Llama 3.1 70B on v6e-32, v6e-64, v6e-128. Also, add requirements file for JAX stable stack 0.6.1 generated by pipreqs.

Achieving good MFU for Llama 3.1 8B and 70B on small clusters (<256 chips) requires changes on top of the existing model configs (e.g. llama3_1_8b_8192 and llama3_1_70b_8192) for certain accelerator types (v6e-32, v6e-64, v6e-128). This PR adds the missing configurations to achieve good MFU for such small clusters and be able to create recipes.

### What is the problem being solved and any relevant context?

Users required to run Llama 3.1 or similar models on small clusters (b/410302796). We were missing some configs to get good MFU on such small clusters. We also need to find the parameters that make a large 70B model fit into small clusters. We will create recipes to share the findings.

### Why this is a good solution?

We thoroughly tested different configurations for small clusters (v6e-8, v6e-16, v6e-32, v6e-64, v6e-128, v6e-256) until we found solutions with good MFU that makes customers happy (b/410302796).

### What would be some information about the specific implementation?

Derived new model configs from existing ones (llama3_1_8b_8192 and llama3_1_70b_8192). Updated batch size, XLA flags and weight dtype. Added dependencies for using latest JAX stable stack 0.6.1 using pipreqs.

### What are the shortcomings of the solution and possible future improvements?

In order to fit Llama 3.1 70B into v6e-32 (32 chips) without OOM, we used bfloat16 for the weight dtype. We did get good MFU, but the model performance was impacted (higher loss). We tested several configurations including different remat policies and batch sizes, but didn't find anything that could avoid the use of bfloat16.

Same as PR #1595, hardcoding requirements.txt file for JSS 0.6.1 is a short-term solution to urgently support current recipes, but there is a separate track to solve MaxText versioning long-term.

FIXES: b/419848240
FIXES: b/424855986

# Tests

The new configurations added in this PR were tested by running in small clusters:
* llama3_1_8b_8192_bs5 on v6e-64
* llama3_1_70b_8192_bs2_bfloat16_no_collective_matmul on v6e-32
* llama3_1_70b_8192_bs2 on v6e-64
* llama3_1_70b_8192_bs4 on v6e-128

Also tested the rest of the small clusters using the existing configs to verify there is no regression:
* llama3_1_8b_8192_no_collective_matmul on v6e-8, v6e-16, v6e-32
* llama3_1_8b_8192 on v6e-128, v6e-256
* llama3_1_70b_8192 on v6e-256

# Checklist


maxtext_jax_ai_image.Dockerfile
```diff
@@ -35,7 +35,15 @@ RUN if [ "$DEVICE" = "tpu" ] && ([ "$JAX_AI_IMAGE_BASEIMAGE" = "us-docker.pkg.de
 RUN apt-get update && apt-get install --yes && apt-get install --yes dnsutils
 # TODO(bvandermoon, parambole): Remove this when it's added to JAX AI Image
 RUN pip install google-cloud-monitoring
-RUN python3 -m pip install -r /deps/requirements_with_jax_ai_image.txt
+
+# Install requirements file that was generated with pipreqs for JSS 0.6.1 using:
+# pipreqs --savepath requirements_with_jax_stable_stack_0_6_1_pipreqs.txt
+# Otherwise use general requirements_with_jax_ai_image.txt
+RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:jax0.6.1-rev1" ]; then \
```
gobbleturk, line 42: Maybe we can discuss with branden and param - we now publish publicly available images of JAI + maxtext, maybe we can point to one of these public images instead

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -868,6 +868,65 @@
 )
 
 
+llama3_1_8b_8192_bs5 = _add_to_model_dictionary(
```
gobbleturk, line 871: can you include the slice size either in the model config name or at least in a comment above the model config name?

cfbh-google: I'll include this as comments.

maxtext_jax_ai_image.Dockerfile
bvandermoon: Discussed offline with @gobbleturk, @parambole, and @Rohan-Bierneni. That image is used for our nightly tests, but is not available in a way that can be built by tpu-recipes currently. We are still blocked while the long-term versioning work is ongoing. We discussed sticking with this (unfortunate) pattern until we can get to that better state

benchmarks/maxtext_trillium_model_configs.py
```diff
@@ -868,6 +868,65 @@
...
+        },
+        xla_flags=(
+            xla_flags_library.DENSE_VMEM_LIMIT_FLAG
+            + xla_flags_library.LAYOUT_FOR_ALL_REDUCE_SCATTER
+            + xla_flags_library.DATA_PARALLEL_OVERLAP
+            + xla_flags_library.CF_FOR_ALL_GATHER
+            + xla_flags_library.ENABLE_SPARSECORE_OFFLOADING_FOR_ALL_REDUCE
+            + xla_flags_library.HOST_OFFLOAD_FLAGS
+        ),
+        pathways_xla_flag_options={
```
bvandermoon: Why do we need pathways flags?

cfbh-google: I just copied the original config from llama3_1_8b_8192 and updated the batch size. I am not sure why the original config needs Pathways.

benchmarks/maxtext_trillium_model_configs.py
cfbh-google: Done.

maxtext_jax_ai_image.Dockerfile
cfbh-google: Got it. In that case I'll leave it as is, and resolve this conversation thread.

benchmarks/maxtext_trillium_model_configs.py
cfbh-google: Discussed offline with @bvandermoon. I removed the unnecessary pathways config.

requirements_with_jax_stable_stack_0_6_1_pipreqs.txt
```diff
@@ -0,0 +1,53 @@
...
+cloud_accelerator_diagnostics==0.1.1
+cloud_tpu_diagnostics==0.1.5
+datasets==3.6.0
+etils==1.12.2
+evaluate==0.4.4
+flax==0.10.6
+grain==0.2.10
+grpcio==1.72.0rc1
+huggingface_hub==0.33.0
+jax==0.6.0
```
bvandermoon, line 14: Could you also pin libtpu and jaxlib?

bvandermoon (commented):
LGTM, just one comment

requirements_with_jax_stable_stack_0_6_1_pipreqs.txt
cfbh-google: Done.

requirements_with_jax_stable_stack_0_6_1_pipreqs.txt
gobbleturk, line 1: I am sad but will approve regardless

gobbleturk (approved):

bvandermoon (approved):

PR #1908, opened 2025-07-01
All Gather Once (FSDP Zero-one as an option)

wei879-100:
# Description

This PR introduces an independent Zero-1 sharding optimization before the first microbatch without pipeline parallelism (PP) as an option. Previously, this "all-gather once" (AG-once) functionality was only available in conjunction with PP. This change enables `model_fsdp_ag_once=True` to work as a standalone optimization without PP, improving performance by significantly reducing communication overhead.

This change is being made to address the problem of redundant all-gather operations in FSDP  training. In the past, with `model_fsdp_ag_once=False` (the default), an all-gather operation on model weights would occur twice per microbatch (once for the forward pass and once for the backward pass). For large models and deep pipelines, these repeated communications can become a bottleneck.

The solution is implemented by introducing a ZeroOneTransformer wrapper around the base Transformer model. This wrapper, when `model_fsdp_ag_once` is set to True, orchestrates the single all-gather operation for the model's parameters using `jax.lax.with_sharding_constraint`.

A potential shortcoming of this solution is the increased memory usage, as the fully gathered weights must be held in memory for the entire step, the next step is to add the gradient accumulation to allow weights are gathered only once before the first microbatch and gradients synced only once after the last microbatch. There is a memory/time tradeoff.

Xplane:
`model_fsdp_ag_once = True` , `pipeline_fsdp_ag_once = False` :  
https://xprof.corp.google.com/trace_viewer/fanwei-14931235522453273514?hosts=t1v-n-aade02bb-w-0&host_index=0&trace_filter_config={}

`model_fsdp_ag_once = False` , `pipeline_fsdp_ag_once = False` :  
https://xprof.corp.google.com/trace_viewer/fanwei-6842216908212951232?hosts=t1v-n-aade02bb-w-0&host_index=0&trace_filter_config={}&view_start=-377.916&view_end=2907.326

FIXES: b/406257368

# Tests

This change has been tested using existing integration tests for MaxText. The following tests were run with `model_fsdp_ag_once=True` to verify the functionality and ensure no regressions:

`test_tpu_base_model_ag_once` (on TPU)

`test_gpu_synthetic_model_ag_once` (on GPU)

`test_gpu_context_parallel_ag_once` (on GPU with context parallelism)

Execute the respective tests internally call train.py with the `model_fsdp_ag_once=True` enabled in the configuration file base.yml.

# Checklist


MaxText/configs/base.yml
```diff
@@ -186,6 +186,9 @@ pipeline_delay_activation_forwarding: False # This delays the activation forward
 # the communication and compute in each iteration are now independent. However this comes at the cost of doubling the pipeline bubble,
 # and you must set the number of microbatches to at least 2 * num_stages (the minimum 2 * num_stages is set by default with this delay).
 
+model_fsdp_ag_once: True # This controls whether the Zero-1 optimization is active.
```
gobbleturk: False by default

MaxText/pyconfig.py
```diff
@@ -927,6 +921,10 @@ def validate_sparse_matmul_parallelism(raw_keys):
     raise ValueError(
         f"The expert dimension {raw_keys['num_experts']} is not divisible by expert parallelism setting {expert_parallelism}."
     )
+  if using_pipeline_parallelism(raw_keys) and (
+      raw_keys["pipeline_fsdp_ag_once"] is False or raw_keys["model_fsdp_ag_once"] is True
```
gobbleturk: Logic:

if using_pp and model_fsdp_ag_once:
fail

error message:
set model_fsdp_ag_once=False, and use pp_fsdp_ag_once=True

MaxText/pyconfig.py
```diff
@@ -513,10 +511,6 @@ def user_init(raw_keys):
     """Transformations between the config data and configs used at runtime"""
     if raw_keys["run_name"] == "":
       raw_keys["run_name"] = os.environ.get("JOBSET_NAME")  # using XPK default
-      if raw_keys["run_name"] == "":
-        now = datetime.datetime.now()
-        timestamp = now.strftime("%Y-%m-%d-%H-%M")
-        raw_keys["run_name"] = f'{raw_keys["model_name"]}_{timestamp}'
```
bvandermoon: Why is this removed?

MaxText/pyconfig.py
```diff
@@ -908,7 +902,7 @@ def validate_deepseek_moe(raw_keys):
 
 
 def validate_sparse_matmul_parallelism(raw_keys):
-  if raw_keys["sparse_matmul"] and (using_sequence_parallelism(raw_keys) or using_pipeline_parallelism(raw_keys)):
+  if raw_keys["sparse_matmul"] and using_sequence_parallelism(raw_keys):
     raise ValueError("Currently we only support Megablox and Ragged dot with data, tensor, and expert parallelism.")
```
bvandermoon: Does the error message also need to be updated since the `using_pipeline_parallalism(...)` check was removed?

MaxText/layers/models.py
```diff
@@ -819,3 +817,72 @@ def __call__(
...
+          decoder_input_tokens,
+          decoder_positions,
+          decoder_segment_ids,
+          encoder_images,
+          enable_dropout,
+          model_mode,
+          previous_chunk,
+          true_length,
+          slot,
+          page_state,
```
bvandermoon, line 869: Nit: Ideally would use the named params for passing these in. Really minor though

bvandermoon (approved):
This looks great, thank you @wei879-100

MaxText/pyconfig.py
wei879-100: due to the Github issues, I solved this

MaxText/pyconfig.py
wei879-100: I did not change this part. I tried to pull the latest version again.

PR #1942, opened 2025-07-08
Enable scan for gemma3

gagika:
# Description

Enable scan support for Gemma3 models which use a repeating pattern of local and global attention defined by GEMMA3_ATTENTION_PATTERN.

The approach uses a new Gemma3ScannableBlock, which encapsulates this repeating pattern. The decoder processes full blocks within an nn.scan operation and applies any remainder layers separately, ensuring the model is correct for any number of layers.

# Tests

Tested 12b and 27b configs

# Checklist

### DIFF
--- a/MaxText/layers/decoders.py
+++ b/MaxText/layers/decoders.py
@@ -672,12 +672,21 @@ class Decoder(nn.Module):
               deterministic,
               model_mode,
           )
+        elif cfg.decoder_block == DecoderBlockType.GEMMA3:
+          y = self._apply_gemma3_scanned_blocks(
+              y,
+              decoder_segment_ids,
+              decoder_positions,
+              deterministic,
+              model_mode,
+              bidirectional_mask,
+              previous_chunk,
+              page_state,
+              slot,
+          )
         else:
-          layer_call_kwargs = {}
           layer_kwargs = {}
-          if cfg.decoder_block == DecoderBlockType.GEMMA3:
-            layer_call_kwargs = {"bidirectional_mask": bidirectional_mask}
-          elif cfg.decoder_block == DecoderBlockType.LLAMA4:
+          if cfg.decoder_block == DecoderBlockType.LLAMA4:
             layer_kwargs = {
                 "nope_layer_interval": self.config.nope_layer_interval,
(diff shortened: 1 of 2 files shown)


MaxText/layers/models.py
```diff
@@ -343,7 +343,10 @@ def get_decoder_layers(self):
     elif self.config.decoder_block == DecoderBlockType.GEMMA3:
       from MaxText.layers import gemma3  # pylint: disable=import-outside-toplevel
 
-      return [gemma3.Gemma3DecoderLayer]
+      if self.config.scan_layers:
+        return [gemma3.Gemma3ScannableBlock]
```
RissyRan: Is that possible to return `Gemma3ScannableBlock` in both case? I see `if cfg.scan_layers` in scannable block. I recall we plan to do so in llama4, but blocked by updating the ckpt conversion or something?

gagika: I think it's simpler to separate them (and keep consistant with Llama4),  Gemma3ScannableBlock composed of Gemma3DecoderLayer layers following attention pattern (Local, local, ..., Global). 
If we use Gemma3ScannableBlock for unscanned model too, it will have only one nested Gemma3DecoderLayer and additional layer id to decide local vs gloabal attention (and the name needs to be updated)

gobbleturk (approved):
Looks great! Can you include in the description in your PR the approach to support scan - repeatable block + remainder?

MaxText/layers/models.py
```diff
@@ -701,6 +715,65 @@ def __call__(
...
+          deterministic,
+          model_mode,
+          **layer_call_kwargs,
+      )
+      y = scanned_layers_output
+
+    # Apply any remaining layers that did not fit into a full scanned block
+    num_remaining_layers = cfg.num_decoder_layers % attention_pattern_length
+    if num_remaining_layers > 0:
+      # We name the remainder block with a 'remainder' suffix to avoid parameter name collisions
```
hengtaoguo: Out of curiosity, does this "remainder" string impact the name mapping in checkpoint conversion scripts?

hengtaoguo (approved):
Thank you Gagik!

PR #1943, opened 2025-07-08
Enable AOT Test for HLO Graph Verification

NuojCheng:
# Description

This test verifies that the HLO graph generated by AOT compilation `train_compile.py` is identical to the one generated by a JIT compilation during a training run by `train.py`. The test consists of the following steps:

1. Execute both `train_compile.py` and `train.py` for a minimal number of steps;
2. Instruct both runs to dump their HLO to local temporary directories;
3. Compare the overlapped HLO file from each run to ensure they are identical. The comparison is done by SHA256 hashing. 

FIXES: b/59595326

# Tests

Both regular and int8 quantization tests are pytest passed.

# Checklist


MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -17,102 +17,151 @@
...
+    """
+    jax.config.update("jax_enable_compilation_cache", False)
+    temp_dir = tempfile.gettempdir()
+    self.dump_dir = os.path.join(temp_dir, "aot_test_dump")
+    xla_dump_options = "--xla_dump_hlo_as_text --xla_dump_hlo_module_re=jit_train_step"
+    os.environ["XLA_FLAGS"] = f"--xla_dump_to={self.dump_dir} {xla_dump_options}"
+
+  def get_device_info(self):
+
+    """Gets TPU device info to generate correct AOT arguments."""
```
shralex: I'm not sure it can help, but check out accelerator_to_spec_map.py
it is actually used during AOT compilation, and provides a lot of the details about the architecture. It might be better if you can use it from there rather than specify it again here.

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -17,102 +17,151 @@
...
-    compile_hlo_file = self.find_file_by_substring(compile_dump_dir, hlo_filename_substring)
-    train_hlo_file = self.find_file_by_substring(train_dump_dir, hlo_filename_substring)
-    print(f"AOT compiled HLO file for test {test_name}: {compile_hlo_file}", flush=True)
-    print(f"Real runs HLO file for test {test_name}: {train_hlo_file}", flush=True)
+    # Step 1: generate train.py HLO graphs
+    train.main(train_argv)
+    shutil.move(self.dump_dir, train_dump_dir)
+    jax.clear_caches()
+
+    # Step 2: generate train_compile.py HL graphs
```
shralex, line 136: is it possible to specify the output HLO file name, so we don't need to search for it in step 3 ?

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright 2024 Google LLC
+Copyright 2025 Google LLC
```
shralex, line 2: Could you please add a comment explaining what the tests in this file do

shralex (commented):
Thanks Nuojin! Added a few minor comments. A more major suggestion -- currently the tests in train_compile_test are not asserting anything, as far as I remember. They just run AOT compilation. Instead of adding more tests here and running AOT comilation twice, could we somehow merge this with the other tests so that we not only run the AOT but also validate the produced HLO ?

MaxText/tests/aot_hlo_identical_test.py
NuojCheng: > is it possible to specify the output HLO file name, so we don't need to search for it in step 3 ?

It is XLA that determines the HLO file names, which usually include some arbitrary module and cl numbers. It is not clear to me how to get rid of the cl name part. Searching file part is not expensive though, the number of HLO file is not large.

MaxText/tests/aot_hlo_identical_test.py
NuojCheng: Thank you for pointing out! I checked the `UserFacingNameToSystemCharacteristics` object, it maps user facing names to specific system properties, while here I want to generate the user face names from jax.devices(). I don't think MaxText currently has this function, maybe we can consider putting this function under accelerator_to_spec_map.py? idk

I changed this function name to `get_device_user_facing_name()`. It sounds better.

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -11,108 +11,158 @@
 limitations under the License.
 """
 
+""" Tests for Verifying HLO Identify from Ahead of Time Compilation"""
```
shralex: Can we change this to be a bit more descriptive:
"These tests verify the HLO graphs generated by AOT compilation (using train_compile.py), by making sure they are identical to HLO generated from a real training run (using train.py)."

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -11,108 +11,158 @@
...
+    train.main(train_argv)
+    shutil.move(self.dump_dir, train_dump_dir)
+    jax.clear_caches()
+
+    # Step 2: generate train_compile.py HL graphs
+    train_compile.main(compile_argv)
+    shutil.move(self.dump_dir, compile_dump_dir)
+    jax.clear_caches()
+
+    # Step 3: specify the HLO files and check their homogeneity
```
shralex: "check their homogeneity" -> "check that they are identical"

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -11,108 +11,162 @@
...
-      shutil.rmtree(directory)
+      raise ValueError(f"Unsupported TPU device kind for AOT test: {device_kind}")
+
+    return f"{prefix}-{topology_devices}"
+
+  def find_HLO_files(self, compile_dump_dir, real_dump_dir):
+    """
+    Find the HLO file
+    """
+    pattern = re.compile(r"^.*\.jit_train_step\..*\.after_optimizations_after_buffer_assignment\.txt$")
```
hengtaoguo, line 74: Thanks! We are comparing "jit_train_step" and "after_optimizations_after_buffer_assignment" in the end? Maybe add a comment to say these are the two key files of interest?

hengtaoguo (approved):

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -11,108 +11,162 @@
...
+    if "TPU v4" in device_kind:
+      prefix = "v4"
+      topology_devices = 2 * num_devices
+    elif "TPU v5 lite" in device_kind:
+      prefix = "v5e"
+    elif "TPU v5p" in device_kind:
+      prefix = "v5p"
+      topology_devices = 2 * num_devices
+    elif "TPU v6" in device_kind:
+      prefix = "v6e"
```
bvandermoon: Nit: Could do a map like this to get rid of the if/else for every TPU version 

```
{
  "TPU v4": ("v4", 2 * num_devices))
  ...
}
```

Then just check for device_kind in the map

MaxText/tests/aot_hlo_identical_test.py
```diff
@@ -11,108 +11,162 @@
...
+    real_files = set(os.listdir(real_dump_dir))
+    compile_hlo, real_hlo = None, None
+    # HLO file satisfying above pattern should uniquely exist
+    for file in compile_files:
+      if pattern.search(file):
+        compile_hlo = file
+    for file in real_files:
+      if pattern.search(file):
+        real_hlo = file
+    return compile_hlo, real_hlo
```
bvandermoon, line 85: Just for my understanding, why can't we just read the paths directly?

NuojCheng: HLO files usually has format "module_0000..jit_train_step.cl_0000..after_optimizations_after_buffer_assignment.txt". The numbers following 'module' and 'cl' are pretty arbitrary with no clear pattern. As a result we use the re and search to find the corresponding HLO files.

MaxText/tests/aot_hlo_identical_test.py
NuojCheng: Thank you for the suggestion. It is slightly different here because we are loooking for `"TPU v4" in device_kind` instead of `"TPU v4" is device_kind`, but I can still use the map in a different way.

MaxText/tests/aot_hlo_identical_test.py
NuojCheng: done!

PR #1815, opened 2025-06-10
Refactor metrics related code and fix calculation for step time metrics

SurbhiJainUSC:
# Description

* This PR removes `writer` from the list of values returned by `setup_mesh_and_model`. This simiplifies `setup_mesh_and_model` as it is unnecessary to create a TensorBoard summary writer object in that method. 
* This PR also refactors the metrics related code in all trainers (train.py, sft_trainer.py, grpo_trainer.py, elastic_train.py).
* This PR also fixes the calculation for metric `perf/step_time_seconds` for training. Currently, this metric for step `x` includes the time taken for checkpointing + HLO dump + eval + profiling by step `x-1`. This PR fixes this by ensuring that time taken by checkpointing + HLO dump + eval + profiling by step `x` is added to metric `perf/step_time_seconds` for step `x` only. Depends on profiling fixes [#1833](https://github.com/AI-Hypercomputer/maxtext/pull/1833).

Thanks to @richjames0 for contributing!

# Tests
E2E testing with train.py and sft_trainer.py - verified the metrics are correctly uploaded to TensorBoard

### Logs for refactored code with metric fix:
```
steps=10 profiler=xplane checkpoint_period=7

# training + checkpointing
completed step: 0, seconds: 27.544, TFLOP/s/device: 1.534, Tokens/s/device: 37.177, total_weights: 3573, loss: 0.993

# training + profiling activated
completed step: 1, seconds: 1.105, TFLOP/s/device: 38.242, Tokens/s/device: 927.006, total_weights: 3253, loss: 0.993

# training
completed step: 2, seconds: 0.996, TFLOP/s/device: 42.403, Tokens/s/device: 1027.874, total_weights: 1692, loss: 0.912
completed step: 3, seconds: 0.996, TFLOP/s/device: 42.426, Tokens/s/device: 1028.421, total_weights: 2944, loss: 1.105
completed step: 4, seconds: 0.994, TFLOP/s/device: 42.517, Tokens/s/device: 1030.628, total_weights: 3466, loss: 0.948

# training + profiling deactivated
completed step: 5, seconds: 16.416, TFLOP/s/device: 2.573, Tokens/s/device: 62.376, total_weights: 2420, loss: 0.947

# training
completed step: 6, seconds: 0.982, TFLOP/s/device: 43.037, Tokens/s/device: 1043.238, total_weights: 1655, loss: 1.059

# training + checkpointing
[...]

PR #1956, opened 2025-07-10
Buffer metrics before writing to GCS/Tensorboard

SurbhiJainUSC:
# Description

This PR adjusts how we calculate step time, reverting to the method used before https://github.com/AI-Hypercomputer/maxtext/pull/1815. Our previous attempt to precisely measure step time involved adding `jax.block_until_ready()`, which, while accurate for individual steps, introduced an undesirable delay between `train_steps` during profiling. This delay stemmed from JAX being prevented from queuing up the next training step. To improve training performance, we're now prioritizing the ability for JAX to queue steps efficiently, even if it means a slight compromise in the precision of individual step time reports.

# Tests

```
completed step: 0, seconds: 18.665, TFLOP/s/device: 2.263, Tokens/s/device: 54.862, total_weights: 3573, loss: 0.993
completed step: 1, seconds: 9.720, TFLOP/s/device: 4.346, Tokens/s/device: 105.355, total_weights: 3253, loss: 0.993
completed step: 2, seconds: 0.164, TFLOP/s/device: 256.874, Tokens/s/device: 6226.779, total_weights: 1692, loss: 0.912
completed step: 3, seconds: 0.834, TFLOP/s/device: 50.645, Tokens/s/device: 1227.665, total_weights: 2944, loss: 1.105
completed step: 4, seconds: 0.957, TFLOP/s/device: 44.151, Tokens/s/device: 1070.256, total_weights: 3466, loss: 0.948
completed step: 5, seconds: 0.961, TFLOP/s/device: 43.949, Tokens/s/device: 1065.339, total_weights: 2420, loss: 0.946
completed step: 6, seconds: 16.063, TFLOP/s/device: 2.630, Tokens/s/device: 63.748, total_weights: 1655, loss: 1.058
completed step: 7, seconds: 0.013, TFLOP/s/device: 3333.582, Tokens/s/device: 80808.081, total_weights: 3335, loss: 1.000
completed step: 8, seconds: 0.964, TFLOP/s/device: 43.820, Tokens/s/device: 1062.227, total_weights: 1676, loss: 0.850
completed step: 9, seconds: 0.962, TFLOP/s/device: 43.895, Tokens/s/device: 1064.035, total_weights: 1676, loss: 0.965
```

Gap between two `jit_train_steps` reduced from 30ms 
image

to 13.8us 
image

# Checklist

### DIFF
--- a/MaxText/elastic_train.py
+++ b/MaxText/elastic_train.py
@@ -209,11 +209,12 @@ def train_loop(config, elastic_manager, recorder, state=None):
   # Write train config params, num model params, and XLA flags to tensorboard
   metric_logger.write_setup_info_to_tensorboard(state.params)
 
+  last_step_completion = datetime.datetime.now()
+
   # Using while loop instead of a for loop because with elasticity
   # the step is restored back to the latest snapshot when a slice is lost
   while step < config.steps:
     try:
-      step_start_time = datetime.datetime.now()
       prof.maybe_activate_profiler(step, state)
 
       max_logging.log(f"{step=} {elastic_manager.elastic_down_event_count=} {elastic_manager.good_slice_count=}")
@@ -225,6 +226,9 @@ def train_loop(config, elastic_manager, recorder, state=None):
           with maybe_record_goodput(recorder, GoodputEvent.STEP, step):
             state, metrics = p_train_step(state, example_batch, nextrng)
 
+        step_time_delta = datetime.datetime.now() - last_step_completion
+        last_step_completion = datetime.datetime.now()
+
         checkpointing.maybe_save_checkpoint(checkpoint_manager, state, config, data_iterator, step)
 
         prof.maybe_deactivate_profiler(step, state)
@@ -232,10 +236,7 @@ def train_loop(config, elastic_manager, recorder, state=None):
       if step == start_step:
         max_utils.print_mem_stats("After params initialized")
 
-      jax.block_until_ready(state)  # ensure training step is completed
-
-      step_time_delta = datetime.datetime.now() - step_start_time
-      metric_logger.record_train_metrics(metrics, step, step_time_delta)
+      metric_logger.buffer_and_write_train_metrics(metrics, step, step_time_delta)
 
       elastic_manager.maybe_snapshot(
           step=step,
@@ -304,7 +305,7 @@ def train_loop(config, elastic_manager, recorder, state=None):
       max_logging.log(f"Training stopped: {str(error)}")
 
   checkpointing.maybe_save_checkpoint(checkpoint_manager, state, config, data_iterator)
-  metric_logger.cleanup()
+  metric_logger.flush_metrics_and_cleanup()
 
   return state
 
--- a/MaxText/metric_logger.py
+++ b/MaxText/metric_logger.py
@@ -57,6 +57,7 @@ class MetricLogger:
     self.performance_metric_queue = self.get_performance_metric_queue(config)
     self.learning_rate_schedule = learning_rate_schedule
     self.cumulative_eval_metrics = {"scalar": defaultdict(float)}
+    self.buffered_train_metrics = None
 
   def write_metrics(self, metrics, step, is_training=True):
     """Entry point for all metrics writing in Train's Main."""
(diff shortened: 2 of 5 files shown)


MaxText/metric_logger.py
```diff
@@ -153,8 +154,17 @@ def get_performance_metric_queue(self, config):
         gcp_workload_monitor.start_performance_reporting_thread(performance_metric_queue)
     return performance_metric_queue
 
+  def buffer_and_write_train_metrics(self, metrics, step, step_time_delta):
```
gobbleturk, line 157: Can you add this buffer approach allows train steps to be executed back to back, in particular allows overlapping the data loading of step n during the execution of step n-1

SurbhiJainUSC: Updated.

MaxText/elastic_train.py
```diff
@@ -225,17 +226,17 @@ def train_loop(config, elastic_manager, recorder, state=None):
...
+        last_step_completion = datetime.datetime.now()
+
         checkpointing.maybe_save_checkpoint(checkpoint_manager, state, config, data_iterator, step)
 
         prof.maybe_deactivate_profiler(step, state)
 
       if step == start_step:
         max_utils.print_mem_stats("After params initialized")
 
-      jax.block_until_ready(state)  # ensure training step is completed
```
shauryagup, line 235: After removing this, wouldn't the step time be simply the time to dispatch the step and not its actual execution time? This behavior could also result in different step times between pathways and mcjax at least on the JAX layer since both of them use different dispatch strategies. 

Do we know what was python doing during the 30 ms reported in the first profile? Was it only spent on dispatching the operation? Or does it also include data loading etc.?

MaxText/metric_logger.py
```diff
@@ -153,8 +154,22 @@ def get_performance_metric_queue(self, config):
...
```
lukebaumann, line 157: Can you optionally block on the metrics so that the user can choose to measure step time (dispatch N to finish N) and have a simpler flow (avoiding buffering of metrics and queuing of steps) vs measuring step time (dispatch N to dispatch N+1) and added complexity?

MaxText/elastic_train.py
SurbhiJainUSC: https://github.com/AI-Hypercomputer/maxtext/pull/1815 was actually fixing the issue with reported step time. Before #1815, we were taking into account the time it takes for step n-1 to complete when reporting step time for step n. With #1815, I tried fixing this issue by ensuring that step time for step n only includes the actual time it takes to complete step n and not step n-1. This was achieved by adding `jax.block_until_ready()`. However, this blocks JAX from queuing any operation until the current training step is completed.

In the first profile, 30ms include the time for data loading for step n before starting training step.

There is a tradeoff between accuracy (reporting the correct step time) vs performance here.

MaxText/metric_logger.py
```diff
@@ -153,8 +154,22 @@ def get_performance_metric_queue(self, config):
...
+    """
+    Buffers metrics for the current training step and simultaneously writes the training metrics
+    for the previous step to GCS and/or TensorBoard. This buffering strategy allows for back-to-back
+    execution of training steps, by overlapping data loading for step n with the execution of step n−1.
+    This significantly boosts training efficiency.
```
lukebaumann, line 162: IIUC in the example step times were about 0.9 seconds and this saves (32 ms - 13 us) ~ 0.030 seconds or 3.3% difference.

Should be changed to: "This boosts training efficiency by a few percent" since "significant" is subjective.

MaxText/elastic_train.py
lukebaumann: > Do we know what was python doing during the 30 ms reported in the first profile? Was it only spent on dispatching the operation? Or does it also include data loading etc.?

As an alternative to this PR's approach, if the loop is reordered so that data loading for the next step happens at the end of the previous iteration and the block on metrics happens after the data loading is dispatched, you may achieve similar performance without needed to buffer the metrics.

MaxText/metric_logger.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, config, learning_rate_schedule):
     self.performance_metric_queue = self.get_performance_metric_queue(config)
     self.learning_rate_schedule = learning_rate_schedule
     self.cumulative_eval_metrics = {"scalar": defaultdict(float)}
+    self.buffered_train_metrics = None
```
lukebaumann, line 60: AFAICT the buffered metric from before a failure may be incomplete but because `elastic_handler` creates a brand new `MetricLogger` there is no change needed.

lukebaumann (commented):
I will run some elastic training runs from this branch and let you know if they pass.

MaxText/metric_logger.py
SurbhiJainUSC: IMO, configuring this will add too much complexity in the code on top of another config variable (which would be a different pain to handle).

MaxText/elastic_train.py
lukebaumann: We can discuss the alternative as a potential future improvement if you are interested.

> Do we know what was python doing during the 30 ms reported in the first profile? Was it only spent on dispatching the operation? Or does it also include data loading etc.?
I think this is unanswered still so I will let @shauryagup resolve the conversation when he is satisfied.

MaxText/metric_logger.py
lukebaumann: Confirmed

MaxText/metric_logger.py
lukebaumann: I think there is a simple implementation of my idea but that discussion won't block this PR.

MaxText/metric_logger.py
lukebaumann: Depending on the configuration, the savings will change but I would be curious if it is ever more than a "few percent".

This suggestion also shouldn't block this PR so I am resolving.

lukebaumann:
> I will run some elastic training runs from this branch and let you know if they pass.

I ran elastic training from this branch on 2 slices and kill each slice at different times for a total of three disruptions. The training reacted exactly as expected.

Thanks!

lukebaumann (approved):

khatwanimohit (approved):

PR #1959, opened 2025-07-11
Migrate PagedAttentionOp to use NNX

bvandermoon:
# Description

Use NNX for the PagedAttentionOp module.

* Change module to NNX
* Move `init_or_get_kv_pages` logic to `__init__` and `get_kv_pages`
* Create bridge to Linen function and use it where the PagedAttentionOp module is invoked

# Tests

Ran decode with `attention=paged` and the output is still generated. I didn't use a checkpoint but the text generation output matches main.

```
python3 -m MaxText.decode MaxText/configs/base.yml \
    model_name=llama3.1-8b \
    tokenizer_path=assets/tokenizer_llama3.tiktoken \
    tokenizer_type=tiktoken \
    attention=paged \
    scan_layers=false \
    per_device_batch_size=1 \
    ici_fsdp_parallelism=1 \
    ici_autoregressive_parallelism=-1 \
    max_target_length=256 \
    prompt="I love to"
```

@vipannalla @patemotter please let me know if there are better inference tests (besides the unit tests) to confirm this is still working properly.

# Checklist

### DIFF
--- a/MaxText/inference/paged_attention.py
+++ b/MaxText/inference/paged_attention.py
@@ -20,6 +20,7 @@ WARNING: THIS FILE IS A WORK IN PROGRESS.
 import functools
 from typing import Optional
 
+import jax
 import jax.numpy as jnp
 from jax.experimental.shard_map import shard_map
 from jax.experimental.pallas.ops.tpu.paged_attention import paged_attention_kernel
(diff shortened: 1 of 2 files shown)


MaxText/inference/paged_attention.py
```diff
@@ -134,8 +244,8 @@ def paged_dot_product_attention_with_max_and_sum(self, query, key, value):
   def paged_attention_v2_prefill(
       self,
       query: Array,
-      key_pages_var: nn.Variable,
-      value_pages_var: nn.Variable,
+      key_pages_var: nnx.Cache,
+      value_pages_var: nnx.Cache,
```
patemotter: Optional: we can rename these variables to no longer use the `_var` suffix since they are not `nn.Variable` type any more. Not a requirement though.

bvandermoon: Done, thanks @patemotter. Good catch. These variables could probably also be read directly from self instead of passed around (not related to this change). Left this out of this change though. The NNX migration already requires a lot of changes so we are trying to leave off refactoring

jrplatin (approved):

richjames0 (approved):
LGTM

PR #1969, opened 2025-07-14
Multimodal quality benchmark

hengtaoguo:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/benchmarks/multimodal/multimodal_eval.py
+++ b/benchmarks/multimodal/multimodal_eval.py
@@ -1,0 +1,254 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This is a simple script for MMLU benchmark for a trained checkpoint.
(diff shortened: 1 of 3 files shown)


benchmarks/multimodal/multimodal_eval.py
```diff
@@ -0,0 +1,289 @@
...
+  # Resize the image if specified. This helps simplify the llama4's tiling, so we have a fixed input size
+  if cfg.image_resize != -1:
+    pil_img = Image.fromarray(parsed_example.image_np)
+    pil_img = pil_img.resize((cfg.image_resize, cfg.image_resize))
+    parsed_example.image_np = np.asarray(pil_img.convert("RGB"))
+
+  return parsed_example
+
+
+def construct_prompt(parsed_dataset_example: ParsedDatasetExample, config, system_message: Optional[str] = None):
```
aireenmei, line 113: If we evaluate with a certain prompt/answer format, should we use the same for SFT? If so, we can put them in multimodal_utils, and use them in sft data processing pipeline.

benchmarks/multimodal/multimodal_eval.py
aireenmei, line 1: Could we reuse the existing eval loop in maxtext? Setting config.eval_interval > 0 will trigger the eval loop https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/train.py#L657, and it will reuse the data processing pipeline similar to train/sft (https://github.com/AI-Hypercomputer/maxtext/blob/53f83d261a5df2dafbe3606ae8407b1b88edb224/MaxText/input_pipeline/_hf_data_processing.py#L361). Also set config.eval_steps for now, as the data input returns error when the whole dataset run out. Looks like the main difference is in the metrics we use for eval, the existing eval uses loss, but we want  token difference. And the feature of exporting results to a csv file is a nice add-on. Seems like we can expand base on the current eval metrics, so user can choose eval metrics {loss, token_diff, etc}.

PR #1970, opened 2025-07-14
Qwen3 Dense 14b config

shralex:
This PR adds support for the 14b variant of Qwen3.

# Tests 

Tested locally, for example (HF -> MaxText):

--- MaxText model top 10 tokens ---
| Token ID   | Token                | Score      |
|------------|----------------------|------------|
| 1349       | read                 | 12.3125    |
| 4296       | cook                 | 11.5625    |
| 5821       | travel               | 11.5000    |
| 3270       | write                | 11.2500    |
| 1281       | make                 | 11.1875    |
| 1486       | play                 | 11.0000    |
| 1490       | see                  | 10.9375    |
| 8180       | eat                  | 10.8125    |
| 1855       | create               | 10.6875    |
| 990        | use                  | 10.6250    |

--- HF model top 10 tokens ---
| Token ID   | Token                | Score      |
|------------|----------------------|------------|
| 1349       | read                 | 12.3125    |
| 4296       | cook                 | 11.5625    |
| 5821       | travel               | 11.5000    |
| 3270       | write                | 11.2500    |
| 1281       | make                 | 11.1250    |
| 1486       | play                 | 11.0000    |
| 1490       | see                  | 10.9375    |
| 8180       | eat                  | 10.8125    |
| 1855       | create               | 10.6875    |
| 990        | use                  | 10.6250    |

--- Similarity Metrics of Top Tokens ---
| Metric                         | Value                |
|--------------------------------|----------------------|
| overlap_count                  | 10/10                |
| jaccard_similarity             | 1.0                  |
| rank_agreement_percentage      | 100.0                |

Average KL divergence per token (D_KL(P_golden || Q_model)): 0.007674

Max KL divergence for a single token in the set: 0.023109

# Checklist


MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -60,7 +60,7 @@
 from MaxText import pyconfig
 from MaxText import optimizers
 from MaxText.layers import models, quantizations
-from MaxText.train import save_checkpoint
```
gobbleturk, line 63: This seems out of place in this PR but I guess its a good fix. There are no unit tests to catch this I guess- the code is broken without this change?

MaxText/configs/models/qwen3-14b.yml
```diff
@@ -0,0 +1,38 @@
...
+decoder_block: "qwen3"
+
+normalization_layer_epsilon: 1.0e-6
+rope_max_timescale: 1000000
+
+use_qk_norm: True
+
+logits_via_embedding: False # different from 0.6 and 4B variants, "tie_word_embeddings": false
+normalize_embedding_logits: False
+enable_dropout: False # deterministic for testing
```
gagika: nit: we don't need the comment `# deterministic for testing`
I think qwen models don't have dropouts (as most of recent models) regardless testing or not.

gagika (approved):

MaxText/configs/models/qwen3-14b.yml
gobbleturk: dropout is also not a core property of the model, e.g. you can choose to train qwen with dropout or without, so dropout should not be specified in this model definition file

MaxText/utils/ckpt_conversion/to_maxtext.py
shralex: Right, I think @YixuanWang-99 's new conversion code is not currently exercised in end to end or unit tests. I agree that we should add testing.

gobbleturk (approved):

MaxText/configs/models/qwen3-14b.yml
shralex: Thanks. I removed the line and verified that logits still match.

PR #1971, opened 2025-07-15
Update MaxText's requirements.txt

kanglant:
This change shouldn't significantly affect MaxText's existing environment setup, and it is crucial for consistent python environment setups.

When updating MaxText's requirements.txt file, please keep these recommendations in mind:
- For each dependency, we recommend setting a lower bound to ensure the python seed env tool works well
- All source links must be pinned to a specific hash to guarantee reproducibility

We picked the lower bounds based on the generated maxtext requirements lock files in python seed env experiments, e.g., https://github.com/vam-google/python_seed_env/blob/python_3_10_gpu_test/maxtext_requirements_lock_3_10.txt. Please update the lower bounds if you found different versions are needed.

See more conversations in https://github.com/AI-Hypercomputer/maxtext/pull/1871.

# Checklist


SamuelMarks:
FWIW: See https://github.com/AI-Hypercomputer/maxtext/blob/7c270a3/requirements.txt from #1826

This works on uv or regular Python:
- 3.10
- 3.11
- 3.12

PR #1976, opened 2025-07-15
Implementation of the test for sharding of Linen and NNX models

hsuan-lun-chiang:
# Description

Add test scripts to generate and compare sharding of models implementing in Linen and NNX.

- Create ```sharding_dump.py``` to output sharding of model for a specific topology and slice.
- Create ```run_sharding_dump.py``` to automatically produce sharding JSON files for all combinations of models, topologies and slices.
- Create ```sharding_compare_test.py``` to compare model shardings implemented in Linen and NNX.

# Tests

Ran the ```run_sharding_dump``` script to automatically produce json file containing sharding information for Linen models. Currently tested with Llama-3.1-70B, Llama-3.1-405B across v6e, v5p, and v5e.

```
python -m MaxText.tests.run_sharding_dump
```

Ran the ```sharding_compare_test``` script to compare sharding of models. Currently we don't have NNX model so the test should be performed after migration to NNX.
```
python -m pytest MaxText/tests/sharding_compare_test.py
```

# Checklist

### DIFF
--- a/MaxText/tests/run_sharding_dump.py
+++ b/MaxText/tests/run_sharding_dump.py
@@ -1,0 +1,53 @@
+"""
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+""" Run script to dump sharding of various combination of model and topology. """
+
+
+from typing import Sequence
+from MaxText.tests.sharding_dump import TEST_CASES
+import os
+import subprocess
+from absl import app
+
+
+def run_single_dump(model_name: str, topology: str, num_slice: str) -> None:
+  """Generate sharding json file for one specific model, topology and slice."""
+  subprocess.run(
+      [
+          "python",
+          "-m",
+          "MaxText.tests.sharding_dump",
+          "MaxText/configs/base.yml",
+          f"compile_topology={topology}",
+          f"compile_topology_num_slices={num_slice}",
+          f"model_name={model_name}",
+      ],
+      check=True,
+  )
+
+
+def main(argv: Sequence[str]) -> None:
+  """Generate sharding json files for every combination of model, topology and slices."""
+  for model_name, topology, num_slice in TEST_CASES:
+    json_path = f"sharding_info/{model_name}/{topology}/slice_{num_slice}/named_shardings.json"
+    if os.path.exists(json_path):
+      continue
+    run_single_dump(model_name, topology, num_slice)
+
+
+if __name__ == "__main__":
+  app.run(main)
--- a/MaxText/tests/sharding_compare_test.py
+++ b/MaxText/tests/sharding_compare_test.py
@@ -1,0 +1,111 @@
+"""
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Compare expected sharding of models with actual sharding of models."""
+
+
+import hashlib
+from MaxText.train_compile import get_shaped_inputs, get_topology_mesh, validate_config
+from MaxText.tests.sharding_dump import named_shardings_to_json, load_named_sharding_json, TEST_CASES
+from MaxText import pyconfig
+import pytest
+import os
+import json
+
+
+def compute_checksum(d: dict) -> str:
+  """Compute a checksum (SHA256) of a dictionary."""
+  # Serialize the dictionary into a JSON string (ensuring consistent ordering of keys)
+  json_str = json.dumps(d, sort_keys=True)
+
+  # Compute the SHA256 checksum of the serialized string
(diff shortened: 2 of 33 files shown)


MaxText/tests/run_sharding_dump.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Sequence
+from MaxText.tests.sharding_dump import TEST_CASES
+import os
+import subprocess
+from absl import app
+
+
+def run_single_test(model_name: str, topology: str, num_slice: str) -> None:
+  """Run test for one specific model, topology and slice."""
+  subprocess.run(
```
gobbleturk, line 29: we would prefer to avoid subprocess if possible, we have seen strange race conditions in the past using it in tests. If this is meant to be run as a unit test, can you instead import main of sharding dump and run that as a regular python command instead of subprocess?

hsuan-lun-chiang: Hi, thanks for the suggestion. We initially tried calling the function directly, but encountered an OOM issue due to a memory leak. Despite using ```jax.clear_caches()``` and ```gc.collect()```, we haven't been able to fully release the memory after retrieving the mesh and abstract states of the model. Using a subprocess helps isolate memory usage and avoid these issues for now.

hsuan-lun-chiang:
> Is there a script to generate these golden shardings? What if a user makes a change where we expect the shardings the change (e.g. we need to update the golden shardings?)

Sorry the comments in the script were a bit misleading, and I've updated them for clarity. 
To answer your question: 
> Is there a script to generate these golden shardings?

```run_sharding_dump.py``` (which calls functions from ```sharding_dump.py```) is used to generate the golden shardings. For testing and comparison between Linen and NNX models, we use ```sharding_compare_test.py```.

> What if a user makes a change where we expect the shardings the change (e.g. we need to update the golden shardings?)

In that case, we need to update the golden shardings by running the script to generate them again.

File: MaxText/tests/sharding_compare_test.py [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+
+  config = pyconfig.initialize(params)
+  validate_config(config)
+
+  topology_mesh = get_topology_mesh(config)
+  _, _, state_mesh_shardings, _ = get_shaped_inputs(topology_mesh, config)
+  json = named_shardings_to_json(state_mesh_shardings)
+  expected_json = load_named_sharding_json(json_path)
+
+  assert compare_named_sharding_jsons(expected_json, f"linen_{model_name}", json, f"nnx_{model_name}"), True
```
shralex: can we change linen_{model_name} to expected_{model_name} and nnx_{model_name} to actual_{model_name}  ? I expect this test will be useful even after nnx conversion. Similarly I suggest to change the comment at the top of the file

hsuan-lun-chiang: Sure, I've updated the docstrings and model names.

MaxText/tests/run_sharding_dump.py
gobbleturk: Not sure we can accept a test that uses subprocess, we have seen crazy race conditions in the past where unrelated tests can fail when any test uses subprocess

gobbleturk:
I'm hesitant to merge this, this will be a burden to maintain, any sharding/mesh change will require updating these goldens

hsuan-lun-chiang:
> I'm hesitant to merge this, this will be a burden to maintain, any sharding/mesh change will require updating these goldens

Would it be possible to trigger this process using a GitHub Action? Ideally, it would automatically regenerate the golden sharding and upload it to a GCS bucket whenever a pull request is merged.

bvandermoon:
> > I'm hesitant to merge this, this will be a burden to maintain, any sharding/mesh change will require updating these goldens
> 
> Would it be possible to trigger this process using a GitHub Action? Ideally, it would automatically regenerate the golden sharding and upload it to a GCS bucket whenever a pull request is merged.

@gobbleturk how often do these get changed? @richjames0 any thoughts on this?

bvandermoon:
> > > I'm hesitant to merge this, this will be a burden to maintain, any sharding/mesh change will require updating these goldens
> > 
> > 
> > Would it be possible to trigger this process using a GitHub Action? Ideally, it would automatically regenerate the golden sharding and upload it to a GCS bucket whenever a pull request is merged.
> 
> @gobbleturk how often do these get changed? @richjames0 any thoughts on this?

@gobbleturk thinking about this a bit more, don't we want the test to fail if the sharding/mesh gets changed? It shouldn't be too difficult to update the test also. It should confirm things are working properly IMO. Thoughts?

MaxText/tests/run_sharding_dump.py
xibinliu: @bvandermoon and @gobbleturk probably it's okay if we use 'subprocess' in unit tests? Can we add a TODO to further working on the OOM issue while proceeding with this PR?

> encountered an OOM issue due to a memory leak

@hsuan-lun-chiang  Can you create a bug for this, with the re-producing steps?

MaxText/tests/sharding_compare_test.py
```diff
@@ -0,0 +1,91 @@
...
+      print(f"\nMesh mismatch at '{key}':")
+      print(f"  mesh1: {mesh1}")
+      print(f"  mesh2: {mesh2}")
+
+    if spec1 != spec2:
+      print(f"\nPartitionSpec mismatch at '{key}':")
+      print(f"  spec1: {spec1}")
+      print(f"  spec2: {spec2}")
+
+  return not only_in_1 and not only_in_2 and all(json1[k] == json2[k] for k in shared_keys)
```
bvandermoon, line 78: Would a checksum work for the actual diff check? Then you could run this to return the diff if they don't match

bvandermoon (approved):
LGTM other than the subprocess comment. Thank you @hsuan-lun-chiang for setting this up. Could you please sort out the subprocess piece with @gobbleturk? We would really want to remove that if possible

MaxText/tests/sharding_compare_test.py
hsuan-lun-chiang: Good idea! I've implemented this.

bvandermoon:
> LGTM other than the subprocess comment. Thank you @hsuan-lun-chiang for setting this up. Could you please sort out the subprocess piece with @gobbleturk? We would really want to remove that if possible

Discussed with Matt offline. We will merge this with subprocess and see how it goes

hsuan-lun-chiang:
> Discussed with Matt offline. We will merge this with subprocess and see how it goes

Understood, thanks for the help.

PR #1983, opened 2025-07-17
Support gemma3 full model bidirectional ckpt conversion.

YixuanWang-99:
# Description

The PR supports the Gemma3 multimodal (vision + text) checkpoint Bidirectional conversion between Maxtext and huggingface. 

- Added parameter mappings, layerwise hook functions, and HF shapes/configs for gemma3 conversion
- Fixed the OOM error by adding Jax config 
- Improve the to_huggingface.py to skip local store when publishing ckpt on huggingface hub, greatly save the disk usage 

# Tests

Tests are done by running decode of the converted checkpoint. 
[Original decode output](https://screenshot.googleplex.com/3mqX2nzXDoU2XjL)
[Converted Maxtext ckpt output](https://screenshot.googleplex.com/898FH54gj76MQVo)
[Converted Huggingface ckpt output](https://screenshot.googleplex.com/4Pc9CykGhiVmYS5)

Converted Huggingface ckpt are published on: https://huggingface.co/yixuan-99/gemma3-4b-it/tree/main

# Checklist


MaxText/utils/ckpt_conversion/utils/utils.py
```diff
@@ -417,40 +441,47 @@ def save_model_files(
...
-            if remove_local_copy:
-              os.remove(local_file_path)
-              max_logging.log(f"   Removed local copy: {local_file_path}")
+        max_logging.log(f"    Tokenizer files saved locally: {saved_tokenizer_files}")
+      files_to_upload = [os.path.join(current_save_path, f) for f in os.listdir(current_save_path)]
+
+      if output_dir.startswith("gs://"):
+        for local_file_path in files_to_upload:
+          if not os.path.exists(local_file_path):
+            max_logging.log(f"   Warning: Tokenizer file {local_file_path} not found locally. Skipping upload to GCS.")
```
hengtaoguo, line 457: IIUC, this code support two ways to save the HF ckpt: (1) upload to GCS bucket and (2) directly publish on HF.

For GCS option, does it still require saving all the files to local directory first? Is it possible to also do the memory-to-GCS in the future?

hengtaoguo (approved):
Can you add some instructions running the Gemma3 full model conversion? Thanks!

Also add a bug for improving the forward_pass_logits_checker, to include multimodal correctness check.

MaxText/utils/ckpt_conversion/utils/utils.py
YixuanWang-99: At this point, for both the GCS and and HF uploading, the .safetensor files are the most significant portion of disk usage, and it is optimized to be uploaded from memory to remote directory in this PR

MaxText/utils/ckpt_conversion/utils/utils.py
```diff
@@ -189,29 +191,33 @@ def save_config_file(
...
+            repo_type="model",
+        )
+      max_logging.log(f"  Successfully uploaded {file_name} to HF repo: {repo_id}")
+    else:
+      # local storage
+      actual_local_file_path = os.path.join(local_path_to_save_to, file_name)
+      config.to_json_file(actual_local_file_path)
+      max_logging.log(f"   Saved {file_name} to {actual_local_file_path}")
+      # upload
+      if output_dir_final.startswith("gs://"):
```
shralex, line 216: can we upload directly to GCS here like in the case of checkpoint files ?

shralex (commented):
Thanks Yixuan! General comment -- is it possible to make this implementation more modular ? Right now, for every type of file there is a separate implementation of uploading to HF, uploading to GCS and to local disk. Can we create an upload method and then use them ? Also for GCS for checkpoint files the implementation is optimized but for other files its storing things locally, while for HF its always uploading directly.

MaxText/utils/ckpt_conversion/utils/utils.py
YixuanWang-99: We need to upload 3 types of files, 1) tokenizer files, this must be saved as local file then upload it, this is because of the nature of save_pretrained() and from_pretrained() function calls from the transformers library. 2) Json files; 3) *.safetensors files. Now I have modified the code to upload the JSON files and safetensors from memory to GCS/HF directly from memory. 

Specifically, for GCS upload, uploading JSON files and safetensors files requires different implementation (upload_json_to_gcs(), upload_state_dict_to_gcs()). For HF upload, they are similar implementation which are encapsulated in function upload_file_to_hf().

PR #1985, opened 2025-07-17
Change shardy to default + make shardy configurable

gobbleturk:
# Description

Change the default XLA backend to shardy. Shardy is an alternative to GSPMD which should give higher performance in the majority of cases. There are no jax/user level changes needed to support shardy other than setting the jax config to use shardy `jax.config.update("jax_use_shardy_partitioner", True)`

Unfortunately the above line was added for all trainers (train.py, sft, etc), it is difficult to refactor into a shared initialize method until we clean up how we call `jax.distributed.initialize()` (tracked in b/432512350)

 
# Tests

Ran on some models, saw no regressions (actually identical performance, but can confirm shardy is used with verbose libtpu logs like `Using Shardy for XLA SPMD propagation`

| Model | Hardware | Shardy? | MFU | Peak HBM (GB) |
|---|---|---|---|---|
| **Llama405B** | 2x v6e-256 | Yes | 39% | 25.8 |
| **Llama405B** | 2x v6e-256 | No | 39% | 25.8 |
| **DeepSeekV3** | 1x v5p-512 | Yes | 29% | 81.6 |
| **DeepSeekV3** | 1x v5p-512 | No | 29% | 81.6 |

Also added unit test for the (new) non-default case so both cases stay protected

# Checklist

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -701,6 +701,7 @@ def initialize(argv: Sequence[str]) -> Tuple[pyconfig.HyperParameters, Any, Any]
   # TODO: mazumdera@ : ensure missing mandatory fields in base.yml are filled in in argv,
   # or fill in here
   config = pyconfig.initialize(argv)
+  jax.config.update("jax_use_shardy_partitioner", config.shardy)
   max_utils.print_system_information()
   validate_train_config(config)
   os.environ["TFDS_DATA_DIR"] = config.dataset_path or ""
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -683,6 +683,8 @@ allow_split_physical_axes: False
 # Apply transformations to the mesh to optimize for TPU v6e
 optimize_mesh_for_tpu_v6e: False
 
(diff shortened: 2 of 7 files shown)


MaxText/train.py
```diff
@@ -701,6 +701,7 @@ def initialize(argv: Sequence[str]) -> Tuple[pyconfig.HyperParameters, Any, Any]
   # TODO: mazumdera@ : ensure missing mandatory fields in base.yml are filled in in argv,
   # or fill in here
   config = pyconfig.initialize(argv)
+  jax.config.update("jax_use_shardy_partitioner", config.shardy)
```
bvandermoon, line 704: Should this also be in `elastic_train.py`?

bvandermoon (approved):

MaxText/train.py
gobbleturk: Yup, good catch, thank you!

lukebaumann (approved):

khatwanimohit (approved):

PR #2013, opened 2025-07-23
Enable TP transpose for Megablox implementation

RissyRan:
# Description

Enable TP transpose for Megablox implementation

# Tests

* unit test
* end-to-end tests

# Checklist


MaxText/configs/base.yml
```diff
@@ -342,6 +342,7 @@ logical_axis_rules: [
                       ['embed_no_exp', ['fsdp', 'sequence', 'tensor_transpose', 'context']],
                       ['embed_no_exp', ['fsdp', 'fsdp_transpose', 'sequence', 'context']],
                       ['embed_no_exp', ['fsdp', 'sequence', 'context']],
+                      ['embed_tensor_transpose', ['tensor_transpose']],
```
gobbleturk, line 345: I'm not a fan of having this special rule, but I think I see how it simplifies the implementation in moe.py

An alternative is to use the regular "embed" axis and then we would need an explicit FSDP all gather inside of the shard map (as opposed to an FSDP replicated sharding constraint which we currently have). This might be preferable but I am worried about collisions... your new rule for sure has no collisions =D

MaxText/tests/moe_test.py
```diff
@@ -453,6 +453,51 @@ def test_megablox_expert_parallelism(self):
...
+        dtype=cfg.dtype,
+    )
+
+    devices_array = maxtext_utils.create_device_mesh(cfg)
+    mesh = Mesh(devices_array, cfg.mesh_axes)
+    with nn_partitioning.axis_rules(cfg.logical_axis_rules):
+      variables, _ = self.get_expected_output(rng_model, hidden_states, cfg)
+      tp_transpose_output, _ = self.get_moe_output(variables, hidden_states, cfg, mesh)
+      tp_output, _ = self.get_moe_output(variables, hidden_states, cfg2, mesh)
+      self.assertTrue(jax.numpy.allclose(tp_output, tp_transpose_output, rtol=1e-05, atol=1e-05, equal_nan=False))
```
gobbleturk, line 499: I'm amazed this passes with such strict tolerances in bfloat16, it wouldn't surprise me if we had to relax these in the future

gobbleturk (approved):
Thanks Ran!

LGTM, I'm not super happy with the singleton logical axis rule but I think thats on us to eventually refactor our sharding at some point to avoid collision.

MaxText/configs/base.yml
RissyRan: Yeah, good point. I met some issue regarding `context` collisions if I followed up the same format as `embed_no_exp` as bellow. So instead of using unknown sharding configs (i.e. with `fsdp_transpose`  `sequence` etc), I just simplify it to know what I exactly want to do.

```
                      ['embed_no_exp_fsdp', ['fsdp_transpose', 'sequence', 'tensor_transpose', 'context']],
                      ['embed_no_exp_fsdp', ['sequence', 'tensor_transpose', 'context']],
                      ['embed_no_exp_fsdp', ['fsdp_transpose', 'sequence', 'context']],
                      ['embed_no_exp_fsdp', ['sequence', 'context']],
```

gagika (approved):

PR #2028, opened 2025-07-25
Enable gemma3 multimodal SFT

aireenmei:
# Description
This PR supports gemma3 SFT on chartQA dataset, the main changes in multimodal_utils are migrating functions we want to run on CPU from jnp to np. Note that the following existing limitations on the gemma3 support won't be addressed in this PR, but I worked around it and added TODO comments for the future:
* vision encoder does not support scan yet, so the scan_layers flag only applies to text model
* vision encoder is not sharded.

b/433982967

# Tests

tested gemma3-4b on v5p with scan_layer= {true, false}

# Checklist

### DIFF
--- a/MaxText/multimodal_utils.py
+++ b/MaxText/multimodal_utils.py
@@ -84,12 +84,14 @@ class PreprocessorOutput:
 
 
 def resize_image(image, model_name):
+  """resize image if needed"""
   image_sizes = {
-    "gemma3": (GEMMA_DEFAULT_IMAGE_SIZE, GEMMA_DEFAULT_IMAGE_SIZE),
-    "llama4": (LLAMA4_TILE_SIZE, LLAMA4_TILE_SIZE),
+      #resize for vanilla llama4 support, will be removed once size support is added to multimodal llama4
+      "llama4": (LLAMA4_TILE_SIZE, LLAMA4_TILE_SIZE),
   }
-  target_size = image_sizes[model_name.split('-')[0]]
-  if target_size:
+  model_prefix = model_name.split("-")[0]
+  if model_prefix in image_sizes:
+    target_size = image_sizes[model_prefix]
     image = image.resize(target_size)
   return image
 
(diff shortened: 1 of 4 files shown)


MaxText/configs/models/gemma3-4b.yml
```diff
@@ -30,3 +30,4 @@ use_post_attn_norm: true
 use_post_ffw_norm: true
 local_rope_max_timescale: 10_000
 rope_max_timescale: 1_000_000
+image_placeholder: "<start_of_image>"
```
hengtaoguo: This is nice, I almost forgot we have this flag for model-specific image placeholder. We can use flag this during eval too. Thanks for adding it to the model's yml.

MaxText/multimodal_utils.py
```diff
@@ -589,7 +592,7 @@ def get_num_tokens_for_this_image(this_aspect_ratio, num_patches_per_chunk):
 
 
 def add_extra_tokens_for_images_gemma3(
-    tokens: np.ndarray | jnp.ndarray,
+    tokens: np.ndarray | List,
```
hengtaoguo, line 597: This is nice to have! Could you remind why we need to support List as an input type?

MaxText/configs/models/gemma3-4b.yml
aireenmei: uh, now I see that this flag may cause some confusion, and it may be misleading adding it to the model yml, because we already have it hardcoded as GEMMA_IMAGE_PLACEHOLDER_IN_PROMPT. I added the flag image_placeholder with the intention of allowing custom image_placeholder in case they have a different one in the dataset. It will be replaced by model specific placeholder https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/multimodal_utils.py#L415. So we shouldn't tie it to the model's yml. How about we remove it here, and explain it in the coming docs?

MaxText/multimodal_utils.py
aireenmei: The output after HF tokenizer is in list format. I added "if not isinstance(tokens, np.ndarray)" for conversion later in this function

aireenmei:
> Just to confirm, Gemma3 doesn't require additional image resizing, because the pre_process_gemma3_image will force to resize GEMMA_DEFAULT_IMAGE_SIZE = 896 right?

Thanks for pointing out. I remove gemma3 from resize_image function and added comments.

gagika (approved):

PR #2033, opened 2025-07-28
Orchestration Agent

wei879-100:
# Description

This PR introduces an automated orchestration agent designed to analyze PyTorch codebase files and generate a hierarchical dependency plan. This plan dictates the precise order in which components should be converted to JAX. This represents a significant change from previous manual or ad-hoc approaches. Previously, developers had to manually trace dependencies between files and code blocks, a process that is time-consuming for complex model conversions. This agent accelerates the conversion timeline and improves the reliability of the resulting JAX model.

Large codebases like transformers feature intricate dependency webs. A single model file might depend on dozens of other files for configuration, utilities, and base model layers. Within each file, functions and classes call each other in a specific sequence. This agent solves the problem by programmatically building a complete dependency graph and topologically sorting it to find the optimal path.

The agent operates in a two-phase process, orchestrated by `ochastrationAgent.py`:

**Inter-File Dependency Analysis:** The agent first calls `GetFilesInHierarchicalOrder.py`. Starting from a user-specified entry point (e.g., a model file URL), this script recursively scans all import statements. It intelligently resolves both absolute and relative imports within the context of the GitHub repository's base path, effectively mapping out the entire file-level dependency graph. It then performs a topological sort on this graph to produce an ordered list of files, from foundational utilities to the final model file.

**Intra-File Component Analysis:** Once the file order is established, the agent processes each file one by one using the `SplitPythonFile.py` script. This script parses the source code of a single file into an Abstract Syntax Tree (AST). It identifies all top-level definitions (classes, functions, and global variables) and builds a "micro" dependency graph based on which definitions reference others within the same file. Crucially, it uses Tarjan's algorithm to detect and group any circular dependencies (Strongly Connected Components) into single, inseparable components. The output is a list of code blocks for that file, sorted in the correct order for execution.

The `ochastrationAgent.py` script combines these two phases, creating a comprehensive, hierarchical plan. It logs the file-level order and then, for each file, details the standalone and dependent internal components, saving the complete plan to `FilesWithComponents.txt`.

# Tests
The functionality of the orchestration agent was tested by running it against a complex model file within the Hugging Face Transformers repository. The agent successfully generated a dependency-ordered list of all required files and their internal components.

To reproduce the test, run the main orchestration script from the command line:

```
python ochastrationAgent.py \
    --base-path "https://github.com/huggingface/transformers/blob/main/src/" \
    --entry-file-path "https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py"
```
You can change the base-path and entry-file-path with different URLs.

**Verification Steps:**

1. Confirm that the script completes without errors.
2. Inspect the output file `AllFiles.json`. This file contains the sorted list of dependent file paths relative to the base path.
3. Inspect the primary output file `FilesWithComponents.txt`. Verify that it lists the standalone files first, followed by the dependent files in their sorted order. For each file, it breaks down the internal components (classes, functions) into "Standalone Modules" and "Dependent Modules."

The dependency relationships listed for both files and components logically match the source code. For example, a file like modeling_llama.py should be listed after its dependencies (e.g., configuration_llama.py), and a class like LlamaModel should be listed after the base classes it inherits from.

Sample output [here](https://paste.googleplex.com/5788005834686464).

# Checklist


MaxText/experimental/agent/ochestration_agent/SplitPythonFile.py
```diff
@@ -0,0 +1,342 @@
...
+        nodes = list(self.definitions.keys())
+        disc = {node: -1 for node in nodes}
+        low = {node: -1 for node in nodes}
+        on_stack = {node: False for node in nodes}
+        stack = []
+        self.time = 0
+        sccs = []
+
+        def tarjan(u):
+            disc[u] = low[u] = self.time
```
gobbleturk: I'm not familiar with tarjan but perhaps replace self.time with self.step?

MaxText/experimental/agent/ochestration_agent/Utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Args:
+        blob_url (str): The URL of the GitHub blob.
+
+    Returns:
+        str: The URL of the raw content of the GitHub blob.
+    """
+    parsed = urlparse(blob_url)
+    if 'github.com' not in parsed.netloc or '/blob/' not in parsed.path:
+        raise None # Not a valid blob URL
```
gobbleturk: should we raise an informative ValueError here?

MaxText/experimental/agent/ochestration_agent/Utils.py
```diff
@@ -0,0 +1,133 @@
...
+    
+    parts = parsed.path.split('/')
+    if len(parts) < 5:
+        raise None # Not a valid blob URL
```
gobbleturk: also ValueError here?

YixuanWang-99:
Can you revise the example output as a link instead of a downloading file?

MaxText/experimental/agent/ochestration_agent/SplitPythonFile.py
wei879-100: Great suggestion, I changed it.

MaxText/experimental/agent/ochestration_agent/Utils.py
wei879-100: Yes, I added here

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def HaveModule(target_name, file_url):
+    """
+    Checks if a Python module at a given URL contains a specific function, class, or variable.
+    This version reads content from the URL instead of a local file.
+    """
+    flag, content = get_github_file_content(file_url)
+    if not flag:
+        return False  # Fail silently if the file can't be fetched
```
gobbleturk: Why fail silently here? Some URLs we expect don't point to real files or?

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+            for alias in node.names:
+                if alias.asname == target_name or alias.name == target_name:
+                    module = node.module or ""
+                    level = node.level
+                    full_module = "." * level + module if level > 0 else module
+                    return ("ImportFrom", full_module)
+    return False
+
+
+def Resolve_complex_import(module_path_base_url, importPackage, base_url, current_dir_url, Try=0, Message=""):
```
gobbleturk: lowercase variable names?

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+logging.basicConfig(level=logging.INFO,  # You can use DEBUG, INFO, WARNING, ERROR, CRITICAL
+                    format='%(asctime)s - %(name)s.%(funcName)s - %(levelname)s - %(message)s',
+                    datefmt='%H:%M:%S')
+logger=logging.getLogger(__name__)
+
+
+def HaveModule(target_name, file_url):
+    """
+    Checks if a Python module at a given URL contains a specific function, class, or variable.
+    This version reads content from the URL instead of a local file.
```
gobbleturk: Can you include the possible return values here? Its surprising this method doesn't always return a boolean

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+            # The package exists, but the import is not in __init__. It could be a submodule.
+            potential_file_in_pkg_url = url_join(module_path_base_url, f"{importPackage}.py")
+            if check_github_file_exists(potential_file_in_pkg_url)[0] and potential_file_in_pkg_url.startswith(base_url):
+                return potential_file_in_pkg_url
+    return None
+
+def resolve_import_path(importer_url, module_name, level, base_url, importPackage=None):
+    """
+    Resolves the absolute URL of an imported module.
+    Follows the logic of your original function but with URLs.
```
gobbleturk: I don't follow this comment - whose original function?

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
+import ast,json
+from collections import deque
+from Utils import find_cycle, check_github_file_exists, get_github_file_content, url_join
+import argparse,logging
+
```
gobbleturk, line 39: Can you include top level comments here what this file does, and an example invocation?

MaxText/experimental/agent/orchestration_agent/SplitPythonFile.py
```diff
@@ -0,0 +1,343 @@
+import ast
+import os.path
+import sys
+import json
+from collections import defaultdict, deque
+import argparse
+from Utils import get_github_file_content
+
```
gobbleturk, line 36: Can you also include a top level comment here what this file does and example invocations?

MaxText/experimental/agent/orchestration_agent/orchestrationAgent.py
```diff
@@ -0,0 +1,116 @@
+import json
+import os.path
+from Utils import check_github_file_exists
+import argparse,logging
+from GetFilesInHierarchicalOrder import get_dependency_sorted_files
+from SplitPythonFile import getModulesInOrder
+
```
gobbleturk, line 39: Can you include top level comments here what this file is and example invocations?

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+# Set up basic configuration
+logging.basicConfig(level=logging.INFO,  # You can use DEBUG, INFO, WARNING, ERROR, CRITICAL
+                    format='%(asctime)s - %(name)s.%(funcName)s - %(levelname)s - %(message)s',
+                    datefmt='%H:%M:%S')
+logger=logging.getLogger(__name__)
+
+
+def HaveModule(target_name, file_url):
```
hengtaoguo: Let's uniform our function naming convention. Please use snake_case (e.g., have_module, find_file_dependencies) for all function names, ensuring all letters are lowercase.

MaxText/experimental/agent/orchestration_agent/Utils.py
```diff
@@ -0,0 +1,133 @@
...
+    Returns:
+        str: The URL of the raw content of the GitHub blob.
+    """
+    parsed = urlparse(blob_url)
+    if 'github.com' not in parsed.netloc or '/blob/' not in parsed.path:
+        raise ValueError(f"Invalid GitHub blob URL: {blob_url}") # Not a valid blob URL
+    
+    parts = parsed.path.split('/')
+    if len(parts) < 5:
+        raise ValueError(f"Imcomplete GitHub blob URL: {blob_url}") # Not a valid blob URL
```
hengtaoguo: nit: "Incomplete"

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
hengtaoguo, line 1: Can you make sure to run `bash code_style.sh` before submission? You can exclude the style changes from files other than your own work.

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+                    level = node.level
+                    full_module = "." * level + module if level > 0 else module
+                    return ("ImportFrom", full_module)
+    return False
+
+
+def Resolve_complex_import(module_path_base_url, importPackage, base_url, current_dir_url, Try=0, Message=""):
+    """
+    Tries to resolve a complex import (e.g., from a package) using URLs,
+    replicating the user's original recursive logic.
```
hengtaoguo: Is my understanding correct? If your code does from mypackage.submodule import myfunction, the "resolved URL" would be the GitHub link to the .py file that actually contains the definition of myfunction.

If so, would be nice to include an example in the doc string. This also applies for other complex functions.

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+        return None
+    # Check for a direct .py file containing the definition
+    potential_py_url = f"{module_path_base_url}.py"
+    if check_github_file_exists(potential_py_url)[0] and HaveModule(importPackage, potential_py_url) == True:
+        return potential_py_url
+
+    # Check for a package (directory with __init__.py)
+    potential_pkg_init_url = url_join(module_path_base_url, '__init__.py')
+    if check_github_file_exists(potential_pkg_init_url)[0] and potential_pkg_init_url.startswith(base_url):
+        isHaveModule = HaveModule(importPackage, potential_pkg_init_url)
```
hengtaoguo: Same as the variables names, let's use snake_case for consistency, such as has_module.

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+    if check_github_file_exists(potential_url_py)[0]:
+        return potential_url_py
+
+    # If it's not a direct file, it might be a complex package import
+    return Resolve_complex_import(module_path_base_url, importPackage, base_url, current_dir_url)
+
+
+def find_file_dependencies(file_path_url, base_path_url, exclude_conditional_imports=True):
+    """
+    Parses a Python file from a URL to find all its internal dependencies.
```
hengtaoguo: An naive question: When a dependency file has nested dependencies, how does your program determine when to stop resolving the dependency chain? Or does it only include the current file's dependencies without nesting further?

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Exclude imports inside functions/classes (use --no-exclude-conditional-imports to disable)",
+    )
+    args = parser.parse_args()
+    if not args.entry_file_path.startswith(args.base_path):
+        raise ValueError("Entry file path must be inside the base path.")
+    return args
+
+def saveResultsInFile(sorted_files, dependencies,args,outFile="FileOrder.txt"):
```
hengtaoguo: nit: doc string for every function.

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
```diff
@@ -0,0 +1,303 @@
...
+    with open("AllFIles.json","w") as f:
```
hengtaoguo: maybe use all_files.json?

hengtaoguo (approved):
Thanks for the hard work! +1 to Matt on adding example command to the beginning of every agent you submitted.

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
wei879-100: Great suggestion! changed them

MaxText/experimental/agent/orchestration_agent/GetFilesInHierarchicalOrder.py
wei879-100: Excellent question! This is designed to resolve the entire nested dependency chain, not just the immediate imports of the entry file. It builds a complete map of all interconnected files within the project's scope before determining the final order.

Here’s a step-by-step breakdown of how it achieves this:

1. Graph Traversal: The process starts with the single entry file. It uses a queue (specifically, a deque named `files_to_process`) to manage which files to analyze next. This is essentially a BFS algorithm over the file dependency graph.

2. Discovering Nested Dependencies: It takes a file from the front of the queue and parses that file to find all of its direct dependencies. For each dependency it finds, if that file hasn't been analyzed before, it's added to the back of the queue to be processed in a future step.

**When to Stop:**  Continues until the queue is empty. The queue only becomes empty when every single reachable file (and all of its dependencies, and so on) has been processed.

So, it doesn't stop at the first level. It systematically follows every valid import path until it has discovered every file that is part of the dependency web. Only after this complete graph is built does it perform the topological sort to give you the final hierarchical order.

I've added some comments to the `get_dependency_sorted_files function`.

SamuelMarks:
Thanks.

FWIW: There was discussion about this yesterday, particularly in the maintenance burden and scope-creep to add this entire patch into MaxText. My recommendation was that a: large portion of the logic can be replaced with [griffe](https://github.com/mkdocstrings/griffe) (either via the CLI or SDK).

PR #2034, opened 2025-07-28
Add training and data loading hooks for SFT

SurbhiJainUSC:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

Local E2E testing

# Checklist


MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+    metrics = {
+        "scalar": {
+            "eval/total_loss": self.eval_metadata["total_loss"],
+            "eval/avg_loss": avg_loss,
+            "eval/total_weights": self.eval_metadata["total_weights"],
+        }
+    }
+
+    self.metric_logger.write_metrics(metrics, train_ctx.train_steps, is_training=False)
+    self.eval_metadata.clear()
```
gagika, line 144: is it intentional to clear at the end of every eval step?

That way: self.eval_metadata["total_loss"] and self.eval_metadata["eval_step_count"] will never accumulate beyond the current step

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+  def on_train_step_end(self, train_ctx: peft_trainer.PeftTrainer, train_loss: float):
+    """Called at the end of training step."""
+    step_time_delta = datetime.datetime.now() - self.train_metadata["step_start_time"]
+    self.train_metadata["total_loss"] += train_loss
+
+    # We check against `train_ctx.train_steps - 1` because this hook is called
+    # after `train_ctx.train_steps` has been incremented for the current
+    # step.
+    if self.metadata["first_train_step"] == train_ctx.train_steps - 1:
+      max_utils.print_mem_stats("After params initialized")
```
gagika, line 109: why do you need this print?

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+
+    if train_ctx.data_hooks.train_batch is not None:
+      self.train_metadata["total_weights"] += jnp.sum(train_ctx.data_hooks.train_batch["targets_segmentation"] != 0)
+    self.train_metadata["step_start_time"] = datetime.datetime.now()
+
+  @override
+  def on_train_step_end(self, train_ctx: peft_trainer.PeftTrainer, train_loss: float):
+    """Called at the end of training step."""
+    step_time_delta = datetime.datetime.now() - self.train_metadata["step_start_time"]
+    self.train_metadata["total_loss"] += train_loss
```
gagika: is it storing loss per step or is accumulating, if it's accumulating then we don't want self.train_metadata.clear() at every step, if it's for each step, perhaps we can have `self.train_metadata["total_loss"] = train_loss`

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+from MaxText.utils import gcs_utils
+from MaxText import exceptions
+from MaxText import maxtext_utils
+from MaxText import max_logging
+from MaxText import max_utils
+from MaxText.data_loader import DataLoader
+from MaxText.input_pipeline.input_pipeline_interface import create_data_iterator
+from tunix.sft import peft_trainer
+from tunix.sft.hooks import DataHooks, TrainingHooks
+from typing_extensions import override
```
gagika, line 33: is MaxText depending on Tunix and Tunix on MaxText? doesn't this cause a circular dependancy?

richjames0: MaxText invokes Tunix which then has callbacks into MaxText but it's factored reasonably enough that there isn't a hard circular dependency per se

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+          delete_local_after=self.config.dump_hlo_delete_local_after,
+          all_host_upload=self.config.dump_hlo_upload_all,
+      )
+
+    self.metadata["first_train_step"] = train_ctx.train_steps
+
+  @override
+  def on_train_end(self, train_ctx: peft_trainer.PeftTrainer):  # pylint: disable=unused-argument
+    """Called at the end of training."""
+    self.metadata.clear()
```
richjames0: Is this correct, as __init__ sets it to something non-empty?

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+      self.metric_logger.flush_metrics_and_cleanup()
+
+  @override
+  def on_train_step_start(self, train_ctx: peft_trainer.PeftTrainer):
+    """Called at the beginning of a training step."""
+    if self.config.enable_goodput_recording:
+      record_goodput(self.goodput_recorder, f"record_{GoodputEvent.STEP.value}_start_time", train_ctx.train_steps)
+
+    if train_ctx.data_hooks.train_batch is not None:
+      self.train_metadata["total_weights"] += jnp.sum(train_ctx.data_hooks.train_batch["targets_segmentation"] != 0)
```
richjames0: this is a tiny bit subtle (i initially wondered whether the != 0 was necessary, although i now see it obviously is) so maybe a comment is warranted

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,177 @@
...
+          self.config.dump_hlo_gcs_dir,
+          module_name=self.config.dump_hlo_module_name,
+          delete_local_after=self.config.dump_hlo_delete_local_after,
+          all_host_upload=self.config.dump_hlo_upload_all,
+      )
+
+    self.metadata["first_train_step"] = train_ctx.train_steps
+
+  @override
+  def on_train_end(self, train_ctx: peft_trainer.PeftTrainer):  # pylint: disable=unused-argument
```
richjames0, line 76: Use your judgement on this but I have a vague feeling that we ought to assert that `.*_start` is called before `.*_end` to catch bugs from the Tunix side which would be quite nasty if hit. I wouldn't go as far as creating new state to check but I think here there's enough state already that you could assert on. But your call, just an itch I'm feeling

MaxText/sft/hooks.py
SurbhiJainUSC: MaxText depends on Tunix, Tunix doesn't. So, there won't be any circular dependency.

MaxText/sft/hooks.py
SurbhiJainUSC: Yes, it is storing loss per step. I will change it to `self.train_metadata["total_loss"] = train_loss`

MaxText/sft/hooks.py
SurbhiJainUSC: This is similar to the current state of sft_trainer: https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/sft_trainer.py#L139. If this is not useful, I can remove this.

MaxText/sft/hooks.py
SurbhiJainUSC: Yes, I am intentionally clearing the `eval_metadata`. This is the sequence of eval:
```
for step in eval_step:
   on_eval_step_start()
   eval_step()
on_eval_step_end()
```
We are clearing `eval_metadata` in `on_eval_step_end()` to ensure that next evaluation run could start with a fresh value of `eval_step_count` and `total_loss`.

MaxText/sft/hooks.py
SurbhiJainUSC: Good point, updated!

MaxText/sft/hooks.py
SurbhiJainUSC: Added a comment.

MaxText/sft/hooks.py
SurbhiJainUSC: Updated!

richjames0 (approved):
LGTM

khatwanimohit (approved):
LGTM

MaxText/sft/hooks.py
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+    self.metric_logger.write_metrics(metrics, train_ctx.train_steps, is_training=False)
+    self.eval_metadata.clear()
+
+    if avg_loss <= self.config.target_eval_loss:
+      raise exceptions.StopTraining(f"Target loss {self.config.target_eval_loss=} is achieved.")
+
+
+class SFTDataHooks(DataHooks):
```
gagika, line 150: Do we need DataHooks?

Instead, Tunix could consume a DataSource (from MaxText or any other source) with a minimal, checkpointable interface:
`next_train(), next_eval(), state(), load_state(), reset(), etc`.

This keeps data lifecycle (prefetch, retries) in one place, makes it checkpointable, and moves eval cadence/early stop into the trainer.
I would simplify swapping in other data sources and remove coupling between hooks and data loading.

gagika (commented):
One comment, feel free to address in a follow up PR as well.

MaxText/sft/hooks.py
SurbhiJainUSC: I will follow up with you on this comment offline and will address in a follow up PR, if needed.

PR #2043, opened 2025-07-29
[Qwix] int8 and fp8 configs

khatwanimohit:
# Description

This PR adds an option to use Qwix for quantization and adds configs similar to their AQT counterpart.

Following AQT configs are used for the Qwix implementation 
- `_get_int8_quant_config`
int8 config with `bwd_weight_grad_tile_size=1/num_slices` to replace the LocalAQT used in drhs
- `_get_aqt_fp8_default_config`
fp8 config with different qtype for forward and backward pass, constant bounds can be set for calibration using `config.quantization_calibration_method='fixed,-0.6,0.6'`, per_tensor scales used for calibration

FIXES: b/433326986

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -117,7 +117,9 @@ save_quantized_params_path: ""
 # when left as is, corresponds to training
 # accepted values are "inference"
 model_call_mode: ""
-
+use_qwix_quantization: False # Whether to use qwix for quantization. If set to True, the model will be quantized using qwix.
+# Quantization calibration method used for weights and activations. Supported methods can be found in https://github.com/google/qwix/blob/dc2a0770351c740e5ab3cce7c0efe9f7beacce9e/qwix/qconfig.py#L70-L80
+quantization_calibration_method: "absmax"
 # Shard the range finding operation for quantization. By default this is set to number of slices.
 quantization_local_shard_count: -1
 
--- a/MaxText/layers/quantizations.py
+++ b/MaxText/layers/quantizations.py
@@ -26,6 +26,8 @@ from aqt.jax.v2.flax import aqt_flax
 from aqt.jax.v2 import tiled_dot_general
 from aqt.jax.v2 import calibration
 
+import qwix
+
 import jax
 import jax.numpy as jnp
 from jax.tree_util import tree_flatten_with_path, tree_unflatten
@@ -545,6 +547,8 @@ def get_quant_mode(quant_mode_str: str = "train"):
 
 def configure_quantization(config: Config, quant_mode_str: str = "train"):
   """Configure quantization based on user config and quant mode."""
+  if config.use_qwix_quantization:
+    return None
   quant_cfg = _get_quant_config(config)
   if quant_cfg:
     if quant_cfg == "fp8":
@@ -606,3 +610,54 @@ def remove_quantized_params(params, aqt_vars):
 
 def configure_kv_quant(config):
   return None if not config.quantize_kvcache else KVQuant(config)
+
+
+def get_basic_config(config, dtype):
+  rules = [
+      qwix.QtRule(
+        module_path='decoder/.*layers.*',  # Apply to all modules
+        weight_qtype=dtype,
+        act_qtype=dtype,
+        bwd_weight_grad_tile_size = 1 / config.quantization_local_shard_count
+      )
+    ]
+  return rules
+
+
+def get_fp8_config(config):
+  """ fp8 config rules with per-tensor calibration.
+  """
+  rules = [
+      qwix.QtRule(
+        module_path='decoder/.*layers.*',  # Apply to all modules
+        weight_qtype=jnp.float8_e4m3fn,
+        act_qtype=jnp.float8_e4m3fn,
+        bwd_qtype=jnp.float8_e5m2,
+        bwd_use_original_residuals=True,
+        disable_channelwise_axes=True, # per_tensor calibration
+        weight_calibration_method = config.quantization_calibration_method,
+        act_calibration_method = config.quantization_calibration_method,
+        bwd_calibration_method = config.quantization_calibration_method,
+      )
+    ]
+  return rules
+
+def get_qt_provider(config):
+  """Get quantization rules based on the config."""
+  match config.quantization:
+    case "int8":
+      return qwix.QtProvider(get_basic_config(config, jnp.int8))
+    case "fp8":
+      return qwix.QtProvider(get_basic_config(config, jnp.float8_e4m3fn))
+    case "fp8_full":
+      return qwix.QtProvider(get_fp8_config(config))
+  return None
+
+
+def maybe_quantize_model(model, config):
+  """Quantize the model if quantization is enabled."""
+  if config.use_qwix_quantization:
+    quantization_provider = get_qt_provider(config)
+    if quantization_provider:
+      model = qwix.quantize_model(model, quantization_provider)
+  return model
--- a/MaxText/train_utils.py
+++ b/MaxText/train_utils.py
@@ -37,6 +37,7 @@ def create_model(config, mesh):
   # Model definition
   quant = quantizations.configure_quantization(config)
   model = get_transformer_model(config, mesh, quant)
+  model = quantizations.maybe_quantize_model(model, config)
   return model
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,6 +28,7 @@ pyink
 pre-commit
 pytype
 pillow>=11.1.0
+qwix@git+https://github.com/google/qwix.git
 sentencepiece==0.2.0
 tensorflow-text>=2.13.0
 tensorflow>=2.13.0
--- a/MaxText/tests/quantizations_test.py
+++ b/MaxText/tests/quantizations_test.py
@@ -17,13 +17,15 @@ limitations under the License.
 """ Tests for the quantizations """
 import unittest
 import os.path
-
+import sys
 import pytest
 
 import numpy as np
 
+import jax
 from jax import numpy as jnp
 from jax import random, lax
+from jax.sharding import Mesh
 
 from flax import linen as nn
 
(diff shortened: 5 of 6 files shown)


MaxText/layers/quantizations.py
```diff
@@ -606,3 +610,46 @@ def remove_quantized_params(params, aqt_vars):
...
+      qwix.qt.QtRule(
+        module_path='.*',  # Apply to all modules
+        weight_qtype=jnp.float8_e4m3fn,
+        act_qtype=jnp.float8_e4m3fn,
+        bwd_qtype=jnp.float8_e5m2,
+        bwd_use_original_residuals=True,
+        disable_channelwise_axes=True, # per_tensor calibration
+        weight_calibration_method = config.quantization_calibration_method,
+        act_calibration_method = config.quantization_calibration_method,
+        bwd_calibration_method = config.quantization_calibration_method,
```
BirdsOfAFthr, line 640: Can users specify the constant bound for 6 different tensors?

MaxText/layers/quantizations.py
```diff
@@ -606,3 +610,46 @@ def remove_quantized_params(params, aqt_vars):
...
+        module_path='.*',  # Apply to all modules
+        weight_qtype=dtype,
+        act_qtype=dtype,
+        bwd_weight_grad_tile_size = 1 / config.quantization_local_shard_count
+      )
+    ]
+  return rules
+
+
+def get_fp8_config(config):
```
BirdsOfAFthr, line 627: Would it be possible to run this on DeepSeek v3 to verify perf on GFC and convergence on VFC

khatwanimohit: @suexu1025 for visibility to run perf experiments

MaxText/layers/quantizations.py
khatwanimohit: Are the constant bounds different for all 6 tensors? Can they be grouped in these 3? I can separate out these 3 configs if its better. 
cc: @liudangyi

liudangyi: IIUC the constant bound we're using today is only for headroom analysis so the actual value doesn't matter. In real world, using a constant bound for every op is almost always incorrect.

MaxText/configs/base.yml
```diff
@@ -117,7 +117,9 @@ save_quantized_params_path: ""
 # when left as is, corresponds to training
 # accepted values are "inference"
 model_call_mode: ""
-
+qwix: False # Whether to use qwix for quantization. If set to True, the model will be quantized using qwix.
```
liudangyi: If it's temporary then I'm fine. Otherwise, the name is very obscure for other users. Maybe `use_qwix_quantization: bool` is better?

MaxText/layers/quantizations.py
liudangyi: I don't think we should block the submission of this PR on the performance or the convergence. Those depend on many factors (hardware availability, XLA flags, etc) and could really take a long time. The quantizations_test.py verifies the correctness, and we could always tune the recipes after.

MaxText/train_utils.py
```diff
@@ -37,6 +39,11 @@ def create_model(config, mesh):
   # Model definition
   quant = quantizations.configure_quantization(config)
   model = get_transformer_model(config, mesh, quant)
+  if config.qwix:
+    assert quant is None, "Using Qwix, quantizations.configure_quantization should return None."
+    quantization_rules = quantizations.get_rules(config)
+    if quantization_rules:
+      model = qwix.quantize_model(model, qwix.QtProvider(quantization_rules))
```
liudangyi: I like how neat it is!

liudangyi (approved):

MaxText/layers/quantizations.py
BirdsOfAFthr: Just to clarify, these recipes are already ‘tuned’. We checked in these AQT FP8 recipes after verifying perf and quality

So I think it could be helpful to verify them as we migrate to Qwix recipes.

MaxText/layers/quantizations.py
BirdsOfAFthr: Static scaling is an important feature we’re exploring for FP8, and we’ve already observed performance improvements across LEMs, LLMs, and diffusion models without quality drop when applying it. I can share the benchmark links privately since this comment is on GitHub.

Static scaling does require profiling to compute factors, but it creates opportunities to improve performance while maintaining quality.

re:the constant bound we're using today is only for headroom analysis
Would Qwix be able to support static scaling with per-tensor granularity? We’d ideally need it to handle six different tensors (fwd 2, bwd 4).

MaxText/configs/base.yml
khatwanimohit: yes, this is temporary ( maybe 2 weeks ) until I migrate everything to Qwix and remove all AQT code

MaxText/layers/quantizations.py
liudangyi: I highly doubt that there can be quality improvements using constant bound (let's don't use static scale because it means something else in Qwix's terminology: the real static scale).

This CL does enable using constant bound by specifying `*_calibration_method` to `"fixed,-7,7"`. We just didn't provide separate configs for the bwd 4 tensors.

MaxText/layers/quantizations.py
liudangyi: This PR doesn't turn AQT off. All changes are gated behind a flag with no impact on existing configs. It's a standard practice to land incremental changes.

We of course need to verify the numerics and the performance before we remove AQT.

MaxText/configs/base.yml
BirdsOfAFthr: Could you please leave this option until we have e2e verification / benchmarks that all Qwix APIs work well? I'm afraid that we remove it prematurely before we have any concrete evidence. With this option, I can still use AQT to do benchmarks.

MaxText/train_utils.py
liudangyi: nit: maybe move the `qwix.quantize_model(model, qwix.QtProvider(quantization_rules))` line into quantizations.py? Benefits are

* train_utils.py no longer depends on qwix.
* Other quantization solutions (like GPU fp8) may use a different provider.

khatwanimohit: awesome suggestion! done

MaxText/configs/base.yml
khatwanimohit: changed the config to `use_qwix_quantization`. 
yes, we will not remove any AQT related code until we have verified correctness and perf

MaxText/layers/quantizations.py
khatwanimohit: thank you @liudangyi for verifying the fp8_full recipe with dynamic scaling and comparing it with AQT which showed parity in performance and loss value

MaxText/layers/quantizations.py
```diff
@@ -606,3 +610,53 @@ def remove_quantized_params(params, aqt_vars):
...
+        weight_calibration_method = config.quantization_calibration_method,
+        act_calibration_method = config.quantization_calibration_method,
+        bwd_calibration_method = config.quantization_calibration_method,
+      )
+    ]
+  return rules
+
+def get_qt_provider(config):
+  """Get quantization rules based on the config."""
+  if config.quantization == "int8":
```
richjames0: nit: i would use a match statement

SurbhiJainUSC (approved):

MaxText/layers/quantizations.py
```diff
@@ -606,3 +610,54 @@ def remove_quantized_params(params, aqt_vars):
 
 def configure_kv_quant(config):
   return None if not config.quantize_kvcache else KVQuant(config)
+
+
+def get_basic_config(config, dtype):
+  rules = [
+      qwix.QtRule(
+        module_path='.*',  # Apply to all modules
```
liudangyi: For a more exact configuration match, the module_path should be `'decoder/.*layers.*'`. I realize that Qwix additionally quantizes the `token_embedder` and the `decoder/logits_dense` module, which are not quantized in AQT.

requirements.txt
```diff
@@ -28,6 +28,7 @@ pyink
 pre-commit
 pytype
 pillow>=11.1.0
+qwix@git+https://github.com/google/qwix.git
```
bvandermoon, line 31: This dependency should also be in `constraints_gpu` and `requirements_with_jax_stable_stack_0_6_1_pipreqs.txt`

dannikay:
I found that with qwix does not work well with python3.10 which is what the MaxText docker image uses. You end up with:

```
  File "/usr/local/lib/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/local/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/deps/MaxText/__init__.py", line 30, in <module>
    from MaxText import train_utils
  File "/deps/MaxText/train_utils.py", line 21, in <module>
    from MaxText.layers import quantizations
  File "/deps/MaxText/layers/quantizations.py", line 29, in <module>
    import qwix
  File "/usr/local/lib/python3.10/site-packages/qwix/__init__.py", line 21, in <module>
    from qwix._src.providers.lora import LoraProvider, LoraRule, apply_lora_to_model
  File "/usr/local/lib/python3.10/site-packages/qwix/_src/providers/lora.py", line 28, in <module>
    from qwix._src.core import einsum
  File "/usr/local/lib/python3.10/site-packages/qwix/_src/core/einsum.py", line 21, in <module>
    from qwix._src.core import dot_general
  File "/usr/local/lib/python3.10/site-packages/qwix/_src/core/dot_general.py", line 351
    return array[*indices]
```

Either we need to bump the docker image python version to 3.11 or we have to conditionally import qwix.

dannikay:
I uploaded #2071 to conditionally import qwix.

PR #2057, opened 2025-07-31
Loss scaling for GA

hx89:
Scale up loss by total_weights instead of multiply the gradient by total_weights when gradient accumulation(GA) is used.

It's mathematically equivalent to the original implementation, but has performance benefits:
1. Makes it easier for XLA to move all-reduce out of the GA loop when use zero1+GA to reduce communication overhead
2. Needs less computations, no need to divide loss by total_weights and multiply gradient by total_weights in each GA step

It may also have less numerical error with less computations.

### DIFF
--- a/MaxText/train.py
+++ b/MaxText/train.py
@@ -307,7 +307,19 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
   xent = xent * (data["targets_segmentation"] != 0)
   total_loss = jnp.sum(xent)
   total_weights = jnp.sum(data["targets_segmentation"] != 0)
-  loss = total_loss / (total_weights + EPS)
+
+  # If gradient accumulation is enabled, we don't need to divide total_loss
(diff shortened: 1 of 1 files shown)


MaxText/train.py
```diff
@@ -307,7 +307,10 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
   xent = xent * (data["targets_segmentation"] != 0)
   total_loss = jnp.sum(xent)
   total_weights = jnp.sum(data["targets_segmentation"] != 0)
-  loss = total_loss / (total_weights + EPS)
+  if config.gradient_accumulation_steps > 1:
+    loss = total_loss
```
gobbleturk, line 320: Can you include a short comment here why we choose to omit EPS in the GA case?

hx89: Added the comment, EPS was used to avoid division by zero, but it's not needed when gradient accumulation is enabled since there's no division.

PR #2061, opened 2025-08-01
Update config validation rules for sparse matmul

RissyRan:
# Description

* Allow sequence parallelism sharding for MoE - this should be aligned with context parallelism (we tested recently) only difference is in attention part which should be consistent with other models
* Update PP sharding rule, it seems does not align with validation comment. 
* Add constrains with EP combination of PP (b/434699033) or TP (b/435539039)

# Tests

* local test for sequence parallelism - [log](https://paste.googleplex.com/6218230775152640)
* other tests in bug to constrain the sharding usage

# Checklist


shuningjin:
Thanks for testing different parallelism! Looks good to me.

MaxText/pyconfig.py
```diff
@@ -939,8 +939,25 @@ def validate_deepseek_moe(raw_keys):
...
+      raw_keys["sparse_matmul"]
+      and raw_keys["decoder_block"] == "deepseek"
+      and (
+          using_fsdp_and_transpose_parallelism(raw_keys)
+          and using_expert_parallelism(raw_keys)
+          and using_tensor_parallelism(raw_keys)
+      )
+  ):
+    raise ValueError(
+        "For DeepSeek v3, sparse matmul doesn't support using fsdp, expert, and tensor parallelism together."
```
gobbleturk: oh really? Why not?

MaxText/pyconfig.py
```diff
@@ -958,22 +975,20 @@ def validate_sparse_matmul_parallelism(raw_keys):
     raise ValueError(
         f"The expert dimension {raw_keys['num_experts']} is not divisible by expert parallelism setting {expert_parallelism}."
     )
-  if (
-      using_pipeline_parallelism(raw_keys)
-      and raw_keys["pipeline_parallel_layers"] is True
-      and raw_keys["model_fsdp_ag_once"] is True
-  ):
+
+  pp_correct_config = raw_keys["pipeline_fsdp_ag_once"] is True and raw_keys["model_fsdp_ag_once"] is False
```
gobbleturk: huh I didn't realize this logic was here, it should be inside of the [set_and_validate_pipeline_config](https://github.com/AI-Hypercomputer/maxtext/blob/86be2a6d56c66be4d6c50435509650ffad7ba5ec/MaxText/pyconfig.py#L785) method above

MaxText/pyconfig.py
```diff
@@ -958,22 +975,20 @@ def validate_sparse_matmul_parallelism(raw_keys):
...
+  if raw_keys["sparse_matmul"] and using_pipeline_parallelism(raw_keys) and not pp_correct_config:
     raise ValueError("You should use the pipeline_fsdp_ag_once = True and leave model_fsdp_ag_once = False.")
```
gobbleturk: pipeline_fsdp_ag_once=True should be set instead of model_fsdp_ag_once regardless of sparse_matul, I think we should move this check into the above [set_and_validate_pipeline_config](https://github.com/AI-Hypercomputer/maxtext/blob/86be2a6d56c66be4d6c50435509650ffad7ba5ec/MaxText/pyconfig.py#L785) method and drop the sparse_matmul check, it should just be

if using_pipeline_parallelism:
    if model_fsdp_ag_once:
        raise ValueError("You should set pipeline_fsdp_ag_once with pipeline parallelism instead of model_fsdp_ag_once")

MaxText/pyconfig.py
RissyRan: It shows internal check error so far: b/435539039

MaxText/pyconfig.py
RissyRan: Cool! I will move them there.

MaxText/pyconfig.py
RissyRan: Sounds good. Not sure why it's there at the very beginning, and I thought it's on purpose for sparse matmul. Will move it out.

MaxText/pyconfig.py
```diff
@@ -784,6 +784,9 @@ def validate_multiple_slices(raw_keys):
 
 def set_and_validate_pipeline_config(raw_keys):
   if using_pipeline_parallelism(raw_keys):
+    pp_correct_config = raw_keys["pipeline_fsdp_ag_once"] is True and raw_keys["model_fsdp_ag_once"] is False
```
gobbleturk: if model_fsdp_ag_once==True:
     value_error(you should only set pipeline_fsdp_once=True, leave model_fsdp_ag_once=False with pipeline parallelism)

RissyRan: Updated, thanks!

gobbleturk (approved):

bvandermoon (approved):

PR #2067, opened 2025-08-01
Fix import from flax

khatwanimohit:
# Description

Fix flax error for Object

FIXES: MaxText TAP builds

# Tests

unit tests

# Checklist

### DIFF
--- a/MaxText/layers/nnx_wrappers.py
+++ b/MaxText/layers/nnx_wrappers.py
@@ -24,9 +24,9 @@ from flax.core import FrozenDict
 from flax.core import meta
 from flax.nnx import graph
 from flax.nnx import variablelib
+from flax.nnx import Object
 from flax.nnx.bridge import module as bdg_module
 from flax.nnx.module import Module
-from flax.nnx.object import Object
 from flax.nnx.rnglib import Rngs
 import jax
 from jax import tree_util as jtu


MaxText/layers/nnx_wrappers.py
```diff
@@ -24,9 +24,9 @@
 from flax.core import meta
 from flax.nnx import graph
 from flax.nnx import variablelib
+from flax.nnx import Object
 from flax.nnx.bridge import module as bdg_module
 from flax.nnx.module import Module
-from flax.nnx.object import Object
```
bvandermoon, line 29: Won't this cause issues for current Flax versions?

khatwanimohit: I checked with my local flax and the import worked. I believe the Object class was importable both ways. 
But I agree with you that we should halt on this because I see other flax PRs from Cristian regarding using Pytree instead of Object

khatwanimohit: @bvandermoon I believe since the tests pass, we should push this PR to alteast unblock TAP tests

hengtaoguo (approved):

gobbleturk (approved):

gobbleturk:
FYI @frgossen has a slightly different fix (cl/789929028), but we can submit this one instead

PR #2068, opened 2025-08-01
Upgrade Flax to 0.11.0

SurbhiJainUSC:
# Description

1. Upgrade Flax to 0.11.0

2. In 3.11, there was a change that makes `nnx.Dropout` fork the `dropout` RngStream on `__init__` so it makes it a requirement if `rngs` is passed. This PR adds a `dropout` key when initializing a model.

# Tests

Unit tests/integration tests

# Checklist

### DIFF
--- a/MaxText/layers/pipeline.py
+++ b/MaxText/layers/pipeline.py
@@ -424,7 +424,7 @@ class Pipeline(nn.Module):
         in_axes=(0, 0, 0, None, None),
         spmd_axis_name="stage",
         variable_axes={"params": 0, "_overwrite_with_gradient": 0},
-        split_rngs={"params": self.is_initializing()},
+        split_rngs={"params": self.is_initializing(), "dropout": self.config.enable_dropout},
         metadata_params={
             nn.PARTITION_NAME: "layers",
             "sub_weight_split_dims_mapping": (None),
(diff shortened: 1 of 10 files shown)


MaxText/layers/pipeline.py
```diff
@@ -682,7 +682,7 @@ def __call__(
                 "non_trainable": 0,
                 "hyper_params": 0,
             },
-            split_rngs={"params": True},
+            split_rngs={"params": True, "dropout": True},
```
bvandermoon: Why true instead of the config value here?

SurbhiJainUSC: Hmm, I just used the same value as params here. I can update it to use config value instead.

SurbhiJainUSC: Updated!

shralex (approved):

gobbleturk (approved):

requirements.txt
```diff
@@ -4,7 +4,7 @@ array-record
 cloud-accelerator-diagnostics
 cloud-tpu-diagnostics
 datasets>=4.0.0
-flax==0.10.6
+flax>=0.11.0
```
bvandermoon, line 7: Is this also needed in `gpu_constraints.txt`?

bvandermoon (approved):

requirements.txt
SurbhiJainUSC: Updated

PR #2075, opened 2025-08-04
port Transformer

cgarciae:
# Description
* Port `Transformer` to NNX and creates the `transformer_to_linen` wrapper.

# Tests
* Llama2-7B gives same perf with `enable_nnx` on/off:
```
python3 -m MaxText.train MaxText/configs/base.yml \
    run_name=<run_name> \
    base_output_directory=gs://<gcs_bucket> \
    dataset_type=synthetic \
    steps=10 \
    model_name=llama2-7b
```
* Llama2-7B gives decode output with `enable_nnx` on/off:
```
python3 -m MaxText.decode MaxText/configs/base.yml \
    model_name=llama2-7b \
    tokenizer_path=assets/tokenizer_llama3.tiktoken \
    tokenizer_type=tiktoken \
    scan_layers=false \
    per_device_batch_size=1 \
    ici_fsdp_parallelism=1 \
    ici_autoregressive_parallelism=-1 \
    max_target_length=256 \
    prompt="I love to" \
    attention=dot_product
```
* Llama3.3-70B Maxengine/Jetstream gives same accuracy for `enable_nnx` on/off: https://diff.googleplex.com/#key=O4rlQHQDz8Cg

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -797,3 +797,6 @@ projector_dropout_for_vit: 0.0
 # Subslice shape in the form of "x,y,z" when using pathways (single controller). 
 # Example: "8,8" to use a 8x8 subgrid (64 chips) of a full pod (16x16) of trillium.
 subslice_shape: ""
+
+# NNX
+enable_nnx: false
--- a/MaxText/layers/decoders.py
+++ b/MaxText/layers/decoders.py
@@ -25,6 +25,7 @@ from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
 
 from flax import linen as nn
+from flax import nnx
 from flax.linen.partitioning import ScanIn
 
 from MaxText.common_types import DecoderBlockType, Config, MODEL_MODE_TRAIN, MODEL_MODE_PREFILL, MODEL_MODE_AUTOREGRESSIVE
@@ -133,7 +134,7 @@ class DecoderLayer(nn.Module):
         ar_cache_axis_order=tuple(map(int, cfg.ar_cache_axis_order.split(","))),
         compute_axis_order=tuple(map(int, cfg.compute_axis_order.split(","))),
         reshape_q=cfg.reshape_q,
-        model_mode=model_mode,
+        model_mode=self.model_mode,
     )
 
     attention_lnx = attention_layer(
@@ -142,7 +143,7 @@ class DecoderLayer(nn.Module):
         decoder_positions,
         decoder_segment_ids=decoder_segment_ids,
         deterministic=deterministic,
-        model_mode=model_mode,
+        model_mode=self.model_mode,
     )
 
     if model_mode == MODEL_MODE_PREFILL:
@@ -159,7 +160,7 @@ class DecoderLayer(nn.Module):
         dtype=cfg.dtype,
         weight_dtype=cfg.weight_dtype,
         name="mlp",
-        model_mode=model_mode,
+        model_mode=self.model_mode,
         config=cfg,
         quant=self.quant,
     )(lnx, deterministic=deterministic)
@@ -221,7 +222,7 @@ class SequentialBlockDecoderLayers(nn.Module):
   ) -> jnp.ndarray:
     for lyr in range(self.num_decoder_layers):
       inputs = self.decoder_layer(
-          config=self.config, mesh=self.mesh, name=f"layers_{lyr}", quant=self.quant, model_mode=model_mode
+          config=self.config, mesh=self.mesh, name=f"layers_{lyr}", quant=self.quant, model_mode=self.model_mode
       )(
           inputs,
           decoder_segment_ids,
@@ -243,7 +244,6 @@ class Decoder(nn.Module):
   """A stack of decoder layers as a part of an encoder-decoder architecture."""
 
   config: Config
-  shared_embedding: nn.Module
   mesh: Mesh
   quant: None | Quant = None
   model_mode: str = MODEL_MODE_TRAIN
@@ -422,9 +422,7 @@ class Decoder(nn.Module):
     else:
       raise ValueError(f"Incorrect decoder_block name {self.config.decoder_block.value=}")
 
-  def scan_decoder_layers(
-      self, cfg, decoder_layer, length, metadata_axis_name, mesh, in_axes_tuple, model_mode, **kwargs
-  ):
+  def scan_decoder_layers(self, cfg, decoder_layer, length, metadata_axis_name, mesh, in_axes_tuple, **kwargs):
     """scan decoder layers, calls `flax.linen.transforms.scan`"""
     initializing = self.is_mutable_collection("params")
     params_spec = cfg.param_scan_axis if initializing else ScanIn(cfg.param_scan_axis)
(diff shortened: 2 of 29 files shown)


MaxText/maxtext_utils.py
```diff
@@ -1098,10 +1099,7 @@ def get_kv_cache_annotations(model, config, rng, mesh, page_state: Optional[Page
   """Get a shaped abstraction of the state (including optimizer)"""
 
   def init_kv_cache(model, config):
-    input_shape = (
-        config.global_batch_size_to_load,
-        1,
-    )
+    input_shape = (config.micro_batch_size_to_train_on, 1)
```
bvandermoon, line 1107: This should be changed in `get_prefill_kv_cache_annotations also, right?

bvandermoon (commented):
Discussed the inference memory issues more with the team this morning. I am looking into adding a config flag to enable/disable NNX for the remaining modules. We can use that flag for this and https://github.com/AI-Hypercomputer/maxtext/pull/2178. That way we can still merge to main. Hoping it is fairly straightforward, taking a closer look now

MaxText/maxtext_utils.py
cgarciae: yeah that looks about right

MaxText/layers/linears.py
```diff
@@ -289,6 +289,29 @@ def dense_general(
...
+    rng_collection: str = 'dropout',
+    rngs: nnx.Rngs| None = None,
+  ):
+    self.rate = rate
+    self.broadcast_dims = broadcast_dims
+    self.deterministic = deterministic
+    self.rng_collection = rng_collection
+
+    if isinstance(rngs, nnx.Rngs):
+      self.rngs = rngs.fork()
```
bvandermoon: Could you add the fix you previously mentioned for old Flax versions here? I see an attribute error on Flax 0.10.6 from this line

```
 self.rngs = rngs.fork() if hasattr(type(rngs), 'fork') else rngs
```

MaxText/maxtext_utils.py
bvandermoon: Reopening this, could you please also change this above?

MaxText/layers/decoders.py
```diff
@@ -587,11 +587,11 @@ def _apply_output_head(self, y, deterministic, model_mode):
   @nn.compact
   def __call__(
       self,
+      shared_embedding: nn.Module | nnx.Module,
```
bvandermoon, line 590: Why do we need to pass this `shared_embedding` around now?

MaxText/maxengine.py
```diff
@@ -1416,6 +1416,7 @@ def init(abstract_params, page_state):
           mutable=["cache"],
           page_state=page_state,
           slot=0,
+          # nnx_method="no_op",
```
bvandermoon: Nit: please remove commented code

bvandermoon: Or is it needed here?

MaxText/configs/base.yml
```diff
@@ -796,3 +796,6 @@ projector_dropout_for_vit: 0.0
 # Subslice shape in the form of "x,y,z" when using pathways (single controller). 
 # Example: "8,8" to use a 8x8 subgrid (64 chips) of a full pod (16x16) of trillium.
 subslice_shape: ""
+
+# NNX
+enable_nnx: false
```
A9isha: nit: please add the new line

MaxText/layers/decoders.py
A9isha: `class Decoder` now becomes NNX, and shouldn't have a Linen attribute `shared_embedding`.  I wonder why we cannot use the NNX version of [embeddimgs](https://github.com/AI-Hypercomputer/maxtext/blob/a0a77e1d6adda33f49c8e14b3163ffe818300c56/MaxText/layers/models.py#L60). Does the bridge not support such nested NNX modules?

A9isha: My followup qs is also why can't we pass around `model_mode` anymore and need to save it as an attribute?

MaxText/layers/decoders.py
```diff
@@ -751,7 +749,7 @@ def __call__(
                   decoder_segment_ids,
                   decoder_positions,
                   deterministic,
-                  model_mode,
+                  self.model_mode,
```
A9isha, line 752: sorry for the rookie question - why are we able to pass `model_mode` here but not above?

MaxText/layers/models.py
```diff
@@ -52,6 +55,22 @@ class Transformer(nn.Module):
   model_mode: str = MODEL_MODE_TRAIN  # May be different than the model_mode passed to __call__
   # pylint: enable=attribute-defined-outside-init
 
+  def init(self, *args, model_mode: str = MODEL_MODE_TRAIN, **kwargs):
+    """Initializes the model."""
+    if model_mode != self.model_mode:
+      module = self.clone(model_mode=model_mode)
```
A9isha: I don't understand this :(

MaxText/layers/models.py
```diff
@@ -161,12 +180,213 @@ def __call__(
...
+      true_length: int | None = None,
+      slot: int | None = None,
+      page_state: page_manager.PageState | None = None,
+      decoder_target_tokens: jax.Array | None = None,
+      decoder_target_mask: jax.Array | None = None,
+  ):
+    """Applies Transformer decoder-branch on encoded-input and target.
+
+    Args:
+      true_length: (Optional) Prompt length before padding
```
A9isha, line 318: wondering why only 2 of the args are described here...

MaxText/tests/model_test.py
```diff
@@ -166,10 +167,8 @@ def test_train_vs_prefill_and_autoregress(self):
         mutable=["cache"],
     )
 
-    self.assertTrue(
-        jax.numpy.allclose(
-            full_train_logits[:, :PREFILL_RANGE, :], partial_prefill_logits, rtol=1e-01, atol=1e-01, equal_nan=False
-        )
+    np.testing.assert_allclose(
```
A9isha, line 170: curious why assertTrue doesn't work?

MaxText/train_utils.py
```diff
@@ -35,10 +38,28 @@
...
+    model_mode: str = MODEL_MODE_TRAIN,
+) -> nn.Module: ...
+@overload
+def from_pretrained(
+    config: pyconfig.HyperParameters,
+    devices: Sequence[jax.Device] | None = None,
+    *,
+    model_mode: str = MODEL_MODE_TRAIN,
+    rngs: nnx.Rngs,
+) -> models.Transformer: ...
```
A9isha, line 55: why do we need two overridden methods? is it only to have non-null rngs? then should return type of the second one skip models.Transformer?

A9isha (commented):
Hats off to you Cristian for making this work!!

MaxText/layers/decoders.py
cgarciae: In principle we can, at some point @bvandermoon and I thought it was better to only have it in the constructor but maybe we can have both.

MaxText/layers/linears.py
cgarciae: good catch!

MaxText/layers/models.py
cgarciae: I think its always been like this

MaxText/tests/model_test.py
cgarciae: this one is better because it gives you more info on the error message

MaxText/train_utils.py
cgarciae: this is helping the type system know that if you pass `rngs` the return is of type `Transformer` (NNX) else its Linen.

A9isha (approved):
Thank you Cristian - hopefully the tests go green soon

PR #2088, opened 2025-08-05
NNX Migration for Mistral models

NuojCheng:
# Description

# Tests

# Checklist


MaxText/layers/mistral.py
```diff
@@ -59,47 +70,43 @@ def __call__(
...
+    lnx_rms = RMSNorm(
         num_features=inputs.shape[-1],
-        dtype=cfg.dtype,
-        weight_dtype=cfg.weight_dtype,
-        name="pre_self_attention_layer_norm",
+        dtype=self.config.dtype,
+        weight_dtype=self.config.weight_dtype,
         kernel_axes=("norm",),
-        epsilon=cfg.normalization_layer_epsilon,
+        epsilon=self.config.normalization_layer_epsilon,
```
bvandermoon: Looking good so far, thanks @NuojCheng. You will want to move NNX module initiation like `RMSNorm`, `Attention`, etc. to the `__init__` of `MistralDecoderLayer`

NuojCheng: Make sense!

src/MaxText/layers/mistral.py
```diff
@@ -21,131 +21,155 @@
...
+from MaxText.common_types import MODEL_MODE_PREFILL
 
 
 # -----------------------------------------
 # The Decoder Layer for Mistral
 # -----------------------------------------
 
 
-class MistralDecoderLayer(nn.Module):
+class MistralDecoderLayerNNX(nnx.Module):
```
bvandermoon: Let's keep the same module name rather than appending everything with NNX

src/MaxText/layers/mistral.py
```diff
@@ -21,131 +21,155 @@
...
+        intermediate_dropout_rate=config.dropout_rate,
+        dtype=config.dtype,
+        weight_dtype=config.weight_dtype,
+        config=config,
+        quant=self.quant,
+        model_mode=model_mode,
+        rngs=self.rngs,
+    )
+
+    self.dropout = nnx.Dropout(rate=config.dropout_rate, broadcast_dims=(-2,), rngs=self.rngs)
```
bvandermoon: Please use the `linears.Dropout` from MaxText instead of `nnx.Dropout`. There are some known issues with the existing `nnx.Dropout` in MaxText

bvandermoon (commented):
LGTM just a couple small comments. Are the inference decode test results in the description from after the rebase with #2370? I see some very slight memory diff

Also discussed this offline already, but do you mind running some Maxengine/Jetstream test and loading from a checkpoint?

src/MaxText/layers/mistral.py
NuojCheng: Thank you for noticing that. Fixed.

src/MaxText/layers/mistral.py
NuojCheng: Fixed.

NuojCheng:
> I see both of accuracy is 0 from JetStream. Did you kill the process or others?
> 
> ```
> Results
> 
> {'accuracy': 0.0, 'gen_num': 5000}
> ```

I did not kill the process. I am not sure what happened here, maybe using the wrong tokenizer? (I used the mistral tokenizer under assets instead of huggingface in llama2)

src/MaxText/layers/mistral.py
```diff
@@ -21,131 +21,155 @@
...
+        model_mode=model_mode,
+        rngs=self.rngs,
+    )
+
+    self.dropout = Dropout(rate=config.dropout_rate, broadcast_dims=(-2,), rngs=self.rngs)
+    
+    if model_mode == MODEL_MODE_PREFILL:
+      self.activation_axis_names = ("activation_batch", "prefill_activation_norm_length", "activation_embed")
+    else:
+      self.activation_axis_names = ("activation_batch", "activation_norm_length", "activation_embed")
```
bvandermoon: Please remove this if it is added based on https://github.com/AI-Hypercomputer/maxtext/pull/2370. Llama2 previously had it in the `__call__` function, so it was just being move to `__init__`. Looks like Mistral doesn't have that

NuojCheng: Thank you. Updated!

bvandermoon (approved):
LGTM, thanks for the great thorough testing so far. Please do follow up here when you have the results of the golden logits test you mentioned you were running

src/MaxText/layers/mistral.py
```diff
@@ -21,131 +21,151 @@
...
+    self.config = config
+    self.mesh = mesh
+    self.quant = quant
+    self.rngs = rngs
+
+    batch_size, seq_len = max_utils.get_batch_seq_len_for_mode(config, model_mode)
+    dummy_inputs_shape = (batch_size, seq_len, config.emb_dim)
+
+    self.pre_self_attention_layer_norm = RMSNorm(
+        num_features=dummy_inputs_shape[-1],
```
SurbhiJainUSC: Should this be `config.emb_dim`?

bvandermoon: Yes, it should: https://github.com/AI-Hypercomputer/maxtext/pull/2178#discussion_r2283002384

Nice catch @SurbhiJainUSC

NuojCheng: updated. Thank you for the catch!

PR #2094, opened 2025-08-06
Migrate moe.py to NNX

RissyRan:
# Description

* Migrate `moe.py` file to use NNX
* Update unit tests. Also we have G3 version of MoE tests ([test1](https://source.corp.google.com/piper///depot/google3/third_party/py/maxtext/tests/moe_google_test.py?q=moe_google_test), [test2](https://source.corp.google.com/piper///depot/google3/third_party/py/maxtext/tests/moe_ragged_dot_test.py)), and plan to address then when cl is available from copybara (or as a follow up if not blocking)

# Tests

* Unit tests passing, end-to-end model tests passing
* Test with/without the change for 15 steps of training (base commit: `f9d55998a9a0db1370d4391ff24bfac3c1a28b4a`)

  * Mixtral (small 5 layers): tflops/s/device both at ~101, and same memory usage ([with change](https://paste.googleplex.com/6266857287581696), [without change](https://paste.googleplex.com/5306154305191936)).
  * DeepSeek (small 5 layers):  tflops/s/device both at ~62, and same memory usage ([with change](https://paste.googleplex.com/6335829764734976), [without change](https://paste.googleplex.com/6300186233798656))
  * Mixtral quantization: tflops/s/device both at ~108, and same memory usage ([with change](https://paste.googleplex.com/4724932470374400), [without change](https://paste.googleplex.com/5311955363168256))
  * DeepSeek quantization: tflops/s/device both at ~63, and same memory usage ([with change](https://paste.googleplex.com/6750287062368256), [without change](https://paste.googleplex.com/6602365301686272))

*  Forward logit check for ds v2 - [pass](https://paste.googleplex.com/4505554583814144)

# Checklist

### DIFF
--- a/MaxText/layers/moe.py
+++ b/MaxText/layers/moe.py
@@ -22,20 +22,20 @@ from typing import Iterable, Optional, Tuple, Union
 
 from aqt.jax.v2 import aqt_tensor as aqt
 import flax.linen as nn
+from flax import nnx
 import jax
 from jax import ad_checkpoint as adc
 from jax.experimental import shard_map
 from jax.experimental import xla_metadata
 import jax.numpy as jnp
+import numpy as np
+
 from MaxText import common_types as ctypes
 from MaxText import max_logging
 from MaxText import max_utils
 from MaxText.kernels import megablox as mblx
-from MaxText.layers import attentions
-from MaxText.layers import initializers
-from MaxText.layers import linears
-from MaxText.layers import quantizations
-import numpy as np
+from MaxText.layers import attentions, linears, quantizations, nnx_wrappers
+from MaxText.layers.initializers import NdInitializer, nd_dense_init, default_bias_init, variable_to_logically_partitioned
 
 
 set_xla_metadata = xla_metadata.set_xla_metadata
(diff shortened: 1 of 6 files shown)


MaxText/layers/moe_linen.py
```diff
@@ -0,0 +1,128 @@
...
+import jax.numpy as jnp
+from MaxText import common_types as ctypes
+from MaxText.layers import moe
+from MaxText.layers import linears
+from MaxText.layers import quantizations
+from MaxText.layers import nnx_wrappers
+from MaxText.layers.initializers import NdInitializer, nd_dense_init, variable_to_logically_partitioned
+
+
+def get_gate_logit(
```
cgarciae: currently we are putting the `to_linen` wrappers next to their NNX Module counter part, these will be deleted very soon so I don't think it worth creating a new file.

MaxText/layers/moe.py
```diff
@@ -72,74 +72,122 @@ def random_routing(rng_key, gate_logits, num_experts_per_tok):
...
-  def __call__(self, inputs: ctypes.Array) -> Tuple[ctypes.Array, Optional[ctypes.Array]]:
+  @property
+  def quant_dot_general(self) -> nnx_wrappers.ToNNX | None:
+    if self._quant_dot_general_name is None:
+      return None
+    return getattr(self, self._quant_dot_general_name)
 
-    features = linears.canonicalize_tuple(self.features)
-    axis = linears.canonicalize_tuple(self.axis)
+  def __call__(self, inputs: ctypes.Array, _initializing: bool = False) -> Tuple[ctypes.Array, Optional[ctypes.Array]]:
```
cgarciae: `ctypes` is probably not right here, use `jax.Array`:
```suggestion
  def __call__(self, inputs: jax.Array, _initializing: bool = False) -> tuple[jax.Array, jax.Array | None]:
```

MaxText/layers/moe.py
```diff
@@ -1415,27 +1466,15 @@ def retrieve_quantized_weight(
...
-        name="gate",
-        use_bias=cfg.routed_bias,
-        score_func=cfg.routed_score_func,
-        matmul_precision=cfg.matmul_precision,
-    )(inputs)
+    gate_logits, pre_bias_logits = self.gate(inputs)
+
+    w0_kernel = jnp.asarray(self.wi_0.value, self.dtype)
+    w1_kernel = jnp.asarray(self.wi_1.value, self.dtype)
+    wo_kernel = jnp.asarray(self.wo.value, self.dtype)
```
cgarciae: Because `wi_0` `wi_1` and `wo` can also be a `zeros` array `.value` will fail. You can use `[...]` syntax as a type-agnostic way of getting the value, for Arrays this is an identity function.
```suggestion
    w0_kernel = jnp.asarray(self.wi_0[...], self.dtype)
    w1_kernel = jnp.asarray(self.wi_1[...], self.dtype)
    wo_kernel = jnp.asarray(self.wo[...], self.dtype)
```

MaxText/layers/moe.py
```diff
@@ -1451,59 +1490,74 @@ def __call__(
...
-        activations=cfg.mlp_activations,
-        intermediate_dropout_rate=cfg.dropout_rate,
-        dtype=cfg.dtype,
-        weight_dtype=cfg.weight_dtype,
-        name="shared_experts",
-        config=cfg,
+        rngs=self.rngs,
+    )
+
+    self.shared_moe = linears.MlpBlock(
```
cgarciae: seems cleaner to assign it to a `shared_experts` attribute:
```suggestion
    self.shared_experts = linears.MlpBlock(
```
if needed create a `shared_moe` property if you need the alias.

MaxText/layers/moe_linen.py
RissyRan: Thanks for the comment! The original moe.py is very long already with 1500+ lines. So create a new file and will be easier for clean-up (delete the file directly instead of looking for functions in original file)?

MaxText/layers/moe.py
RissyRan: Actually let me update it accordingly. Thanks!

MaxText/layers/moe.py
RissyRan: Thank you! Will update.

MaxText/layers/moe.py
RissyRan: Sounds good! Will update and delete those code

```
    self.shared_moe_ckpt_name = "shared_experts"
    setattr(self, self.shared_moe_ckpt_name, self.shared_moe)
    ```

File: MaxText/layers/moe_linen.py (hunk 1 above)
Comment (RissyRan) [reply MaxText/layers/moe_linen.py]: After a second thought, I like your idea to keep it inside of `moe.py`, in this case we don't need to update BUILD file in G3. Will update.

File: MaxText/layers/moe.py [hunk 5]
```diff
@@ -1451,59 +1488,175 @@ def __call__(
...
+    """
+    self.config = config
+    self.mesh = mesh
+    self.kernel_init = kernel_init
+    self.kernel_axes = kernel_axes
+    self.weight_dtype = weight_dtype
+    self.dtype = dtype
+    self.quant = quant
+    self.rngs = rngs
+    self.routed_moe = RoutedMoE(
```
cgarciae: Setting a Module twice can cause issues. I would be better to just set it once with its Linen name:

```suggestion
    self.MoeBlock_0 = RoutedMoE(
```
if you want to create an alias we can define a property:

```
@property
def routed_moe(self):
  return self.MoeBlock_0

File: MaxText/layers/moe.py [hunk 6]
```diff
@@ -1451,59 +1488,175 @@ def __call__(
...
-    )(inputs)
-    logical_axis_names = ("activation_batch", "activation_norm_length", "activation_embed")
-    shared_experts = nn.with_logical_constraint(shared_experts, logical_axis_names)
+        rngs=self.rngs,
+    )
+
+    # NOTE: the naming mismatch here is to ensure reverse compatibility with
+    # existing checkpoints.
+    self.routed_moe_ckpt_name = "MoeBlock_0"
+    setattr(self, self.routed_moe_ckpt_name, self.routed_moe)
```
cgarciae: Consider removing this in place of the above

MaxText/layers/moe.py
```diff
@@ -1451,59 +1488,175 @@ def __call__(
...
 
+  def __call__(self, inputs: jax.Array) -> jax.Array:
+    routed_experts, _ = getattr(self, self.routed_moe_ckpt_name)(inputs)
```
cgarciae:
```suggestion
    routed_experts, _ = self.MoeBlock_0(inputs)
```

MaxText/layers/moe.py
RissyRan: Great! Updated this way.

MaxText/layers/moe.py
RissyRan: Updated accordingly.

MaxText/tests/moe_test.py
```diff
@@ -177,7 +180,7 @@ def setUp(self):
     )
     self.rng = jax.random.PRNGKey(42)
     quant = Fp8Quantization()
-    self.model = mlp_block(
+    self.model = linears.mlp_block(
```
bvandermoon, line 183: Nit: should this be the NNX version?

bvandermoon (approved):
LGTM, great to see this. Thank you @RissyRan

MaxText/tests/moe_test.py
RissyRan: It seems this one is added by @cgarciae to test Linen initialization? We could delete it if we don't need it.

RissyRan:
> MaxText/tests/moe_test.py

Good point! I just tested by loading a ckpt, and it works for Mixtral 8x7b: https://paste.googleplex.com/6527470786052096

PR #2101, opened 2025-08-07
Refactor models to use `mlp_block` +  `use_pre_norm`

phu0ngng:
# Description

This PR refactors some of the models to use the existing `use_pre_norm` in the `mlp_block` instead of calling a separate norm before the MLP block. 
It should not affect the performance or functionality of MaxText, but rather be code cleanup. 
Besides, this PR also paved the way for us to integrate the TransformerEngine LayerNormMLP block into MaxText for our E2E testing.

# Tests

No changes in testing

# Checklist

### DIFF
--- a/MaxText/layers/linears.py
+++ b/MaxText/layers/linears.py
@@ -314,7 +314,7 @@ class MlpBlock(nnx.Module):
       rngs: nnx.Rngs,
   ) -> None:
     """A MlpBlock module.
-    
+
     Args:
       config: Config object containing model parameters.
       in_features: Number of input features.
@@ -422,8 +422,19 @@ class MlpBlock(nnx.Module):
     """Applies Transformer MlpBlock module."""
     cfg = self.config
 
+
     if self.mlp_layer_norm is not None:
       inputs = self.mlp_layer_norm(inputs)
+      if self.model_mode == MODEL_MODE_PREFILL:
+        inputs = nn.with_logical_constraint(inputs, ("activation_batch",
+                                                     "prefill_activation_norm_length",
+                                                     "activation_embed")
+                                            )
+      else:
+        inputs = nn.with_logical_constraint(inputs, ("activation_batch",
+                                                     "activation_norm_length",
+                                                     "activation_embed")
+                                            )
 
     # Iterate over specified MLP input activation functions.
     # e.g. ('relu',) or ('gelu', 'linear') for gated-gelu.
--- a/MaxText/layers/llama2.py
+++ b/MaxText/layers/llama2.py
@@ -127,20 +127,9 @@ class LlamaDecoderLayer(nn.Module):
     attention_lnx = nn.with_logical_constraint(attention_lnx, activation_axis_names)
     intermediate_inputs = inputs + attention_lnx
 
-    # Fully Connected
-    hidden_states = rms_norm(
-        num_features=intermediate_inputs.shape[-1],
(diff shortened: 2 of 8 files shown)


MaxText/layers/linears.py
```diff
@@ -408,7 +408,7 @@ def get_norm_layer(self, num_features: int):
         DecoderBlockType.DEEPSEEK,
         DecoderBlockType.LLAMA4,
     ):
-      return functools.partial(normalizations.RMSNorm, num_features=num_features)
+      return functools.partial(normalizations.rms_norm, num_features=num_features)
```
NuojCheng: All other changes look good to me, but for this line we wanna return an nnx.module instead a linen one, right? @bvandermoon

bvandermoon: Good find @NuojCheng. Yes this should be the NNX version `RMSNorm`

phu0ngng: Hi, I updated the code accordingly. Thanks for a prompt review.

NuojCheng (approved):
LGTM

MaxText/layers/mistral.py
```diff
@@ -117,21 +117,9 @@ def __call__(
     )
     intermediate_inputs = inputs + attention_lnx
 
-    # Fully Connected
-    hidden_states = rms_norm(
-        num_features=intermediate_inputs.shape[-1],
-        dtype=cfg.dtype,
-        weight_dtype=cfg.weight_dtype,
-        name="post_self_attention_layer_norm",
```
bvandermoon, line 125: Does this still work with existing checkpoints? I don't think name is being passed in anymore

MaxText/layers/llama2.py
```diff
@@ -127,20 +127,9 @@ def __call__(
     attention_lnx = nn.with_logical_constraint(attention_lnx, activation_axis_names)
     intermediate_inputs = inputs + attention_lnx
 
-    # Fully Connected
-    hidden_states = rms_norm(
-        num_features=intermediate_inputs.shape[-1],
-        dtype=cfg.dtype,
-        weight_dtype=cfg.weight_dtype,
-        name="post_self_attention_layer_norm",
```
bvandermoon, line 135: Does this still work with existing checkpoints? I don't think name is being passed in anymore

bvandermoon: @phu0ngng Based on our tests, it looks like this does not work properly with existing checkpoints. Can we roll this back or have a roll forward to include `name`?

Also, could you please provide a little more context around this from the description: `Besides, this PR also paved the way for us to integrate the TransformerEngine LayerNormMLP block into MaxText for our E2E testing.`

phu0ngng: Hi @bvandermoon, 

It is correct that the `name` could not be passed, as we are now using the NNX module, as requested above, instead of the linen module as in the past. 

I would prefer to roll forward to include `name` and can send a PR today if you are ok with using the linen model here.

phu0ngng: We are testing some of our training recipes for FP8 by replacing your `MLPBlock` with our TransformerEngine `LayerNormMLP`. So with this refactor in MaxText, the replacement will be pretty straightforward. We also plan to use this for testing incoming recipes, i.e. NVFP4. 

The integration is mostly used in-house right now, but we are happy to upstream them eventually if there is interest.

PR #2154, opened 2025-08-12
Add `norm_ckpt_name` to `mlp_block`

phu0ngng:
# Description

PR GH-2101 obmits the checkpointing name for normalization layers. This PR added the name back so that the `mlp_block` can checkpoint the normalization with the correct name.

# Checklist

### DIFF
--- a/MaxText/layers/gemma.py
+++ b/MaxText/layers/gemma.py
@@ -117,6 +117,7 @@ class GemmaDecoderLayer(nn.Module):
         config=cfg,
         quant=self.quant,
         use_pre_norm=True,
+        norm_ckpt_name="pre_ffw_norm",
     )(residual, deterministic=deterministic)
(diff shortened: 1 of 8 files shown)


bvandermoon:
Thanks for the roll forward @phu0ngng. I think this still changes the nesting of the RMSNorm in the MlpBlock. Therefore it won't work for existing checkpoints. I am not sure there is a way to fix the nesting piece. I think we will need to roll back the initial PR. Can you please help me understand how the original change impacts your testing?

bvandermoon (commented):
Thanks for the roll forward @phu0ngng. I think this still changes the nesting of the RMSNorm in the MlpBlock. Therefore it won't work for existing checkpoints. I am not sure there is a way to fix the nesting piece. I think we will need to roll back the initial PR. Can you please help me understand how the original change impacts your testing?

phu0ngng:
Close as this `mlp_block` refactoring causes disruptions in training with old checkpoints.

PR #2155, opened 2025-08-12
Migrate LlamaDecoderLayer to NNX

bvandermoon:
# Description

Migrate LlamaDecoderLayer to NNX instead of Linen

* Change module type to NNX
* Initialize submodules in `__init__` instead of `__call__`
* Pass `self.model_mode` to module constructors instead of the one passed to call
* New `to_linen_class` NNX bridge to handle the way we pass decoder classes around. This is needed for checkpointing to work
* New init/apply functions in Transformer module to ensure the proper `model_mode` value is passed around

# Tests

commands to reproduce.

# Checklist


MaxText/layers/llama2.py
```diff
@@ -18,39 +18,115 @@
...
+      model_mode: str,
+      mesh: Mesh,
+      quant: Optional[Quant] = None,
+      rngs: Optional[nnx.Rngs] = None,
+  ):
+
+    self.config = config
+    self.mesh = mesh
+    self.quant = quant
+    self.rngs = rngs
```
NuojCheng: I think it could be 
`self.rngs = rngs if rngs else nnx.Rngs(0)`
otherwise error may incur in [RMSNorm initialization](https://github.com/AI-Hypercomputer/maxtext/blob/00d6ea88b4c0c415521d64cf6f963374862dfad9/MaxText/layers/normalizations.py#L53)

MaxText/layers/llama2.py
```diff
@@ -42,15 +44,88 @@
...
-  mesh: Mesh
-  model_mode: str
-  quant: Optional[Quant] = None
+  def __init__(
+      self,
+      config: Config,
+      model_mode: str,
+      mesh: Mesh,
+      quant: Optional[Quant] = None,
+      rngs: Optional[nnx.Rngs] = None,
```
cgarciae: Rngs should be required
```suggestion
      *,
      rngs: nnx.Rngs,
```

MaxText/layers/llama2.py
```diff
@@ -42,15 +44,88 @@
...
+  ):
+
+    self.config = config
+    self.mesh = mesh
+    self.quant = quant
+    self.rngs = rngs
```
cgarciae: Don't assign it as an attribute. Its not used in call and introduces potential references sharing.

PR #2162, opened 2025-08-12
Migrate gemma to NNX

SurbhiJainUSC:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist


src/MaxText/layers/gemma.py
```diff
@@ -14,30 +14,109 @@
...
+        activations=config.mlp_activations,
+        intermediate_dropout_rate=config.dropout_rate,
+        dtype=config.dtype,
+        weight_dtype=config.weight_dtype,
+        quant=self.quant,
+        model_mode=self.model_mode,
+        rngs=self.rngs,
+    )
+
+    self.dropout = nnx.Dropout(rate=config.dropout_rate, broadcast_dims=(-2,), rngs=self.rngs)
```
bvandermoon: Please use the `linears.Dropout` from MaxText instead of `nnx.Dropout`. There are some known issues with the existing `nnx.Dropout` in MaxText. Same for the one in `gemma2.py`

src/MaxText/layers/gemma2.py
```diff
@@ -14,30 +14,200 @@
...
+        rngs=self.rngs,
+    )
+
+    self.post_ffw_norm_global = RMSNorm(
+        num_features=config.emb_dim,
+        dtype=config.dtype,
+        weight_dtype=config.weight_dtype,
+        kernel_axes=("norm",),
+        rngs=self.rngs,
+    )
```
bvandermoon: Since it is only used in `__call__` when `self.config.use_post_ffw_norm` is True, it should also only be initialized in that scenario to avoid adding the extra params. It's look like there may be some others in this PR as well like `post_self_attention_norm_global`

src/MaxText/layers/gemma.py
SurbhiJainUSC: Updated!

src/MaxText/layers/gemma2.py
SurbhiJainUSC: That's a good suggestion, updated!

SurbhiJainUSC:
> Thank you Surbhi. Can you please also share the before profiles in the description? So we can check that the HLOs are identical

@bvandermoon - I have added before & after profiles.

bvandermoon:
> > Thank you Surbhi. Can you please also share the before profiles in the description? So we can check that the HLOs are identical
> 
> @bvandermoon - I have added before & after profiles.

Thanks @SurbhiJainUSC. The gemma profiles look good, but it looks like the gemma2 profiles have different HLOs. The general memory usage still looks very similar though. Could you check if anything is standing out as different in the gemma2 logic? I took a look but didn't see anything standing out

I assume the same JAX version, etc. were used between these runs, right?

SurbhiJainUSC:
> > > Thank you Surbhi. Can you please also share the before profiles in the description? So we can check that the HLOs are identical
> > 
> > 
> > @bvandermoon - I have added before & after profiles.
> 
> Thanks @SurbhiJainUSC. The gemma profiles look good, but it looks like the gemma2 profiles have different HLOs. The general memory usage still looks very similar though. Could you check if anything is standing out as different in the gemma2 logic? I took a look but didn't see anything standing out
> 
> I assume the same JAX version, etc. were used between these runs, right?

Yes, I am using the same dependencies for both the runs.

SurbhiJainUSC:
> > > Thank you Surbhi. Can you please also share the before profiles in the description? So we can check that the HLOs are identical
> > 
> > 
> > @bvandermoon - I have added before & after profiles.
> 
> Thanks @SurbhiJainUSC. The gemma profiles look good, but it looks like the gemma2 profiles have different HLOs. The general memory usage still looks very similar though. Could you check if anything is standing out as different in the gemma2 logic? I took a look but didn't see anything standing out
> 
> I assume the same JAX version, etc. were used between these runs, right?

Found the issue. Had to set `float32_qk_product=True` and `float32_logits=True` instead of getting their values from config.

Here is the updated profile with this change: http://shortn/_GNZXxUvODi. HLO in this profiler (after NNX) matches with profile (without NNX): http://shortn/_R75pqHR9kt

src/MaxText/layers/gemma.py
```diff
@@ -14,30 +14,109 @@
...
+        inputs_kv_shape=dummy_inputs_shape,
+        mesh=self.mesh,
+        dtype=config.dtype,
+        weight_dtype=config.weight_dtype,
+        dropout_rate=config.dropout_rate,
+        float32_qk_product=config.float32_qk_product,
+        float32_logits=config.float32_logits,
+        quant=self.quant,
+        kv_quant=quantizations.configure_kv_quant(config),
+        use_ragged_attention=config.use_ragged_attention,
```
NuojCheng, line 89: I added `'reshape_q=config.reshape_q` input, not sure it is necessary for Gemma

src/MaxText/layers/gemma.py
```diff
@@ -14,30 +14,109 @@
...
+        ragged_block_size=config.ragged_block_size,
+        model_mode=self.model_mode,
+        rngs=self.rngs,
+    )
+
+    self.pre_ffw_norm = RMSNorm(
+        num_features=config.emb_dim,
+        dtype=config.dtype,
+        weight_dtype=config.weight_dtype,
+        kernel_axes=("norm",),
```
NuojCheng, line 99: I have `epsilon=config.normalization_layer_epsilon,`

src/MaxText/layers/gemma.py
```diff
@@ -98,46 +145,26 @@ def __call__(
         model_mode=model_mode,
     )
```
NuojCheng, line 146: I added slot=slot, page_state=page_state configs for self_attention, not sure if paged attention has been deprecated or not

src/MaxText/layers/gemma.py
SurbhiJainUSC: There is no epsilon in old version:
```
attn_output = rms_norm(
        num_features=attention_lnx.shape[-1],
        dtype=cfg.dtype,
        weight_dtype=cfg.weight_dtype,
        name="pre_ffw_norm",
        kernel_axes=("norm",),
    )(attention_lnx)
```

NuojCheng (approved):
LGTM, just with some nits. Thank you Surbhi!

src/MaxText/layers/gemma.py
SurbhiJainUSC: Similarly, I don't see this in old code either.

NuojCheng: I copied that from LLama2 decoder NNX migration PR.  It looks like an inference-specific feature https://github.com/AI-Hypercomputer/maxtext/blob/ee917cdcd6e6fd0dce7b9813f2fa54c060c9194e/src/MaxText/layers/attention_op.py#L1259 and unclear when it was added or whether it ever get used.

bvandermoon (approved):

src/MaxText/layers/gemma.py
SurbhiJainUSC: I checked Llama2 decoder layer PR, and `reshape_q` argument was present in linen version of Llama2. I don't see this in Gemma linen, so I don't think so if there is a need to add that.

PR #2174, opened 2025-08-14
Break up attentions.py

richjames0:
# Description

This PR breaks MLA and AttentionOp out of attentions.py as that file was unwieldy, difficult to reason about and difficult to apply significant changes to (e.g. NNX migration).

# Tests

Relevant tests have been updated and all pass.

# Checklist


MaxText/common_types.py
```diff
@@ -90,3 +90,11 @@ class DecoderBlockType(enum.Enum):
...
   SIMPLE_MLP = "simple_mlp"
   LLAMA4 = "llama4"
+
+
+class AttentionType(enum.Enum):
+  GLOBAL = "global"  # default, with causality
+  LOCAL_SLIDING = "local_sliding"
+  CHUNK = "chunk"
+  MLA = "mla"
+  FULL = "full"
```
gobbleturk, line 100: off the bat my question is what is the difference between global and full? Full means no causality?

gobbleturk: maybe include a comment here?

richjames0: this is a great question but one for @gagika - I just moved it to common_types from attentions :D

RissyRan: I have a sense that is bidirectional mask for vision part.

MaxText/layers/attention_mla.py
```diff
@@ -0,0 +1,719 @@
+#  Copyright 2023 Google LLC
```
RissyRan, line 1: Nit: 2025?

gagika (approved):
thanks

MaxText/common_types.py
gagika: yes, full has bidirectional mask (it is used vision soft tokens).

PR #2175, opened 2025-08-14
Add valid test cases list for sharding tests

hsuan-lun-chiang:
# Description

This PR

- Add cpu_only mark ensure the test runs on a cpu runner.
- Add code to generate a valid test cases list (as a json file) to resolve the ambiguity between a missing golden sharding file and an invalid golden sharding for a specific topology / model combination.
- Change relative path of golden shardings in test code.

# Tests

Generate golden shardings and valid test cases list:

```
python -m MaxText.tests.run_sharding_dump
```

Run sharding test:

```
python -m pytest MaxText/tests/sharding_compare_test.py
```

# Checklist


tests/sharding_dump.py
```diff
@@ -200,10 +200,36 @@
     # "a3"
 ]
 
-SLICES = [1, 4, 8192]
+SLICES = [
```
shralex, line 215: It seems like this and the previous static lists aren't being used anymore. Can we remove them ?

tests/sharding_dump.py
```diff
@@ -250,10 +276,9 @@ def save_named_sharding_dict(output_path: str | Path, sharding_dict: dict) -> No
     json.dump(sharding_dict, f, indent=2)
 
 
-def load_named_sharding_json(json_path: str | Path) -> dict:
+def load_named_sharding_json(json_path: Path) -> dict:
```
shralex, line 291: Can we please add a top level comment in this file explaining what it does

tests/sharding_compare_test.py
```diff
@@ -78,34 +72,43 @@ def compare_named_sharding_jsons(json1: dict, model1_name: str, json2: dict, mod
   return not only_in_1 and not only_in_2 and all(json1[k] == json2[k] for k in shared_keys)
 
 
-@pytest.mark.parametrize("model_name, topology, num_slice", TEST_CASES)
+@pytest.mark.cpu_only
+@pytest.mark.parametrize("model_name, topology, num_slice", valid_test_cases())
```
shralex, line 83: can we add a comment somewhere saying that sharding_info/ stores the list of valid (model_name, topology, num_slice) tuples and the expected sharding for various combinations.

tests/sharding_dump.py
shralex: please also mention the new json file where configurations need to be added

tests/sharding_dump.py
hsuan-lun-chiang: Hi, they are still used in run_sharding_dump.py to regenerate valid test cases whenever we need to update golden shardings.

shralex: is that necessary now that you have the combinations in the json file ?

PR #2178, opened 2025-08-14
Migrate LlamaDecoderLayer to NNX

bvandermoon:
# Description

Migrate LlamaDecoderLayer to NNX instead of Linen

* Change module type to NNX
* Initialize submodules in `__init__` instead of `__call__`
* Pass `self.model_mode` to module constructors instead of the one passed to call
* New `to_linen_class` NNX bridge to handle the way we pass decoder classes around. This is needed for checkpointing to work
* New init/apply functions in Transformer module to ensure the proper `model_mode` value is passed around

# Tests

commands to reproduce.

# Checklist


MaxText/layers/llama2.py
```diff
@@ -21,34 +21,116 @@
...
+    batch_size = 1 if model_mode == MODEL_MODE_PREFILL else config.global_batch_size_to_train_on
+
+    if model_mode == MODEL_MODE_PREFILL:
+      seq_len = config.max_prefill_predict_length
+    elif model_mode == MODEL_MODE_AUTOREGRESSIVE:
+      seq_len = 1
+    else:
+      seq_len = config.max_target_length
+
+    dummy_inputs_shape = (batch_size, seq_len, config.emb_dim)
```
bvandermoon, line 201: These can be different: https://github.com/AI-Hypercomputer/maxtext/blob/f0ffb2d6b57a1a5177a12a6dc253edec20b01d8d/MaxText/pyconfig.py#L590

@gobbleturk do you know the right one to use here? Seems they are the same for me for Llama2-7B. cc: @NuojCheng

MaxText/layers/decoders.py
```diff
@@ -85,12 +85,12 @@ def __call__(
...
+    if self.model_mode == MODEL_MODE_PREFILL:
       logical_axis_names = ("activation_batch", "prefill_activation_length", "activation_embed")
     else:
       logical_axis_names = ("activation_batch", "activation_length", "activation_embed")
 
-    if model_mode == MODEL_MODE_PREFILL:
+    if self.model_mode == MODEL_MODE_PREFILL:
       inputs = nn.with_logical_constraint(inputs, logical_axis_names)
     else:
       inputs = nn.with_logical_constraint(inputs, logical_axis_names)
```
gagika: since you are touching this part, could you also fix it.

if/else blocks are doing the same thing, we can just call (without if/else):
`inputs = nn.with_logical_constraint(inputs, logical_axis_names)`

MaxText/layers/decoders.py
```diff
@@ -105,7 +105,7 @@ def __call__(
         epsilon=cfg.normalization_layer_epsilon,
         kernel_axes=("norm",),
     )(inputs)
-    if model_mode == MODEL_MODE_PREFILL:
+    if self.model_mode == MODEL_MODE_PREFILL:
       lnx = nn.with_logical_constraint(lnx, logical_axis_names)
     else:
       lnx = nn.with_logical_constraint(lnx, logical_axis_names)
```
gagika: same as above, no need for if/else.

MaxText/layers/models.py
```diff
@@ -54,6 +54,15 @@ class Transformer(nn.Module):
...
 
+  def init(self, *args, model_mode: str = MODEL_MODE_TRAIN, **kwargs):
+    """Initializes the model."""
+    module = self.clone(model_mode=model_mode)
+    return nn.Module.init(module, *args, **kwargs)
+
+  def apply(self, *args, model_mode: str = MODEL_MODE_TRAIN, **kwargs):
+    """Initializes the model."""
+    module = self.clone(model_mode=model_mode)
+    return nn.Module.apply(module, *args, **kwargs)
```
gagika: Could you comment why those functions with clones are needed?

MaxText/layers/llama2.py
```diff
@@ -61,7 +143,6 @@ def __call__(
       previous_chunk=None,
   ):
     cfg = self.config
-    mesh = self.mesh
 
     if model_mode == MODEL_MODE_PREFILL:
```
gagika: since your PR made `model_mode` a static argument, passing it in both `__init__` and `__call__` can be confusing. You could move `activation_axis_names` initialization in __init__ function (or depend model_mode from config).

if model_mode == MODEL_MODE_PREFILL:
      activation_axis_names = ("activation_batch", "prefill_activation_norm_length", "activation_embed")
else:
      activation_axis_names = ("activation_batch", "activation_norm_length", "activation_embed")

gagika: Feel free to keep it as if if we expect model_mode to change during runtime, e.g. going from prefill to autoregressive decode.

cc @cgarciae

MaxText/layers/llama2.py
gobbleturk: emb_dim is the right one to use

MaxText/layers/llama2.py
```diff
@@ -21,34 +21,116 @@
...
+    batch_size = config.micro_batch_size_to_train_on
+
+    if model_mode == MODEL_MODE_PREFILL:
+      seq_len = config.max_prefill_predict_length
+    elif model_mode == MODEL_MODE_AUTOREGRESSIVE:
+      seq_len = 1
+    else:
+      seq_len = config.max_target_length
+
+    dummy_inputs_shape = (batch_size, seq_len, config.emb_dim)
```
dubstack, line 201: I believe you want to use a shape that is sharded along the batch dim?

cgarciae: @dubstack can you clarify what you mean about "sharded shape"?

MaxText/layers/llama2.py
bvandermoon: Thank you @gobbleturk

MaxText/layers/llama2.py
bvandermoon: Thanks @gagika. Updated so that `activation_axis_names` is set in `__init__` now. As a P1, we will want to remove the `model_mode` that is passed to `__call__` later

MaxText/layers/models.py
bvandermoon: Added a code comment for these. They are needed to ensure the same `model_mode` is passed to `__init__` and `__call__`

MaxText/layers/decoders.py
bvandermoon: Updated, thanks @gagika

MaxText/layers/decoders.py
bvandermoon: Updated, thanks @gagika

PR #2183, opened 2025-08-15
Fix: Create the gcs folder when creating Orbax EMC checkpointer

RexBearIU:
# Description

The Orbax EMC checkpointer in its current state does not automatically create necessary subdirectories within a GCS bucket, leading to a file not found error when a training job attempts to save the 0 checkpoint. This fix ensures that the checkpointer can create the required subfolders on GCS before attempting to write a checkpoint.

FIXES: b/437830796

# Tests

This fix has been validated with a dedicated test case that reproduces the original bug. Details and results can be found in the [Buganizer ticket](https://buganizer.corp.google.com/issues/437830796#comment1) and the [Original Buganizer ticket](436674299)

# Checklist


RexBearIU:
The problem is not existed when using Maxtext and Orbax HEAD.

PR #2188, opened 2025-08-15
Self Debugging Agent

wei879-100:
# Description
This PR introduces the Self-Debugging Agent, a new experimental tool within MaxText that automates the process of converting PyTorch code to JAX. The agent not only generates JAX code but also creates test cases, executes them, and autonomously debugs the generated code if tests fail, using Gemini to interpret errors and propose fixes. Manually converting code between deep learning frameworks is a tedious and error-prone process. While LLMs can automate the initial conversion, the generated code often contains subtle bugs, syntax errors, or incompatibilities that require significant manual debugging. The primary problem is the high failure rate and manual effort required in automated code conversion. A simple "translate and pray" approach is often insufficient. This agent addresses this by creating a closed-loop system where the model that generates the code is also responsible for fixing it, using the concrete feedback from test failures. This solution is powerful because it leverages the test suite as a verification and feedback mechanism for the LLM. By providing the model with the original code, the failing JAX code, the test case, and the resulting stack trace, the agent enables the LLM to make targeted, context-aware corrections. This significantly increases the probability of a successful, fully-verified code conversion without human intervention.

This agent's workflow is orchestrated by `self_debugging_agent.py` and follows these steps:

**Code Generation:** Converts a PyTorch file to JAX using `convert_code_from_torch_to_jax`, with retries to fix initial syntax errors. This step is performed by the code generation agent.
**Test Case Generation:** Gemini generates pytest to compare the PyTorch and JAX implementations. This step is performed by the code evaluation agent.
**Test Execution:** The test is run, and the output (pass/fail status and stack trace) is captured.
**Self-Debugging:** If tests fail, the code_debugging function is invoked. It uses a specialized prompt from `prompt_debugging.py` to feed the error context back to the Gemini, which then attempts to fix the JAX code and/or the test case. This loop continues for a configurable number of tries.
**Logging:** Final results are logged, and overall file-level and test-case-level accuracy metrics are computed.

# Tests
The agent was tested by running it on a dataset of standalone PyTorch modules. The primary method of testing is to execute the agent itself and observe its success rate in converting and debugging the files.

To reproduce the tests, you can run the agent from the command line.

Prerequisites:

Ensure all dependencies are installed.

1. Set up your `.env` file with a GOOGLE_API_KEY and Model for the `GeminiAgent`.

2. Place the source PyTorch files in the directory specified by the `--pytorch_path argument` (e.g., `MaxText/experimental/agent/code_generation_agent/dataset/PyTorch/`).

Execution Commands:

To run the agent on all PyTorch files in the specified directory:
```python MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py```
To run the agent on a single file:
```python MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py --module_name <your_file_name>.py```
The agent's output will log the pass/fail status for each file and provide final accuracy metrics, which serve as the test result. Detailed instructions are also available in MaxText/experimental/agent/self_debugging_agent/README.md

[Here](https://paste.googleplex.com/4834291833634816) is a comparison of the final results from the self-debugging agent versus the initial output from the code generation agent.

# Checklist

### CODE REVIEW

File: MaxText/experimental/agent/code_generation_agent/llm_code_generation.py [hunk 1]
```diff
@@ -74,13 +74,12 @@
 
 load_dotenv()
 import os, logging, sys
-from llm_agent import GeminiAgent
 
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
NuojCheng, line 78: What the root dir for running these scripts? Do you have to add this line?

MaxText/experimental/agent/code_generation_agent/llm_code_generation.py
```diff
@@ -74,13 +74,12 @@
...
 
+from llm_agent import GeminiAgent  # If you are in debugging mode, please change to `from .llm_agent import GeminiAgent`
 from orchestration_agent.Utils import parse_python_code
-
-from prompt_code_generation import CodeGeneration
+from prompt_code_generation import CodeGeneration  # If you are in debugging mode, please change to `from .prompt_code_generation import CodeGeneration`
```
NuojCheng: It would be nice if we can avoid manually changing it but instead setting up a config flag for that...

MaxText/experimental/agent/self_debugging_agent/README.md
```diff
@@ -0,0 +1,73 @@
+# Self-Debugging Agent
+
+## Overview
+
+The Self-Debugging Agent is an experimental tool designed to automate the conversion of PyTorch code to JAX, generate corresponding test cases, and debug any issues that arise during the conversion or testing process. It leverages Gemini to facilitate code generation, test case creation, and debugging.
```
NuojCheng: I wonder why it is called `self-debugging agent` while its main function is to convert pytorch to JAX. It sounds more like a pytorch-to-jax agent with self-debugging feature.

MaxText/experimental/agent/self_debugging_agent/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+To start the self-debugging process, run the main script from the `self_debugging_agent` directory.
+
+To process all files in the source directory:
+```bash
+python self_debugging_agent.py
+```
+To process a single file:
+```bash
+python self_debugging_agent.py --module_name <file_name>.py
```
NuojCheng: `--file_name` instead of `--module_name`?

MaxText/experimental/agent/self_debugging_agent/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Configuration
+
+You can customize the agent's behavior using command-line arguments:
+
+-   `--pytorch_path`: Path to the directory containing source PyTorch files. (Default: `../code_generation_agent/dataset/PyTorch/`)
+-   `--jax_path`: Path to the directory where converted JAX files will be saved. (Default: `./dataset/jax_converted/`)
+-   `--testcase_path`: Path to the directory where generated test cases will be saved. (Default: `./dataset/test_cases/`)
+-   `--base_jax_path`: Base path for JAX files. (Default: `../code_generation_agent/dataset/jax_converted/`)
+-   `--base_testcase_path`: Base path for test cases. (Default: `../code_generation_agent/dataset/test_cases/`)
```
NuojCheng: Why we set `--testcase_path` and `--base_testcase_path` separately?

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
```diff
@@ -0,0 +1,449 @@
...
+-   `orchestration_agent/Utils.py`: Contains `parse_python_code` for extracting
+    code from LLM responses.
+-   `utils.py`: Provides utility functions like `save_in_file_and_check_code_syntax`
+    and `parse_json_response`.
+"""
+
+import argparse
+import os, logging, sys
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
NuojCheng: same with above, can we somehow avoid add this line?

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
```diff
@@ -0,0 +1,449 @@
...
+)
+parser.add_argument("--code_debug_error_tries", type=int, default=5, help="for how many times to tries in case code failed.")
+parser.add_argument(
+    "--code_generation_tries",
+    type=int,
+    default=2,
+    help="for how many times to tries in case code failed and debugging also failed.",
+)
+parser.add_argument("--error_penalty", type=int, default=10, help="Penalty for errors in test case generation or execution.")
+parser.add_argument("--module_name", type=str, nargs="+", help="Name of the module in case want to check one case only.")
```
NuojCheng: Based on following usages, module_name(s) should be a list.

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
```diff
@@ -0,0 +1,449 @@
...
+    type=str,
+    default="../code_generation_agent/dataset/jax_converted/",
+    help="Base path for JAX files.",
+)
+parser.add_argument(
+    "--base_testcase_path",
+    type=str,
+    default="../code_generation_agent/dataset/test_cases/",
+    help="Base path for test cases.",
+)
```
NuojCheng, line 132: again I don't know why we have `testcase_path` and `base_testcase_path`

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
```diff
@@ -0,0 +1,449 @@
...
+    with open(test_file_path, "w") as f:
+      f.write("import os,sys\nsys.path.append(os.path.abspath('..'))\n")
+      f.write(best_code["test_case_code"])
+    return best_code["passed"], best_code["failed"]
+
+  except Exception as e:
+    logger.error("Exception in code generation %s", e)
+    logger.error("The code file is %s", python_file.split("/")[-1])
+    # logger.error("The jax Code is %s", jax_code)
+    # logger.error("The testcase Code is %s", test_case_code)
```
NuojCheng: delete them?

MaxText/experimental/agent/code_evaluation_agent/utils.py
gobbleturk, line 1: I would avoid white space changes in otherwise untouched files if possible

MaxText/experimental/agent/self_debugging_agent/README.md
gobbleturk: I agree, great comment! Perhaps if the self-debugging is critical it can be called self_debugging_translation_agent

MaxText/experimental/agent/self_debugging_agent/README.md
```diff
@@ -0,0 +1,73 @@
...
+-   **`code_evaluation_agent/prompt_code_evaluation.py`**: Contains prompts for generating the initial `pytest` test cases.
+-   **`code_evaluation_agent/utils.py`**: Provides utilities for executing `pytest` and capturing its results (`run_pytest_capture_output`) and for identifying the primary component in a code file (`get_last_defined_module`).
+-   **`code_generation_agent/llm_agent.py`**: Contains the `GeminiAgent` class for interacting with the Google Gemini model.
+-   **`orchestration_agent/Utils.py`**: Contains the `parse_python_code` utility for extracting code blocks from LLM responses.
+
+## Usage
+
+### Setup
+
+1.  **Install Dependencies**: Ensure all required packages from the parent directories are available in your Python environment.
```
gobbleturk, line 37: Does this include pytorch dependencies? Can you show example instructions to install dependencies?

MaxText/experimental/agent/self_debugging_agent/README.md
```diff
@@ -0,0 +1,73 @@
...
+2.  **Configure Environment**: Make sure your `.env` file is configured with the necessary `GOOGLE_API_KEY` and `Model` for the `GeminiAgent`.
+3.  **Prepare Data**:
+    *   Place your source PyTorch files in the directory specified by the `--pytorch_path` argument (default: `../code_generation_agent/dataset/PyTorch/`).
+    *   The agent will create output directories for JAX code and test cases if they don't exist.
+4.  **Modify the import in llm_code_generation.py**:
```
gobbleturk: Noujin may have recommended this as well, but I think we should try to avoid users having to modify code and especially import statements if possible

MaxText/experimental/agent/code_generation_agent/llm_code_generation.py
wei879-100: Great, I added the root dir

MaxText/experimental/agent/code_generation_agent/llm_code_generation.py
wei879-100: Yes, I changed, added the root dir

MaxText/experimental/agent/self_debugging_agent/README.md
wei879-100: I think I got some confusion in the description here, actually it analyze the error message and failed code and try to call the code generation agent to regenerate the code. I change the description accordingly.

MaxText/experimental/agent/self_debugging_agent/README.md
wei879-100: since during the code evaluation agent, when some test cases failed, it does not mean the JAX code converted has error, also might due to the test cases generated by Gemini. Therefore, the debugging agent is debug for both test cases generated and the JAX code generated. Here, the testcase_path means the new debugged test cases. base_testcase_path means the old test cases generated.

MaxText/experimental/agent/self_debugging_agent/README.md
wei879-100: No, it does not include pytorch dependencies. The error message will show some dependency needed to be installed such as the Google genai module.

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
wei879-100: Yes, I changed.

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
wei879-100: Yes, again, for the code evaluation agent, when some test cases failed, it does not mean the JAX code converted has error, also might due to the test cases generated by Gemini. Therefore, the debugging agent is debug for both test cases generated and the JAX code generated. Here, the testcase_path means the new debugged test cases. base_testcase_path means the old test cases generated.

MaxText/experimental/agent/self_debugging_agent/self_debugging_agent.py
wei879-100: module_name is intended to be a list of strings, and the use of nargs="+" in argparse ensures this is the case. Even if we only supply one module name on the command line, it will be stored in a list.

NuojCheng (approved):
LGTM

PR #2199, opened 2025-08-18
Fix `norm` axis for `tensor_sequence_parallelism`

phu0ngng:
# Description

The `norm` logical axis should not have `tensor_sequence`, as we don't partition the norm input along the hidden dimension in `tensor_sequence_parallelism`, nor does the `kernel` of norm, which is the `gamma`.
This causes the gamma to initialize distributedly, resulting in an additional all-gather on `gamma`.

# Checklist


richjames0 (approved):
lgtm

PR #2203, opened 2025-08-19
Support multimodal in logit checker + match gemma3 logits with HF

aireenmei:
# Description
I found some mismatch in gemma3 logits compared with HF (b/437988753).
Key changes to fill the gap:
* Match the vocab size: HF uses a bigger vocab size which includes special image tokens. 
* Special image token ID matches with HF
* Use linear Rope scaling with factor 8 for global attention layers in language model
* This will be a separate PR due to potential conflicts with the ongoing NNX migration: Allow setting gemma3 vision encoder precision through config.matmul_precision

# Tests
* add unittest in check_gemma3_layers.py to check RoPE implementation
* logits test:
Use generate_hf_golden_logits.py to generate HF logits on a m1 CPU, dtype=float32
Run forward_pass_logit_checker on a v5p TPU, with all the dype and precision set to float32
Results: see b/437988753

# Checklist


MaxText/layers/gemma3.py
hengtaoguo, line 1: Thank you Aireen for digging deep and finding this precision issue!

Is there any chance that you could split out the "precision" changes in gemma3.py into a separate PR? Asking because we are doing the NNX migration, and all of these layers will be rewritten from nn to nnx ([example](https://screenshot.googleplex.com/5E5APfEEBHCxbbY)). I guess we can plugin these "precision" changes after the migration? Either way works for me.

aireenmei: I see, yes I do notice that may create conflicting points with the current NNX migration. I will move them to a separate PR. Thanks Hengtao!

MaxText/configs/models/gemma3-12b.yml
```diff
@@ -21,7 +21,7 @@ base_num_kv_heads: 8
 base_mlp_dim: 15360
 head_dim: 256
 mlp_activations: ["gelu","linear"]
-vocab_size: 262_144
+vocab_size: 262_208
```
gagika: why do we need this change here an in other model configs?

doesn't this change impact checkpoint checkpoint conversion?

There embedding lookup and unembed layers depend on the vocab size.

aireenmei: Gemma3 has a text vocab_size of 262144, but use ids 262114-262207 for image tokens. The HF ckpt has output layer use vocab size of 262_208, which reflects on the output logits shape. Previously we truncated weights when converting HF ckpt (see where I add warnings in `MaxText/utils/ckpt_conversion/utils/param_mapping.py`).

hengtaoguo (approved):

MaxText/configs/models/gemma3-12b.yml
aireenmei: I reverted the vocab size change for now because changing vocab size requires updating other test scripts, we can do it later with a separate PR if needed. 
But it looks like changing vocab size does not impact the modality of the model because the diff tokens are placeholder image tokens (<start_of_image> is in text vocab), their embedding will be replaced by real image embedding. And we don't expect to output those tokens. 
So we may not need to change vocab size. I added code to truncate the output logits to a matching size in `forward_pass_logit_checker.py`, with loggings about the truncation. So we can still compare HF vs Maxtext logits.

gagika (approved):
Thanks for all the fixes!

PR #2208, opened 2025-08-20
Add Zero1+GA to improve performance

hx89:
Add Zero1+gradient-accumulation (GA) support in Maxtext to reduce the communication overhead and improve performance.

When Zero1+GA is enabled by setting shard_optimizer_over_data=True with DP, we will only have AG before the GA loop and AR after the GA loop, and there are no communication of gemm weights and activations within GA loop.

prevent_cse need to be set to False when GA > 1 to avoid extra ops generated by XLA.

### DIFF
--- a/src/MaxText/layers/decoders.py
+++ b/src/MaxText/layers/decoders.py
@@ -432,7 +432,7 @@ class Decoder(nn.Module):
       # Apply remat policy to layer
       layer = nn.remat(
           block_layer,
-          prevent_cse=not self.config.scan_layers,
+          prevent_cse=maxtext_utils.should_prevent_cse_in_remat(self.config),
           policy=policy,
           static_argnums=(4, 5),  # Deterministic and model mode are static arguments.
       )
--- a/src/MaxText/maxtext_utils.py
+++ b/src/MaxText/maxtext_utils.py
@@ -53,9 +53,9 @@ def get_input_data_sharding(config, mesh):
   return nn.logical_to_mesh_sharding(P(*config.input_data_sharding_logical_axes), mesh, config.logical_axis_rules)
 
 
-def get_functional_train_with_signature(train_step, data_sharding, state_mesh_shardings, model, config):
+def get_functional_train_with_signature(train_step, data_sharding, state_mesh_shardings, model, config, params_shardings=None):
   """Get the shardings (both state and data) for `train_step`."""
-  functional_train = functools.partial(train_step, model, config, state_mesh_shardings)
+  functional_train = functools.partial(train_step, model, config, state_mesh_shardings, params_shardings)
   functional_train.__name__ = "train_step"
   in_shardings = (state_mesh_shardings, data_sharding, None)  # State, batch, rng
   out_shardings = (state_mesh_shardings, None)  # State, metrics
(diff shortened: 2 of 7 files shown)


src/MaxText/train_utils.py
```diff
@@ -274,3 +281,13 @@ def validate_train_config(config):
         " different attention mechanism. With synthetic data, the format is not"
         " important as packing is not applied."
     )
+
+  # Automatically disable shardy when gradient accumulation is enabled
+  if config.gradient_accumulation_steps > 1 and config.shardy:
```
gobbleturk: can you move this logic to pyconfig?

gobbleturk: Can you also add hardware_type = "gpu" or "gpu_multiprocess" to this check? This particular issue is only seen on gpu

src/MaxText/train.py
```diff
@@ -210,12 +210,24 @@ def train_step(model, config, state_mesh_shardings, state, data, dropout_rng):
     extra_dpo_args = [reference_params]
     _loss_fn = dpo_loss_fn
 
+  params = state.params
+
   if config.gradient_accumulation_steps > 1:
+    def convert_to_bf16(param):
```
gobbleturk: Why not use param.astype(jnp.bfloat16) directly?

src/MaxText/maxtext_utils.py
```diff
@@ -938,6 +938,62 @@ def setup_training_state(model, data_iterator, tx, config, rng, mesh, checkpoint
...
+
+    if size % mesh.shape['data'] == 0 and (partition is None or 'tensor' not in partition):
+      added_component = ('data',) + partition
+      new_pspec = jax.sharding.PartitionSpec(*(pspec[:idx] + (added_component,) + pspec[idx+1:]))
+      new_sharding = jax.sharding.NamedSharding(sharding.mesh, new_pspec)
+      # return sharding.with_spec(new_pspec)
+      return new_sharding
+  return sharding
+
+def maybe_update_params_sharding_with_opt(config, state_mesh_shardings):
```
gobbleturk, line 1023: can you include a comment what this function does and what it returns?

src/MaxText/layers/decoders.py
```diff
@@ -427,7 +427,7 @@ def map_fn(path, value):
       # Apply remat policy to layer
       layer = nn.remat(
           block_layer,
-          prevent_cse=not self.config.scan_layers,
+          prevent_cse=not (self.config.scan_layers or self.config.gradient_accumulation_steps > 1),
```
gobbleturk: we need an additional check here that "config.hardware==gpu (or gpu_multiprocess)"

Given this is significant logic that is likely to change we should move it into a small function in maxtext utils, e.g.
def prevent_cse(config):
   return not self.config.scan_layers or (self.config.gradient_accumulation_steps > 1 and config.hardware==...)

src/MaxText/train.py
gobbleturk: I guess to handle fp8 guess?

src/MaxText/train.py
```diff
@@ -210,12 +210,24 @@ def train_step(model, config, state_mesh_shardings, state, data, dropout_rng):
     extra_dpo_args = [reference_params]
     _loss_fn = dpo_loss_fn
 
+  params = state.params
+
   if config.gradient_accumulation_steps > 1:
```
gobbleturk, line 217: doesn't have to be part of this PR but this makes train very hard to read - I think this amount of logic and code warrants an entire new file gradient_accumulation.py

src/MaxText/train.py
```diff
@@ -210,12 +210,24 @@ def train_step(model, config, state_mesh_shardings, state, data, dropout_rng):
...
+    def convert_to_bf16(param):
+      if param.dtype == jnp.float32:
+        return param.astype(jnp.bfloat16)
+      else:
+        return param
+    # Cast params to bf16 and unshard before gradient accumulation loop so that 
+    # all-gather is done once in bf16 
+    params_bf16 = jax.tree_util.tree_map(convert_to_bf16, params)
```
gobbleturk: This will only get run with gradient accumulation? Shouldn't this instead only get run when shard_optimizer_over_data is true?

src/MaxText/train.py
hx89: yes it's safer to makes sure the conversion is only done on fp32 param.

src/MaxText/maxtext_utils.py
```diff
@@ -938,6 +938,62 @@ def setup_training_state(model, data_iterator, tx, config, rng, mesh, checkpoint
...
+      partition = ()
+
+    if isinstance(partition, str):
+      partition = (partition,)
+
+    if size % mesh.shape['data'] == 0 and (partition is None or 'tensor' not in partition):
+      added_component = ('data',) + partition
+      new_pspec = jax.sharding.PartitionSpec(*(pspec[:idx] + (added_component,) + pspec[idx+1:]))
+      new_sharding = jax.sharding.NamedSharding(sharding.mesh, new_pspec)
+      # return sharding.with_spec(new_pspec)
```
NuojCheng: could you remove this line?

src/MaxText/train.py
NuojCheng: Agree. I will move them in a separately PR.

hx89:
Thanks for reviewing! I updated the PR to address the comments.

PR #2209, opened 2025-08-20
Fix for megablox, treat quantize as a bool

rdyro:
This should avoid rounding the entire last tile in the non-quantized path.

# Description

When masking the last reduction-axis tile, correctly decide on non-quantized path based on the bool argument. Previously the argument passed was a bool, but a None/not-None was expected within the function.

FIXES: b/439845640

### DIFF
--- a/MaxText/kernels/megablox/gmm.py
+++ b/MaxText/kernels/megablox/gmm.py
@@ -433,13 +433,13 @@ def gmm(
         def _init_out():
           out[...] = existing_out[...]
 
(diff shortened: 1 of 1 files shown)


RissyRan:
Thanks @rdyro! Testing it on my side.

gagika (approved):
thanks

RissyRan (approved):
I have seen loss is lower: https://screenshot.googleplex.com/BorZgA8Twwbo5Rx

RissyRan (approved):

PR #2215, opened 2025-08-20
Add stochastic decoding sampling strategy, allowing run-time sampling parameters change.

babyplutokurt:
# Description

This PR introduces a new "stochastic" decoding sampling strategy and enables the modification of sampling parameters at runtime.
Previously, the decoding sampling strategy and its associated parameters (e.g., top_k, top_p, temperature) were fixed by the static configuration in base.yml. This required a inference server restart to change sampling behavior. This change addresses that limitation by allowing these parameters to be specified on a per-request basis, providing greater flexibility for inference.

The new stochastic strategy combines top-k, top-p (nucleus), and temperature sampling in a sequential manner. The core generation methods in MaxEngine (prefill, generate, etc.) have been updated to accept algorithm, topk, nucleus_topp, and temperature as optional arguments. If these arguments are provided in a generation call, they override the static configuration for that specific request. If they are omitted, the system defaults to the values set in base.yml.

# Checklist


MaxText/inference_utils.py
```diff
@@ -143,6 +143,8 @@ def sampling(logits, rng, algorithm, topk=0, nucleus_topp=0, temperature=1.0):
     return sample_nucleus_topp_logits(logits, nucleus_topp, temperature, rng)
   elif algorithm == "topk":
     return sample_topk_logits(logits, topk, temperature, rng)
+  elif algorithm == "stochastic":
+    return sample_topk_topp_weighted(logits, topk, nucleus_topp, temperature, rng)
```
patemotter: I would make either the algorithm name or the function name more similar to the other one. I don't think it's wrong this way, but it's not immediately obvious to me that "stochastic" and "sample_topk_topp_weighted" are associated with each other. Or a little comment explaining how they are related is an option too. 

If you are feeling extra brave a little snippet comment for each of the algos could be very helpful. But that's not necessarily your responsibility.

MaxText/inference_utils.py
```diff
@@ -172,3 +174,49 @@ def sample_topk_logits(logits, topk, temperature, rng):
...
+
+def sample_topk_topp_weighted(logits, topk, nucleus_topp, temperature, rng):
+  """Combines top-k, top-p, and temperature sampling.
+
+  Sequence of operations:
+  1. Filter to top-k logits.
+  2. Filter the remaining logits by top-p (nucleus).
+  3. Apply temperature scaling.
+  4. Sample from the final distribution.
+  """
```
patemotter: This should probably have a full docstring describing the inputs and outputs.

MaxText/inference_utils.py
```diff
@@ -172,3 +174,49 @@ def sample_topk_logits(logits, topk, temperature, rng):
...
+  if topk <= 0:
+    raise ValueError(f"topk must be positive, got {topk=}")
+  if not (0.0 < nucleus_topp <= 1.0):
+    raise ValueError(f"nucleus_topp must be in (0, 1], got {nucleus_topp=}")
+
+  # 1. Top-K filtering
+  topk_logits, topk_idxs = jax.lax.top_k(logits, topk)
```
patemotter, line 216: Nit: I know MaxText already uses the `topk` formatting elsewhere so this just comes down to style preference, but it does seem needlessly confusing to mix it with the `top_k` formatting that we use from JAX. So this line has 3 of one kind and 1 of another.

MaxText/maxengine.py
```diff
@@ -529,22 +535,27 @@ def _prefill_jit(
         "generated_tokens": generated_tokens,
         "tokens": first_generated_token,
         "prompt_logp": prompt_logp,
+        "token_logp": token_logp,         # First sampled token logp is also needed
```
patemotter: Nit: Do we need this comment?

patemotter (commented):
Overall LGTM, some minor comments most of which are entirely cosmetic and optional.

MaxText/inference_utils.py
```diff
@@ -172,3 +174,49 @@ def sample_topk_logits(logits, topk, temperature, rng):
   topk_token = jnp.expand_dims(jax.random.categorical(rng, topk_logits / temperature).astype(jnp.int32), axis=-1)
   sampled_tokens = jnp.squeeze(jnp.take_along_axis(topk_idxs, topk_token, axis=-1), axis=-1).astype(jnp.int32)
   return sampled_tokens
+
+
+def sample_topk_topp_weighted(logits, topk, nucleus_topp, temperature, rng):
```
RissyRan, line 179: My understanding is that while server is running, users could pass in one config of any choice, and do the sampling? If so, is that possible to recall existing func, like `sample_topk_logits`, `sample_nucleus_topp_logits`, etc using if/else? 

Or users could use all sampling methods in one single benchmark? I didn't see early return after topk filter.

MaxText/inference_utils.py
RissyRan: Yes, I have the same feeling that "stochastic" that is not representative the combined algorithm. Maybe "dynamic" or "multi-mode"?

Also, missing greedy?

RissyRan (commented):
Could you add some unit tests for your new combined algorithm?

MaxText/maxengine.py
```diff
@@ -399,20 +399,26 @@ def prefill_aot(  # pylint: disable=too-many-positional-arguments
...
+      rng: PRNGKeyType | None = None,
+      slot: int | None = None,
+      page_state: PageState | None = None,
       return_prompt_logp: bool = False,
+      algorithm: str | None = None,
+      topk: int | None = None,
+      nucleus_topp: float | None = None,
+      temperature: float | None = None,
   ) -> tuple[Prefix, engine_api.ResultTokens]:
     """Computes a kv-cache for a new generate request.
```
jrplatin: Update the docstring here?

MaxText/maxengine.py
```diff
@@ -529,22 +535,27 @@ def _prefill_jit(
...
+      rng: PRNGKeyType | None = None,
+      request_id: uuid.UUID | None = None,  # pylint: disable=unused-argument
+      slot: int | None = None,
       return_prompt_logp: bool = False,
+      algorithm: str | None = None,
+      topk: int | None = None,
+      nucleus_topp: float | None = None,
+      temperature: float | None = None,
   ) -> tuple[Prefix, engine_api.ResultTokens]:
     """Public API for prefill that updates page state outside JIT."""
```
jrplatin, line 580: Since you're already editing this, I'd recommend adding a docstring

MaxText/inference_utils.py
babyplutokurt: Yeah I agree, but I intentionally used topk here to keep it consistent with the other functions in this file, like sample_topk_logits and sampling.

MaxText/tests/maxtext_utils_test.py
```diff
@@ -532,5 +532,116 @@ def test_respects_segments_and_masking(self):
...
+
+class TestSamplingFunctions(unittest.TestCase):
+  """Test suite for sampling utility functions."""
+
+  def setUp(self):
+    """Set up common logits and RNG for tests."""
+    self.rng = jax.random.PRNGKey(0)
+    # Logits with a clear ranking for a vocabulary of 10
+    self.logits = jnp.array([[0.1, 0.5, 0.2, 1.5, 0.8, 2.5, 0.3, 1.8, 0.7, 0.4]])
+    # Descending order of indices by logit value: [5, 7, 3, 4, 8, 1, 9, 6, 2, 0]
```
patemotter: Seems like this would make more sense to have as an actual class variable that is computed programmatically based on the values in self.logits. Then re-using it throughout the various unit tests. 

You can do this via `self.expected_order = jnp.argsort(self.logits, axis=None, descending=True)`.

patemotter (commented):
Overall LGTM, only changes are to the unit tests to make them a bit more resilient.

MaxText/inference_utils.py
```diff
@@ -172,3 +174,77 @@ def sample_topk_logits(logits, topk, temperature, rng):
...
+  3.  **Temperature scaling**: The logits of the filtered tokens are scaled
+      by the `temperature`. Higher temperatures result in a flatter
+      distribution (more randomness), while lower temperatures make the
+      distribution sharper (less randomness).
+  4.  **Sampling**: A token is sampled from the final probability
+      distribution using multinomial sampling.
+
+  Args:
+    logits: The unnormalized log probabilities of the vocabulary tokens,
+      with shape `[..., vocab_size]`.
```
RissyRan: Nit, let's specify `...` to be `batch, sequence`? Similar comment in the return.

MaxText/configs/base.yml
```diff
@@ -613,7 +613,7 @@ compiled_trainstep_file: "" # Name of saved serialized compiled train_step, e.g.
 compile_topology: '' # Target hardware version, e.g. 'v5e-256'
 compile_topology_num_slices: -1 # Number of target slices, set to a positive integer.
 
-decode_sampling_strategy: "greedy" # decode_sampling_strategy should be one of greedy, weighted, nucleus, or topk
+decode_sampling_strategy: "greedy" # decode_sampling_strategy should be one of greedy, weighted, nucleus, topk, or multinomial(top_k -> top_p -> weighted temperature)
```
RissyRan: Will `composite` be a better name here to indicate combination of 3 strategies?

MaxText/tests/maxtext_utils_test.py
```diff
@@ -532,5 +532,116 @@ def test_respects_segments_and_masking(self):
...
+      )
+    with self.assertRaises(ValueError):
+      inference_utils.sample_topk_topp_weighted(
+          self.logits, topk=-1, nucleus_topp=1.0, temperature=1.0, rng=self.rng
+      )
+    with self.assertRaises(ValueError):
+      inference_utils.sample_topk_topp_weighted(
+          self.logits, topk=10, nucleus_topp=0.0, temperature=1.0, rng=self.rng
+      )
+    with self.assertRaises(ValueError):
```
RissyRan, line 614: Shall we have assertion/test for negative temperature? i.e. invalid values.

babyplutokurt: negative_temperature is not allowed in sample_topk_topp_weighted, `scaled_logits = filtered_topk_logits / jnp.maximum(temperature, 1e-6)`

jrplatin (approved):

Issue #2207, opened 2025-08-19
Remat Policy `minimal` does not work for MoE models


kocchop:
### Bug report

The remat policy `minimal` which essentially uses `jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims` does not work with the expert gemms in MoE models because the expert gemms have einsum notation that has the `batch_dims`. One possible workaround is to use the remat policy `custom` and then manually specifying the different checkpoint names such as `mlpwi_0=device mlpwi_1=device mlpwo=device`.

For dense models, the `minimal` policy works.

<img width="1140" height="219" alt="Image" src="https://github.com/user-attachments/assets/985576ea-6586-4fea-9cf4-9f4af8bf7e2c" />

As we can see, for most part the `custom` remat policy with explicit checkpoint names' memory allocation requirement matches with defined remat policies except for `minimal` for the `mixtral-8x7b` case.

RissyRan:
Thanks for reporting the issue! Do you have repro commands to share? For instance, in the MoE, what's the values for `sparse_matmul`, `megablox`, `capacity_factor` etc?

gobbleturk:
ya this makes sense, thanks for filing! I think we should either remove minimal as an option, or change it to actually do the minimal remat by making use of "custom" under the hood and setting everything to device

kocchop:
> Thanks for reporting the issue! Do you have repro commands to share? For instance, in the MoE, what's the values for `sparse_matmul`, `megablox`, `capacity_factor` etc?

Hi @RissyRan I ran on 8xGPUs with FSDP=1 and EP=8 parallelism config and `sparse_matmul=false megablox=false capacity_factor=1.0`

For defined policy used `remat_policy=minimal_flash`
Equivalent with custom would be `remat_policy=custom query_proj=device key_proj=device value_proj=device context=device out_proj=device mlpwi_0=device mlpwi_1=device mlpwo=device`

RissyRan:
> ya this makes sense, thanks for filing! I think we should either remove minimal as an option, or change it to actually do the minimal remat by making use of "custom" under the hood and setting everything to device

```
      if cfg.remat_policy == "minimal":
        policy = jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims
```

[...]

PR #2217, opened 2025-08-21
Update remat rules

RissyRan:
# Description

To fix https://github.com/AI-Hypercomputer/maxtext/issues/2207:
* Remove `checkpoint_dots_with_no_batch_dims` with specific rules
* Update `minimal_flash` rule to `minimal_with_context` to align current naming
* Test changes accordingly

# Tests
Expect compile/unit tests pass.

# Checklist

### DIFF
--- a/MaxText/layers/decoders.py
+++ b/MaxText/layers/decoders.py
@@ -259,13 +259,34 @@ class Decoder(nn.Module):
           config=self.config, mesh=self.mesh, layers=pipeline_stage_module, remat_policy=remat_policy
       )
 
+  def minimal_policy(self, with_context=False):
+    """Helper for creating minimal checkpoint policies."""
+    names = [
+        "query_proj",
+        "value_proj",
+        "key_proj",
+        "qkv_proj",
+        "out_proj",
+        "mlpwi_0",
+        "mlpwi_1",
+        "mlpwi",
+        "mlpwo",
+    ]
+    if with_context:
+      names.append("context")
+    return jax.checkpoint_policies.save_only_these_names(*names)
+
   def get_remat_policy(self):
     """Get remat policy"""
     policy = None
     cfg = self.config
     if cfg.remat_policy != "none":
-      if cfg.remat_policy == "minimal":
-        policy = jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims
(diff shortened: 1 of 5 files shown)


MaxText/layers/decoders.py
```diff
@@ -264,8 +264,29 @@ def get_remat_policy(self):
     policy = None
     cfg = self.config
     if cfg.remat_policy != "none":
-      if cfg.remat_policy == "minimal":
-        policy = jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims
+      if cfg.remat_policy == "minimal_with_context":
```
khatwanimohit, line 284: Should we also include `mlpwi` in both these configs just to be safe and cater for cases where fused_mlp=true

RissyRan: Good point! I have updated it for [this](https://github.com/AI-Hypercomputer/maxtext/blob/7898576359bacde81be25cb3038e348aac1f943b/MaxText/layers/linears.py#L432).

khatwanimohit (approved):

MaxText/layers/decoders.py
```diff
@@ -264,8 +264,31 @@ def get_remat_policy(self):
...
```
gobbleturk, line 284: style nit: I would have small helper functions for these e.g.

def minimal_with_context_policy():
    return jax.checkpoint_policies.save_only_these_names(
            "query_proj",
            "value_proj",
            "key_proj",
            "qkv_proj",
            "context",
            "out_proj",
            "mlpwi_0",
            "mlpwi_1",
            "mlpwi",
            "mlpwo",
        )
        
       This should help keep the larger function get_remat_policy() easier to read and maintain

gobbleturk, line 284: Can include a comment that this is saving everything - or if you think this is obvious no need (same for other minimal policies)

gobbleturk (approved):
Thanks Ran!

MaxText/layers/decoders.py
RissyRan: Updated.

RissyRan: Added comment. Thanks!

PR #2222, opened 2025-08-21
Add an option to save a checkpoint on completion

xuefgu:
# Description

Add an option to save a checkpoint on completion. The option doesn't change the default behavior but allows skipping the last checkpoint-saving for various purposes (e.g. faster testing).

# Tests

Tested with the new flag being True and False and it works as intended.

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -44,6 +44,7 @@ load_full_state_path: ""
 # async_checkpointing is true, else a synchronous one is used. If you have
 # problems with the checkpointer we recommend trying the synchronous one.
 enable_checkpointing: True
+save_checkpoint_on_completion: True
 async_checkpointing: True
 checkpoint_period: 10_000
 # enables one replica to read the ckpt then broadcast to the rest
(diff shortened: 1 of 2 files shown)


NuojCheng (approved):

MaxText/configs/base.yml
```diff
@@ -44,6 +44,7 @@ load_full_state_path: ""
 # async_checkpointing is true, else a synchronous one is used. If you have
 # problems with the checkpointer we recommend trying the synchronous one.
 enable_checkpointing: True
+save_checkpoint_on_completion: True
```
bvandermoon, line 47: Has this been the default historically? I am wondering if it would be worthwhile to set the default to False. It seems to slow down a lot of testing having this enabled

bvandermoon (approved):

MaxText/configs/base.yml
xuefgu: It hadn't been the default but I also didn't want to change the existing behavior just yet. We can discuss later what the default behavior should be. For now, the option allows us to turn off the on-completion checkpointing.

PR #2241, opened 2025-08-26
Onboard GPT OSS model

RissyRan:
# Description

Onboard GPT OSS model:
* Add two GPT OSS model configs
* Add necessary configs in base.yml file (e.g., mlp_activations_limit, mlp_bias, attention_bias, attention_sink)
* Add sink feature for dot_product and enable for flash_attention
* Add decoder layers for scan and unscan features

**Next step: we will work on ckpt conversion and verify logits**

# Tests

* Unit test `check_gpt_vs_reference.py` against reference implementation for attention & mlp blocks
* End-to-end functional tests for [scan](https://paste.googleplex.com/6225469984473088) and [unscan](https://paste.googleplex.com/5797500786835456) on gpt-oss-20b model

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -138,6 +138,7 @@ base_mlp_dim: 7168
 base_num_decoder_layers: 16
 head_dim: 128
 mlp_activations: ["silu", "linear"]
+mlp_activations_limit: -1.0
 dropout_rate: 0.0
 logits_via_embedding: False
 normalize_embedding_logits: True  # whether to normalize pre-softmax logits if logits_via_embedding is true
@@ -183,7 +184,8 @@ first_num_dense_layers: 0 # number of initial dense layers in the model
 shared_experts: 1
 routed_scaling_factor: 1.0 # scaling factor for routing scores
 routed_score_func: "" # scoring function for routing
-routed_bias: False # a flag if a bias term is added for routing
+routed_bias: False # a flag if a learnable bias is added for routing
+mlp_bias: False # a flag if a learnable bias is added for MLP matmul
 n_routing_groups: -1 # number of groups for routing, disabled by default
 topk_routing_group: -1 # number of top groups to route inputs. For EP,
 # sending activations to a maximum of topk_routing_group distinct devices can yield performance benefits.
@@ -268,6 +270,8 @@ param_scan_axis: 1
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+attention_bias: False # If True, adds a learnable bias to the query, key, and value projections 
+attention_sink: False
 sliding_window_size: 0
 chunk_attn_window_size: 0
 attn_logits_soft_cap: 0.0
--- a/MaxText/configs/models/gpt-oss-120b.yml
+++ b/MaxText/configs/models/gpt-oss-120b.yml
@@ -1,0 +1,55 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# model config for gpt-oss-120b
+# https://huggingface.co/openai/gpt-oss-120b/blob/main/config.json
+
+# tokenizer_type: "huggingface"
+
+# Attention
+base_emb_dim: 2880
+base_num_query_heads: 64
+base_num_kv_heads: 8
+head_dim: 64
+sliding_window_size: 128
+attention_bias: True
+attention_sink: True
+
+# RoPE
+rope_type: "yarn"
+rope_max_timescale: 150_000
+max_position_embeddings: 131072
+original_max_position_embeddings: 4096
+rope_factor: 32
+beta_fast: 32
+beta_slow: 1
+
+# MLP
+base_mlp_dim: 2880
+base_moe_mlp_dim: 2880
+mlp_activations: ["sigmoid","linear"]
+mlp_activations_limit: 7.0
+routed_bias: True
+mlp_bias: True
+num_experts: 128
+num_experts_per_tok: 4
+
+# General
+base_num_decoder_layers: 6 #36
+vocab_size: 201088
+normalization_layer_epsilon: 1.0e-5
+enable_dropout: False
+logits_via_embedding: False
+decoder_block: "gpt_oss"
+inhomogeneous_layer_cycle_interval: 2
--- a/MaxText/layers/attention_op.py
+++ b/MaxText/layers/attention_op.py
@@ -20,6 +20,7 @@ from functools import partial
 import math
 
 import numpy as np
+from packaging import version
 
 from jax import lax
 from jax.ad_checkpoint import checkpoint_name
@@ -129,6 +130,12 @@ def apply_mask_to_logits(logits: Array, mask: Array):
     Masked logits.
   """
   return jnp.where((mask >= DEFAULT_MASK_VALUE * 0.5), logits, DEFAULT_MASK_VALUE)
+
+
+def validate_flash_attention_with_sinks_on_gpu(sinks: Array | None) -> None:
+  """Helper function to check for sinks with flash attention on GPU."""
+  if sinks is not None:
+    raise ValueError("The flash attention with sinks is not supported on GPU yet.")
 
 
 # TODO(agagik): change splash_attention_mask._ComputableMask to be non protected
@@ -677,6 +684,7 @@ class AttentionOp(nnx.Module):
       use_ragged_attention: bool = False,
       previous_chunk: Any = None,
       bidirectional_mask: Any = None,
+      sinks: Array = None,
       *,
       qk_product_einsum: Callable[..., Array],
       wv_product_einsum: Callable[..., Array],
@@ -712,6 +720,7 @@ class AttentionOp(nnx.Module):
           model_mode,
           previous_chunk,
           bidirectional_mask=bidirectional_mask,
+          sinks=sinks,
           qk_product_einsum=qk_product_einsum,
           wv_product_einsum=wv_product_einsum,
       )
@@ -727,8 +736,9 @@ class AttentionOp(nnx.Module):
               """Decode not supported with flash attention.
                               Use `dot_product` instead."""
           )
-        return self.tpu_flash_attention(query, key, value, decoder_segment_ids, self.attn_logits_soft_cap), None, None
+        return self.tpu_flash_attention(query, key, value, decoder_segment_ids, self.attn_logits_soft_cap, sinks), None, None
       else:
+        validate_flash_attention_with_sinks_on_gpu(sinks)
         if model_mode == MODEL_MODE_AUTOREGRESSIVE:
           # fallback to dot_product as pallas gpu flash attention doesn't support decode stage
           return self.apply_attention_dot(
@@ -763,6 +773,7 @@ class AttentionOp(nnx.Module):
           out = gpu_pallas_attention.mha(query, key, value, decoder_segment_ids, sm_scale=1.0, causal=True)
           return out, None, None
     elif self.attention_kernel == "cudnn_flash_te":
+      validate_flash_attention_with_sinks_on_gpu(sinks)
       if isinstance(key, KVTensor):
         key = key.dequant()
       if isinstance(value, KVTensor):
@@ -774,6 +785,7 @@ class AttentionOp(nnx.Module):
         )
       return self.cudnn_flash_attention(query, key, value, decoder_segment_ids, model_mode), None, None
     elif self.attention_kernel == "cudnn_flash_jax":
+      validate_flash_attention_with_sinks_on_gpu(sinks)
       if isinstance(key, KVTensor):
         key = key.dequant()
       if isinstance(value, KVTensor):
@@ -877,6 +889,7 @@ class AttentionOp(nnx.Module):
       value: Array,
       decoder_segment_ids: Array | None,
       attn_logits_soft_cap: float | None = None,
+      sinks: Array = None,
   ) -> Array:
     """TPU Flash Attention."""
 
@@ -1029,6 +1042,7 @@ class AttentionOp(nnx.Module):
             segment_axis_names_splash_kernel,
             None,  # no sharding for cp_size
             None,  # no sharding for load_balanced_context_parallel
+            None,  # no sharding for sinks
         ),
         out_specs=axis_names_q,
         check_rep=False,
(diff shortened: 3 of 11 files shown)


MaxText/layers/attention_op.py
```diff
@@ -1065,8 +1071,13 @@ def wrap_flash_attention(
           decoder_segment_ids_tuple = splash_attention_kernel.SegmentIds(decoder_segment_ids_q, decoder_segment_ids_kv)
       else:
         decoder_segment_ids_tuple = None
-      attention_output = jax.vmap(splash_kernel)(query, key, value, segment_ids=decoder_segment_ids_tuple)
-
+      # TODO(ranran): remove if/else branch once b/441336842 is fixed
+      if version.parse(jax.__version__) < version.parse("0.7.2.dev20250824"):
+        attention_output = jax.vmap(splash_kernel)(query, key, value, decoder_segment_ids_tuple)
```
shuningjin, line 1085: This seems to silently ignore attention sink. Shall we add validation in pyconfig.py when using attention sink and flash attention?
```
if raw_key["attention_sink"] and raw_key["attention"] == "flash" and version.parse(jax.__version__) < version.parse("0.7.2.dev20250824"):
  raise ValueError(...)
```

RissyRan: Most of tests will fail if we add this validation. Currently, the github runner are using 0.7.0 JAX version in the candidate image, and the flash attention feature with sinks is only supports starting 0.7.2 dev. That's the reason we have this branch here :)

MaxText/configs/models/gpt-oss-120b.yml
```diff
@@ -0,0 +1,56 @@
...
+num_experts: 128
+num_experts_per_tok: 4
+
+# General
+base_num_decoder_layers: 36
+vocab_size: 201088
+normalization_layer_epsilon: 1.0e-5
+enable_dropout: False
+logits_via_embedding: False
+use_bias: True
```
shuningjin: `ValueError: Key use_bias does not exist in config MaxText/configs/base.yml.`

MaxText/layers/attention_op.py
```diff
@@ -727,7 +730,7 @@ def apply_attention(
               """Decode not supported with flash attention.
                               Use `dot_product` instead."""
           )
-        return self.tpu_flash_attention(query, key, value, decoder_segment_ids, self.attn_logits_soft_cap), None, None
+        return self.tpu_flash_attention(query, key, value, decoder_segment_ids, self.attn_logits_soft_cap, sinks), None, None
       else:
         if model_mode == MODEL_MODE_AUTOREGRESSIVE:
```
gagika, line 742: seams attention sinks is not passed for MODEL_MODE_AUTOREGRESSIVE mode, could you add it too (can be follow up PR).

MaxText/layers/attention_op.py
```diff
@@ -677,6 +678,7 @@ def apply_attention(
       use_ragged_attention: bool = False,
       previous_chunk: Any = None,
       bidirectional_mask: Any = None,
+      sinks: Array = None,
```
gagika, line 687: could you error out if sinks is provided, but there is no implementation, e.g. for cudnn_flash

MaxText/layers/attention_op.py
```diff
@@ -1210,19 +1223,42 @@ def compute_local_attention(
...
-        local_max, (local_max.shape[0], local_max.shape[1], local_max.shape[2] * local_max.shape[3], 1)
-    )
-    local_sum = jnp.reshape(
-        local_sum, (local_sum.shape[0], local_sum.shape[1], local_sum.shape[2] * local_sum.shape[3], 1)
-    )
+      local_max = jnp.reshape(
+          local_max, (local_max.shape[0], local_max.shape[1], local_max.shape[2] * local_max.shape[3], 1)
+      )
+      local_sum = jnp.reshape(
+          local_sum, (local_sum.shape[0], local_sum.shape[1], local_sum.shape[2] * local_sum.shape[3], 1)
```
gagika: is there any repeated code to reuse in this if/else?

MaxText/tests/check_gpt_vs_reference.py
```diff
@@ -0,0 +1,427 @@
...
+            "wi_0_bias": to_jax(predefined_weights["experts.gate_up_proj_bias"][..., ::2]),
+            "wi_1": to_jax(predefined_weights["experts.gate_up_proj"][..., 1::2]),
+            "wi_1_bias": to_jax(predefined_weights["experts.gate_up_proj_bias"][..., 1::2]),
+            "wo": to_jax(predefined_weights["experts.down_proj"]),
+            "wo_bias": to_jax(predefined_weights["experts.down_proj_bias"]),
+        },
+    }
+    actual_output, _ = jax.jit(jax_model.apply)(moe_variables, jax_hidden_states)
+    normalized_expected = to_jax(expected_output) / jnp.linalg.norm(to_jax(expected_output))
+    normalized_actual = actual_output / jnp.linalg.norm(actual_output)
```
gagika, line 267: any reason of comparing normalized values and not unnormalized?

MaxText/configs/models/gpt-oss-120b.yml
RissyRan: Good catch! Will remove and test.

RissyRan: Test: https://paste.googleplex.com/4781335079813120 after removal.

MaxText/layers/attention_op.py
RissyRan: It seems that branch is GPU with flash? If so, I haven't tested flash on GPU yet. But dot_product should work out of box.

```
    elif self.attention_kernel in ("flash", "autoselected"):
      if target_hardware == "tpu":
        if isinstance(key, KVTensor):
          key = key.dequant()
        if isinstance(value, KVTensor):
          value = value.dequant()

        if model_mode == MODEL_MODE_AUTOREGRESSIVE:
          raise ValueError(
              """Decode not supported with flash attention.
                              Use `dot_product` instead."""
          )
        return self.tpu_flash_attention(query, key, value, decoder_segment_ids, self.attn_logits_soft_cap, sinks), None, None
      else:
        if model_mode == MODEL_MODE_AUTOREGRESSIVE:
          # fallback to dot_product as pallas gpu flash attention doesn't support decode stage
          return self.apply_attention_dot(
              query,
              key,
              value,
              decoder_segment_ids,
              model_mode,
              bidirectional_mask=bidirectional_mask,
              qk_product_einsum=qk_product_einsum,
              wv_product_einsum=wv_product_einsum,
          )
```

MaxText/tests/check_gpt_vs_reference.py
RissyRan: Good question. This mlp block output is much larger, around 20 or so, instead of regular ~1 with logits. So added a norm there to reduce the tolerance. Let me know your thoughts!

MaxText/layers/attention_op.py
RissyRan: Sounds good, will combine as one.

MaxText/layers/attention_op.py
RissyRan: Sure! Will raise the error accordingly.

MaxText/layers/attention_op.py
RissyRan: I have added check for GPU with flash attention using sinks.

MaxText/tests/check_gpt_vs_reference.py
gagika: ~1 seams high, but sometimes with Random weights it's higher than with trained weights.

is it with fp32 or bf16?

RissyRan: No, I guess I meant to say, the logits from reference is ~20-30, and delta is about 0.1-2ish, and tolerance is about 1e-1. But our regular logits output usually ~1, not ~20-30. I am using fp32 I think. So after normalizatoin, the tolerance can be 1e-3.

MaxText/layers/attention_op.py
```diff
@@ -1210,19 +1232,26 @@ def compute_local_attention(
...
-    )
+    b, n_kv, g, t, s = attn_weights.shape
+    n_q = n_kv * g
+    logits = jnp.reshape(attn_weights, (b, n_q, t, s))
+    if sinks is not None:
+      # broadcast sinks to match the attn weights dimension and combine
+      sinks_param = sinks.astype(attn_weights.dtype)  # (n_q,)
+      sinks_logits = sinks_param[jnp.newaxis, :, jnp.newaxis, jnp.newaxis]  # (1, n_q, 1, 1)
+      sinks_logits = jnp.broadcast_to(sinks_logits, (b, n_q, t, 1))
+      logits = jnp.concatenate([logits, sinks_logits], axis=-1)
```
gagika, line 1243: It would be good to see performance hit of adding 1 item (after concat, the last dimension becomes s+1 = 129, which won't be optimal on TPUs/GPUs).

We could also have an implementation without concatenation.

For local attention with small window (128) wondering if flash attention will be faster or dot product.

MaxText/layers/gpt_oss.py
```diff
@@ -0,0 +1,245 @@
...
+    cfg = self.config
+    mesh = self.mesh
+
+    inputs = nn.with_logical_constraint(inputs, ("activation_batch", "activation_norm_length", "activation_embed"))
+    inputs = checkpoint_name(inputs, "decoder_layer_input")
+    lnx_rms = rms_norm(
+        num_features=inputs.shape[-1],
+        dtype=cfg.dtype,
+        weight_dtype=cfg.weight_dtype,
+        name="pre_self_attention_layer_norm",
```
gagika: the reference implementation in HF or https://github.com/openai/gpt-oss/blob/main/gpt_oss/torch/model.py 
uses fp32 for  pre_self_attention_layer_norm and post_self_attention_layer_norm.

MaxText/layers/moe.py
```diff
@@ -909,18 +941,37 @@ def wrapper(x, logits, pre_bias_logits, w0, w1, wo):
...
+      if self.config.mlp_bias:
+        layer_w1 = layer_w1 + w1_bias
       if self.get_tensor_transpose_parallelism_size() > 1:
         layer_w1 = jax.lax.psum(layer_w1, "tensor_transpose")
       layer_w1 = adc.checkpoint_name(layer_w1, "mlpwi_1")
       # pylint: disable=protected-access
-      layer_act = linears._convert_to_activation_function(self.config.mlp_activations[0])(layer_w0)
-      intermediate_layer = jnp.multiply(layer_act, layer_w1)
+
+      if self.config.decoder_block == ctypes.DecoderBlockType.GPT_OSS:
```
gagika, line 962: it would be good to avoid model specific checks when possible and make it more about features e.g.

```
def swiglu_parametric(w0, w1, spec: MoeActivationSpec, dtype):
    # w0: "GLU" half, w1: "linear" half
    if spec.clamp_glu_max is not None:
        w0 = jnp.minimum(w0, spec.clamp_glu_max)
    if spec.clamp_lin_min is not None or spec.clamp_lin_max is not None:
        lo = -jnp.inf if spec.clamp_lin_min is None else spec.clamp_lin_min
        hi =  jnp.inf if spec.clamp_lin_max is None else spec.clamp_lin_max
        w1 = jnp.clip(w1, lo, hi)

    if spec.kind == "swiglu":
        act = jax.nn.sigmoid(spec.alpha * w0)
        out = (w0 * act) * (w1 + spec.linear_shift)
    else:
        # fall back to your existing activation pipeline
        act = linears._convert_to_activation_function(spec.kind)(w0)  # noqa
        out = act * w1

    return out.astype(dtype)
```

and use data class or our base config:

```
@dataclass(frozen=True)
class MoeActivationSpec:
    kind: str = "swiglu"              # "swiglu", "gelu", etc.
    alpha: float = 1.0                # 1.702 for GPT-OSS
    clamp_glu_max: float | None = None  # 7.0 for GPT-OSS; None = no clamp
    clamp_lin_min: float | None = None  # -7.0 for GPT-OSS
    clamp_lin_max: float | None = None  # 7.0 for GPT-OSS
    linear_shift: float = 0.0           # 1.0 for GPT-OSS
    
 ```

**Review by gagika — COMMENTED** [review]
A few more comments, if you agree, feel free to address them in a separate PR.

File: MaxText/layers/gpt_oss.py (hunk 8 above)
Comment (RissyRan) [reply MaxText/layers/gpt_oss.py]: Nice catch! Will update `weight_dtype=jnp.float32,`

File: MaxText/layers/attention_op.py (hunk 7 above)
Comment (RissyRan) [reply MaxText/layers/attention_op.py:1243]: Thanks! Let's discuss offline?

File: MaxText/layers/moe.py (hunk 9 above)
Comment (RissyRan) [reply MaxText/layers/moe.py:962]: Let's discuss offline?

File: MaxText/tests/check_gpt_vs_reference.py (hunk 6 above)
Comment (RissyRan) [reply MaxText/tests/check_gpt_vs_reference.py:267]: Also attach comparison without norm: [link](https://diff.googleplex.com/#key=jIvLuXvtrGqu)

File: MaxText/configs/base.yml [hunk 10]
```diff
@@ -268,6 +270,8 @@ param_scan_axis: 1
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+attention_bias: False # If True, adds a learnable bias to the query, key, and value projections
```
aireenmei, line 273: Do we need to expose this in base.yaml? If it's linked to model definition, we can hardcode in gpt_oss.py. For llama4, the vision encoder use bias, but text model does not, so we hardcoded [here](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/layers/llama4.py#L639). It may cause confusion which attention this is referring to, or we need one for each modality.

MaxText/configs/base.yml
```diff
@@ -268,6 +270,8 @@ param_scan_axis: 1
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+attention_bias: False # If True, adds a learnable bias to the query, key, and value projections 
+attention_sink: False
```
aireenmei, line 274: Similarly, do we need to expose this in base.yaml? I just read a bit about attention sink, maybe it's yes so user can apply it to other models.

MaxText/layers/attention_op.py
RissyRan: Discussed offline, I will have some tests during initial benchmarks with/without this sink feature.

MaxText/layers/moe.py
RissyRan: Discussed offline, we will do the refactor separately, and track in b/441837666

MaxText/configs/base.yml
RissyRan: Thanks Aireen for the review! I think both could work, but I am slightly prefer to expose this. I think this bias turn is also like attention_type, we could "hardcode" it in each model decoder layer. For instance, deepseek could be mla specific, mixtral could be just global. Yeah, expose configs could give more flexibility when customize models or let customers to train a new model without bias from config yml file.

MaxText/configs/base.yml
RissyRan: Yeah, similar to comments above. I think we could think it a way that it gives more flexibility when train other models, especially when MaxText is moving from reference implementation to a library. Please let me know your thoughts!

MaxText/configs/base.yml
aireenmei: Synced offline, we will add this for text model for now. We can add another for vit in the future if needed

aireenmei (approved):
Thanks Ran!

gagika (approved):
Thanks

PR #2242, opened 2025-08-26
Add vocabulary tiling to reduce redundant memory

NuojCheng:
# Description

This PR introduces **vocabulary tiling**, a new feature for MaxText designed to significantly reduce peak memory consumption during training.

This optimization is particularly beneficial for two key scenarios:

1.  **Training large-vocabulary models on limited devices:** It makes it possible to train models like Gemma 2B and Gemma-2 2B, which have large vocabulary sizes, on hardware with limited memory.
2.  **Training with long sequence lengths:** It makes training on long context lengths more feasible by avoiding the need to store the entire, large logits tensor in memory for the final loss computation.

The core idea of vocabulary tiling is to avoid explicitly materializing the full final logits tensor. Instead, the logits activation is chunked (or "tiled") along the batch-sequence dimension.

As illustrated in the diagram below, the forward and backward passes are repeated `num_vocab_tiling` times. In each iteration, a small slice of the logits is computed, used to calculate the loss, and the Vector-Jacobian product is immediately backpropagated. This iterative process avoids holding the complete, memory-intensive logits tensor.

Illustration of the vocabulary tiling process, showing chunked logits with repeated forward and backward passes.
*Figure 1: The vocabulary tiling process. The forward and backward passes are repeated for each tile, preventing the full logits tensor from being stored in memory.*

For a more in-depth technical explanation, please see the design document.

Doc: go/maxtext-vocab-tiling
FIXES: b/429255841

# Tests

## Correctness Tests

Test losses and embedding table gradient differences in `MaxText/tests/vocab_tiling_test.py` with 1\% relative error tolerance in following cases:
- Default non-tied embedding (`logits_via_embedding=False`).
- Tied embedding and default sharding (FSDP).
- Data parallelism.
- Tensor parallelism.
- Context parallelism.

## Performance Tests

See [doc](https://docs.google.com/document/d/1OdjQb-kVO2sVNbNyEwYmVmEFP9Tz_JdEria_aARnNe0/edit?resourcekey=0-xu47ZgKDZMCBxCGEgidnGA&tab=t.0#bookmark=id.3u2uxaamsc6x).

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -341,6 +341,7 @@ logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_batch_no_exp', ['data', 'fsdp', 'fsdp_transpose']],
                       ['activation_embed_and_logits_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
+                      ['activation_embed_and_logits_batch_sequence', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'expert']],
                       ['activation_heads', ['tensor', 'tensor_transpose', 'sequence','tensor_sequence','autoregressive']],
                       ['activation_kv_heads', ['tensor', 'tensor_transpose', 'sequence','tensor_sequence']],
                       ['activation_length', ['sequence', 'context', 'expert']],
@@ -451,6 +452,13 @@ ici_expert_parallelism: 1
 # The number of TPU slices is automatically determined, you should not set this explicitly. For ahead of time compilation,
 # you should set compile_toplogy_num_slices, which will in turn set this value. For non-TPU environments this is set to 1.
 num_slices: -1
+
+# Vocab Tiling Configs
+# Enables a memory-saving optimization by computing the cross-entropy loss in chunks.
+# The logits are tiled into `num_vocab_tiling` parts along the batch-sequence axis,
+# reducing peak memory usage. This is highly recommended for models with large
+# vocabularies (e.g., Gemma). Set to a value greater than 1 to enable.
+num_vocab_tiling: 1
 
 # Tokenizer
 vocab_size: 32_000 # powers of 2 for sharding
--- a/src/MaxText/layers/decoders.py
+++ b/src/MaxText/layers/decoders.py
@@ -610,7 +610,7 @@ class Decoder(nn.Module):
       )  # We do not quantize the logits matmul.
     if self.model_mode in (MODEL_MODE_PREFILL, MODEL_MODE_AUTOREGRESSIVE):
       logits = nn.with_logical_constraint(logits, (None, None, "activation_vocab"))
-    else:
+    elif cfg.num_vocab_tiling == 1:
       logits = nn.with_logical_constraint(
           logits, ("activation_embed_and_logits_batch", "activation_length_no_exp", "activation_vocab")
       )
@@ -825,7 +825,13 @@ class Decoder(nn.Module):
     # After the final transformer layer, `y` holds the raw, un-normalized hidden state.
     hidden_state = y
 
-    logits = self._apply_output_head(shared_embedding, hidden_state, deterministic)
+    # When vocab tiling is enabled in training mode, full logits won't generate to reduce memory
+    # Instead, we keep track on the hidden states, which has smaller size compared to full logits
+    if cfg.num_vocab_tiling > 1 and self.model_mode == MODEL_MODE_TRAIN:
+      logits = None
+      self.sow('intermediates', 'hidden_states', hidden_state)
+    else:
+      logits = self._apply_output_head(shared_embedding, hidden_state, deterministic)
 
     # The API of the Decoder is now a tuple, providing both the main output
     # and the raw hidden state needed for auxiliary tasks.
--- a/src/MaxText/layers/models.py
+++ b/src/MaxText/layers/models.py
@@ -94,6 +94,18 @@ class TransformerLinenPure(nn.Module):
           config=self.config, mesh=self.mesh, name="mtp_block", transformer_layer_module=mtp_layer, decoder=self.decoder
       )
 
+  def logits_from_hidden_states(self, hidden_states, deterministic):
+    """
+    Compute logits from hidden states (wrapping decoder._apply_output_head).
+    This function is only used for vocabulary tiling.
+    """
+    logits = self.decoder._apply_output_head(
+        shared_embedding=self.shared_embedding,
+        y=hidden_states,
+        deterministic=deterministic,
+    )
+    return logits
+
   def __call__(
       self,
       decoder_input_tokens: jnp.ndarray,
@@ -249,7 +261,7 @@ class Transformer(nnx.Module):
 
     decoder_linen = Decoder(config=cfg, mesh=mesh, quant=self.quant, model_mode=self.model_mode)
     self.decoder = nnx_wrappers.ToNNX(decoder_linen, rngs=rngs)
-
+    self.hidden_states = None
     if self.model_mode == MODEL_MODE_PREFILL:
       seq_len = cfg.max_prefill_predict_length
     elif self.model_mode == MODEL_MODE_AUTOREGRESSIVE:
@@ -348,6 +360,10 @@ class Transformer(nnx.Module):
         bidirectional_mask=bidirectional_mask,
         image_embeddings=image_embeddings,
     )
+
+    # Materialize hidden state when vocab tiling is enabled
+    if self.config.num_vocab_tiling > 1:
+      self.hidden_states = hidden_state
 
     # If we are initializing the model AND MTP is enabled, we must create
     # dummy target tensors. This allows Flax to trace the MTPBlock and create
--- a/src/MaxText/maxtext_utils.py
+++ b/src/MaxText/maxtext_utils.py
@@ -1415,3 +1415,182 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
   # Apply the constraint to the model's current variables. This tells JAX to
   # gather the weights into this layout.
   return jax.lax.with_sharding_constraint(variables, physical_constraint_no_fsdp)
+
+
+# Vocab Tiling Helper Functions
+def compute_loss_nnx(intermediate_outputs, logits, data, config, model, params, is_train):
+  one_hot_targets = jax.nn.one_hot(data["targets"], config.vocab_size)
+  xent, _ = max_utils.cross_entropy_with_logits(logits, one_hot_targets, 0.0)
+  xent = nn.with_logical_constraint(xent, ("activation_embed_and_logits_batch", "activation_length"))
+  # Mask out paddings at the end of each example.
+  xent = xent * (data["targets_segmentation"] != 0)
+  total_loss = jnp.sum(xent)
+  return total_loss
+
+
+def compute_loss_linen(intermediate_outputs, logits, data, config, model, params, is_train):
+  if config.num_vocab_tiling > 1:
+    hidden_state_key = ("intermediates", "decoder", "hidden_states")
+    hidden_states = get_nested_value(intermediate_outputs, hidden_state_key)[0]
+    total_loss = get_vocab_tiling_loss_linen(hidden_states, data, config, model, params, is_train)
+  else:
+    one_hot_targets = jax.nn.one_hot(data["targets"], config.vocab_size)
+    xent, _ = max_utils.cross_entropy_with_logits(logits, one_hot_targets, 0.0)
+    xent = nn.with_logical_constraint(xent, ("activation_embed_and_logits_batch", "activation_length"))
+    # Mask out paddings at the end of each example.
+    xent = xent * (data["targets_segmentation"] != 0)
+    total_loss = jnp.sum(xent)
+  return total_loss
+
+
+def get_vocab_tiling_loss_linen(
+    hidden_states,
+    data,
+    config,
+    model,
+    params,
+    is_train,
+):
+  labels = data["targets"]
+  segmentation = data["targets_segmentation"]
+  deterministic = not config.enable_dropout if is_train else True
+
+  param_spec= nn.get_partition_spec(params)
+  hidden_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(("activation_embed_and_logits_batch", "activation_length_no_exp", "activation_embed")))
+  label_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(("activation_embed_and_logits_batch", "activation_length_no_exp")))
+  reshaped_hidden_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(("num_tile", 'activation_embed_and_logits_batch_sequence', "activation_embed")))
+  reshaped_data_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(("num_tile", 'activation_embed_and_logits_batch_sequence')))
+  chunked_hidden_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(('activation_embed_and_logits_batch_sequence', "activation_embed")))
+  chunked_data_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(('activation_embed_and_logits_batch_sequence',)))
+  chunked_logits_spec = jax.sharding.NamedSharding(model.mesh, nn.logical_to_mesh_axes(('activation_embed_and_logits_batch_sequence', "activation_vocab")))
+  
+  hidden_states = jax.lax.with_sharding_constraint(hidden_states, hidden_spec)
+  labels = jax.lax.with_sharding_constraint(labels, label_spec)
+  segmentation = jax.lax.with_sharding_constraint(segmentation, label_spec)
+  #TODO (chengnuojin) all gather only embedding table instead of all params after NNX module is enabled
+  gathered_params = all_gather_over_fsdp(params, param_spec, model.mesh, config.logical_axis_rules)
+
+  # Customized forward and backward maps for the embedding tiling
+  @jax.custom_vjp
+  def chunked_cross_entropy_loss(gathered_params, hidden_states, labels, segmentation):
+      """
+      Calculates the total cross-entropy loss using vocab tiling.
+      """
+      total_loss, _ = _chunked_cross_entropy_loss_fwd(gathered_params, hidden_states, labels, segmentation)
+      return total_loss
+
+
+  def _chunked_cross_entropy_loss_fwd(gathered_params, hidden_states, labels, segmentation):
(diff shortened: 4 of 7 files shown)


MaxText/train.py
```diff
@@ -300,6 +304,119 @@ def move(path, value):
   return new_state, metrics
 
 
+def get_seq_tiling_loss(
```
gobbleturk: This should be defined in some other file (we want to keep train.py as clean as possible). Possibly in decoders.py or maxtext_utils.py

MaxText/configs/base.yml
```diff
@@ -440,6 +440,10 @@ ici_expert_parallelism: 1
 # you should set compile_toplogy_num_slices, which will in turn set this value. For non-TPU environments this is set to 1.
 num_slices: -1
 
+# Vocab Tiling to reduce memory usage but potentially increase step time
```
gobbleturk: I would include a bigger description here - reduce the memory of the vocab activations, especially useful when the vocab size is large

MaxText/train.py
```diff
@@ -300,6 +304,119 @@ def move(path, value):
...
+  @jax.custom_vjp
+  def chunked_cross_entropy_loss(params, reshaped_hidden_states, reshaped_labels, reshaped_segmentation):
+      """
+      Calculates the total cross-entropy loss using lax.scan
+      """
+      def fwd_scan_body(loss_accumulator, chunk_data):
+          hidden_chunk, label_chunk, segmentation_chunk = chunk_data
+          
+          # Calculate logits for the current chunk
+          chunk_logits = model.apply(
```
gobbleturk: Is it possible to call the method directly instead of use model.apply?

E.g. something like

chunk_logits = logits_from_hidden_states(hidden_state, params['vocab_unembedding'])? May have to modify the signature of logits_from_hidden_states. I'm not sure exactly how flax magic works here, perhaps the vocab_unembedding parameter doesn't exist yet

MaxText/train.py
NuojCheng: SG Will do

MaxText/configs/base.yml
NuojCheng: Agree!

MaxText/train.py
gobbleturk: (this might be easier with nnx)

NuojCheng: I tried replacing it but it is especially challenging for linen when `logits_via_embedding=False`. I agree that it would be easier for nnx.

NuojCheng: I will use this method before we fully migrating to NNX

src/MaxText/configs/base.yml
```diff
@@ -450,6 +451,13 @@ ici_expert_parallelism: 1
 # you should set compile_toplogy_num_slices, which will in turn set this value. For non-TPU environments this is set to 1.
 num_slices: -1
 
+# Vocab Tiling Configs
+# Enables a memory-saving optimization by computing the cross-entropy loss in chunks.
+# The logits are tiled into `num_vocab_tiling` parts along the vocabulary axis,
```
gobbleturk: Isn't it across the batch/sequence axis?

src/MaxText/layers/decoders.py
```diff
@@ -825,7 +825,11 @@ def __call__(
     # After the final transformer layer, `y` holds the raw, un-normalized hidden state.
     hidden_state = y
 
-    logits = self._apply_output_head(shared_embedding, hidden_state, deterministic)
+    if cfg.num_vocab_tiling > 1 and self.model_mode == MODEL_MODE_TRAIN:
```
richjames0, line 830: it's hard to know whether this is correct - could you add a comment to the code explaining what's happening please :)

src/MaxText/layers/models.py
```diff
@@ -94,6 +94,20 @@ def setup(self):
           config=self.config, mesh=self.mesh, name="mtp_block", transformer_layer_module=mtp_layer, decoder=self.decoder
       )
 
+  def logits_from_hidden_states(self, hidden_states, deterministic):
```
richjames0, line 97: i love breaking stuff into functions far more than the next man but is this function actually worthwhile since it just forwards to a single other function?

src/MaxText/layers/models.py
```diff
@@ -410,6 +441,20 @@ class ZeroOneTransformer(nn.Module):
   def setup(self):
     self.model = transformer_as_linen(self.config, self.mesh, self.quant, self.model_mode)
 
+  def logits_from_hidden_states(self, hidden_states, deterministic):
+    """
+    A pure function to compute logits from hidden states (tiled).
```
gobbleturk: Include a more helpful comment (we assume all functions are pure in jax) I don't think you actually scan this function, I don't think you can given its signature?

src/MaxText/layers/models.py
```diff
@@ -297,6 +311,19 @@ def no_op(self, *args, **kwargs):
   def init_cache(self, cache_size: int, batch_size: int, dtype=jnp.float32):
     return True
 
+  def logits_from_hidden_states(self, hidden_states, deterministic):
```
richjames0: i'm confused - is this function defined twice? here and above

src/MaxText/layers/models.py
```diff
@@ -410,6 +441,20 @@ class ZeroOneTransformer(nn.Module):
   def setup(self):
     self.model = transformer_as_linen(self.config, self.mesh, self.quant, self.model_mode)
 
+  def logits_from_hidden_states(self, hidden_states, deterministic):
```
richjames0: three times?

src/MaxText/layers/models.py
gobbleturk: All three should call a shared implementataiotn in max_utils

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
   # Apply the constraint to the model's current variables. This tells JAX to
   # gather the weights into this layout.
   return jax.lax.with_sharding_constraint(variables, physical_constraint_no_fsdp)
+
+
+# Vocab Tiling Helper Functions
```
richjames0, line 1420: @gobbleturk I feel like maxtext_utils (and max_utils for that matter) is/are turning into huge grab bags of random stuff with no structure. should we perhaps take this opportunity to break this out into e.g. maxtext_utils_tiling.py?

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+def compute_loss_nnx(intermediate_outputs, logits, data, config, model, params, is_train):
+  if config.num_vocab_tiling > 1:
```
gobbleturk: is it possible to catch this earlier in pyconfig?

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+    xent = nn.with_logical_constraint(xent, ("activation_embed_and_logits_batch", "activation_length"))
+    # Mask out paddings at the end of each example.
+    xent = xent * (data["targets_segmentation"] != 0)
+    total_loss = jnp.sum(xent)
+  return total_loss
+
+
+def compute_loss_linen(intermediate_outputs, logits, data, config, model, params, is_train):
+  if config.num_vocab_tiling > 1:
+    nested_key = ("intermediates", "decoder", "hidden_states")
```
gobbleturk: hidden_state_key

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+    hidden_states = get_nested_value(intermediate_outputs, nested_key)[0]
+    total_loss = get_vocab_tiling_loss_linen(hidden_states, data, config, model, params, is_train)
```
gobbleturk, line 1435: I like that this is modularized into a method get_vocab_tiling_loss_linen!

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+  
+  hidden_states = jax.lax.with_sharding_constraint(hidden_states, hidden_spec)
+  labels = jax.lax.with_sharding_constraint(labels, label_spec)
+  segmentation = jax.lax.with_sharding_constraint(segmentation, label_spec)
+  #TODO (chengnuojin) all gather only embedding table instead of all params after NNX module is enabled
+  gathered_params = all_gather_over_fsdp(params, param_spec, model.mesh, config.logical_axis_rules)
+
+  # Customized forward and backward maps for the embedding tiling
+  @jax.custom_vjp
+  def chunked_cross_entropy_loss(gathered_params, hidden_states, labels, segmentation):
```
richjames0, line 1475: is this function needed?

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+    hidden_states,
+    data,
+    config,
+    model,
+    params,
+    is_train,
+):
+  labels = data["targets"]
+  segmentation = data["targets_segmentation"]
+  deterministic = not config.enable_dropout if is_train else True
```
gobbleturk, line 1456: can we pass in deterministic instead of re-running this logic?

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+      """
+      Calculates the total cross-entropy loss using lax.scan
+      """
+      total_loss, _ = _chunked_cross_entropy_loss_fwd(gathered_params, hidden_states, labels, segmentation)
+      return total_loss
+
+
+  def _chunked_cross_entropy_loss_fwd(gathered_params, hidden_states, labels, segmentation):
+      batch_size, seq_len, emb_dim = hidden_states.shape
+      assert (batch_size * seq_len) % config.num_vocab_tiling == 0, "Per device batch size times sequence length should be divisible by the number of vocab tiles."
```
richjames0: i would raise instead of assert since this isn't a coding issue - it could be a config issue

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+  labels = jax.lax.with_sharding_constraint(labels, label_spec)
+  segmentation = jax.lax.with_sharding_constraint(segmentation, label_spec)
+  #TODO (chengnuojin) all gather only embedding table instead of all params after NNX module is enabled
+  gathered_params = all_gather_over_fsdp(params, param_spec, model.mesh, config.logical_axis_rules)
+
+  # Customized forward and backward maps for the embedding tiling
+  @jax.custom_vjp
+  def chunked_cross_entropy_loss(gathered_params, hidden_states, labels, segmentation):
+      """
+      Calculates the total cross-entropy loss using lax.scan
```
gobbleturk: eventually use scan or similar (this method isn't being scanned which is misleading)

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+          (reshaped_hidden_states, reshaped_labels, reshaped_segmentation)
+      )
+      residuals = (gathered_params, reshaped_hidden_states, reshaped_labels, reshaped_segmentation, batch_size, seq_len, emb_dim)
+
+      return total_loss, residuals
+
+  def _chunked_cross_entropy_loss_bwd(residuals, loss_cotangent):
+      gathered_params, reshaped_hidden_states, reshaped_labels, reshaped_segmentation, batch_size, seq_len, emb_dim = residuals
+
+      def bwd_scan_body(carry, chunk_data):
```
gobbleturk: include a comment - what is carry and chunk_data (What are these variables and what are there shapes)

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,179 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+      grad_params, grad_reshaped_hidden_states = jax.lax.scan(
+          bwd_scan_body,
+          initial_carry,
+          (reshaped_hidden_states, reshaped_labels, reshaped_segmentation)
+      )
+      grad_reshaped_hidden_states = jax.lax.with_sharding_constraint(grad_reshaped_hidden_states, reshaped_hidden_spec)
+      # Convert grad_params to bf16
+      grad_params = jax.tree_util.tree_map(lambda x, y: y.astype(x.dtype), gathered_params, grad_params)
+      # Chain-rule to accumulate gradients
+      grad_params = jax.tree_util.tree_map(lambda g: g * loss_cotangent, grad_params)
```
gobbleturk: is loss_cotangent 1?

src/MaxText/configs/base.yml
```diff
@@ -339,6 +339,7 @@ logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_batch_no_exp', ['data', 'fsdp', 'fsdp_transpose']],
                       ['activation_embed_and_logits_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
+                      ['activation_embed_and_logits_batch_sequence', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert', 'sequence', 'context']],
```
gobbleturk: I think we want to keep order the same as physical axes defined above - so expert should be last (not confident this is true)

src/MaxText/configs/base.yml
NuojCheng: Thank you for the catch. Will fix

src/MaxText/layers/models.py
NuojCheng: I will remove two of them and only keep one.

src/MaxText/maxtext_utils.py
NuojCheng: Will fix.

src/MaxText/maxtext_utils.py
NuojCheng: ty

src/MaxText/maxtext_utils.py
NuojCheng: I will keep as it is since we have not defined deterministic in train.py yet.

src/MaxText/layers/decoders.py
NuojCheng: Will do

src/MaxText/layers/models.py
NuojCheng: We have model: Transformer and model.decoder: Decoder. Currectly, MaxText implements the last unembedding step (mapping hidden states to logits) using a function '_apply_output_head' belonging to model.decoder: Decoder. Since I instead want to apply unembedding directly from model: Transformer, I have to write a seperate function 'logits_from_hidden_states' under the transformer as a wrapper so that I can easily call it. I am not sure if there is a better way to do so.

src/MaxText/layers/models.py
NuojCheng: I will only keep one of them, Sorry for the confusion.

src/MaxText/layers/models.py
NuojCheng: I will remove them. This is because there are three types of Transformers defined in model.py but actually only one of them get used.

src/MaxText/maxtext_utils.py
NuojCheng: I will keep the code in maxtext_utils.py  so far but let me know if a decision gets made. I agree with this comment.

src/MaxText/maxtext_utils.py
NuojCheng: Yes unfortunatly that's how custom_vjp is defined, for example https://github.com/AI-Hypercomputer/maxtext/blob/240a59aedf57088ed7e46367821524f66de50f3e/src/MaxText/max_utils.py#L525-L554

src/MaxText/maxtext_utils.py
NuojCheng: Good point! I will move it to the config check

src/MaxText/maxtext_utils.py
NuojCheng: thannk you, will update it

src/MaxText/maxtext_utils.py
NuojCheng: wil add

src/MaxText/maxtext_utils.py
NuojCheng: I checked its values (per step) as follow
Using jax.debug.print: The value of loss cotangent is 1.2481900739658158e-05
Using jax.debug.print: The value of loss cotangent is 1.270502707484411e-05
Using jax.debug.print: The value of loss cotangent is 1.2468983186408877e-05
Using jax.debug.print: The value of loss cotangent is 1.2645102287933696e-05
Using jax.debug.print: The value of loss cotangent is 1.3046995263721328e-05
Using jax.debug.print: The value of loss cotangent is 1.2260611583769787e-05
Using jax.debug.print: The value of loss cotangent is 1.3167250472179148e-05
My understand is that they are 1/(batch_size*seq_size), in this case are 48 and 2048, fluctuating with different segmentation values. This is because the output total_loss for the forward function is sum over all the cross entropy loss while the final loss is the average among them, so the loss_contangent is reflecting a scaler adjusting the total number of losses we consider (which is batch_size * seq_len).

src/MaxText/configs/base.yml
NuojCheng: Thank you. Will update.

src/MaxText/maxtext_utils.py
```diff
@@ -1415,3 +1415,183 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules):
...
+      reshaped_labels = jax.lax.with_sharding_constraint(reshaped_labels, reshaped_data_spec)
+      reshaped_segmentation = segmentation.reshape((config.num_vocab_tiling, vocab_tile_size))
+      reshaped_segmentation = jax.lax.with_sharding_constraint(reshaped_segmentation, reshaped_data_spec)
+      
+      # Scan body accumulates loss from each tile givne chunked hidden states and labels
+      def _fwd_scan_body(loss_accumulator, chunk_data):
+          hidden_chunk, label_chunk, segmentation_chunk = chunk_data
+          hidden_chunk = jax.lax.with_sharding_constraint(hidden_chunk, chunked_hidden_spec)
+          label_chunk = jax.lax.with_sharding_constraint(label_chunk, chunked_data_spec)
+          segmentation_chunk = jax.lax.with_sharding_constraint(segmentation_chunk, chunked_data_spec)
```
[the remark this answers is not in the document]
RissyRan: Nice catch! Gemini!

PR #2298, opened 2025-09-05
Ensure HF_TOKEN is set for checkpoint conversion tests on Airflow

SurbhiJainUSC:
# Description

This PR resolves a bug where the `HF_TOKEN` was not being correctly set as an environment variable, which was causing issues with Airflow tests. Before this fix, the Airflow tests would fail because they couldn't authenticate with the Hugging Face Hub. This was due to `HF_TOKEN` set to `""` in the bash scripts.

# Tests

Tested with local airflow server

# Checklist


end_to_end/tpu/gemma3/4b/test_gemma3_to_mt.sh
```diff
@@ -1,23 +1,26 @@
...
 set -ex
 idx=$(date +%Y-%m-%d-%H-%M)
 MODEL_NAME='gemma3-4b'
 export MODEL_VARIATION='4b'
-HF_TOKEN='' # Important!!! Save your hf access token here
 HF_GOLDEN_MODEL='google/gemma-3-4b-it'
 TOKENIZER_PATH="${MAXTEXT_ASSETS_ROOT:-${MAXTEXT_REPO_ROOT:-$PWD}/assets}"'/tokenizer.gemma3'
 # To convert the multimodal model, make sure the use_multimodal is set to be true
-USE_MULTIMODAL=true
+USE_MULTIMODAL=false
```
hengtaoguo, line 23: Could we leave this flag as true? Text-only is a subset of multimodal model weights. For the full (multimodal) checkpoint, if you set `use_multimodal=false`, it should behave the same way as text-only. So it would be great to keep the full ckpt conversion check here.

hengtaoguo (approved):
Thanks for the fix and tests!

end_to_end/tpu/gemma3/4b/test_gemma3_to_mt.sh
bvandermoon: Also in `end_to_end/tpu/gemma3/4b/test_gemma3_to_hf.sh`

bvandermoon (commented):
Where is the HF token coming from now? I think I am just missing it

end_to_end/tpu/gemma3/4b/test_gemma3_to_mt.sh
SurbhiJainUSC: This script was failing for me when I set `USE_MULTIMODAL=true`
```
AssertionError: Multimodal does not support running hf model on-the-fly, please generate hf golden logits using generate_hf_golden_logits.py
```

SurbhiJainUSC:
> Where is the HF token coming from now? I think I am just missing it

I have added a comment in the script that it is pre-requisite to set HF_TOKEN as an environment variable before running this script. This will help us to use these scripts to run on Airflow.

end_to_end/tpu/gemma3/4b/test_gemma3_to_mt.sh
SurbhiJainUSC: Had an offline discussion and it looks like forward pass logit checker is not compatible with multimodal. @hengtaoguo will work on adding the support, till then we will set `USE_MULTIMODAL=false`.

shralex (approved):

PR #2301, opened 2025-09-05
Enable padding_batch for data pipelines

aireenmei:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/MaxText/configs/base.yml
+++ b/MaxText/configs/base.yml
@@ -464,6 +464,8 @@ eval_data_columns: ['text'] # for DPO dataset containing "chosen" and "rejected"
 eval_image_column: 'image'
 packing: True
 num_epoch: 1  # only grain and tfds pipeline supports num_epoch > 1
+generate_padding_batch_train: False
+generate_padding_batch_eval: False
 
 # direct preference optimization (DPO)
 use_dpo: False
--- a/MaxText/data_loader.py
+++ b/MaxText/data_loader.py
@@ -35,6 +35,7 @@ class DataLoader:
   def __init__(self, config, mesh, data_iterator, goodput_recorder):
     self.config = config
     self.goodput_recorder = goodput_recorder
+    data_iterator.reset()
     self.data_iterator = data_iterator
     self.last_batch = None
     self.input_data_shardings = maxtext_utils.get_input_data_sharding(config, mesh)
@@ -51,10 +52,10 @@ class DataLoader:
         self.last_batch = jax.lax.with_sharding_constraint(example_batch, self.input_data_shardings)
         self.check_example_batch()
       except Exception as e:  # pylint: disable=broad-except
-        if "StopIteration" in str(e):
-          raise exceptions.StopTraining("You may have run out of training data.")
+        if isinstance(e, StopIteration):
+          raise exceptions.StopTraining(f"Received {type(e)} exception: ({e}), You may have run out of training data.")
         else:
-          raise exceptions.StopTraining(f"`load_next_batch()` failed ({e}).")
+          raise exceptions.StopTraining(f"`load_next_batch()` failed with {type(e)} exception: ({e}).")
     return self.last_batch
 
   def check_example_batch(self):
--- a/MaxText/input_pipeline/_grain_data_processing.py
+++ b/MaxText/input_pipeline/_grain_data_processing.py
@@ -33,7 +33,7 @@ from MaxText import tokenizer
 
 def find_data_files(data_file_pattern):
   data_files = glob.glob(str(Path(data_file_pattern).expanduser().resolve()))
-  assert len(data_files) > 0, f"No file found with pattern {data_file_pattern}."
+  assert len(data_files) > 0, f"No file found with pattern {data_file_pattern=}."
   max_logging.log(f"Found {len(data_files)} files for train/eval with grain")
   return data_files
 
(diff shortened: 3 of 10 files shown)


src/MaxText/train.py
```diff
@@ -372,7 +372,7 @@ def train_loop(config, recorder, state=None):
     state_mesh_shardings = _merge_dpo_state(state_mesh_shardings, state_mesh_shardings.params["params"])
 
   p_train_step, p_eval_step = train_utils.jit_train_and_eval_step(
-      config, model, mesh, state, state_mesh_shardings, train_step, eval_step, eval_data_iterator
```
SurbhiJainUSC: You might need to update other trainers too, for example: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/sft_trainer.py#L74

src/MaxText/multihost_dataloading.py
```diff
@@ -103,21 +77,72 @@ def __init__(self, dataloader: tf.data.Dataset | Iterable, global_mesh: Mesh):
...
+    if self.out_of_data and self.generate_padding_batch:
+      local_data = self._make_padding_batch()
+
+    else:
+      SLEEP_TIME = 10
+      MAX_DATA_LOAD_ATTEMPTS = 30
+
+      data_load_attempts = 0
+      loaded_data_success = False
+      while not loaded_data_success and data_load_attempts < MAX_DATA_LOAD_ATTEMPTS:
```
SurbhiJainUSC: Line 112-136 can be simplified:
```
for attempt in range(MAX_DATA_LOAD_ATTEMPTS):
   try:
        local_data = next(self.local_iterator)
        break # exit the loop on success
    except tf.errors.FailedPreconditionError as e:
        max_logging.log()
        time.sleep(SLEEP_TIME)
    except StopIteration as e and self.generate_padding_batch:
         max_logging.log()
         self.out_of_data = True
         local_data = self._make_padding_batch()
         break
    except Exception as e:
         raise e # raise exception to show full stack trace
```

Wdyt?

aireenmei: Great suggestion! I updated.

src/MaxText/train.py
aireenmei: Looks like this change is not needed for my feature, and it's used by other files as you pointed out. I reverted it

docs/guides/data_input_pipeline.md
```diff
@@ -37,11 +37,11 @@ The approaches to solve these challenges depend on whether your dataset supports
...
+* **Uneven completion**: Data indices are distributed evenly among hosts. Without packing, the data imbalance between hosts will be at most one batch. To handle the final steps where some hosts run out of data, you can enable the `generate_padding_batch_{train/eval}` flag. This directs hosts to generate empty "padding" batches until the training or evaluation steps are met. **Note**: When sequence packing is enabled, the difference in the number of packed examples per host can be larger. The `generate_padding_batch_{train/eval}` flag still solves this. However, as more hosts begin generating padding, you will observe a decrease in total_weights and a slower change in the training loss. If all hosts exhaust their data before the target step count is reached, both total_weights and loss will drop to 0.
```
RissyRan: Nit: shall we do something like `generate_padding_batch_train`/`generate_padding_batch_eval` to avoid confusing? Similar for other occurrences

src/MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -162,8 +161,8 @@ def preprocessing_pipeline(
     packing=True,
     shift=True,
     num_threads=1,
-    drop_remainder=False,
-    generate_padding_example=False,
+    drop_remainder=True,
```
RissyRan, line 164: Why this is True now? Curious.

aireenmei: Good question. The old hf specific approach allows padding partial batch, so no need to drop remainder. But this new approach only pads the full batch, so has to drop remainder, unless the pipeline has a custom function to pad partial batch to full. In this case, hf pipeline is similar to tfds, which always has drop_remainder=True (https://github.com/AI-Hypercomputer/maxtext/blob/f4ab533995e55eae01b2be26503f421eda008bea/src/MaxText/input_pipeline/_tfds_data_processing.py#L93). Grain already has a function to pad partial batch, so it allows drop_remainder=False. It's possible to add similar padding to hf and tfds if there's customer request. But at the moment, grain is the recommended pipeline, and we prioritize migrating customer to grain.

docs/guides/data_input_pipeline.md
aireenmei: revised

aireenmei:
> Thanks for the feature! LGTM overall. A few high level questions:
> 
> 1. could you help attach your tests in the PR description.
> 2. I didn't see new test is added for this feature, wondering how do we test/know it is working as expected? Probably just paste your test in the description?

Thanks for the suggestion! I updated in the description

src/MaxText/multihost_dataloading.py
```diff
@@ -103,21 +77,64 @@ def __init__(self, dataloader: tf.data.Dataset | Iterable, global_mesh: Mesh):
...
+      for _ in range(MAX_DATA_LOAD_ATTEMPTS):
+        try:
+          local_data = next(self.local_iterator)
+          break  # exit the loop on success
+        except tf.errors.FailedPreconditionError as e:
+          max_logging.log(f"Failed to get next data batch due to {e}, retrying")
+          time.sleep(SLEEP_TIME)
+        except Exception as e:
+          if isinstance(e, StopIteration) or "StopIteration" in str(e):
+            if self.generate_padding_batch:
```
RissyRan, line 119: what's the behavior here if `self.generate_padding_batch=False`?

RissyRan: Discussed offline.

requirements_with_jax_stable_stack_0_6_1_pipreqs.txt
```diff
@@ -8,7 +8,7 @@ datasets==3.6.0
 etils==1.12.2
 evaluate==0.4.4
 flax==0.11.0
-grain==0.2.10
+grain==0.2.12
```
parambole, line 11: @aireenmei - Maybe we can leave these as is as the versions were pinned with 0.6.1

aireenmei: Discussed offline, since my change uses features only available in this grain version, it will break if user use this image with maxtext head. cc @bvandermoon @cfbh-google for viz

parambole: Discussed offline. Looks like we are using this file with Maxtext@HEAD and @aireenmei mentioned that these changes are required for Maxtext at HEAD. 

Approving for now. @bvandermoon Please let us know if you have any concerns.

PR #2232, opened 2025-08-24
Add doc for logs and metrics

shuningjin:
# Description

Add MaxText doc: Understanding Logs & Metrics

Reference
- sheet: http://shortn/_WpJ0ksNSkV
- guide: http://shortn/_2CkORMDiPV
- For 3.1: https://docs.jax.dev/en/latest/notebooks/host-offloading.html#activation-offloading
- For 3.3 & 4.1: https://github.com/AI-Hypercomputer/maxtext/blob/e31039e4e34622f5afdea48df770d4a54f057f4e/docs/guides/performance_metrics.md

## Test

Increase batch size from 1 to 24

**Command4: main demo**
```
python3 -m MaxText.train MaxText/configs/base.yml \
base_output_directory=gs://runner-maxtext-logs run_name=demo \
model_name=deepseek2-16b \
per_device_batch_size=24 max_target_length=2048 steps=10 dataset_type=synthetic enable_checkpointing=false
```
https://paste.googleplex.com/6275408357752832

Command5: output path (section 1)
```
python3 -m MaxText.train MaxText/configs/base.yml \
base_output_directory=gs://runner-maxtext-logs run_name=demo-$(date +%Y-%m-%d-%H-%M-%S) \
model_name=deepseek2-16b \
per_device_batch_size=24 max_target_length=2048 steps=10 dataset_type=synthetic \
enable_tensorboard=True \
profiler=xplane skip_first_n_steps_for_profiler=5 profiler_steps=3 \
gcs_metrics=True \
save_config_to_gcs=True \
enable_checkpointing=True 
```
https://paste.googleplex.com/5453878690316288

Command6: real dataset to show total weight (section 4)
```
python3 -m MaxText.train MaxText/configs/base.yml \
base_output_directory=gs://runner-maxtext-logs run_name=demo \
model_name=deepseek2-16b \
per_device_batch_size=24 max_target_length=2048 steps=10 enable_checkpointing=false \
dataset_type=tfds dataset_path=gs://maxtext-dataset dataset_name='c4/en:3.0.1'
```
https://paste.googleplex.com/5032928475873280

## Tests (deprecated)

Command1: main demo
```
python3 -m MaxText.train MaxText/configs/base.yml \
base_output_directory=gs://runner-maxtext-logs run_name=demo \
model_name=deepseek2-16b \
per_device_batch_size=1 max_target_length=2048 steps=10 dataset_type=synthetic enable_checkpointing=false
```
Example log: https://paste.googleplex.com/5789309067853824

[...]

PR #2306, opened 2025-09-08
Review 'Understand Logs and Metrics' doc

melissawm:
# Description

Follow-up to #2232

This PR includes some suggestions for the recently merged document "Understand Logs and Metrics". 

A few notes:
- I suggest using the "none" lexer for the code blocks here, since they don't actually represent code but outputs in the log. This causes the output to not be highlighted (as opposed to having distracting colors which are not useful in this context).
- I have updated the links to the code to match the current structure with `src/MaxText`. I used permalinks, so hopefully this won't have to be done every single time code moves around the repo. However in this case I thought it was important since it's a completely new strucutre and the old links (while they work) might be confusing for readers.

# Tests

Docs built successfully locally.

# Checklist

### DIFF
--- a/docs/tutorials/sft.md
+++ b/docs/tutorials/sft.md
@@ -20,7 +20,7 @@ Supervised fine-tuning (SFT) is a process where a pre-trained large language mod
 This tutorial demonstrates step-by-step instructions for setting up the environment and then training the model on a Hugging Face dataset using SFT.
 
 We use [Tunix](https://github.com/google/tunix), a JAX-based library designed for post-training tasks, to perform SFT.
- 
+
 In this tutorial we use a single host TPU VM such as `v6e-8/v5p-8`. Let's get started!
 
 ## Install dependencies


docs/tutorials/sft.md
```diff
@@ -20,7 +20,7 @@ Supervised fine-tuning (SFT) is a process where a pre-trained large language mod
 This tutorial demonstrates step-by-step instructions for setting up the environment and then training the model on a Hugging Face dataset using SFT.
 
 We use [Tunix](https://github.com/google/tunix), a JAX-based library designed for post-training tasks, to perform SFT.
-
+
```
bvandermoon: Nit: I think there is an extra whitespace added here. It might fail the linter

bvandermoon (approved):

docs/tutorials/sft.md
melissawm: Fixed!

SurbhiJainUSC (approved):

PR #2310, opened 2025-09-08
Update pyproject.toml to adopt seed-env for dep management

kanglant:
The `pyproject.toml` file is now generated by [seed-env](https://github.com/google-ml-infra/actions/tree/main/python_seed_env), which provides a complete closure of all direct and transitive dependencies with their versions as lower bounds. This allows engineers and researchers to reproduce the exact same environment at any point by simply checking out a maxtext commit and running `uv pip install . --resolution=lowest`, or running `uv pip install 'maxtext==<version>' --resolution=lowest` after the package is released.

The full command used to generate this pyproject.toml is `seed-env --local-requirements=requirements.txt --seed-commit=5a4a3b9147811616f4d8ae2c69240d4685d6673f --python-version=3.12`. 

The pyproject.toml is generated using specific JAX's lock files as a base environment by default. This PR uses [a python 3.12 lock file from JAX 0.7.0](https://github.com/jax-ml/jax/blob/5a4a3b9147811616f4d8ae2c69240d4685d6673f/build/requirements_lock_3_12.txt) as seed. This approach ensures a stable and tested foundation, on top of which `seed-env` intelligently resolves MaxText's dependencies from `requirements.txt`.

The maxtext's `requirements.txt` file has also been cleaned up in this PR. I've removed unnecessary lower bounds and pinned all VCS dependencies with a specific commit hash to ensure reproducibility. We'll need to set up a workflow to regularly update the commit hashes for our VCS dependencies and run `seed-env` to update the `pyproject.toml` file, as these dependencies are under active development.

In the future, we should have a CI/CD workflow to run seed-env and automatically update the `pyproject.toml` file when needed.

I also updated `setup.sh` to use `uv pip install ...` for faster dependency resolution and installation.

# Tests

In addition to the presubmits, I also set up a TPU VM, ran setup.sh, and ran the MaxText.train and MaxText.decode commands for local testing.

# Checklist


docs/guides/knowledge_distillation.md
```diff
@@ -52,7 +52,7 @@ git clone https://github.com/AI-Hypercomputer/maxtext.git
 python3 -m venv ~/venv-maxtext
 source ~/venv-maxtext/bin/activate
 cd maxtext
-pip install -r requirements.txt
+uv pip install -r requirements.txt
```
bvandermoon, line 55: Customers will still be able to run without `uv`, right? It's just that you are saying `uv` should be recommended now?

pyproject.toml
```diff
@@ -8,34 +8,259 @@ path = "src/MaxText/__init__.py"
...
     "Programming Language :: ML",
     "Environment :: GPU"
 ]
+dependencies = [
+    "absl-py>=2.3.1",
+    "aiofiles>=24.1.0",
+    "aiohappyeyeballs>=2.6.1",
+    "aiohttp>=3.12.15",
+    "aiosignal>=1.4.0",
+    "annotated-types>=0.7.0",
```
bvandermoon, line 29: How do you decide the right JAX commit for the command you shared in the PR description? Is it possible to tie this to a JAX version instead?

```
seed-env --local-requirements=requirements.txt --seed-commit=5a4a3b9147811616f4d8ae2c69240d4685d6673f --python-version=3.12
```

requirements.txt
```diff
@@ -3,42 +3,42 @@ aqtp
...
+google-cloud-aiplatform
 google-cloud-monitoring
-google-jetstream@git+https://github.com/AI-Hypercomputer/JetStream.git
-grain[parquet]>=0.2.6
+grain[parquet]
 huggingface_hub
-jax>=0.4.30
-jaxlib>=0.4.30
+jax<0.7.1
+jaxlib<0.7.1
```
bvandermoon, line 14: Is this due to MaxText incompatibilities?

bvandermoon (approved):
LGTM, thanks @kanglant

bvandermoon (commented):
`uv pip install . --resolution=lowest` will work for the `benchmark_runner` too, right? Even when it is not moved under `src` and is not part of the PyPi package

docs/guides/knowledge_distillation.md
kanglant: Yes. It is more recommended to use `uv pip install . --resolution=lowest` to install the maxtext package and the deps with pinned versions. Let me update it in the later PR.

pyproject.toml
kanglant: Right now, JAX adds the release tag before its lock files are updated with the new JAX version, i.e., if you use a tag like `jax-v0.7.1`, the jaxlib version in the corresponding lock file will still be 0.7.0. This means we have to manually select a JAX commit that comes right after the lock file update. We'll be updating our release process so you can use a github tag or branch with seed-env in the future.

requirements.txt
kanglant: This is due to a mosaic error with jaxlib 0.7.1: https://github.com/jax-ml/jax/issues/31245. I think it's better to change `<` to `!=`. Let me update it in the chained PR.

bvandermoon (approved):
I was able to build a docker image and run a successfull `benchmark_runner` training on this branch:

Build docker image:
```
bash docker_build_dependency_image.sh DEVICE=tpu MODE=stable
``` 

Run workload:
```
python3 -m benchmarks.benchmark_runner xpk \
    --project=$PROJECT \
    --zone=$ZONE \
    --device_type=v6e-256 \
    --num_slices=1  \
    --cluster_name=${CLUSTER_NAME} \
    --base_output_directory=${OUTPUT_DIR} \
    --model_name="llama3_1_70b_8192" \
    --base_docker_image=maxtext_base_image
```

PR #2315, opened 2025-09-08
Fix random keys after NNX migration

RissyRan:
# Description

Thank you @cgarciae for the fix!

Fix random keys issue: b/442608470. This `use_random_routing`  feature is mainly used for debug and testing.
* Have to pass the keys outside of shard_map to work around
* Update both **sparse_matmul and dense_matmul** for alignment

# Tests

* Test locally
  * sparse_matmul: https://paste.googleplex.com/5365694891950080
  * dense_matmul: https://paste.googleplex.com/6508950325428224

# Checklist

### DIFF
--- a/src/MaxText/layers/moe.py
+++ b/src/MaxText/layers/moe.py
@@ -295,6 +295,7 @@ class RoutedMoE(nnx.Module):
     self.weight_dtype = weight_dtype
     self.dtype = dtype
     self.quant = quant
+    self.rngs = rngs
 
     self.wi_kernel_axes = ("exp", "embed_no_exp", "mlp")
     self.wo_kernel_axes = ("exp", "mlp", "embed_no_exp")
@@ -311,7 +312,7 @@ class RoutedMoE(nnx.Module):
         use_bias=self.config.routed_bias,
         score_func=self.config.routed_score_func,
         matmul_precision=self.config.matmul_precision,
-        rngs=rngs,
+        rngs=self.rngs,
     )
 
     # pylint: disable=protected-access
@@ -332,7 +333,7 @@ class RoutedMoE(nnx.Module):
     else:
       self.wi_0 = nnx.Param(
           self.kernel_init(
-              rngs.params(),
+              self.rngs.params(),
               (num_experts, self.config.emb_dim, intermediate_dim),
               weight_dtype,
               kernel_in_axis,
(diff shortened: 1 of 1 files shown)


src/MaxText/layers/moe.py
```diff
@@ -1064,7 +1069,7 @@ def wrapper(x, logits, pre_bias_logits, w0, w1, wo, w0_bias, w1_bias, wo_bias):
 
       return output, None
 
-    return wrapper(inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kernel, wo_kernel, w0_bias, w1_bias, wo_bias)
+    return wrapper(inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kernel, wo_kernel, w0_bias, w1_bias, wo_bias, self.rngs.fork())
```
bvandermoon: We have been using [hasattr snippets](https://github.com/AI-Hypercomputer/maxtext/blob/1220cb6b1ee35605afb3bbd517124704549068b9/src/MaxText/layers/linears.py#L309) to continue supporting older versions of Flax that don't have `fork`

RissyRan: Thanks Branden for the pointer! I actually re-test it without the fork, and it also works. I think the trick here is to pass the key outside of shard_map, instead of call it inside.

I have tested locally for both sparse and dense matmuls: https://paste.googleplex.com/5502532315119616

src/MaxText/layers/moe.py
```diff
@@ -390,12 +391,15 @@ def get_tensor_transpose_parallelism_size(self):
...
+  def get_topk(self, gate_logits, pre_bias_logits, rngs=None):
     """get topk."""
     # shape of top_k_weights & top_k_indices:
     # (batch, sequence, num_experts_per_tok).
     if self.config.use_random_routing:
-      rng = self.make_rng("random_routing")
+      if rngs is None:
+        raise ValueError("The random key cannot be None for random routing.")
+      # Re-use the 'dropout' RNG stream to ensure random routing
+      rng = rngs.dropout()
```
bvandermoon, line 407: Why isn't this just `self.rng`?

RissyRan: Discussed offline. Using self.rng inside of shard_map will cause context trace error. Workaround is to pass this key outside of shrad_map, due to the current bridge implementation.

PR #2317, opened 2025-09-09
Create a CI workflow that build maxtext package and run tests against it

kanglant:
Create MaxText Package Tests workflow, which builds maxtext package and run cpu/tpu unit/integration tests. 

Save the built maxtext distributions as action artifacts. Artifacts are retained within the workflow run for 90 days.

Create a docker image, which has all the system level deps for maxtext and no python libraries pre-installed except `pip` and `uv`, for the package testing in a clean python environment.

Test with local configs/assets/test_assets files for now.

# Tests

The MaxText Package Tests presubmit passed. 

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 path = "src/MaxText/__init__.py"
 
 [project]
-name = "MaxText"
+name = "maxtext"
 description = "MaxText is a simple, performant and scalable Jax LLM!"
 dynamic = ["version"]
 requires-python = ">=3.12"
--- a/.github/workflows/build_and_test_maxtext.yml
+++ b/.github/workflows/build_and_test_maxtext.yml
@@ -1,0 +1,113 @@
+# Copyright 2025 Google LLC
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 path = "src/MaxText/__init__.py"
 
 [project]
-name = "MaxText"
+name = "maxtext"
```
SamuelMarks, line 9: We're not doing the rename yet; soon; but not yet

Rohan-Bierneni:
Do we want to run these tests as part of every unit test even though they are duplicate of the old ones, the difference is setup is using maxtext package? I think we could have them run only on scheduled runs or a nightly run to make sure they work properly so we can save on resources and also on total unit test time? @kanglant @bvandermoon

bvandermoon:
> Do we want to run these tests as part of every unit test even though they are duplicate of the old ones, the difference is setup is using maxtext package? I think we could have them run only on scheduled runs or a nightly run to make sure they work properly so we can save on resources and also on total unit test time? @kanglant @bvandermoon

@Rohan-Bierneni we are planning to make these the main MaxText CI tests. @kanglant is currently keeping the old ones in place until we build confidence in the new ones. There should not be both running for long. Also as FYI, she is not updating the GPU tests for now

parambole (approved):
Awesome! Thanks for adding this.

PR #2325, opened 2025-09-09
Add rule to load quantization ckpt

RissyRan:
# Description

This is only affects loading AQT quantized ckpt for MoE models, and we suggest users to use [Qwix](https://github.com/google/qwix) instead. 

Context:
* We met issues to load AQT quantization ckpt for decoding after NNX migration, b/442608470
* The fix is doable (like attention does with some [dummy values & workaround](https://github.com/AI-Hypercomputer/maxtext/blob/01c7137d4e13878e38baae44dc99e588eaa50a70/src/MaxText/layers/attention_op.py#L477-L525)).
* Confirmed with team that Qwix is working end-to-end, so we'd like to add an assertion to the pyconfig to indicate this is an known issue, and we will do the clean up of AQT once convergence of Qwix is fully verified.

# Tests

Test locally:
* Functional end-to-end without `load_parameters_path` for both AQT & Qwix
* Assertion when using AQT + `load_parameters_path`: [link](https://paste.googleplex.com/5113077431402496)

# Checklist

### DIFF
--- a/src/MaxText/pyconfig.py
+++ b/src/MaxText/pyconfig.py
@@ -215,6 +215,7 @@ def validate_keys(keys):
 
   validate_multiple_slices(keys)
   if keys["num_experts"] > 1:
+    validate_quantization_checkpoint(keys)
     validate_mlp_dim(keys)
     validate_sparse_matmul_parallelism(keys)
     validate_ragged_dot(keys)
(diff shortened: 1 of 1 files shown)


RissyRan:
Discussed with team, we still need AQT to load quantized ckpt at this moment.

PR #2335, opened 2025-09-11
Tunix import fix for GRPO & GRPO fix in DEBUG mode

mydatascience:
# Description

Fixing Tunix imports (pointing to the latest main) and grpo py script in DEBUG mode.
Currently GRPO is broken.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,7 +236,7 @@ dependencies = [
     "tqdm>=4.67.1",
     "transformers>=4.56.1",
     "treescope>=0.1.10",
-    "tunix @ https://github.com/google/tunix/archive/4c5561be36d8a2f1f0858c2685554ca4e1a65fd2.zip",
+    "tunix @ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip",
     "typing-extensions>=4.14.1",
     "typing-inspection>=0.4.1",
     "tzdata>=2025.2",
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,7 @@ tensorflow-text
 tensorflow
 tiktoken
 transformers
-tunix @ https://github.com/google/tunix/archive/4c5561be36d8a2f1f0858c2685554ca4e1a65fd2.zip
+tunix @ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip
 google-jetstream @ https://github.com/AI-Hypercomputer/JetStream/archive/daedc21c393f23449fb54ddc4f75fca34348ea9c.zip
 mlperf-logging @ https://github.com/mlcommons/logging/archive/38ab22670527888c8eb7825a4ece176fcc36a95d.zip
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -36,7 +36,7 @@ tensorflow-text
 tensorflow
 tiktoken
 transformers
-tunix@ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip
+tunix @ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip
```
SurbhiJainUSC, line 39: I don't think `pip install` would work when spacing is present before and after `@`.

mydatascience: It does for files. Other lines here look this way. For checkouts - it won't

SurbhiJainUSC: I get this:
 pip install tunix @ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip

```
ERROR: Invalid requirement: '@': Expected package name at the start of dependency specifier
    @
    ^
```

pyproject.toml
```diff
@@ -236,7 +236,7 @@ dependencies = [
     "tqdm>=4.67.1",
     "transformers>=4.56.1",
     "treescope>=0.1.10",
-    "tunix @ https://github.com/google/tunix/archive/4c5561be36d8a2f1f0858c2685554ca4e1a65fd2.zip",
+    "tunix @ https://github.com/google/tunix/archive/d770659621eb16ef6588268e26fa687fa068df20.zip",
```
SurbhiJainUSC, line 239: Can you pin the commit `81ad1e6de0528d399eb2fbddb6ea3f92c7cb3d4f`, as Tunix at head is currently broken for SFT?

cc: @bvandermoon

requirements.txt
SurbhiJainUSC: @kanglant - Were you able to install the packages when spacing is present?

pyproject.toml
```diff
@@ -236,7 +236,7 @@ dependencies = [
     "tqdm>=4.67.1",
     "transformers>=4.56.1",
     "treescope>=0.1.10",
-    "tunix @ https://github.com/google/tunix/archive/4c5561be36d8a2f1f0858c2685554ca4e1a65fd2.zip",
+    "tunix@git+https://github.com/google/tunix.git",
```
bvandermoon: Discussed offline, we are trying to get every dependency pinned to a specific version for reproducibility in PyPi and benchmarks. But there is currently no Tunix pin that works for both SFT and GRPO. We will need to go with this for now

requirements.txt
mydatascience: we decided to go back to the head of tunix, no spaces

bvandermoon: Why does Tunix need the @, but other packages don't?

mydatascience: we don't have tunix in pip that's why we need to add @ something

PR #2336, opened 2025-09-11
Fix train step counter in on_train_step_end() hook for SFT

SurbhiJainUSC:
# Description

[This commit](https://github.com/google/tunix/commit/ef0b4e582e048d4316a24fc5e16bee16c402c9ce) in Tunix changed the training loop, causing the `train_step` to be incremented before logging metrics. As a result, Tunix logs the metrics at x+1 for step=x. This created an issue with MaxText, which expects the `on_train_step_end()` hook to be called with train_step=x, when logging metrics for step x. As a result, the SFTTrainingHooks.on_train_step_end() hook failed with an AssertionError. 
```
AssertionError: SFTTrainingHooks.on_train_step_start() must be called before SFTTrainingHooks.on_train_step_end()
```

This PR resolves the problem by adjusting the on_train_step_end() hook to account for this change in the step increment timing.

# Tests

`bash end_to_end/tpu/llama3.1/8b/run_sft.sh`

# Checklist

### DIFF
--- a/src/MaxText/sft/hooks.py
+++ b/src/MaxText/sft/hooks.py
@@ -103,26 +103,29 @@ class SFTTrainingHooks(TrainingHooks):
       train_loss: float,
       step_time: float,
   ):
-    """Called at the end of training step."""
-    assert train_step in self.train_metadata, (
-        "SFTTrainingHooks.on_train_step_start() must be called before"
-        " SFTTrainingHooks.on_train_step_end()"
+    """Called at the end of training step.
+    This hook is called by Tunix after the step counter has been incremented for logging purposes.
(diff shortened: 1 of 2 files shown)


src/MaxText/sft/hooks.py
```diff
@@ -92,7 +92,7 @@ def on_train_step_start(self, train_ctx: peft_trainer.PeftTrainer):
     total_weights = jnp.sum(train_ctx.data_hooks.train_batch["targets_segmentation"] != 0)
 
     self.train_metadata[train_ctx.train_steps] = {
-      "total_weights": total_weights,
+        "total_weights": total_weights,
```
A9isha: nit: whitespace added

PR #2350, opened 2025-09-15
Support multiple images per example in Gemma3

aireenmei:
# Description

The Gemma3 model was pre-trained with a mix of text + image of long seq_len (128k for 4B model), and there is no limit in the number of images per example. Our initial support only allows one image per example. This PR expands the support.
* User can provide multiple images to run decode.py
* When using sft, the "image" column in the dataset can be a single image, or a list of images of different lengths

Feature request: b/435464855

# Tests
* [decode script](https://paste.googleplex.com/5911819562647552); [decode output](https://paste.googleplex.com/6257301044461568)
* For SFT, the testing dataset ChartQA only contains a single image. For testing purpose, I added a transformation function to make the "image" column a list of images of varied length:
```
def ensure_images_is_list(example):
   if not isinstance(example["images"], list):
      example["images"] = [example["images"]] * np.random.randint(0, 3)
   print(f"{len(example['images'])=}")
   return example
dataset = dataset.map(ensure_images_is_list)
```
* [sft script](https://paste.googleplex.com/5742407480115200); [sft output](https://paste.googleplex.com/6030701925367808)
* the original dataset with single image per example is also tested to confirm it's still supported.

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -798,9 +798,13 @@ freeze_vision_encoder_params: True
 dtype_mm: "float32"  # Data type for multimodal model's vision encoder
 remat_policy_for_vit: "minimal"  # Remat policy for multimodal model's vision encoder. Check `remat_policy` for options.
 image_size_for_vit: 896 # Default for Gemma3, and should be overwritten by model's config
-image_path: "" # Local image path used for decoding
+image_path: "" # Local image path used for decoding, can be multiple paths separated by comma, exp "/path/image1.jpg,/path/image2.jpg"
 image_placeholder: "<|image|>"
 posemb_type_for_vit: "learn"
+# max_num_images_per_example only applies for training when your image column is a list of images.
+# -1 means no limit, and will pad to the max possible number of images determined by sequence length.
+# Set it to avoid unnecessary padding if you know the maximum number of images per example.
+max_num_images_per_example: -1
 
 ### llama4 multi modal configs
 hidden_size_for_vit: 1408
--- a/src/MaxText/decode.py
+++ b/src/MaxText/decode.py
@@ -16,7 +16,7 @@
 
 import os
 from typing import Sequence
-
+import numpy as np
 import jax
 import jax.numpy as jnp
 
@@ -101,13 +101,14 @@ def main(argv: Sequence[str]) -> None:
   prefill_length = config.max_prefill_predict_length
   processor_output = multimodal_utils.PreprocessorOutput()
   if config.use_multimodal:
-    text = multimodal_utils.reformat_prompt(
-        text, image_placeholder=config.image_placeholder, model_name=config.model_name
+    image_path = config.image_path.split(",")
+    images = [multimodal_utils.load_image_from_path(p) for p in image_path]
+    processor_outputs = [multimodal_utils.pre_process_image(img, model_name=config.model_name) for img in images]
+    image_offsets = sum(
+        [multimodal_utils.get_image_offsets(config.model_name, processor_output=po) for po in processor_outputs]
     )
-    # TODO(hengtaoguo): Support multiple images as input.
-    images = multimodal_utils.load_image_from_path(config.image_path)
-    processor_output = multimodal_utils.pre_process_image(images, model_name=config.model_name)
-    prefill_length -= multimodal_utils.get_image_offsets(config.model_name, processor_output=processor_output)
+    prefill_length -= image_offsets
+    text = multimodal_utils.reformat_prompt(text, image_placeholder=config.image_placeholder, model_name=config.model_name, num_images=len(images))
 
   metadata = engine.get_tokenizer()
   tokenizer_model = engine.build_tokenizer(metadata)
@@ -119,9 +120,9 @@ def main(argv: Sequence[str]) -> None:
   tokens, true_length = tokenizer_model.encode(text, is_bos=not has_chat_template, prefill_lengths=[prefill_length])
   if config.use_multimodal:
     tokens = multimodal_utils.prepare_text_for_image_fusion(
-        tokens, model_name=config.model_name, processor_output=processor_output
+        tokens, model_name=config.model_name, processor_output=processor_outputs
     )
-    true_length += multimodal_utils.get_image_offsets(config.model_name, processor_output=processor_output)
+    true_length += image_offsets
 
   assert (
       true_length <= config.max_prefill_predict_length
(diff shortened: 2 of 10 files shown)


src/MaxText/decode.py
```diff
@@ -119,9 +120,9 @@ def main(argv: Sequence[str]) -> None:
   tokens, true_length = tokenizer_model.encode(text, is_bos=not has_chat_template, prefill_lengths=[prefill_length])
   if config.use_multimodal:
     tokens = multimodal_utils.prepare_text_for_image_fusion(
-        tokens, model_name=config.model_name, processor_output=processor_output
+        tokens, model_name=config.model_name, processor_output=processor_outputs
```
hengtaoguo, line 123: Now it will expand each of the <start_of_image> to placeholder tokens?

src/MaxText/decode.py
```diff
@@ -101,13 +101,14 @@ def main(argv: Sequence[str]) -> None:
...
+    processor_outputs = [multimodal_utils.pre_process_image(img, model_name=config.model_name) for img in images]
+    image_offsets = sum(
+        [multimodal_utils.get_image_offsets(config.model_name, processor_output=po) for po in processor_outputs]
     )
-    # TODO(hengtaoguo): Support multiple images as input.
-    images = multimodal_utils.load_image_from_path(config.image_path)
-    processor_output = multimodal_utils.pre_process_image(images, model_name=config.model_name)
-    prefill_length -= multimodal_utils.get_image_offsets(config.model_name, processor_output=processor_output)
+    prefill_length -= image_offsets
+    text = multimodal_utils.reformat_prompt(text, image_placeholder=config.image_placeholder, model_name=config.model_name, num_images=len(images))
```
hengtaoguo, line 111: Could you remind me, if <start_of_image> is not provided, is it directly appended to the front?

src/MaxText/configs/base.yml
```diff
@@ -798,9 +798,13 @@ freeze_vision_encoder_params: True
 dtype_mm: "float32"  # Data type for multimodal model's vision encoder
 remat_policy_for_vit: "minimal"  # Remat policy for multimodal model's vision encoder. Check `remat_policy` for options.
 image_size_for_vit: 896 # Default for Gemma3, and should be overwritten by model's config
-image_path: "" # Local image path used for decoding
+image_path: "" # Local image path used for decoding, can be multiple paths separated by comma
```
hengtaoguo: Thanks! Maybe add an example in the comment like "/path/image1.jpg,/path/image2.jpg"?

src/MaxText/configs/base.yml
```diff
@@ -798,9 +798,13 @@ freeze_vision_encoder_params: True
...
 image_placeholder: "<|image|>"
 posemb_type_for_vit: "learn"
+# max_num_images_per_example only applies when your image column is a list of images.
+# -1 means no limit, and will pad to the max possible number of images determined by sequence length.
+# Set it to avoid unnecessary padding if you know the maximum number of images per example.
+max_num_images_per_example: -1
```
hengtaoguo, line 807: What if the number of image tokens is larger than the max_prefill_predict_length? I was wondering if we have a check on that anywhere...

src/MaxText/input_pipeline/_input_pipeline_utils.py
```diff
@@ -400,58 +415,58 @@ def map(self, element):
...
       if data_column != "images":
         element[f"{data_column}_segmentation"] = (element[data_column] != self.pad_id).astype(np.int32)
         element[f"{data_column}_position"] = np.arange(element[data_column].shape[0], dtype=np.int32)
+        if self.add_true_length:
+          element[f"{data_column}_true_length"] = np.array([element[data_column].shape[0]], dtype=np.int32)
     for key, _ in element.items():
-      if key != "images":
-        element[key] = _pad(element[key], self.max_length, self.pad_id)
+      if key == "images":
+        if isinstance(element["images"], list):
```
hengtaoguo, line 463: I agree this is a reasonable starting point where one "images" column could contain a list of images. It is easier for the users to compose their own dataset following this format.

Trying to find a real multi-image VQA dataset and found this [SlideVQA](https://huggingface.co/datasets/NTT-hil-insight/SlideVQA), where it actually contains 20 columns page_1~20, and each column contains an image. We could also revisit this kind of format in future support if necessary?

src/MaxText/decode.py
aireenmei: yes, that's correct.

src/MaxText/decode.py
aireenmei: yes, that's correct.

src/MaxText/configs/base.yml
aireenmei: This only applies to training (will add to the comment for clarity), where we don't know the number of images in future examples, but it requires the batches to be the same shape. For decode there's no limit, if the max_prefill_predict_length is too short user will get this error: https://github.com/AI-Hypercomputer/maxtext/blob/87131c16a79e3b8e4880ce1c5a3ab02635acdb26/src/MaxText/decode.py#L126-L128

src/MaxText/input_pipeline/_input_pipeline_utils.py
aireenmei: Thanks for finding the dataset! For dataset like this, it should be straightfoward to merge the columns into a list, and downstream will work similarly. I agree we can revisit if this is a more popular format or we hear customer requests.

src/MaxText/configs/base.yml
aireenmei: done

aireenmei:
> Thanks for the great work!!
> 
> One general question: if a gbs=4, and each example has different number of images [1, 2, 1, 2], what does the input to VisionEncoder look like? Previously when we only support 1 image per example, the input is [4, 896, 896, 3]. I believe now it's stacked to [6, 896, 896, 3]? I'm wondering how to divide the first dimension back and assign them to their own examples.

The Gemma3 image encoder input is [b,n,h,w,c] (https://github.com/AI-Hypercomputer/maxtext/blob/87131c16a79e3b8e4880ce1c5a3ab02635acdb26/src/MaxText/layers/gemma3.py#L594). So for 1 image it's [4, 1, 896, 896, 3]. For the batch in your example it will be [4, max_num_images_per_example, 896, 896, 3]. User can set max_num_images_per_example or it will be calculated to be (max_target_len // 256) -1 (GEMMA_NUM_PLACEHOLDER_TOKENS_PER_IMAGE=256).

SurbhiJainUSC (approved):

Issue #2344, opened 2025-09-15
[feature request] Support chunking (splitting) long sequences instead of truncation during tokenization


bzantium:
### Feature or Model Request

* **What problem are you trying to solve?**
    The current data pipeline (`TokenizeAndTrim`) simply truncates any text sequence that exceeds `max_target_length`. This causes all data past this limit in long documents (e.g., from large corpora like C4) to be completely lost during training. This is a highly inefficient use of data.

* **Why is this problem important?**
    Data loss leads directly to wasted compute resources and storage. Furthermore, it risks biasing the model by only training on the *beginning* of every document, which reduces the model's opportunity to learn from long-context information. It is far more efficient to utilize 100% of the purchased or preprocessed data.

* **Describe your requested feature or solution.**
    1.  Implement a new class, `TokenizeAndChunk`, that inherits from `grain.experimental.FlatMapTransform` (which supports 1:N mappings) as an alternative to the existing `TokenizeAndTrim` (`grain.MapTransform`).
    2.  Instead of truncating the token list (i.e., `[:seq_len]`), this new transform should split the *entire* token sequence into multiple chunks, each of `max_target_length`, and return them as a list (`[chunk_1, chunk_2, ...]`).
    3.  Add a boolean configuration flag, **`use_truncation`**, to control this behavior:
        * When **`use_truncation = True`**, the pipeline will use the existing `TokenizeAndTrim` (`MapTransform`) to truncate sequences.
        * When **`use_truncation = False`**, the pipeline will use the new `TokenizeAndChunk` (`FlatMapTransform`) to split sequences into multiple examples.

* **Describe alternatives you’ve considered (if any).**
    The only alternative is the current "truncation" behavior, which, as noted, is data-inefficient.

* **Additional context or examples.**
    For example, if `max_target_length = 512` and an input document tokenizes to 1200 tokens:
    * **Current (`use_truncation = True`):** Produces **one** example of 512 tokens. (688 tokens are lost).
    * **Proposed (`use_truncation = False`):** Produces **three** examples: one of 512 tokens, one of 512 tokens, and one of 176 tokens. (0 tokens are lost).

    This implementation requires using `dataset.apply(TokenizeAndChunk(...))` in the pipeline (when the flag is false) instead of `dataset.map(TokenizeAndTrim(...))`.

### Additional Context

* AI GDE and Kakao

PR #947, opened 2024-10-05
DPO

rdyro:
Adding DPO training to the main `train.py` script via config options.

The main new configs are `use_dpo` and other `dpo_...` values. Also `train_data_columns` and 
`eval_data_columns` are now lists of string field `['text']` for LLM pre-training and `['chosen', 'rejected']` for DPO.

Currently, only pre-processed datasets are supported already containing the concatenated full prompts: 'chosen' and 'rejected', like in [https://huggingface.co/datasets/Anthropic/hh-rlhf](https://huggingface.co/datasets/Anthropic/hh-rlhf)

Grain, HF and TFDS data pipelines are all compatible.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #2354, opened 2025-09-16
feat(input_pipeline): Add support for chunking long sequences instead truncation

bzantium:
This PR introduces "chunking" as an alternative to "truncation" in the Grain input pipeline.

Previously, the `TokenizeAndTrim` operation (`MapTransform`) would truncate any document longer than `max_target_length`, discarding all subsequent tokens. This change introduces a new `TokenizeAndChunk` operation (`FlatMapTransform`) that splits a single long document into *multiple* training examples, each no longer than `max_target_length`.

This new behavior is controlled by a new configuration flag, `use_truncation`.

  - **Why is this change being made?**
    The default truncation behavior is highly data-inefficient for corpora with many long documents (like C4). It wastes significant amounts of data, compute, and storage, and may bias the model by only ever training on the beginning of documents.

  - **The problem being solved and any relevant context:**
    This PR solves the problem of data loss during tokenization for long sequences. By using a 1:N `FlatMapTransform`, we can map one long input document to a list of multiple, valid training chunks, ensuring 100% of the tokenized data is used.

  - **Why this is a good solution:**
    This solution is efficient and flexible. It utilizes the `FlatMapTransform` provided by Grain, which is designed for this 1:N mapping. It is also fully backwards-compatible, as the new chunking behavior is "opt-in" by setting `use_truncation = False` in the config. The default behavior remains truncation.

  - **Some information about the specific implementation:**

    1.  **`_grain_tokenizer.py`**: A new `TokenizeAndChunk` class has been added. It inherits from `grain.experimental.FlatMapTransform` and implements the `flat_map` method to split a list of token IDs into multiple chunks.
    2.  **`_grain_data_processing.py`**: The `pretrain_preprocessing_pipeline` function has been updated with a conditional check for `config.use_truncation`:
          - If `True`, it uses the existing `dataset.map(TokenizeAndTrim(...))`.
          - If `False`, it uses `dataset.apply(TokenizeAndChunk(...))`.
    3.  **Requirement:** The `dataset.apply()` method and support for `FlatMapTransform` are recent features in Grain. This PR requires a version of Grain installed directly from the main branch.
        ```bash
        pip install git+https://github.com/google/grain.git
        ```

  - **Shortcomings of the solution and possible future improvements.**
    The `max_fan_out` attribute in `TokenizeAndChunk` is set with a class-level default (`2048`). If a document is exceptionally long and produces more chunks than this, it will error. This could be exposed as a configuration option in the future if needed.

### Tests

This change is tested with a new, self-contained unit test file: `tests/tokenizer_transform_test.py`.

  - This test **does not** require real data (like C4) or JAX/TPU.
  - It uses a `MockTokenizer` to provide known, deterministic tokenization (`"a b c" -> [1, 2, 3]`).
  - It uses an in-memory `grain.MapDataset.source` with a small, known dataset to test edge cases (short text, long text, and multi-chunk text).
  - Four separate test cases were added to verify the logic:
    1.  **`test_tokenize_and_trim`**: Verifies the original 1:1 truncation logic is correct.
    2.  **`test_tokenize_and_chunk`**: Verifies the new 1:N chunking logic (e.g., an input with 7 tokens and `max_len=5` correctly produces two new examples with 5 and 2 tokens).
    3.  **`test_trim_and_pad_chaining`**: Verifies that the output of `TokenizeAndTrim` can be correctly chained into a subsequent `PadToMaxLength` transform.
    4.  **`test_chunk_and_pad_chaining`**: Verifies that *all* outputs from `TokenizeAndChunk` are correctly chained into `PadToMaxLength` (e.g., both the 5-token chunk and the 2-token chunk are correctly padded).

To reproduce, you can run the new test file directly:

```bash
python -m unittest tests/tokenizer_transform_test.py
```

Fixes: #2344 

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -495,6 +495,11 @@ tokenize_train_data: True  # False if the dataset is pre-tokenized
 tokenize_eval_data: True  # False if the dataset is pre-tokenized
 add_bos: True
 add_eos: True
+# If False, use chunking for long sequences instead of truncation.
+# Note: use_truncation=False is only available in grain's pretrain preprocessing pipeline.
+# See the TokenizeAndTrim and TokenizeAndChunk classes in
+# `src/MaxText/input_pipeline/_grain_tokenizer.py` for implementation details.
+use_truncation: True
 
 # Dataset
 per_device_batch_size: 12.0
--- a/src/MaxText/input_pipeline/_grain_data_processing.py
+++ b/src/MaxText/input_pipeline/_grain_data_processing.py
@@ -95,9 +95,7 @@ def pretrain_preprocessing_pipeline(dataset, config, data_columns, tokenize, gra
     dataset = dataset.map(_input_pipeline_utils.NormalizeFeatures(data_columns, tokenize))
 
   assert len(data_columns) == 1
-  rekey_dict = {"inputs": "text", "targets": "text"}
-  dataset = dataset.map(_input_pipeline_utils.Rekey(rekey_dict))
-  data_columns = ("inputs", "targets")
+  text_column = data_columns[0]
 
   tokenizer_model = tokenizer.build_tokenizer(
       config.tokenizer_path,
@@ -115,11 +113,15 @@ def pretrain_preprocessing_pipeline(dataset, config, data_columns, tokenize, gra
     pad_id = -1
 
   if tokenize:
-    dataset = dataset.map(
-        _grain_tokenizer.TokenizeAndTrim(
-            data_columns, config.max_target_length, config.add_bos, config.add_eos, tokenizer_model
-        )
-    )
+    if config.use_truncation:
+      dataset = dataset.map(_grain_tokenizer.TokenizeAndTrim(text_column, config.max_target_length, tokenizer_model))
+    else:
+      dataset = dataset.apply(_grain_tokenizer.TokenizeAndChunk(text_column, config.max_target_length, tokenizer_model))
+
+  data_columns = ("inputs", "targets")
+  rekey_dict = {col: text_column for col in data_columns}
+  dataset = dataset.map(_input_pipeline_utils.Rekey(rekey_dict))
+
   # Pack and Batch examples.
   batch_size = config.global_batch_size_to_load // jax.process_count()
   if config.expansion_factor_real_data > 1:
@@ -176,11 +178,7 @@ def dpo_preprocessing_pipeline(dataset, config, data_columns, tokenize, grain_wo
     pad_id = -1
 
   if tokenize:
-    dataset = dataset.map(
-        _grain_tokenizer.TokenizeAndTrim(
-            data_columns, config.max_target_length, config.add_bos, config.add_eos, tokenizer_model
-        )
-    )
+    dataset = dataset.map(_grain_tokenizer.TokenizeAndTrim(data_columns, config.max_target_length, tokenizer_model))
 
   dataset = dataset.map(_input_pipeline_utils.PadOrTrimToMaxLength(config.max_target_length, pad_id))
   batch_size = config.global_batch_size_to_load // jax.process_count()
--- a/src/MaxText/input_pipeline/_grain_tokenizer.py
+++ b/src/MaxText/input_pipeline/_grain_tokenizer.py
@@ -24,35 +24,34 @@ from MaxText import tokenizer
 
 
 @dataclasses.dataclass
-class TokenizeAndTrim(grain.MapTransform):
-  """Tokenize and trim features to sequence length."""
+class TokenizerTransformBase:
+  """Base class for tokenizer transforms with common functionality."""
 
   # pylint: disable=attribute-defined-outside-init
   feature_names: str | Sequence[str]
   sequence_length: int | Sequence[int]
-  add_bos: bool
-  add_eos: bool
   tokenizer: tokenizer.SentencePieceTokenizerGrain | tokenizer.HFTokenizer
 
   def __post_init__(self):
     self._processor = None
     self._initialize_processor_lock = threading.Lock()
+    # Convert single values to lists for consistent processing
     if isinstance(self.feature_names, str):
       self.feature_names = [self.feature_names]
     if isinstance(self.sequence_length, int):
       self.sequence_length = [self.sequence_length] * len(self.feature_names)
 
-  def map(self, element: dict[str, Any]) -> dict[str, Any]:
-    """Maps to each element."""
+  def _get_processor(self):
     if self._processor is None:
       with self._initialize_processor_lock:
-        if self._processor is None:  # Ensures only one thread initializes SPP.
+        if self._processor is None:  # Ensures only one thread initializes processor.
           self._processor = self.tokenizer
-    for feature_name, sequence_length in zip(self.feature_names, self.sequence_length, strict=True):
-      text = element[feature_name]
-      token_ids = self._processor.encode(text)[:sequence_length]
-      element[feature_name] = np.asarray(token_ids, dtype=np.int32)
-    return element
+    return self._processor
+
+  def _encode(self, text: str) -> list[int]:
+    """Common method to encode text using the tokenizer."""
+    processor = self._get_processor()
+    return processor.encode(text)
 
   def __getstate__(self):
     state = self.__dict__.copy()
@@ -64,3 +63,49 @@ class TokenizeAndTrim(grain.MapTransform):
     self.__dict__.update(state)
     self._processor = None
     self._initialize_processor_lock = threading.Lock()
+
+
+@dataclasses.dataclass
+class TokenizeAndTrim(TokenizerTransformBase, grain.MapTransform):
+  """Tokenize and trim features to sequence length."""
+
+  def map(self, element: dict[str, Any]) -> dict[str, Any]:
+    """Maps to each element."""
+    for feature_name, max_length in zip(self.feature_names, self.sequence_length, strict=True):
+      text = element[feature_name]
+      token_ids = self._encode(text)[:max_length]
+      element[feature_name] = np.asarray(token_ids, dtype=np.int32)
+    return element
+
+
+@dataclasses.dataclass
+class TokenizeAndChunk(TokenizerTransformBase, grain.experimental.FlatMapTransform):
+  """Tokenize and chunk features into multiple examples of sequence length."""
+
+  max_fan_out: int = 2048
+
+  def __post_init__(self):
+    super().__post_init__()
+    # TokenizeAndChunk only supports single feature for chunking
+    assert len(self.feature_names) == 1, "TokenizeAndChunk only supports single feature name"
+    assert len(self.sequence_length) == 1, "TokenizeAndChunk only supports single sequence length"
+    self.feature_name = self.feature_names[0]  # For backward compatibility
+    self.sequence_length = self.sequence_length[0]  # Convert back to int for chunking
+
+  def flat_map(self, element: dict[str, Any]) -> list[dict[str, Any]]:
+    """Tokenize and chunk text into multiple examples of sequence length."""
+    text = element[self.feature_name]
+    chunk_size = self.sequence_length
+
+    token_ids = self._encode(text)
+
+    if not token_ids:
+      return []
+
+    output_elements = []
+    for start_idx in range(0, len(token_ids), chunk_size):
+      chunk = np.asarray(token_ids[start_idx : start_idx + chunk_size], dtype=np.int32)
+      new_element = {self.feature_name: chunk}
+      output_elements.append(new_element)
+
+    return output_elements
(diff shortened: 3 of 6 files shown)


src/MaxText/configs/base.yml
```diff
@@ -458,6 +458,7 @@ tokenize_train_data: True  # False if the dataset is pre-tokenized
 tokenize_eval_data: True  # False if the dataset is pre-tokenized
 add_bos: True
 add_eos: True
+use_truncation: True # If False, use chunking for long sequences instead of truncation
```
aireenmei: Could you point to the class for users who are interested in the detailed implementation.

src/MaxText/input_pipeline/_grain_tokenizer.py
```diff
@@ -64,3 +64,64 @@ def __setstate__(self, state):
...
+      return []
+
+    output_elements = []
+
+    for i in range(0, len(token_ids), max_len):
+      chunk = token_ids[i : i + max_len]
+      chunk_np = np.asarray(chunk, dtype=np.int32)
+      new_element = element.copy()
+
+      for feature_name in self.feature_names:
```
aireenmei: Replacing all the features with the chunked primary_feature can be a bit confusing here, and complicates this module. I think this way is cleaner: TokenizeAndChunk takes in a primary_feature param and only change that column, leaving all other features unchanged (copied to all chunks). Then in _grian_data_processing.py, move the rekey block (https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/input_pipeline/_grain_data_processing.py#L98-L100) to after tokenization to map 'text' to 'inputs' and 'targets'. What do you think?

aireenmei (commented):
Thanks for the feature! And great unit tests! Just some minor comments.

bzantium:
Thanks for the great feedback!

I've pushed the changes addressing all your points:

* **`base.yml` comment**: Added the comment to the `use_truncation` flag to point to the implementation classes as you suggested.
* **Tokenizer Refactoring**:
    * I've simplified both `TokenizeAndTrim` and `TokenizeAndChunk` to operate on a single `text_column`.
    * The `Rekey` transform has been moved to execute *after* the tokenization step for both code paths, which cleans up the pipeline logic nicely.

to: @aireenmei

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -95,9 +95,9 @@ def pretrain_preprocessing_pipeline(dataset, config, data_columns, tokenize, gra
     dataset = dataset.map(_input_pipeline_utils.NormalizeFeatures(data_columns, tokenize))
 
   assert len(data_columns) == 1
-  rekey_dict = {"inputs": "text", "targets": "text"}
-  dataset = dataset.map(_input_pipeline_utils.Rekey(rekey_dict))
+  text_column = "text"
```
aireenmei: we can have text_column = data_columns[0]? Let's move the rekey_dict definition to right before the Rekey op in line 131.

src/MaxText/input_pipeline/_grain_tokenizer.py
```diff
@@ -28,30 +28,26 @@ class TokenizeAndTrim(grain.MapTransform):
   """Tokenize and trim features to sequence length."""
 
   # pylint: disable=attribute-defined-outside-init
-  feature_names: str | Sequence[str]
```
aireenmei, line 31: Let's keep TokenizeAndTrim as it was to allow supporting multiple columns, I think it's needed for the DPO support (https://github.com/AI-Hypercomputer/maxtext/pull/947)

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -115,11 +115,21 @@ def pretrain_preprocessing_pipeline(dataset, config, data_columns, tokenize, gra
...
-        )
-    )
+    if config.use_truncation:
+      dataset = dataset.map(
+          _grain_tokenizer.TokenizeAndTrim(
+              text_column, config.max_target_length, config.add_bos, config.add_eos, tokenizer_model
+          )
+      )
+    else:
+      dataset = dataset.apply(
```
SurbhiJainUSC: What is the reason for using `dataset.apply` and not `dataset.map` similar to how `TokenizeAndTrim` uses at line 119?

aireenmei:
Looks like the github actions tests need to be triggered by a maintainer. Please take a look at the test failures. You can also run them locally

src/MaxText/input_pipeline/_grain_tokenizer.py
```diff
@@ -64,3 +60,55 @@ def __setstate__(self, state):
     self.__dict__.update(state)
     self._processor = None
     self._initialize_processor_lock = threading.Lock()
+
+
+@dataclasses.dataclass
```
SurbhiJainUSC, line 68: Most of the code looks similar to `TokenizeAndTrim` class. Can we create a base class, something like `TokenizerTransform` and then `TokenizeAndTrim` and `TokenizeAndChunk` can inherit from that class?

src/MaxText/input_pipeline/_grain_tokenizer.py
bzantium: You're right. I will roll back TokenizeAndTrim and TokenizeAndChunk can consume multiple columns.

src/MaxText/input_pipeline/_grain_data_processing.py
bzantium: That is because of `MapDataset`'s `map` only takes 1:1 mapping transformations (return `MapMapDataset`) but `apply` supports 1:N transformations (return `FlatMapMapDataset`). You can find it here: https://github.com/google/grain/blob/main/grain/_src/python/dataset/transformations/map.py
I also want it to support this with `map` for consistency 😂.
```py
  def map(
      self, transform: transforms.MapTransform | Callable[[T], S]
  ) -> MapDataset[S]:
    """Returns a dataset containing the elements transformed by ``transform``.

    Example usage::

      ds = MapDataset.range(5)
      ds = ds.map(lambda x: x + 10)
      list(ds) == [10, 11, 12, 13, 14]

    Args:
      transform: Either a ``MapTransform`` containing the ``map`` method or a
        callable that takes an element and returns a new element.

    Returns:
      A dataset containing the elements of the original dataset transformed by
      ``transform``.
    """
    # Loaded lazily due to a circular dependency (dataset <-> map).
    # pylint: disable=g-import-not-at-top
    from grain._src.python.dataset.transformations import (
        map as map_dataset,
    )
    # pylint: enable=g-import-not-at-top
    return map_dataset.MapMapDataset(parent=self, transform=transform)
```
```py
  def apply(
      self,
      transformations: transforms.Transformation | transforms.Transformations,
  ) -> IterDataset:
    """Returns a dataset with the given transformation(s) applied.

    Syntactic sugar to avoid dispatch by transformation type.

    Example usage::

      ds = grain.MapDataset.range(5).to_iter_dataset()
      ds = ds.apply([AddOne(), grain.transforms.Batch(2)])
      list(ds) == [np.ndarray([1, 2]), np.ndarray([3, 4]), np.ndarray([5])]

    Args:
      transformations: one or more transformations to apply.

    Returns:
      Dataset with the given transformations applied.
    """
    return apply_transformations(self, transformations)

def apply_transformations(
    ds: _ConsistentDatasetType,
    transformations: transforms.Transformation | transforms.Transformations,
) -> _ConsistentDatasetType:
  """Applies transformations to a dataset.

  DEPRECATED: Use `ds.apply(transformations)` instead.

  Args:
    ds: `MapDataset` or `IterDataset` to apply the transformations to.
    transformations: one or more transformations to apply.

  Returns:
    Dataset of the same type with transformations applied.
  """
  if not isinstance(transformations, Sequence):
    transformations = (transformations,)
  for transformation in transformations:
    match transformation:
      case transforms.Batch():
        ds = ds.batch(
            transformation.batch_size,
            drop_remainder=transformation.drop_remainder,
        )
      case transforms.MapTransform():
        ds = ds.map(transformation)
      case transforms.RandomMapTransform():
        ds = ds.random_map(transformation)
      case transforms.MapWithIndex():
        ds = ds.map_with_index(transformation)
      case transforms.FlatMapTransform():
        # Loaded lazily due to a circular dependency (dataset <-> flatmap).
        # pylint: disable=g-import-not-at-top
        from grain._src.python.dataset.transformations import flatmap
        # pylint: enable=g-import-not-at-top
        if isinstance(ds, MapDataset):
          ds = flatmap.FlatMapMapDataset(ds, transformation)
        else:
          ds = flatmap.FlatMapIterDataset(ds, transformation)
      case transforms.Filter():
        ds = ds.filter(transformation)
      case _:
        raise NotImplementedError(
            f"Transformation type: {transformation} is not supported."
        )
  return ds
  ```

src/MaxText/input_pipeline/_grain_tokenizer.py
bzantium: Yeah, that would be better!

src/MaxText/input_pipeline/_grain_data_processing.py
bzantium: Got it. I've updated the variable to `text_column` and moved the `rekey_dict` as requested. The changes have been pushed.

src/MaxText/input_pipeline/_grain_tokenizer.py
bzantium: Thanks for the context about the DPO support. I've reverted `TokenizeAndTrim` back to its original implementation (but refactored as recommended below) to support multiple columns, and I've pushed the change.

aireenmei (approved):
LGTM, thanks!

bzantium:
@aireenmei yeah, I think all look good for now. thanks for the detailed reviews!

bzantium:
I think `dataset.apply` requires latest `grain` so need to change requirements.txt like grain[parquet] @ https://github.com/google/grain/archive/bab58eabf0c94d16002b5d82dda8d8320edd3c7b.zip)

aireenmei:
> I think `dataset.apply` requires latest `grain` so need to change requirements.txt like grain[parquet] @ https://github.com/google/grain/archive/bab58eabf0c94d16002b5d82dda8d8320edd3c7b.zip)

The github runner uses requirements_with_jax_ai_image.txt. Could you change it with the PR and see if that fix the test?

Rohan-Bierneni:
Hi @bzantium can you rebase your pr and run the tests again. There were some changes to testing infra last week that require a rebase for successful image builds in unit tests.

bzantium:
@aireenmei @Rohan-Bierneni @SurbhiJainUSC 
I revised the code to use `WithOptionsIterDataset` and `apply_transformations` instead of `dataset.apply` because `grain` needs to be pre-compiled because it generates python file using pb2 with compilation but installation with latest commit does not perform this (pypi supports this so I decided to use grain==0.2.12).
Moreover, I found that add_bos/eos are used when building tokenizer not on transformation so I removed those arguments. I've checked current code works fine on my TPU cluster.

aireenmei:
> @aireenmei @Rohan-Bierneni @SurbhiJainUSC I revised the code to use `WithOptionsIterDataset` and `apply_transformations` instead of `dataset.apply` because `grain` needs to be pre-compiled because it generates python file using pb2 with compilation but installation with latest commit does not perform this (pypi supports this so I decided to use grain==0.2.12). Moreover, I found that add_bos/eos are used when building tokenizer not on transformation so I removed those arguments. I've checked current code works fine on my TPU cluster.

Thanks for the update! If grain==0.2.12 doesn't work, let me know if we need to request a new release from the grain team.

bzantium:
@aireenmei thanks for the fast reply! I've checked current implementation works fine with grain==0.12.2 on tpu-v6e but I think if it's possible for grain team to release new version, that would be nice in terms of this PR (can make code more neat as previous) and other upcoming features that I want to add soon.

bzantium:
https://pypi.org/project/grain/0.2.13/
new grain version is now available! I will test with the new code and push soon.

bzantium:
I've successfully tested this code on my tpus... why test failure happened...?!

aireenmei:
The corresponding tpu test are passing for MaxText head. Could you rebase again? Maxtext gpu unit test seems to not pick up the new Grain version. But at head it has picked up the newest one (https://github.com/AI-Hypercomputer/maxtext/actions/runs/18940014406/job/54076248442). I manually added pull ready label, not sure if this can make the test run automatically once you rebase and repush the branch

bzantium:
@aireenmei 
Thanks for checking. I also saw in the logs that the failed test was using `grain==0.2.12`.
I've just rebased the branch and modified the `requirements_with_jax_stable_stack_0_6_1_pipreqs.txt` file. 
It seems the tests didn't run automatically after the rebase.

aireenmei:
Tests are looking good. Could you fix the linter error? The script code_style.sh maybe helpful. Then squash the commits into one.

bzantium:
@aireenmei I think it might be fine this time.

parambole (approved):

aireenmei:
@bzantium Could you squash the 22 commits into one? MaxText allows merging 1 commit per PR

bzantium:
@aireenmei I squashed commits into one!

PR #2363, opened 2025-09-17
Remove batch truncation for fractional batch sizes.

chaserogo:
# Description

When running with a `per_device_batch_size < 1`, rather than splitting larger batches from the data loader into the smaller batch sizes for training, the larger batch is truncated. This PR changes it so that instead the larger batch is split into smaller batches which are looped over instead, so that no data is discarded.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/src/MaxText/multihost_dataloading.py
+++ b/src/MaxText/multihost_dataloading.py
@@ -18,8 +18,9 @@
 Adapted from Sholto's:
 https://github.com/sholtodouglas/multihost_dataloading
 """
+import itertools
 from functools import partial
-from typing import Union, Sequence
+from typing import Union, Sequence, Optional
 from collections.abc import Iterator, Iterable
 import time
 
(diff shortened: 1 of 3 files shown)


aireenmei:
Thanks for contributing to maxtext! Sorry for the delay in review. Have you completed the CLA? Could you rebase the PR, check the checklist, so we can rerun the tests?

aireenmei (commented):
Could you add a unit test in multihost_dataloading_test.py for this feature?

src/MaxText/multihost_dataloading.py
```diff
@@ -91,12 +87,39 @@ def reset(self):
     self.out_of_data = False
     self.last_local_data = None
 
+    sharded_iter = self._base_iter()
+    if self.microbatch_size_to_run:
+       self.local_iterator = itertools.chain.from_iterable(
```
SurbhiJainUSC, line 92: This can be simplified:
```
def microbatch_generator():
    for global_batch in sharded_iter:
        yield from self.explode_to_micro(global_batch)

if self.microbatch_size_to_run:
  self.local_iterator = microbatch_generator()
else:
  self.local_iterator = sharded_iter
```

PR #2367, opened 2025-09-18
Fix error in AttentionTest when upgrading to jax 0.7.2 

Rohan-Bierneni:
# Description

This is the error with some of the AttentionTests when upgrading to jax 0.7.2:
`
ValueError: Received incompatible devices for jitted computation. Got argument x of _identity_fn with shape bfloat16[4,512,2048] and device ids [0] on platform TPU and explicit output sharding with device ids [0, 2, 1, 3] on platform TPU
`

Looking at the test file, seems like the input tensors (lnx, decoder_segment_ids, and decoder_positions) are not sharded to match the mesh across all 4 devices. They are defined on device[0] instead of matching the model. So we should explicitly shard the input tensors to match mesh_cp before calling the model. Implemented the fix by using a jax.device_put with NamedSharding that matches the 4-device mesh defined in the testcase. 

# Tests

Tests passed locally with changes: https://paste.googleplex.com/5374926655717376
Extra verification if the unit tests here pass as well.

# Checklist

### DIFF
--- a/tests/attention_test.py
+++ b/tests/attention_test.py
@@ -26,7 +26,7 @@ from absl.testing import parameterized
 
 import numpy as np
 
-from jax.sharding import Mesh
+from jax.sharding import Mesh, NamedSharding, PartitionSpec
 import jax
 import jax.numpy as jnp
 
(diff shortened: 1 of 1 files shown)


NuojCheng (approved):
I think we can merge it after you update the test image back. Thank you @Rohan-Bierneni !

PR #2370, opened 2025-09-19
Migrate LlamaDecoderLayer to NNX

RissyRan:
# Description

This is co-work with @bvandermoon and @shuningjin 

Example to migrate llama2 to NNX. This change is based on @bvandermoon's [branch](https://github.com/AI-Hypercomputer/maxtext/compare/bvandermoon-nnx-single-kvcache-memory-issue) and keep it untouched for sanity check. This PR squashes all commits into one.

FIX: b/445953113, b/439070153

# Tests

Details in b/445953113#comment13

1. decode cmd: https://paste.googleplex.com/4995561203826688
- decode diff: https://diff.googleplex.com/#key=8kLLbM6jhJky, same as main (memstat after init, output)
2. train cmd: https://paste.googleplex.com/5801717128101888
- train diff: https://diff.googleplex.com/#key=9wKB5thwoIP9, same as main (TFLOP for last steps)
3. jetstream cmd, PDB=6: https://paste.googleplex.com/4992195518136320 (b/445953113#comment7)
- jetstream diff: https://diff.googleplex.com/#key=fqj8G4bz5Bom, same as main (acc, similar speed)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #2373, opened 2025-09-19
Better Sharding for dsv3 moe layer

suexu1025:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/src/MaxText/layers/moe.py
+++ b/src/MaxText/layers/moe.py
@@ -300,8 +300,8 @@ class RoutedMoE(nnx.Module):
     self.quant = quant
     self.rngs = rngs
 
-    self.wi_kernel_axes = ("exp", "embed_no_exp", "mlp")
-    self.wo_kernel_axes = ("exp", "mlp", "embed_no_exp")
+    self.wi_kernel_axes = ("embed_no_exp", "exp", "mlp")
+    self.wo_kernel_axes = ("embed_no_exp", "mlp", "exp")
 
     self.gate = GateLogit(
         in_features_shape=self.config.emb_dim,
@@ -900,9 +900,9 @@ class RoutedMoE(nnx.Module):
 
     # w0, w1, wo needs to be un sharded on fsdp / fsdp_transpose axis, so use
     # mlp_no_fsdp axis
-    w0_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    w1_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    wo_pspec = nn.logical_to_mesh_axes(("exp", "mlp_no_fsdp", "embed_tensor_transpose"))
+    w0_pspec = nn.logical_to_mesh_axes(("embed_tensor_transpose", "exp", "mlp_no_fsdp"))
+    w1_pspec = nn.logical_to_mesh_axes(("embed_tensor_transpose", "exp", "mlp_no_fsdp"))
+    wo_pspec = nn.logical_to_mesh_axes(("embed_tensor_transpose", "mlp_no_fsdp", "exp"))
     if isinstance(w0_kernel, aqt.QTensor):
       w0_pspec = aqt.partition_spec(w0_pspec, (1,), w0_kernel.dtype, use_bias=False)
     if isinstance(w1_kernel, aqt.QTensor):
--- a/src/MaxText/layers/moe.py
+++ b/src/MaxText/layers/moe.py
@@ -300,8 +300,13 @@ class RoutedMoE(nnx.Module):
     self.quant = quant
     self.rngs = rngs
 
-    self.wi_kernel_axes = ("embed_no_exp", "exp", "mlp")
-    self.wo_kernel_axes = ("embed_no_exp", "mlp", "exp")
+    # special sharding for dsv3
+    if self.config.num_experts == 256:
+      self.wi_kernel_axes = ("embed_no_exp", "exp", "mlp")
+      self.wo_kernel_axes = ("embed_no_exp", "mlp", "exp")
+    else:
+      self.wi_kernel_axes = ("exp", "embed_no_exp", "mlp")
+      self.wo_kernel_axes = ("exp", "mlp", "embed_no_exp")
 
     self.gate = GateLogit(
         in_features_shape=self.config.emb_dim,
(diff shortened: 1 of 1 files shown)


src/MaxText/layers/moe.py
```diff
@@ -300,8 +300,13 @@ def __init__(
     self.quant = quant
     self.rngs = rngs
 
-    self.wi_kernel_axes = ("exp", "embed_no_exp", "mlp")
-    self.wo_kernel_axes = ("exp", "mlp", "embed_no_exp")
+    # special sharding for dsv3
```
gobbleturk: @richjames0 for all sharding changes =D

src/MaxText/layers/moe.py
```diff
@@ -300,8 +300,13 @@ def __init__(
...
+    if self.config.num_experts == 256:
```
gobbleturk: We need a better logic for such a specific conditional. It should be controllable via a base.yml config argument like ("expert_first_dim")

This change doesn't surprise me that expert first is the most performant. Are there are any downsides to just flipping it here? Other than having to modify all of our checkpoint conversion scripts =D

gobbleturk: If modifying the checkpoint conversion scripts is really the hardest part we can support both options (with a base.yml config) for a while with a deprecation warning on the current behavior to migrate to the new one.

gobbleturk (commented):
@richjames0 for thoughts =D

src/MaxText/layers/moe.py
```diff
@@ -900,9 +905,15 @@ def gmm(inputs, kernel, group_sizes, expert_assignments):
 
     # w0, w1, wo needs to be un sharded on fsdp / fsdp_transpose axis, so use
     # mlp_no_fsdp axis
-    w0_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    w1_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    wo_pspec = nn.logical_to_mesh_axes(("exp", "mlp_no_fsdp", "embed_tensor_transpose"))
+    # special sharding for dsv3 to remove overhead between gmm/AG
+    if self.config.num_experts == 256:
```
RissyRan: Similar comment with Matt. For instance, we could have a flag similar to `expert_shard_attention_option`, and we could have `expert_shard_mlp_option` or similar. 

This condition `self.config.num_experts == 256` may raise up question, how about 128 or 512 experts?

RissyRan (commented):
Thanks Qinwen! Could you also attach test results in the description?

src/MaxText/layers/moe.py
suexu1025: I added `shard_by_num_of_expert` to only enable it when it expert_dim dividable by fdsp_shard_dim

src/MaxText/layers/moe.py
suexu1025: I added shard_by_num_of_expert to only enable it when it expert_dim dividable by fdsp_shard_dim

suexu1025:
> Thanks Qinwen! Could you also attach test results in the description?
pls see the test results here
https://screenshot.googleplex.com/5xAhsHPUdHFS6zc

src/MaxText/layers/moe.py
```diff
@@ -427,6 +432,9 @@ def get_topk(self, gate_logits, pre_bias_logits, rngs=None):
 
     return top_k_weights, top_k_indices
 
+  def shard_by_num_of_expert(self) -> bool:
```
gobbleturk: I would remove this - a user would not expect that setting "fsdp_shard_on_exp" is not enough. If you want to perform these checks I will perform them in pyconfig and fail loudly when they are not met when fsdp_shard_on_exp is true

src/MaxText/layers/moe.py
```diff
@@ -300,8 +300,13 @@ def __init__(
     self.quant = quant
     self.rngs = rngs
 
-    self.wi_kernel_axes = ("exp", "embed_no_exp", "mlp")
-    self.wo_kernel_axes = ("exp", "mlp", "embed_no_exp")
+    # special sharding for dsv3
+    if self.shard_by_num_of_expert():
+      self.wi_kernel_axes = ("embed_no_exp", "exp", "mlp")
```
gobbleturk: Should this actually be ("embed_no_exp", None, "mlp")? We never actually want the expert physical axis to shard this dimension

gobbleturk: Similar for below, I think "exp" should be replaced by None in all of these? This is identical to what you have when exp=1 which we can enforce in pyconfig

src/MaxText/layers/moe.py
suexu1025: I changed the validation to pyconfig

RissyRan:
> > Thanks Qinwen! Could you also attach test results in the description?
> > pls see the test results here
> > https://screenshot.googleplex.com/5xAhsHPUdHFS6zc

Thanks! I meant to say the original links, so that we could directly to see profiles or anything. One example [here](https://github.com/AI-Hypercomputer/maxtext/pull/2370), usually put original links for reviewers.

src/MaxText/layers/moe.py
RissyRan: Could you move `# special sharding for dsv3` inside of `if self.config.fsdp_shard_on_exp:` branch?

src/MaxText/layers/moe.py
```diff
@@ -900,9 +906,15 @@ def gmm(inputs, kernel, group_sizes, expert_assignments):
 
     # w0, w1, wo needs to be un sharded on fsdp / fsdp_transpose axis, so use
     # mlp_no_fsdp axis
-    w0_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    w1_pspec = nn.logical_to_mesh_axes(("exp", "embed_tensor_transpose", "mlp_no_fsdp"))
-    wo_pspec = nn.logical_to_mesh_axes(("exp", "mlp_no_fsdp", "embed_tensor_transpose"))
+    # special sharding for dsv3 to remove overhead between gmm/AG
```
RissyRan: Same. Could you move this comment inside of `if self.config.fsdp_shard_on_exp:`?

src/MaxText/pyconfig.py
```diff
@@ -1049,6 +1050,11 @@ def validate_ring_of_experts_parallelism(raw_keys):
   if raw_keys["use_ring_of_experts"] and not using_expert_parallelism(raw_keys):
     raise ValueError("Ring-of-experts requires expert-parallelism to be enabled.")
 
+def validate_shard_fsdp_on_expert_parallelism(raw_keys):
+  if raw_keys["fsdp_shard_on_exp"] and raw_keys["num_experts"] % raw_keys["ici_fsdp_parallelism"]!=0:
```
RissyRan, line 1054: Could you help add one more assertion for TP sharding? if `raw_keys["fsdp_shard_on_exp"] and (using_tensor_parallelism(raw_keys) or using_expert_parallelism(raw_keys) --> raise only supported fsdp in this config`

src/MaxText/layers/moe.py
suexu1025: updated

src/MaxText/pyconfig.py
suexu1025: thank you, added

src/MaxText/pyconfig.py
```diff
@@ -1049,6 +1050,11 @@ def validate_ring_of_experts_parallelism(raw_keys):
   if raw_keys["use_ring_of_experts"] and not using_expert_parallelism(raw_keys):
     raise ValueError("Ring-of-experts requires expert-parallelism to be enabled.")
 
+def validate_shard_fsdp_on_expert_parallelism(raw_keys):
+  if raw_keys["fsdp_shard_on_exp"] and raw_keys["num_experts"] % raw_keys["ici_fsdp_parallelism"]!=0: 
+    raise ValueError("fsdp_shard_on_exp requires num_experts is divisiable by ici_fsdp_parallelism.")
+  if raw_keys["fsdp_shard_on_exp"] and (using_tensor_parallelism(raw_keys) or useing_expert_parallelism(raw_keys)): 
+    raise ValueError("fsdp_shard_on_exp requires ici_expert_parallelism = 1 and ici_tensor_parallelism/ici_tensor_transpose_parallelism = 1.")
```
RissyRan, line 1057: Nit: probably just say `fsdp_shard_on_exp does not support EP and TP shardings`?

RissyRan (approved):
Thanks @suexu1025

src/MaxText/layers/attention_mla.py
```diff
@@ -398,7 +398,7 @@ def _init_projections(self, inputs_q_shape: Tuple, inputs_kv_shape: Tuple) -> No
           out_features_shape=self.q_lora_rank,
           axis=-1,
           kernel_init=self.kernel_init,
-          kernel_axes=("embed", "q_lora"),
+          kernel_axes=("embed", "q_lora_up_proj"),
```
RissyRan, line 401: This part is helpful for all MLA models or only DS V3? 

cc @richjames0 another case.

PR #2382, opened 2025-09-23
GPT-OSS: Add user guide and tests

shuningjin:
# Description

Main change: for gpt-oss
- add user guide: 
  - `end_to_end/tpu/gpt_oss/run_gpt_oss.md`
  - use gpt-oss-20b on v5p-8 (sparse_matmul + megablox) as example
- add test script
  - `end_to_end/tpu/gpt_oss/20b/test_gpt_oss.sh`
  - `end_to_end/tpu/gpt_oss/120b/test_gpt_oss.sh`
- pyconfig validation

Other change: format output from forward_logit_checker, correct typos in other files

# Tests

single script for tpu: checkpoint conversion, logit check, train, sft, decode (instead of splitting conversion to another)

gpt-oss-20b, end-to-end test on v5p-8 cluster, takes around 45min
- cmd: https://paste.googleplex.com/5491652676419584
- log: https://cloudlogging.app.goo.gl/DzrB6DY1qw576Q747
- to faciliate the test, I have uploaded the dequantized checkpoint and golden logits to GCS
- note: gpt-oss with flash attention requires jax>=0.7.2

The test will also be onboarded to XLML: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/913

## Additional Notes

HF tokenizer:
- Here we directly use the HF tokenizer, instead of storing it as assets in the repo. This practice seems fine, as long as the HF access token is handled correctly.

HF access token:
- For MaxText native training, the access token is not needed because the HF model/tokenizer (gpt-oss) and HF dataset (used in SFT) are not gated. 
- However, decoding with HF tokenizer still needs access token despite that the model is not gated. The tokenizer is built from JetStream utils, which handles things differently from MaxText.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #2394, opened 2025-09-24
Moved tunix to setup.sh to only install when DEVICE=tpu

Rohan-Bierneni:
# Description

Tunix has a requirement on jax[tpu], which is causing errors and bugs in setting up maxtext environment for GPUs since it is in requirements.txt.

When using the JAX AI Image flow, since jax is installed as part of the image and tunix is installed when using JAX AI Image as a base image, there is a weird bug where libtpu is being installed in the final maxtext GPU image. This is probably due to jax and tunix not being installed in the same step, which would normally throw an error but is hidden in this 2-stage flow.

When using setup.sh, since both jax[cuda12] and tunix are being installed in the same step, pip correctly recognizes that tunix depends on jax[tpu] and will throw an error here.

To fix this, we will move tuni from requirements.txt to setup.sh and install the pinned commit when MODE=stable and the unpinned repo when MODE=nightly. We already do something similar for transformer_engine which is a gpu-only dependency.

Note: when tunix supports GPUs, we can move it back to requirements.txt
Note2: Added a small change for migration to jax-ai-image from jax-stable-stack in the entrypoint for JAX AI Images.

# Tests

Gpu image built in the unit test run does not contain libtpu: https://screenshot.googleplex.com/AxYbecHHimTdRx8

Tpu image has tunix installed: https://screenshot.googleplex.com/3JM78mv6ZJR76ju

PR is working as intended

# Checklist

### DIFF
--- a/.github/workflows/RunTests.yml
+++ b/.github/workflows/RunTests.yml
@@ -111,6 +111,7 @@ jobs:
       device_name: a100-40gb-4
       cloud_runner: linux-x86-a2-48-a100-4gpu
       pytest_marker: 'not cpu_only and not tpu_only and not integration_test'
+      pytest_addopts: '--ignore=tests/sft_hooks_test.py'
       xla_python_client_mem_fraction: 0.65
       tf_force_gpu_allow_growth: true
       container_resource_option: "--shm-size 2g --runtime=nvidia --gpus all --privileged"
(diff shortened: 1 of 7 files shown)


RissyRan:
It seems this label doesn't work here, but works as expected in other [PR](https://github.com/AI-Hypercomputer/maxtext/pull/2382). Wondering if it's due to the `draft` status?

Updated: tried another [PR](https://github.com/AI-Hypercomputer/maxtext/pull/2393) in ready status (not draft), and it works well.

parambole (approved):
LGTM. Thank you for the quick fix.

bvandermoon (approved):

PR #2413, opened 2025-09-26
Llama4 image tiling and basic batching for Multimodal SFT

NicoGrande:
# Description

This PR adds image tiling support for llama4 Multimodal SFT, which allows images to preserve their original resolution and aspect ratio when input to the ViT. More specifically, this PR adds:

- `pad_to_max_tiles` method in `multimodal_utils.py` which pads an image tensor (B, T, C, H, W) across the the tile dimension to ensure uniformity when batching image data. This method also returns a mask indicating true tiles for that image.
- Changes to `_merge_mm_embeddings_inner` in `multimodal_utils.py` to ignore padded tiles when merging vision embeddings and text embeddings
- Unit tests in `multimodal_utils.py` to cover new functionality.

FIXES: b/447624435

# Tests

**Llama4 Tests:**
SFT [logs](https://cloudlogging.app.goo.gl/Y98Dh8Te8asUPqqX7)

Decode [logs](https://cloudlogging.app.goo.gl/AQj3i8iiQec62rFp7)

MM Eval [logs](https://cloudlogging.app.goo.gl/Va8As4CCMdwxW6TZA)

**Gemma3 Tests:**
Confirmed local decode works with the following command:
```
python -m MaxText.decode src/MaxText/configs/base.yml model_name=gemma3-4b tokenizer_path=assets/tokenizer.gemma3 load_parameters_path=gs://maxtext-gemma/unified/gemma3/4b/unscanned/2025-08-09-01-17/0/items per_device_batch_size=1 run_name=ht_test max_prefill_predict_length=272 max_target_length=300 steps=1 async_checkpointing=false scan_layers=false use_multimodal=true prompt=\'Describe\ image\ \<start_of_image\>\' image_path=\'src/MaxText/test_assets/test_image.jpg\' attention=\'dot_product\'
```

# Checklist


src/MaxText/maxengine.py
```diff
@@ -327,17 +327,16 @@ def quantize_params(self, state, rng: PRNGKeyType | None = None):
...
           jnp.ones((1, self.config.max_prefill_predict_length), dtype=jnp.int32),
-          encoder_images=jnp.ones(
-              multimodal_utils.get_dummy_image_shape_for_init(
-                  self.config.model_name, batch_size=self.config.micro_batch_size_to_train_on
-              ),
-              dtype=jnp.float32,
-          )
-          if self.config.use_multimodal
+          encoder_images=jnp.ones(image_shape, dtype=jnp.float32) if self.config.use_multimodal else None,
+          encoder_image_masks=jnp.ones(image_shape[:2], dtype=jnp.int32)
```
aireenmei, line 339: Should we also add `if self.config.use_multimodal else None` for encoder_image_masks?

src/MaxText/multimodal_utils.py
```diff
@@ -263,6 +267,48 @@ def pad_to_best_fit_jax(
...
+    raise ValueError(f"Number of tiles {num_tiles} exceeds max_num_tiles {max_num_tiles}")
+
+  # Create a new array filled with zeros for padding
+  padded_tiles = np.zeros((max_num_tiles, num_channels, height, width), dtype=images.dtype)
+  mean_array = np.array(mean_values, dtype=images.dtype).reshape((1, num_channels, 1, 1))
+  std_array = np.array(std_values, dtype=images.dtype).reshape((1, num_channels, 1, 1))
+
+  # Normalize the tile padding to zero mean and unit variance
+  padded_tiles *= pixel_value_rescale_factor
+  padded_tiles -= mean_array
```
aireenmei: Did we do similar normalization in the original code? I see line 403 we do another normalization later. Could you help me understand?

src/MaxText/input_pipeline/_input_pipeline_utils.py
```diff
@@ -434,6 +437,25 @@ def _pad_text(self, x, max_length, pad_id):
     pad_amount = [(0, pad_amount)] + [(0, 0)] * (len(x.shape) - 1)
     return np.pad(x, pad_amount, constant_values=pad_id)[: self.max_length]
 
+  def _pad_image_mask(self, image_masks):
```
aireenmei: This is very similar to the _pad_image function, could we combine them into a _pad_image_or_mask function?

aireenmei (commented):
Great work! Since many modules are shared between Llama4 and Gemma3, could you also test Gemma3 decode with this script in this PR https://github.com/AI-Hypercomputer/maxtext/pull/2350?

src/MaxText/layers/llama4.py
```diff
@@ -754,15 +758,18 @@ def __call__(
...
     Args:
-      inputs: Input tensor of shape [batch_size, num_tiles, num_channels_for_vit, tile_size_for_vit, tile_size_for_vit]
+      inputs: Input tensor of shape:
+              [batch_size * num_images, num_tiles, num_channels_for_vit, tile_size_for_vit, tile_size_for_vit]
       deterministic: Whether to use deterministic mode (disables dropout)
 
     Returns:
-      Final hidden states from the vision encoder of shape [batch_size, num_tiles, num_patches, vision_output_dim_for_vit]
+      Final hidden states from the vision encoder of shape:
+      [batch_size * num_images, num_tiles, num_patches, vision_output_dim_for_vit]
```
hengtaoguo, line 767: Since now we support multiple images and tiles at the same time, are we doing paddings for both dimensions?

src/MaxText/maxengine.py
hengtaoguo, line 339: Maybe also add a comment line explaining the meaning of `encoder_image_masks`?

src/MaxText/multimodal_utils.py
```diff
@@ -495,6 +545,8 @@ def prepare_text_for_image_fusion(texts, model_name, processor_output=None):
     num_images = len(processor_output) if isinstance(processor_output, list) else 1
     return add_extra_tokens_for_images_gemma3(texts, max_num_images=num_images)
   if model_name in ["llama4-17b-16e", "llama4-17b-128e"]:
+    # TODO(nicogrande): support multiple images
```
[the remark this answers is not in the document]
hengtaoguo: Following this, what would happen if there are multiple images with different number of tiles in one example?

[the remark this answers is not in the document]
NicoGrande: Each image would be padded across the tile dimension individually, so there would be no issues with stacking. Suppose we had 2 images corresponding to the same query. Also suppose the maximum number of images is configured to be 5. In this case, each image would be padded to 16 across the tile dimension, and the image dimension would be padded to 5. However, to maintain 5-dimensional data, we would fold the image dimension into the batch dimension. This means we would end up with a (5 * B, 16, C, H, W) dimensional tensor.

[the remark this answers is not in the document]
NicoGrande: I can add multi-image support in a follow up PR.

src/MaxText/maxengine.py
NicoGrande: The condition `if self.config.use_multimodal and "llama4" in self.config.model_name` is already present in line 339

src/MaxText/multimodal_utils.py
NicoGrande: Yes, it is the same normalization. The normalization that happens in line 403 is applied on the original input image, as well as the padded regions appended to the image to achieve the desired tile shape. The code referenced in this comment applies the same transformation on the new padded tiles, which are added later in line 413. The normalization itself needs to be applied slightly differently as the input to the `_normalize_images` is expected to be 3-dim.

src/MaxText/layers/llama4.py
NicoGrande: Yes, both dimensions would need to be padded. Note, we do not yet support multiple images for llama4. This will be added in a follow up PR.

hengtaoguo (approved):

src/MaxText/maxengine.py
NicoGrande: Done!

src/MaxText/input_pipeline/_input_pipeline_utils.py
NicoGrande: Done!

PR #2437, opened 2025-10-01
add grad_dtype configurable

suexu1025:
# Description

1. add the gradient dtype configuration.
2. perf diff 4%. 
grad_in_fp32: https://xprof.corp.google.com/overview_page/qinwen-10410456020302047643
grad_in_bf16: https://xprof.corp.google.com/overview_page/qinwen-14894563938690716073

3. loss diff:
final eval loss 4.55[cloudlog](https://cloudlogging.app.goo.gl/T46zPU6xM3js9dMZ6)
final eval loss 4.60[cloudlog](https://cloudlogging.app.goo.gl/2ydfHieG1XWhxXN28)

# Checklist


src/MaxText/train.py
```diff
@@ -255,6 +255,8 @@ def reshape_to_microbatch_accumulations(batch_arr):
         extra_dpo_args = [reference_params]
     grad_func = jax.value_and_grad(_loss_fn, argnums=4, has_aux=True)
     (loss, aux), raw_grads = grad_func(model, config, data, dropout_rng, state.params, *extra_dpo_args, is_train=True)
+
+  raw_grads = jax.tree_util.tree_map(lambda x: x.astype(jnp.bfloat16) if x.dtype == jnp.float32 else x, raw_grads)
```
bvandermoon: Should this be based on config? Or is it always the case that this gets mapped back to `jnp.bfloat16`?

suexu1025: nice catch, updated.

bvandermoon: Thanks @suexu1025. And are we also always checking for `x.dtype == jnp.float32`? Or can that be config-based too?

suexu1025: yes, for here we only want `grad_dtype` to convert from fp32 to lower precision

PR #2462, opened 2025-10-07
Specify maxtext[tpu] package using a requirements.txt generated by se…

kanglant:
…ed-env. This is a prerequisite for the gpu extra in a follow-up PR.

# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist


setup.sh
```diff
@@ -163,7 +163,7 @@ fi
 
 # Install maxtext package
 if [ -f 'pyproject.toml' ]; then
-  python3 -m uv pip install -e . --no-deps --resolution=lowest
+  python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest
   install_maxtext_github_deps
 fi
```
bvandermoon, line 168: Is this currently also running on GPUs? Maybe I am just missing it, but I don't see a tpu nesting around this block

base_requirements/requirements.txt
```diff
@@ -0,0 +1,42 @@
...
+tensorboard-plugin-profile
+tensorboardx
+tensorflow-datasets
+tensorflow-text
+tensorflow
+tiktoken
+transformers
+qwix
+google-jetstream @ https://github.com/AI-Hypercomputer/JetStream/archive/daedc21c393f23449fb54ddc4f75fca34348ea9c.zip
+mlperf-logging @ https://github.com/mlcommons/logging/archive/38ab22670527888c8eb7825a4ece176fcc36a95d.zip
```
bvandermoon, line 42: Do we need to delete the old `requirements.txt` also? Or will that happen later?

setup.sh
kanglant: I've split the package support for GPU and TPU extras into separate PRs. This specific PR focuses on creating the TPU extra. Since the current maxtext package is already TPU-only, I haven't changed the core logic in setup.sh; I've just updated it to adopt the TPU extra package. Once the GPU support is added in a subsequent PR, we can refactor setup.sh to handle the installation of the GPU package as well.

base_requirements/requirements.txt
kanglant: We will delete the old `requirements.txt` in later PRs along with the refactor of setup.sh.

setup.sh
bvandermoon: Thanks @kanglant. Does this block need to have something like this instead then?

```
if [ -f 'pyproject.toml' && $DEVICE == "tpu" ];
```

kanglant: I just updated setup.sh.

For the stable TPU package, it now contains all necessary dependencies, so we can install it directly using the recommended method (lines 169-170).

Since this PR doesn't yet include the stable GPU package, I've moved the installation of the old requirements.txt and maxtext package with no-deps to the bottom block for stable gpu. In the later GPU package PR, we'll remove the dependency on that old requirements.txt from setup.sh.

WDYT? @bvandermoon @SamuelMarks

setup.sh
```diff
@@ -155,23 +155,8 @@ if [[ "$MODE" == "nightly" ]]; then
...
-    # stable or stable_stack mode: Install with pinned commits
-    echo "Installing requirements.txt with pinned commits."
-    python3 -m uv pip install --no-cache-dir -U -r requirements.txt
-fi
-
-# Install maxtext package
-if [ -f 'pyproject.toml' ]; then
-  python3 -m uv pip install -e . --no-deps --resolution=lowest
-  install_maxtext_github_deps
 fi
```
bvandermoon, line 168: Do we still need this for `MODE=stable_stack`?

kanglant: I think that's just a comment that was forgotten and should be deleted. See line 250.

bvandermoon: Got it, thanks @kanglant. You are right that `stable_stack` only copies, but does not use `setup.sh`: https://github.com/AI-Hypercomputer/maxtext/blob/main/maxtext_jax_ai_image.Dockerfile

bvandermoon (approved):
@kanglant were you able to test the setup script/Docker build works? Do you have the `benchmark_runner` set up to run a sample workload?

richjames0 (approved):
lgtm

setup.sh
SamuelMarks: As said earlier, I recommend creating three to five .txt files for dependencies and referencing them using https://pypi.org/project/hatch-requirements-txt/ … e.g.:
- `requirements-base.txt`
- `requirements-gpu.txt`
- `requirements-gpu-nightly.txt`
- `requirements-tpu.txt`
- `requirements-tpu-nightly.txt`

`requirements-base.txt` should be what's installed by default, and for all. The rest are self-explanatory.

kanglant: We've previously discussed this and concluded that having a default requirements-base.txt for the default maxtext package doesn't make sense, as maxtext should always be targeted for either TPU or GPU. We want to use an empty dep list instead.

To maintain clear requirement file separation, I've created different folders to group the "base" requirements files (which users can update manually and only contains direct maxtext deps in addition to JAX) and the generated requirements files (which are created by the `seed-env` command based on the "base"). 

This PR adds support for maxtext[tpu], and a follow-up PR will address maxtext[gpu]. Support for nightly builds will be discussed and implemented later.

kanglant:
> @kanglant were you able to test the setup script/Docker build works? Do you have the `benchmark_runner` set up to run a sample workload?

After offline discussion, we will refactor the setup scripts and docker build in subsequent PRs.

PR #2463, opened 2025-10-07
Create maxtext gpu package and unit/integration tests

kanglant:
# Description

This PR adds GPU extras to the maxtext package, allowing users to install the necessary dependencies for a GPU environment by running `uv pip install -e .[cuda12] --resolution=lowest`. It also integrates GPU package tests into the CI workflow.

FIXES: b/450562915

# Tests

I've updated the package test and it is passed.

# Checklist

### DIFF
--- a/clean_py_env_gpu.Dockerfile
+++ b/clean_py_env_gpu.Dockerfile
@@ -1,0 +1,74 @@
+# This Dockerfile is used to build a Docker image containing all the
+# system-level dependencies required for maxtext. Since the maxtext
+# package itself includes all the necessary Python dependencies, this
+# image is designed to have a clean Python environment with just pip
+# and uv installed.
+#
+# Build a docker image by running:
+# `docker build --build-arg DEVICE=cuda12 -t <docker_image_name>:<tag> -f clean_py_env_gpu.Dockerfile .`
+# Or
+# `docker build --build-arg PYTHON_VERSION=3.11 --build-arg DEVICE=cuda12 -t <docker_image_name>:<tag> -f clean_py_env_gpu.Dockerfile .`
+#
+# How to upload the image to Google Container Registry (GCR):
+# e.g., DEVICE=cuda12 and PYTHON_VERSION=3.12
+#.  gcloud init
+#   gcloud auth configure-docker
+#   docker tag <docker_image_name>:<tag> gcr.io/tpu-prod-env-multipod/maxtext-unit-test-cuda12:py312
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -27,6 +27,8 @@ dependencies = []
 
 [tool.hatch.metadata.hooks.requirements_txt.optional-dependencies]
 tpu = ["generated_requirements/tpu-requirements.txt"]
+gpu = ["generated_requirements/cuda12-requirements.txt"]
+cuda12 = ["generated_requirements/cuda12-requirements.txt"]
```
bvandermoon: The idea here is that we will add other cuda versions like `cuda13`, etc. over time, but we will have one of them be the default for gpu, right? How will we decide which is the default?

clean_py_env_gpu.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+#   docker push gcr.io/tpu-prod-env-multipod/maxtext-unit-test-gpu:py312
+
+# Default to Python 3.12.
+ARG PYTHON_VERSION=3.12
+
+FROM nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04
+# TODO: enable cuda 13, e.g., nvcr.io/nvidia/cuda-dl-base:25.08-cuda13.0-devel-ubuntu24.04.
+
+# Set the working directory in the container
+WORKDIR /maxtext
```
bvandermoon, line 26: Is this image used in the PR tests currently? How is it integrated into the workflow? I don't see where it is referenced

pyproject.toml
kanglant: That's a good point. To avoid any ambiguity for users, it's better to have them specify the exact CUDA version they need rather than relying on a default gpu option. Let me remove it.

clean_py_env_gpu.Dockerfile
kanglant: Yes. The image is `gcr.io/tpu-prod-env-multipod/maxtext-unit-test-gpu:py312`. You can find it being used in .github/workflows/run_tests_against_package.yml on line 60.

kanglant: Retagged. The cuda12 job is now using `gcr.io/tpu-prod-env-multipod/maxtext-unit-test-cuda12:py312`

pyproject.toml
bvandermoon: Thank you @kanglant. Can you please also update the example command in the description?

kanglant: Done!

PR #2468, opened 2025-10-08
Pinning JAX=0.7.0 to allow google-tunix=0.1.1

hengtaoguo:
# Description

Current `cpu_unit_tests` are [failing](https://github.com/AI-Hypercomputer/maxtext/actions/runs/18350713027/job/52270253358). The reason is that google-tunix 0.1.1 is downgrading to [JAX 0.7.1](https://screenshot.googleplex.com/5wusABhmVmhaHfr), which is known to have pallas mosaic errors. This PR re-installs `jax==0.7.0` after tunix installation, so that we have can keep `google-tunix==0.1.1` for new features. We should recover when tunix stopped pinning JAX 0.7.1.

[b/450286600](https://b.corp.google.com/issues/450286600)

# Tests

N/A

# Checklist

### DIFF
--- a/maxtext_jax_ai_image.Dockerfile
+++ b/maxtext_jax_ai_image.Dockerfile
@@ -50,6 +50,8 @@ RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg
 # Install google-tunix for TPU devices, skip for GPU
 RUN if [ "$DEVICE" = "tpu" ]; then \
         python3 -m pip install 'google-tunix>=0.1.0'; \
+        # TODO: Once tunix stopped pinning jax 0.7.1, we should remove our 0.7.0 version pin (b/450286600)
+        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0'; \
   fi
 
 # Now copy the remaining code (source files that may change frequently)


SurbhiJainUSC:
Thank you Hengtao for fixing this.

maxtext_jax_ai_image.Dockerfile
```diff
@@ -50,6 +50,8 @@ RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg
 # Install google-tunix for TPU devices, skip for GPU
 RUN if [ "$DEVICE" = "tpu" ]; then \
         python3 -m pip install 'google-tunix>=0.1.0'; \
+        # TODO: Once tunix stopped pinning jax 0.7.1, we should remove our 0.7.0 version pin (b/450286600)
+        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0'; \
```
bvandermoon, line 54: What if a user is building with a JAX 0.6.x base image? @parambole @Rohan-Bierneni

hengtaoguo: I'm not sure about 0.6.x; but when I tried building on top of a 0.7.0 image, some dependencies in our requirements will force uninstall 0.7.0 and upgrade to a newer version (likely 0.7.2). Then Tunix 0.1.1 will downgrade from 0.7.2 to 0.7.1 at the very end.

https://screenshot.googleplex.com/Bt4ZFB9R8eKgdft

hengtaoguo: So my guess is: even the user starts from 0.6.x base image, it will result in 0.7.1 eventually after satisfying all dependencies (flax+orbax) and tunix.

bvandermoon: @parambole @Rohan-Bierneni please confirm

bvandermoon: Approving the PR to unblock the fix for merging MaxText PRs. We can find a better approach later for old Docker images if needed (like only setting JAX version for specific base images)

parambole: @bvandermoon @hengtaoguo 

>  it will result in 0.7.1 eventually after satisfying all dependencies (flax+orbax) and tunix.

If tunix has a hard requirement on a JAX version then it will update the JAX version to it. In the past we have seen issues related to this i.e. it updates the JAX version but not the other bundled LibTPU versions which can cause a problem.

PR #2471, opened 2025-10-08
Adding multi-image support for Llama4 SFT and decode.

NicoGrande:
# Description

This PR extends multi-image support to multi-modal models that leverage image-tiling for their vision encoder (llama4). To do this, the PR makes the following changes:
- Modifies pre-processing functions in `multimodal_utils.py` to output multiple images in a single `PreprocessorOutput` instance.
- Modifies `_input_pipeline_utils.py` to use the new `PreprocessorOutput` structure, and includes new grain functions to reshape data into the corresponding shape expected by the model.
- Modifies `_hf_data_processing.py` to use new grain function for SFT.
- Modifies `decode.py` to leverage new `PreprocessorOutput` structure for decoding.

FIXES: b/449743930 

# Tests

## Llama4:
SFT on Slide-QA: [logs](https://cloudlogging.app.goo.gl/gkLctCAipYN4PnX8A)
Decode: [logs](https://cloudlogging.app.goo.gl/9RzbQUawgeacDuEMA)

## Gemma3:
SFT on Slide-QA: [logs](https://cloudlogging.app.goo.gl/g6nrSH71UL9Utt878)
Decode (tested locally on v4-8):
```
python -m MaxText.decode src/MaxText/configs/base.yml model_name=gemma3-4b tokenizer_path=assets/tokenizer.gemma3 load_parameters_path=gs://maxtext-gemma/unified/gemma3/4b/unscanned/2025-08-09-01-17/0/items per_device_batch_size=1 run_name=ht_test max_prefill_predict_length=550 max_target_length=570 steps=1 async_checkpointing=false scan_layers=false use_multimodal=true prompt=\'Describe\ difference\ between\ these\ images\ \<start_of_image\>\<start_of_image\>\' image_path=\'src/MaxText/test_assets/test_image.jpg,src/MaxText/test_assets/test_image.jpg\' attention=\'dot_product\'

Input `<start_of_turn>user
Describe difference between these images <start_of_image><start_of_image><end_of_turn>
<start_of_turn>model
` -> `Okay, let's break down the differences between the two images.

**Both images show the same`
```

# Checklist


src/MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -47,6 +47,28 @@ def vision_sft_preprocessing_pipeline(
...
     dataset = dataset.shuffle(seed=config.data_shuffle_seed)
 
+  # If multiple image columns are provided, merge them into a single 'images' column.
+  if isinstance(image_column, list):
+    # First, truncate the list of image columns if it exceeds the configured max.
+    if config.max_num_images_per_example > 0 and len(image_column) > config.max_num_images_per_example:
+      image_column = image_column[: config.max_num_images_per_example]
+
+    # Define a mapping function that accepts the list of columns as an argument.
+    def merge_image_columns(example, column_names):
```
aireenmei: Could we make it a general merge_columns function and put in _input_pipeline_utils.py?

src/MaxText/multimodal_utils.py
```diff
@@ -346,81 +336,110 @@ def split_to_tiles(images: np.ndarray, num_tiles_height: int, num_tiles_width: i
...
+      global_tiles = np.array(global_tiles_pil)
+      global_tiles = _normalize_images(
+          global_tiles * LLAMA4_PIXEL_VALUE_RESCALE_FACTOR, mean=LLAMA4_IMAGE_MEAN, std=LLAMA4_IMAGE_STD
+      )
+      global_tiles = np.transpose(global_tiles, (2, 0, 1))
+      global_tiles = np.expand_dims(global_tiles, axis=0)
+      image_tiles = np.concatenate((image_tiles, global_tiles), axis=0)
+
+    # Pad the tiles to the maximum number of tiles
+    image_tiles, image_mask = pad_to_max_tiles(image_tiles, max_num_tiles=20)
```
aireenmei: What's the reason of changing LLAMA4_TILES_NUM to 20 here? There are similar changes in several places, could you help me understand?

src/MaxText/input_pipeline/_input_pipeline_utils.py
```diff
@@ -437,8 +428,10 @@ def _pad_text(self, x, max_length, pad_id):
     pad_amount = [(0, pad_amount)] + [(0, 0)] * (len(x.shape) - 1)
     return np.pad(x, pad_amount, constant_values=pad_id)[: self.max_length]
 
-  def _pad_image_or_mask(self, tensor: np.ndarray) -> np.ndarray:
-    """Pads an input tensor (image or mask) to a maximum number of items.
+  def _pad_image_and_mask(
+      self, preprocessed_image: multimodal_utils.PreprocessorOutput
+  ) -> tuple[np.ndarray, np.ndarray | None]:
```
aireenmei: Should the return type be preprocessed_image?

src/MaxText/multimodal_utils.py
NicoGrande: Of course! The main issue is that we were not padding enough previously for large images when training llama4. The reason if because very large images will be tiled to at most 16 tiles, but then we add an extra global tile making it 17. This means that the maximum pad size of 16 was not enough previously. Setting the max_num_tiles to 17 however is problematic since it is not easily divisible and as such we cannot use sequence parallelism. The next multiple of 2 and 4 I thought could work was 20 so I set it to that.

src/MaxText/input_pipeline/_input_pipeline_utils.py
NicoGrande: Yes, thank you for catching this!

src/MaxText/multimodal_utils.py
aireenmei: Thanks for the clarification! Could we make a constant such as LLAMA4_TILES_PAD_TO with comments to explain and use it in other places?

NicoGrande: Just added this! Thanks for the suggestion.

src/MaxText/input_pipeline/_hf_data_processing.py
NicoGrande: Done!

aireenmei (approved):

hengtaoguo (approved):

PR #2478, opened 2025-10-09
Fix create_nnx_model() to load NNX checkpoint, verified with vllm_decode

SurbhiJainUSC:
# Description

- Update `create_nnx_model()` to load both Linen and NNX checkpoint
- FIXES: b/450116695

# Tests

Tested loading both linen and NNX checkpoint and compare the params before and after updating the model.

Loading linen checkpoint: https://paste.googleplex.com/5840546769797120

Loading NNX checkpoint (without pathways): https://paste.googleplex.com/5153322256433152

Loading NNX checkpoint (with pathways): https://paste.googleplex.com/5967825172824064

vLLM Decoding Results: http://shortn/_QHI6tVbeYP

# Checklist

### DIFF
--- a/src/MaxText/model_creation_utils.py
+++ b/src/MaxText/model_creation_utils.py
@@ -137,33 +137,64 @@ def create_nnx_model(config):
     model = nnx.merge(graphdef, sharded_state)
 
     if config.load_parameters_path:
-      target_for_restore = jax.tree.map(
-          lambda v: v.value,
-          sharded_state,
-          is_leaf=lambda n: isinstance(n, nnx.Variable),
-      )
-
       try:
         ckptr = ocp.Checkpointer(
             ocp.PyTreeCheckpointHandler(
-                restore_concurrent_gb=None,
-                save_concurrent_gb=None,
-                use_ocdbt=True,
-                use_zarr3=True,
+                restore_concurrent_gb=config.checkpoint_storage_concurrent_gb,
+                save_concurrent_gb=config.checkpoint_storage_concurrent_gb,
+                use_ocdbt=config.checkpoint_storage_use_ocdbt,
+                use_zarr3=config.checkpoint_storage_use_zarr3,
             )
         )
+
         # This is a memory optimization. We don't want to restore the entire checkpoint - only the params.
         # Rather than passing the entire abstract state, which could unnecessarily restore opt_state and
         # waste memory, we instead restore the params field of the checkpoint (which itself may be a dictionary
         #  containing a key named 'params').
-        restore_args = ocp.checkpoint_utils.construct_restore_args(target_for_restore)
+
+        # Get the structure of checkpoint in `config.load_parameters_path`
+        metadata = ckptr.metadata(config.load_parameters_path)
+
+        is_nnx_checkpoint = True
+        if (
+            "params" in metadata.item_metadata.tree.keys()
+            and "params" in metadata.item_metadata.tree.get("params", {}).keys()
+        ):
+          # structure of linen checkpoint: {'params': {'params': {'decoder': ...}}}
+          is_nnx_checkpoint = False
+          target_for_restore = jax.tree.map(
+              lambda v: v.value,
+              sharded_state,
+              is_leaf=lambda n: hasattr(n, "value"),
+          )
+
+          item_to_restore = {"params": {"params": target_for_restore}}
+          restore_args = {"params": {"params": ocp.checkpoint_utils.construct_restore_args(target_for_restore)}}
+        else:
+          # structure of nnx checkpoint: {'decoder': {'value': ...}}
+          target_for_restore = jax.tree.map(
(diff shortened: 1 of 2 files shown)


src/MaxText/model_creation_utils.py
```diff
@@ -137,33 +137,64 @@ def create_sharded_state():
...
+          target_for_restore = jax.tree.map(
+              lambda v: v.value,
+              sharded_state,
+              is_leaf=lambda n: isinstance(n, nnx.Variable),
+          )
+
+          item_to_restore = {"params": {"params": target_for_restore}}
+          restore_args = {"params": {"params": ocp.checkpoint_utils.construct_restore_args(target_for_restore)}}
+        else:
+          # structure of nnx checkpoint: {'decoder': {'value': ...}}
```
hengtaoguo, line 174: For your comments:
```
# structure of linen checkpoint: {'params': {'params': {'decoder': ...}}}
# structure of nnx checkpoint: {'decoder': {'value': ...}}
```

Is the `...` part structural identical between linen and nnx? Then for the below variable `checkpoint`, what's the tree structure of it?

SurbhiJainUSC: Yes, `...` is the identical part. Here is what `checkpoint` variable would look like for both Linen and NNX: https://paste.googleplex.com/4702000029761536

src/MaxText/vllm_decode.py
```diff
@@ -15,12 +15,12 @@
...
 Example command:
-  python3 -m MaxText.vllm_decode MaxText/configs/sft.yml \
+  python3 -m MaxText.vllm_decode MaxText/configs/base.yml \
     model_name=llama3.1-8b tokenizer_path=meta-llama/Llama-3.1-8B-Instruct \
     tokenizer_type=huggingface hf_access_token=<your_hf_token> \
     load_parameters_path=<your_checkpoint_path> \
-    per_device_batch_size=1 run_name=vllm_decode_test \
-    use_chat_template=True prompt="Suggest some famous landmarks in London." \
+    per_device_batch_size=1 run_name=vllm_decode_test max_target_length=64 \
+    use_chat_template=False prompt="Suggest some famous landmarks in London." \
```
A9isha, line 23: what is the issue if we use `sft.yml`?

src/MaxText/model_creation_utils.py
```diff
@@ -137,33 +137,64 @@ def create_sharded_state():
...
         # waste memory, we instead restore the params field of the checkpoint (which itself may be a dictionary
         #  containing a key named 'params').
-        restore_args = ocp.checkpoint_utils.construct_restore_args(target_for_restore)
+
+        # Get the structure of checkpoint in `config.load_parameters_path`
+        metadata = ckptr.metadata(config.load_parameters_path)
+
+        is_nnx_checkpoint = True
+        if (
+            "params" in metadata.item_metadata.tree.keys()
```
A9isha, line 160: I think the better way to decide if it is an NNX checkpoint is to check for nnx.Variable in the leaf.

src/MaxText/model_creation_utils.py
```diff
@@ -137,33 +137,64 @@ def create_sharded_state():
...
+            and "params" in metadata.item_metadata.tree.get("params", {}).keys()
+        ):
+          # structure of linen checkpoint: {'params': {'params': {'decoder': ...}}}
+          is_nnx_checkpoint = False
+          target_for_restore = jax.tree.map(
+              lambda v: v.value,
+              sharded_state,
+              is_leaf=lambda n: isinstance(n, nnx.Variable),
```
A9isha: if it is not an NNX checkpoint i.e., `is_nnx_checkpoint = False`, why do we have `nnx.Variable` check?

src/MaxText/vllm_decode.py
SurbhiJainUSC: This `vllm_decode` script is not specifically for SFT. It can be used for decoding any NNX MaxText model. So, it doesn't make sense to use `sft.yml` configuration. Hence, I changed it to `base.yml`.

src/MaxText/model_creation_utils.py
SurbhiJainUSC: This will check the params structure in the checkpoint present in GCS. Linen checkpoint has nested params structure but NNX checkpoint doesn't. So, this checks the structure of checkpoint in GCS before any loading begins.

src/MaxText/model_creation_utils.py
SurbhiJainUSC: Updated the check.

src/MaxText/model_creation_utils.py
A9isha: Right, my comment was more to say that instead of relying on the existence of "params" in the keys, a better way to reliably infer that it is an NNX checkpoint is to check for the existence of `nnx.variable`.

But, it is okay for now.

PR #2479, opened 2025-10-09
init KVCache fix for grpo trainer

khatwanimohit:
# Description

originally kvcache was initialized using `config.global_batch_size_to_train_on` and model was initialized using `max_utils.get_batch_seq_len_for_mode`. These ways assumed that the whole mesh is going to be used for model initialization. 

In grpo_trainer with disaggregated setup, we initialize the training and inference model with a submesh. This PR uses mesh.size to calculate global batch size used for initializing the model. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/src/MaxText/maxtext_utils.py
+++ b/src/MaxText/maxtext_utils.py
@@ -1060,7 +1060,7 @@ def get_prefill_kv_cache_annotations(model, config, rng, mesh, page_state: None 
 
   def init_kv_cache(model, config):
     input_shape = (
-        config.micro_batch_size_to_train_on,
+        int(config.per_device_batch_size * mesh.size),
         config.max_prefill_predict_length,
     )
     image_shape = multimodal_utils.get_dummy_image_shape_for_init(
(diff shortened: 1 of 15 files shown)


khatwanimohit:
> Thanks Mohit. For my understanding, can you please share the behavior you saw here? Was it throwing an exception/not being partitioned properly?

It was throwing an error at [attention_op.py](https://github.com/AI-Hypercomputer/maxtext/blob/19b9f7f86e55b09662220eb0d89567ae14476dc0/src/MaxText/layers/attention_op.py#L530) complaining the batch size of query and key, value don't match. 
That error helped me find the reason because we were initializing key, values batch size based on jax.devices()

src/MaxText/maxtext_utils.py
```diff
@@ -1060,7 +1060,7 @@ def get_prefill_kv_cache_annotations(model, config, rng, mesh, page_state: None
 
   def init_kv_cache(model, config):
     input_shape = (
-        config.micro_batch_size_to_train_on,
+        int(config.per_device_batch_size * mesh.size),
```
bvandermoon, line 1063: @gobbleturk is this the right value to use here?

khatwanimohit:
Thank you @gobbleturk for suggesting to change in https://github.com/AI-Hypercomputer/maxtext/blob/07205e24f770501214735cf80ff6358d6c831915/src/MaxText/pyconfig.py#L1166 instead which made me realize we already support initializing a model on a submesh using config.subslice_shape. Closing this PR

PR #2480, opened 2025-10-10
[WIP]Refactor resource management

mydatascience:
# Description

Better handling of resources namely TPU so there is one place to init it and gracefully release 

# Checklist

### DIFF
--- a/src/MaxText/max_utils.py
+++ b/src/MaxText/max_utils.py
@@ -20,26 +20,19 @@ from collections.abc import Sequence
 import functools
 from functools import partial
 import os
-import socket
 import subprocess
-import time
 from typing import Any
 
-from etils import epath
 import flax
 import jax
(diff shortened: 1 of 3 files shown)


shauryagup:
Can you clarify the motivation for this a bit more? Is there a particular multi-tenancy setup that you are running that needs this resource management? My initial thought is that this resource management should be done by a higher layer whichever is scheduling workloads that needs to share resources like GKE etc.

Also, to note is that this logic seems highly coupled with multi controller JAX and would likely not be needed with pathways (there is no jax.distributed.initialize() needed for pathways)

mydatascience:
The logic here is to have some resource management. Right now jax.distributed.initialize is called but jax.distributed.shutdown is not. Also jax.distributed.initialize is inside some functions in maxtext. For a colab it might result in non-releasing TPU even when the job is done

PR #2486, opened 2025-10-10
fix xlml logits test in gemma, llama4

aireenmei:
# Description

b/445750145
Changes in forward_pass_logit_checker: Previously, this script test either KL divergence when max_kl_divergence is provided, or otherwise numerical similarity (atol, rtol), until my [PR](https://github.com/AI-Hypercomputer/maxtext/pull/2203) made it test atol all the time with the default threshold, causing new logit failures that's not a regression (https://b.corp.google.com/issues/445750145#comment10). The later added new tests use both max_kl_divergence and atol, and we should allow this flexibility. With this PR, user needs to explicitly provide atol to run atol test. I inspected all the current scripts that use forward_pass_logit_checker, and only a few (llama2 and llama3) have neither max_kl_divergence nor atol, thus relies on the default atol, rtol value for testing. I added "atol=0.1 rtol=0.1" explicitly to these tests, so the behavior should be consistent.

# Tests

Llama4 logits test pass: https://cloudlogging.app.goo.gl/FA6biJx9rg89dbh47
other test should pass because the threshold is matching the current test. I'll monitor xlml tests after PR merged.

# Checklist


tests/forward_pass_logit_checker.py
```diff
@@ -351,6 +340,17 @@ def main(config, test_args):  # pylint: disable=W0621
             f"KL divergence values exceed the specified threshold of {test_args.max_kl_div}. "
             f"Max divergence: {jax.numpy.max(kl_div)}"
         )
+      else:
```
shuningjin: Is it possible to test for both when having flag `--atol --rtol --kl_div`? Perhaps change this `else` to `if test_args.atol is not None` , and change the default `atol` value to None. Thank you!

aireenmei: Thanks for the suggestion! I revised.

tests/forward_pass_logit_checker.py
```diff
@@ -451,8 +452,8 @@ def main(config, test_args):  # pylint: disable=W0621
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
 
   parser = argparse.ArgumentParser()
-  parser.add_argument("--atol", type=float, required=False, default=0.1)
-  parser.add_argument("--rtol", type=float, required=False, default=0.1)
+  parser.add_argument("--atol", type=float, required=False, default=None)
```
shuningjin, line 455: Could you also add a check to prevent empty test, when both max_kl_div and atol is none?

aireenmei: Good suggestion! Added

tests/forward_pass_logit_checker.py
```diff
@@ -329,16 +329,17 @@ def main(config, test_args):  # pylint: disable=W0621
...
-      max_logging.log(
-          f"Checking Numerical Differences between train logits and golden logits against "
-          f"atol={test_args.rtol} rtol={test_args.atol}."
-      )
-      rtol_val = float(test_args.rtol)
-      atol_val = float(test_args.atol)
-      assert jax.numpy.allclose(
-          train_logits_slice, golden_logits_slice, rtol=rtol_val, atol=atol_val, equal_nan=False
-      ), f"Logits do not match closely enough. Required rtol={test_args.rtol}, atol={test_args.atol}."
+      if test_args.atol is not None:
```
RissyRan, line 332: Shall we also check rtol is not None?

tests/forward_pass_logit_checker.py
```diff
@@ -451,8 +452,8 @@ def main(config, test_args):  # pylint: disable=W0621
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
 
   parser = argparse.ArgumentParser()
-  parser.add_argument("--atol", type=float, required=False, default=0.1)
-  parser.add_argument("--rtol", type=float, required=False, default=0.1)
+  parser.add_argument("--atol", type=float, required=False, default=None)
+  parser.add_argument("--rtol", type=float, required=False, default=0)  # only test on atol when rtol=0
```
RissyRan: Is this change on purpose?

aireenmei: Discussed offline, I'll have the default value set to 1e-05, which aligns with the default used for jnp.allclose

PR #2487, opened 2025-10-10
Add support for Qwen3-Next Model Architecture

parambole:
### TL;DR

-   **What**: This PR integrates the **Qwen3-Next** architecture into MaxText. Key features include a hybrid attention mechanism combining a **Gated Delta Net** (a form of linear attention) with standard attention, and an ultra-sparse MoE block that includes a **shared expert**.

-   **How**: By implementing several new, highly-optimized JAX layers (`Qwen3NextGatedDeltaNet`, `Qwen3NextSparseMoeBlock`, `Qwen3NextRMSNorm`), creating a hybrid `Qwen3NextDecoderLayer` that alternates between attention types based on the `full_attention_interval`, and validating each new component against a PyTorch reference in a robust new test suite.

* * * * *

### Detailed Description

This pull request introduces an implementation of the Qwen3-Next architecture, as described in the [official Qwen3-Next blog post](https://qwen.ai/blog?id=4074cca80393150c248e508aa62983f9cb7d27cd&from=research.latest-advancements-list).

#### **Architectural Implementation (`src/MaxText/layers/qwen3.py`)**

-   **Hybrid Attention**: The core of this architecture is its hybrid attention system. This PR implements:

    -   **`Qwen3NextGatedDeltaNet`**: A full JAX implementation of the Gated Delta Net for efficient linear attention. The core recurrence is handled by the pure JAX `jax_chunk_gated_delta_rule`, a parallel scan algorithm.

    -   **`Qwen3NextFullAttention`**: A **placeholder** for the model's standard attention layers.

-   **Custom Normalization**: A custom **`Qwen3NextRMSNorm`** layer is included where the learnable weight is initialized to zeros and applied as `(1.0 + weight)`, matching the reference implementation's specific behavior.

* * * * *

#### **Decoder and Model Integration**

-   **`Qwen3NextDecoderLayer`**: This new decoder layer acts as the hybrid controller. Based on its `layer_idx` and the `full_attention_interval` config parameter, it conditionally invokes either the `Qwen3NextGatedDeltaNet` or the `Qwen3NextFullAttention` layer before passing the result to the `Qwen3NextSparseMoeBlock`.

-   **Scanned Layers**: The implementation also introduces a **`Qwen3NextScannableBlock`**.

-   **Configuration**: A new model config `qwen3-next-80b-a3b.yml` is added, along with the necessary parameters in `base.yml` to control the new architectural features.

* * * * *

#### **Testing and Validation**

This PR includes a new test file, `tests/check_qwen3_next_vs_reference.py`.

-   It directly copies the reference PyTorch implementations for each new component.

-   It provides granular unit tests that validate the numerical output of each JAX module (`Qwen3NextGatedDeltaNet`, `Qwen3NextSparseMoeBlock`, etc.) against its corresponding PyTorch reference.

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -860,3 +860,19 @@ subslice_shape: ""
 # NNX
 enable_nnx: false
 shard_optimizer_over_data: False
+
+################################## Qwen3-Next Specific Configs ##################################
+# Kernel size for the 1D convolution in the Gated Delta Net
+gdn_conv_kernel_dim: 4
+# Head dimension for the key/query in the Gated Delta Net
+gdn_key_head_dim: 128
+# Head dimension for the value in the Gated Delta Net
+gdn_value_head_dim: 128
+# Number of key/query heads in the Gated Delta Net
+gdn_num_key_heads: 16
+# Number of value heads in the Gated Delta Net
+gdn_num_value_heads: 32
+# Chunk size for the parallel scan algorithm in the Gated Delta Net.
+gdn_chunk_size: 64
+# Whether to apply L2 normalization to query and key tensors inside the Gated Delta Rule kernel.
+use_qk_norm_in_gdn: True
--- a/src/MaxText/configs/models/qwen3-next-80b-a3b.yml
+++ b/src/MaxText/configs/models/qwen3-next-80b-a3b.yml
@@ -1,0 +1,47 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# MaxText/configs/models/qwen3-next-80b-a3b.yml
+
+# Set the decoder block to our new implementation
+decoder_block: "qwen3_next"
+
+# Core Architectural Parameters
+base_emb_dim: 2048
+base_num_decoder_layers: 48
+base_num_query_heads: 16
+base_num_kv_heads: 2
+head_dim: 256
+vocab_size: 151936
+normalization_layer_epsilon: 1.0e-6
+
+# MoE Specific Parameters
+# Set base_mlp_dim to match base_moe_mlp_dim to pass validation for fully MoE models.
+base_mlp_dim: 512
+base_moe_mlp_dim: 512
+num_experts: 512
+num_experts_per_tok: 10
+norm_topk_prob: True
+
+# Qwen3-Next Specific Parameters for Linear Attention (Gated Delta Net)
+inhomogeneous_layer_cycle_interval: 4
+gdn_conv_kernel_dim: 4
+gdn_key_head_dim: 128
+gdn_value_head_dim: 128
+gdn_num_key_heads: 16
+gdn_num_value_heads: 32
+gdn_chunk_size: 64
+
+# RoPE Settings
+rope_max_timescale: 10000000
--- a/src/MaxText/layers/decoders.py
+++ b/src/MaxText/layers/decoders.py
@@ -401,6 +401,8 @@ class Decoder(nn.Module):
         return [qwen3.Qwen3DecoderLayerToLinen]
       case DecoderBlockType.QWEN3_MOE:
         return [qwen3.Qwen3MoeDecoderLayerToLinen]
+      case DecoderBlockType.QWEN3_NEXT:
+        return [qwen3.Qwen3NextScannableBlockToLinen] if self.config.scan_layers else [qwen3.Qwen3NextDecoderLayerToLinen]
       case DecoderBlockType.SIMPLE:
         return [simple_layer.SimpleDecoderLayerToLinen]
       case DecoderBlockType.SIMPLE_MLP:
@@ -452,6 +454,7 @@ class Decoder(nn.Module):
         DecoderBlockType.GEMMA3,
         DecoderBlockType.QWEN3,
         DecoderBlockType.QWEN3_MOE,
+        DecoderBlockType.QWEN3_NEXT,
         DecoderBlockType.GPT_OSS,
         DecoderBlockType.SIMPLE,
         DecoderBlockType.SIMPLE_MLP,
@@ -824,6 +827,8 @@ class Decoder(nn.Module):
                   "is_nope_layer": llama4.determine_is_nope_layer(lyr, self.config.nope_layer_interval),
                   "is_moe_layer": llama4.determine_is_moe_layer(lyr, self.config.interleave_moe_layer_step),
               }
+            if cfg.decoder_block == DecoderBlockType.QWEN3_NEXT:
+              layer_kwargs = {"layer_idx": lyr}
             if cfg.decoder_block == DecoderBlockType.GPT_OSS:
               layer_kwargs = {"attention_type": gpt_oss.get_attention_type(layer_id=lyr)}
             layer = RemattedBlockLayer(
--- a/src/MaxText/layers/qwen3.py
+++ b/src/MaxText/layers/qwen3.py
@@ -16,24 +16,905 @@
 # pylint: disable=arguments-differ
 # pylint: disable=no-name-in-module
 
+from typing import cast
+
+import jax
+import jax.nn
 from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
 import jax.numpy as jnp
 
 from flax import linen as nn
 from flax import nnx
+from flax.linen import initializers as linen_initializers
 
 from MaxText import max_utils
-from MaxText.common_types import Config
-from MaxText.layers import initializers
+from MaxText.common_types import Config, DType, Array
+from MaxText.layers import attentions
+from MaxText.layers import initializers as max_initializers
+from MaxText.layers import linears
+from MaxText.layers import moe
 from MaxText.layers import nnx_wrappers
 from MaxText.layers import quantizations
+from MaxText.layers.normalizations import RMSNorm, l2norm
+from MaxText.layers.quantizations import AqtQuantization as Quant
+from MaxText.inference import page_manager
 from MaxText.layers.attentions import Attention
 from MaxText.layers.linears import MlpBlock
 from MaxText.layers.moe import RoutedMoE
-from MaxText.layers.normalizations import RMSNorm
-from MaxText.layers.quantizations import AqtQuantization as Quant
-from MaxText.inference import page_manager
+
+
+# -----------------------------------------
+# Qwen3-Next Layer Implementations
+# -----------------------------------------
+
+
+def jax_chunk_gated_delta_rule(
+    query: Array,
+    key: Array,
+    value: Array,
+    g: Array,
+    beta: Array,
+    chunk_size: int = 64,
+    initial_state: None | Array = None,
+    use_qk_norm_in_gdn: bool = False,
+) -> tuple[Array, None | Array]:
+  """
+  A JAX implementation of the chunked Gated Delta Rule, a parallel scan algorithm.
+  This function implements the core recurrent logic of the Gated Delta Network in
+  a hardware-efficient way by splitting the sequence into chunks and using
+  jax.lax.scan for the recurrent part.
+
+  Tensor Shape Abbreviations:
+    B: batch_size, S: sequence_length, H: num_heads,
+    D_k: key/query_head_dim, D_v: value_head_dim,
+    N: num_chunks, C: chunk_size
+
+  Args:
+    query: Query tensor. Shape (B, S, H, D_k)
+    key: Key tensor. Shape (B, S, H, D_k)
+    value: Value tensor. Shape (B, S, H, D_v)
+    g: Log decay tensor. Shape (B, S, H)
+    beta: Gate tensor. Shape (B, S, H)
+    chunk_size: The size of each chunk for processing.
+    initial_state: Optional initial state for the recurrence. Shape (B, H, D_k, D_v)
+    use_qk_norm_in_gdn: Whether to apply L2 normalization to query and key.
+
+  Returns:
+    Output tensor. Shape (B, S, H, D_v)
+    Final recurrent state. Shape (B, H, D_k, D_v) or None
+  """
+
+  # =========================================================================
+  # STAGE 1: PREPARATION & PADDING
+  # =========================================================================
+  initial_dtype = query.dtype
+  if use_qk_norm_in_gdn:
+    query = l2norm(query, dim=-1, eps=1e-6)
+    key = l2norm(key, dim=-1, eps=1e-6)
+
+  # Transpose (B, S, H, D) -> (B, H, S, D)
+  query = jnp.transpose(query, (0, 2, 1, 3)).astype(jnp.float32)
+  key = jnp.transpose(key, (0, 2, 1, 3)).astype(jnp.float32)
+  value = jnp.transpose(value, (0, 2, 1, 3)).astype(jnp.float32)
+  # Transpose (B, S, H) -> (B, H, S)
+  beta = jnp.transpose(beta, (0, 2, 1)).astype(jnp.float32)
+  g = jnp.transpose(g, (0, 2, 1)).astype(jnp.float32)
+
+  batch_size, num_heads, sequence_length, k_head_dim = key.shape
+  v_head_dim = value.shape[-1]
+  pad_size = (chunk_size - sequence_length % chunk_size) % chunk_size
+
+  # Padding to make sequence_length divisible by chunk_size
+  if pad_size > 0:
+    query = jnp.pad(query, ((0, 0), (0, 0), (0, pad_size), (0, 0)))  # (B, H, S_padded, D_k)
+    key = jnp.pad(key, ((0, 0), (0, 0), (0, pad_size), (0, 0)))  # (B, H, S_padded, D_k)
+    value = jnp.pad(value, ((0, 0), (0, 0), (0, pad_size), (0, 0)))  # (B, H, S_padded, D_v)
+    beta = jnp.pad(beta, ((0, 0), (0, 0), (0, pad_size)))  # (B, H, S_padded)
+    g = jnp.pad(g, ((0, 0), (0, 0), (0, pad_size)))  # (B, H, S_padded)
+
+  total_sequence_length = sequence_length + pad_size
+  # query shape: (B, H, S_padded, D_k)
+  scale = jax.lax.rsqrt(jnp.array(query.shape[-1]).astype(jnp.float32))
+  query = query * scale
+
+  v_beta = value * jnp.expand_dims(beta, -1)  # (B, H, S_padded, D_v)
+  k_beta = key * jnp.expand_dims(beta, -1)  # (B, H, S_padded, D_k)
+
+  # Reshape to chunks
+  num_chunks = total_sequence_length // chunk_size
+  # query_c shape: (B, H, N, C, D_k)
+  query_c = query.reshape(batch_size, num_heads, num_chunks, chunk_size, k_head_dim)
+  key_c = key.reshape(batch_size, num_heads, num_chunks, chunk_size, k_head_dim)
+  k_beta_c = k_beta.reshape(batch_size, num_heads, num_chunks, chunk_size, k_head_dim)
+  v_beta_c = v_beta.reshape(batch_size, num_heads, num_chunks, chunk_size, v_head_dim)
+  g_c = g.reshape(batch_size, num_heads, num_chunks, chunk_size)  # (B, H, N, C)
+
+  mask = jnp.triu(jnp.ones((chunk_size, chunk_size), dtype=bool), k=0)  # (C, C)
+
+  # =========================================================================
+  # STAGE 2: INTRA-CHUNK CALCULATION (PARALLEL)
+  # =========================================================================
+  # g_cumsum shape: (B, H, N, C)
+  g_cumsum = jnp.cumsum(g_c, axis=-1)
+  # g_diff shape: (B, H, N, C, C)
+  g_diff = jnp.expand_dims(g_cumsum, -1) - jnp.expand_dims(g_cumsum, -2)
+
+  # Apply tril to zero out the upper triangle of g_diff. This is crucial because
+  # the upper triangle contains large positive values that would cause exp() to overflow.
+  g_diff_tril = jnp.tril(g_diff)
+
+  # Exponentiate the lower triangular g_diff. Since these values are non-positive,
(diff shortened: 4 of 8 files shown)


src/MaxText/layers/qwen3.py
```diff
@@ -113,6 +115,818 @@ def self_attention_with_norm(
 
   return hidden_states, residual_after_attention
 
+# -----------------------------------------
+# Qwen3-Next Layer Implementations
+# -----------------------------------------
+
+class Qwen3NextRMSNorm(nn.Module):
```
bvandermoon: Can you make these NNX modules instead?

parambole: Hey @bvandermoon - Yes I will do it during my main branch rebase which is a merge conflict as main has qwen nnx changes.

parambole: @bvandermoon - Done

src/MaxText/layers/decoders.py
```diff
@@ -401,6 +401,8 @@ def get_decoder_layers(self):
         return [qwen3.Qwen3DecoderLayerToLinen]
       case DecoderBlockType.QWEN3_MOE:
         return [qwen3.Qwen3MoeDecoderLayerToLinen]
+      case DecoderBlockType.QWEN3_NEXT:
+        return [qwen3.Qwen3NextDecoderLayerLinen]
```
gobbleturk: I see the previous are named LayerToLinen vs this is just LayerLinen, is this intentional?

src/MaxText/layers/decoders.py
```diff
@@ -924,3 +939,50 @@ def _apply_gemma3_scanned_blocks(
...
+        deterministic,
+        model_mode,
+        previous_chunk,
+        page_state,
+        slot,
+    )
+
+    RemattedBlockLayer = self.set_remat_policy([qwen3.Qwen3NextScannableBlockLinen], policy)[0]
+
+    scanned_block = self.scan_decoder_layers(
```
gobbleturk: An alternative implementation is to use jax.lax.cond and scan every layer - @gagika implemented this for gemma

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+from MaxText.layers.linears import MlpBlock
+from MaxText.layers.moe import RoutedMoE
+
+
+# -----------------------------------------
+# Qwen3-Next Layer Implementations
+# -----------------------------------------
+
+
+def l2norm(x: Array, dim: int = -1, eps: float = 1e-6) -> Array:
```
gobbleturk: Is there an existing jax function we could use for this instead?

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+  g_diff_exp = jnp.exp(g_diff_tril).astype(jnp.float32)
+
+  # The result g_diff_exp is already lower triangular and serves as the decay_mask.
+  # decay_mask shape: (B, H, N, C, C)
+  decay_mask = g_diff_exp
+
+  # --- Precompute within-chunk attention ---
+  prec = jax.lax.Precision.HIGHEST
+  # attn shape: (B, H, N, C, C)
+  attn = -jnp.matmul(k_beta_c, jnp.swapaxes(key_c, -1, -2), precision=prec) * decay_mask
```
gobbleturk, line 159: should the precision here be user configurable? I believe precision only has an effect if the inputs are f32

tests/check_qwen3_next_vs_reference.py
```diff
@@ -0,0 +1,669 @@
...
+        super().__init__()
+        self.num_experts = config.num_experts
+        for _ in range(config.num_experts):
+            self.append(Qwen3NextMLP_PT(config, intermediate_size=config.moe_intermediate_size))
+
+    def forward(self, hidden_states: torch.Tensor, top_k_index: torch.Tensor, top_k_weights: torch.Tensor) -> torch.Tensor:
+        final_hidden_states = torch.zeros_like(hidden_states)
+        expert_mask = torch.nn.functional.one_hot(top_k_index, num_classes=self.num_experts).permute(2, 1, 0)
+        expert_hit = torch.greater(expert_mask.sum(dim=(-1, -2)), 0).nonzero()
+        for expert_idx in expert_hit:
```
gobbleturk: Correctness tests are hard. Ideally tests are "obviously correct" e.g. they don't have any logic and are easy to read but I understand the need here

src/MaxText/configs/base.yml
```diff
@@ -859,3 +859,21 @@ subslice_shape: ""
 
 # NNX
 enable_nnx: false
+
+################################## Qwen3-Next Specific Configs ##################################
+# The interval for using full attention. e.g., 4 means every 4th layer is full attention.
+full_attention_interval: 4
```
shuningjin: nit: maybe reuse exisiting `inhomogeneous_layer_cycle_interval` instead? we use it for llama4 and gpt-oss

https://github.com/AI-Hypercomputer/maxtext/blob/5ca855e27f93185a850e61b4a49ca2ba55ba9f55/src/MaxText/configs/base.yml#L206-L209

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+from MaxText.layers import attentions
+from MaxText.layers import initializers as max_initializers
+from MaxText.layers import linears
+from MaxText.layers import moe
 from MaxText.layers import nnx_wrappers
 from MaxText.layers import quantizations
-from MaxText.layers.attentions import Attention
-from MaxText.layers.linears import MlpBlock
-from MaxText.layers.moe import RoutedMoE
+from MaxText.layers.normalizations import rms_norm
```
shuningjin: nit: can remove this

src/MaxText/layers/qwen3.py
shuningjin: maybe use this config: https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/layers/moe.py#L578

src/MaxText/configs/base.yml
RissyRan: Wondering if we need this config specifically. Currently, both [Gemma3](https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/layers/gemma3.py#L37 and [GPT-OSS](https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/layers/gpt_oss.py#L45) have this mixed attn partterns.

tests/check_qwen3_next_vs_reference.py
```diff
@@ -0,0 +1,669 @@
...
+from flax import nnx
+
+from MaxText import pyconfig
+from MaxText.layers import qwen3, linears, moe
+from MaxText.layers import initializers as max_initializers
+from MaxText import maxtext_utils
+from MaxText.common_types import Config
+from MaxText.globals import MAXTEXT_PKG_DIR
+
+# ----------------------------------------------------------------------
```
RissyRan, line 36: Could you add a comment for the source?

src/MaxText/layers/qwen3.py
```diff
@@ -253,10 +1144,45 @@ def __call__(
...
+    Qwen3NextRMSNormGated,
+    base_metadata_fn=max_initializers.variable_to_logically_partitioned,
+)
+
+Qwen3NextRMSNormLinen = nnx_wrappers.to_linen_class(
+    Qwen3NextRMSNorm,
+    base_metadata_fn=max_initializers.variable_to_logically_partitioned,
+)
+
+Qwen3NextDecoderLayerLinen = nnx_wrappers.to_linen_class(
```
shuningjin: nit: I see only `Qwen3NextDecoderLayerLinen` and `Qwen3NextScannableBlockLinen` are used. Could you remove the other, and rename to `Qwen3NextDecoderLayerToLinen` to be consistent?

tests/check_qwen3_next_vs_reference.py
RissyRan: Could you explain a little bit more, Matt? Those are reference implementation in PT (copied from source). We will run MaxText against the reference implementation and compare outputs.

src/MaxText/layers/decoders.py
shuningjin: Why is it not returning scannable block as in llama4 and gpt-oss? I see how you scan qwen3next is also different later on. (I think qwen3next is using inhomogenous attention pattern, and the scan should be very similar to these two?) @RissyRan 

https://github.com/AI-Hypercomputer/maxtext/blob/5ca855e27f93185a850e61b4a49ca2ba55ba9f55/src/MaxText/layers/decoders.py#L398-L411

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+      model_mode: str,
+      previous_chunk=None,
+      page_state: None | page_manager.PageState = None,
+      slot: None | int = None,
+  ):
+    residual = inputs
+
+    # First LayerNorm, applied before the attention block.
+    hidden_states = self.input_layernorm(inputs)
+    hidden_states = nn.with_logical_constraint(hidden_states, ("activation_batch", "activation_length", "activation_embed"))
```
shuningjin: Please use `activation_norm_length`. Could you set `self.activation_axis_names = ("activation_batch", "activation_norm_length", "activation_embed")` in init, and use `nn.with_logical_constraint(self.activation_axis_names)` for all constraint in this call. This is what we use for other decoder layers.

src/MaxText/configs/base.yml
```diff
@@ -859,3 +859,21 @@ subslice_shape: ""
...
+# Head dimension for the value in the Gated Delta Net
+linear_value_head_dim: 128
+# Number of key/query heads in the Gated Delta Net
+linear_num_key_heads: 16
+# Number of value heads in the Gated Delta Net
+linear_num_value_heads: 32
+# Chunk size for the parallel scan algorithm in the Gated Delta Net.
+linear_chunk_size: 64
+# Whether to apply L2 normalization to query and key tensors inside the Gated Delta Rule kernel.
+use_qk_l2norm_in_kernel: True
```
RissyRan: Do you think we could reuse use_qk_norm or rename it to use_qk_norm_in_gdn? I know HF has those config names, wondering inside of MaxText, shall we rename them to something like Gated Delta Net specific? 

linear_conv_kernel_dim --> gdn_conv_kernel_dim (no strong opinion on this, but `linear` in MaxText seems unclear to me.

src/MaxText/configs/models/qwen3-next-80b-a3b.yml
```diff
@@ -0,0 +1,52 @@
...
+linear_conv_kernel_dim: 4
+linear_key_head_dim: 128
+linear_value_head_dim: 128
+linear_num_key_heads: 16
+linear_num_value_heads: 32
+
+# RoPE Settings
+rope_max_timescale: 10000000
+
+# Use the dense_matmul path that we verified in our unit tests.
```
RissyRan: This is not model specific, let's add assertion in pyconfig.py file and only allow to use sparse_mamtul for now.

src/MaxText/configs/models/qwen3-next-80b-a3b.yml
```diff
@@ -0,0 +1,52 @@
...
+sparse_matmul: False
+
+# General Model Settings
+enable_dropout: False
```
RissyRan: Let's remove it, not model specific

src/MaxText/layers/qwen3.py
RissyRan: Also, let's move to https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/layers/normalizations.py#L4

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+    eps: Small epsilon to prevent division by zero.
+
+  Returns:
+    L2 normalized array with the same shape as x.
+  """
+  inv_norm = jax.lax.rsqrt((x * x).sum(axis=dim, keepdims=True) + jnp.array(eps, dtype=x.dtype))
+  return x * inv_norm
+
+
+def jax_chunk_gated_delta_rule(
```
RissyRan, line 52: Shall we actually move this gated delta rule to attention.py? or even a separate file like attention_mla.py?

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+  # Exponentiate the lower triangular g_diff. Since these values are non-positive,
+  # exp() will not overflow and will produce values between 0 and 1.
+  g_diff_exp = jnp.exp(g_diff_tril).astype(jnp.float32)
+
+  # The result g_diff_exp is already lower triangular and serves as the decay_mask.
+  # decay_mask shape: (B, H, N, C, C)
+  decay_mask = g_diff_exp
+
+  # --- Precompute within-chunk attention ---
+  prec = jax.lax.Precision.HIGHEST
```
RissyRan, line 157: Is this can be controlled by https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/configs/base.yml#L103

tests/check_qwen3_next_vs_reference.py
```diff
@@ -0,0 +1,669 @@
...
+# START: Copied PyTorch functions
```
shuningjin, line 37: could you add the source link, maybe this: https://github.com/huggingface/transformers/blob/a9731a725eb1d7b3b7e11f0ad35a819fa4ee8b20/src/transformers/models/qwen3_next/modeling_qwen3_next.py?

nit: I feel it might be better for keep the function/class names the same as the original code.

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+  # Trim padding: (B, H, S, D_v)
+  core_attn_out = core_attn_out[:, :, :sequence_length, :]
+
+  # Transpose back to (B, S, H, D_v)
+  core_attn_out = jnp.transpose(core_attn_out, (0, 2, 1, 3)).astype(initial_dtype)
+
+  return core_attn_out, final_state if output_final_state else None
+
+
+class Qwen3NextRMSNorm(nnx.Module):
```
RissyRan, line 285: Could we double check and reuse if possbile? https://github.com/AI-Hypercomputer/maxtext/blob/f1fcf8c771f77421301739bdfea0cdc84e32362d/src/MaxText/layers/normalizations.py#L30 

If not, let's move to normalization layer

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+    x_dtype = x.dtype
+    x = x.astype(jnp.float32)
+    variance = jnp.mean(jnp.square(x), axis=-1, keepdims=True)
+    x = x * jax.lax.rsqrt(variance + self.eps)
+    # Add 1.0 to the learnable weight
+    x = x * (1.0 + weight.astype(jnp.float32))
+    return x.astype(x_dtype)
+
+
+class Qwen3NextRMSNormGated(nnx.Module):
```
RissyRan, line 321: Similar comment as above.

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+    initial_state: Optional initial state for the recurrence. Shape (B, H, D_k, D_v)
+    use_qk_l2norm_in_kernel: Whether to apply L2 normalization to query and key.
+
+  Returns:
+    Output tensor. Shape (B, S, H, D_v)
+    Final recurrent state. Shape (B, H, D_k, D_v) or None
+  """
+
+  # =========================================================================
+  # STAGE 1: PREPARATION & PADDING
```
RissyRan, line 89: Thank you for great comments!

Let's leverage Gemini, and break them down to smaller helper functions for better readability?

RissyRan (commented):
Great work!

src/MaxText/layers/qwen3.py
```diff
@@ -16,24 +16,915 @@
...
+    weight = self.weight.value
+
+    # RMS Normalization logic
+    hidden_states_f32 = hidden_states.astype(jnp.float32)
+    variance = jnp.mean(jnp.square(hidden_states_f32), axis=-1, keepdims=True)
+    normalized_states = hidden_states_f32 * jax.lax.rsqrt(variance + self.eps)
+    normalized_states = normalized_states * weight.astype(jnp.float32)
+
+    # Gated Activation using SiLU (Sigmoid-weighted Linear Unit)
+    gated_states = normalized_states * nn.silu(gate.astype(jnp.float32))
```
shuningjin: shall we migrate the linen activation functions as well? for `nn.silu`, use `jax.nn.silu` or `nnx.silu`?  (and `nn.sigmoid`, `nn.softplus` in the below) @bvandermoon

src/MaxText/layers/decoders.py
```diff
@@ -924,3 +939,50 @@ def _apply_gemma3_scanned_blocks(
...
+      decoder_positions,
+      deterministic,
+      model_mode,
+      previous_chunk,
+      page_state,
+      slot,
+  ):
+    """Applies Qwen3-Next scanned decoder blocks."""
+    cfg = self.config
+    mesh = self.mesh
```
bvandermoon: Is it possible to move this function to `qwen3.py` since it is model-specific? I realize already have this here for `gemma3`, but it would be better to keep model-specific pieces out of this file

bvandermoon (commented):
Generally looking good to me, thanks @parambole

src/MaxText/layers/decoders.py
parambole: Thank you for catching that. Done!

src/MaxText/layers/qwen3.py
parambole: > Is there an existing jax function we could use for this instead?

I wasn't able to find an existing JAX function for this. 

> Also, let's move to

Done

src/MaxText/configs/base.yml
parambole: Done. I have reused the existing pattern.

tests/check_qwen3_next_vs_reference.py
parambole: Done!

src/MaxText/configs/base.yml
parambole: Great suggestion. I have updated the names.

src/MaxText/configs/models/qwen3-next-80b-a3b.yml
parambole: Done!

src/MaxText/configs/base.yml
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/qwen3.py
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/qwen3.py
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/qwen3.py
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/decoders.py
parambole: Hey @bvandermoon I have removed this method all together thank you for the review.

tests/check_qwen3_next_vs_reference.py
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/qwen3.py
parambole: Done! Thank you for the suggestion.

src/MaxText/layers/qwen3.py
parambole: Thanks for the suggestion! The inputs to the matmuls within jax_chunk_gated_delta_rule are indeed cast to float32, so the precision setting does have an effect.

During testing against the PyTorch reference, I found that using jax.lax.Precision.HIGHEST was necessary to achieve numerical parity, likely due to the nature of the recurrent scan. Lower precision settings led to deviations.

To ensure correctness for this specific kernel, I've kept it hardcoded to HIGHEST for now and added a comment explaining this. Making it configurable might lead to unexpected numerical mismatches if lower precision is selected.

src/MaxText/configs/models/qwen3-next-80b-a3b.yml
parambole: Done!

src/MaxText/layers/qwen3.py
parambole: Thanks for the suggestion on reorganizing! I agree that jax_chunk_gated_delta_rule implements a more general linear attention mechanism that could potentially live in attention.py or a dedicated attention_mla.py file in the future.

For this initial PR, I've kept it within qwen3.py since it's currently only used here and is closely tied to the Qwen3NextGatedDeltaNet implementation. We might also replace this JAX version with a more optimized kernel later, which could influence the best long-term location.

Happy to revisit moving it in a follow-up PR if needed!

src/MaxText/layers/qwen3.py
parambole: Good point to check for reuse!

Qwen3NextRMSNorm differs from the standard RMSNorm in normalizations.py in two key ways to match the reference implementation:

It initializes the weight parameter to zeros (standard RMSNorm uses ones).

It applies the scale as (1.0 + weight) (standard RMSNorm applies it as * weight).

Because these behaviors are quite specific to Qwen3-Next and currently only used here, I've kept the definition within qwen3.py for now to group model-specific components together. Let me know if you strongly prefer moving it!

src/MaxText/layers/qwen3.py
parambole: Thanks for the feedback! I agree that breaking down jax_chunk_gated_delta_rule could improve readability.

For now, I've kept it as a single function because the different stages are quite interconnected for this specific algorithm. Also, we might eventually replace this pure JAX implementation with a more optimized kernel, so I thought it might be better to hold off on refactoring this version extensively.

Happy to revisit this if the JAX implementation remains long-term!

src/MaxText/layers/qwen3.py
parambole: Good point to check for reuse!

Because these behaviors are quite specific to Qwen3-Next and currently only used here, I've kept the definition within qwen3.py for now to group model-specific components together. Let me know if you strongly prefer moving it!

src/MaxText/layers/qwen3.py
parambole: While we generally use config.matmul_precision, I found during testing against the PyTorch reference that this specific recurrent kernel (jax_chunk_gated_delta_rule) requires jax.lax.Precision.HIGHEST to maintain numerical accuracy.

Since the inputs inside the function are cast to float32, the precision setting does have an impact, and lower settings caused deviations from the reference output. To ensure correctness, I've kept it hardcoded to HIGHEST for now and added a comment explaining this decision.

shuningjin (approved):
Thanks for the great effort!

PR #2492, opened 2025-10-13
Adjust batch and seq sharding rules based on EP_AS_CONTEXT config

NuojCheng:
# Description

Update batch and sequence sharding rules based on EP_AS_CONTEXT config.

# Tests

Results generated from v6e-4 TPU VM.

- `smoke_train ici_context_parallelism=4 dataset_type=synthetic profiler=xplane`
    - Webdiff: https://diff.googleplex.com/#key=ciEy3v1yNDN8
    - Before changes: http://xprof.corp.google.com/trace_viewer/chengnuojin-8289786618770909522
    - After changes: http://xprof.corp.google.com/trace_viewer/chengnuojin-7852601387453077593

- `smoke_train ici_context_parallelism=4 dataset_type=synthetic expert_shard_attention_option=context profiler=xplane`
    - Webdiff: https://diff.googleplex.com/#key=36cA7ApZucEk
    - Before changes: http://xprof.corp.google.com/trace_viewer/chengnuojin-10976403094821055906
    - After changes: http://xprof.corp.google.com/trace_viewer/chengnuojin-17276207501518055418

Conclusion: The new change improves FLOPs from ~140 to 155 under smoke training tests.

# Checklist


src/MaxText/layers/pipeline.py
```diff
@@ -53,14 +53,21 @@ class Pipeline(nn.Module):
   mesh: Mesh
   remat_policy: Any = None
 
-  def setup(self):
+  def setup(self):  # pylint: disable=missing-function-docstring
```
RissyRan, line 56: Nit: why not add it now?

src/MaxText/layers/decoders.py
```diff
@@ -88,15 +89,17 @@ def __call__(
...
     mesh = self.mesh
-    if model_mode == MODEL_MODE_PREFILL:
-      logical_axis_names = ("activation_batch", "prefill_activation_length", "activation_embed")
-    else:
-      logical_axis_names = ("activation_batch", "activation_length", "activation_embed")
 
-    if model_mode == MODEL_MODE_PREFILL:
-      inputs = nn.with_logical_constraint(inputs, logical_axis_names)
+    if self.model_mode == MODEL_MODE_PREFILL:
+      logical_axis_names = ("activation_batch", "prefill_activation_length", "activation_mlp")
```
RissyRan, line 94: Nit: I see this annotation is used a couple of times. Could we put it in the [common](https://github.com/AI-Hypercomputer/maxtext/blob/d6193a5000c68e20ab97357e07f58fc24976e2de/src/MaxText/common_types.py#L4) or something similar? 

It's nice to have, but not a blocker as we have multiple places in MaxText already, and haven't put them into util.

```
inputs_train_logical_axis_names = ("activation_batch", "activation_length_no_exp", "activation_mlp")
inputs_perfill_logical_axis_names = ("activation_batch", "prefill_activation_length", "activation_mlp")
inputs_train_ep_as_context_axis_names = ("activation_batch_no_exp", "activation_length", "activation_mlp")
```

RissyRan:
Please also add any tests you have run. Thanks!

src/MaxText/layers/pipeline.py
NuojCheng: I don't feel there is a need to add funciton docstring for this setup, which is self-explained pre well.

src/MaxText/layers/decoders.py
NuojCheng: I absolutely agree. However, it will involve a lot of changes, since currently we have logical axes specified basically everywhere in MaxText. I would prefer we do all of them together with Rich's sharding proposal instead of in this PR.

PR #2506, opened 2025-10-15
Update patch to ensure maxtext images are downgraded to 0.7.0

Rohan-Bierneni:
# Description

Maxtext unit tests are having a mismatch between jax version and libtpu version. They initially pull the 0.7.2 JAII, then downgrade only the jax and jaxlib versions to 0.7.2, but not the corresponding libtpu version. This is causing pallas libtpu error for unit tests using pathways backend. 

Notice 1: The downgrade to jax 0.7.0 is because google-tunix has a requirement to be <= 0.7.1 jax. Once tunix removes the jax version pin, we will revert these changes to the normal flow.

# Tests

Building a maxtext image and will verify the pip freeze: https://paste.googleplex.com/6454746957348864

# Checklist

### DIFF
--- a/maxtext_jax_ai_image.Dockerfile
+++ b/maxtext_jax_ai_image.Dockerfile
@@ -49,10 +49,18 @@ RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg
 
 # Install google-tunix for TPU devices, skip for GPU
 RUN if [ "$DEVICE" = "tpu" ]; then \
+        echo "TPU device detected. Installing google-tunix."; \
         python3 -m pip install 'google-tunix>=0.1.2'; \
-        # TODO: Once tunix stopped pinning jax 0.7.1, we should remove our 0.7.0 version pin (b/450286600)
-        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0'; \
-  fi
+\
+        if [[ "$JAX_AI_IMAGE_BASEIMAGE" == *"nightly"* ]]; then \
+            echo "Nightly image detected. Uninstalling base JAX and installing pre-release."; \
+            pip uninstall -y jax jaxlib libtpu; \
+            pip install -U --pre jax jaxlib libtpu requests -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/ -f https://storage.googleapis.com/jax-releases/libtpu_releases.html; \
+        else \
+            echo "Non-nightly image. Installing JAX 0.7.0."; \
+            python3 -m pip install jax[tpu]==0.7.0; \
+        fi; \
+    fi
 
 # Now copy the remaining code (source files that may change frequently)
 COPY . .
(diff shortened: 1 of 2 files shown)


gobbleturk (approved):

maxtext_jax_ai_image.Dockerfile
```diff
@@ -51,7 +51,7 @@ RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg
 RUN if [ "$DEVICE" = "tpu" ]; then \
         python3 -m pip install 'google-tunix>=0.1.2'; \
         # TODO: Once tunix stopped pinning jax 0.7.1, we should remove our 0.7.0 version pin (b/450286600)
-        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0'; \
+        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0' 'libtpu==0.0.19'; \
```
hengtaoguo: Is there any way to skip this line when `MODE=nightly`? Or only do so with stable mode?

Rohan-Bierneni: Since we use mode=stable_stack for all images built off of JAII, we distinguish them normally via the image passed to BASEIMAGE. And the only way to distinguish nightly vs stable versions would be if it has "nightly" in the text

hengtaoguo: Got it, do you think this code snippet will help in both cases? It forces to upgrade JAX version to the newest in the nightly version, so the libtpu could match ([b/451429959](https://b.corp.google.com/issues/451429959)).

```
        if [[ "$JAX_AI_IMAGE_BASEIMAGE" == *"nightly"* ]]; then \
            echo "Nightly image detected"; \
            python3 -m pip install --upgrade jax[tpu]; \
        else \
            echo "Non-nightly image"; \
            python3 -m pip install jax[tpu]==0.7.0; \
        fi
```

cc @gobbleturk @bvandermoon Basically we are (1) upgrading JAX to latest 0.8.0 for nightly images (2) downgrading JAX to 0.7.0 for github CI stable stack images. Both cases are after the Tunix installation so we can keep JAX version. Not sure if this is a worthwhile patch given this problem is transient.

Rohan-Bierneni: For nightly images I think we have to install via these commands:
`
pip uninstall -y jax jaxlib libtpu
`
`
pip install -U --pre jax jaxlib libtpu requests -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/ -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
`

Rohan-Bierneni: This works, was able to force install jax nightly only for nightly jaii images: https://paste.googleplex.com/5962851466477568

maxtext_jax_ai_image.Dockerfile
```diff
@@ -47,11 +47,25 @@ RUN if [ "$DEVICE" = "tpu" ] && [ "$JAX_STABLE_STACK_BASEIMAGE" = "us-docker.pkg
...
+            python3 -m pip install jax[tpu]==0.7.0; \
+        fi; \
+    fi
+
 # Install google-tunix for TPU devices, skip for GPU
 RUN if [ "$DEVICE" = "tpu" ]; then \
         python3 -m pip install 'google-tunix>=0.1.2'; \
         # TODO: Once tunix stopped pinning jax 0.7.1, we should remove our 0.7.0 version pin (b/450286600)
-        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0'; \
+        python3 -m pip install 'jax==0.7.0' 'jaxlib==0.7.0' 'libtpu==0.0.19'; \
```
bvandermoon: Why not just install `jax[tpu]`?

Rohan-Bierneni: jax[tpu] would install jax 0.8.0, which hasn't been tested yet with maxtext for compatability. And there is known issues with 0.7.2 and 0.7.1, so maxtext would be on 0.7.0 until 0.8.0 compatability is confirmed

PR #2514, opened 2025-10-16
[setup.sh] Fix Python version check to correctly lexicographically compare

SamuelMarks:
# Description

[setup.sh] Fix Python version check to correctly lexicographically compare

# Tests

```sh
$ python3 --version
Python 3.12.8
$ PY_VERSION=$(python3 --version 2>&1)
$ PY_VERSION=${PY_VERSION##* }
$ PY_VERSION=${PY_VERSION%.*}
$ printf '%s\n' "$PY_VERSION"
3.12
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
gte
$ PY_VERSION='3.13'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
gte
$ PY_VERSION='3.11'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
less than
$ PY_VERSION='3.12'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
gte
```

# Checklist

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -49,7 +49,7 @@ fi
 PY_VERSION=${PY_VERSION##* }
 PY_VERSION=${PY_VERSION%.*}
 # shellcheck disable=SC2072
-if [ "$PYTHON_VERSION" \< '3.12' ]; then
+if [[ '3.12' > "$PY_VERSION" ]]; then
     echo -e "\n\e[31mERROR: Outdated Python Version! You are currently using ${PY_VERSION}.*, but MaxText requires Python version 3.12 or higher.\e[0m"
     # Ask the user if they want to create a virtual environment with uv
     read -p "Would you like to create a Python 3.12 virtual environment using uv? (y/n) " -n 1 -r


setup.sh
```diff
@@ -49,7 +49,7 @@ fi
 PY_VERSION=${PY_VERSION##* }
 PY_VERSION=${PY_VERSION%.*}
 # shellcheck disable=SC2072
-if [ "$PYTHON_VERSION" \< '3.12' ]; then
+if [[ ! "$PY_VERSION" > '3.11' ]]; then
```
bvandermoon: Is the < intentionally changed to > 3.11? Isn't this code intended to run only when python < 3.12?

SamuelMarks: yeah it's a bit annoying but that's how you do it, see the `!`.

See: https://stackoverflow.com/a/21295242

and see my test in the description of this PR if you want to experiment with different cases:
```sh
$ PY_VERSION='3.13'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
gte
$ PY_VERSION='3.11'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
less than
$ PY_VERSION='3.12'
$ if [[ ! "$PY_VERSION" > '3.11' ]]; then echo less than ; else echo gte ; fi
gte
```

bvandermoon: Thanks Samuel, makes sense. I guess we can go with this if needed, but it's a bit confusing like you mentioned. Would this work instead? It's clearer:

```
if [[ '3.12' > "$PY_VERSION" ]]; then
```

SamuelMarks: That works also. Arguable that it's clearer. Your call which one you want in `main`.

bvandermoon: Yes, let's go with that one. Thanks Samuel

PR #2361, opened 2025-09-17
Refactor: third phase of RESTRUCTURE.md

SamuelMarks:
# Description

Third phase of [RESTRUCTURE.md](./RESTRUCTURE.md). Follows #2144 & #2304.

# Tests

Whence CI passes manual testing can happen; then this is ready for merge.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #2515, opened 2025-10-16
[setup.sh] Install extra_deps_from_github.txt unconditionally on nightly ; [*.Dockerfile] Cache extra_deps_from_github.txt

SamuelMarks:
# Description

[setup.sh] Install extra_deps_from_github.txt unconditionally ; [*.Dockerfile] Cache extra_deps_from_github.txt at same time as requirements.txt ; [src/install_maxtext_extra_deps/install_github_deps.py] Use new location

# Tests

N/A

# Checklist

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -137,32 +137,29 @@ if [[ $DEVICE == "tpu" ]]; then
     fi
 fi
 
-# Save the script folder path of maxtext
-run_name_folder_path=$(pwd)
-
-# Install dependencies from requirements.txt
-cd "$run_name_folder_path" && python3 -m uv pip install --upgrade pip
 if [[ "$MODE" == "nightly" ]]; then
     echo "Nightly mode: Installing requirements.txt, stripping commit pins from git+ repos."
-    cp requirements.txt requirements.txt.nightly-temp
+
     # Create a temp file, strip commit pins from git+ repos in requirements.txt
     # Remove/update this section based on the pinned github repo commit in requirements.txt
-    sed -i -E \
-      -e 's|^mlperf-logging @ https?://github.com/mlcommons/logging/archive/.*\.zip$|mlperf-logging@git+https://github.com/mlperf/logging.git|' \
+    sed -E \
       -e 's|^([^ ]*) @ https?://github.com/([^/]*\/[^/]*)/archive/.*\.zip$|\1@git+https://github.com/\2.git|' \
-      requirements.txt.nightly-temp
+      -e '/JetStream/d' \
+      -e '/mlperf-logging/d' \
+      requirements.txt > requirements.txt.nightly-temp
 
     echo "--- Installing modified nightly requirements: ---"
     cat requirements.txt.nightly-temp
     echo "-------------------------------------------------"
     
-    python3 -m uv pip install --no-cache-dir -U -r requirements.txt.nightly-temp
+    python3 -m uv pip install --no-cache-dir -U -r requirements.txt.nightly-temp \
+                                                -r "${MAXTEXT_REPO_ROOT?}"'/extra_deps_from_github.txt'
     rm requirements.txt.nightly-temp
 else
     # stable or stable_stack mode: Install with pinned commits
     echo "Installing tpu-requirements.txt with pinned commits."
     tpu_requirements_txt=
-    for candidate in 'generated_requirements' "${MAXTEXT_REPO_ROOT}"'/generated_requirements' "$PWD"; do
+    for candidate in 'generated_requirements' "${MAXTEXT_REPO_ROOT?}"'/generated_requirements' "$PWD"; do
       if [ -f "$candidate"'/tpu-requirements.txt' ]; then
         tpu_requirements_txt="$candidate"'/tpu-requirements.txt'
         break
@@ -174,7 +171,8 @@ else
       >&2 printf 'Could not find "tpu-requirements.txt", looked in: %s\n' "${searched%?}"
       exit 2
     else
-      python3 -m uv pip install --resolution=lowest -r "$tpu_requirements_txt"
+      python3 -m uv pip install --resolution=lowest -r "$tpu_requirements_txt" \
+                                                    -r "${MAXTEXT_REPO_ROOT?}"'/extra_deps_from_github.txt'
     fi
 fi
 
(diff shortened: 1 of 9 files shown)


setup.sh
```diff
@@ -141,7 +143,7 @@ fi
 run_name_folder_path=$(pwd)
 
 # Install dependencies from requirements.txt
-cd "$run_name_folder_path" && python3 -m uv pip install --upgrade pip
+cd "$run_name_folder_path"
```
bvandermoon: Why is this removed?

setup.sh
```diff
@@ -118,6 +118,8 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE || ($MODE == "nightly"
      exit 1
 fi
 
+uv pip install -r './extra_deps_from_github.txt'
+
```
bvandermoon: Shouldn't this go after `python3 -m uv pip install --resolution=lowest -r "$tpu_requirements_txt"`?

SamuelMarks: It could. I put it earlier so that the dependencies wouldn't conflict with pinned version dependencies later. Your call.

setup.sh
SamuelMarks: We are not using `pip` to install, we are using `uv pip` (completely different implementation in Rust).

src/install_maxtext_extra_deps/install_github_deps.py
```diff
@@ -32,11 +32,11 @@ def main():
   This script looks for 'extra_deps_from_github.txt' relative to its own location.
   It executes 'uv pip install -r <path_to_extra_deps.txt> --resolution=lowest'.
   """
-  script_dir = Path(__file__).resolve().parent
+  MAXTEXT_REPO_ROOT = os.environ.get("MAXTEXT_REPO_ROOT", os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```
kanglant: Why this file is moved?

SamuelMarks: It needs to go with all the other .txt files because the #2361 looks in just that one place for dependencies; the one top level `dependencies` folder.

setup.sh
kanglant: Please move it after tpu_requirements_txt installation. For stable package with every dep pinned, there shouldn't be a new conflict. I would be surprised if there is a new conflict. It means there're something wrong.

src/install_maxtext_extra_deps/install_github_deps.py
```diff
@@ -59,7 +59,6 @@ def main():
       "install",
       "-r",
       str(extra_deps_file),
-      "--no-deps",
```
kanglant, line 62: Please do not remove this `--no-deps` flag. This is intentional. If people want to add new transitive lib, they should update the commit hash of the github dep in the base requirement files, run seed-env to generate and update lock files, move the github deps to the extra_deps_from_github.txt. In this way, the lock files contain the transitive lib.

setup.sh
SamuelMarks: Ok

src/install_maxtext_extra_deps/install_github_deps.py
kanglant: All the package tests failed with this change. please take a look

kanglant: We need to ensure that external users who simply run pip install maxtext can successfully execute this command without needing to manually copy any MaxText files.

kanglant: Since we confirmed that GitHub dependencies cannot be placed in pyproject.toml, I think we have no choice but to include this github dep data as a .txt file inside the maxtext package.

setup.sh
kanglant: Why do we need `$run_name_folder_path`?

setup.sh
```diff
@@ -162,7 +157,7 @@ else
     # stable or stable_stack mode: Install with pinned commits
     echo "Installing tpu-requirements.txt with pinned commits."
     tpu_requirements_txt=
-    for candidate in 'generated_requirements' "${MAXTEXT_REPO_ROOT}"'/generated_requirements' "$PWD"; do
+    for candidate in 'generated_requirements' "${MAXTEXT_REPO_ROOT?}"'/generated_requirements' "$PWD"; do
```
bvandermoon, line 162: Why this update? Was it causing some issue previously?

SamuelMarks: Being extra cautious that this var is always defined; even if not run in the Dockerfile we provide.

setup.sh
```diff
@@ -178,6 +173,8 @@ else
     fi
 fi
 
+uv pip install -r "${MAXTEXT_REPO_ROOT?}"'/extra_deps_from_github.txt'
```
kanglant: I think this line needs to be run in the same block following the installation of the dependencies specified in tpu-requirements.txt.

Placing it here will overwrite those dependencies when running in nightly mode (see line 146).

For a long-term solution, we should plan to restructure setup.sh entirely to clearly define the installation flows, making the script much easier to understand and maintain.

SamuelMarks: @kanglant Ok so do you want `uv pip install -r "${MAXTEXT_REPO_ROOT?}"'/extra_deps_from_github.txt'` to be at the bottom of the `else` condition and at the top of the `if [[ "$MODE" == "nightly" ]]; then` condition then?

kanglant: Moving it to the `else` condition after running`python3 -m uv pip install --resolution=lowest -r "$tpu_requirements_txt"` is enough. We do not need to put it at the top of the nightly condition now. The nightly condition is using the old [requirements.txt](https://github.com/AI-Hypercomputer/maxtext/blob/main/requirements.txt#L41-L42), which contains the the github deps.

kanglant: Also, shouldn't the path be `"${MAXTEXT_REPO_ROOT?}"'src/install_maxtext_extra_deps/extra_deps_from_github.txt'`? Please verify the github sources are installed successfully in the docker image before merging.

SamuelMarks: No because docker moves it in the `COPY`

kanglant: @SamuelMarks could you move this line to the else condition?

PR #2521, opened 2025-10-18
Fix exception causes in data_loader.py

cool-RR:
This makes the correct message appear between the 2 chained exceptions.


bvandermoon (commented):
Thank you @cool-RR for the contribution. Would you mind adding an example in the PR description showing how this log looks before/after this change?

Issue #683, opened 2024-06-04
Llama3


peregilk:
Has anyone tried to train the newest models on MaxText. For instance Llama3 and Mistral v.0.3?

It is a bit unclear to me how much work this might be to support these models here. Do you for instance have to implement GQA in Llama3. 

If this is mainly updating the config files, it would be extremely helpful if there were working config files for the newest versions of Llama, Gemma and Mistral.

borisdayma:
It looks like there is a PR ongoing: https://github.com/google/maxtext/pull/653

gobbleturk:
Merged!

PR #2529, opened 2025-10-21
Qwen3-Next Gated FullAttention Implementation

Rohan-Bierneni:
# Description

This pr adds the Qwen3-Next Gated Full Attention implementation to the existing Qwen3-Next code in qwen3.py. 

Current implementation in maxtext uses normal Attention, but qwen3 next requires attention with 2 slight tweaks: an output gate and partial ROPE applied to 25% of head_dim. This pr adds this functionality by building a custom attention class for Qwen3-Next on top of AttentionOp for the core attention calculation.

FIXES: b/448407748

# Tests

I have added a testcase that compares the pytorch ref to the jax ref and compares the output tensors after the gated fullattention layer. 

This tests is passing after running this command: `pytest -vvs tests/check_qwen3_next_vs_reference.py::TestQwen3Next::test_full_attention_jax_vs_pytorch`: https://paste.googleplex.com/5626786360721408

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -905,6 +905,8 @@ gdn_num_value_heads: 32
 gdn_chunk_size: 64
 # Whether to apply L2 normalization to query and key tensors inside the Gated Delta Rule kernel.
 use_qk_norm_in_gdn: True
+# The ratio of dimension to apply ROPE on
+partial_rotary_factor: 1.0
 
 # Use tokamax library for gmm kernel implementation
 use_tokamax_gmm: false
--- a/src/MaxText/layers/attentions.py
+++ b/src/MaxText/layers/attentions.py
@@ -65,10 +65,11 @@ from MaxText.layers.embeddings import (
     LlamaVisionRotaryEmbedding,
     RotaryEmbedding,
     YarnRotaryEmbedding,
+    Qwen3NextRotaryEmbedding,
 )
 from MaxText.layers.initializers import nd_dense_init, NdInitializer, variable_to_logically_partitioned, default_bias_init
 from MaxText.layers.linears import DenseGeneral, canonicalize_tuple, normalize_axes
-from MaxText.layers.normalizations import RMSNorm
+from MaxText.layers.normalizations import RMSNorm, Qwen3NextRMSNorm
 from MaxText.layers.quantizations import AqtQuantization as Quant
 
 # pylint: disable=line-too-long, g-doc-args, g-doc-return-or-yield, bad-continuation, g-inconsistent-quotes
@@ -416,6 +417,8 @@ class Attention(nnx.Module):
     self.model_mode = model_mode
     self.rngs = rngs
 
+    self.is_qwen3_next = self.config.decoder_block == DecoderBlockType.QWEN3_NEXT
+
     # Module attribute names must match names previously passed to Linen for checkpointing
     self.KVCache_0 = (
         self.init_kv_caches(inputs_kv_shape=inputs_kv_shape)
@@ -478,6 +481,9 @@ class Attention(nnx.Module):
     else:
       self.sinks = None
 
+    self.query_norm = None
+    self.key_norm = None
+
     is_llama4_decoder_block = self.config.decoder_block == DecoderBlockType.LLAMA4
     if self.use_qk_norm and not is_llama4_decoder_block:
       self.query_norm = RMSNorm(
@@ -498,9 +504,21 @@ class Attention(nnx.Module):
           kernel_axes=("norm",),
           rngs=self.rngs,
       )
-    else:
-      self.query_norm = None
-      self.key_norm = None
+    elif self.is_qwen3_next:
+      self.query_norm = Qwen3NextRMSNorm(
+          num_features=self.config.head_dim,
+          eps=self.config.normalization_layer_epsilon,
+          dtype=self.config.dtype,
+          weight_dtype=self.config.weight_dtype,
+          rngs=self.rngs,
+      )
+      self.key_norm = Qwen3NextRMSNorm(
+          num_features=self.config.head_dim,
+          eps=self.config.normalization_layer_epsilon,
+          dtype=self.config.dtype,
+          weight_dtype=self.config.weight_dtype,
+          rngs=self.rngs,
+      )
 
     self._maybe_shard_with_logical = functools.partial(
         maybe_shard_with_logical,
@@ -538,9 +556,15 @@ class Attention(nnx.Module):
     kernel_axes = (
         (None, None, None) if self.config.ici_context_autoregressive_parallelism > 1 else ("embed", "q_heads", "kv")
     )
+    in_features = self.convert_dense_general_inputs_shape(inputs_q_shape)
+    out_features = (self.num_query_heads, self.head_dim)
+
+    if self.is_qwen3_next:
+      out_features = (self.num_query_heads, self.head_dim * 2)
+
     return DenseGeneral(
-        in_features_shape=self.convert_dense_general_inputs_shape(inputs_q_shape),
-        out_features_shape=(self.num_query_heads, self.head_dim),
+        in_features_shape=in_features,
+        out_features_shape=out_features,
         axis=-1,
         kernel_init=query_init,
         kernel_axes=kernel_axes,
@@ -642,13 +666,22 @@ class Attention(nnx.Module):
 
   def init_out_w(self, output_dim: int) -> nnx.Module:
     """out projection"""
+    in_features = (self.num_query_heads, self.head_dim)
+    out_features = output_dim
     out_kernel_axis = (
         (None, None, None) if self.config.ici_context_autoregressive_parallelism > 1 else ("heads", "kv", "embed")
     )
+    axis = (-2, -1)
+
+    if self.is_qwen3_next:
+      in_features = self.num_query_heads * self.head_dim
+      out_kernel_axis = ("mlp", "embed")
+      axis = (-1,)
+
     return DenseGeneral(
-        in_features_shape=(self.num_query_heads, self.head_dim),
-        out_features_shape=output_dim,
-        axis=(-2, -1),
+        in_features_shape=in_features,
+        out_features_shape=out_features,
+        axis=axis,
         kernel_init=self.kernel_init,
         kernel_axes=out_kernel_axis,  # trade speed with memory
         dtype=self.dtype,
@@ -718,6 +751,16 @@ class Attention(nnx.Module):
           interleave=self.config.rope_interleave,
           truncate=self.config.rope_truncate,
           attention_scaling=self.config.rope_attention_scaling,
+          rngs=self.rngs,
+      )
+    elif self.is_qwen3_next:
+      rotary_embedding = Qwen3NextRotaryEmbedding(
+          min_timescale=self.config.rope_min_timescale,
+          max_timescale=self.config.rope_max_timescale,
+          embedding_dims=self.config.head_dim,
+          partial_rotary_factor=self.config.partial_rotary_factor,
+          cast_as_fprop_dtype=True,
+          fprop_dtype=self.config.dtype,
           rngs=self.rngs,
       )
     else:
@@ -890,9 +933,17 @@ class Attention(nnx.Module):
       value_sharding = NamedSharding(self.mesh, nn.logical_to_mesh_axes(self.value_axis_names))
       value = self.kv_projection(inputs_kv, proj_name="value", out_sharding=value_sharding)
 
+    gate = None
+    if self.is_qwen3_next:
+      # Split query into query & gate.
+      query, gate = jnp.split(query, 2, axis=-1)
+      batch_size, seq_len, _, _ = gate.shape
+      gate = gate.reshape(batch_size, seq_len, self.config.num_query_heads * self.config.head_dim)
+
     is_llama4_decoder_block = self.config.decoder_block == DecoderBlockType.LLAMA4
     # NOTE: llama 4 does L2 normalization after RoPE
-    if self.use_qk_norm and not is_llama4_decoder_block:
+    # Apply Qwen3Next specific RMS Norm
+    if (self.use_qk_norm and not is_llama4_decoder_block) or self.is_qwen3_next:
       query = self.query_norm(query)
       key = self.key_norm(key)
 
@@ -964,7 +1015,9 @@ class Attention(nnx.Module):
           bidirectional_mask,
           self.sinks,
       )
-
+    if self.is_qwen3_next:
+      out = out.reshape(batch_size, seq_len, self.config.num_query_heads * self.config.head_dim)
+      out = out * jax.nn.sigmoid(gate)
     if model_mode == MODEL_MODE_PREFILL:
       out = self._maybe_shard_with_logical(out, self.prefill_out_axis_names)
     elif model_mode == MODEL_MODE_TRAIN and self.config.expert_shard_attention_option == EP_AS_CONTEXT:
--- a/src/MaxText/layers/embeddings.py
+++ b/src/MaxText/layers/embeddings.py
@@ -380,6 +380,97 @@ def llama_rotary_embedding_as_linen(
   )
 
 
+def qwen3_next_rotary_embedding_as_linen(
+    *,
+    min_timescale: int,
+    max_timescale: int,
+    embedding_dims: int = 0,
+    partial_rotary_factor: float = 0.25,
+    cast_as_fprop_dtype: bool = True,
+    fprop_dtype: DType = jnp.bfloat16,
+    name: str | None = None,
+):
+  """Initializes the Qwen3NextRotaryEmbedding module and returns it as a Linen module.
+
+  Args:
+    min_timescale: Start of the geometric index. Determines the periodicity of
+      the added signal.
+    max_timescale: End of the geometric index. Determines the frequency of the
+      added signal.
+    embedding_dims: Dimension of the embedding to be generated.
+    partial_rotary_factor: Ratio of dimensions to apply ROPE to.
+    cast_as_fprop_dtype: Whether to cast the output to the fprop dtype.
+    fprop_dtype: The dtype of the output.
+    name: Name of the Linen module.
+  """
+  return nnx_wrappers.to_linen(
+      Qwen3NextRotaryEmbedding,
+      min_timescale=min_timescale,
+      max_timescale=max_timescale,
+      embedding_dims=embedding_dims,
+      partial_rotary_factor=partial_rotary_factor,
+      cast_as_fprop_dtype=cast_as_fprop_dtype,
+      fprop_dtype=fprop_dtype,
+      metadata_fn=variable_to_logically_partitioned,
+      name=name,
+  )
+
+
+class Qwen3NextRotaryEmbedding(RotaryEmbedding):
+  """Qwen3 Next variant of ROPE (partial ROPE)"""
+
+  def __init__(
+      self,
+      min_timescale: int,
+      max_timescale: int,
+      embedding_dims: int = 0,
+      cast_as_fprop_dtype: bool = True,
+      fprop_dtype: DType = jnp.bfloat16,
+      partial_rotary_factor: float = 0.25,
+      rngs: nnx.Rngs = None,
+  ):
+    """Initializes the Qwen3NextRotaryEmbedding module.
+
+    Args:
+      min_timescale: Start of the geometric index. Determines the periodicity of
+        the added signal.
+      max_timescale: End of the geometric index. Determines the frequency of the
+        added signal.
+      embedding_dims: Dimension of the embedding to be generated.
+      partial_rotary_factor: Ratio of dimensions to apply ROPE to
+      rngs: rng keys passed in by nnx.bridge.to_linen.
+    """
+    self.head_dim = embedding_dims
+    self.partial_rotary_factor = partial_rotary_factor
+    self.rotary_dim = int(self.head_dim * self.partial_rotary_factor)
+
+    super().__init__(
+        min_timescale=min_timescale,
+        max_timescale=max_timescale,
+        embedding_dims=self.rotary_dim,
+        cast_as_fprop_dtype=cast_as_fprop_dtype,
+        fprop_dtype=fprop_dtype,
+        rngs=rngs,
+    )
+
+  def __call__(self, inputs: jax.Array, position: None | jax.Array = None) -> jax.Array:
+    """Applies LLaMA variant of rotary position embedding.
+
+    Args:
+      inputs: The input sequence on which to apply the Rotary position
+        embedding. It is assumed of shape [B, S, H, D].
+      position: Optional position array [B, S]. Only needed when the sequence
+        is packed.
+
+    Returns:
+      A jax.Array of shape [B, S, H, D - rotary_dim] with rotary position embeddings applied.
+    """
+    inputs_rot, inputs_pass = jnp.split(inputs, [self.rotary_dim], axis=-1)
+    inputs_rot = super().__call__(inputs_rot, position)
+    inputs = jnp.concatenate([inputs_rot, inputs_pass], axis=-1)
+    return inputs
+
+
 class LLaMARotaryEmbedding(RotaryEmbedding):
   """LLaMA variant of ROPE."""
 
--- a/src/MaxText/layers/normalizations.py
+++ b/src/MaxText/layers/normalizations.py
@@ -18,6 +18,7 @@ from typing import Any
 
 from flax import linen as nn
 from flax import nnx
+from flax.linen import initializers as linen_initializers
 from jax import lax
 import jax
 import jax.numpy as jnp
@@ -26,7 +27,7 @@ from MaxText import max_logging
 from MaxText import max_utils
 from MaxText.layers import nnx_wrappers
 from MaxText.layers.initializers import Initializer, variable_to_logically_partitioned
-from MaxText.common_types import Array, ShardMode
+from MaxText.common_types import Array, DType, ShardMode
 
 
 class RMSNorm(nnx.Module):
@@ -42,6 +43,7 @@ class RMSNorm(nnx.Module):
       kernel_axes: tuple[None | str, ...] = (),
       scale_init: Initializer = nn.initializers.ones,
       parameter_memory_host_offload: bool = False,
+      scale_offset: float = 0.0,
       *,
       rngs: nnx.Rngs,
   ):
@@ -53,6 +55,7 @@ class RMSNorm(nnx.Module):
     self.kernel_axes = kernel_axes
     self.scale_init = scale_init
     self.parameter_memory_host_offload = parameter_memory_host_offload
+    self.scale_offset = scale_offset
     self.scale = nnx.Param(
         scale_init(rngs.params(), (num_features,), weight_dtype),
         sharding=kernel_axes,
@@ -73,8 +76,83 @@ class RMSNorm(nnx.Module):
       out_sharding = None
 
     scale = jnp.asarray(scale, self.dtype)
+    effective_scale = scale + self.scale_offset  # Apply offset
     # broadcast 2nd input then element-wise mul
-    return jnp.einsum("i...k,...k->i...k", y, scale, out_sharding=out_sharding)
+    return jnp.einsum("i...k,...k->i...k", y, effective_scale, out_sharding=out_sharding)
+
+
+def Qwen3NextRMSNorm(num_features: int, eps: float, dtype: DType, weight_dtype: DType, *, rngs: nnx.Rngs):
+  """
+  Used for input and post attention layernorms
+  in Qwen3NextDecoderLayer.
+
+  This normalization layer is specific to Qwen3-Next. Key characteristics:
+  1.  The learnable scale parameter `scale` is initialized to ZEROS.
+  2.  The scale is applied as `(1.0 + self.scale)`, making the initial scale effectively 1.0.
+      This matches the PyTorch implementation of Qwen3NextRMSNorm.
+  """
+  return nnx.data(
+      RMSNorm(
+          num_features=num_features,
+          epsilon=eps,
+          dtype=dtype,
+          weight_dtype=weight_dtype,
+          scale_init=linen_initializers.zeros,
+          scale_offset=1.0,
+          rngs=rngs,
+      )
+  )
+
+
+class Qwen3NextRMSNormGated(nnx.Module):
+  """
+  This applies RMS Normalization and then a gated activation function (SiLU).
+  This is used within the Qwen3NextGatedDeltaNet.
+
+  The normalization is performed by an internal `RMSNorm` instance (`self.rms_norm`),
+  which has its own learnable `scale` parameter, initialized to ONES.
+
+  Attributes:
+    num_features: The number of features in the input.
+    eps: A small epsilon value to prevent division by zero in RMSNorm.
+    dtype: The datatype of the computation.
+    weight_dtype: The datatype of the internal RMSNorm scale.
+  """
+
+  def __init__(self, num_features: int, eps: float, dtype: DType, weight_dtype: DType, *, rngs: nnx.Rngs):
+    self.num_features = num_features
+    self.eps = eps
+    self.dtype = dtype
+    self.weight_dtype = weight_dtype
+    self.rms_norm = nnx.data(
+        RMSNorm(
+            num_features=num_features,
+            epsilon=eps,
+            dtype=dtype,
+            weight_dtype=weight_dtype,
+            scale_init=nnx.initializers.ones,
+            rngs=rngs,
+        )
+    )
+
+  def __call__(self, hidden_states: Array, gate: Array) -> Array:
+    """
+    Applies RMSNorm and then a SiLU gate.
+
+    Args:
+      hidden_states: The input array to be normalized (o). Shape: (..., F)
+      gate: The gating array for the activation (z). Shape: (..., F)
+            where F is num_features.
+
+    Returns:
+      The normalized and gated output array. Shape: (..., F)
+    """
+    normalized_states = self.rms_norm(hidden_states)
+
+    # Gated Activation using SiLU (Sigmoid-weighted Linear Unit)
+    gated_states = normalized_states * jax.nn.silu(gate.astype(jnp.float32))
+
+    return gated_states.astype(self.dtype)
 
 
 def rms_norm(
--- a/src/MaxText/layers/qwen3.py
+++ b/src/MaxText/layers/qwen3.py
@@ -26,17 +26,16 @@ import jax.numpy as jnp
 
 from flax import linen as nn
 from flax import nnx
-from flax.linen import initializers as linen_initializers
 
 from MaxText import max_utils
-from MaxText.common_types import Config, DType, Array
+from MaxText.common_types import Config, DType, Array, BATCH, LENGTH_NO_EXP, EMBED
 from MaxText.layers import attentions
 from MaxText.layers import initializers as max_initializers
 from MaxText.layers import linears
 from MaxText.layers import moe
 from MaxText.layers import nnx_wrappers
 from MaxText.layers import quantizations
-from MaxText.layers.normalizations import RMSNorm, l2norm
+from MaxText.layers.normalizations import RMSNorm, l2norm, Qwen3NextRMSNorm, Qwen3NextRMSNormGated
 from MaxText.layers.quantizations import AqtQuantization as Quant
 from MaxText.inference import page_manager
 from MaxText.layers.attentions import Attention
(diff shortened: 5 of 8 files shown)


src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+      quant=self.quant,
+      kv_quant=quantizations.configure_kv_quant(cfg),
+      num_query_heads=cfg.num_query_heads,
+      num_kv_heads=cfg.num_kv_heads,
+      dropout_rate=cfg.dropout_rate,
+      dtype=cfg.dtype,
+      rngs=rngs,
+    )
+
+    # 8. KVCache (if not training)
```
parambole: @Rohan-Bierneni - I believe we should omit caching specific logic from here and other places used in the file.

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
-      deterministic: bool,
-      model_mode: str,
-      # TODO(parambole): Add cache arguments
+        self,
+        inputs: jnp.ndarray,
+        decoder_segment_ids: None | jnp.ndarray,
+        decoder_positions: None | jnp.ndarray,
+        deterministic: bool,
+        model_mode: str,
+        previous_chunk=None, # Added for cache update
```
parambole: Please remove "Added" and other inline comments that are not relevant here?

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
-        inputs,
-        inputs,
-        decoder_positions,
-        decoder_segment_ids=decoder_segment_ids,
-        deterministic=deterministic,
-        model_mode=model_mode,
+    cfg = self.config
+    batch_size, seq_len, _ = inputs.shape
+
+    print(f"inputs: {inputs.shape}") # DEBUG
```
parambole: Please remove the print statement.

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+
+    # Project and split Query and Gate
+    # q_proj_out shape: (B, S, E * 2)
+    q_proj_out = self.q_proj(inputs)
+    print(f"q_proj_out: {q_proj_out.shape}")
```
parambole: Please remove the print statements here and other places in the file.

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+            model_mode=model_mode,
+            previous_chunk=previous_chunk,
+        )
+        cached_values = [prefill_kv_cache, ar_kv_cache]
+    
+    # Compute Attention
+    # Transpose to (B, H, S, D) for attention op
+    # query_states = jnp.transpose(query_states, (0, 2, 1, 3))
+    # key_states = jnp.transpose(key_states, (0, 2, 1, 3))
+    # value_states = jnp.transpose(value_states, (0, 2, 1, 3))
```
parambole: Please remove debug comments.

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+      key_rot, key_pass = jnp.split(key_states, [rotary_dim], axis=-1)
+
+      query_rot = self.rotary_embedding(query_rot, decoder_positions)
+      key_rot = self.rotary_embedding(key_rot, decoder_positions)
+
+      # shape doesnt change stays as (B, S, H_q, D)
+      query_states = jnp.concatenate([query_rot, query_pass], axis=-1)
+      key_states = jnp.concatenate([key_rot, key_pass], axis=-1)
+
+    # KVCache Update/Lookup
```
parambole: Please remove caching logic

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+    # RESHAPE to (B, S, H_q, D) BEFORE norm
+    query_states = query_states.reshape(batch_size, seq_len, cfg.num_query_heads, cfg.head_dim)
+    print(f"query_states after norm reshape: {query_states.shape}")
+    # (B, S, H_q, D)
+    gate = gate.reshape(batch_size, seq_len, cfg.num_query_heads, cfg.head_dim)
+    # (B, S, H_kv, D)
+    key_states = key_states.reshape(batch_size, seq_len, cfg.num_kv_heads, cfg.head_dim)
+    print(f"key_states after norm reshape: {key_states.shape}")
+    # (B, S, H_kv, D)
+    value_states = value_states.reshape(batch_size, seq_len, cfg.num_kv_heads, cfg.head_dim)
```
parambole: @Rohan-Bierneni - Can you please elaborate on why are these reshapes required here?

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,237 @@ def __init__(
...
+    if rotary_dim > 0:
+      query_rot, query_pass = jnp.split(query_states, [rotary_dim], axis=-1)
+      key_rot, key_pass = jnp.split(key_states, [rotary_dim], axis=-1)
+
+      query_rot = self.rotary_embedding(query_rot, decoder_positions)
+      key_rot = self.rotary_embedding(key_rot, decoder_positions)
+
+      # shape doesnt change stays as (B, S, H_q, D)
+      query_states = jnp.concatenate([query_rot, query_pass], axis=-1)
+      key_states = jnp.concatenate([key_rot, key_pass], axis=-1)
```
parambole: @Rohan-Bierneni - Can we move this to a method ? and potentially move this to src/MaxText/layers/embeddings.py ?

tests/check_qwen3_next_vs_reference.py
```diff
@@ -28,6 +30,12 @@
 from jax.sharding import Mesh
 from flax import nnx
 
+from collections.abc import Callable
+from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
+from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
+from transformers.utils import TransformersKwargs, auto_docstring, logging
+from transformers.processing_utils import Unpack
```
parambole: @Rohan-Bierneni - Why do we need these transfomers packages? can we do them using `torch` ?

tests/check_qwen3_next_vs_reference.py
```diff
@@ -39,6 +47,38 @@
...
+    mask = torch.triu(torch.ones(q_seq_len, kv_seq_len, dtype=torch.bool), diagonal=1)
+    masked_fill_value = -torch.finfo(dtype).max / 2
+    return torch.zeros(q_seq_len, kv_seq_len, dtype=dtype).masked_fill(mask, masked_fill_value)
+
+def rotate_half(x):
+    """Rotates half the hidden dims of the input."""
+    x1 = x[..., : x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2 :]
+    return torch.cat((-x2, x1), dim=-1)
+
```
parambole, line 83: Remove extra line.

tests/check_qwen3_next_vs_reference.py
```diff
@@ -179,6 +219,37 @@ def forward(self, x):
     output = output * (1.0 + self.weight.float())
     return output.type_as(x)
 
+class Qwen3NextRotaryEmbedding_PT(nn.Module):
```
parambole, line 247: @Rohan-Bierneni - can you point me to the equivalent of this in the JAX and qwen3.py ?

tests/check_qwen3_next_vs_reference.py
```diff
@@ -393,6 +464,86 @@ def forward(self, hidden_states):
...
+      query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+      if past_key_values is not None:
+          key_states = torch.cat([past_key_values[0], key_states], dim=2)
+          value_states = torch.cat([past_key_values[1], value_states], dim=2)
+
+      key_states = repeat_kv(key_states, self.num_key_value_groups)
+      value_states = repeat_kv(value_states, self.num_key_value_groups)
+
+      attn_weights = torch.matmul(query_states, key_states.transpose(2, 3)) # * self.scaling
```
parambole: Please add a comment on why the scaling factor is not included here.

parambole (commented):
@Rohan-Bierneni I have left a few comments. Also were you also able to verify how the `Qwen3NextDecoderLayer` behaves now with `GatedAttention`?

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Removed caching logic from gated full attention implementation.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Resolved.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Removed debug statements

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +560,187 @@ def __init__(
     self.layer_idx = layer_idx
     self.quant = quant
     cfg = self.config
-
-    # TODO(rbierneni): Implement the actual Qwen3NextAttention logic.
-    # This is a placeholder. The actual Qwen3NextAttention in the PyTorch code has:
-    # 1.  q_proj projection: hidden_size -> num_attention_heads * head_dim * 2.
```
shuningjin, line 568: It seems to me, the main change is
- gate
- Qwen3NextRMSNorm can be incoportated like this
https://github.com/AI-Hypercomputer/maxtext/blob/58dba0a78f8ca34e20a778892b2ce4409f2067d8/src/MaxText/layers/attentions.py#L482
- the query projection dimension? `q_proj projection: hidden_size -> num_attention_heads * head_dim * 2`

Other things might be re-used
- key proj, value proj, out proj can be reused in attention
- since you are using `attentions.RotaryEmbedding`, rope can be re-used in attention

shuningjin (commented):
Is it possible to reuse `attentions.Attention`? If the changes are relatively small, can you directly modify that file and have some if/else? 

https://github.com/AI-Hypercomputer/maxtext/blob/58dba0a78f8ca34e20a778892b2ce4409f2067d8/src/MaxText/layers/attentions.py#L231

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Thanks for catching this! There was a bug in my testcase which was causing the output of DenseGeneral to be a 3D tensor, with shape (B, S, E) instead of (B, S, H, D). The in_features to the linear layer was recieving the wrong input in the testcase and has now been updated. I also was able to get the testcase to pass using exact code from pytorch impl for the attention mechanism. Now gating is applied correctly to D and working as intended.

Rohan-Bierneni: Removed these lines since output of DenseGeneral is now (B, S, H, D) no need for the reshaping.

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: Imported the eager_attention function and used dircet torch code in testcase. Also modified it slightly so we don't need all of its functionality. Now as a result was able to remove all of these imports.

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: This is the equivalent code in qwen3.py:



    self.rotary_dim = int(cfg.head_dim * self.partial_rotary_factor)
    if self.rotary_dim > 0:
      self.rotary_embedding = attentions.RotaryEmbedding(
        min_timescale=cfg.rope_min_timescale,
        max_timescale=cfg.rope_max_timescale,
        embedding_dims=self.rotary_dim,
        fprop_dtype=cfg.dtype,
        rngs=rngs,
      )
    else:
      self.rotary_embedding = None


Note that we manually apply the partial ROPE logic and just call `attentions.RotaryEmbedding` on the first 25% of head_dim.

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: Actually we should include the scaling factor here, but perform the scaling before passing to AttentionOp in the jax implementation. Otherwise our final implementation for the gated attention will be missing a core operation of attention

Rohan-Bierneni: I added this logic before calling attention_op:

```
scaling = jax.lax.rsqrt(jnp.array(cfg.head_dim, dtype=cfg.dtype))
query_states = query_states * scaling
```

src/MaxText/layers/qwen3.py
Rohan-Bierneni: I created a new class Qwen3NextRotaryEmbedding in embeddings.py and moved the logic there. Now in qwen3.py, the __call__ function only has this 2 lines to perform partial rope:

```
# Apply Rotary Embeddings
    query_states = self.rotary_embedding(query_states, decoder_positions)
    key_states= self.rotary_embedding(key_states, decoder_positions)
```

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: This code is now moved to embeddings.py as part of the Qwen3NextRotaryEmbedding class

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Taking a look at this and should be possible to use Attention instead of attentionop. I moved the partial RotaryEmbedding logic to embeddings.py but should not make a difference.

src/MaxText/layers/embeddings.py
```diff
@@ -353,6 +353,104 @@ def llama_rotary_embedding_as_linen(
...
+        is packed.
+
+    Returns:
+      A jax.Array of shape [B, S, H, D - rotary_dim] with rotary position embeddings applied.
+    """
+    # ensure input is 4D
+    if len(inputs.shape) != 4:
+      raise ValueError("Input is assumed to be a rank 4 tensor of shape [B, S, N, H].")
+    if inputs.shape[3] != self.head_dim:
+      raise ValueError(f"Input hidden dim {inputs.shape[3]} does not match configured head dim {self.head_dim}")
```
parambole: @Rohan-Bierneni - I am not sure what would cause such a condition. Can you please elaborate ?

src/MaxText/layers/qwen3.py
```diff
@@ -551,8 +552,6 @@ def __call__(self, hidden_states: Array) -> Array:
 
 
 class Qwen3NextFullAttention(nnx.Module):
-  """Placeholder for Qwen3-Next full attention."""
```
parambole, line 554: @Rohan-Bierneni - Can you please add a docstring ?

src/MaxText/layers/qwen3.py
```diff
@@ -562,66 +561,154 @@ def __init__(
     self.layer_idx = layer_idx
     self.quant = quant
     cfg = self.config
+    self.partial_rotary_factor = 0.25
```
parambole: @Rohan-Bierneni - I see that we are interchangeably using a constant and value from config. Can we please make this configurable here and elsewhere ( with an appropriate default ). 

Thanks!

src/MaxText/layers/embeddings.py
```diff
@@ -353,6 +353,104 @@ def llama_rotary_embedding_as_linen(
...
+        min_timescale=min_timescale,
+        max_timescale=max_timescale,
+        embedding_dims=self.rotary_dim,
+        cast_as_fprop_dtype=cast_as_fprop_dtype,
+        fprop_dtype=fprop_dtype,
+        rngs=rngs,
+    )
+
+  def __call__(self, inputs: jax.Array, position: None | jax.Array = None) -> jax.Array:
+    """Applies LLaMA variant of rotary position embedding.
```
parambole, line 457: > LLaMA variant of rotary

Can you please elaborate ?

src/MaxText/layers/embeddings.py
Rohan-Bierneni: I had these checks to make sure the entire head was being passed as input, as it would be split inside the class in the call to RotaryEmbedding(). I'll remove it since the tests pass now.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Added this docstring:
```
"""Qwen3-Next Full Attention Layer.

  This module implements the full self-attention mechanism as used in
  Qwen3-Next models for layers that do not use the Gated Delta Network.
  It wraps the main `attentions.attention_op` class, which is used to perform the core attention operation:
  
  Qwen3 Next Attention differs from standard attention by the following features:
    - Query and Gate splitting from a single q projection.
    - Application of a sigmoid gate to the attention output.
    - Usage of `Qwen3NextRMSNorm` for query and key normalization.
    - Usage of `Qwen3NextRotaryEmbedding` for partial rotary position embeddings.
      - Partial ROPE is applied to the first 25% of head dimensions

  Attributes:
    config: MaxText configuration object.
    mesh: The device mesh for sharding.
    model_mode: The operational mode (e.g., 'train', 'prefill').
    layer_idx: The index of the current layer.
    quant: Optional quantization configuration.
    q_proj: Dense layer for query and gate.
    k_proj: Dense layer for key.
    v_proj: Dense layer for value.
    o_proj: Dense layer for output projection.
    q_norm: RMSNorm for queries.
    k_norm: RMSNorm for keys.
    rotary_embedding: Partial rotary embedding module.
    attention_op: The core attention calculation module.
  """
```

src/MaxText/layers/normalizations.py
```diff
@@ -68,6 +69,41 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
...
+    self.eps = eps
+    self.dtype = dtype
+    self.weight_dtype = weight_dtype
+
+    self.weight = nnx.Param(linen_initializers.zeros(rngs.params(), (self.num_features,), self.weight_dtype))
+
+  def __call__(self, x: Array) -> Array:
+    """Applies RMSNorm to the input tensor."""
+    weight = self.weight.value
+    x_dtype = x.dtype
```
shuningjin: use `self.dtype`?

src/MaxText/layers/normalizations.py
```diff
@@ -68,6 +69,41 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
...
+      and applies it multiplicatively (`y * scale`).
+  """
+
+  def __init__(self, num_features: int, eps: float, dtype: DType, weight_dtype: DType, *, rngs: nnx.Rngs):
+    self.num_features = num_features
+    self.eps = eps
+    self.dtype = dtype
+    self.weight_dtype = weight_dtype
+
+    self.weight = nnx.Param(linen_initializers.zeros(rngs.params(), (self.num_features,), self.weight_dtype))
```
shuningjin: use `self.scale` as in RMSNorm. consistent name is easier for checkpoint

nit: nn.initializers.zeros (as the above uses nn.initializers.ones)

src/MaxText/layers/qwen3.py
```diff
@@ -563,67 +553,153 @@ def __init__(
...
+    # Apply QK Normalization (per-head)
+    query_states = self.q_norm(query_states)
+    key_states = self.k_norm(key_states)
+
+    # Apply Rotary Embeddings
+    query_states = self.rotary_embedding(query_states, decoder_positions)
+    key_states= self.rotary_embedding(key_states, decoder_positions)
+
+    # Apply scaling before attention op
+    scaling = jax.lax.rsqrt(jnp.array(cfg.head_dim, dtype=cfg.dtype))
```
shuningjin: just use `cfg.head_dim**-0.5`?

tests/check_qwen3_next_vs_reference.py
```diff
@@ -28,6 +30,8 @@
 from jax.sharding import Mesh
 from flax import nnx
 
+from collections.abc import Callable
```
shuningjin: nit: remove unused

tests/check_qwen3_next_vs_reference.py
```diff
@@ -179,6 +238,37 @@ def forward(self, x):
...
+      super().__init__()
+      self.dim = int(config.head_dim * config.partial_rotary_factor)
+      self.max_seq_len_cached = config.max_position_embeddings
+      self.base = config.rope_theta
+      inv_freq = 1.0 / (self.base ** (torch.arange(0, self.dim, 2).float().to(device) / self.dim))
+      self.register_buffer("inv_freq", inv_freq, persistent=False)
+      self._set_cos_sin_cache(seq_len=self.max_seq_len_cached, device=self.inv_freq.device,
+                              dtype=torch.get_default_dtype())
+
+  def _set_cos_sin_cache(self, seq_len, device, dtype):
```
shuningjin: I didn't see this in reference. Could you clarify?
https://github.com/huggingface/transformers/blob/a9731a725eb1d7b3b7e11f0ad35a819fa4ee8b20/src/transformers/models/qwen3_next/modeling_qwen3_next.py#L176-L209

Is it related to KVCache? Current Qwen3NextFullAttention does not have cache though.

If possible, try to keep the reference code as is. Otherwise, be explicit in the class comment what you have modified.

tests/check_qwen3_next_vs_reference.py
```diff
@@ -848,6 +1015,131 @@ def run_jax(x):
...
+      rtol = 1e-6
+      atol = 1e-6
+      is_close = np.allclose(pt_out_np, jax_out_np, rtol=rtol, atol=atol)
+      num_elements = pt_out_np.size
+      mismatched_elements = np.sum(~np.isclose(pt_out_np, jax_out_np, rtol=rtol, atol=atol))
+      percent_mismatch = (mismatched_elements / num_elements) * 100
+
+      print(f"  Max Abs Diff: {max_abs_diff:.2e}, Mean Abs Diff: {mean_abs_diff:.2e}")
+      print(f"  Mismatched Elements: {mismatched_elements}/{num_elements} ({percent_mismatch:.2f}%)")
+      print(f"  Is Close (rtol={rtol}, atol={atol}): {is_close}")
```
shuningjin: since these are for temporary debugging, probably remove these before final merge?

src/MaxText/layers/qwen3.py
```diff
@@ -563,67 +553,153 @@ def __init__(
...
+        min_timescale=cfg.rope_min_timescale,
+        max_timescale=cfg.rope_max_timescale,
+        embedding_dims=cfg.head_dim,
+        partial_rotary_factor=cfg.partial_rotary_factor,
+        cast_as_fprop_dtype=True,
+        fprop_dtype=cfg.dtype,
         rngs=rngs,
+      )
+    else: 
+      self.rotary_embedding = None
```
shuningjin: When rotary_dim=0, how is this applied in `__call__`? Should it raise error, or just use identity function?

shuningjin (commented):
Thank you for the good work! I left some minor comments.

src/MaxText/layers/embeddings.py
```diff
@@ -353,6 +353,98 @@ def llama_rotary_embedding_as_linen(
...
+        added signal.
+      embedding_dims: Dimension of the embedding to be generated.
+      partial_rope_factor: Ratio of dimensions to apply ROPE to
+      rngs: rng keys passed in by nnx.bridge.to_linen.
+    """
+    self.head_dim = embedding_dims
+    self.partial_rotary_factor = partial_rotary_factor
+    self.rotary_dim = int(self.head_dim * self.partial_rotary_factor)
+
+    if self.rotary_dim % 2 != 0:
```
RissyRan: Shall we also have an initial assertion in pyconfig file for early return?

src/MaxText/configs/base.yml
```diff
@@ -881,6 +881,8 @@ gdn_num_value_heads: 32
 gdn_chunk_size: 64
 # Whether to apply L2 normalization to query and key tensors inside the Gated Delta Rule kernel.
 use_qk_norm_in_gdn: True
+# The ratio of dimension to apply ROPE on
+partial_rotary_factor: 0.25
```
RissyRan: Shall we set this as 0 or 1 by default? and Qwen3-next config as 0.25?

src/MaxText/layers/embeddings.py
```diff
@@ -353,6 +353,98 @@ def llama_rotary_embedding_as_linen(
...
+    Args:
+      inputs: The input sequence on which to apply the Rotary position
+        embedding. It is assumed of shape [B, S, H, D].
+      position: Optional position array [B, S]. Only needed when the sequence
+        is packed.
+
+    Returns:
+      A jax.Array of shape [B, S, H, D - rotary_dim] with rotary position embeddings applied.
+    """
+    if self.rotary_dim > 0:
```
RissyRan: Is this value always > 0? 

head_dim = 256 & partial_rotary_factor=0.25? If so, shall we remove `if self.rotary_dim > 0:` in Qwen3NextRotaryEmbedding?

src/MaxText/layers/normalizations.py
RissyRan: I may miss this part. Wondering if this 0 init values are on purpose, i.e. same as reference. If not, shall we leverage existing https://github.com/AI-Hypercomputer/maxtext/blob/a8499ddc059bb2daf4c123a6c65014bc62070f50/src/MaxText/layers/normalizations.py#L53-L55?

The init values are important for training model from scratch and we met customer issue before.

src/MaxText/layers/normalizations.py
```diff
@@ -68,6 +69,41 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
 
     scale = jnp.asarray(scale, self.dtype)
     return y * scale
+  
+class Qwen3NextRMSNorm(nnx.Module):
```
RissyRan: Thanks for the update!

src/MaxText/layers/qwen3.py
```diff
@@ -281,43 +282,6 @@ def scan_body(prev_state, x):
...
-    x_dtype = x.dtype
-    x = x.astype(jnp.float32)
-    variance = jnp.mean(jnp.square(x), axis=-1, keepdims=True)
-    x = x * jax.lax.rsqrt(variance + self.eps)
-    # Add 1.0 to the learnable weight
-    x = x * (1.0 + weight.astype(jnp.float32))
-    return x.astype(x_dtype)
-
-
 class Qwen3NextRMSNormGated(nnx.Module):
```
RissyRan: Can we also move this to normalizations.py? I think I left a comment on Param's PR, and haven't followed it up due to OOO.

src/MaxText/layers/qwen3.py
```diff
@@ -551,8 +515,34 @@ def __call__(self, hidden_states: Array) -> Array:
 
 
 class Qwen3NextFullAttention(nnx.Module):
-  """Placeholder for Qwen3-Next full attention."""
+  """Qwen3-Next Full Attention Layer.
```
RissyRan, line 471: If we decide to write a new full attn, let's move to src/MaxText/layers/attentions.py or have a separate attn like src/MaxText/layers/attention_mla.py?

Personally, I would prefer to cooperate into existing attentions.py file, to avoid duplicate code blocks.

tests/check_qwen3_next_vs_reference.py
```diff
@@ -416,6 +577,7 @@ def setUp(self):
             "weight_dtype=float32",
             "matmul_precision=highest",
             "decoder_block=qwen3_next",
+            "attention=dot_product",
```
RissyRan, line 630: Shall we have assertion in pyconfig file and indicate flash attn is not supported yet?  If supported, can we also test it?

parambole: @Rohan-Bierneni - This should support flash attention?

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: Resolved.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Based on internal discussion, will integrate this with Attention() class so will remove this code.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Integrated successfully with Attentions() class but had to add conditionals to route model-specific logic for gating mechanism, partial ROPE, and Qwen3NextRMSNorm.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: For now, integrated the Gated FullAttention with existing attentions.py file, and @parambole is working on porting the Gated Delta Net to a separate attention file. This change will be in a separate PR.

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: The Gated FullAttention layer should be supported with splash attention, especially since now we integrated it with main attentions.py file. Will try running with attention="splash" and update.

tests/check_qwen3_next_vs_reference.py
Rohan-Bierneni: I have removed this code block and only included pytorch ref code for the default ROPE computation. The code I included is this class: https://github.com/huggingface/transformers/blob/02c324f43fe0ef5d484e846417e5f3bf4484524c/src/transformers/models/qwen3_next/modular_qwen3_next.py#L184. Also, since in transformers repo Qwen3NextRotaryEmbedding inherits from Gemma2RotaryEmbedding, I also included the code from there: https://github.com/huggingface/transformers/blob/02c324f43fe0ef5d484e846417e5f3bf4484524c/src/transformers/models/gemma2/modeling_gemma2.py#L86.

Note that I removed the option to config for different ROPE types (default, yarn, linear, etc) and only included the default implementation.

src/MaxText/layers/normalizations.py
Rohan-Bierneni: The reference implementation is also initializing to zeroes (Qwen3NextRMSNorm use Gemma3RMSNorm existing impl): https://github.com/huggingface/transformers/blob/02c324f43fe0ef5d484e846417e5f3bf4484524c/src/transformers/models/gemma3/modeling_gemma3.py#L128. Since the ref implementation does this, I will keep it as is. @parambole is this fine?

src/MaxText/layers/attentions.py
```diff
@@ -642,6 +665,21 @@ def qkv_projection(self, inputs: Array, proj_name: str, out_sharding: NamedShard
...
   def init_out_w(self, output_dim: int) -> nnx.Module:
     """out projection"""
+    if self.isQwen3Next:
+      out_kernel_axis = ("mlp","embed")
+      return DenseGeneral(
+          in_features_shape=(self.num_query_heads * self.head_dim),
+          out_features_shape=output_dim,
+          axis=(-1,),
+          kernel_init=self.kernel_init,
+          kernel_axes=out_kernel_axis,  # trade speed with memory
```
RissyRan: nit: directly apply`("mlp","embed")` to avoid confusion for line #683

src/MaxText/pyconfig.py
```diff
@@ -286,6 +286,9 @@ def validate_keys(keys):
       raise ValueError(
           "For Qwen3-Next, sparse_matmul must be False for now. The dense path has been verified against reference."
       )
+    rotary_dim = int(keys["head_dim"] * keys["partial_rotary_factor"])
```
RissyRan: Sorry I missed this one in previous review. Can we wrap 

`if keys["decoder_block"] == "qwen3_next":` into a helper function, and put inside of this? https://github.com/AI-Hypercomputer/maxtext/blob/57e0ece219f3c4016fc41c4a3b27f499824f575d/src/MaxText/pyconfig.py#L262

src/MaxText/layers/qwen3.py
```diff
@@ -282,88 +283,6 @@ def scan_body(prev_state, x):
...
-    normalized_states = hidden_states_f32 * jax.lax.rsqrt(variance + self.eps)
-    normalized_states = normalized_states * weight.astype(jnp.float32)
-
-    # Gated Activation using SiLU (Sigmoid-weighted Linear Unit)
-    gated_states = normalized_states * jax.nn.silu(gate.astype(jnp.float32))
-
-    return gated_states.astype(self.dtype)
-
-
 class Qwen3NextGatedDeltaNet(nnx.Module):
```
RissyRan, line 284: Did we mention to move this in a separate file? I am fine with next PR

src/MaxText/configs/base.yml
```diff
@@ -882,6 +882,8 @@ gdn_num_value_heads: 32
 gdn_chunk_size: 64
 # Whether to apply L2 normalization to query and key tensors inside the Gated Delta Rule kernel.
 use_qk_norm_in_gdn: True
+# The ratio of dimension to apply ROPE on
+partial_rotary_factor: 1.0
```
RissyRan, line 909: Nit: It seems this is only used in Qwen3 right? let's add an assertion in pyconfig.py file to indicate this is only used in qwen3-next block? Otherwise, customers set this value, but actually doesn't take effect.

RissyRan (commented):
A few minor comments.

RissyRan:
Oh, I think you will need to squash those 26 commits into 1

src/MaxText/layers/attentions.py
```diff
@@ -498,9 +504,21 @@ def __init__(
           kernel_axes=("norm",),
```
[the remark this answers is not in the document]
RissyRan: +1

src/MaxText/layers/normalizations.py
```diff
@@ -72,9 +73,90 @@ def __call__(self, x: jnp.ndarray, out_sharding: NamedSharding | None = None) ->
...
+  Attributes:
+    num_features: The number of features in the input.
+    eps: A small epsilon value to prevent division by zero in RMSNorm.
+    dtype: The datatype of the computation.
+    weight_dtype: The datatype of the weights.
+  """
+
+  def __init__(self, num_features: int, eps: float, dtype: DType, weight_dtype: DType, *, rngs: nnx.Rngs):
+    self.num_features = num_features
+    self.eps = eps
```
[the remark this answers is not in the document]
RissyRan: I recall @shuningjin also mentioned this before?

src/MaxText/layers/attention_op.py
```diff
@@ -33,8 +33,8 @@
 from jax.experimental.pallas.ops.tpu.splash_attention import splash_attention_kernel
 from jax.experimental.pallas.ops.tpu.splash_attention import splash_attention_mask
 
-from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_kernel as tokamax_splash_kernel
-from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_mask as tokamax_splash_mask
+# from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_kernel as tokamax_splash_kernel
+# from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_mask as tokamax_splash_mask
```
shuningjin: should keep?

src/MaxText/layers/attentions.py
```diff
@@ -416,6 +417,8 @@ def __init__(
     self.model_mode = model_mode
     self.rngs = rngs
 
+    self.isQwen3Next = self.config.decoder_block == DecoderBlockType.QWEN3_NEXT
```
shuningjin: nit: self.isQwen3Next -> self.is_qwen3_next

src/MaxText/layers/attentions.py
```diff
@@ -890,11 +940,19 @@ def __call__(
       value_sharding = NamedSharding(self.mesh, nn.logical_to_mesh_axes(self.value_axis_names))
       value = self.kv_projection(inputs_kv, proj_name="value", out_sharding=value_sharding)
 
+    gate = None
```
shuningjin, line 936: nit: remove?

src/MaxText/layers/attentions.py
```diff
@@ -964,7 +1022,9 @@ def __call__(
           bidirectional_mask,
           self.sinks,
       )
-
+    if self.isQwen3Next:
+      out = out.reshape(batch_size, seq_len, self.config.num_query_heads * self.config.head_dim)
+      out = out * nn.sigmoid(gate)
```
shuningjin: nit: jax.nn.sigmoid (nn is linen here)

src/MaxText/layers/moe.py
```diff
@@ -36,7 +36,7 @@
 from MaxText.layers import attentions, linears, quantizations, nnx_wrappers
 from MaxText.layers.initializers import NdInitializer, nd_dense_init, default_bias_init, variable_to_logically_partitioned
 
-from tokamax._src.ops.ragged_dot import api as tokamax_api
+# from tokamax._src.ops.ragged_dot import api as tokamax_api
```
shuningjin: keep?

src/MaxText/layers/attentions.py
```diff
@@ -538,9 +556,14 @@ def query_init(*args):
     kernel_axes = (
         (None, None, None) if self.config.ici_context_autoregressive_parallelism > 1 else ("embed", "q_heads", "kv")
     )
+    in_features = self.convert_dense_general_inputs_shape(inputs_q_shape)
+    out_features = (self.num_query_heads, self.head_dim)
+    if self.isQwen3Next:
+      out_features = (self.num_query_heads, self.head_dim * 2)
```
shuningjin, line 563: nit: if / else might be clearer

shuningjin (approved):
Thank you! LGTM.

src/MaxText/layers/attentions.py
Rohan-Bierneni: Refactored code to use only 1 Densegeneral and change configs if using qwen3next decoder block.

src/MaxText/layers/normalizations.py
```diff
@@ -72,9 +73,90 @@ def __call__(self, x: jnp.ndarray, out_sharding: NamedSharding | None = None) ->
     if self.shard_mode != ShardMode.EXPLICIT:
       out_sharding = None
 
-    scale = jnp.asarray(scale, self.dtype)
+    scale = jnp.asarray(scale, self.dtype)
```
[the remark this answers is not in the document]
Rohan-Bierneni: Resolved.

tests/check_qwen3_next_vs_reference.py
```diff
@@ -415,7 +608,9 @@ def setUp(self):
             "dtype=float32",
```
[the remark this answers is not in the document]
Rohan-Bierneni: Resolved.

Rohan-Bierneni:
Manually adding pull ready tag since all tests pass, have 3 approvals, and all comments are resolved. Seems to be bug with a skipped check causing the tag to not be added.

PR #2541, opened 2025-10-24
Refactor: 3.1 phase of RESTRUCTURE.md, focussing on setup.sh and related Python dependency & Dockerfile files

SamuelMarks:
# Description

Refactor: 3.1 phase of RESTRUCTURE.md, focussing on setup.sh and related Python dependency & Dockerfile files

Note: merge this before #2361; after which that PR is roughly halved.

# Tests

CI and manual:
```sh
$ bash ./dependencies/scripts/docker_build_dependency_image.sh DEVICE='tpu' MODE='nightly'
```
```sh
$ bash ./dependencies/scripts/docker_build_dependency_image.sh DEVICE='tpu' MODE='stable'
```

```sh
$ export MODEL_NAME='llama3_1_70b_8192_synthetic' \
         PROJECT="${GOOGLE_CLOUD_PROJECT?}" \
         ZONE="${GOOGLE_CLOUD_ZONE?}" \
         CLUSTER_NAME="${GOOGLE_CLOUD_CLUSTER_NAME?}" \
         OUTPUT_DIR="${GOOGLE_CLOUD_BUCKET?}" \
         BASE_OUTPUT_DIR="${GOOGLE_CLOUD_BUCKET?}"'/output/' \
         DATASET_PATH="${GOOGLE_CLOUD_BUCKET?}"'/' \
         WORKLOAD='job_name_goes_here'

$ python3 -m MaxText.train MaxText/configs/base.yml \
      run_name="${USER}"'002' \
      base_output_directory="${OUTPUT_DIR?}" \
      dataset_type='synthetic' \
      steps='10' \
      model_name='llama2-7b'

$ command='python3 -m MaxText.train MaxText/configs/base.yml
      base_output_directory='"${BASE_OUTPUT_DIR?}"'
      dataset_path='"${DATASET_PATH?}"'
      steps=100
      per_device_batch_size=1'

$ xpk workload create \
      --base-docker-image 'maxtext_base_image' \
      --zone "${ZONE?}"
      --cluster "${CLUSTER_NAME?}" \
      --workload "${WORKLOAD?}" \
      --tpu-type='v6e-256' \
      --num-slices='1' \
      --command "${command?}"

$ python3 -m benchmarks.benchmark_runner xpk \
      --base_docker_image='maxtext_base_image' \
      --project="${PROJECT?}" \
      --zone="${ZONE?}" \
      --cluster_name="${CLUSTER_NAME?}" \
      --device_type='v6e-256' \
      --num_slices='1' \
      --base_output_directory="${OUTPUT_DIR?}" \
      --model_name="${MODEL_NAME?}"
```

# Checklist


SamuelMarks:
> @SamuelMarks discussed offline - can you please include the manual benchmark runner tests you ran in the PR description?

Are my tests in the PR description sufficient, or do you want more models+configurations tested?

.github/workflows/build_and_upload_images.sh
```diff
@@ -24,7 +24,16 @@
...
 # bash build_and_upload_images.sh PROJECT=<project> MODE=stable DEVICE=tpu CLOUD_IMAGE_NAME=${USER}_runner
 
+if [ "${BASH_SOURCE-}" ]; then
+  this_file="${BASH_SOURCE[0]}"
+elif [ "${ZSH_VERSION-}" ]; then
+  # shellcheck disable=SC2296
+  this_file="${(%):-%x}"
+else
+  this_file="${0}"
+fi
```
bvandermoon, line 34: Why is this needed? Can we remove this?

src/install_maxtext_extra_deps/install_github_deps.py
```diff
@@ -1,91 +0,0 @@
...
-
-This script is designed to be run to install dependencies specified in
-'extra_deps_from_github.txt', which is expected to be in the same directory.
-It first ensures 'uv' is installed and then uses it to install the packages
-listed in the requirements file.
-"""
-
-import subprocess
-import sys
-from pathlib import Path
```
bvandermoon, line 25: Why is this file deleted? Don't we still need this?

SamuelMarks: No we don't need this, all we need is the .txt file and instructions to call `-r` on it. This replaces the unfamiliar previous execution of a Python file/module. Now it's just a regular .txt file that makes sense—is expected—in ones workflow.

.github/workflows/build_and_upload_images.sh
SamuelMarks: No we need it otherwise the script doesn't know where it is. In Python you have `__file__` magic; this is how you get similar in shell.

Paths relative to this enable always-correct specification of [later] dockerfile location.

src/install_maxtext_extra_deps/install_github_deps.py
bvandermoon: @SamuelMarks I thought this was still needed for PyPI since we can't upload non-PyPI dependencies?

@kanglant can you please take a look also?

SamuelMarks: Yeah but the Python file doesn't do anything that `pip install -r <txt_file>` doesn't. And that command is a regular part of peoples workflow, so no disadvantage in just making that the recommendation.

It's weird to have to execute a Python script that calls Popen on pip

bvandermoon: Would these [installation instructions](https://github.com/AI-Hypercomputer/maxtext?tab=readme-ov-file#installation) need to change? Ideally we should be trying to reduce the number of reasons to require cloning MaxText. My understanding is this would be another reason it is required

kanglant: The external users will need to run `uv pip install -r https://raw.githubusercontent.com/AI-Hypercomputer/maxtext/refs/heads/main/dependencies/requirements/extra_deps_from_github.txt` instead.

kanglant:
Why do we need to relocate the Dockerfiles and the requirement files?

SamuelMarks:
> Why do we need to relocate the Dockerfiles and the requirement files?

@kanglant This was agreed upon a long time ago in the RESTRUCTURE.md file. See internal design docs for more details.

src/install_maxtext_extra_deps/install_github_deps.py
SamuelMarks: Yeah and I documented that. It's always one extra line, only now it's a clearer line and you can execute easily without cloning the whole MaxText repo.

bvandermoon: Can't this file get out of sync with the PyPI version? Since it is installing from Github HEAD

SamuelMarks: Yeah but we don't have tags [`git tag`] currently. When we start having tagged releases this'll be trivial to do correctly; I don't want to have magic numbers in the codebase (like a random commit hash that people will forget to update).

https://github.com/AI-Hypercomputer/maxtext/tags

kanglant: We should add a tag during each MaxText release (e.g., `maxtext-v0.1.0`). Let's add a note in the `README.md` now, mentioning that users can download GitHub dependencies using a versioned path like `https://github.com/.../blob/maxtext-v<version>/dependencies/...`. We can then fully update the `README.md` to reflect this final behavior in the next MaxText pypi release.

kanglant: For users installing maxtext 0.1.0 from PyPI, the README.md instructs them to continue using `install_maxtext_github_deps`. This will be updated in the next release to: `uv pip install -r https://github.com/.../blob/maxtext-v<maxtext_version>/dependencies/...`.

bvandermoon: I'm not following why this is needed though. IMO it is easier for the customer to just run the existing workflow which is provided when they pip install maxtext. If we make this change, they will need to take an extra step to find the right tag and track down a longer URL for the command

SamuelMarks: `git tag` is the release approach recommended by goreleaser, what I've been pushing for. `/blob/{PYPI_VERSION}` should be the scheme, as this is the most common approach.

Cleanest way of handling dependencies whence pypi restricts github deps.

bvandermoon: Discussed offline with @kanglant and @SamuelMarks. We will go back to the old way of installation for this

bvandermoon:
> > Why do we need to relocate the Dockerfiles and the requirement files?
> 
> @kanglant This was agreed upon a long time ago in the RESTRUCTURE.md file. See internal design docs for more details.

@SamuelMarks can we still have the base_requirements and generated_requirements structure within the new directory? These didn't exist at the time but it would help with organization to keep them

tools/setup/setup.sh
```diff
@@ -129,7 +129,6 @@ if [[ $DEVICE == "tpu" ]]; then
             python3 -m uv pip install -U crcmod
             # Copy libtpu.so from GCS path
             gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
-            exit 0
```
bvandermoon, line 132: Why is this removed?

tools/setup/setup.sh
```diff
@@ -138,48 +137,68 @@ if [[ $DEVICE == "tpu" ]]; then
 fi
 
 if [[ "$MODE" == "nightly" ]]; then
-    echo "Nightly mode: Installing requirements.txt, stripping commit pins from git+ repos."
+    if [ "$DEVICE" = "gpu" ]; then
+        dep_name='dependencies/requirements/generated_requirements/cuda12-requirements.txt'
+    else
+        dep_name='dependencies/requirements/generated_requirements/'"${DEVICE?}"'-requirements.txt'
+    fi
```
bvandermoon, line 145: Why is this added here?

tools/setup/setup.sh
SamuelMarks: Keeping it in will exit the whole script. That's a bug. This script isn't called "setup_libtpu.sh"; so that branch should only ever early-exit on failure.

tools/setup/setup.sh
SamuelMarks: To branch betwixt accelerator type. The old dependency file is not meant to be used.

.github/workflows/build_and_upload_images.sh
kanglant: Could you please clarify a couple of points regarding this script?

1. Is the script still needed? I'm not sure if the script `.github/workflows/build_and_upload_images.sh` is still required, as the `UploadDockerImages.yml` workflow seems to have been updated to use `docker/build-push-action` instead of calling this shell script. Could you confirm if this script is still active?
2. Necessity of `MAXTEXT_REPO_ROOT`: I understand that the logic sets `MAXTEXT_REPO_ROOT` to the absolute path of the repository root (which is two levels up to this file). However, in the old `UploadDockerImages.yml`, the script was called from the project root (e.g., `bash .github/workflows/build_and_upload_images.sh ...`), meaning the working directory was already at the repository root. If the working directory is consistently at the project root, we should be able to use relative paths like `dependencies/scripts/docker_build_dependency_image.sh` directly. Could you explain why you felt setting the `MAXTEXT_REPO_ROOT` variable was necessary in this context?

bvandermoon: For (1), yes this script is still needed because some users run it manually

@SamuelMarks for (2)

kanglant: If people are still running this script manually, setting the `MAXTEXT_REPO_ROOT` with the absolute path would be helpful. This makes the script runnable from any directory, as it no longer relies on the user's current working directory to locate necessary files.

The only question is how many users are currently running this script manually. Perhaps we should also consider moving this script to the `dependencies` folder?

bvandermoon: Makes sense, thanks both. Agree we should move it to the `dependencies/scripts` folder. I was thinking of the `docker_upload_runner.sh` script when I was saying it is still run manually. I am not sure if this `build_and_upload_images.sh` script is still used or not. But agree for now we can move it to the same location as the other two scripts

tools/setup/setup.sh
bvandermoon: Won't the rest of the script override the custom libtpu here though? Since this is handling `MODE=libtpu-only`

.github/workflows/build_and_upload_images.sh
SamuelMarks: More files can be moved and more refactoring can happen. Is there some pressing reason this needs to happen in this PR, and not in #2361 or a follow-on PR?

tools/setup/setup.sh
SamuelMarks: I can add an `unset LIBTPU_GCS_PATH` here instead of the removed `exit 0`; but it's not a big deal as the later occurrences do the same copy (yes it would be good to avoid the additional network transfer(s)).

.github/workflows/build_and_upload_images.sh
bvandermoon: Let's just leave this file here for now. Looking at the code comments, it looks like it is not intended to be used directly by customers and was added specifically for the Github runners

tools/setup/setup.sh
bvandermoon: Looks like this exit was added intentionally. Sent over a doc offline:

```
For libtpu-only mode, LIBTPU_GCS_PATH is required. It is used for updating custom libtpu without re-install jax and jaxlib.
```

So the flow is to first install all dependencies, then run it again with `MODE=libtpu-only`. Not saying this is the right pattern or not, but let's leave it as is. The way this is set up in this PR, the custom libtpu will be overwritten by the later JAX/libtpu installation

SamuelMarks: Ok

maxtext_runner.Dockerfile
```diff
@@ -16,7 +16,6 @@ WORKDIR /deps
 # Copy assets separately
 COPY src/MaxText/assets/ "${MAXTEXT_ASSETS_ROOT}"
 COPY src/MaxText/test_assets/ "${MAXTEXT_TEST_ASSETS_ROOT}"
-COPY generated_requirements .
```
bvandermoon, line 19: Was this removed intentionally? I guess it just isn't used?

tools/setup/setup.sh
```diff
@@ -138,48 +138,68 @@ if [[ $DEVICE == "tpu" ]]; then
...
+                                                    -r 'dependencies/requirements/extra_deps_from_github.txt'
     fi
 fi
 
 # Install maxtext package
 if [ -f 'pyproject.toml' ]; then
-  python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest
-  install_maxtext_github_deps
+  case "$DEVICE" in
+      'gpu') python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest ;;
```
shralex: The TPU and GPU are inverted here -- we should install tpu dependencies for tpu and gpu dependencies for gpu.

maxtext_runner.Dockerfile
shralex: Also would like to know. Thanks!

tools/setup/setup.sh
SamuelMarks: Good find, thanks

maxtext_runner.Dockerfile
SamuelMarks: That whole folder has been moved under dependencies so it's redundant and incorrect to retain.

.github/workflows/run_tests_against_package.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           source .venv/bin/activate
           maxtext_wheel=$(ls maxtext-*-py3-none-any.whl 2>/dev/null)
           uv pip install ${maxtext_wheel}[${MAXTEXT_PACKAGE_EXTRA}] --resolution=lowest
-          install_maxtext_github_deps
+          uv pip install -r src/install_maxtext_extra_deps/extra_deps_from_github.txt
```
shralex, line 88: Can you explain this change ? 
Also, since we have dependencies / scripts folder now, do we also need this folder under src ?
CC: @bvandermoon

dependencies/dockerfiles/maxtext_jax_ai_image.Dockerfile
```diff
@@ -16,9 +16,9 @@ ENV MAXTEXT_REPO_ROOT=/deps
 WORKDIR /deps
 
 # Copy setup files and dependency files separately for better caching
-COPY setup.sh ./
-COPY requirements.txt requirements_with_jax_ai_image.txt requirements_with_jax_stable_stack_0_6_1_pipreqs.txt src/install_maxtext_extra_deps/extra_deps_from_github.txt generated_requirements ./
-
+COPY tools/setup /deps/tools/setup/
+COPY dependencies/requirements/ /deps/dependencies/requirements/
```
shralex, line 20: We no longer want to copy files separately ? 

IMO we should try not to change logic in this PR, so should continue copying files separately, which also corresponds to the code comment that is here.

src/MaxText/examples/demo_decoding.ipynb
```diff
@@ -86,7 +86,7 @@
     "\n",
     "# Install MaxText and dependencies\n",
     "!uv pip install maxtext --resolution=lowest\n",
-    "!install_maxtext_github_deps\n",
+    "!python3 -m MaxText.install_maxtext_extra_deps\n",
```
shralex, line 89: Can you explain this change ?

src/MaxText/examples/sft_llama3_demo.ipynb
```diff
@@ -109,7 +109,7 @@
         "\n",
         "# 2. Install MaxText and its dependencies\n",
         "!uv pip install maxtext --resolution=lowest\n",
-        "!install_maxtext_github_deps"
+        "!python3 -m MaxText.install_maxtext_extra_deps\n"
```
shralex, line 112: Can you explain this change ?

src/MaxText/examples/sft_qwen3_demo.ipynb
```diff
@@ -100,6 +100,7 @@
         "\n",
         "# Install MaxText and its dependencies\n",
         "!uv pip install -e .[tpu] --resolution=lowest\n",
+        "!python3 -m MaxText.install_maxtext_extra_deps\n",
```
shralex, line 103: Can you explain this change ?

shralex (commented):
Added some more comments / questions.

Given the extensive changes to docker files and setup.sh, how are we testing these two ?

src/MaxText/examples/sft_llama3_demo.ipynb
SamuelMarks: This is the correct way to execute a module. MaxText was installed in an earlier `!` command.

src/MaxText/examples/demo_decoding.ipynb
SamuelMarks: This is the correct way to execute a module. MaxText was installed in an earlier `!` command. Modules can be executed irrespective of current working directory.

src/MaxText/examples/sft_qwen3_demo.ipynb
SamuelMarks: This is the correct way to execute a module. MaxText was installed in an earlier `!` command. Modules can be executed irrespective of current working directory. Without this line the github—i..e, not pypi—dependencies won't be installed.

.github/workflows/run_tests_against_package.yml
SamuelMarks: Earlier I tried moving everything under the dependencies directory. Very happy to do that; but @bvandermoon and @kanglant are against it. Your call.

src/MaxText/examples/demo_decoding.ipynb
shralex: What I mean is - this change calls "install_maxtext_extra_deps" instead of "install_maxtext_github_deps" previously

src/MaxText/examples/sft_llama3_demo.ipynb
shralex: same -- why is the name of the module changing ?

src/MaxText/examples/sft_qwen3_demo.ipynb
shralex: What's the difference between this module and install_maxtext_github_deps ?

.github/workflows/run_tests_against_package.yml
kanglant: @shralex 

The reason this script is currently placed here is that the github dep file must be included within the MaxText Python package as **package data**. This allows users who install the package to call the `install_maxtext_github_deps` function. If we move the file to the `dependencies/` folder, it will no longer be included in the package.

If we decide to move this file to the `dependencies/` folder, the `install_maxtext_github_deps` module is no longer needed. Consequently, users (who haven't cloned the MaxText repo) would need to install GitHub dependencies directly from the MaxText GitHub source via a command like: `uv pip install -r https://github.com/.../blob/maxtext-v<maxtext_version>/dependencies/extra_deps_from_github.txt`, where `maxtext-v<maxtext_version>` is a maxtext python release tag. The downside of this approach is that PyPI users would have to manually update the `<maxtext_version>` in the URL for every new MaxText release.

kanglant: A workaround is to ask users run `MAXTEXT_VERSION=$(pip show maxtext | grep -m 1 "Version:" | awk '{print $2}')` and then construct the GitHub URL.

bvandermoon: Discussed offline, we will keep this file in /src for the reason @kanglant gave in this thread

shralex: Thanks for explaining!

dependencies/dockerfiles/maxtext_jax_ai_image.Dockerfile
SamuelMarks: They're all in distinct folders now, one that contains all the requirement (.txt) files and, the other, setup (.sh) files.

PR #2544, opened 2025-10-24
Integrate SDK for managed diagnostics

xibinliu:
# Description
Integrate SDK for managed diagnostics

- include new SDK google-cloud-mldiagnostics
   seed-env: --seed-commit=3d8c7df56aa55f363d5453fc82b64c5093ea2863
- add new config params
- modify profiler.py to add ML run and profiling
- modify metrics_logger.py to upload metrics

Four modes will be supported:
1. Only upload configs and metrics, not profiling (managed_mldiagnostics=True)
2. Upload configs and metrics, and do profiling on the first device (managed_mldiagnostics=True and profiler=xplane)
3. Upload configs and metrics, and do profiling on the all devices (managed_mldiagnostics=True, profiler=xplane, and upload_all_profiler_results=True)
4. on-demand profiling support (TODO. Not in this PR)

# IMPORTANT
Since the GCP UI support is not formally rolled out yet, currently this feature only works in supercomputer-testing / us-central1. Enabling this feature in other projects and regions will fail.

# Tests
Command:
1. Enable the feature with `managed_mldiagnostics=True`, no profiling
```
python3 -m MaxText.train src/MaxText/configs/base.yml run_name="xibin-run22" model_name="gpt3-52k" base_output_directory=gs://xibin-images/  dataset_type=synthetic steps=22 managed_mldiagnostics=True managed_profiler_run_group="xibin-demo" log_period=5
```

Note:  `managed_mldiagnostics_run_group` is optional

2. Enable the feature with `managed_mldiagnostics=True profiler=xplane` for single device profiling
```
python3 -m MaxText.train src/MaxText/configs/base.yml run_name="xibin-run23" model_name="gpt3-52k" base_output_directory=gs://xibin-images/  dataset_type=synthetic steps=22 profiler=xplane managed_mldiagnostics=True  log_period=5
```
3. Enable the feature with `managed_mldiagnostics=True profiler=xplane upload_all_profiler_results=True` to profile on all TPU devices
```
python3 -m MaxText.train src/MaxText/configs/base.yml run_name="xibin-run24" model_name="gpt3-52k" base_output_directory=gs://xibin-images/  dataset_type=synthetic steps=22 profiler=xplane managed_mldiagnostics=True upload_all_profiler_results=True  log_period=5
```

See all uploaded runs in the [managed profiler GCP UI](https://pantheon-hourly-sso.corp.google.com/cluster-director/diagnostics/list?b=hypercompute_cluster:AUTOPUSH,coliseum_backend:AUTOPUSH,coliseum_host_backendz:AUTOPUSH&e=GceDiagonUiLaunch::GceDiagonUiEnabled&mods=force_hide_emergency_banner,disable_console_nav_discovery,-is_integration_test,integration_test,enable_router_tracing,disable_learn_assistant,hide_section_nav&project=supercomputer-testing)

# Checklist


src/MaxText/profiler.py
```diff
@@ -41,6 +46,32 @@ def __init__(self, config, offset_step=0):
     if config.profiler != "" and self.start_initial_profile_step >= config.steps:
       raise ValueError("Profiling requested but initial profiling step set past training final step")
 
+    # Set up the managed profiler. Only set it up on the first device.
+    self.managed_profiler = False if jax.process_index() != 0 else config.managed_profiler
```
bvandermoon: Is this saying that the manager should only run on the first device? Or that the profiles should only be collected on the first device? I am basically wondering if there is still a way to choose whether to profile every device or just a single one

src/MaxText/pyconfig.py
```diff
@@ -41,6 +41,9 @@
 
 _MAX_PREFIX = "M_"
 
+# Don't log the following keys.
+KEYS_NO_LOGGING = ["hf_access_token"]
+
```
bvandermoon, line 43: @SamuelMarks FYI we will need this to be compatible with Pydantic in #1836. It's just a constant so should be straightforward

src/MaxText/profiler.py
```diff
@@ -41,6 +46,32 @@ def __init__(self, config, offset_step=0):
...
+        if key in KEYS_NO_LOGGING:
+          continue
+        # Work-around of the Diagon bug (b/454722730): convert all values into string.
+        string_value = str(value)
+        try:
+          # Verify the value can be serialized to json. If not, we'll skip it.
+          json.dumps(value, allow_nan=False)
+          config_dict[key] = string_value
+        except TypeError:
+          pass
```
bvandermoon: What about something like this instead? I think it's a bit cleaner and makes it more clear what's happening (Please double check the logic)

```
config_dict = { key: str(value) for key, value in config.get_keys().items() if should_log_key(key) }

def should_log_key(key):
    # logic to determine whether to log key
```

bvandermoon (commented):
LGTM just a few comments. Thanks @xibinliu, great to see this

src/MaxText/profiler.py
xibinliu: Thanks for the review. This is a good point! The SDK currently restricts the ML runs to have the same run_name, that's why I started it only on the first device.

But with a second thought, I changed the logic to append jax `process_index` to the `run_name`, if the uploading is from all hosts. With this, we expect the ML runs will be name: `run_name-0`, `run_name-1`, etc.

src/MaxText/profiler.py
xibinliu: Got it. Modified as recommended.

src/MaxText/pyconfig.py
xibinliu: Not sure what I should do for this comment. But I changed it to the tuple making it immutable.

src/MaxText/profiler.py
```diff
@@ -41,6 +46,37 @@ def __init__(self, config, offset_step=0):
     if config.profiler != "" and self.start_initial_profile_step >= config.steps:
       raise ValueError("Profiling requested but initial profiling step set past training final step")
 
+    # Set up the managed profiler on the first device, or all devices, depending on the config.
+    proc_id = jax.process_index()
```
SurbhiJainUSC: Instead of line 50-52, can't we just say:
```
if config.managed_profiler`:
  self.prof = None
  .....
 ```

Comment (xibinliu) [reply src/MaxText/profiler.py]: Hi Surbhi, the logic is needed because:
1. if upload_all_profiler_results, we need do this on all TPU devicers.
2. if not upload_all_profiler_results, we just do it on the first device.

The flag self.managed_profiler will be changed based on the above conditions.

**Comment by xibinliu** [comment]
> How is the correct GCP project info picked up?

The ML Run (managed profiler UI) is always created under the project / regions where the workload is running

File: src/MaxText/pyconfig.py (hunk 2 above)
Comment (bvandermoon) [reply src/MaxText/pyconfig.py:43]: Thanks @xibinliu. No action item on this one, just calling out that this will need to update in the other PR also

**Comment by bvandermoon** [comment]
> > How is the correct GCP project info picked up?
> 
> The ML Run (managed profiler UI) is always created under the project / regions where the workload is running

Thanks @xibinliu. In that case, the info is coming from XPK, right?

**Comment by xibinliu** [comment]
> > > How is the correct GCP project info picked up?
> > 
> > 
> > The ML Run (managed profiler UI) is always created under the project / regions where the workload is running
> 
> Thanks @xibinliu. In that case, the info is coming from XPK, right?

The mldiagnostics SDK figured it out by itself. Even the run is on a GCE VM, it can also find the right the project / region info.

File: src/MaxText/profiler.py (hunk 4 above)
Comment (xibinliu) [reply src/MaxText/profiler.py]: This part has been changed in the latest update, because the sdk provides a way to profiling different devices.
```
    if self.managed_mldiagnostics.enabled and self.mode == "xplane":
      # Handle the special profiling logic for managed_mldiagnostics
      if self.prof is None:
        # Starts xprof collector.
        # Only profiling on the first device, if not upload_all_profiler_results. None is for all devices.
        self.prof = xprof(process_index_list=None if self.upload_all_profiler_results else [0])
      self.prof.start()
```

File: src/MaxText/managed_mldiagnostics.py [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+
+from google_cloud_mldiagnostics import machinelearning_run
+
+from MaxText.pyconfig import KEYS_NO_LOGGING
+
+
+class ManagedMLDiagnostics:
+  """
+  ML Diagnostics Run, implemented with the Singleton pattern.
+  Ensures that only one instance of can exist.
```
SurbhiJainUSC: nit: `one instance of ....`

xibinliu: Thanks! This has been fixed.

src/MaxText/metric_logger.py
```diff
@@ -101,6 +118,9 @@ def write_metrics(self, metrics, step, is_training=True):
       if self.config.gcs_metrics and jax.process_index() == 0:
         self.write_metrics_for_gcs(metrics, step, is_training)
 
+      if self.managed_mldiagnostics.enabled:
```
SurbhiJainUSC: Can we replace it with `if self.config.managed_mldiagnostics:`?

src/MaxText/metric_logger.py
```diff
@@ -82,6 +98,7 @@ def __init__(self, config, learning_rate_schedule):
     self.learning_rate_schedule = learning_rate_schedule
     self.cumulative_eval_metrics = {"scalar": defaultdict(float)}
     self.buffered_train_metrics = None
+    self.managed_mldiagnostics = ManagedMLDiagnostics(config)
```
SurbhiJainUSC: Where are we using `self.managed_mldiagnostics`?
If this is not used anywhere, can we just say:
```
 if self.config.managed_mldiagnostics:
    ManagedMLDiagnostics(config)
```

src/MaxText/profiler.py
```diff
@@ -40,6 +43,8 @@ def __init__(self, config, offset_step=0):
     self.finished_initial_profile_step = self._set_last_profiler_step(config.profiler_steps, config.steps)
     if config.profiler != "" and self.start_initial_profile_step >= config.steps:
       raise ValueError("Profiling requested but initial profiling step set past training final step")
+    self.managed_mldiagnostics = ManagedMLDiagnostics(config)
```
SurbhiJainUSC: Similar comment as above

src/MaxText/managed_mldiagnostics.py
```diff
@@ -0,0 +1,80 @@
...
+from MaxText.pyconfig import KEYS_NO_LOGGING
+
+
+class ManagedMLDiagnostics:
+  """
+  ML Diagnostics Run, implemented with the Singleton pattern.
+  Ensures that only one instance of the class can exist.
+  """
+
+  enabled = False
```
SurbhiJainUSC: This can be removed as well.

xibinliu: Done.

SurbhiJainUSC: Did you forget to push your changes?

xibinliu: Thanks for catching this. I pushed the change.

src/MaxText/metric_logger.py
xibinliu: Done!

src/MaxText/managed_mldiagnostics.py
```diff
@@ -0,0 +1,77 @@
...
+  def __init__(self, config):
+    """
+    Initializes the ManagedMLDiagnostics, ensuring this method runs only once.
+    """
+    # We need a flag to ensure __init__ only runs once,
+    # as the object is returned multiple times by __new__.
+    if hasattr(self, "_initialized"):
+      return
+    self._initialized = True
+    if not config.managed_mldiagnostics:
```
SurbhiJainUSC, line 52: This `if` is not required since you are initializing this class only when `self.config.managed_mldiagnostics` is True.

xibinliu: Hi Surbhi, I have thought about this and added this check as defensive code. Initialization could be in multiple places but just want this to be globally guarded here. WDYT?

PR #2600, opened 2025-11-05
Enable packed sequences and ring attention for CUDNN Flash Attention

kocchop:
# Description

Added THD packed format support and configurable context parallel strategies (all_gather/ring) for TransformerEngine's DotProductAttention. Included comprehensive GPU tests for packed attention (sm90+) and ring attention modes.

- Added max_segments_per_seq config for packed sequence control
- Supported context_parallel_strategy selection (all_gather vs ring)
- Add GPU tests for both packed and ring attention scenarios

Note: Context parallelism with packing temporarily disabled pending full support soon.

# Tests

Appropriate tests for packed sequence with fused attention and ring attention have been added.

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -519,6 +519,11 @@ packing: True
 num_epoch: 1  # only grain and tfds pipeline supports num_epoch > 1
 generate_padding_batch_train: False
 generate_padding_batch_eval: False
+# Maximum number of segments that can be packed into a single sequence
+# This needs to be passed to TransformerEngine's DotProductAttention layer for packing
+# NOTE: This parameter is only relevant for GPU packed attention.
+# When using TPU, this parameter will be ignored.
+max_segments_per_seq: 32
 # Rampup batch size, similar to Megatron-LM, see
 # https://github.com/NVIDIA/Megatron-LM/blob/2a01637aa54ccdaf7ea9afc1f1b80f58c53d7f3c/megatron/core/num_microbatches_calculator.py#L233-L237
 # The ramp-up proceeds in stages from `per_device_batch_size_start` up to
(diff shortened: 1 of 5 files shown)


src/MaxText/configs/base.yml
```diff
@@ -519,6 +519,9 @@ packing: True
 num_epoch: 1  # only grain and tfds pipeline supports num_epoch > 1
 generate_padding_batch_train: False
 generate_padding_batch_eval: False
+# Maximum number of segments that can be packed into a single sequence
+# This needs to be passed to TransformerEngine's DotProductAttention layer for packing
+max_segments_per_seq: 32
```
gobbleturk, line 526: can you add a valueError in pyconfig if this is set on TPU it errors out - this is only supported on GPU?

src/MaxText/configs/base.yml
```diff
@@ -814,6 +817,7 @@ cost_estimate_flops_bwd: -1 # -1 means using splash default cost estmiation, any
 dq_reduction_steps: 0 #the number of reduction steps. For now, only 3 or all the kv steps are supported.
 ### Determine if we want to use load balance for context parallelism
 context_parallel_load_balance: True
+context_parallel_strategy: "all_gather" # "all_gather" or "ring"
```
gobbleturk, line 822: can you add a valueError in pyconfig that ring is only supported on GPU?

src/MaxText/configs/base.yml
kocchop: I can make it a warning for TPUs. But it does not get used in the TPU path. So, not sure if erroring out would make sense

src/MaxText/configs/base.yml
kocchop: added!

src/MaxText/configs/base.yml
kocchop: added a warning @gobbleturk please check

gobbleturk: I realize it doesn't get used in the TPU path but I'm worried some of our TPU users might just read this variable name "max_segments_per_seq" and try it out on TPU. They will not be happy to see it silently doesn't do anything - I would prefer an error

kocchop: Hi @gobbleturk I removed the warning and squashed in a single commit. Could you please check?

PR #2612, opened 2025-11-06
Adding MaxTextForCausalLM interface.

NicoGrande:
# Description

This PR introduces the MaxTextForCausalLM interface, a general wrapper for all MaxText models that makes them discoverable and invokable by VLLM. Additionally, this PR introduces some changes to `model_creation_utils.py` to allow NNX models to be initialized with parameters passed in by VLLM. 

This PR also includes the packaging code enabling for the MaxTextForCausalLM architecture to be registered into VLLM-TPU as a plugin. 

# Tests

MaxTextForCausalLM can be registered with VLLM by running the following command in the same environment as VLLM: 

`cd src/MaxText/integrations/vllm && pip install .`

# Checklist

### DIFF
--- a/src/MaxText/configs/vllm.yml
+++ b/src/MaxText/configs/vllm.yml
@@ -1,0 +1,61 @@
+# Copyright 2023–2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+base_config: "base.yml"
+attention: "vllm_rpa"
+# NNX required for vLLM integration
+enable_nnx: True
+# Avoid re-initializing JAX distributed system when using vLLM
+skip_jax_distributed_system: True
+# Scanned layers are not supported with vLLM integration
+scan_layers: False
+
+
+# -------------- Logical Axis Rules --------------
+mesh_axes: ['data', 'model', 'expert']
+logical_axis_rules: [
+                      ['activation_batch', ['data', 'expert']],
+                      ['activation_batch_no_exp', ['data']],
+                      ['activation_embed_and_logits_batch', ['data', 'expert']],
+                      ['activation_embed_and_logits_batch_sequence', ['data', 'expert']],
+                      ['activation_heads', ['model']],
+                      ['activation_kv_heads', ['model']],
+                      ['activation_length', ['expert']],
+                      ['activation_q_length', ['expert']],
+                      ['activation_embed', ['model']],
+                      ['activation_mlp', ['model']],
+                      ['activation_kv', ['model']],
+                      ['activation_prefill_kv_batch', ['data', 'expert']],
+                      ['activation_kv_batch', ['data', 'expert']],
+                      ['activation_kv_batch_no_exp', ['data']],
+                      ['activation_kv_head_dim', ['model']],
+                      ['activation_vocab', ['model']],
+                      ['activation_exp', ['expert']],
+                      ['decode_batch', ['data', 'expert']],
+                      ['mlp', ['model']],
+                      ['mlp_no_fsdp', ['model']],
+                      ['vocab', ['model']],
+                      ['heads', ['model']],
+                      ['q_heads', ['model']],
+                      ['kv_heads', ['model']],
+                      ['embed', ['expert']],
+                      ['q_lora', ['expert']],
+                      ['kv_lora', ['expert']],
+                      ['norm', ['model']],
+                      ['cache_heads', ['model']],
+                      ['exp', ['expert']],
+                      ['paged_kv_heads', ['model']],
+                    ]
+data_sharding: [['data', 'model', 'expert']]
+input_data_sharding_logical_axes: ['activation_embed_and_logits_batch']
--- a/src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
+++ b/src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
@@ -1,0 +1,33 @@
+# Copyright 2023–2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MaxText vLLM adapter package."""
+
+from tpu_inference.logger import init_logger
+from tpu_inference.models.common.model_loader import register_model
+from .adapter import MaxTextForCausalLM
+
+
+logger = init_logger(__name__)
+
+
+def register():
+  """Register MaxTextForCausalLM model with tpu_inference and vllm.
+
+  Note, this function is invoked directly by the vLLM engine during startup. As such,
+  it leverages vLLM logging to report its status.
+  """
+  logger.info("Registering MaxTextForCausalLM model with tpu_inference and vllm.")
+  register_model("MaxTextForCausalLM", MaxTextForCausalLM)
+  logger.info("Successfully registered MaxTextForCausalLM model.")
--- a/src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
+++ b/src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
@@ -1,0 +1,269 @@
+# Copyright 2023–2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""vLLM adapter for MaxText models."""
+
+import jax
+import jax.numpy as jnp
+
+from etils import epath
+from flax import nnx
+from jax.sharding import Mesh
+from MaxText import model_creation_utils
+from MaxText import pyconfig
+from MaxText.common_types import MODEL_MODE_AUTOREGRESSIVE
+from MaxText.globals import MAXTEXT_PKG_DIR
+
+from tpu_inference.layers.common.attention_metadata import AttentionMetadata
+from vllm.config import VllmConfig
+
+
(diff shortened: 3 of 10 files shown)


src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from .adapter import MaxTextForCausalLM
+
+
+logger = init_logger(__name__)
+
+
+def register():
+  logger.info("Registering MaxTextForCausalLM model with tpu_inference and vllm.")
+  register_model("MaxTextForCausalLM", MaxTextForCausalLM)
+  logger.info("Successfully registered MaxTextForCausalLM model.")
```
richjames0, line 33: should files end with a blank line?

src/MaxText/model_creation_utils.py
```diff
@@ -112,17 +117,23 @@ def create_model(config, mesh, model_mode: str = MODEL_MODE_TRAIN, rngs: nnx.Rng
...
+def create_nnx_model(config, mesh=None, devices=None, model_mode=None, rng_key=None):
   """Creates a NNX model with sharded parameters, possibly loading from a checkpoint."""
 
-  def _create_model():
-    init_rng = jax.random.PRNGKey(config.init_weights_seed)
-    return from_config(config, devices, rngs=nnx.Rngs(params=init_rng, dropout=1))
+  def _create_model(mesh: Mesh | None = None, model_mode: str | None = None, rng_key: jax.Array | None = None):
+    if rng_key is None:
+      rng_key = jax.random.PRNGKey(config.init_weights_seed)
+    rngs = nnx.Rngs(params=rng_key, dropout=1)
```
richjames0: i don't understand this interface so i'll just flag that this unconditional `dropout=1` scares me a little

src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
NicoGrande: I thought so too! This is what I got when running the pre-commit hooks however. I can manually add them.

src/MaxText/model_creation_utils.py
NicoGrande: This is a bit confusing to me also, however I set it this way since this is how it currently is defined in main: https://github.com/AI-Hypercomputer/maxtext/blob/bfdb7edb1cdb5c3d4679a034360ad744ea197790/src/MaxText/model_creation_utils.py#L120

Is my understanding correct that this is setting a specific seed for nnx dropout layers? Is the lack of determinism for seed generation the scary part?

src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
NicoGrande: Oh, it looks like the blank lines are there in my IDE, they just aren't rendered in the diff for some reason.

src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from tpu_inference.logger import init_logger
+from tpu_inference.models.common.model_loader import register_model
+from .adapter import MaxTextForCausalLM
+
+
+logger = init_logger(__name__)
+
+
+def register():
+  logger.info("Registering MaxTextForCausalLM model with tpu_inference and vllm.")
```
SurbhiJainUSC, line 31: Can we use `max_logging` instead?

NicoGrande: These specific lines are invoked by vLLM and are related to registration of an architecture with vLLM, which can be independent from the MaxText code. Because of this, I would argue it makes more sense for this to be surfaced as a vLLM log. Is there something about `max_logging` that makes it beneficial over this?

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Returns:
+      A JAX array of logits (Q, vocab_size).
+    """
+    if self.logits is not None:
+      return self.logits
+
+    embeddings = self.model.token_embedder
+    # pylint: disable=protected-access
+    return self.model.decoder._apply_output_head(embeddings, hidden_states, True, self.model_mode)
```
gagika: should we change this to be non protected?

_apply_output_head -> apply_output_head

src/MaxText/model_creation_utils.py
gagika: I guess dropout won't be used during inference, can we have: 

rngs = nnx.Rngs(params=seed, dropout=None)  # disable dropout RNG for inference

gagika (commented):
thanks

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
NicoGrande: Yes, happy to make this change, let me go ahead and add this.

src/MaxText/model_creation_utils.py
NicoGrande: I think this method can also be used to initialize models for training. I can add conditional logic to disable the dropout RNG based on the model mode.

NicoGrande: I needed to remove the explicit `dropout=None` since this was causing errors.

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -0,0 +1,268 @@
...
+    embeddings = self.model.token_embedder
+    return self.model.decoder.apply_output_head(embeddings, hidden_states, True, self.model_mode)
+
+  def load_weights(self, rng_key: jax.Array) -> None:
+    """Loads model parameters on the provided mesh.
+
+    Args:
+      rng_key: A JAX random key for model initialization.
+    """
+    self.model, _ = model_creation_utils.create_nnx_model(
```
khatwanimohit, line 184: is this function loading the whole model along with opt state on TPUs? I guess vLLM is just expecting us to load the weights on the given mesh.

src/MaxText/configs/vllm.yml
```diff
@@ -0,0 +1,63 @@
...
+                      ['activation_q_length', ['expert']],
+                      ['activation_embed', ['model']],
+                      ['activation_mlp', ['model']],
+                      ['activation_kv', ['model']],
+                      ['activation_prefill_kv_batch', ['data', 'expert']],
+                      ['activation_kv_batch', ['data', 'expert']],
+                      ['activation_kv_batch_no_exp', ['data']],
+                      ['activation_kv_head_dim', ['model']],
+                      ['activation_vocab', ['model']],
+                      ['activation_vocab', ['model']],
```
gagika: there are duplicate

src/MaxText/configs/vllm.yml
```diff
@@ -0,0 +1,63 @@
...
+                      ['vocab', ['model']],
+                      ['heads', ['model']],
+                      ['q_heads', ['model']],
+                      ['kv_heads', ['model']],
+                      ['embed', ['expert']],
+                      ['q_lora', ['expert']],
+                      ['kv_lora', ['expert']],
+                      ['norm', ['model']],
+                      ['cache_heads', ['model']],
+                      ['cache_heads', ['model']],
```
gagika: same here, duplicate

src/MaxText/integration/vllm/setup.py
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Setup for MaxText vLLM adapter package."""
+
+from setuptools import setup
+
+setup(
```
khatwanimohit, line 19: for my understanding, what is this file doing? when is this setup supposed to be called

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -0,0 +1,268 @@
...
+      raise ValueError("Model must be an instance of type nnx.Module.")
+
+    if input_ids.ndim < 2:
+      input_ids = jnp.expand_dims(input_ids, axis=0)
+
+    input_positions = attention_metadata.input_positions
+    if input_positions.ndim < 2:
+      input_positions = jnp.expand_dims(input_positions, axis=0)
+
+    aux_hidden_states = []
```
gagika, line 146: what is it for?, right now it doesn't seam to be used.

src/MaxText/model_creation_utils.py
```diff
@@ -46,6 +47,7 @@ def from_config(
 def from_config(
     config: pyconfig.HyperParameters,
     devices: Sequence[jax.Device] | None = None,
+    mesh: Mesh | None = None,
```
khatwanimohit, line 50: this can be done in a future PR, devices argument was kept here so we can initialize a maxtext model on a subset of jax.devices() for RL. I understand we are adding mesh so can pass a vLLM mesh, we should remove devices argument and just allow users to pass a mesh to this function.

src/MaxText/configs/vllm.yml
NicoGrande: Thank you for catching! Will Fix.

src/MaxText/configs/vllm.yml
NicoGrande: Same!

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
NicoGrande: `aux_hidden_states` is an additional expected output in the vLLM interface we are implementing. We do not use it so we just return an empty list. I added the variable for readability but can just return [] instead. 

An example of this in the `tpu-inference` llama3 implementation: https://github.com/vllm-project/tpu-inference/blob/c46669b38c33881d18a8e0a671c7d3ae192e1026/tpu_inference/models/jax/llama3.py#L287-L299

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
NicoGrande: Yes, this is correct.

src/MaxText/integration/vllm/setup.py
NicoGrande: This is the function that allows this interface to be registered with vLLM as a plugin by pip installing it. As you can see, the `register` method is invoked in `setup`. The `register` function itself is defined in the `maxtext_vllm_adapter` `__init__.py`.

src/MaxText/model_creation_utils.py
NicoGrande: Great! Lets add this in a follow up PR

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
gagika: if you are setting `config.load_parameters_path` then it should just load model weights.

src/MaxText/integration/vllm/maxtext_vllm_adapter/adapter.py
gagika: I see, please add some comment describing the use case and update return type to reflect it e.g.
tuple[list[jax.Array], jax.Array, list[jax.Array]]:

NicoGrande: Done!

gagika (approved):
thanks

src/MaxText/integration/vllm/maxtext_vllm_adapter/__init__.py
shralex: can you please comment to this fact somewhere in the doc

PR #2613, opened 2025-11-06
Add preprocessing utils for Qwen3-Omni

hengtaoguo:
# Description

- Add image/video/audio preprocessing utils for Qwen3-Omni in `MaxText.multimodal.qwen3_omni_preprocessor.preprocess_mm_data_qwen3_omni()`, returning dataclass `Qwen3OmniPreprocessorOutput` containing all preprocessed data (`pixel_values`, `pixel_grid_thw`, `video_values`, `video_grid_thw`, `video_second_per_grid`, `audio_values`, `audio_mask`).
- Add unit test comparing MaxText implementation with Qwen3-Omni's processor on HuggingFace.
- [WIP] Refactor [multimodal_utils.py]:
  - `MaxText.multimodal.utils`: Commonly used basic functions such as image loading and normalization.
  - `MaxText.multimodal.{MODEL}_preprocessor.py`: Model-specific preprocessing utils.
  - `MaxText.multimodal.preprocessor.py`: Centralized function `preprocess_mm_data()` will route to model-specific preprocessing logics based on model name.

# Tests

Passing unit tests for MaxText `preprocess_mm_data_qwen3_omni` vs HuggingFace `Qwen3OmniMoeProcessor`:

```
python -m unittest tests.check_qwen3_embedding_vs_reference.TextQwen3OmniPreprocessing
```

This unit tests compares the data processor outputs between HF and MaxText implementation on the same inputs. Since it requires pytorch within HF processor, we are not adding it to the CI tests.

Multi-image loading also tested.

Note: Since `jax.image.resize` still grabs TPU even we specify CPU, we are going to use PIL for video resize by looping through all frames. Here's a code copy for `jax.image.resize`: [link](https://paste.googleplex.com/5176547155771392).

Now, using PIL image resize, the interpolation detail is different from the original [torchvision.transforms.functional.resize](https://github.com/QwenLM/Qwen2.5-Omni/blob/d8a31ca56c0456b6edfcbcbf4bdbb6ae2200ef42/qwen-omni-utils/src/qwen_omni_utils/v2_5/vision_process.py#L429), causing slightly larger difference. So we adopted `5e-2` in the unit test for video outputs comparison:

```
video diff stats: max 0.02987128496170044, min 0.0, mean 0.0011402600212022662
mt video stats: max 1.0368887186050415, min -1.055976152420044, mean -0.6622966527938843
hf video stats: max 1.036394476890564, min -1.0549969673156738, mean -0.6623122692108154
```

# Checklist


src/MaxText/layers/decoders.py
```diff
@@ -580,6 +580,8 @@ def _apply_embedding(
             image_masks=image_masks,
         )
       # TODO(hengtaoguo): Add support for other multimodal models such as Llama4, refactor if needed
+      elif cfg.model_name in ["qwen3-omni-30b-a3b"]:
```
eitanporat: what is the reasoning behind this change?

src/MaxText/multimodal/preprocessor.py
```diff
@@ -0,0 +1,47 @@
...
+
+  Returns:
+    A `PreprocessorOutput` object containing the processed multimodal data.
+  """
+  processor_outputs = multimodal_utils.PreprocessorOutput()
+
+  if config.model_name in ["gemma3-4b", "gemma3-12b", "gemma3-27b"]:
+
+    images = [multimodal_utils.load_image_from_path(p) for p in config.image_path.split(",")]
+    processor_outputs = multimodal_utils.pre_process_gemma3_image(images)
```
eitanporat, line 35: maybe rename the functions to `preprocess_mm_data_gemma3` ?

maybe it would be better to use a factory pattern here

src/MaxText/multimodal/preprocessor.py
```diff
@@ -0,0 +1,47 @@
...
+  elif config.model_name in ["llama4-17b-16e", "llama4-17b-128e"]:
+
+    images = [multimodal_utils.load_image_from_path(p) for p in config.image_path.split(",")]
+    processor_outputs = multimodal_utils.pre_process_llama4_image(images)
+  elif config.model_name in ["qwen3-omni-30b-a3b"]:
+    from MaxText.multimodal.qwen3_omni_processor import preprocess_mm_data_qwen3_omni  # pylint: disable=import-outside-toplevel
+
+    processor_outputs = preprocess_mm_data_qwen3_omni(config)
```
eitanporat, line 43: why does it accept a config?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+MAX_RATIO = 200
+
+# Video constants.
+VIDEO_MIN_PIXELS = 128 * 28 * 28
+VIDEO_MAX_PIXELS = 768 * 28 * 28
+VIDEO_TOTAL_PIXELS = 128000 * 28 * 28 * 0.9
+FRAME_FACTOR = 2
+FPS = 2.0
+FPS_MIN_FRAMES = 4
+FPS_MAX_FRAMES = 768
```
eitanporat: should move to the config

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+  max_pixels = max(min(VIDEO_MAX_PIXELS, VIDEO_TOTAL_PIXELS / nframes * FRAME_FACTOR), int(VIDEO_MIN_PIXELS * 1.05))
+  resized_height, resized_width = smart_resize(
+      height,
+      width,
+      factor=IMAGE_FACTOR,
+      min_pixels=VIDEO_MIN_PIXELS,
+      max_pixels=max_pixels,
+  )
+
+  with jax.default_device(jax.devices("cpu")[0]):
```
eitanporat: why is this necessary i am curious.
could it support multiple cpus for example?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+
+  if db_range is not None:
+    if db_range <= 0.0:
+      raise ValueError("db_range must be greater than zero")
+    spectrogram_array = np.clip(spectrogram_array, a_min=spectrogram_array.max() - db_range, a_max=None)
+
+  return spectrogram
+
+
+def spectrogram(
```
eitanporat: maybe it's better to move the audio parts to a processor for audio. many audio models use this functionality.

src/MaxText/multimodal/utils.py
```diff
@@ -0,0 +1,71 @@
...
+  Change the image mean and std based on parameters mean and std.
+  Args:
+    images: The images to normalize.
+    mean: tuple[float, float, float].
+    std: tuple[float, float, float].
+  Returns:
+    The normalized images.
+  """
+  images -= np.asarray(mean)
+  images /= np.asarray(std)
```
eitanporat, line 75: curious why it's called mean std. as this isn't used as mean as std.

it transforms the mean and std in the following way. if mean[images]=m and std[images]=s then
if x = normalize_images(images, mean, std), mean[x] = (m-mean)/std and std[x] = s/std

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+def _np_extract_fbank_features(waveform_batch: np.ndarray, device: str) -> np.ndarray:
+  """
+  Compute the log-mel spectrogram of the provided audio, gives similar results to Whisper's original torch
+  implementation with 1e-5 tolerance.
+  """
+  if device != "cpu":
+    raise ValueError(
+        f"Got device `{device}` for feature extraction, but feature extraction on CUDA accelerator "
```
eitanporat: i guess this was copied from torch but this comment should be changed.

eitanporat:
is the functionality implemented on cpu in numpy in the torch variant. if so, is there a reason not to want to reuse it?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+
+  spectrogram_array = np.clip(spectrogram_array, a_min=min_value, a_max=None)
+  spectrogram_array = 10.0 * (np.log10(spectrogram_array) - np.log10(reference))
+
+  if db_range is not None:
+    if db_range <= 0.0:
+      raise ValueError("db_range must be greater than zero")
+    spectrogram_array = np.clip(spectrogram_array, a_min=spectrogram_array.max() - db_range, a_max=None)
+
+  return spectrogram
```
eitanporat: you probably have an error here... it will err

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+    processor_outputs.video_grid_thw = video_grid_thw
+    processor_outputs.video_second_per_grid = np.asarray([config.temporal_patch_size_for_vit], dtype=np.float32)
+    processor_outputs.num_videos = 1  # Only one video for now.
+
+  if config.audio_path is not None or (config.video_path is not None and config.use_audio_in_video):
+    mt_audio = _load_audio(config.video_path)
+    mt_audio, mt_audio_mask = pre_process_audio_qwen3_omni(mt_audio)
+    processor_outputs.audio_values = mt_audio
+    processor_outputs.audio_mask = mt_audio_mask
+
```
eitanporat, line 488: this will break if you want audio and not video!

eitanporat:
could you add the new requirements to the pyproject toml (`decord` and `librosa`)?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+  processor_outputs = Qwen3OmniPreprocessorOutput()
+
+  if config.image_path is not None:
+    images = [mm_utils.load_image_from_path(p) for p in config.image_path.split(",")]
+    pixel_values, pixel_grid_thw = pre_process_qwen3_image(images, config)
+    processor_outputs.pixel_values = pixel_values
+    processor_outputs.pixel_grid_thw = pixel_grid_thw
+    processor_outputs.num_images = len(images)
+
+  if config.video_path is not None:
```
eitanporat, line 474: did it work for you? in the base.yml i see config.video_path = "" so this doesn't actually work for me.

hengtaoguo:
> is the functionality implemented on cpu in numpy in the torch variant. if so, is there a reason not to want to reuse it?

This has been a long-standing constraint, we intentionally exclude `torch` from our dependency. So we cannot use torch resize functions and need to reimplement everything in numpy/jnp.

src/MaxText/multimodal/qwen3_omni_processor.py
aireenmei: I have the same question. Could we replace jax.image.resize with other alternatives to avoid jax grabbing TPU for preprocessing. Perhaps we can apply the resize function from PIL on video frames? We did the same for gemma3 (https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/multimodal_utils.py#L359-L367) and llama4

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+from dataclasses import dataclass
+
+from MaxText import max_logging
+from MaxText.multimodal import utils as mm_utils
+
+# Image constants.
+IMAGE_MEAN = 127.5
+IMAGE_STD = 127.5
+IMAGE_FACTOR = 28
+MIN_PIXELS = 4 * 28 * 28
```
aireenmei: Add some comments to explain the constants? Same for video constants. It's not easy to tell from name.

src/MaxText/multimodal_utils.py
```diff
@@ -452,10 +455,11 @@ def pre_process_llama4_image(image: np.ndarray | list[np.ndarray]) -> Preprocess
       pixel_mask=image_mask,
       aspect_ratios=aspect_ratios_array,
   )
+  processor_output.num_images = len(image)
   return processor_output
 
 
-def pre_process_image(image, model_name):
+def pre_process_image(image, model_name, config=None):
```
aireenmei: Why is config needed here?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+
+  Returns:
+      tuple: (numpy.ndarray with shape (T, C, H, W), sample_fps as float)
+  """
+  video_config = {
+      "video": video_path,
+      "video_start": video_start,
+      "video_end": video_end,
+  }
+  vr = decord.VideoReader(video_path)
```
aireenmei: Add try/except to handle errors?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+def _load_audio(data_path: str) -> np.ndarray:
+  """Load audio from a file path.
+
+  Args:
+      data_path (str): The path to the audio file.
+
+  Returns:
+      np.ndarray: The loaded audio waveform.
+  """
+  audio = librosa.load(data_path, sr=SAMPLE_RATE)[0]
```
aireenmei: Add try/except to handle errors?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1032 @@
...
+  if config.video_path is not None:
+    video_array, _ = _read_video_decord(config.video_path)
+    video_processed, video_grid_thw = preprocess_video(video_array, config)
+    processor_outputs.video_values = video_processed
+    processor_outputs.video_grid_thw = video_grid_thw
+    processor_outputs.video_second_per_grid = np.asarray([config.temporal_patch_size_for_vit], dtype=np.float32)
+    processor_outputs.num_videos = 1  # Only one video for now.
+
+  if config.audio_path is not None or (config.video_path is not None and config.use_audio_in_video):
+    mt_audio = _load_audio(config.video_path)
```
aireenmei: Should this be `_load_audio(config.audio_path)` when audio_path is provided?

src/MaxText/multimodal/qwen3_omni_processor.py
```diff
@@ -0,0 +1,1026 @@
...
+    )
+    img_np = np.permute_dims(img_np, (0, 1, 4, 7, 5, 8, 3, 2, 6, 9))  # HWC to CHW
+    img_np = np.reshape(
+        img_np, (batch_size, grid_t * grid_h * grid_w, channel * temporal_patch_size * patch_size * patch_size)
+    )
+    img_grid_thw = np.asarray([grid_t, grid_h, grid_w], dtype=np.int32)
+    images_out.append(img_np)
+    grids_thw.append(img_grid_thw)
+
+  return images_out[0][0, :, :], grids_thw[0]
```
aireenmei: Why only return the first item here?

src/MaxText/multimodal/preprocessor.py
hengtaoguo: Yes, I will rename them to `preprocess_mm_data_gemma3 ` and `preprocess_mm_data_llama4` in next PR, because I need to move them to this dedicated multimodal folder.

src/MaxText/multimodal/preprocessor.py
hengtaoguo: I think it can provide better control and can generalize to the future models. For Omni, it is using many flags including config.image/video/audio_path, config.temporal_patch_size_for_vit and config.use_audio_in_video.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Thanks! These flags were not configurable in the [huggingface implementation](https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/blob/main/preprocessor_config.json). Instead, they are hard coded in the [omni utils](https://github.com/QwenLM/Qwen2.5-Omni/blob/d8a31ca56c0456b6edfcbcbf4bdbb6ae2200ef42/qwen-omni-utils/src/qwen_omni_utils/v2_5/vision_process.py#L31-L36), which appears to be Omni-specific. How about keep them here for now, and if future models share the same configs we can factor them out?

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: This is a good suggestion. The reason is that:

- `jax.image.resize` is faster when processing an image cube (video) size of (N frames, H, W). While for PIL library, I had a version running for-loop for each frame, but is much slower than a single jax operator. Doing for-loops in preprocessing for a large batch could harm the efficiency.
- To avoid grabbing TPU here, I forced it to run on CPU.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Done! Removed all CUDA related codes.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Done, moved general functions to a separate utils.py.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Great suggestion! Done

src/MaxText/multimodal_utils.py
hengtaoguo: Reverted.

However, in the new dedicated multimodal preprocessor, config is needed because it already contains many flags to use such as "image/video/audio paths, model name, use_image/video/audio, use_audio_in_video". In the future, this list of required flags could keep growing. So I assume we should directly pass the config instead of adding new flags to the function call each time.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Done!

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Done! Catching FileNotFoundError and RuntimeError now.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Thanks for the catch! The intent here is to support "use_audio_in_video" use case, and I didn't address the part for loading audio only files. I added a TODO here to solve later.

src/MaxText/multimodal/qwen3_omni_processor.py
hengtaoguo: Good catch! I've updated and now it supports multiple images output. Namely, all images are concatenated along the sequence dimension like HuggingFace, e.g. (seq1 + seq2, 1536)

src/MaxText/multimodal_utils.py
```diff
@@ -452,10 +455,11 @@ def pre_process_llama4_image(image: np.ndarray | list[np.ndarray]) -> Preprocess
       pixel_mask=image_mask,
       aspect_ratios=aspect_ratios_array,
   )
```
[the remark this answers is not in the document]
hengtaoguo: Reverted, will refactor in next PR

src/MaxText/multimodal_utils.py
```diff
@@ -90,9 +90,11 @@ class PreprocessorOutput:
   pixel_values: None | np.ndarray = None
   pixel_mask: None | np.ndarray = None
   aspect_ratios: None | np.ndarray = None
+  num_images: int = 0
```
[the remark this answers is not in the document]
hengtaoguo: Going to move other preprocessors to the dedicated folder in next PR.

PR #2616, opened 2025-11-06
Modifying decoders and attention for vllm

NicoGrande:
# Description

This PR introduces new optional arguments `kv_cache` and `attention_metadata` for model decoder blocks and for the `attentions.py` module. These arguments are provided by vLLM when executing a MaxText model from vLLM Engine and are used when calling the ragged paged attention kernel in `tpu-inference`.

This PR builds on the work started in #2612.

**Note: This PR changes the expected method signature from Attention layers and decoder layers.**

# Tests

Includes a new unit-test in `attention_test.py`. 

Additionally, end-to-end tests were performed locally on a v6e VM using the following test command:

In one process, start the vLLM server. This requires a local `config.json` file for the corresponding model you are trying to test from HuggingFace. Modify this file such that `architectures: "MaxTextForCausalLM"` is set.
```
HF_TOKEN=<YOUR_HF_TOKEN> TPU_BACKEND_TYPE=jax \
  python -m vllm.entrypoints.cli.main serve \
  <HF_MODEL> \
  --max-num-batched-tokens=32 \
  --max-model-len=32 \
  --max-num-seqs=1 \
  --tensor-parallel-size=4 \
  --hf_config_path=<PATH_TO_LOCAL_HF_CONFIG_JSON> \
  --additional-config='{"maxtext_config": {"model_name": "<MAXTEXT_MODEL_NAME>", "max_prefill_predict_length": 28, "max_target_length": 32, "ici_tensor_parallelism": 4, "load_parameters_path": "<MAXTEXT_CHECKPOINT_PATH>"}}'
```

In a second process, issue the query to the model:
```
curl http://localhost:8000/v1/completions \
   -H "Content-Type: application/json" \
   -d '{
       "model": "<HF_MODEL>",
       "prompt": ["Seattle is a"],
       "max_tokens": 16,
       "temperature": 0
   }'
```
Results for different tested models are shown below:
## llama3.1-8b
```
output: " city that is known for its coffee culture, and it's not hard to see"
```

## gemma3-4b
```
output: " vibrant city with a lot to offer, and it's a great place to"
```

## qwen3-8b
```
output: " city in the state of Washington, in the Pacific Northwest region of the United States"
```
# Checklist

### DIFF
--- a/src/MaxText/layers/attention_op.py
+++ b/src/MaxText/layers/attention_op.py
@@ -847,11 +847,14 @@ class AttentionOp(nnx.Module):
         raise NotImplementedError(target_hardware)
       return impl(query, key, value, lengths, self.ragged_block_size)
 
+    # 'vllm_rpa' uses the same dot-attention wrapper but routes to the vLLM
+    # ragged paged attention kernel in `Attention.__call__`.
     elif (
         self.attention_kernel == "dot_product"
         or (self.attention_kernel == "autoselected" and model_mode == MODEL_MODE_AUTOREGRESSIVE)
         or (self.attention_kernel == "autoselected" and length < 128)
         or (self.attention_kernel == "paged")
+        or (self.attention_kernel == "vllm_rpa")
     ):
       return self.apply_attention_dot(
           query,
--- a/src/MaxText/layers/attentions.py
+++ b/src/MaxText/layers/attentions.py
@@ -889,6 +889,51 @@ class Attention(nnx.Module):
     )
     return [prefill_kv_cache, ar_kv_cache]
 
+  def forward_serve_vllm(
+      self,
+      query: Array,
+      key: Array,
+      value: Array,
+      rpa_kv_cache: list[Array] | None = None,
+      rpa_metadata: dict[str, Any] | None = None,
+  ) -> tuple[list[Array], Array]:
+    """Forward function for vLLM serving with RPA attention."""
+    try:
+      # pylint: disable=import-outside-toplevel
+      # pytype: disable=import-error
+      from tpu_inference.layers.jax.attention_interface import sharded_ragged_paged_attention as rpa_ops
+    except ImportError as e:
+      raise ImportError(
+          "vLLM RPA attention ops require the vllm-tpu package. Please install it with `pip install vllm-tpu`."
+      ) from e
+
+    if self.config.attention_sink:
+      raise NotImplementedError("Attention sink is not supported in MaxText vLLM RPA attention.")
+
+    if rpa_kv_cache is None or rpa_metadata is None:
+      raise ValueError("kv_cache and attention_metadata must be provided when using vLLM.")
+
+    query = query.reshape(-1, query.shape[2], query.shape[3])
+    key = key.reshape(-1, key.shape[2], key.shape[3])
+    value = value.reshape(-1, value.shape[2], value.shape[3])
+
+    attention_chunk_size = self.config.chunk_attn_window_size if self.config.chunk_attn_window_size > 0 else None
+    q_scale, k_scale, v_scale = None, None, None
+
+    md = rpa_metadata
+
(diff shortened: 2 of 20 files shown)


src/MaxText/layers/models.py
```diff
@@ -386,6 +397,8 @@ def __call__(
       decoder_target_tokens: Target tokens for the decoder (optional, used in MTP).
       decoder_target_mask: Target mask for the decoder (optional, used in MTP).
       nnx_method: Method to call on the NNX module (optional).
+      kv_caches: List of KV caches for each attention layer, used when invoking from vLLM (optional).
+      attention_metadata: Mapping to store attention metadata, used when invoking from vLLM (optional).
 
     Returns:
       Logits from the Transformer model.
```
gagika: Could you update return to be `logits, hidden_state, kv_caches` if called from vLLM

src/MaxText/layers/attentions.py
```diff
@@ -1000,6 +1040,15 @@ def __call__(
           query, key, value, decoder_segment_ids, model_mode, previous_chunk, slot=slot, page_state=page_state
       )
       out = unnormalized_out / (exp_sum + 1e-9) if exp_sum is not None else unnormalized_out
+
+    elif self.config.attention == "vllm_rpa" and attention_metadata and model_mode != MODEL_MODE_TRAIN:
+      batch, seq_len, num_heads, head_dim = query.shape
```
cgarciae, line 1080: if `self.config.attention == "vllm_rpa"` but there is not attention metadata should we raise instead?

NicoGrande: Great catch! This is actually a small hack I put together to pass the MaxText model compilation stage. To get around this, the more correct approach would be to pass a `dummy_attention_metadata` object in the following line: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/layers/models.py#L307-L311 

If we do that, then we can raise an error inside the `elif` to fail in case no `attention_metadata` is provided.

src/MaxText/layers/models.py
```diff
@@ -199,6 +204,10 @@ def __call__(
           model_mode=model_mode,
       )
 
+    if self.config.attention == "vllm_rpa":
+      # In vLLM, logits are computed separately after updating the KV cache.
+      return logits, hidden_state, kv_caches
```
gagika, line 211: We little bit discussed it in person

Can we skip computing the logits for vllm_rpa use case as it might not be needed (and when it's needed it can be computed from hidden_state) and this will avoid need for caching it?

it's not urgent, can be a follow up PR.

gagika (approved):
Thanks Nico,
Feel free to address the comments in a follow up PR.

src/MaxText/layers/models.py
NicoGrande: Great, added this!

src/MaxText/layers/attentions.py
NicoGrande: Done!

src/MaxText/layers/models.py
NicoGrande: Sounds great, will add this in a follow up PR.

src/MaxText/layers/attention_op.py
```diff
@@ -852,6 +852,7 @@ def apply_attention(
         or (self.attention_kernel == "autoselected" and model_mode == MODEL_MODE_AUTOREGRESSIVE)
         or (self.attention_kernel == "autoselected" and length < 128)
         or (self.attention_kernel == "paged")
+        or (self.attention_kernel == "vllm_rpa")
```
gagika, line 857: could you comment somewhere:

```
# 'vllm_rpa' uses the same dot-attention wrapper but routes to the vLLM
# ragged paged attention kernel in `Attention.__call__`.
```

gagika (approved):

src/MaxText/layers/attention_op.py
NicoGrande: Added!

PR #2636, opened 2025-11-07
Reduce memory requirements for checkpoint conversion 

khatwanimohit:
# Description

This PR introduces a lazy_load configuration option that reduces peak RAM usage by deferring the loading and transformation of weights until the exact moment Orbax needs to save them to disk.

Note: lazy loading feature is temporarily disabled for multimodal models because the key names in the files are different from what are seen when loading a HF model using AutoConfig.  

## Key Changes

- Lazy Loading Mechanism: Implemented LazyHFLoader and LazyTensor classes. Instead of eagerly loading all weights, we now create lightweight proxy objects that only load their specific tensor data from disk when Orbax calls __array__() during the saving phase.
- Orbax Integration: Registered a custom LazyTensorHandler with Orbax to allow it to correctly recognize and process our proxy objects as if they were standard NumPy arrays.
- Config: Added a new boolean flag lazy_load to the MaxText configuration to enable this mode. It defaults to False to preserve existing behavior for smaller models.

FIXES: b/458745828

# Tests

x here denotes number of B parameters

## Llama.3.1-70B

old RAM usage: 616 GB (8.8x GB)
new RAM usage: 86.2 GB (1.2x GB)
Logs: https://paste.googleplex.com/6288510168465408#l=571

## LLama3.1-8B

old RAM usage: 51 GB (6.3x GB)
new RAM usage: 31GB (4x GB)
Logs: https://paste.googleplex.com/6128000672333824
forward pass logits test: https://paste.googleplex.com/5595374018494464

## Qwen3-4B

old RAM usage: 37 GB ( 9.2x GB)
new RAM usage: 15.7 GB ( 3.7x GB )
Logs: https://paste.googleplex.com/4933249042350080#l=543
forward pass logit test: https://paste.googleplex.com/5515770054443008

# Checklist

### DIFF
--- a/src/MaxText/utils/ckpt_conversion/to_maxtext.py
+++ b/src/MaxText/utils/ckpt_conversion/to_maxtext.py
@@ -25,6 +25,9 @@ Key Parameters (to be set in the config file or as command-line overrides):
                          Defaults to "./mt_output/".
   scan_layers: (bool) Whether the MaxText model was trained with scanned layers.
                This must match the training configuration of the checkpoint.
+  lazy_load: (bool) If True, uses an on-demand loading strategy to minimize RAM
+             usage during conversion. Recommended if, 2 * model_size (GB) >= system RAM
+             Defaults to False.
 
 Environment Variables:
   HF_AUTH_TOKEN: (Required) HuggingFace authentication token, needed to
@@ -33,29 +36,44 @@ Environment Variables:
 Example Usage:
   To convert a gemma2-2b model and save it to a specific directory:
 
-  HF_AUTH_TOKEN="hf_YOUR_TOKEN" python src/MaxText/utils/ckpt_conversion/to_maxtext.py \
-    --model_name="gemma2-2b" \
-    --base_output_directory="/path/to/your/output/directory" \
-    --scan_layers=False
+    /usr/bin/time -v python src/MaxText/utils/ckpt_conversion/to_maxtext.py \
+    MaxText/configs/base.yml model_name="gemma2-2b" \
+    base_output_directory="/path/to/your/output/directory" \
+    hf_access_token=$HF_TOKEN hardware=cpu skip_jax_distributed_system=True \
+    scan_layers=False
 
   For models with scanned layers (e.g., some custom architectures), you might
   need to set scan_layers=True and param_scan_axis accordingly.
+
+  To convert a 70B model with minimal RAM usage:
+
+   /usr/bin/time -v python src/MaxText/utils/ckpt_conversion/to_maxtext.py \
+    MaxText/configs/base.yml model_name="meta-llama/Llama-3.1-70B" \
+    base_output_directory="gs://my-bucket/maxtext-checkpoints" \
+    hf_access_token=$HF_TOKEN hardware=cpu skip_jax_distributed_system=True \
+    --lazy_load_tensors=True
 """
 
+import argparse
 import os
 import sys
-from typing import Sequence, List, Dict, Any
-
+import json
+import threading
+from functools import partial
+from typing import Sequence, List, Any, Callable
+from benchmarks.benchmark_utils import str2bool
 import numpy as np
 import jax
 import psutil
-from absl import app
 from flax.training import train_state
 import flax.linen as nn
 from flax.linen import partitioning as nn_partitioning
 from transformers import AutoConfig, AutoModelForCausalLM
 from tqdm import tqdm
+from huggingface_hub import hf_hub_download, list_repo_files
+from safetensors import safe_open
 
+from orbax.checkpoint import type_handlers
 from MaxText import checkpointing
 from MaxText import max_logging
 from MaxText import max_utils
@@ -68,6 +86,13 @@ from MaxText.utils.ckpt_conversion.utils.param_mapping import HOOK_FNS, PARAM_MA
 from MaxText.utils.ckpt_conversion.utils.utils import apply_hook_fns, HF_IDS
 
 jax.config.update("jax_platform_name", "cpu")
+
+
+def print_ram_usage(stage=""):
+  memory = psutil.virtual_memory()
+  max_logging.log(
+      f"[{stage}] RAM Usage: {memory.used / (1024**3):.2f}/{memory.total / (1024**3):.2f} GB ({memory.percent:.1f}%)"
+  )
 
 
 class MemoryMonitorTqdm(tqdm):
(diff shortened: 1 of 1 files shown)


src/MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -25,6 +25,9 @@
                          Defaults to "./mt_output/".
   scan_layers: (bool) Whether the MaxText model was trained with scanned layers.
                This must match the training configuration of the checkpoint.
+  lazy_load: (bool) If True, uses an on-demand loading strategy to minimize RAM
+             usage during conversion. Recommended if, 2 * model_size (GB) >= system RAM
+             Defaults to False.
```
hengtaoguo, line 30: This is awesome!!

src/MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -103,8 +134,173 @@ def format_meter(
...
+    return self.size * self.dtype.itemsize
+
+  @property
+  def itemsize(self):
+    return self.dtype.itemsize
+
+  def __array__(self, dtype=None):
+    # This method is called just-in-time by Orbax when saving this specific leaf.
+    try:
+      arr = self._load_fn()
```
hengtaoguo, line 263: I'm very curious about what's happening under-the-hood.  
- If we are calling one tensor from the checkpoint, what is happening now? Does it only load that specific shard (0001 out of all shards)of the safetensors? 
- And after this tensor is loaded and transformed, do we release the memory for this safetensor shard? 
- What if a following tensor also belongs to the same shard, and does it required to load the same shard all over again? 

Thanks so much Mohit!

src/MaxText/configs/base.yml
```diff
@@ -67,6 +67,8 @@ checkpoint_storage_use_zarr3: True
 # larger models requires higher concurrent GB for I/O
 # default concurrent gb for PytreeCheckpointHandler is 96GB
 checkpoint_storage_concurrent_gb: 96
+# lazyily load tensors for checkpoint conversion
```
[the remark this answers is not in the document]
shuningjin: nit (optional): if this is only used in to_maxtext, maybe we can use local argument there like this
https://github.com/AI-Hypercomputer/maxtext/blob/188c4268ffcf41cb2ee7412239e9732b10a1ccf7/tests/forward_pass_logit_checker.py#L475

src/MaxText/utils/ckpt_conversion/to_maxtext.py
shuningjin: I am very curious as well.

src/MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -55,7 +68,18 @@
...
+import torch
+from huggingface_hub import hf_hub_download, list_repo_files
 
+# Try importing safetensors, but don't fail immediately if missing (fallback to torch)
+try:
+  from safetensors import safe_open
+
+  HAS_SAFETENSORS = True
+except ImportError:
+  HAS_SAFETENSORS = False
```
RissyRan: Do we still support torch checkpoint?

src/MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -103,8 +134,173 @@ def format_meter(
...
+    if "model.safetensors.index.json" in files:
+      self.is_safetensors = True
+      index_file = "model.safetensors.index.json"
+    elif "model.safetensors" in files:
+      self.is_safetensors = True
+      # Single file case
+      self.shard_map = {None: "model.safetensors"}
+      return
+    elif "pytorch_model.bin.index.json" in files:
+      index_file = "pytorch_model.bin.index.json"
```
RissyRan: is this support on purpose? I thought we plan to only support safetensors. Otherwise, the mapping could be different for the same model. We had some bad experience in llama4, and would like to get rid of this if possible.

src/MaxText/utils/ckpt_conversion/to_maxtext.py
RissyRan: +1, @khatwanimohit could you add some explanation in the code or some pointers? I think this help future reader too!

src/MaxText/utils/ckpt_conversion/to_maxtext.py
khatwanimohit: removed all the pytorch loading code

src/MaxText/configs/base.yml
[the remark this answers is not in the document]
khatwanimohit: thank you for suggesting this, removed that parameter from base.yml

src/MaxText/utils/ckpt_conversion/to_maxtext.py
khatwanimohit: removed all the pytorch loading code

src/MaxText/utils/ckpt_conversion/to_maxtext.py
khatwanimohit: I have added more comments around this. 
After loading the index file we identify which tensor belongs to which shard, when `get_tensor` is called, it opens the specific shard file using `safetensors.safe_open` which uses `get_tensor()` to load only that specific tensor from the shard. At this point only 1 tensor's worth of RAM is used.

src/MaxText/utils/ckpt_conversion/to_maxtext.py
```diff
@@ -340,16 +557,34 @@ def main(argv: Sequence[str]) -> None:
...
   max_logging.log(f"Conversion complete. Checkpoint saved to {output_directory}")
 
 
 if __name__ == "__main__":
-  app.run(main)
+  jax.config.update("jax_default_prng_impl", "unsafe_rbg")
+  os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"  # Suppress TensorFlow logging
+
+  parser = argparse.ArgumentParser()
+  parser.add_argument("--lazy_load_tensors", type=bool, default=False, help="Whether to use lazy loading of HF tensors.")
```
shuningjin: Thanks for the update! use `type=str2bool` instead of `type=bool`, and `required=False`
- otherwise `--lazy_load_tensors=False` is still evaluate to true, as "False" is a non-empty string
- from MaxText.inference_utils import str2bool
- ref: https://github.com/AI-Hypercomputer/maxtext/blob/ce7160339f55b470afdaf6a809da77d163629270/src/MaxText/utils/ckpt_scripts/convert_deepseek_family_ckpt.py#L714

RissyRan (approved):
Thank you!

src/MaxText/utils/ckpt_conversion/to_maxtext.py
khatwanimohit: thank you, updated

PR #2638, opened 2025-11-08
Temporarily remove profiler (de)activation metric logging

NuojCheng:
Introduce a new config `hide_profiler_step_metric` to omit metric logging for profiler start/stop and subsequent steps.

This change temporarily removes metric logging for four specific steps:
1.  The step where the profiler is activated.
2.  The step immediately following profiler activation.
3.  The step where the profiler is deactivated.
4.  The step immediately following profiler deactivation.

These steps produce outlier timings because profiler activation/deactivation overhead is currently included in the step time, which skews MFU and other benchmark metrics. This is a short-term fix to address the benchmarking inaccuracies described in b/456828037.

A long-term solution, moving the profiler to a separate subprocess to isolate its overhead, is planned.

# Tests

1. Tested with the command:
`smoke_train skip_first_n_steps_for_profiler=10 profiler_steps=3 profiler=xplane steps=20`

The log difference can be viewed here: https://diff.googleplex.com/#key=JIA9QUANVfdW

2. Tested with the command:
`smoke_train skip_first_n_steps_for_profiler=10 profiler_steps=3 profiler=xplane steps=20 hide_profiler_step_metric=true`

The log difference can be viewed here: https://diff.googleplex.com/#key=LbBZkoUsDAW5

# Checklist

### DIFF
--- a/src/MaxText/metric_logger.py
+++ b/src/MaxText/metric_logger.py
@@ -104,43 +104,93 @@ class MetricLogger:
   def log_metrics(self, metrics, step, is_training):
     """Logs metrics via max_logging."""
     if is_training:
-      loss = metrics["scalar"]["learning/loss"]
-      # Do not show flops and tokens during batch size rampup
-      if step >= self.config.rampup_end_step:
-        log_message = (
-            f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
-            f"TFLOP/s/device: {metrics['scalar']['perf/per_device_tflops_per_sec']:.3f}, "
-            f"Tokens/s/device: {metrics['scalar']['perf/per_device_tokens_per_sec']:.3f}, "
-            f"total_weights: {metrics['scalar']['learning/total_weights']}, "
-            f"loss: {loss:.3f}"
-        )
-      else:
-        log_message = (
-            "[Rampup Batch Size Phase]: "
-            f"completed step: {step}, seconds: {metrics['scalar']['perf/step_time_seconds']:.3f}, "
-            f"total_weights: {metrics['scalar']['learning/total_weights']}, "
-            f"loss: {loss:.3f}"
-        )
-
-      if self.config.mtp_num_layers > 0:
-        mtp_loss = metrics["scalar"].get("learning/mtp_loss", 0.0)
-        main_model_loss = loss - mtp_loss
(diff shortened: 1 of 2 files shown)


src/MaxText/metric_logger.py
```diff
@@ -104,43 +104,80 @@ def write_metrics(self, metrics, step, is_training=True):
...
-        )
-
-    max_logging.log(log_message)
+      self._log_eval_metrics(metrics, step)
+
+  def _log_training_metrics(self, metrics, step):
+    """Handles training-specific metric logging."""
+    # Skip logging if in profiler activation/deactivation steps
+    # TODO(b/456828037): Switch to subprocess profiling to avoid timing artifacts at boundary steps.
+    if self._is_profiler_boundary_step(step):
```
Obliviour: I think its worth adding a log saying "completed step: {step}  skipping step metrics due to profilier activation/deactivation"

NuojCheng: Updated! Not sure whether Ubench they use regex to grab MFU info from this or not. I am worried that outputs with different format may break their logic. Will check with them next week

bvandermoon (approved):

src/MaxText/metric_logger.py
```diff
@@ -104,43 +104,93 @@ def write_metrics(self, metrics, step, is_training=True):
...
+      self._log_eval_metrics(metrics, step)
+
+  def _log_training_metrics(self, metrics, step):
+    """Handles training-specific metric logging."""
+    # Skip logging if in profiler activation/deactivation steps
+    # TODO(b/456828037): Switch to subprocess profiling to avoid timing artifacts at boundary steps.
+    scalars = metrics["scalar"]
+    loss = scalars["learning/loss"]
+    is_rampup = step < self.config.rampup_end_step
+    is_metric_hidden_step = self.config.hide_profiler_step_metric & self._is_profiler_boundary_step(step)
```
gagika: Could you use `and` instead of `&` for boolean and?

e.g. either

self.config.hide_profiler_step_metric and self._is_profiler_boundary_step(step)

or 
bool(self.config.hide_profiler_step_metric) and self._is_profiler_boundary_step(step)

NuojCheng: Good catch! Updated. Thank you Gagik.

PR #2645, opened 2025-11-10
Enable micro batching with Tunix-managed profiling

xuefgu:
# Description

Enable micro batching with Tunix-managed profiling

# Tests

Manually tested 4 ways: w/ and w/o micro batching, w/ and w/o profiling

# Checklist

### DIFF
--- a/src/MaxText/configs/rl.yml
+++ b/src/MaxText/configs/rl.yml
@@ -81,6 +81,9 @@ batch_size: 1
 # Increase `batch_size` and `MAX_STEPS` for better results.
 # num_batches: 3738
 num_batches: 4  # 200
+# A batch can be split into multiple micro batches for memory management
+# and/or async sampling and training.
+micro_batch_size: -1
 # Keep `num_test_batches` low so that evaluation runs quickly. It can be
 # increased to a max. of 330 (if batch size is 4).
 num_test_batches: 5  # 200
--- a/src/MaxText/rl/train_rl.py
+++ b/src/MaxText/rl/train_rl.py
@@ -61,7 +61,7 @@ import tensorflow_datasets as tfds
 from tunix.rl import rl_cluster as rl_cluster_lib
 from tunix.rl.rollout import base_rollout
 from tunix.rl.grpo.grpo_learner import GrpoConfig, GrpoLearner
-from tunix.sft import metrics_logger
+from tunix.sft import metrics_logger, profiler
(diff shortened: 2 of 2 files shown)


src/MaxText/configs/rl.yml
```diff
@@ -81,6 +81,9 @@ batch_size: 1
 # Increase `batch_size` and `MAX_STEPS` for better results.
 # num_batches: 3738
 num_batches: 4  # 200
+# A batch can be split into multiple micro batches for memory management
+# and/or async sampling and training.
+micro_batch_size: -1
```
bvandermoon, line 86: Nit: We are overloading this term a bit since we already use [`micro_batch_size_to_train_on`](https://github.com/search?q=repo%3AAI-Hypercomputer%2Fmaxtext+micro_batch_size&type=code) in MaxText. Does RL also use that value?

richjames0: maybe it's fine as that one is more specific (just for training) and this is higher level (rl level) but could also add e.g. `rl_`. good spot!

src/MaxText/rl/train_rl.py
```diff
@@ -271,16 +273,24 @@ def rl_train(tmvp_config):
...
   max_logging.log(f"TensorBoard logs directory: {tmvp_config.tensorboard_dir}")
   # Metrics logger
   metrics_logging_options = metrics_logger.MetricsLoggerOptions(
       log_dir=tmvp_config.tensorboard_dir, flush_every_n_steps=tmvp_config.log_period
   )
 
-  # Profiler configurations
-  # TODO: xfgu@: add profiling
   profiler_options = None
+  if tmvp_config.profiler == "xplane":
```
richjames0, line 287: nit: in an ideal world this would be a separate commit

richjames0 (approved):
lgtm

src/MaxText/configs/rl.yml
xuefgu: That one is a derived config which isn't to be explicitly specified, whereas the one at hand is a direct mandate from the users. Good point though!

Also, I would prefer not adding the `rl_` prefix as it already lives in `rl.yaml`.

src/MaxText/rl/train_rl.py
xuefgu: Agreed.

PR #2683, opened 2025-11-13
Add audio encoder support

eitanporat:
# Description

Adding audio encoder support - in decoders.py 

# Tests

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -867,7 +867,9 @@ temperature_tuning: False
 
 # Multimodal flags
 use_multimodal: False
+use_audio: False
 freeze_vision_encoder_params: True
+freeze_audio_encoder_params: True
 dtype_mm: "float32"  # Data type for multimodal model's vision encoder
 remat_policy_for_vit: "minimal"  # Remat policy for multimodal model's vision encoder. Check `remat_policy` for options.
 image_size_for_vit: 896 # Default for Gemma3, and should be overwritten by model's config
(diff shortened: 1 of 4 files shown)


src/MaxText/configs/base.yml
```diff
@@ -867,7 +867,9 @@ temperature_tuning: False
 
 # Multimodal flags
 use_multimodal: False
+use_audio: False
```
aireenmei, line 870: It's a bit confusing that we have both use_multimodal and use_audio flags. Should we change use_multimodel to use_vision? Or is it possible to use one flag use_multimodal for all non-text modalities and depends on the provided input types to know audio/vision or both? Just thinking out loud. cc @hengtaoguo

src/MaxText/layers/decoders.py
```diff
@@ -569,6 +578,16 @@ def _apply_embedding(
       else:
         raise ValueError(f"Unsupported model_name for multimodal: {cfg.model_name}")
 
+    # Merge the audio embeddings with the text embeddings for qwen3-omni models
+    if audio_embeddings is not None and cfg.use_multimodal:
+      if cfg.model_name in ["qwen3-omni-30b-a3b"]:
```
aireenmei, line 583: Add a corresponding else block for unsupported model_name?

src/MaxText/configs/base.yml
eitanporat: @hengtaoguo said we should do it in a seperate PR.
i think it should be `use_image` `use_audio`

this makes the most sense imo

PR #2694, opened 2025-11-14
[tools/setup/setup.sh] Nightly cleanup

SamuelMarks:
# Description

[tools/setup/setup.sh] Cleanup redundant and unnecessary nightly dependencies ; use `--no-deps` ; minor other cleanup / DRYness

# Tests

CI and manual:
```sh
$ bash ./dependencies/scripts/docker_build_dependency_image.sh DEVICE='tpu' MODE='nightly'
```
```sh
$ bash ./dependencies/scripts/docker_build_dependency_image.sh DEVICE='tpu' MODE='stable'
```

# Checklist

### DIFF
--- a/tools/setup/setup.sh
+++ b/tools/setup/setup.sh
@@ -128,6 +128,10 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE || ($MODE == "nightly"
     exit 1
 fi
 
+if [[ -n "$LIBTPU_GCS_PATH" ]]; then
+    python3 -m uv pip install -U 'crcmod'
+fi
+
 if [[ $DEVICE == "tpu" ]]; then
     libtpu_path="$HOME/custom_libtpu/libtpu.so"
     if [[ "$MODE" == "libtpu-only" ]]; then
@@ -135,8 +139,6 @@ if [[ $DEVICE == "tpu" ]]; then
         if [[ -n "$LIBTPU_GCS_PATH" ]]; then
             # Install custom libtpu
             echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
-            # Install required dependency
-            python3 -m uv pip install -U crcmod
             # Copy libtpu.so from GCS path
             gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
             exit 0
@@ -147,69 +149,27 @@ if [[ $DEVICE == "tpu" ]]; then
     fi
 fi
 
-if [[ "$MODE" == "nightly" ]]; then
-    if [ "$DEVICE" = "gpu" ]; then
-        dep_name='dependencies/requirements/generated_requirements/cuda12-requirements.txt'
-    else
-        dep_name='dependencies/requirements/generated_requirements/'"${DEVICE?}"'-requirements.txt'
-    fi
-    printf 'Nightly mode: Installing "%s", stripping commit pins from git+ repos.\n' "$dep_name"
-    nightly_txt="${dep_name##*/}"
-    nightly_txt="${nightly_txt%.txt}"'-nightly-temp.txt'
-
-    # Create a temp file, strip commit pins from git+ repos in requirements.txt
-    # Remove/update this section based on the pinned github repo commit in requirements.txt
-    sed -E \
-      -e 's|^([^ ]*) @ https?://github.com/([^/]*\/[^/]*)/archive/.*\.zip$|\1@git+https://github.com/\2.git|' \
-      -e '/JetStream/d' \
-      -e '/mlperf-logging/d' \
-      "$dep_name" > "$nightly_txt"
-
-    echo "--- Installing modified nightly requirements: ---"
-    cat -- "$nightly_txt"
-    echo "-------------------------------------------------"
-    
-    python3 -m uv pip install --no-cache-dir -U -r "$nightly_txt" \
-                                                -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
-    rm -fv -- "$nightly_txt"
+if [ "$DEVICE" = "gpu" ]; then
+    dep_name='dependencies/requirements/generated_requirements/cuda12-requirements.txt'
 else
-    # stable or stable_stack mode: Install with pinned commits
-    if [ "$DEVICE" = "gpu" ]; then
-        dep_basename='cuda12-requirements.txt'
-    else
-        dep_basename="${DEVICE?}"'-requirements.txt'
-    fi
-
-    printf 'Installing "%s" with pinned commits.\n' "$dep_basename"
-    requirements_txt=
-    for candidate in 'dependencies/requirements/generated_requirements' 'dependencies/requirements' "${MAXTEXT_REPO_ROOT}"'/dependencies/requirements' "$PWD"; do
-      if [ -f "$candidate"'/'"$dep_basename" ]; then
-        requirements_txt="$candidate"'/'"$dep_basename"
-        break
-      else
-        searched="$searched"':'
-      fi
-    done
-    if [ -z "${requirements_txt}" ]; then
-      >&2 printf 'Could not find "%s", looked in: %s\n' "$dep_basename" "${searched%?}"
-      exit 2
-    else
-      python3 -m uv pip install --resolution=lowest -r "$requirements_txt" \
-                                                    -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
-    fi
+    dep_name='dependencies/requirements/generated_requirements/'"${DEVICE?}"'-requirements.txt'
 fi
+
+printf 'Installing "%s" with pinned commits.\n' "${dep_name##*/}"
+python3 -m uv pip install --resolution=lowest -r "$dep_name" \
+                                              -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
 
 # Install maxtext package
 if [ -f 'pyproject.toml' ]; then
   case "$DEVICE" in
-      'gpu') python3 -m uv pip install -e .[cuda12] --no-deps --resolution=lowest ;;
-      'tpu') python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest ;;
+      'gpu') python3 -m uv pip install -e .'[cuda12]' --no-deps --resolution='lowest' ;;
+      'tpu') python3 -m uv pip install -e .'[tpu]' --no-deps --resolution='lowest' ;;
       *)
           >&2 printf 'Unsupported device\n'
           exit 6
           ;;
   esac
-  python3 -m uv pip install --resolution=lowest -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
+  python3 -m uv pip install --resolution='lowest' -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
 fi
 
 # Delete custom libtpu if it exists
@@ -223,13 +183,12 @@ if [[ "$MODE" == "stable" || ! -v MODE ]]; then
         echo "Installing stable jax, jaxlib for tpu"
         if [[ -n "$JAX_VERSION" ]]; then
             echo "Installing stable jax, jaxlib, libtpu version ${JAX_VERSION}"
-            python3 -m uv pip install -U jax[tpu]==${JAX_VERSION} -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+            python3 -m uv pip install -U --no-deps 'jax[tpu]'=="${JAX_VERSION}" \
+                -f 'https://storage.googleapis.com/jax-releases/libtpu_releases.html'
         fi
         if [[ -n "$LIBTPU_GCS_PATH" ]]; then
             # Install custom libtpu
             echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
-            # Install required dependency
-            python3 -m uv pip install -U crcmod
             # Copy libtpu.so from GCS path
             gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
         fi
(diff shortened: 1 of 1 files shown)


tools/setup/setup.sh
```diff
@@ -261,42 +240,43 @@ elif [[ $MODE == "nightly" ]]; then
...
-        python3 -m uv pip install 'git+https://github.com/google/tunix.git'
-        echo "Installing jax-nightly, jaxlib-nightly"
-        # Install jax-nightly
-        python3 -m uv pip install --pre -U jax -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/
-        # Install jaxlib-nightly
-        python3 -m uv pip install --pre -U jaxlib -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/
+        if [[ -n "$JAX_VERSION" ]]; then
+            echo "Installing jax-nightly, jaxlib-nightly ${JAX_VERSION}"
+            python3 -m uv pip install --no-deps -U --pre jax=="${JAX_VERSION}" jaxlib=="${JAX_VERSION}" \
+                -i 'https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/'
```
bvandermoon, line 229: If JAX version is used, we will also need to use a libtpu installation from the same day so that it's compatible

bvandermoon (commented):
Thanks Samuel. Could you please also run a training for both of the `MODE` options being updated here? As well as with/without explicitly setting `JAX_VERSION`?

tools/setup/setup.sh
SamuelMarks: I can do that - hey do you have a working link https://us-python.pkg.dev/ml-oss-artifacts-published/jax-public-nightly-artifacts-registry/simple from https://docs.jax.dev/en/latest/installation.html#jax-nightly-installation

(I'm cross-referencing with https://storage.googleapis.com/jax-releases/libtpu_releases.html)

EDIT: Posted here also https://github.com/jax-ml/jax/issues/33341

bvandermoon:
> Thanks Samuel. Could you please also run a training for both of the `MODE` options being updated here? As well as with/without explicitly setting `JAX_VERSION`?

It would also be good to `pip show` that the versions are what you would expect for each build

tools/setup/setup.sh
```diff
@@ -220,36 +178,34 @@ fi
...
     if [[ $DEVICE == "tpu" ]]; then
-        # TODO: Once tunix has support for GPUs, move it from here to requirements.txt
-        echo "Installing google-tunix for stable TPU environment"
-        python3 -m uv pip install 'google-tunix>=0.1.2'
         echo "Installing stable jax, jaxlib for tpu"
         if [[ -n "$JAX_VERSION" ]]; then
             echo "Installing stable jax, jaxlib, libtpu version ${JAX_VERSION}"
-            python3 -m uv pip install -U jax[tpu]==${JAX_VERSION} -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+            python3 -m uv pip install -U jax[tpu]=="${JAX_VERSION}" \
+                -f 'https://storage.googleapis.com/jax-releases/libtpu_releases.html'
```
bvandermoon, line 187: Looks like this should be with `--no-deps`

tools/setup/setup.sh
```diff
@@ -220,36 +178,34 @@ fi
...
         fi
         if [[ -n "$LIBTPU_GCS_PATH" ]]; then
             # Install custom libtpu
             echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
             # Install required dependency
-            python3 -m uv pip install -U crcmod
+            python3 -m uv pip install -U 'crcmod'
```
bvandermoon: Should this be with `--no-deps`?

tools/setup/setup.sh
```diff
@@ -261,42 +217,43 @@ elif [[ $MODE == "nightly" ]]; then
...
             # Copy libtpu.so from GCS path
             gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
         else
             # Install libtpu-nightly
             echo "Installing libtpu-nightly"
-            python3 -m uv pip install -U --pre libtpu -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+            python3 -m uv pip install -U --pre 'libtpu' -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
```
bvandermoon: This should be with `--no-deps`

tools/setup/setup.sh
SurbhiJainUSC: Looks like this is needed only when `device=tpu` and `LIBTPU_GCS_PATH` is set. Can we remove all occurrences of `pip install -U crcmod` and move it top? It will lead to only one-time installation of this package.

SamuelMarks: Guarded by `if [[ -n "$LIBTPU_GCS_PATH" ]]` I assume; as it's not needed in other branches

tools/setup/setup.sh
bvandermoon: I think the second link you have (https://storage.googleapis.com/jax-releases/libtpu_releases.html) is the recommended way to install libtpu nightly at this point

tools/setup/setup.sh
```diff
@@ -220,83 +180,73 @@ fi
...
 elif [[ $MODE == "nightly" ]]; then
 # Nightly mode
 
     # Uninstall existing jax, jaxlib and  libtpu-nightly
-    python3 -m uv pip show jax && python3 -m uv pip uninstall jax
-    python3 -m uv pip show jaxlib && python3 -m uv pip uninstall jaxlib
-    python3 -m uv pip show libtpu-nightly && python3 -m uv pip uninstall libtpu-nightly
+    python3 -m uv pip show 'jax' && python3 -m uv pip uninstall 'jax'
+    python3 -m uv pip show 'jaxlib' && python3 -m uv pip uninstall 'jaxlib'
+    python3 -m uv pip show 'libtpu-nightly' && python3 -m uv pip uninstall 'libtpu-nightly'
```
bvandermoon, line 211: Does uninstalling update any transitive dependencies?

SamuelMarks: This is identical to previous functionality. `uv pip uninstall` does not have options to remove or not remove transitive dependencies. I don't believe it does, see for example:
```sh
(maxtext-uv-3-12) $ uv pip uninstall fastapi
Using Python 3.12.12 environment at: venvs/maxtext-uv-3-12
Uninstalled 1 package in 56ms
 - fastapi==0.119.1
(maxtext-uv-3-12) $ uv pip install fastapi
Using Python 3.12.12 environment at: venvs/maxtext-uv-3-12
Resolved 11 packages in 450ms
Prepared 2 packages in 32ms
Installed 2 packages in 4ms
 + annotated-doc==0.0.4
 + fastapi==0.121.2
```

tools/setup/setup.sh
SamuelMarks: Are you sure you want that, even for `libtpu`?

tools/setup/setup.sh
```diff
@@ -147,69 +149,27 @@ if [[ $DEVICE == "tpu" ]]; then
...
-      'gpu') python3 -m uv pip install -e .[cuda12] --no-deps --resolution=lowest ;;
-      'tpu') python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest ;;
+      'gpu') python3 -m uv pip install -e .'[cuda12]' --no-deps --resolution='lowest' ;;
+      'tpu') python3 -m uv pip install -e .'[tpu]' --no-deps --resolution='lowest' ;;
       *)
           >&2 printf 'Unsupported device\n'
           exit 6
           ;;
   esac
-  python3 -m uv pip install --resolution=lowest -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
```
SurbhiJainUSC, line 212: Why are we install deps from Github again? It looks redundant to me. We have already installed them at line 159, right?

tools/setup/setup.sh
SurbhiJainUSC: If `tpu-requirements.txt` were generated with JAX==0.8.0 (say) and then if we install `JAX_VERSION=0.7.2` without deps installation, then we are not even sure if it will work or not. There might be some dependency conflicts. How do we ensure that everything works?

tools/setup/setup.sh
SamuelMarks: Yes but this happens after. So caching in Docker is setup such that the first run doesn't have our package at all so it never hits that branch. Installing from `extra_deps_from_github.txt` in stage 1 is correct, then in stage 2 it may install from there again but will be cached so at most will override conflicts from `pip install` if `--no-deps` is ever removed. The duplication is intended.

SurbhiJainUSC:
> > Thanks Samuel. Could you please also run a training for both of the `MODE` options being updated here? As well as with/without explicitly setting `JAX_VERSION`?
> 
> It would also be good to `pip show` that the versions are what you would expect for each build

@SamuelMarks - Can you also test the setup when `MODE` and `JAX_VERSION` are specified?

tools/setup/setup.sh
```diff
@@ -223,75 +183,70 @@ if [[ "$MODE" == "stable" || ! -v MODE ]]; then
...
             echo "Installing libtpu.so from $LIBTPU_GCS_PATH to $libtpu_path"
-            # Install required dependency
-            python3 -m uv pip install -U crcmod
             # Copy libtpu.so from GCS path
             gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
         else
             # Install libtpu-nightly
             echo "Installing libtpu-nightly"
-            python3 -m uv pip install -U --pre libtpu -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+            python3 -m uv pip install -U --pre 'libtpu' \
```
SurbhiJainUSC, line 246: This should be with `--no-deps` right?

tools/setup/setup.sh
```diff
@@ -128,15 +128,17 @@ if [[ -n $JAX_VERSION && ! ($MODE == "stable" || -z $MODE || ($MODE == "nightly"
     exit 1
 fi
 
+if [[ -n "$LIBTPU_GCS_PATH" ]]; then
+    python3 -m uv pip install -U 'crcmod'
```
kanglant, line 132: Can you check if this dep is necessary? If so, please add a comment to explain why.

tools/setup/setup.sh
```diff
@@ -147,69 +149,27 @@ if [[ $DEVICE == "tpu" ]]; then
...
+printf 'Installing "%s" with pinned commits.\n' "${dep_name##*/}"
+python3 -m uv pip install --resolution=lowest -r "$dep_name" \
+                                              -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
+
 # Install maxtext package
 if [ -f 'pyproject.toml' ]; then
   case "$DEVICE" in
-      'gpu') python3 -m uv pip install -e .[cuda12] --no-deps --resolution=lowest ;;
-      'tpu') python3 -m uv pip install -e .[tpu] --no-deps --resolution=lowest ;;
+      'gpu') python3 -m uv pip install -e .'[cuda12]' --no-deps --resolution='lowest' ;;
```
kanglant, line 165: We'll need to add DEVICE=cuda12 and DEVICE=cuda13 values to support both CUDA versions. This is just a reminder for future work, not a change required for this PR.

tools/setup/setup.sh
kanglant: Do you mean that the `--no-deps` will be removed from the package installation commands in the future, so the second github deps installation command here is to act as a safeguard to override any conflicting package versions?

tools/setup/setup.sh
kanglant: As we discussed, if users choose to install a JAX version different from the one pinned by MaxText, we must present a warning. We cannot guarantee consistent dependency environments while offering this level of JAX version flexibility (maintaining lock files for every JAX/libtpu combination is not feasible).

Since JAX==0.8.0 is already included in tpu-requirements.txt, you don't need to install it again here. I recommend the following steps in this block:
1. Define Variables: Introduce variables here for the pinned JAX version (and optionally the corresponding libtpu version) used during requirement generation.
2. Conditional Installation: Only execute the installation logic if the requested JAX version differs from the pinned one.
3. Warning/Flag Choice: When installing the different version, you can either:
 * Install without `--no-deps` but issue a strong warning that the resulting environment may be inconsistent.
 * Install with `--no-deps` but issue a strong warning that the dependency versions may be incompatible with the new JAX version.

tools/setup/setup.sh
```diff
@@ -223,75 +183,70 @@ if [[ "$MODE" == "stable" || ! -v MODE ]]; then
...
 elif [[ $MODE == "nightly" ]]; then
 # Nightly mode
 
     # Uninstall existing jax, jaxlib and  libtpu-nightly
-    python3 -m uv pip show jax && python3 -m uv pip uninstall jax
-    python3 -m uv pip show jaxlib && python3 -m uv pip uninstall jaxlib
-    python3 -m uv pip show libtpu-nightly && python3 -m uv pip uninstall libtpu-nightly
+    python3 -m uv pip show 'jax' && python3 -m uv pip uninstall 'jax'
+    python3 -m uv pip show 'jaxlib' && python3 -m uv pip uninstall 'jaxlib'
+    python3 -m uv pip show 'libtpu-nightly' && python3 -m uv pip uninstall 'libtpu-nightly'
```
kanglant, line 211: Given that libtpu has nightly versions like `libtpu==0.0.32.dev20251210` available in the [index](https://storage.googleapis.com/jax-releases/libtpu_releases.html) , should we also uninstall existing `libtpu` package?

tools/setup/setup.sh
kanglant: Please explain. We use the `--no-deps` flag for JAX nightly artifacts, but not for the `libtpu` nightly. Could you explain the reasoning for this difference? Add a short comment in the script would be helpful for users.

tools/setup/setup.sh
SamuelMarks: Pre-existing dep. If it works without this you can remove this dep.

tools/setup/setup.sh
SamuelMarks: No because if libtpu depends on something then that version needs to be what it provides. Otherwise we'll be reporting bugs to libtpu that are not their bug but ours.

SurbhiJainUSC:
Comments will be addressed as part of https://github.com/AI-Hypercomputer/maxtext/pull/2833. Closing this PR.

PR #2703, opened 2025-11-17
Use threadpool when mixing multiple datasets in Grain

aireenmei:
# Description
Use theadpool to speed up by initializing multiple data sources in parallel. Also use Iterdataset.mix instead of mapdataset.mix to prepare for supporting more features.

# Tests
Manually added timing loggings around the changed block, tested 3 times in v5p, with 2 data sources, each 1024 files:
Before this change: 10.37s, 11.40s, 11.24s
After this change: 8.33s, 7.79s, 9.48s

# Checklist

### DIFF
--- a/src/MaxText/configs/base.yml
+++ b/src/MaxText/configs/base.yml
@@ -578,8 +578,15 @@ grain_eval_files: ''
 grain_file_type: 'arrayrecord' # arrayrecord or parquet
 grain_worker_count: 1
 grain_per_worker_buffer_size: 1
+# num_threads and prefetch_buffer_size are per-worker per-dataset. Used in ReadOptions (https://google-grain.readthedocs.io/en/latest/tutorials/data_loader_tutorial.html#per-worker-readoptions)
+# The default value matches that in the Grain package. If mixing multiple data sources, consider lowering these values to reduce memory usage.
+grain_num_threads: 16  
+grain_prefetch_buffer_size: 500
 grain_worker_count_eval: 1
 grain_per_worker_buffer_size_eval: 1
+grain_num_threads_eval: 16
+grain_prefetch_buffer_size_eval: 500
+grain_data_source_max_workers: 16  # Max workers for ThreadPoolExecutor when mixing multiple Grain data sources.
 # for using pathways
 colocated_python_data_input: False  # experimental feature, under testing
 
--- a/src/MaxText/input_pipeline/_grain_data_processing.py
+++ b/src/MaxText/input_pipeline/_grain_data_processing.py
@@ -18,6 +18,7 @@ import glob
 from pathlib import Path
 import functools
 import ml_collections
+from concurrent import futures
 
 import jax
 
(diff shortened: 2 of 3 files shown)


src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -56,23 +57,41 @@ def get_datasets(
...
+      executor.shutdown(wait=True)
+
+      # Apply shuffle, repeat, sharding, and conversion to IterDataset to each dataset before mixing
+      for d, _ in enumerate(dataset_list):
+        if shuffle:
+          dataset_list[d] = dataset_list[d].shuffle(seed=shuffle_seed)
+        dataset_list[d] = dataset_list[d].repeat(num_epoch)
+        dataset_list[d] = dataset_list[d][dataloading_host_index::dataloading_host_count]  # sharding
+        dataset_list[d] = dataset_list[d].to_iter_dataset()
+      # Use IterDataset.mix instead of MapDataset.mix
```
iindyk: comment here that this is to have per-mixture component checkpoint in order to later be able to change the mixture after checkpointing

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -56,23 +57,41 @@ def get_datasets(
...
+      dataset_list = list(executor.map(create_dataset_from_pattern, data_file_patterns))
+      executor.shutdown(wait=True)
+
+      # Apply shuffle, repeat, sharding, and conversion to IterDataset to each dataset before mixing
+      for d, _ in enumerate(dataset_list):
+        if shuffle:
+          dataset_list[d] = dataset_list[d].shuffle(seed=shuffle_seed)
+        dataset_list[d] = dataset_list[d].repeat(num_epoch)
+        dataset_list[d] = dataset_list[d][dataloading_host_index::dataloading_host_count]  # sharding
+        dataset_list[d] = dataset_list[d].to_iter_dataset()
```
iindyk: you may need to reduce the number of threads since you hade only one prefetch and buffer before, but now you have one per mixture component

iindyk: I would suggest running some representative benchmarks for this change

src/MaxText/input_pipeline/_grain_data_processing.py
aireenmei: Added

src/MaxText/input_pipeline/_grain_data_processing.py
aireenmei: I now make grain_num_threads and grain_prefetch_buffer_size configurable to account for different use cases. We will test after merging this PR. cc @bernardhan33

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -53,26 +54,57 @@ def get_datasets(
...
+      for d, _ in enumerate(dataset_list):
+        if shuffle:
+          dataset_list[d] = dataset_list[d].shuffle(seed=shuffle_seed)
+        dataset_list[d] = dataset_list[d].repeat(num_epoch)
+        dataset_list[d] = dataset_list[d][dataloading_host_index::dataloading_host_count]  # sharding
+        dataset_list[d] = dataset_list[d].to_iter_dataset(
+            read_options=grain.ReadOptions(
+                num_threads=grain_num_threads,
+                prefetch_buffer_size=grain_prefetch_buffer_size,
+            )
```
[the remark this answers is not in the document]
aireenmei: That's a good suggestion. I update it to be configurable

src/MaxText/configs/base.yml
```diff
@@ -578,8 +578,15 @@ grain_eval_files: ''
 grain_file_type: 'arrayrecord' # arrayrecord or parquet
 grain_worker_count: 1
 grain_per_worker_buffer_size: 1
+# Settings for per-worker ReadOptions (https://google-grain.readthedocs.io/en/latest/tutorials/data_loader_tutorial.html#per-worker-readoptions)
+# The default value matches that in the Grain package. If mixing multiple data sources, consider lowering these values to reduce memory usage.
+grain_num_threads: 16
```
RissyRan, line 583: If using mixing data sources, is this `grain_num_threads` per data source or in total? Potentially you could rename it to `grain_num_threads_per_dataset` to avoid potential confusion when using multiple data sources. Not a blocker change.

aireenmei: Thanks for the suggestion. I added clarification in the comments above that both "num_threads" and "prefetch_buffer_size" applies per-worker per-dataset

PR #2722, opened 2025-11-19
Fix hf_train_files and hf_eval_files after pydantic and fix model_mode=None in create_nnx_model

SurbhiJainUSC:
# Description

This PR fixes two issues:

1. https://github.com/AI-Hypercomputer/maxtext/pull/2612 introduced `model_mode` in `create_nnx_model()`, which was set to None when `create_nnx_model()` is called by SFT & GRPO trainer. This resulted in the following error:
```
/content/maxtext/src/MaxText/model_creation_utils.py in _create_model(mesh, model_mode, rng_key)
    130       rngs = nnx.Rngs(params=rng_key)  # disable dropout RNG for inference
    131 
--> 132     return from_config(config, devices, mesh, rngs=rngs, model_mode=model_mode)
    133 
    134   _create_model_partial = partial(_create_model, mesh=mesh, model_mode=model_mode, rng_key=rng_key)

/content/maxtext/src/MaxText/model_creation_utils.py in from_config(config, devices, mesh, model_mode, rngs)
     89     mesh = Mesh(devices_array, config.mesh_axes, axis_types=axis_types)
     90 
---> 91   model = create_model(config, mesh, model_mode=model_mode, rngs=rngs)
     92 
     93   # Return only the model
ValueError: Unknown model_mode: None
```

2. [Pydantic PR](https://github.com/AI-Hypercomputer/maxtext/commit/81d653c93f6868984090e5959a8e26c972d13fab) changed the `hf_train_files` and `hf_eval_files` to `""`, which was suppose to be `None` as default.

# Tests

Tested locally by running SFT and GRPO.

# Checklist

### DIFF
--- a/src/MaxText/model_creation_utils.py
+++ b/src/MaxText/model_creation_utils.py
@@ -117,7 +117,7 @@ def create_model(config, mesh, model_mode: str = MODEL_MODE_TRAIN, rngs: nnx.Rng
   return model
 
 
-def create_nnx_model(config, mesh=None, devices=None, model_mode=None, rng_key=None):
+def create_nnx_model(config, mesh=None, devices=None, model_mode=MODEL_MODE_TRAIN, rng_key=None):
   """Creates a NNX model with sharded parameters, possibly loading from a checkpoint."""
 
   def _create_model(mesh: Mesh | None = None, model_mode: str = MODEL_MODE_TRAIN, rng_key: jax.Array | None = None):
--- a/src/MaxText/pyconfig.py
+++ b/src/MaxText/pyconfig.py
@@ -97,11 +97,6 @@ def _prepare_for_pydantic(raw_keys: dict[str, Any]) -> dict[str, Any]:
   pydantic_kwargs = {}
   valid_fields = types.MaxTextConfig.model_fields.keys()
 
-  # This is a workaround for tests that use `dataset_type='hf'` but do not
-  # specify `tokenizer_type='huggingface'`, which they should.
-  if raw_keys.get("dataset_type") == "hf" and "tokenizer_type" not in raw_keys:
-    raw_keys["tokenizer_type"] = "huggingface"
-
   for key, value in raw_keys.items():
     if key not in valid_fields:
       logger.warning("Ignoring invalid/unsupported field from YAML/CLI: %s", repr(key))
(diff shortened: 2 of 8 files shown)


src/MaxText/model_creation_utils.py
```diff
@@ -131,7 +131,7 @@ def _create_model(mesh: Mesh | None = None, model_mode: str = MODEL_MODE_TRAIN,
 
     return from_config(config, devices, mesh, rngs=rngs, model_mode=model_mode)
 
-  _create_model_partial = partial(_create_model, mesh=mesh, model_mode=model_mode, rng_key=rng_key)
+  _create_model_partial = partial(_create_model, mesh=mesh, rng_key=rng_key)
```
khatwanimohit: I think this will break the vLLM path cause now are you not passing model_mode to `_create_model` and it will always be defaulted to `MODEL_MODE_TRAIN`?

src/MaxText/model_creation_utils.py
khatwanimohit: instead can you change the model_model here from None to `MODEL_MODE_TRAIN` and pass the model_mode to _create_model

src/MaxText/model_creation_utils.py
SurbhiJainUSC: VLLM is passing model_mode: http://shortn/_IxQLeHIkx2

khatwanimohit: If I'm understanding this correctly, vLLM is passing model_mode to `create_nnx_model` and with this change, we won't be passing that model_mode to _create_model ?

src/MaxText/model_creation_utils.py
SurbhiJainUSC: Updated.

src/MaxText/model_creation_utils.py
SurbhiJainUSC: Got it, updated.

src/MaxText/pyconfig.py
```diff
@@ -119,7 +119,10 @@ def _prepare_for_pydantic(raw_keys: dict[str, Any]) -> dict[str, Any]:
       if key == "data_sharding" and isinstance(new_value, list) and new_value and isinstance(new_value[0], str):
         new_value = [new_value]
 
-    if key in ("run_name", "hf_train_files", "hf_eval_files") and new_value is None:
+    if key in ("hf_train_files", "hf_eval_files") and new_value=="":
```
shralex: Could we add a comment explaining this, something like: 
Ensure that an empty value provided in the configuration is treated as "not present" (None), which aligns with the Optional[str] type and the default None set in types.py

mydatascience:
We need it for GRPO/GSPO to work

PR #2726, opened 2025-11-20
Add Qwen3 Audio Encoder

eitanporat:
# Description

- Add audio encoder support for qwen3
* classes such as `AudioEncoder` and `Qwen3OmniAudioEncoder`
* added the `use_audio` flag to the configuration.
* various audio specific flags

- added merging of audio tokens for qwen3
- added support in maxengine and maxtext_utils
- added support in decode.py

# Tests

tests/check_qwen3_omni_audio_vs_reference.py

# Checklist

### DIFF
--- a/src/MaxText/configs/models/qwen3-omni-30b-a3b.yml
+++ b/src/MaxText/configs/models/qwen3-omni-30b-a3b.yml
@@ -56,3 +56,24 @@ num_position_embeddings_for_vit: 2304
 deepstack_visual_indexes_for_vit: [8, 16, 24]
 
 use_multimodal: true
+use_audio: true
+# Audio Encoder Configuration (need to set use_audio=true to enable)
+# Based on https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
+d_model_for_audio: 1280
+encoder_layers_for_audio: 32
+encoder_attention_heads_for_audio: 20
+encoder_ffn_dim_for_audio: 5120
+max_source_positions_for_audio: 1500
+num_mel_bins_for_audio: 128
+downsample_hidden_size_for_audio: 480
+output_dim_for_audio: 2048
+attention_dropout_for_audio: 0.0
+n_window_for_audio: 50
+n_window_infer_for_audio: 400
+conv_chunksize_for_audio: 500
+num_conv_layers_for_audio: 3
+max_timescale_for_audio: 10000.0
+max_sample_len_for_audio: 10000
+
+freeze_audio_encoder_params: false
+freeze_vision_encoder_params: false
--- a/src/MaxText/decode.py
+++ b/src/MaxText/decode.py
@@ -150,6 +150,8 @@ def main(argv: Sequence[str]) -> None:
           padded_tokens=tokens,
           images=processor_outputs.pixel_values if config.use_multimodal else None,
           image_masks=processor_outputs.pixel_mask if config.use_multimodal and "llama4" in config.model_name else None,
+          audio_values=processor_outputs.audio_values if config.use_multimodal and config.use_audio else None,
+          audio_masks=processor_outputs.audio_mask if config.use_multimodal and config.use_audio else None,
           true_length=true_length,
           rng=rng_prefill,
           slot=i,
--- a/src/MaxText/layers/decoders.py
+++ b/src/MaxText/layers/decoders.py
@@ -557,6 +557,8 @@ class Decoder(nn.Module):
       image_embeddings=None,
       bidirectional_mask=None,
       image_masks=None,
+      audio_embeddings=None,
+      audio_mask=None,
   ):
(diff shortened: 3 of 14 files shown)


src/MaxText/configs/models/qwen3-omni-30b-a3b.yml
```diff
@@ -56,3 +56,24 @@ num_position_embeddings_for_vit: 2304
...
+attention_dropout_for_audio: 0.0
+n_window_for_audio: 50
+n_window_infer_for_audio: 400
+conv_chunksize_for_audio: 500
+num_conv_layers_for_audio: 3
+max_timescale_for_audio: 10000.0
+max_sample_len_for_audio: 10000
+
+freeze_audio_encoder_params: false
+freeze_vision_encoder_params: false
```
hengtaoguo, line 79: Can you remove these two flags from the model config? I want to control them in SFT config instead.

src/MaxText/layers/decoders.py
```diff
@@ -700,6 +713,8 @@ def __call__(
       image_masks: None | jnp.ndarray = None,
       kv_caches: list[jax.Array] | None = None,
       attention_metadata=None,
+      audio_embeddings: None | jnp.ndarray = None,
+      audio_mask: None | jnp.ndarray = None,
```
hengtaoguo, line 717: Can we change the name to plural `audio_masks` to align with `image_masks`?

src/MaxText/layers/embeddings.py
```diff
@@ -1431,3 +1460,52 @@ def __call__(self, num_frames: int, height: int, width: int) -> Array:
       interpolated = interpolated.astype(self.fprop_dtype)
 
     return interpolated
+  
+@dataclasses.dataclass(repr=False)
+class SinusoidsPositionEmbedding(nnx.Module):
```
hengtaoguo, line 1468: Can we reuse `PositionalEmbedding` above? It also implements classic sinusoidal position embeddings. Are there any differences between them?

src/MaxText/layers/encoders.py
```diff
@@ -66,4 +66,38 @@ def __call__(self, input_images, deterministic=False):
     if len(self.vision_encoder_layer) > 1:
       # vision embedder / projection layer, not frozen in most cases, trained / finetuned together with main model
       embeddings = self.vision_encoder_layer[1](config=cfg, mesh=mesh)(embeddings)
+
+
+class AudioEncoder(nn.Module):
```
hengtaoguo, line 71: Is it convenient for you to directly make it an NNX module? We are going to migrate VisionEncoder to NNX too.

src/MaxText/layers/qwen3.py
```diff
@@ -1641,6 +1642,357 @@ def qwen3omni_visionprojector_as_linen(config: Config, mesh: Mesh) -> nn.Module:
...
+    Returns:
+        Projected output of shape (num_chunks, seq_len, output_dim_for_audio)
+    """
+    hidden_states = self.proj1(hidden_states)
+    hidden_states = jax.nn.gelu(hidden_states)
+    hidden_states = self.proj2(hidden_states)
+    return hidden_states
+
+
+class Qwen3OmniAudioModel(nnx.Module):
```
hengtaoguo, line 1933: Hi why do we have an `AudioModel`? IIUC, in the AudioEncoder, you were calling `[qwen3.qwen3omni_audioencoder_as_linen, qwen3.qwen3omni_audioprojector_as_linen]`, just like we did for the vision encoder [encoder, projection].

src/MaxText/decode.py
```diff
@@ -150,6 +150,8 @@ def main(argv: Sequence[str]) -> None:
           padded_tokens=tokens,
           images=processor_outputs.pixel_values if config.use_multimodal else None,
           image_masks=processor_outputs.pixel_mask if config.use_multimodal and "llama4" in config.model_name else None,
+          audio_values=processor_outputs.audio_values if config.use_multimodal and config.use_audio else None,
+          audio_masks=processor_outputs.audio_mask if config.use_multimodal and config.use_audio else None,
```
hengtaoguo, line 154: I wonder if we can just use `use_audio` flag to control audio info?

Upon request we may rename `use_multimodal` to `use_vision` shortly. So these two flags should be decoupled, WDYT?

src/MaxText/layers/qwen3.py
eitanporat: yes it should just be Model

eitanporat: actually how would you control `freeze_vision_encoder_params` and `freeze_audio_encoder_params`
also deepstack is different

hengtaoguo: I mean this `AudioModel` was not used anywhere. You have been using its basic components `audio_encoder + audio_projector` (this is also our preferred way). So should we remove this AudioModel?

Usually we tend to freeze the encoder, and only tune the projector+decoder in SFT.

PR #2753, opened 2025-11-25
Fix tuple unpacking regression in Decoder Layers

gagika:
# Description

This PR fixes a regression introduced in a recent commit where decoder layers began returning a tuple `(hidden_states, kv_cache)` even when `scan_layers=False`. This caused failures in `SequentialBlockDecoderLayers` (used in pipeline parallelism) because the subsequent layer would receive a tuple as input instead of a tensor, leading to a `ValueError` in normalization layers.

### Changes
- Added a check at the beginning of the `__call__` method for all Decoder Layer implementations to handle tuple inputs gracefully:
  ```python
  # Unpack inputs if it's a tuple (e.g. from a previous layer returning (hidden_states, kv_cache))
  if isinstance(inputs, tuple):
    inputs = inputs[0]
    
 ### Changes
- Updated Qwen3NextScannableBlock to correctly unpack the (output, kv_cache) tuple inside its scan loop.

If the change fixes a bug:
FIXES: b/462751017

### Affected Models
This fix covers the following model architectures:
* DeepSeek (Dense & MoE)
* Gemma, Gemma 2, Gemma 3
* GPT-3, GPT-OSS
* Llama 2, Llama 4
* Mistral, Mixtral
* Qwen 3 (Dense, MoE, and Next)
* Simple Layer (Testing)

# Tests
internal google3 tests.

# Checklist

### DIFF
--- a/src/MaxText/layers/deepseek.py
+++ b/src/MaxText/layers/deepseek.py
@@ -243,7 +243,7 @@ class DeepSeekMoELayer(nn.Module):
       logical_axis_names = ("activation_batch", "prefill_activation_norm_length", "activation_embed")
     else:
       logical_axis_names = ("activation_batch", "activation_norm_length", "activation_embed")
-    
+
     # Unpack inputs if it's a tuple (e.g. from a previous layer returning (hidden_states, kv_cache))
     if isinstance(inputs, tuple):
       inputs = inputs[0]
--- a/src/MaxText/layers/mistral.py
+++ b/src/MaxText/layers/mistral.py
@@ -136,7 +136,7 @@ class MistralDecoderLayer(nnx.Module):
       attention_metadata=None,
   ):
     cfg = self.config
-    
+
     # Unpack inputs if it's a tuple (e.g. from a previous layer returning (hidden_states, kv_cache))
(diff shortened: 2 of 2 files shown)


NuojCheng:
Thank you Gagik for the fix! Can we add one more scan_layers=False smoke train test in `train_smoke_test.py`? It is hard to tell if the changes fixed the scan_layers issue...

NicoGrande:
Thanks for catching this Gagik! One quick question from my side -  would it make more sense to unpack the input in the `SequentialBlockDecoderLayers` `__call__` method? This way we maintain the assumption that inputs to individual decoder layers will always be Jax Arrays. It seems we already do something similar for scan layers.

NicoGrande:
> Thanks for catching this Gagik! One quick question from my side - would it make more sense to unpack the input in the `SequentialBlockDecoderLayers` `__call__` method? This way we maintain the assumption that inputs to individual decoder layers will always be Jax Arrays. It seems we already do something similar for scan layers.

On second thought, the approach in this PR is simpler to introduce in order to fix this regression! I'll rebase this onto main and add the smoke test suggested by @NuojCheng

PR #2774, opened 2025-12-02
Use MaxText max_segments_per_seq config variable to control Grain batch packing

gabeweisz:
# Description

When using THD format packed data with TransformerEngine, the user must specify the maximum number of segments that can be packed into a sequence at Jax JIT time. If grain packs more segments than allowed, then this can cause crashes or data corruption.

We have previously [updated grain](https://github.com/google/grain/commit/a1cb3ba314dc525533e14d89d4f3e02e09d65c13) to allow limiting the number of segments to pack into a sequence, and this PR takes the appropriate value from the MaxText configuration and passes it to Grain

# Tests

We have had this fix in place in [our AMD fork](https://github.com/rocm/maxtext) of MaxText for some time, but needed to get the Grain fix upstreamed first before creating this PR.
We have tested this fix extensively internally and have customers using it in production.

MaxText does not currently have any tests that use packed batches, but I can create some if needed.

# Checklist

### DIFF
--- a/src/MaxText/configs/types.py
+++ b/src/MaxText/configs/types.py
@@ -279,9 +279,7 @@ class Checkpointing(BaseModel):
   save_checkpoint_on_completion: bool = Field(
       True, description="If True, saves a final checkpoint upon training completion."
   )
-  enable_continuous_checkpointing: bool = Field(
-      False, description="If True, enables continuous checkpointing."
-  )
+  enable_continuous_checkpointing: bool = Field(False, description="If True, enables continuous checkpointing.")
 
 
 class OrbaxStorage(BaseModel):
@@ -463,9 +461,7 @@ class Attention(BaseModel):
   ragged_block_size: int = Field(256, description="Block size for ragged attention.")
   enable_padding_causal_mask: bool = Field(True, description="Temporary flag for TE padding.")
   use_tokamax_splash: bool = Field(False, description="Whether to use tokamax splash attention.")
-  use_jax_splash: bool = Field(
-      False, description="Whether to use jax splash attention."
-  )
+  use_jax_splash: bool = Field(False, description="Whether to use jax splash attention.")
 
 
 class MoBa(BaseModel):
@@ -880,8 +876,8 @@ class DatasetGeneral(BaseModel):
       description="Packing type when using Grain pipeline. 'first_fit' or 'concat_then_split'.",
   )
   max_segments_per_seq: int = Field(
-      32,
-      description="Maximum number of segments that can be packed into a single sequence.",
+      -1,
+      description="Maximum number of segments that can be packed into a single sequence. -1 or None for no limit.",
   )
   num_epoch: int = Field(1, description="Number of epochs to train for.")
   expansion_factor_real_data: float = Field(-1.0, description="Factor for partial data loading on hosts.")
@@ -2065,6 +2061,8 @@ class MaxTextConfig(
         raise ValueError(
             "Ring context parallelism strategy (context_parallel_strategy='ring') is only supported on GPUs."
         )
+    if self.hardware == "gpu" and self.packing and self.attention == "cudnn_flash_te" and self.max_segments_per_seq <= 0:
+      raise ValueError("max_segments_per_seq must be set when using TransformerEngine attention and packing")
     dcn_product = (
         self.dcn_data_parallelism
         * self.dcn_pipeline_parallelism
--- a/src/MaxText/input_pipeline/_grain_data_processing.py
+++ b/src/MaxText/input_pipeline/_grain_data_processing.py
@@ -242,9 +242,15 @@ def pretrain_preprocessing_pipeline(
 
   if config.packing:
     length_struct = {col: config.max_target_length for col in data_columns}
+    max_segments = config.max_segments_per_seq
+    if max_segments is not None and max_segments <= 0:
+      max_segments = None
     if config.grain_packing_type == "first_fit":
       dataset = grain.experimental.FirstFitPackIterDataset(
-          dataset, length_struct=length_struct, num_packing_bins=batch_size
+          dataset,
+          length_struct=length_struct,
+          num_packing_bins=batch_size,
+          max_sequences_per_bin=max_segments,
       )
     elif config.grain_packing_type == "concat_then_split":
       if config.add_bos and hasattr(tokenizer_model, "bos_id"):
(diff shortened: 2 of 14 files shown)


yeandy (commented):
We may need to add `max_sequences_per_bin=config.max_segments_per_seq` in `make_hf_eval_iterator` too

gabeweisz:
> We may need to add `max_sequences_per_bin=config.max_segments_per_seq` in `make_hf_eval_iterator` too

Done, thanks for the tip

src/MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -386,6 +388,7 @@ def make_hf_train_iterator(
         use_sft=config.use_sft,
         sft_train_on_completion_only=config.sft_train_on_completion_only,
         chat_template_path=config.chat_template_path,
+        max_sequences_per_bin=config.max_segments_per_seq,
```
aireenmei: According to this [PR](https://github.com/AI-Hypercomputer/maxtext/commit/24548581a29be52967957135d54e9fd539fb3bb2), max_segment_per_seq is only relevant for GPU packed attention. But this change will apply it to TPU workloads as well.
To be cleaner, it's better to align the behavior across hardware, and across different pipelines (grain pipeline's FirstFitPackIterDataset also has this parameter). We can set the default value to -1, which means no limit (passing None to PackAndBatchOperation)

gabeweisz: I don't quite follow what you have in mind.
On GPUs, we need to have the same value for TransformerEngine attention and grain, or we will have data corruption.
That is why we're using the same config value.

Are you suggesting that we set the default value for this configuration value [to -1 here](https://github.com/AI-Hypercomputer/maxtext/blob/3d9378d77759a7756e20ae2940ce71dcaa17ef13/src/MaxText/configs/base.yml#L551), and update the comment to say that on GPUs this should be changed to something like 32?
I can make that change (and turn it to -1 here) if that is what is desired. If we do that I can also add a warning if TE attention is used and max_segments_per_seq is not updated

aireenmei: Yes, how about we:
1. Set the default value to -1 in base.yml and a comment why it needs to be set in case of gpu, and a suggested value
2. Under the [validation section](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/configs/types.py#L1974) in type.py, add a check to make sure it's not -1 when hardware=='gpu', and other conditions if there are.
3. In all related gpu workload configs, add max_segment_per_seq=32

Wdyt?

gabeweisz: Done and thank you for the feedback

aireenmei:
Can you also update here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/configs/types.py#L882?

And apply max_sequences_per_bin here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/input_pipeline/_grain_data_processing.py#L247

gabeweisz:
> Can you also update here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/configs/types.py#L882?
> 
> And apply max_sequences_per_bin here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/input_pipeline/_grain_data_processing.py#L247

Done

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -244,7 +244,10 @@ def pretrain_preprocessing_pipeline(
     length_struct = {col: config.max_target_length for col in data_columns}
     if config.grain_packing_type == "first_fit":
       dataset = grain.experimental.FirstFitPackIterDataset(
-          dataset, length_struct=length_struct, num_packing_bins=batch_size
+          dataset,
+          length_struct=length_struct,
+          num_packing_bins=batch_size,
+          max_sequences_per_bin=config.max_segments_per_seq,
```
aireenmei: max_sequences_per_bin=config.max_segments_per_seq if config.max_segments_per_seq > 0 else None

gabeweisz: Right, thank you for catching this

gabeweisz:
Pyink made all of those whitespace changes in the last commit - those were not made manually and I would have prefer not to have them but the code quality checks required them. Commit 7f70fa6135ca220642eb46e2c71f9ecb4e56ddd1 has everything before pyink

src/MaxText/configs/types.py
```diff
@@ -279,9 +279,7 @@ class Checkpointing(BaseModel):
   save_checkpoint_on_completion: bool = Field(
       True, description="If True, saves a final checkpoint upon training completion."
   )
-  enable_continuous_checkpointing: bool = Field(
-      False, description="If True, enables continuous checkpointing."
-  )
+  enable_continuous_checkpointing: bool = Field(False, description="If True, enables continuous checkpointing.")
```
gabeweisz, line 282: pyink did this

src/MaxText/configs/types.py
```diff
@@ -463,9 +461,7 @@ class Attention(BaseModel):
   ragged_block_size: int = Field(256, description="Block size for ragged attention.")
   enable_padding_causal_mask: bool = Field(True, description="Temporary flag for TE padding.")
   use_tokamax_splash: bool = Field(False, description="Whether to use tokamax splash attention.")
-  use_jax_splash: bool = Field(
-      False, description="Whether to use jax splash attention."
-  )
+  use_jax_splash: bool = Field(False, description="Whether to use jax splash attention.")
```
gabeweisz, line 464: pyink did this

gabeweisz:
@aireenmei I apologize but I needed to make some updates for the code quality checks (they didn't run before your approval). Can you please take another look. These were minor edits.

aireenmei:
Could you take a look to see if this error is related: https://github.com/AI-Hypercomputer/maxtext/actions/runs/20354898076/job/58493556179?pr=2774

aireenmei:
I've asked @SurbhiJainUSC to also take a look. @gabeweisz you will need to squash your commits into 1 before merging

gabeweisz:
@aireenmei thanks for the feedback - I have done the squash commit
@SurbhiJainUSC please take a look - I need a review from a reviewer with write access

gabeweisz:
Looks like I still need an approval from a maintainer. How can I get one?

Issue #2543, opened 2025-10-24
Multi-epoch training differs from single-epoch training in the first epoch when using grain with parquet


emergenz:
### Bug report

Essentially the culprit is the shuffling logic in the grain parquet codepath of maxtext:
https://github.com/AI-Hypercomputer/maxtext/blob/58dba0a78f8ca34e20a778892b2ce4409f2067d8/src/MaxText/input_pipeline/_grain_data_processing.py#L70-L84

I just quickly wanted to open the issue before I forget. If you need more information, let me know.

```bash
python3 -m MaxText.train \
  MaxText/configs/base.yml \
  run_name=${job_name}_${slurm_job_id} \
  base_output_directory=$BASE_OUTPUT_DIRECTORY \
  dataset_type=grain \
  grain_file_type=parquet \
  grain_train_files=$GRAIN_TRAIN_FILES \
  grain_eval_files=$GRAIN_EVAL_FILES \
  eval_interval=500 \
  steps=50000 \
  skip_jax_distributed_system=True
```

```bash
python3 -m MaxText.train \
  MaxText/configs/base.yml \
  run_name=${job_name}_${slurm_job_id} \
  base_output_directory=$BASE_OUTPUT_DIRECTORY \
  dataset_type=grain \
  grain_file_type=parquet \
  grain_train_files=$GRAIN_TRAIN_FILES \
  grain_eval_files=$GRAIN_EVAL_FILES \
  eval_interval=500 \
  steps=50000 \
  num_epoch=10 \
  skip_jax_distributed_system=True
```

### Logs/Output

<img width="1782" height="694" alt="Image" src="https://github.com/user-attachments/assets/83ed7a7b-0d72-487a-ac4e-567436db3e8d" />

Pink is the multi-epoch run.

### Environment Information

HEAD: `3789aa17d7d3cf88c3a4db23c97127b1c2005bfc`
Single H100-SXM5
```
Name: jax
Version: 0.7.0
Summary: Differentiate, compile, and transform Numpy code.
Home-page: https://github.com/jax-ml/jax
Author: JAX team
Author-email: jax-dev@google.com
License: Apache-2.0
Location: /fast/home/franz.srambical/crowd-pilot/maxtext/.venv/lib/python3.12/site-packages
Requires: jaxlib, ml_dtypes, numpy, opt_einsum, scipy
[...]

PR #2785, opened 2025-12-04
fix bug in grain pipeline and refactor

aireenmei:
# Description

FIXES: https://github.com/AI-Hypercomputer/maxtext/issues/2543
Also turn off windowshuffle when enable_shuffling=False

# Tests
Manually tested

# Checklist

### DIFF
--- a/src/MaxText/input_pipeline/_grain_data_processing.py
+++ b/src/MaxText/input_pipeline/_grain_data_processing.py
@@ -47,6 +47,30 @@ def find_data_files(data_file_pattern):
   return data_files
 
 
+def _apply_mapdataset_transforms(
+    dataset,
+    shuffle,
+    shuffle_seed,
+    num_epoch,
+    dataloading_host_index,
+    dataloading_host_count,
+    grain_num_threads,
+    grain_prefetch_buffer_size,
+):
+  """Apply standard shuffle, repeat, shard, and iter conversion transforms."""
+  if shuffle:
+    dataset = dataset.shuffle(seed=shuffle_seed)
+  dataset = dataset.repeat(num_epoch)
+  dataset = dataset[dataloading_host_index::dataloading_host_count]  # sharding
+  dataset = dataset.to_iter_dataset(
+      read_options=grain.ReadOptions(
+          num_threads=grain_num_threads,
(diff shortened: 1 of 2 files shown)


src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -152,8 +152,11 @@ def create_dataset_from_pattern(pattern):
         f"Please lower grain_worker_count or increase file shard count."
     )
     dataset = dataset.map(grain.experimental.ParquetIterDataset)
-    dataset = grain.experimental.InterleaveIterDataset(dataset, cycle_length=len(dataset))
-    dataset = grain.experimental.WindowShuffleIterDataset(dataset, window_size=100, seed=shuffle_seed)
+    if shuffle:
+      dataset = grain.experimental.InterleaveIterDataset(dataset, cycle_length=len(dataset) // num_epoch)
```
iindyk: this is pretty much parallelism parameter -- how many files to read in parallel and interleave the results. I would set the cycle length to either `grain_num_threads` or default to something like 16 (Grain will still be able to handle the cases when len(datasets) is smaller).

src/MaxText/input_pipeline/_grain_data_processing.py
```diff
@@ -152,8 +152,11 @@ def create_dataset_from_pattern(pattern):
...
+      dataset = grain.experimental.WindowShuffleIterDataset(dataset, window_size=100, seed=shuffle_seed)
+    else:
+      dataset = grain.experimental.InterleaveIterDataset(dataset, cycle_length=1)  # files are read sequentially when cycle_length=1
```
iindyk: I would still interleave the files even in non-shuffle scenario -- this would help with performance

aireenmei: done

src/MaxText/input_pipeline/_grain_data_processing.py
aireenmei: Thanks for explanation. Synced offline, I now use min(dataset_len_one_epoch, grain_num_threads) for cycle length, grain_num_threads defaults to 16

NicoGrande (approved):

PR #2833, opened 2025-12-15
Add versioning support for nightly

SurbhiJainUSC:
# Description

* Add versioning support for MODE=nightly
* Cleanup setup.sh
Reference: https://github.com/AI-Hypercomputer/maxtext/pull/2694

# Tests

https://paste.googleplex.com/6430443719229440

# Checklist

### DIFF
--- a/tools/setup/setup.sh
+++ b/tools/setup/setup.sh
@@ -14,15 +14,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Description:
-# bash setup.sh MODE={stable,nightly,libtpu-only} LIBTPU_GCS_PATH={gcs_path_to_custom_libtpu} DEVICE={tpu,gpu}
+# ==================================
+# TPU EXAMPLES
+# ==================================
 
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt
+## bash tools/setup/setup.sh MODE=stable
 
-# You need to specify a MODE, default value stable.
-# You have the option to provide a LIBTPU_GCS_PATH that points to a libtpu.so provided to you by Google.
-# In libtpu-only MODE, the LIBTPU_GCS_PATH is mandatory.
-# For MODE=stable you may additionally specify JAX_VERSION, e.g. JAX_VERSION=0.4.13
-# For DEVICE=gpu, you may also specify JAX_VERSION when MODE=nightly, e.g. JAX_VERSION=0.4.36.dev20241109
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt + specified jax, jaxlib, libtpu
+## bash tools/setup/setup.sh MODE=stable JAX_VERSION=0.8.0
+
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt + custom libtpu
+## bash tools/setup/setup.sh MODE=stable LIBTPU_GCS_PATH=gs://my_custom_libtpu/libtpu.so
+
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt + jax-nightly, jaxlib-nightly, libtpu-nightly
+## bash tools/setup/setup.sh MODE=nightly
+
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt + specified jax-nightly, jaxlib-nightly + latest libtpu-nightly
+## bash tools/setup/setup.sh MODE=nightly JAX_VERSION=0.8.2.dev20251211
+
+# Install dependencies in dependencies/generated_requirements/tpu-requirements.txt + jax-nightly, jaxlib-nightly + custom libtpu
+## bash tools/setup/setup.sh MODE=nightly LIBTPU_GCS_PATH=gs://my_custom_libtpu/libtpu.so
+
+# Install custom libtpu only
+## bash tools/setup/setup.sh MODE=libtpu-only LIBTPU_GCS_PATH=gs://my_custom_libtpu/libtpu.so
+
+# ==================================
+# GPU EXAMPLES
+# ==================================
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt
+## bash tools/setup/setup.sh MODE=stable DEVICE=gpu
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax, jaxlib, jax-cuda12-plugin, jax-cuda12-pjrt
+## bash tools/setup/setup.sh MODE=stable DEVICE=gpu JAX_VERSION=0.4.13
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + jax-nightly, jaxlib-nightly
+## bash tools/setup/setup.sh MODE=nightly DEVICE=gpu
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax, jaxlib, jax-cuda12-plugin, jax-cuda12-pjrt
+## bash tools/setup/setup.sh MODE=nightly DEVICE=gpu JAX_VERSION=0.4.36.dev20241109
 
 
 # Enable "exit immediately if any command fails" option
@@ -102,7 +133,6 @@ apt update -y && apt -y install gcsfuse
 rm -rf /var/lib/apt/lists/*
 EOF
 
-
 python3 -m pip install -U setuptools wheel uv
 
 # Set environment variables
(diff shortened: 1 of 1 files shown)


tools/setup/setup.sh
```diff
@@ -14,15 +14,43 @@
...
+## bash tools/setup/setup.sh MODE=libtpu-only LIBTPU_GCS_PATH=gs://my_custom_libtpu/libtpu.so
+
+# ==================================
+# GPU EXAMPLES
+# ==================================
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt
+## bash tools/setup/setup.sh MODE=stable DEVICE=gpu
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax, jaxlib
```
kanglant:
```suggestion
# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax, jaxlib, jax-cuda12-plugin, jax-cuda12-pjrt
```

tools/setup/setup.sh
```diff
@@ -14,15 +14,43 @@
...
+## bash tools/setup/setup.sh MODE=stable DEVICE=gpu JAX_VERSION=0.4.13
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + jax-nightly, jaxlib-nightly
+## bash tools/setup/setup.sh MODE=nightly DEVICE=gpu
+
+# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax, jaxlib
```
kanglant:
```suggestion
# Install dependencies in dependencies/generated_requirements/cuda12-requirements.txt + specified jax nightly artifacts
```

tools/setup/setup.sh
```diff
@@ -14,15 +14,43 @@
...
+## bash tools/setup/setup.sh MODE=nightly DEVICE=gpu JAX_VERSION=0.4.36.dev20241109
```
kanglant, line 56: Does the MaxText team plan to support a `TPU + nightly` configuration where we can also specify the JAX+libtpu nightly versions? I was under the impression that we would at least want to support this in Docker images.

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
+    gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
+}
 
-if [[ "$MODE" == "nightly" ]]; then
-    if [ "$DEVICE" = "gpu" ]; then
-        dep_name='dependencies/requirements/generated_requirements/cuda12-requirements.txt'
-    else
-        dep_name='dependencies/requirements/generated_requirements/'"${DEVICE?}"'-requirements.txt'
+install_maxtext_with_deps() {
+    if [[ "$DEVICE" != "tpu" && "$DEVICE" != "tpu" ]]; then
```
kanglant:
```suggestion
    if [[ "$DEVICE" != "tpu" && "$DEVICE" != "gpu" ]]; then
```

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
-        searched="$searched"':'
-      fi
-    done
-    if [ -z "${requirements_txt}" ]; then
-      >&2 printf 'Could not find "%s", looked in: %s\n' "$dep_basename" "${searched%?}"
-      exit 2
-    else
-      python3 -m uv pip install --resolution=lowest -r "$requirements_txt" \
-                                                    -r 'src/install_maxtext_extra_deps/extra_deps_from_github.txt'
+    if [ -f 'pyproject.toml' ]; then
```
kanglant, line 206: Could you add a comment explaining why the package is installed separately from the dependencies? It would be helpful to mention that this is to optimize Docker image rebuild times by leveraging the cache and to include a Dockerfile example.

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
+if [[ "$MODE" == "stable" ]]; then
+    install_maxtext_with_deps
 
-if [[ "$MODE" == "stable" || ! -v MODE ]]; then
-# Stable mode
     if [[ $DEVICE == "tpu" ]]; then
-        echo "Installing stable jax, jaxlib for tpu"
         if [[ -n "$JAX_VERSION" ]]; then
             echo "Installing stable jax, jaxlib, libtpu version ${JAX_VERSION}"
+            echo -e "\n\nWARNING: You are installing a JAX version that is different from the one pinned by MaxText."
```
kanglant: Please also note that when using a JAX version other than the one pinned, we don't guarantee a consistent Python environment over time, as dependency versions may vary with each installation.

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
-            # Install required dependency
-            python3 -m uv pip install -U crcmod
-            # Copy libtpu.so from GCS path
-            gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
+            install_custom_libtpu
         fi
     elif [[ $DEVICE == "gpu" ]]; then
-        echo "Installing stable jax, jaxlib for NVIDIA gpu"
         if [[ -n "$JAX_VERSION" ]]; then
             echo "Installing stable jax, jaxlib ${JAX_VERSION}"
```
kanglant: Should we add the same warning here for the GPU case as we have for the TPU case regarding unpinned JAX versions?

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
-        export NVTE_FRAMEWORK=jax
-        if [[ -n "$JAX_VERSION" && "$JAX_VERSION" != "0.7.0" ]]; then
-            python3 -m uv pip install transformer-engine[jax]
-        else
-            python3 -m uv pip install git+https://github.com/NVIDIA/TransformerEngine.git@9d031f
+
+            export NVTE_FRAMEWORK=jax
+            if [[ "$JAX_VERSION" == "0.7.0" ]]; then
+                python3 -m uv pip install --no-deps git+https://github.com/NVIDIA/TransformerEngine.git@9d031f
+            fi
```
kanglant: @Rohan-Bierneni Is this section still necessary? I believe maxtext has pinned a version of `transformer-engine`, which includes this commit, in the requirements file.

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
+    if [[ "$DEVICE" != "tpu" ]]; then
+      echo -e "\n\nError: MODE=libtpu-only is only supported for DEVICE=tpu.\n\n"
+      exit 1
+    fi
+    if [[ -z "$LIBTPU_GCS_PATH" ]]; then
+        echo -e "\n\nError: LIBTPU_GCS_PATH must be set when MODE is libtpu-only.\n\n"
+        exit 1
+    fi
+    install_custom_libtpu
 fi
```
kanglant, line 285: Should we add a check for invalid `$MODE` values to ensure the script fails gracefully if an unsupported mode is used?

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
-            # Install libtpu-nightly
             echo "Installing libtpu-nightly"
-            python3 -m uv pip install -U --pre libtpu -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+            python3 -m uv pip install -U --pre --no-deps libtpu -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+        fi
+    elif [[ $DEVICE == "gpu" ]]; then
+        if [[ -n "$JAX_VERSION" ]]; then
+            echo "Installing jax-nightly, jaxlib-nightly ${JAX_VERSION}"
+            python3 -m uv pip install -U --pre --no-deps jax==${JAX_VERSION} jaxlib==${JAX_VERSION} \
+                jax-cuda12-plugin[with-cuda] jax-cuda12-pjrt -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/
```
kanglant:
```suggestion
                jax-cuda12-plugin[with-cuda]==${JAX_VERSION} jax-cuda12-pjrt==${JAX_VERSION} -i https://us-python.pkg.dev/ml-oss-artifacts-published/jax/simple/
```

kanglant: Jax cuda artifacts are usually released along with jax and jaxlib.

tools/setup/setup.sh
```diff
@@ -111,188 +138,131 @@ for ARGUMENT in "$@"; do
...
+        else
+            echo "Installing latest jax-nightly, jaxlib-nightly"
```
kanglant:
```suggestion
            echo "Installing the latest jax-nightly, jaxlib-nightly"
```

tools/setup/setup.sh
SurbhiJainUSC: Isn't this same as note at line 206?

tools/setup/setup.sh
SurbhiJainUSC: Added support for `JAX_VERSION` for TPU nightly mode.

tools/setup/setup.sh
Rohan-Bierneni: I think we can remove this since at the time TE had not released a specific change in stable version so we had to pin to a specific commit. Also, if maxtext is pinning TE at specific commit, then this can be removed as well.

SurbhiJainUSC: Cool, removed this.

tools/setup/setup.sh
kanglant: I believe there are two separate issues we need to communicate to the user:
1. Compatibility: The dependencies we install from the requirements file are tested and compatible with the pinned JAX version. We cannot guarantee that they will work correctly with a different JAX version that a user chooses to install here.
2. Consistency: Installing a custom JAX version can pull in different transitive dependencies over time, making the environment non-reproducible and potentially affecting performance. The current warning covers compatibility, and we should address the consistency issue as well.

tools/setup/setup.sh
```diff
@@ -111,188 +141,148 @@ for ARGUMENT in "$@"; do
...
-
-if [[ "$MODE" == "stable" || ! -v MODE ]]; then
-# Stable mode
     if [[ $DEVICE == "tpu" ]]; then
-        echo "Installing stable jax, jaxlib for tpu"
         if [[ -n "$JAX_VERSION" ]]; then
-            echo "Installing stable jax, jaxlib, libtpu version ${JAX_VERSION}"
+            echo -e "\nInstalling stable jax, jaxlib, libtpu version ${JAX_VERSION}"
+            version_mismatch_warning "jax"
             python3 -m uv pip install -U jax[tpu]==${JAX_VERSION} -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
```
Rohan-Bierneni: Nit. Jax now bundles the libtpu version in when installing via jax[tpu]. We can update this with a more concise install command:
```
python3 -m uv pip install -U jax[tpu]==${JAX_VERSION}
```

tools/setup/setup.sh
```diff
@@ -111,188 +141,148 @@ for ARGUMENT in "$@"; do
...
+    version_mismatch_warning "libtpu"
+    # Delete custom libtpu if it exists
+    if [ -e "$libtpu_path" ]; then
+        rm -v "$libtpu_path"
     fi
-fi
+    # Install 'crcmod' to download 'libtpu.so' from GCS reliably
+    python3 -m uv pip install -U crcmod
+    # Copy libtpu.so from GCS path
+    gsutil cp "$LIBTPU_GCS_PATH" "$libtpu_path"
```
Rohan-Bierneni, line 182: Can we add a check in the beginning of the script to ensure the user has already installed the necessary tools to run this script (gsutil, gcsfuse, curl, git, etc). Just in case the user forgot to install and is running on non-gcp vm or barebones vm, if the script fails at the end after most of the install is done its a bad user experience.

tools/setup/setup.sh
SurbhiJainUSC: Great, updated the script.

tools/setup/setup.sh
SurbhiJainUSC: We are already installing these requirements at line 125 in this script.

PR #2873, opened 2025-12-22
Improve max_logging to use absl logging.

NicoGrande:
# Description

I noticed `max_logging` was simply calling `print` under the hood, which lacks a lot of useful context like caller line number, logging severity and more. This PR updates max_logging to use absl logging instead. The `log` method is preserved for backwards compatibility.

# Tests

Regular MaxText tests.

# Checklist

### DIFF
--- a/src/MaxText/max_logging.py
+++ b/src/MaxText/max_logging.py
@@ -12,8 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Stub for logging utilities. Right now just meant to avoid raw prints."""
+"""Logging utilities."""
+from absl import logging
 
 
 def log(user_str):
-  print(user_str, flush=True)
+  """Logs a message at the INFO level."""
+  # Note, stacklevel=2 makes the log show the caller of this function.
+  logging.info(user_str, stacklevel=2)
+
+
+def debug(user_str):
+  """Logs a message at the DEBUG level."""
+  logging.debug(user_str, stacklevel=2)
+
+
+def info(user_str):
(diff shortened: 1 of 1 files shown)


src/MaxText/max_logging.py
```diff
@@ -12,8 +12,30 @@
...
+  """Logs a message at the INFO level."""
+  logging.info(user_str, stacklevel=2)
+
+
+def debug(user_str):
+  """Logs a message at the DEBUG level."""
+  logging.debug(user_str, stacklevel=2)
+
+
+def info(user_str):
```
gobbleturk, line 30: to confirm this has identical implementation to `log` -its just an alias?

src/MaxText/max_logging.py
```diff
@@ -12,8 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Stub for logging utilities. Right now just meant to avoid raw prints."""
+"""Logging utilities."""
+from absl import logging
 
 
 def log(user_str):
```
gobbleturk, line 19: I notice all use stacklevel=2, what does stacklevel represent? Could you include a comment about what stacklevel means in at least one of these methods?

src/MaxText/max_logging.py
NicoGrande: Yeah exactly - usually you call `info` directly.

src/MaxText/max_logging.py
NicoGrande: Will do!

PR #2831, opened 2025-12-15
Migrate Decoder to NNX

hsuan-lun-chiang:
# Description
Migrate the Transformer decoder layer into NNX.

Note: The following models are currently not supported:

- DeepSeek
- Gemma3
- Llama4

Support for these models will be added in a follow-up PR.

Strategy:
A `pure_nnx_decoder` flag is added to control whether NNX or Linen decoder shall be used.
Initial migration doesn't include the pipeline NNX support. 

# Tests
Conducted these tests. Details in the [GDoc file](https://docs.google.com/document/d/1NbUP3g5glgbC6bMyt44pwM_vQA1NR7U2rBUzfbTDwSs/edit?pli=1&resourcekey=0-9EUahtzL-hCycdu7l0grhQ&tab=t.htq5367h8au0)
1. Test with different model and compare with Linen training
2. Golden logits comparison
3. Inference
4. Checkpoint comparison (Including TreeStructure Comparison)
5. Sharding comparison

TODOs:
- NNX version of unit tests (future PRs)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #2885, opened 2025-12-24
feat: migrate pipeline to nnx

mesakhcienet:
# Description

implement nnx-based pipeline.

This PR extends [PR#2831](https://github.com/AI-Hypercomputer/maxtext/pull/2831)

Main changes:
1. nnx_decoders.py:  implementing the missing pipeline logic in nnx_decoders.py.
2. pipeline.py : add a new class `NNXPipeline`, which is a nnx-based pipeline class.

# Tests

we run the pipeline process with command below:

```bash
MODEL_NAME=llama2-7b
python -m MaxText.train src/maxtext/configs/base.yml \
    run_name=pipeline_test_${MODEL_NAME}_nnx \
    base_output_directory=/dev/shm/pipeline_test_nnx \
    model_name=${MODEL_NAME}\
    dataset_type=synthetic \
    steps=15 \
    debug_sharding=true \
    per_device_batch_size=2 \
    max_target_length=32 \
    ici_pipeline_parallelism=2 \
    num_pipeline_microbatches=4 \
    num_layers_per_pipeline_stage=2 \
    enable_checkpointing=false \
    enable_nnx=true \
    pure_nnx_decoder=true \
    scan_layers_per_stage=false \
    async_checkpointing=false > nnx-porting-log/pipeline/custom_${MODEL_NAME}.log 2>&1

```

# Checklist

### DIFF
--- a/src/maxtext/layers/decoders.py
+++ b/src/maxtext/layers/decoders.py
@@ -33,6 +33,7 @@ from maxtext.layers import linears
 from maxtext.layers import mhc
 from maxtext.layers import normalizations
 from maxtext.layers import pipeline
+from maxtext.layers.nnx_decoders import NNXDecoderLayer, NNXSequentialPipelineStage, NNXScannedPipelineStage
 from maxtext.layers import quantizations
 from maxtext.layers.attentions import attention_as_linen
 from maxtext.layers.embeddings import attend_on_embedding, embed_as_linen, positional_embedding_as_linen
@@ -260,7 +261,7 @@ class SequentialBlockDecoderLayers(nn.Module):
           slot=slot,
       )
       if self.config.scan_layers:
-        inputs = inputs[0]  #  When scan_layers is True the decoder layers return (outputs, None).
+        inputs = inputs[0]  # When scan_layers is True the decoder layers return (outputs, None).
     if self.config.scan_layers:
       return inputs, None  # pytype: disable=bad-return-type
     else:
@@ -305,10 +306,19 @@ class Decoder(nn.Module):
     self.decoder_layer = self.get_decoder_layers()
     self.norm_layer = self.get_norm_layer(num_features=self.config.emb_dim)
     if self.config.using_pipeline_parallelism:
-      pipeline_stage_module = self.get_pipeline_stage_module(self.decoder_layer)
       remat_policy = self.get_remat_policy()
+      nnx_blocks = self._get_nnx_decoder_block_classes()
+
+      # Per-stage builder handed to create_pipeline: the pipeline transform invokes it
+      # once per pipeline stage, passing that stage's rngs, to construct the stage's
+      # decoder block(s). nnx_blocks (the selected decoder block classes) is captured from
+      # this setup scope; remat_policy is applied separately by create_pipeline.
+      def build_pipeline_stage_layers(rngs):
+        """Builds one pipeline stage module from the selected NNX decoder block classes."""
+        return self._build_nnx_pipeline_stage(nnx_blocks, rngs)
+
       self.pipeline_module = pipeline.create_pipeline(
-          config=self.config, mesh=self.mesh, layers=pipeline_stage_module, remat_policy=remat_policy
+          config=self.config, layers=build_pipeline_stage_layers, mesh=self.mesh, remat_policy=remat_policy
       )
 
   def minimal_policy(self, with_context=False, with_quantization=False):
@@ -498,6 +508,42 @@ class Decoder(nn.Module):
         # Default case to handle any unknown decoder block types.
         raise ValueError(f"Incorrect decoder_block name {self.config.decoder_block.value=}")
 
+  def _get_nnx_decoder_block_classes(self):
+    """Returns NNX decoder block classes for pipeline stage creation."""
+    cfg = self.config
+
+    def get_scannable(normal_cls, scannable_cls):
+      return [scannable_cls] if cfg.scan_layers else [normal_cls]
+
+    layer_map = {
+        DecoderBlockType.DEFAULT: [NNXDecoderLayer],
+        DecoderBlockType.LLAMA2: [llama2.LlamaDecoderLayer],
+        DecoderBlockType.MISTRAL: [mistral.MistralDecoderLayer],
+        DecoderBlockType.MIXTRAL: [mixtral.MixtralDecoderLayer],
+        DecoderBlockType.GEMMA: [gemma.GemmaDecoderLayer],
+        DecoderBlockType.GEMMA2: [gemma2.Gemma2DecoderLayer],
+        DecoderBlockType.GEMMA3: [gemma3.Gemma3DecoderLayer],
+        DecoderBlockType.GEMMA4: get_scannable(gemma4.Gemma4DecoderLayer, gemma4.Gemma4ScannableBlock),
+        DecoderBlockType.GEMMA4_SMALL: [gemma4_small.Gemma4SmallDecoderLayer],
+        DecoderBlockType.GPT3: [gpt3.Gpt3DecoderLayer],
+        DecoderBlockType.GPT_OSS: get_scannable(gpt_oss.GptOssDecoderLayer, gpt_oss.GptOssScannableBlock),
+        DecoderBlockType.QWEN2: [qwen2.Qwen2DecoderLayer],
+        DecoderBlockType.QWEN3: [qwen3.Qwen3DecoderLayer],
+        DecoderBlockType.QWEN3_MOE: [qwen3.Qwen3MoeDecoderLayer],
+        DecoderBlockType.QWEN3_CUSTOM_MOE: [qwen3_custom.Qwen3CustomMoeDecoderLayer],
+        DecoderBlockType.QWEN3_5: get_scannable(qwen3_5.Qwen3_5DecoderLayer, qwen3_5.Qwen3_5ScannableBlock),
+        DecoderBlockType.QWEN3_NEXT: get_scannable(qwen3.Qwen3NextDecoderLayer, qwen3.Qwen3NextScannableBlock),
+        DecoderBlockType.SIMPLE: [simple_layer.SimpleDecoderLayer],
+        DecoderBlockType.SIMPLE_MLP: [simple_layer.SimpleMlpDecoderLayer],
+        DecoderBlockType.DEEPSEEK: [deepseek.DeepSeekDenseLayer, deepseek.DeepSeekMoELayer],
+        DecoderBlockType.LLAMA4: get_scannable(llama4.Llama4DecoderLayer, llama4.Llama4ScannableBlock),
+        DecoderBlockType.OLMO3: get_scannable(olmo3.Olmo3DecoderLayer, olmo3.Olmo3ScannableBlock),
+    }
+
+    if cfg.decoder_block not in layer_map:
+      raise ValueError(f"Incorrect decoder_block name {cfg.decoder_block.value=}")
+    return layer_map[cfg.decoder_block]
+
   def set_remat_policy(self, block_layers, policy):
     """Set remat policy"""
     RemattedBlockLayers = []
@@ -525,6 +571,58 @@ class Decoder(nn.Module):
       )
       RemattedBlockLayers.append(layer)
     return RemattedBlockLayers
+
+  def _build_nnx_pipeline_stage(self, decoder_blocks, rngs):
+    """Creates a single NNX pipeline stage module."""
+    cfg = self.config
+    base_stage_cls = decoder_blocks[1] if cfg.decoder_block == DecoderBlockType.DEEPSEEK else decoder_blocks[0]
+
+    if cfg.num_layers_per_pipeline_stage == 1:
+      return base_stage_cls(config=cfg, mesh=self.mesh, quant=self.quant, model_mode=self.model_mode, rngs=rngs)
+    elif cfg.scan_layers_per_stage:
+      return NNXScannedPipelineStage(
+          base_stage_cls, cfg.num_layers_per_pipeline_stage, cfg, self.mesh, self.quant, self.model_mode, rngs=rngs
+      )
+    return NNXSequentialPipelineStage(
+        base_stage_cls, cfg.num_layers_per_pipeline_stage, cfg, self.mesh, self.quant, self.model_mode, rngs=rngs
+    )
+
+  def get_pipeline_stage_module(self, decoder_blocks):
+    """get pipeline stage module"""
+
+    def get_layer_to_pipeline(blocks, cfg):
+      if cfg.decoder_block == DecoderBlockType.DEEPSEEK:
+        return blocks[1]  # return the sparse block
+      else:
+        return blocks[0]
+
+    cfg = self.config
+    base_stage = get_layer_to_pipeline(decoder_blocks, cfg)
+    if cfg.set_remat_policy_on_layers_per_stage:
+      policy = self.get_remat_policy()
+      base_stage = self.set_remat_policy([base_stage], policy)[0]
+    if cfg.num_layers_per_pipeline_stage == 1:
+      stage_module = base_stage(config=cfg, mesh=self.mesh, quant=self.quant, model_mode=self.model_mode)
+    elif cfg.scan_layers_per_stage:
+      stage_module = self.scan_decoder_layers(
+          cfg,
+          base_stage,
+          cfg.num_layers_per_pipeline_stage,
+          "layers_per_stage",
+          self.mesh,
+          in_axes_tuple=(nn.broadcast,) * 4,
+          model_mode=self.model_mode,
+      )
+    else:
+      stage_module = SequentialBlockDecoderLayers(
+          decoder_layer=base_stage,
+          num_decoder_layers=cfg.num_layers_per_pipeline_stage,
+          config=cfg,
+          mesh=self.mesh,
+          quant=self.quant,
+          model_mode=self.model_mode,
+      )
+    return stage_module
 
   def get_norm_layer(self, num_features: int):
     """get normalization layer (return type inherits from nn.Module)"""
@@ -585,42 +683,6 @@ class Decoder(nn.Module):
     return scan_fn(
         config=cfg, mesh=mesh, name=metadata_axis_name, quant=self.quant, **kwargs  # pytype: disable=wrong-keyword-args
     )
-
-  def get_pipeline_stage_module(self, decoder_blocks):
-    """get pipeline stage module"""
-
-    def get_layer_to_pipeline(blocks, cfg):
-      if cfg.decoder_block == DecoderBlockType.DEEPSEEK:
-        return blocks[1]  # return the sparse block
-      else:
-        return blocks[0]
-
-    cfg = self.config
-    base_stage = get_layer_to_pipeline(decoder_blocks, cfg)
-    if cfg.set_remat_policy_on_layers_per_stage:
-      policy = self.get_remat_policy()
-      base_stage = self.set_remat_policy([base_stage], policy)[0]
-    if cfg.num_layers_per_pipeline_stage == 1:
-      stage_module = base_stage(config=cfg, mesh=self.mesh, quant=self.quant, model_mode=self.model_mode)
-    elif cfg.scan_layers_per_stage:
-      stage_module = self.scan_decoder_layers(
-          cfg,
-          base_stage,
-          cfg.num_layers_per_pipeline_stage,
-          "layers_per_stage",
-          self.mesh,
-          in_axes_tuple=(nn.broadcast,) * 4,
-      )
-    else:
-      stage_module = SequentialBlockDecoderLayers(
-          decoder_layer=base_stage,
-          num_decoder_layers=cfg.num_layers_per_pipeline_stage,
-          config=cfg,
-          mesh=self.mesh,
-          quant=self.quant,
-          model_mode=self.model_mode,
-      )
-    return stage_module
 
   @nn.compact
   def _apply_embedding(
@@ -1301,10 +1363,16 @@ class Decoder(nn.Module):
         decoder_positions,
         deterministic,
         model_mode,
-        slot,
-        previous_chunk,
-        bidirectional_mask,
     )
+
+    # Pass slot/previous_chunk/bidirectional_mask by keyword only (via layer_call_kwargs),
+    # never positionally in broadcast_args: Gemma4DecoderLayer and Gemma4ScannableBlock
+    # declare slot and previous_chunk in swapped order, so positional passing misroutes them.
+    layer_call_kwargs = {
+        "slot": slot,
+        "previous_chunk": previous_chunk,
+        "bidirectional_mask": bidirectional_mask,
+    }
 
     if num_full_blocks > 0:
       ScannableBlockToLinen = gemma4.Gemma4ScannableBlockToLinen
@@ -1335,7 +1403,7 @@ class Decoder(nn.Module):
           num_of_layers=block_pattern_len,
           name="scanned_blocks",
       )(
-          y, *broadcast_args
+          y, *broadcast_args, **layer_call_kwargs
       )
 
     # Process any remaining layers that don't fit into a full scanned block
@@ -1349,7 +1417,7 @@ class Decoder(nn.Module):
           attention_type=attention_type,
           layer_idx=layer_id,
       )
-      y = layer(y, *broadcast_args)
+      y = layer(y, *broadcast_args, **layer_call_kwargs)
       if cfg.scan_layers:
         y = y[0]
 
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -69,6 +69,7 @@ from maxtext.multimodal import utils as mm_utils
 from maxtext.utils import max_logging, max_utils, maxtext_utils, sharding
 from maxtext.utils.maxtext_utils_nnx import nnx_ensure_scan_leading_axis
 from maxtext.utils.sharding import create_sharding
+from maxtext.layers.pipeline import create_nnx_pipeline
 
 # ------------------------------------------------------------------------------
 # The network: Decoder Definitions
@@ -252,6 +253,126 @@ def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
   return hidden_states
 
 
+class NNXSequentialPipelineStage(nnx.Module):
+  """Sequential unscanned series of decoder layers formatted for a single pipeline stage."""
+
+  def __init__(
+      self,
+      layer_cls,
+      num_layers: int,
+      config: Config,
+      mesh: Mesh,
+      quant: Quant,
+      model_mode: str,
+      *,
+      rngs: nnx.Rngs,
+  ):
+    self.config = config
+    self.scan_layers = config.scan_layers
+    self.num_layers = num_layers
+    # Dynamically assign layers with explicit string names to ensure correct PyTree paths (layers_0)
+    for i in range(num_layers):
+      layer = layer_cls(config=config, mesh=mesh, quant=quant, model_mode=model_mode, rngs=rngs)
+      setattr(self, f"layers_{i}", layer)
+
+  def __call__(
+      self,
+      inputs,
+      decoder_segment_ids,
+      decoder_positions,
+      deterministic,
+      model_mode,
+      **kwargs,
+  ):
+    for i in range(self.num_layers):
+      layer = getattr(self, f"layers_{i}")
+      out = layer(
+          inputs,
+          decoder_segment_ids,
+          decoder_positions,
+          deterministic,
+          model_mode,
+          **kwargs,
+      )
+      inputs = out[0] if isinstance(out, tuple) else out
+    if self.scan_layers:
+      return inputs, None
+    return inputs
+
+
+class NNXScannedPipelineStage(nnx.Module):
+  """Scanned block of decoder layers formatted for a single pipeline stage."""
+
+  def __init__(
+      self,
+      layer_cls,
+      num_layers: int,
+      config: Config,
+      mesh: Mesh,
+      quant: Quant,
+      model_mode: str,
+      *,
+      rngs: nnx.Rngs,
+  ):
+    self.config = config
+
+    def create_layer_fn(rng):
+      return layer_cls(config=config, mesh=mesh, quant=quant, model_mode=model_mode, rngs=rng)
+
+    forked_rngs = rngs.fork(split=num_layers)
+
+    out_axes = nnx.StateAxes({nnx.Param: config.param_scan_axis, ...: 0})
+    self.scanned_layers = nnx.vmap(
+        create_layer_fn,
+        in_axes=0,
+        out_axes=out_axes,
+        axis_name="layers_per_stage",
+        transform_metadata={nnx.PARTITION_NAME: "layers_per_stage"},
+    )(forked_rngs)
+
+  def __call__(
+      self,
+      inputs,
+      decoder_segment_ids,
+      decoder_positions,
+      deterministic,
+      model_mode,
+      **kwargs,
+  ):
+    graphdef, params, state = nnx.split(self.scanned_layers, nnx.Param, ...)
+
+    scan_axis = self.config.param_scan_axis
+    if scan_axis != 0:
+      params = jax.tree.map(lambda x: jnp.moveaxis(x, scan_axis, 0), params)
+
+    def layer_fn(carry, scanned_vars):
+      current_params, current_state = scanned_vars
+      layer = nnx.merge(graphdef, current_params, current_state)
+      layer_out = layer(
+          carry,
+          decoder_segment_ids,
+          decoder_positions,
+          deterministic,
+          model_mode,
+          **kwargs,
+      )
+      new_carry = layer_out[0] if isinstance(layer_out, tuple) else layer_out
+      return new_carry, nnx.state(layer)
+
+    final_carry, scanned_state = jax.lax.scan(layer_fn, inputs, (params, state))
+
+    if scan_axis != 0:
+      scanned_params, scanned_other = scanned_state.split(nnx.Param, ...)
+      scanned_params = jax.tree.map(lambda x: jnp.moveaxis(x, 0, scan_axis), scanned_params)
+      scanned_state = nnx.State.merge(scanned_params, scanned_other)
+
+    nnx.update(self.scanned_layers, scanned_state)
+
+    if self.config.scan_layers:
+      return final_carry, None
+    return final_carry
+
+
 class NNXDecoder(nnx.Module):
   """A stack of decoder layers as a part of an encoder-decoder architecture, using NNX."""
 
@@ -324,10 +445,85 @@ class NNXDecoder(nnx.Module):
       if config.using_pipeline_parallelism or config.scan_layers:
         raise ValueError("gemma4_small (Gemma4 E2B/E4B) does not support pipeline parallelism or scan_layers.")
       self._init_gemma4_small_layers(rngs)
+    elif config.using_pipeline_parallelism:
+      self._init_pipeline_layers(decoder_block_classes, rngs, mesh)
     elif config.scan_layers:
       self._init_scanned_layers(decoder_block_classes, rngs, mesh)
     else:
       self._init_sequential_layers(decoder_block_classes, rngs)
+
+  def _init_pipeline_layers(self, decoder_block_classes, rngs, mesh):
+    """Initializes decoder layers with pipeline parallelism."""
+    config = self.config
+    assert not (config.engram_layers and self.is_deepseek), (
+        "engram_layers + DeepSeek + pipeline_parallelism is not supported. "
+        "engram interleaving is currently only implemented in the non-pipeline path."
+    )
+
+    def build_pipeline_stage_layers(rngs):
+      return self._get_pipeline_stage_module(decoder_block_classes, rngs)
+
+    self.pipeline_module = create_nnx_pipeline(
+        config=config,
+        stage_factory=build_pipeline_stage_layers,
+        mesh=mesh,
+        remat_policy=self.get_remat_policy(),
+        rngs=rngs,
+    )
+
+    if self.is_deepseek:
+      self._init_pipeline_deepseek(decoder_block_classes, rngs)
+    else:
+      self._init_pipeline_generic(decoder_block_classes, rngs)
+
+  def _init_pipeline_deepseek(self, decoder_block_classes, rngs):
+    """Initializes DeepSeek dense and MoE layers outside pipeline."""
+    config = self.config
+    assert len(decoder_block_classes) == 2
+    dense_cls, moe_cls = decoder_block_classes
+    if config.scan_layers:
+      self.dense_layers = self._create_scanned_layers(
+          dense_cls,
+          length=config.first_num_dense_layers,
+          metadata_axis_name="dense_layers",
+          rngs=rngs,
+      )
+      num_moe_outside = (config.num_decoder_layers - config.first_num_dense_layers) - config.pipeline_parallel_layers
+      if num_moe_outside > 0:
+        self.moe_layers_outside_pipeline = self._create_scanned_layers(
+            moe_cls,
+            length=num_moe_outside,
+            metadata_axis_name="moe_layers",
+            rngs=rngs,
+        )
+    else:
+      self.num_dense_layers = config.first_num_dense_layers
+      for i in range(self.num_dense_layers):
+        self._create_and_register_named_layer(dense_cls, rngs, "dense_layers", i)
+      self.num_moe_outside_pipeline = (
+          config.num_decoder_layers - config.first_num_dense_layers
+      ) - config.pipeline_parallel_layers
+      if self.num_moe_outside_pipeline > 0:
+        for i in range(self.num_moe_outside_pipeline):
+          self._create_and_register_named_layer(moe_cls, rngs, "moe_layers_outside_pipeline", i)
+
+  def _init_pipeline_generic(self, decoder_block_classes, rngs):
+    """Initializes generic decoder layers outside pipeline."""
+    config = self.config
+    remaining_layers = config.num_decoder_layers - config.pipeline_parallel_layers
+    if remaining_layers > 0:
+      base_cls = decoder_block_classes[0]
+      if config.scan_layers:
+        self.layers_outside_pipeline = self._create_scanned_layers(
+            base_cls,
+            length=remaining_layers,
+            metadata_axis_name="layers",
+            rngs=rngs,
+        )
+      else:
+        self.num_layers_outside_pipeline = remaining_layers
+        for i in range(self.num_layers_outside_pipeline):
+          self._create_and_register_named_layer(base_cls, rngs, "layers_outside_pipeline", i)
 
   def _init_scanned_layers(self, decoder_block_classes, rngs, mesh):
     """Initializes decoder layers with scanning (non-pipeline)."""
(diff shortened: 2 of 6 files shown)


src/maxtext/layers/nnx_decoders.py
```diff
@@ -215,29 +222,92 @@ def __call__(
 
 
 def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
-  """Process deepstack visual embeddings by adding them to hidden states at visual token positions.
+  """Process deepstack visual embeddings by adding them to hidden states at visual token positions."""
+  mask_expanded = bidirectional_mask[:, :, jnp.newaxis]
```
charlesli640, line 243: Why should we remove previous comments?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -373,7 +481,6 @@ def _create_and_register_layer(self, layer_cls, rngs, base_name, i, **layer_kwar
     self.layers.append(layer)
 
   def _create_single_layer(self, decoder_layer_class, rngs, **kwargs):
-    """Helper to create a single layer (Linen or NNX)."""
     if issubclass(decoder_layer_class, nnx.Module):
```
charlesli640, line 815: No need to remove original comments

src/maxtext/layers/nnx_decoders.py
```diff
@@ -385,22 +492,14 @@ def _create_single_layer(self, decoder_layer_class, rngs, **kwargs):
       return nnx_wrappers.ToNNX(layer_linen, rngs=rngs)
 
   def _create_scanned_layers(self, decoder_layer_class, length: int, rngs: nnx.Rngs, **layer_kwargs):
```
charlesli640: No need to remove original comments

charlesli640 (changes requested):
1. do code format clean up
2. run pre-commit checking
3. Don't make unnecessary changes, maybe introduced from rebase and incorrect conflict solving.

bvandermoon (commented):
@gobbleturk what testing do you recommend for migrating pipeline parallelism to NNX? I'll send over an internal doc @hsuan-lun-chiang, @mesakhcienet, and others put together that shows the tests they have already run

bvandermoon:
> @gobbleturk what testing do you recommend for migrating pipeline parallelism to NNX? I'll send over an internal doc @hsuan-lun-chiang, @mesakhcienet, and others put together that shows the tests they have already run

@NuojCheng any thoughts here?

NuojCheng (commented):
Some additional train compile test for pipeline NNX migration:
- Train compile test 1: https://paste.googleplex.com/5960957017849856 
- Train compile test 2: https://paste.googleplex.com/5749974483730432 
- Train compile test 3: https://paste.googleplex.com/5201745681711104 
If the train compile tests above can pass without getting OOM + current tests in pipeline_parallelism_test.py can all pass, then I think it is good to go! Please ping me if the PR is ready for review.

NuojCheng:
There are also some linen usage in `pipeline_utils.py`, e.g. 
- https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
- https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307

I don't see them get updated in this PR but I think they probably should be updated?

Another thing is the usage of function in https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162. I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!

mesakhcienet:
> There are also some linen usage in `pipeline_utils.py`, e.g.
> 
> * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
> * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307
> 
> I don't see them get updated in this PR but I think they probably should be updated?
> 
> Another thing is the usage of function in
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162
> 
> . I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!

As far as I know, the current objective is to migrate the Linen pipeline to NNX while preserving the current Linen version. Please advise if any additional progress is required at this time. Thanks!

NuojCheng:
> > There are also some linen usage in `pipeline_utils.py`, e.g.
> > 
> > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
> > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307
> > 
> > I don't see them get updated in this PR but I think they probably should be updated?
> > Another thing is the usage of function in
> > https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162
> > 
> > . I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!
> 
> As far as I know, the current objective is to migrate the Linen pipeline to NNX while preserving the current Linen version. Please advise if any additional progress is required at this time. Thanks!

Shouldn't we have a nnx version of functions in pipeline_utils.py as well?

bvandermoon:
> > There are also some linen usage in `pipeline_utils.py`, e.g.
> > 
> > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
> > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307
> > 
> > I don't see them get updated in this PR but I think they probably should be updated?
> > Another thing is the usage of function in
> > https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162
> > 
> > . I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!
> 
> As far as I know, the current objective is to migrate the Linen pipeline to NNX while preserving the current Linen version. Please advise if any additional progress is required at this time. Thanks!

Are we able to bridge the NNX version back to Linen at a higher layer? If so, then I think we could get rid of the old Linen code that is no longer used

src/maxtext/layers/nnx_decoders.py
```diff
@@ -235,12 +238,90 @@ def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
...
+      params = jax.tree.map(lambda x: jnp.moveaxis(x, scan_axis, 0), params)
+
+    def layer_fn(carry, scanned_vars):
+      current_params, current_state = scanned_vars
+      layer = nnx.merge(graphdef, current_params, current_state)
+      layer_out = layer(carry, decoder_segment_ids, decoder_positions, deterministic, model_mode, **kwargs)
+      new_carry = layer_out[0] if isinstance(layer_out, tuple) else layer_out
+      return new_carry, nnx.state(layer)
+
+    final_carry, scanned_state = jax.lax.scan(layer_fn, inputs, (params, state))
```
cgarciae, line 362: you can alternatively use `nnx.scan` here which already does the `moveaxis` for you.

cgarciae: it also avoids the use of `split` and `merge`

mesakhcienet: thank you for your review, 
as far as I know, we should avoid the nnx.scan and nnx.remat as mentioned in the internal docs chat [here](https://docs.google.com/spreadsheets/d/1gtOGANxAJy5iNKYIscVFD5vzYY3m0zHesCOVfzaF4nk/edit?disco=AAABwsaOMIs)

mesakhcienet:
> > > There are also some linen usage in `pipeline_utils.py`, e.g.
> > > 
> > > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
> > > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307
> > > 
> > > I don't see them get updated in this PR but I think they probably should be updated?
> > > Another thing is the usage of function in
> > > https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162
> > > 
> > > . I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!
> > 
> > 
> > As far as I know, the current objective is to migrate the Linen pipeline to NNX while preserving the current Linen version. Please advise if any additional progress is required at this time. Thanks!
> 
> Are we able to bridge the NNX version back to Linen at a higher layer? If so, then I think we could get rid of the old Linen code that is no longer used

@bvandermoon 
Yes we are able to brigde the NNX version. Meanwhile, we are considering two solutions and would like your input on which direction to take:

Option 1: If we use `nnx_wrappers.ToLinen` to wrap NNX model back to Linen at a higher layer, we can completely remove the old Linen code, but we will need to conduct more tests.

Option 2: Delay the full migration until `nnx_decoders.py` is stable. In the meantime, we retain the current Linen pipeline and use the `enable_nnx` and `pure_nnx_decoder` flags to trigger the NNX version for testing. After the migration is stable, we can remove the flags and linen version of pipeline.

Please let me know which of these two solutions you prefer, thank you.


@NuojCheng 
Apologize i missed the first questions you asked.

                                                                                                                       
The NNX pipeline classes (NNXPipeline, NNXCircularPipeline) already handle these internally with JAX-native equivalents: 
                                                                                                                           
- `nn.remat`→ `jax.checkpoint` (pipeline.py L1551, L1870)                                                                   
- `nn.scan` → `jax.lax.scan + nnx.split/nnx.merge `(pipeline.py L1556, L1876)                                                
- `remove_logically_partition()`→ inline jax.tree.map unboxing (pipeline.py L1519-1522)                                   
                                                                                                                             
So no NNX versions of those functions are needed — the NNX path bypasses them entirely.                                  maybe you have some suggestions or any part that i am wrong? Please let me know. Thank you.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -235,12 +238,90 @@ def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
...
+
+    if scan_axis != 0:
+      scanned_params, scanned_other = scanned_state.split(nnx.Param, ...)
+      scanned_params = jax.tree.map(lambda x: jnp.moveaxis(x, 0, scan_axis), scanned_params)
+      scanned_state = nnx.State.merge(scanned_params, scanned_other)
+
+    self.scanned_layers = nnx.merge(graphdef, scanned_state)
```
cgarciae:
```suggestion
    nnx.update(self.scanned_layers, scanned_state)
```

bvandermoon:
> > > > There are also some linen usage in `pipeline_utils.py`, e.g.
> > > > 
> > > > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L330
> > > > * https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L307
> > > > 
> > > > I don't see them get updated in this PR but I think they probably should be updated?
> > > > Another thing is the usage of function in
> > > > https://github.com/AI-Hypercomputer/maxtext/blob/77f533451361c4bc1470deb9c7d7ff834a164f5f/src/maxtext/utils/pipeline_utils.py#L151-L162
> > > > 
> > > > . I suspect NNX migration can help us get rid of using this function since it is mostly dealing with linen wrapper troubles. Take a look on this if you can. Thank you for the hard work!
> > > 
> > > 
> > > As far as I know, the current objective is to migrate the Linen pipeline to NNX while preserving the current Linen version. Please advise if any additional progress is required at this time. Thanks!
> > 
> > 
> > Are we able to bridge the NNX version back to Linen at a higher layer? If so, then I think we could get rid of the old Linen code that is no longer used
> 
> @bvandermoon Yes we are able to brigde the NNX version. Meanwhile, we are considering two solutions and would like your input on which direction to take:
> 
> Option 1: If we use `nnx_wrappers.ToLinen` to wrap NNX model back to Linen at a higher layer, we can completely remove the old Linen code, but we will need to conduct more tests.
> 
> Option 2: Delay the full migration until `nnx_decoders.py` is stable. In the meantime, we retain the current Linen pipeline and use the `enable_nnx` and `pure_nnx_decoder` flags to trigger the NNX version for testing. After the migration is stable, we can remove the flags and linen version of pipeline.
> 
> Please let me know which of these two solutions you prefer, thank you.
> 
> @NuojCheng Apologize i missed the first questions you asked.
> 
> The NNX pipeline classes (NNXPipeline, NNXCircularPipeline) already handle these internally with JAX-native equivalents:
> 
> * `nn.remat`→ `jax.checkpoint` (pipeline.py L1551, L1870)
> * `nn.scan` → `jax.lax.scan + nnx.split/nnx.merge `(pipeline.py L1556, L1876)
> * `remove_logically_partition()`→ inline jax.tree.map unboxing (pipeline.py L1519-1522)
> 
> So no NNX versions of those functions are needed — the NNX path bypasses them entirely. maybe you have some suggestions or any part that i am wrong? Please let me know. Thank you.

Thank you @mesakhcienet. Let's go with option 1 please. That way we can continue running unit tests along the way, and we don't need to worry about the Linen/NNX versions diverging before the migration is fully done

src/maxtext/layers/pipeline.py
```diff
@@ -1689,45 +1712,73 @@ def fetch_active_stage_weights(self, bsw, loop_iteration, physical_partition_spe
...
+        pps_dict,
+    )
+
     _, repeat_ids = self.get_microbatch_and_repeat_ids(loop_iteration)
     stage0_repeat_id = jnp.maximum(loop_iteration, 0) // self.config.num_pipeline_microbatches
 
     if bsw_pps is not None:
 
-      @jax.shard_map(mesh=self.mesh, in_specs=((bsw_pps, bsw_pps), P("stage")), out_specs=bsw_pps, check_vma=True)
+      @jax.shard_map(mesh=self.mesh, in_specs=((bsw_pps, bsw_pps), P("stage")), out_specs=bsw_pps, check_vma=False)
```
NuojCheng: please set check_vma=True in pipeline code

mesakhcienet: done

src/maxtext/layers/nnx_decoders.py
mesakhcienet: done

mesakhcienet:
@bvandermoon 

Update on the pipeline migration approach:

After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:

The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).

**Reasoning:**
- The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
- Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
- The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
- `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
- `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.

**Plan to converge to Option 1:**
Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.

Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.

bvandermoon:
> @bvandermoon
> 
> Update on the pipeline migration approach:
> 
> After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:
> 
> The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).
> 
> **Reasoning:**
> 
> * The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
> * Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
> * The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
> * `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
> * `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.
> 
> **Plan to converge to Option 1:** Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.
> 
> Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.

Thanks @mesakhcienet. I feel we should push strongly to remove the Linen logic if possible. Even though the NNX and Linen versions are equivalent now, that could change if someone updates the Linen version and misses that the NNX one needs to be updated also. Is testing differences between Linen/NNX the main reason to fork these two? Maybe we could test before/after this commit if needed?

mesakhcienet:
> > @bvandermoon
> > Update on the pipeline migration approach:
> > After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:
> > The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).
> > **Reasoning:**
> > 
> > * The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
> > * Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
> > * The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
> > * `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
> > * `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.
> > 
> > **Plan to converge to Option 1:** Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.
> > Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.
> 
> Thanks @mesakhcienet. I feel we should push strongly to remove the Linen logic if possible. Even though the NNX and Linen versions are equivalent now, that could change if someone updates the Linen version and misses that the NNX one needs to be updated also. Is testing differences between Linen/NNX the main reason to fork these two? Maybe we could test before/after this commit if needed?

Thank you for your response. A few clarifications on the sequencing:

1. **The NNX path is opt-in, not the default.** This PR adds a new `use_nnx_pipeline` flag specifically for choosing between the Linen and NNX pipeline implementations, and it defaults to `false`. Combined with `enable_nnx` and `pure_nnx_decoder` (both also `false` by default), existing users stay on Linen and aren't affected by NNX changes. Linen remains the source of truth until we explicitly flip the defaults, so divergence is much less risky than if NNX were the live path.

2. **Removing Linen fully right now carries unknown risk.** I'm not yet confident what breaks downstream if the Linen pipeline classes are deleted outright — there may be call sites or configurations that depend on them in ways the current test coverage doesn't surface. Keeping both lets us de-risk the removal incrementally.

3. **NNX isn't unverified.** I've run it against the train-compile configs Nuoj asked about plus the end-to-end run in the PR description. It's just not under unit-test coverage yet.

4. **Existing unit tests are intentionally left on the Linen path** so the battle-tested version keeps rigorous coverage during the migration. Plan is to flip them to NNX once the decoder migration is done.

5. **Decoder NNX migration is still in flight** (open PRs adding missing models + their unit tests). Adding NNX pipeline unit tests now would likely need rework once that lands.

6. **This PR is already large** — I'd rather keep it focused on the migration and handle the test flip in a follow-up.

So my preference is to stick with the hybrid for now and converge to NNX-only once the decoder PRs land and we've validated the removal is safe. Open to revisiting if you'd rather sequence it differently. Thank you!

bvandermoon:
> > > @bvandermoon
> > > Update on the pipeline migration approach:
> > > After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:
> > > The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).
> > > **Reasoning:**
> > > 
> > > * The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
> > > * Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
> > > * The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
> > > * `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
> > > * `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.
> > > 
> > > **Plan to converge to Option 1:** Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.
> > > Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.
> > 
> > 
> > Thanks @mesakhcienet. I feel we should push strongly to remove the Linen logic if possible. Even though the NNX and Linen versions are equivalent now, that could change if someone updates the Linen version and misses that the NNX one needs to be updated also. Is testing differences between Linen/NNX the main reason to fork these two? Maybe we could test before/after this commit if needed?
> 
> Thank you for your response. A few clarifications on the sequencing:
> 
> 1. **The NNX path is opt-in, not the default.** This PR adds a new `use_nnx_pipeline` flag specifically for choosing between the Linen and NNX pipeline implementations, and it defaults to `false`. Combined with `enable_nnx` and `pure_nnx_decoder` (both also `false` by default), existing users stay on Linen and aren't affected by NNX changes. Linen remains the source of truth until we explicitly flip the defaults, so divergence is much less risky than if NNX were the live path.
> 2. **Removing Linen fully right now carries unknown risk.** I'm not yet confident what breaks downstream if the Linen pipeline classes are deleted outright — there may be call sites or configurations that depend on them in ways the current test coverage doesn't surface. Keeping both lets us de-risk the removal incrementally.
> 3. **NNX isn't unverified.** I've run it against the train-compile configs Nuoj asked about plus the end-to-end run in the PR description. It's just not under unit-test coverage yet.
> 4. **Existing unit tests are intentionally left on the Linen path** so the battle-tested version keeps rigorous coverage during the migration. Plan is to flip them to NNX once the decoder migration is done.
> 5. **Decoder NNX migration is still in flight** (open PRs adding missing models + their unit tests). Adding NNX pipeline unit tests now would likely need rework once that lands.
> 6. **This PR is already large** — I'd rather keep it focused on the migration and handle the test flip in a follow-up.
> 
> So my preference is to stick with the hybrid for now and converge to NNX-only once the decoder PRs land and we've validated the removal is safe. Open to revisiting if you'd rather sequence it differently. Thank you!

I can be open to this approach since some models are not integrated into the decoder layer. But can you please ensure there is a plan to avoid divergence between the NNX and Linen versions? I am concerned that if the Linen version changes, we could miss some functionality in the NNX version

I am also concerned that if we don't run the unit tests on the NNX version now, it will end up being more painful when we go to make the real cutover. Is there anything we can do to be more confident here?

mesakhcienet:
> > > > @bvandermoon
> > > > Update on the pipeline migration approach:
> > > > After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:
> > > > The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).
> > > > **Reasoning:**
> > > > 
> > > > * The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
> > > > * Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
> > > > * The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
> > > > * `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
> > > > * `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.
> > > > 
> > > > **Plan to converge to Option 1:** Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.
> > > > Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.
> > > 
> > > 
> > > Thanks @mesakhcienet. I feel we should push strongly to remove the Linen logic if possible. Even though the NNX and Linen versions are equivalent now, that could change if someone updates the Linen version and misses that the NNX one needs to be updated also. Is testing differences between Linen/NNX the main reason to fork these two? Maybe we could test before/after this commit if needed?
> > 
> > 
> > Thank you for your response. A few clarifications on the sequencing:
> > 
> > 1. **The NNX path is opt-in, not the default.** This PR adds a new `use_nnx_pipeline` flag specifically for choosing between the Linen and NNX pipeline implementations, and it defaults to `false`. Combined with `enable_nnx` and `pure_nnx_decoder` (both also `false` by default), existing users stay on Linen and aren't affected by NNX changes. Linen remains the source of truth until we explicitly flip the defaults, so divergence is much less risky than if NNX were the live path.
> > 2. **Removing Linen fully right now carries unknown risk.** I'm not yet confident what breaks downstream if the Linen pipeline classes are deleted outright — there may be call sites or configurations that depend on them in ways the current test coverage doesn't surface. Keeping both lets us de-risk the removal incrementally.
> > 3. **NNX isn't unverified.** I've run it against the train-compile configs Nuoj asked about plus the end-to-end run in the PR description. It's just not under unit-test coverage yet.
> > 4. **Existing unit tests are intentionally left on the Linen path** so the battle-tested version keeps rigorous coverage during the migration. Plan is to flip them to NNX once the decoder migration is done.
> > 5. **Decoder NNX migration is still in flight** (open PRs adding missing models + their unit tests). Adding NNX pipeline unit tests now would likely need rework once that lands.
> > 6. **This PR is already large** — I'd rather keep it focused on the migration and handle the test flip in a follow-up.
> > 
> > So my preference is to stick with the hybrid for now and converge to NNX-only once the decoder PRs land and we've validated the removal is safe. Open to revisiting if you'd rather sequence it differently. Thank you!
> 
> I can be open to this approach since some models are not integrated into the decoder layer. But can you please ensure there is a plan to avoid divergence between the NNX and Linen versions? I am concerned that if the Linen version changes, we could miss some functionality in the NNX version
> 
> I am also concerned that if we don't run the unit tests on the NNX version now, it will end up being more painful when we go to make the real cutover. Is there anything we can do to be more confident here?

@bvandermoon 

Thank you sir. Thinking it through, I'd actually like to go with your original suggestion — remove the Linen pipeline entirely in this PR rather than keep both. Maintaining the hybrid with proper safeguards (dual-path CI, equivalence tests, etc.) ends up being a similar amount of work to just doing the cutover with solid NNX tests, and the cutover leaves us in a cleaner place.

Proposed plan for this PR:

1. Remove the Linen pipeline classes as originally proposed.
2. Update the existing unit tests (mostly shoud be on `pipeline_parallelism_test.py`) to run against the NNX path so the pipeline stays under unit-test coverage through the migration.

Does this work for you? I'll need a bit of time to get the NNX tests passing before this is ready for merge. On our side, we'll also need #3114 to land into main branch first before we can merge these changes — happy to adjust the plan if you'd sequence the test work differently.

src/maxtext/configs/base.yml
```diff
@@ -1161,6 +1161,8 @@ subslice_shape: ""
 enable_nnx: False
 pure_nnx_decoder: False
 pure_nnx: False
+use_nnx_pipeline: False  # Set to False to use native Linen pipeline (with custom VJP)
```
NuojCheng: I wonder if CI tests can pass if `use_nnx_pipeline=true`?

mesakhcienet: I fixed the pipeline tests and just need to rebase the branch now. I'll ping you when it's ready. Thanks!

bvandermoon:
> > > > > @bvandermoon
> > > > > Update on the pipeline migration approach:
> > > > > After further investigation, we've adjusted from the original Option 1 (remove Linen entirely) to a hybrid approach. Here's why:
> > > > > The current branch keeps **both** Linen (`PipelineLinen`/`CircularPipelineLinen`) and NNX (`NNXPipeline`/`NNXCircularPipeline`) pipeline classes in `pipeline.py`, gated by a `use_nnx_pipeline` config flag (default: `False`).
> > > > > **Reasoning:**
> > > > > 
> > > > > * The NNX pipeline wrapped via `ToLinen` still needs validation against the full pipeline test suite (circular pipeline, FSDP+AG, multi-repeat, etc.) to confirm numerical equivalence with the original Linen version.
> > > > > * Keeping the Linen version as a reference allows us to A/B test and verify correctness before removing it.
> > > > > * The Linen classes are verified line-by-line identical to main branch -- no divergence risk.
> > > > > * `nnx_decoders.py` uses `create_nnx_pipeline()` directly for the pure NNX path.
> > > > > * `decoders.py` uses `create_pipeline()` which dispatches on the flag -- `False` gives native Linen, `True` gives NNX-wrapped-in-ToLinen.
> > > > > 
> > > > > **Plan to converge to Option 1:** Once the NNX pipeline passes all integration tests, we'll remove the Linen classes and flip the default, effectively completing Option 1.
> > > > > Let me know if you'd prefer we prioritize removing the Linen code sooner. Thank you.
> > > > 
> > > > 
> > > > Thanks @mesakhcienet. I feel we should push strongly to remove the Linen logic if possible. Even though the NNX and Linen versions are equivalent now, that could change if someone updates the Linen version and misses that the NNX one needs to be updated also. Is testing differences between Linen/NNX the main reason to fork these two? Maybe we could test before/after this commit if needed?
> > > 
> > > 
> > > Thank you for your response. A few clarifications on the sequencing:
> > > 
> > > 1. **The NNX path is opt-in, not the default.** This PR adds a new `use_nnx_pipeline` flag specifically for choosing between the Linen and NNX pipeline implementations, and it defaults to `false`. Combined with `enable_nnx` and `pure_nnx_decoder` (both also `false` by default), existing users stay on Linen and aren't affected by NNX changes. Linen remains the source of truth until we explicitly flip the defaults, so divergence is much less risky than if NNX were the live path.
> > > 2. **Removing Linen fully right now carries unknown risk.** I'm not yet confident what breaks downstream if the Linen pipeline classes are deleted outright — there may be call sites or configurations that depend on them in ways the current test coverage doesn't surface. Keeping both lets us de-risk the removal incrementally.
> > > 3. **NNX isn't unverified.** I've run it against the train-compile configs Nuoj asked about plus the end-to-end run in the PR description. It's just not under unit-test coverage yet.
> > > 4. **Existing unit tests are intentionally left on the Linen path** so the battle-tested version keeps rigorous coverage during the migration. Plan is to flip them to NNX once the decoder migration is done.
> > > 5. **Decoder NNX migration is still in flight** (open PRs adding missing models + their unit tests). Adding NNX pipeline unit tests now would likely need rework once that lands.
> > > 6. **This PR is already large** — I'd rather keep it focused on the migration and handle the test flip in a follow-up.
> > > 
> > > So my preference is to stick with the hybrid for now and converge to NNX-only once the decoder PRs land and we've validated the removal is safe. Open to revisiting if you'd rather sequence it differently. Thank you!
> > 
> > 
> > I can be open to this approach since some models are not integrated into the decoder layer. But can you please ensure there is a plan to avoid divergence between the NNX and Linen versions? I am concerned that if the Linen version changes, we could miss some functionality in the NNX version
> > I am also concerned that if we don't run the unit tests on the NNX version now, it will end up being more painful when we go to make the real cutover. Is there anything we can do to be more confident here?
> 
> @bvandermoon
> 
> Thank you sir. Thinking it through, I'd actually like to go with your original suggestion — remove the Linen pipeline entirely in this PR rather than keep both. Maintaining the hybrid with proper safeguards (dual-path CI, equivalence tests, etc.) ends up being a similar amount of work to just doing the cutover with solid NNX tests, and the cutover leaves us in a cleaner place.
> 
> Proposed plan for this PR:
> 
> 1. Remove the Linen pipeline classes as originally proposed.
> 2. Update the existing unit tests (mostly shoud be on `pipeline_parallelism_test.py`) to run against the NNX path so the pipeline stays under unit-test coverage through the migration.
> 
> Does this work for you? I'll need a bit of time to get the NNX tests passing before this is ready for merge. On our side, we'll also need #3114 to land into main branch first before we can merge these changes — happy to adjust the plan if you'd sequence the test work differently.

That sounds great. Thank you @mesakhcienet

src/maxtext/layers/decoders.py
```diff
@@ -305,10 +306,14 @@ def setup(self):
     self.decoder_layer = self.get_decoder_layers()
     self.norm_layer = self.get_norm_layer(num_features=self.config.emb_dim)
     if self.config.using_pipeline_parallelism:
-      pipeline_stage_module = self.get_pipeline_stage_module(self.decoder_layer)
       remat_policy = self.get_remat_policy()
+      nnx_blocks = self._get_nnx_decoder_block_classes()
+
+      def stage_factory(rngs):
```
NuojCheng: maybe give it a better name, and add some comments

src/maxtext/layers/decoders.py
```diff
@@ -498,6 +503,44 @@ def get_decoder_layers(self):
...
+
+    def get_scannable(normal_cls, scannable_cls):
+      return [scannable_cls] if cfg.scan_layers else [normal_cls]
+
+    def get_deepseek():
+      if cfg.use_batch_split_schedule:
+        return [deepseek_batchsplit.DeepSeekDenseLayer, deepseek_batchsplit.DeepSeekMoELayer]
+      return [deepseek.DeepSeekDenseLayer, deepseek.DeepSeekMoELayer]
+
+    layer_map = {
```
NuojCheng, line 518: why we do not directly use `get_decoder_layer()` but create a separate layer map?

src/maxtext/layers/decoders.py
```diff
@@ -526,6 +569,58 @@ def map_fn(path, value):
...
+      stage_module = SequentialBlockDecoderLayers(
+          decoder_layer=base_stage,
+          num_decoder_layers=cfg.num_layers_per_pipeline_stage,
+          config=cfg,
+          mesh=self.mesh,
+          quant=self.quant,
+          model_mode=self.model_mode,
+      )
+    return stage_module
+
```
NuojCheng, line 626: @gobbleturk could you review changes in decoder.py as well?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1142,152 +1298,229 @@ def __call__(
...
+              y, self.layers_outside_pipeline, _ = self._apply_layers_sequentially(
+                  self.layers_outside_pipeline, y, *layer_args, length=remaining, **layer_kwargs
+              )
+            elif (not cfg.scan_layers) and hasattr(self, "num_layers_outside_pipeline"):
+              for i in range(self.num_layers_outside_pipeline):
+                layer = getattr(self, f"layers_outside_pipeline_{i}")
+                out = layer(y, *layer_args, **layer_kwargs)
+                y = out[0] if isinstance(out, tuple) else out
+
+    else:
```
NuojCheng, line 1708: below is a huge if-else block with many branches. I wonder if there is a way to simplify the logic for better readabiltiy.

src/maxtext/layers/pipeline.py
```diff
@@ -812,33 +843,30 @@ def __call__(
...
-      segment_idx = None
 
     loop_state = self.init_states(inputs)
 
+    # MISS-1: Output shape assumes decoder-only models (micro_batch_size_to_train_on,
+    # max_target_length, emb_dim). After the use_nnx_pipeline flag was removed, ToLinen-NNX
+    # is the only path through create_pipeline — there is no native-Linen alternative that
+    # would compute the output shape generically. If pipeline parallelism is reused for
+    # encoder/seq2seq, this short-circuit must be updated to compute shape from the stage's
+    # abstract eval rather than hardcoding the decoder shape.
```
NuojCheng: Maybe you could remove context about removed flags and simplify the comment here

src/maxtext/layers/pipeline.py
```diff
@@ -852,157 +880,145 @@ def __call__(
...
-
 
-class CircularPipeline(PipelineBase):
-  """Implements an circular pipeline schedule with asynchronous weight prefetching.
 
-  Circular pipelining reduces the pipeline "bubble" by interleaving multiple pipeline
-  stages on the same physical devices. To hide the communication overhead of Fully
-  Sharded Data Parallelism (FSDP), this module utilizes a Buffer Sliding Window (BSW)
-  to prefetch and all-gather the weights for the *next* pipeline repeat while the
-  *current* repeat is executing.
```
NuojCheng, line 935: why previous comments are deleted?

src/maxtext/layers/pipeline.py
```diff
@@ -852,157 +880,145 @@ def __call__(
...
+      return None
 
-  def gather_microbatch_inputs_vmap(self, xs, ids, ids_dim):
-    """Slices out the specific sequence inputs (e.g., positions, segments) for the current microbatch."""
+    xs = jnp.asarray(xs)
+    ndim = xs.ndim
 
     def _gather_one(x, i):
-      idx = tuple(i if d == ids_dim else slice(None) for d in range(x.ndim))
+      # Captures: ndim (int), ids_dim (int), self.mesh (Python), self.config (Python) -- safe.
```
NuojCheng: what does this comment mean

src/maxtext/layers/pipeline.py
```diff
@@ -1012,158 +1028,37 @@ def _gather_one(x, i):
...
-    microbatch_0_idx = self.iterations_to_complete_first_microbatch() % self.microbatches_per_stage
-    output = jnp.roll(output, shift=-microbatch_0_idx, axis=1)
-    output = self._maybe_shard_with_logical(output, self.state_io_logical)
-    return output
+  def from_all_variables_to_repeat_weights(self, weights_state, loop_iteration):
+    """Slices out the specific repeat's weights from the full weights state.
 
-  def fetch_active_stage_weights(self, bsw, loop_iteration, physical_partition_spec=None, is_initializing=None):
-    """The module fetches the actively prefetched weights
-    from the Buffer Sliding Window to avoid mid-iteration FSDP all-gathers.
```
NuojCheng, line 1114: please do not remove previous comments, same for other functions

src/maxtext/layers/pipeline.py
```diff
@@ -1173,18 +1068,28 @@ def from_repeat_weights_to_bsw(
       # TODO (chengnuojin) set use_shardmap=true after JAX >= 10.0.0 and use all_gather(..., to='invarying')
       use_shardmap=False,  # using shardmap produces additional reduce-scatter in backward pass
   ):
-    """Executes the FSDP-like all-gathers to fully materialize a block of weights for the BSW."""
+    """Executes the FSDP-like all-gathers to fully materialize a block of weights for the BSW.
+
+    Captures via self: mesh (Python), config (Python) -- safe.
```
NuojCheng: remove?

src/maxtext/layers/pipeline.py
```diff
@@ -1173,18 +1068,28 @@ def from_repeat_weights_to_bsw(
...
+    else:
+      bsw_partition_specs = None
 
     def _from_repeat_weights_to_bsw_shardmap(
         repeat_weights,
         physical_partition_spec,
         axes_to_gather,
     ):
-      repeat_weights_pps = jax.tree.map(lambda p: P(*p[1:]), physical_partition_spec)
+      # Captures: self.mesh (Python), bsw_partition_specs (PartitionSpec pytree -- Python) -- safe.
```
NuojCheng: remove?

src/maxtext/layers/pipeline.py
```diff
@@ -1173,18 +1068,28 @@ def from_repeat_weights_to_bsw(
...
+      repeat_weights_partition_specs = jax.tree.map(
+          lambda p: P(*p[1:]) if isinstance(p, P) else p,
+          physical_partition_spec,
+          is_leaf=pipeline_utils.is_spec_leaf,
+      )
 
-      # Dynamically gather the index pytrees for all specified axes
```
NuojCheng, line 1187: why remove comments?

src/maxtext/layers/pipeline.py
```diff
@@ -1193,39 +1098,48 @@ def _from_repeat_weights_to_bsw_shardmap(
       axis_pytrees = list(axis_indices_dict.values())
 
       def should_skip_gather(axis_name, path_keys):
-        """Defines specific rule-based exceptions for gathering certain axes."""
+        # Captures: nothing (pure function on args).
         if axis_name == "expert" and "MoeBlock_0" in path_keys:
           return True
-        # Add more exclusion rules for other axes here if needed in the future
```
NuojCheng, line 1199: .

src/maxtext/layers/nnx_decoders.py
```diff
@@ -241,6 +242,85 @@ def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
...
+  def __init__(
+      self, layer_cls, num_layers: int, config: Config, mesh: Mesh, quant: Quant, model_mode: str, *, rngs: nnx.Rngs
+  ):
+    self.config = config
+
+    def create_layer_fn(rng):
+      return layer_cls(config=config, mesh=mesh, quant=quant, model_mode=model_mode, rngs=rng)
+
+    # Workaround for Deepseek MTP test failure.
+    # TODO: Handle this properly.
```
NuojCheng: please specify which error it is instead of using general except

src/maxtext/layers/decoders.py
mesakhcienet: Currently the `get_decoder_layer()` returns `ToLinen` decoder models, but in our pipeline we need `NNX` version of decoder models.

src/maxtext/layers/pipeline.py
```diff
@@ -39,26 +42,22 @@
...
 
 
-class PipelineBase(nn.Module):
-  """Base module that implements shared pipelining logic across stages."""
-
-  config: Config
-  layers: nn.Module
-  mesh: Mesh
-  remat_policy: Any = None
+class NNXPipelineBase(nnx.Module):
```
NuojCheng: If the goal is to deprecate linen version, we could just call it PipelineBase

src/maxtext/layers/pipeline.py
```diff
@@ -186,151 +184,27 @@ def get_microbatch_and_repeat_ids(self, loop_iteration):
     return microbatch_ids, repeat_ids
 
   def get_pipeline_remat_policy(self):
-    """Returns the pipeline remat policy for this pipeline."""
+    """Returns the pipeline remat policy for this pipeline.
+
+    Saves two named tensors during jax.checkpoint recomputation:
+      - "iteration_input": routed microbatch data entering the decoder
+      - "decoder_layer_input": input to the decoder layer itself
+    Everything else is recomputed during backward to save memory.
```
NuojCheng: This docstring is not accurate. The remat policy, even using pipeline parallelism, should be controlled by maxtext flags, e.g. `remat_policy=custom`, `context=device`...

src/maxtext/configs/base.yml
```diff
@@ -1178,8 +1178,8 @@ position_id_per_seconds: 25
 subslice_shape: ""
 
 # NNX
-enable_nnx: false
-pure_nnx_decoder: false
+enable_nnx: true
+pure_nnx_decoder: true
```
NuojCheng: does this change (using nnx decoder) have to be a part of this PR?

mesakhcienet: This change belongs on a different PR ,
it is just here to prove to reviewers that unit tests pass with the nnx decoder enabled.

FYI, currently we are on progress to push the `nnx_decoders.py` update on this [PR](https://github.com/AI-Hypercomputer/maxtext/pull/4132). We expect that PR to be merged first, since it contains similar changes to current branch's `nnx_decoders.py`

NuojCheng (approved):
LGTM

bvandermoon (approved):

PR #2906, opened 2025-12-31
Add Packing Support for Context Parallelism (Ring Attention)

kocchop:
# Description

Enables sequence packing for context parallelism with `ring` strategy using TransformerEngine's DotProductAttention. Includes comprehensive GPU tests for ring attention with packing for sm90+.

- Currently supports packing only for ring attention
- Replaced local sequence reordering with TE `reorder_causal_load_balancing` api
- Currently the load balancing strategy is automatically picked based on the packing config

Current support matrix for context parallelism on GPU:

CP Strategy | Packing | Load Balance Strategy | SWA
-- | -- | -- | --
`all_gather` | ❌ | `dual_chunk_swap` | ❌
`ring` | ✅ (with real data) | `dual_chunk_swap` (non-packed) `striped` (packed) | ✅ (with packing)

# Tests

Added a GPU integration test that works for sm90+.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -1038,6 +1038,7 @@ use_splash_scheduler: False # to use tokamax splash attention scheduler.
 ### Determine if we want to use load balance for context parallelism
 context_parallel_load_balance: True
 context_parallel_strategy: "all_gather" # "all_gather" or "ring"
+context_parallel_reorder_strategy: "auto" # "auto", "dual_chunk_swap", or "striped"
 
 ### Paged Attention ###
 # These settings take effect only when `attention=paged`.
--- a/src/maxtext/utils/max_utils.py
+++ b/src/maxtext/utils/max_utils.py
@@ -887,26 +887,85 @@ def reorder_sequence(tensor, cp_size: int, seq_dim: int = 1, to_contiguous: bool
   return reordered.reshape(ori_tensor_shape)
 
 
-@partial(jax.jit, static_argnums=1)
-def reorder_causal_load_balanced(batch, cp_size):
-  """Reorders the example batch sequences"""
-  return {
-      key: reorder_sequence(
-          value,  # Pass each key's value inside batch separately
-          cp_size=cp_size,
-      )
-      if key
-      in [
-          "inputs",
-          "targets",
-          "inputs_position",
-          "targets_position",
-          "inputs_segmentation",
-          "targets_segmentation",
-      ]
-      else value
-      for key, value in batch.items()
+@partial(jax.jit, static_argnums=(1, 2, 3))
+def reorder_causal_load_balanced(batch, cp_size, reorder_strategy, hardware="tpu"):
+  """Reorders the example batch sequences using a hardware-appropriate backend.
+
+  On GPU (hardware="gpu" or "gpu_multiprocess"), uses Transformer Engine's
+  reorder_causal_load_balancing which supports both DUAL_CHUNK_SWAP and STRIPED strategies.
+  On TPU/CPU, falls back to the pure-JAX reorder_sequence (DUAL_CHUNK_SWAP only).
+
+  Args:
+    batch: The batch to reorder.
+    cp_size: The size of the compute parallelism.
+    reorder_strategy: The ReorderStrategy enum value (DUAL_CHUNK_SWAP or STRIPED).
+    hardware: The hardware type string ("tpu", "gpu", "gpu_multiprocess", "cpu").
+
+  Returns:
+    The reordered batch.
+
+  Reorder Strategy:
+  - DUAL_CHUNK_SWAP: This strategy splits each query into two chunks and do the mirror swap between
+    GPUs. This is currently used for non-THD load balance. It requires the max_seqlens be the
+    multiple of 2 * cp_size.
+    Examples:
+    - Before reorder: GPU0: [0, 1, 2, 3]; GPU1: [4, 5, 6, 7]; GPU2: [8, 9, 10, 11]; GPU3: [12, 13, 14, 15];
+    - After reorder: GPU0: [0, 1, 14, 15]; GPU1: [4, 5, 10, 11]; GPU2: [8, 9, 6, 7]; GPU3: [12, 13, 2, 3]
+
+  - STRIPED: This strategy distributes the tokens in a striped (interleaved) manner across
+    the sequence. This is currently used for THD load balance.
+    Example: Consider 4 GPUs with seqlens=16.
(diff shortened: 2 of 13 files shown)


src/MaxText/layers/attention_op.py
```diff
@@ -1385,18 +1394,26 @@ def cudnn_flash_attention(
 
     # Handle packing configurations
     if self.config.packing and self.config.dataset_type != "synthetic":
+      if using_context_parallelism and not using_load_balanced_ring_cp:
+        raise AssertionError("Packing is only supported for load balanced ring attention with context parallelism.")
```
richjames0: Nit: AssertionError feels weird here to me. Maybe an argumenterror?

src/MaxText/maxtext_utils.py
```diff
@@ -107,19 +107,24 @@ def get_functional_eval_with_signature(eval_step, data_sharding, state_mesh_shar
...
+  reordered = max_utils.reorder_causal_load_balanced(batch, cp_size, reorder_strategy)
   for _, v in batch.items():
     if isinstance(v, jax.Array):
       reordered = sharding.maybe_shard_with_name(reordered, v.sharding, shard_mode)
       break
   return reordered
 
 
-def get_reorder_callable(cp_size, shard_mode):
+def get_reorder_callable(cp_size, shard_mode, reorder_strategy=0):  # 0=DualChunkSwap, 1=Striped
```
richjames0: As I read this late at night I imagine you're using an integer here so it's comprehensible by JAX but could this be made into an enum without breaking things (at worse using `.value`?

richjames0 (approved):
a couple of nits but lgtm

src/MaxText/maxtext_utils.py
```diff
@@ -107,19 +107,24 @@ def get_functional_eval_with_signature(eval_step, data_sharding, state_mesh_shar
   return functional_eval, in_shardings, out_shardings, static_argnums, donate_argnums
 
 
-def shard_reorder_causal_load_balanced(batch, cp_size, shard_mode):
+def shard_reorder_causal_load_balanced(batch, cp_size, shard_mode, reorder_strategy=0):
```
gobbleturk: Can we make reorder_strategy configurable via base.yml?

kocchop: done, ptal @gobbleturk

src/MaxText/maxtext_utils.py
kocchop: changed to enum @richjames0

src/MaxText/layers/attention_op.py
kocchop: converted to `ValueError`

src/maxtext/utils/max_utils.py
```diff
@@ -791,13 +791,50 @@ def reorder_sequence(tensor, cp_size: int, seq_dim: int = 1, to_contiguous: bool
   return reordered.reshape(ori_tensor_shape)
 
 
-@partial(jax.jit, static_argnums=1)
-def reorder_causal_load_balanced(batch, cp_size):
-  """Reorders the example batch sequences"""
+@partial(jax.jit, static_argnums=(1, 2))
+def reorder_causal_load_balanced(batch, cp_size, reorder_strategy):
+  """Reorders the example batch sequences
```
gobbleturk: Great comment explaining the two of them with examples, thank you!

src/MaxText/maxtext_utils.py
gobbleturk: Comment - Terminology/naming is hard, folks have been using the name "striped" to refer to DUAL_CHUNK_SWAP. I guess the string "striped" has to be passed to transformer engine for the other strategy (I would prefer the name "interleaved")...

I highly appreciate your comment with examples of the two strategies to clearly show what they mean in our codebase anyway!

src/maxtext/configs/base.yml
```diff
@@ -881,6 +881,7 @@ dq_reduction_steps: 0 #the number of reduction steps. For now, only 3 or all the
 ### Determine if we want to use load balance for context parallelism
 context_parallel_load_balance: True
 context_parallel_strategy: "all_gather" # "all_gather" or "ring"
+context_parallel_reorder_strategy: "auto" # "auto", "dual_chunk_swap", or "striped"
```
RissyRan, line 1041: Thanks for adding this strategy! Could you help add some explanation here for each from `reorder_causal_load_balanced`?

src/maxtext/utils/train_utils.py
```diff
@@ -186,26 +187,43 @@ def setup_train_loop(config, recorder, devices=None):
...
+            "Please use 'ring' strategy instead."
+        )
 
     # Apply reordering wrapper to data iterators if context parallelism is enabled
     with jax.set_mesh(mesh):
       if context_parallel_size > 1 and config.context_parallel_load_balance:
-        data_iterator = map(maxtext_utils.get_reorder_callable(context_parallel_size, config.shard_mode), data_iterator)
+
+        # Determine load balancing reorder strategy based on whether packing is enabled
+        if config.context_parallel_reorder_strategy == ReorderStrategy.AUTO:
```
RissyRan, line 251: Wondering if AUTO usually gives the best result? Or if there will be an compatible issue if user select ReorderStrategy.STRIPED, but without packing?

Trying to understand if we just provides 2 strategies are good enough.

src/maxtext/configs/base.yml
kocchop: Hi @RissyRan, I was ooo, could not follow up. The explanation for each of the reorder strategies are there with some example in `max_utils.py`. Are you asking to include it in the `base.yml`?

src/maxtext/utils/train_utils.py
kocchop: Hi @RissyRan, you can think of `AUTO` as a heuristics for users who may not want to deal with different reordering + cp strategy intricacies to begin with. We plan to add more support for reordering, packing with other CP strategies in future. That's why proposing this. If user has high level confidence and understanding, then they can just choose themselves, otherwise `AUTO` will try to pick based on the CP type, packing etc. configs. Let me know if it makes sense.

src/MaxText/maxtext_utils.py
kocchop: Hi @gobbleturk , so do you want to change the MaxText strategy name to `interleaved`?

src/maxtext/configs/base.yml
RissyRan: Yes, could you add some short explanation here. Users usually refer to base config instead of checking source codes. Some concise explanation will be great!

PR #2922, opened 2026-01-08
Add doc-only change detection to skip tests

hengtaoguo:
# Description
- Add `doc_only_check` job to detect when only `.md|ipynb` files are modified (Later, you can skip more file types such as `jpg|png|mp4` if needed):
  - If `skip_tests==true`: Skip the build process and all unit tests.
  - If `skip_tests==false`: Execute all remaining build process and unit tests like before.

- Add `all_tests_passed` to aggregate build and all unit tests status. Later in GitHub branch protection rule, we only check this test. Benefits of doing so:
  - Clear intent: Single "gate" that represents overall health
  - Future-proof: Add/remove/rename test jobs without updating branch protection
 

# Tests

Skip test (doc-only): https://screenshot.googleplex.com/ANutL8W8zpaRP5e
Non-skip test (success): https://screenshot.googleplex.com/BKxyPkxUgAYZvxR
Non-skip test (failure): https://screenshot.googleplex.com/4JU7AtHVWpJ7SMT
Non-skip test (cancel): https://screenshot.googleplex.com/8om8PyDT3vSxJd3

# Checklist

### DIFF
--- a/.github/workflows/build_and_test_maxtext.yml
+++ b/.github/workflows/build_and_test_maxtext.yml
@@ -37,7 +37,44 @@ concurrency:
 permissions:
   contents: read
 jobs:
+  doc_only_check:
+    name: Check for Documentation-Only Changes
+    runs-on: ubuntu-latest
+    outputs:
+      skip_tests: ${{ steps.check.outputs.skip_tests }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Check if only documentation changed
+        id: check
+        run: |
+          if [ "${{ github.event_name }}" != "pull_request" ]; then
+            echo "Not a pull request, running all tests"
+            echo "skip_tests=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          git fetch origin ${{ github.base_ref }}
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD)
+
+          echo "Changed files:"
+          echo "$CHANGED_FILES"
+
+          if [ -z "$CHANGED_FILES" ]; then
+            echo "No files changed"
+            echo "skip_tests=false" >> $GITHUB_OUTPUT
+          elif echo "$CHANGED_FILES" | grep -v -E '\.(md|ipynb)$' > /dev/null; then
+            echo "Non-documentation files changed, running tests"
+            echo "skip_tests=false" >> $GITHUB_OUTPUT
+          else
+            echo "Only documentation (.md|ipynb) files changed, skipping tests"
+            echo "skip_tests=true" >> $GITHUB_OUTPUT
+          fi
+
   build_and_upload_maxtext_package:
+    needs: doc_only_check
+    if: needs.doc_only_check.outputs.skip_tests != 'true'
     uses: ./.github/workflows/build_package.yml
     with:
       device_type: tpu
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test_maxtext.yml
```diff
@@ -177,6 +214,42 @@ jobs:
       container_resource_option: "--shm-size 2g --runtime=nvidia --gpus all --privileged"
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
 
+  all_tests_passed:
```
SurbhiJainUSC, line 217: Since you have already added `if: needs.doc_only_check.outputs.skip_tests != 'true'` to `build_and_upload_maxtext_package` job and all CPU/TPU/GPU tests job dependends on `build_and_upload_maxtext_package`, I think all the tests job will automatically skip. So, we might not need this block.

hengtaoguo: Currently, branch protection rule checks [all these tests](https://screenshot.googleplex.com/AcJAv8YpQdz7XTh) for every PR. Take a look at this counter-example [PR](https://screenshot.googleplex.com/C7sp28eG9kiETu7): 

- You can see that `MaxText Package Tests / build_and_upload_maxtext_package (pull_request) Skipped 17 minutes ago` is successfully skipped.
- However, `build_and_upload_maxtext_package / build_and_upload Expected — Waiting for status to be reported` is waiting forever, because branch protection rule is checking for its status. **It is blocking the submission indefinitely.**

**Why do two differently-named status checks appear for the same job?**

When workflows call reusable workflows using `uses:`, GitHub automatically generates status check names in the format `{workflow_name} / {caller_job_id}`, but branch protection was configured to expect the old format `{caller_job_id} / {called_job_id}`. This name mismatch means the required check never receives a status update, blocking the PR indefinitely.

So my solution is: Add an `all_tests_passed` summary job that:
1. Always runs and reports status (never shows "Expected")
2. Reports success when tests pass OR are properly skipped (doc-only changes)
3. Reports failure if any test fails
4. Is future-proof—adding/removing/renaming test jobs won't require updating branch protection

In this way, we will always have one gate check for the branch protection.

hengtaoguo: It works before because the workflow name contains "build_and_upload_maxtext_package", and GitHub's fuzzy matching considered it "close enough" to satisfy the protection rule. 

However, when we skip this test now, the actual workflow reports a "skipped" status instead of "success". GitHub branch protection rules only accept "success" as passing - "skipped" is treated as "not reported" and fails to satisfy the requirement. Since there's no successful status check to match against, the "Expected" check from the misconfigured branch protection rule will block the PR forever.

SurbhiJainUSC: Thank you Hengtao for the great explanation.

SurbhiJainUSC (approved):

.github/workflows/build_and_test_maxtext.yml
```diff
@@ -177,6 +214,38 @@ jobs:
...
+          echo "GPU tests: ${{ needs.maxtext_gpu_unit_tests.result }}"
+          echo "GPU integration: ${{ needs.maxtext_gpu_integration_tests.result }}"
+          
+          # Fail only if any job failed or was cancelled (skipped is OK)
+          if [ "${{ contains(needs.*.result, 'failure') }}" == "true" ] || [ "${{ contains(needs.*.result, 'cancelled') }}" == "true" ]; then
+            echo "❌ One or more jobs failed or were cancelled"
+            exit 1
+          fi
+          
+          echo "✅ All required tests passed successfully"
```
NuojCheng: do we want emoji?

NuojCheng (approved):

.github/workflows/build_and_test_maxtext.yml
hengtaoguo: Let me take that out for better robustness, thanks!

PR #2924, opened 2026-01-09
Add run_tests_coordinator for the test workflow

xibinliu:
# Description

Add run_tests_coordinator

This file is called by build_and_test_maxtext.yml to run tpu/cpu/gpu unit and integration tests. It simplifies the caller's input.

Moved post_training unit tests that require the post_training dependencies into tests/unit/post_train, so we can ignore it for pre-training tests.

# Checklist


.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,119 @@ jobs:
...
+        fail-fast: false
+        matrix:
+          image_name: [maxtext_gpu_jax_stable, maxtext_gpu_jax_nightly]
+    with:
+      device_type: ${{ matrix.cuda }}
+      device_name: a100-40gb-4
+      image_name_and_tag: ${{  matrix.image_name }}:${{ needs.setup.outputs.image_date }}
+      cloud_runner: linux-x86-a2-48-a100-4gpu
+      pytest_marker: 'not cpu_only and not tpu_only and integration_test'
+      pytest_addopts: '--ignore=tests/sft_hooks_test.py'
```
SurbhiJainUSC: No need to ignore these tests: https://github.com/AI-Hypercomputer/maxtext/blob/main/.github/workflows/build_and_test_maxtext.yml#L219. Same comment for GPU unit tests.

.github/workflows/build_and_test_maxtext.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       device_type: cpu
       device_name: X64
       cloud_runner: linux-x86-n2-16
-      image_type: ${{ matrix.image_type }}
+      image_name_and_tag: maxtext-unit-test-tpu:${{ matrix.image_type }}
```
SurbhiJainUSC: Can we call it base_image?

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,119 @@ jobs:
...
+      image_name_and_tag: ${{  matrix.image_name }}:${{ needs.setup.outputs.image_date }}
+      cloud_runner: linux-x86-ct6e-180-4tpu
+      pytest_marker: 'not cpu_only and not gpu_only and not integration_test'
+      xla_python_client_mem_fraction: 0.75
+      tf_force_gpu_allow_growth: false
+      container_resource_option: "--privileged"
+      is_scheduled_run: ${{ github.event_name == 'schedule' }}
+      maxtext_installed: true
+
+  post-training-images-tpu-integration-tests:
```
SurbhiJainUSC: Please run jupyter notebook tests as well for post-training image: https://github.com/AI-Hypercomputer/maxtext/blob/main/.github/workflows/build_and_test_maxtext.yml#L105

.github/workflows/build_and_test_maxtext.yml
xibinliu: Done

.github/workflows/UploadDockerImages.yml
xibinliu: Done!

.github/workflows/UploadDockerImages.yml
xibinliu: Done

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,149 @@ jobs:
...
+      base_image: ${{  matrix.image_name }}:${{ needs.setup.outputs.image_date }}
+      cloud_runner: linux-x86-a2-48-a100-4gpu
+      pytest_marker: 'not cpu_only and not tpu_only and integration_test'
+      xla_python_client_mem_fraction: 0.65
+      tf_force_gpu_allow_growth: true
+      container_resource_option: "--shm-size 2g --runtime=nvidia --gpus all --privileged"
+      is_scheduled_run: ${{ github.event_name == 'schedule' }}
+      maxtext_installed: true
+
+  pre-training-images_jupyter_notebooks:
```
SurbhiJainUSC: This is not required because this workflow only runs post-training notebooks.

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,149 @@ jobs:
...
+    with:
+      device_type: tpu
+      device_name: v6e-4
+      base_image: ${{  matrix.image_name }}:${{ needs.setup.outputs.image_date }}
+      cloud_runner: linux-x86-ct6e-180-4tpu
+      maxtext_installed: true
+    secrets:
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+  post-training-images_jupyter_notebooks:
```
SurbhiJainUSC: nit: `post-training-images-tpu-notebook-tests`

.github/workflows/UploadDockerImages.yml
xibinliu: Done!

.github/workflows/UploadDockerImages.yml
xibinliu: Done!

SurbhiJainUSC (approved):

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,133 @@ jobs:
...
+
+  pre-training-images-gpu-unit-tests:
+    needs: [setup, gpu-pre-training]
+    uses: ./.github/workflows/run_tests_against_package.yml
+    strategy:
+        fail-fast: false
+        matrix:
+          image_name: [maxtext_gpu_jax_stable, maxtext_gpu_jax_nightly]
+    with:
+      device_type: ${{ matrix.cuda }}
```
shralex: where is matrix.cuda defined ? same for integration tests

.github/workflows/run_tests_against_package.yml
```diff
@@ -89,12 +97,24 @@ jobs:
...
+            # Move to the directory where code is baked into the image. See the Dockerfile.
+            # This is necessary because GHA sets an empty workspace by default.
+            cd /deps
+            PYTHON_EXE="python3"
+          else
+            PYTHON_EXE=".venv/bin/python3"
+          fi
+
+          # Install pytest-cov to run test with codecov
+          $PYTHON_EXE -m pip install --quiet pytest-cov
```
shralex: since we're skipping pytest-cov upload why install this

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,133 @@ jobs:
       dockerfile: ${{ matrix.dockerfile }}
       maxtext_sha: ${{ needs.setup.outputs.maxtext_sha }}
       image_date: ${{ needs.setup.outputs.image_date }}
+
```
shralex, line 143: Should we add CPU tests as well ?

.github/workflows/UploadDockerImages.yml
xibinliu: This has been fixed in the refactored code

.github/workflows/run_tests_against_package.yml
xibinliu: Added logic to bypass the codecov related code for image test.

.github/workflows/UploadDockerImages.yml
xibinliu: Added CPU test to pre-training TPU images.

.github/workflows/run_tests_coordinator.yml
```diff
@@ -0,0 +1,94 @@
...
+            && '--shm-size 2g --runtime=nvidia --gpus all --privileged' 
+            || '--privileged' }}
+
+      # Metadata and Scaling
+      base_image: ${{ inputs.base_image }}
+      is_scheduled_run: ${{ inputs.is_scheduled_run }}
+      maxtext_installed: ${{ inputs.maxtext_installed }}
+      worker_group: ${{ inputs.worker_group }}
+      total_workers: ${{ inputs.total_workers }}
+      maxtext_sha: ${{ inputs.maxtext_sha }}
```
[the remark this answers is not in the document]
xibinliu: resolved

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,77 @@ jobs:
...
+    needs: [setup, tpu-pre-training]
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [maxtext_jax_stable, maxtext_jax_nightly]
+        worker: [1, 2]
+    uses: ./.github/workflows/run_tests_coordinator.yml
+    with:
+      flavor: cpu-unit
+      base_image: ${{  matrix.image_name }}:py312
```
shralex: should be matric.image ?

.github/workflows/UploadDockerImages.yml
```diff
@@ -15,7 +15,7 @@
 # This workflow builds and pushes MaxText images for both TPU and GPU devices.
 # It runs automatically daily at 12am UTC, on Pull Requests, or manually via Workflow Dispatch.
 
-name: Build Images
+name: Build and Test Images
```
shralex, line 18: Should we split this file or add an option to distinguish stable image building and testing from nightly building and testing ? Assuming that nightly might be frequently broken, my concern is that bundling the two would prevent us from publishing stable images to pypi

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,77 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [maxtext_jax_stable, maxtext_jax_nightly]
+        flavor: [tpu-unit, tpu-integration]
+    uses: ./.github/workflows/run_tests_coordinator.yml
+    with:
+      flavor: ${{ matrix.flavor }}
+      base_image: ${{ matrix.image }}:${{ needs.setup.outputs.image_date }}
+      is_scheduled_run: ${{ github.event_name == 'schedule' }}
```
shralex: this should probably just be true throughout this file. The idea for this flag is to run more tests when this isn't affecting a PR

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,77 @@ jobs:
...
+        worker: [1, 2]
+    uses: ./.github/workflows/run_tests_coordinator.yml
+    with:
+      flavor: cpu-unit
+      base_image: ${{  matrix.image_name }}:py312
+      worker_group: ${{ matrix.worker }}
+      total_workers: 2
+      is_scheduled_run: ${{ github.event_name == 'schedule' }}
+
+  post-training-images-tpu-notebook-tests:
```
shralex: could you move this up to be after post-training-tpu-tests, and remove "images" from the job name for consistency

.github/workflows/run_tests_coordinator.yml
```diff
@@ -0,0 +1,97 @@
...
+        type: string
+      base_image:
+        description: 'The docker image to run tests against'
+        required: true
+        type: string
+      is_scheduled_run:
+        required: false
+        type: boolean
+        default: false
+      worker_group:
```
shralex: why do we need to make worker_group and total_workers parameters ? do these change across invocations from uploadDockerImages and build_and_test_maxtext ?

.github/workflows/run_tests_coordinator.yml
```diff
@@ -0,0 +1,97 @@
...
+      device_type: >-
+        ${{ contains(inputs.flavor, 'tpu') && 'tpu' 
+            || (contains(inputs.flavor, 'gpu') && 'cuda12' || 'cpu') }}
+      
+      device_name: >-
+        ${{ contains(inputs.flavor, 'tpu') && 'v6e-4' 
+            || (contains(inputs.flavor, 'gpu') && 'a100-40gb-4' || 'X64') }}
+            
+      cloud_runner: >-
+        ${{ contains(inputs.flavor, 'tpu') && 'linux-x86-ct6e-180-4tpu'
```
shralex: Instead of using complex selection operators here consider creating a setup job that maps the flavor to specific outputs that would then be used as parameters. This centralizes the logic and provides implicit validation. Something like this:

  configure:
    outputs:
      device_type: ${{ steps.map.outputs.device_type }}
      device_name: ${{ steps.map.outputs.device_name }}
      # ... other outputs
    steps:
      - id: map
        run: |
          case "${{ inputs.flavor }}" in
            "tpu-unit")
              echo "device_type=tpu" >> $GITHUB_OUTPUT
              echo "device_name=v6e-4" >> $GITHUB_OUTPUT
              # also for pytest markers etc
              ;;
            "gpu-unit")
              echo "device_type=cuda12" >> $GITHUB_OUTPUT
              echo "device_name=a100-40gb-4" >> $GITHUB_OUTPUT
              ;;
            *)
              echo "::error::Unsupported flavor: ${{ inputs.flavor }}"
              exit 1
              ;;
          esac
 
 execute-test-package:
    needs: configure
    uses: ./.github/workflows/run_tests_against_package.yml
    with:
      device_type: ${{ needs.configure.outputs.device_type }}
      # ...

.github/workflows/UploadDockerImages.yml
xibinliu: Done.

.github/workflows/UploadDockerImages.yml
xibinliu: The trigger could be scheduled or manual for this workflow. But it probably okay to run full tests even with manual trigger. I now changed `is_scheduled_run` to `true` throughout the file.

.github/workflows/run_tests_coordinator.yml
xibinliu: Good point. I removed the "worker_group" and "total_workers" as parameters from the coordinator, and set them based on the other input (cpu or not).

.github/workflows/UploadDockerImages.yml
xibinliu: fixed.

.github/workflows/run_tests_coordinator.yml
xibinliu: The "configure" job has to be run a VM, isn't this heavier than just using selection operators?

.github/workflows/UploadDockerImages.yml
xibinliu: The build and tests jobs are independent for stable and nightly images, so they won't impact each other. If we want to keep them in one file depends on how the pypi publishing process use the output from this workflow. @SurbhiJainUSC could you comment here?

SurbhiJainUSC: This workflow uploads the docker image to GCR. This is independent of pypi release process. Also, we will be able to release stable docker image to GCR even if nightly docker image fails.

shralex: @SurbhiJainUSC how would we condition the release on just the stable tests passing here? Gemini says: "it is not possible to condition a downstream job on only a specific element of a matrix using the standard" and proposes to split the matrix into pre-training-tpu-stable-tests and pre-training-tpu-nightly-tests

.github/workflows/run_tests_coordinator.yml
shralex: I'm worried about the complexity / maintainability of these expressions. Given that this workflow is not blocking PRs, I would prefer to pay the 30s overhead of a setup job to simplify a bit. Alternatively, there might be an easier way, like this:

# Inside .github/workflows/run_tests_coordinator.yml
with:
  device_type: >-
    ${{ fromJSON('{
      "tpu-unit": "tpu",
      "tpu-integration": "tpu",
      "gpu-unit": "cuda12",
      "gpu-integration": "cuda12",
      "cpu-unit": "cpu"
    }')[inputs.flavor] }}
  
  pytest_marker: >-
    ${{ fromJSON('{
      "tpu-unit": "not cpu_only and not gpu_only and not integration_test",
      "tpu-integration": "not cpu_only and not gpu_only and integration_test",
      "gpu-unit": "not cpu_only and not tpu_only and not integration_test",
      "gpu-integration": "not cpu_only and not tpu_only and integration_test",
      "cpu-unit": "cpu_only"
    }')[inputs.flavor] }}

.github/workflows/UploadDockerImages.yml
SurbhiJainUSC: We have a different workflow for PyPI release: https://github.com/AI-Hypercomputer/maxtext/blob/main/.github/workflows/pypi_release.yml.

I think you are getting confused between PyPI release and docker images release. We have different workflows for both. `UploadDockerImages.yml` releases docker images to GCR.

.github/workflows/UploadDockerImages.yml
```diff
@@ -128,3 +128,75 @@ jobs:
       dockerfile: ${{ matrix.dockerfile }}
       maxtext_sha: ${{ needs.setup.outputs.maxtext_sha }}
       image_date: ${{ needs.setup.outputs.image_date }}
+
+  # TEST JOBS
```
SurbhiJainUSC, line 144: We should test the docker image before we add the required tags to the image. The flow should be:
```
1. Build the image.
2. Push the image with a dummy tag, but no latest tag.
3. Test the image.
4. Push the image & add the required tags.
```

OR
```
1. Build the image.
3. Test the image.
4. Push the image & add the required tags.
```

.github/workflows/UploadDockerImages.yml
xibinliu: changed the workflow to:

image build
upload image, but with a <date>-temp tag only
test
if tests pass, remove <date>-temp tag, add "latest", "date", and other hash tags

.github/workflows/UploadDockerImages.yml
xibinliu: changed the workflow to:

image build
upload image, but with a -temp tag only
test
if tests pass, remove -temp tag, add "latest", "date", and other hash tags

.github/workflows/run_tests_coordinator.yml
xibinliu: I used the new python "JSON Lookup Table" as you suggested.

.github/workflows/run_jupyter_notebooks.yml
```diff
@@ -25,15 +25,20 @@ on:
       device_name:
         required: true
         type: string
-      image_type:
+      base_image:
```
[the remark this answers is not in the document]
xibinliu: done

.github/workflows/run_tests_coordinator.yml
```diff
@@ -0,0 +1,92 @@
+# Copyright 2026 Google LLC
```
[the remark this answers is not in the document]
xibinliu: done

.github/workflows/run_tests_against_package.yml
```diff
@@ -59,16 +59,21 @@ on:
         type: number
         default: 1
       maxtext_sha:
-        required: true
+        required: false
```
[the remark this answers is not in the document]
xibinliu: Added description about the condition usage of the parameters.

.github/workflows/build_and_push_docker_image.yml
```diff
@@ -126,23 +122,19 @@ jobs:
...
           # Convert date to YYYYMMDD format
           clean_date=$(echo "${{ inputs.image_date }}" | sed 's/[-:]//g' | cut -c1-8)
 
           # Add MaxText tag
           maxtext_hash=$(git rev-parse --short HEAD)
-          gcloud container images add-tag "$SOURCE_IMAGE:latest" "$SOURCE_IMAGE:maxtext_${maxtext_hash}_${clean_date}" --quiet
+          gcloud container images add-tag "$SOURCE_IMAGE:$TEMP_TAG" "$SOURCE_IMAGE:maxtext_${maxtext_hash}_${clean_date}" --quiet
 
           # Add post-training dependencies tags
-          if [ "${{ inputs.is_post_training }}" == "true" ]; then
```
SurbhiJainUSC, line 141: Why did you remove `is_post_training` flag? This is to prevent adding post-training dependency labels to MaxText pre-training images.

xibinliu: For pre-training, there are no those dependencies directories, because they are not checked out. The `if [ -d "./$dir" ]; then` check will guard the tags not to be added.

.github/workflows/build_and_push_docker_image.yml
```diff
@@ -126,23 +124,19 @@ jobs:
         shell: bash
         run: |
           SOURCE_IMAGE="gcr.io/tpu-prod-env-multipod/${{ inputs.image_name }}"
-
-          # Add date tag
-          gcloud container images add-tag "$SOURCE_IMAGE:latest" "$SOURCE_IMAGE:${{ inputs.image_date }}" --quiet
+          TEMP_TAG="${{ inputs.image_date }}-temp"
```
SurbhiJainUSC: Instead of adding a `temp` tag to the image before testing, let's add `build-${{ github.run_id }}`. This will guarantee that even if we re-run the workflow on the same commit, the new image won't pull a cached version from a previous failed run.

Also, we should just add `maxtext` and post-training (if applicable) commit ids as the tag to the untested image. Date and `latest` tag should be added after testing is green.

xibinliu: > Instead of adding a temp tag to the image before testing, let's add build-${{ github.run_id }}

Done.

> we should just add maxtext and post-training (if applicable) commit ids as the tag to the untested image. Date and latest tag should be added after testing is green.

Yes, the implementation has this same logic.

.github/workflows/build_and_test_maxtext.yml
```diff
@@ -113,72 +113,47 @@ jobs:
...
+      base_image: maxtext-unit-test-tpu:py312
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
-      worker_group: ${{ matrix.worker_group }}
-      total_workers: 2
       maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
 
-  maxtext_tpu_unit_tests:
-    needs: build_and_upload_maxtext_package
+  gpu-tests:
+    needs: [doc_only_check, build_and_upload_maxtext_package]
```
shralex: since doc_only_check is a pre-requisite for build_and_upload_maxtext_package, there is no need to add doc_only_check here again. Similarly below.

.github/workflows/build_and_test_maxtext.yml
```diff
@@ -208,85 +181,38 @@ jobs:
...
-      xla_python_client_mem_fraction: 0.65
-      tf_force_gpu_allow_growth: true
-      container_resource_option: "--shm-size 2g --runtime=nvidia --gpus all --privileged"
-      is_scheduled_run: ${{ github.event_name == 'schedule' }}
-      maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
-
   all_tests_passed:
     name: All Required Tests Passed
-    needs: [doc_only_check, build_and_upload_maxtext_package, maxtext_cpu_unit_tests, maxtext_tpu_unit_tests, maxtext_tpu_integration_tests, maxtext_tpu_pathways_unit_tests, maxtext_tpu_pathways_integration_tests, maxtext_gpu_unit_tests, maxtext_gpu_integration_tests]
+    needs: [doc_only_check, build_and_upload_maxtext_package, tpu-tests, gpu-tests, cpu-tests, maxtext_tpu_pathways_unit_tests, maxtext_tpu_pathways_integration_tests]
```
shralex: I don't think you need to require docs or build_and_upload_maxtext_package here since other tests require these already

tests/unit/post_train/sft_hooks_test.py
```diff
@@ -31,6 +31,7 @@
 from maxtext.utils import maxtext_utils
 
 
+@pytest.mark.post_training
```
shralex: I think the marker will work even without its own folder. Not sure we need this folder for 1 file. Also, can we call the marker post_training_only ?

.github/workflows/build_and_test_maxtext.yml
```diff
@@ -113,72 +113,47 @@ jobs:
...
-      tf_force_gpu_allow_growth: false
-      container_resource_option: "--privileged"
+      flavor: ${{ matrix.flavor }}
+      base_image: maxtext-unit-test-cuda12:py312
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
       maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
 
-  maxtext_tpu_integration_tests:
-    needs: build_and_upload_maxtext_package
+  cpu-tests:
```
shralex: no pytest_marker ?

.github/workflows/build_and_test_maxtext.yml
shralex, line 160: this doesn't use the coordinator ? also below

tests/unit/post_train/sft_hooks_test.py
xibinliu: I'll remove the marker actually. The marker doesn't work because the pytest will load the the whole file and import the dependencies, and then read the marker. But the dependency for post training could be missing for pre-training images.

With a folder, we can skip the whole folder and the dependencies will not be imported/checked.

.github/workflows/build_and_test_maxtext.yml
xibinliu: Because eventually pathways tests and normal tpu tests uses different workflow files, if we combine them in the same coordinator, there will be "if" check for certain actions and we will see many "skipped" actions in the CI flow.  This is why we issue the pathways tests from the top.

Also pathways inputs is relatively simple, so we don't need another coordinator wrapper.

.github/workflows/build_and_test_maxtext.yml
xibinliu: Done

.github/workflows/build_and_test_maxtext.yml
xibinliu: Done

gobbleturk:
This PR has been reverted in https://github.com/AI-Hypercomputer/maxtext/pull/3134

PR #2930, opened 2026-01-12
Disable a flaky test

xuefgu:
# Description

Disable grain_data_processing_test.py::GrainArrayRecordAutoTuneTest::test_for_loop_repeatable due to flakiness. Fix is tracked in b/475255774.

# Tests

N/A

# Checklist

### DIFF
--- a/tests/grain_data_processing_test.py
+++ b/tests/grain_data_processing_test.py
@@ -226,6 +226,10 @@ class GrainArrayRecordAutoTuneTest(GrainArrayRecordProcessingTest):
   def test_batch_determinism(self):
     super().test_batch_determinism()
 
+  @pytest.mark.skip(reason="Flaky test - see b/475255774.")
+  def test_for_loop_repeatable(self):
+    super().test_for_loop_repeatable()
+
 
 class GrainArrayRecordBestFitPackingTest(GrainArrayRecordProcessingTest):
   """Test grain data processing with best_fit packing strategy."""


tests/grain_data_processing_test.py
```diff
@@ -226,6 +226,10 @@ def setUp(self):
   def test_batch_determinism(self):
     super().test_batch_determinism()
 
+  @pytest.mark.skip(reason="Flaky test - see b/475255774.")
```
gobbleturk, line 229: shouldn't we put the skip on the existing tests, e.g. here https://github.com/AI-Hypercomputer/maxtext/blob/0cd06b6d4f7a4a5de3323bd7c405fcd987386a84/tests/hf_data_processing_test.py#L91?

aireenmei: I think we only want to skip this test for class GrainArrayRecordAutoTuneTest, which is inherited from GrainArrayRecordProcessingTest

aireenmei: If we add the skip decorator, we don't need to use pass to set the test empty. We can write it similar to the way we skip test_batch_determinism test

tests/grain_data_processing_test.py
```diff
@@ -226,6 +226,10 @@ def setUp(self):
...
+  def test_for_loop_repeatable(self):
+    pass
```
aireenmei: Can you change pass to super().test_for_loop_repeatable? Similar to how we skip test_batch_determinism

aireenmei (changes requested):

tests/grain_data_processing_test.py
xuefgu: Done

tests/grain_data_processing_test.py
xuefgu: Done

tests/grain_data_processing_test.py
gobbleturk: I see this is some OOP inheritance stuff... I do prefer this solution  to using `pass`, thank you!

PR #2931, opened 2026-01-12
Fix dummy weight support in vLLM decode.

NicoGrande:
# Description

This PR fixes support for running vLLM decode with dummy weights. This was previously failing when `load_parameters_config=None` as Pydantic expects a string with "" being the default. 

Additionally this PR adds the logical axis rules context manager necessary to shard the model upon initialization. 

# Tests

Running vLLM decode with dummy weights.

# Checklist

### DIFF
--- a/src/MaxText/vllm_decode.py
+++ b/src/MaxText/vllm_decode.py
@@ -40,12 +40,14 @@ from typing import Any, Sequence
 
 from absl import app
 from absl import flags
+from flax.linen import partitioning as nn_partitioning
 import jax
 import transformers
 
 from MaxText import model_creation_utils
 from MaxText import pyconfig
 from MaxText.common_types import Config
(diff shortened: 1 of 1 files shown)


src/MaxText/vllm_decode.py
```diff
@@ -42,10 +42,12 @@
 from absl import flags
 import jax
 import transformers
+from flax.linen import partitioning as nn_partitioning
```
hengtaoguo: nit: We could sort these imports following PEP8 style guide.

richjames0 (approved):
lgtm

src/MaxText/vllm_decode.py
NicoGrande: Just changed the order, thank you for pointing out!

PR #2933, opened 2026-01-13
DeepSeek3.2: Onboard sparse attention

RissyRan:
# Description

Main author by @shuningjin 

## Background

DeepSeek V3.2 differs from DeepSeek V3 solely in the attention mechanism, aiming for efficiency in long-context scenario. While DeepSeek V3 uses Multi-head Latent Attention (MLA), DeepSeek V3.2 uses **DeepSeek Sparse Attention (DSA)**. DSA augments MLA with two components: 
- **Indexer**: parametric, qk product to get index score
- **Top-k token selection**: non-parametric, select top-k key/value for each query, introducing sparsity to qkv attention

## What this PR does

Fix: b/475925910

**1. Naive implementation of DeepSeek Sparse Attention (DSA)**

- Indexer: 
  - **qk product: currently implemented with dot product** to get index scores. To be optimized.
  - (minor) RoPE: indexer applies partial RoPE to q and k based on YaRN extension. It uses the same YaRN frequency as MLA, but with concatenated layout rather than interleaved layout.
  -   Based on index scores, get top-k indices and index mask

- Top-k selection for qkv attention:
  - This is currently implemented inside **dot product attention, by adding index mask** to regular attention mask. To be optimized.
- training only (no prefill / decode)
- See changes `attention_mla.py`, `attention_op.py`

**2. Onboard deepseek3.2-671b config**
- `deepseek3.2-671b.yml`
- deepseek v3.2 vs. v3: HF config [diff](https://diff.googleplex.com/#key=5cunmR62Fxpv): additional config for indexer
```
"index_head_dim": 128, "index_n_heads": 64, "index_topk": 2048,
```
- number of parameter: (1) Similar to v3, HF safetensor of v3.2 contains an extra layer for MTP which we omit. (2) Note that indexer contains extra parameter. (3) By [counting](https://github.com/AI-Hypercomputer/maxtext/blob/9f4ce1c84b94cead6a443d0b719994dc310f1948/src/MaxText/max_utils.py#L73), v3 has `671026419200` (671.03B) and v3.2 has`671877944064` (671.88B) parameters.

**3. unit test: ahead-of-time train compile for deepseek3.2-671b**

**4. unit test: compare output against torch code for Indexer and MLA**
- `deepseek32_vs_reference.py`
- The original torch reference can only run on GPU, due to quantization and fp8 kernel (act_quant, fp8_gemm, fp8_index). https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/model.py
- We adapted the reference to run on CPU 
  - changes: https://diff.googleplex.com/#key=UN66QnHPkvzg, ([original code after formatting](https://paste.googleplex.com/6700548998955008))
  - Remove quantization and use float32 for dtype and weight_dytpe
  - Replace fp8 kernel with naive dot product
  - (minor) Replace fast_hadamard_transform.hadamard_transform with F.linear

## Reference
- DeepSeek-V3.2 paper: https://arxiv.org/pdf/2512.02556
- HuggingFace: https://huggingface.co/deepseek-ai/DeepSeek-V3.2
- Github repo: https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/model.py

## Future work
- verify end-to-end training logits for deepseek3.2
- more efficient implementation of DSA

# Tests

Unit test against torch code (adapted from reference): indexer, MLA
```
python3 -m pytest -v --pyargs tests.unit.deepseek32_vs_reference_test -rP -s
```
Unit test for train compile 
```
python3 -m pytest -v --pyargs tests.unit.train_compile_test -rP -s -k "test_deepseek32"
```

# Checklist

### DIFF
--- a/src/MaxText/configs/models/deepseek3.2-671b.yml
+++ b/src/MaxText/configs/models/deepseek3.2-671b.yml
@@ -1,0 +1,59 @@
+# Copyright 2023–2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# model config for DeepSeek V3.2 - 671B
+# Identical to deepseek3-671b config, except adding indexer config.
+
+base_emb_dim: 7168
+base_num_query_heads: 128
+base_num_kv_heads: 128
+base_mlp_dim: 18432
+base_moe_mlp_dim: 2048
+base_num_decoder_layers: 61
+first_num_dense_layers: 3
+mlp_activations: ["silu","linear"]
+vocab_size: 129280
+enable_dropout: False
+logits_via_embedding: False
+normalization_layer_epsilon: 1.0e-6
+num_experts: 256
+num_experts_per_tok: 8
+shared_experts: 1
+routed_scaling_factor: 2.5
+routed_score_func: "sigmoid"
+routed_bias: True
+decoder_block: "deepseek"
+# MLA
+attention_type: "mla"
+q_lora_rank: 1536
+kv_lora_rank: 512
+qk_nope_head_dim: 128
+qk_rope_head_dim: 64
+v_head_dim: 128
+# RoPE
+mscale: 1.0
+rope_type: "yarn"
+rope_max_timescale: 10_000 # DeepSeek uses  "rope_theta": 10000
+max_position_embeddings: 163840
+original_max_position_embeddings: 4096
+rope_factor: 40
+beta_fast: 32
+rope_interleave: True
+rope_truncate: True
+rope_attention_scaling: False
+# Indexer for DeepSeek Sparse Attention
+use_sparse_indexer: True
+index_n_heads: 64
+index_head_dim: 128
+index_topk: 2048
--- a/src/MaxText/configs/types.py
+++ b/src/MaxText/configs/types.py
@@ -207,6 +207,7 @@ ModelName = Literal[
     "deepseek3-671b-2dfsdp",
     "deepseek3-test",
     "deepseek3-tiny",
+    "deepseek3.2-671b",
     "kimi-k2-1t",
     "gemma-7b",
     "gemma-2b",
@@ -500,6 +501,15 @@ class MlaAttention(BaseModel):
   qk_nope_head_dim: NonNegativeInt = Field(128, description="Dimension for non-RoPE part of QK heads in MLA.")
   qk_rope_head_dim: NonNegativeInt = Field(64, description="Dimension for RoPE part of QK heads in MLA.")
   v_head_dim: NonNegativeInt = Field(128, description="Dimension of V heads in MLA.")
+
+
+class AttentionIndexer(BaseModel):
+  """Configuration for DeepSeek Sparse Attention (DSA): DeepSeek3.2-style MLA with indexer."""
+
+  use_sparse_indexer: bool = Field(False, description="Whether to use sparse indexer for MLA.")
+  index_head_dim: NonNegativeInt = Field(128, description="Head dim for indexer query and key.")
+  index_n_heads: NonNegativeInt = Field(64, description="Number of query heads in indexer.")
+  index_topk: NonNegativeInt = Field(2048, description="Number of tokens selected by the query token in indexer.")
 
 
 class Llama4Attention(BaseModel):
@@ -1686,6 +1696,7 @@ class MaxTextConfig(
     Attention,
     MlaAttention,
     MoBa,
+    AttentionIndexer,
     Llama4Attention,
     SplashAttention,
     PagedAttention,
@@ -2120,6 +2131,11 @@ class MaxTextConfig(
         raise ValueError("`local_checkpoint_period` must be > 0 for emergency checkpointing.")
     if self.moba and self.attention not in ("dot_product"):
       raise ValueError("MoBA is only supported with dot_product attention.")
+    if self.use_sparse_indexer:
+      if self.q_lora_rank == 0:
+        raise NotImplementedError("Sparse indexer has not implemented for q_lora_rank = 0.")
+      if self.attention not in ("dot_product"):
+        raise ValueError("Sparse indexer is only supported dot_product attention")
     if self.attention_type == AttentionType.CHUNK.value and (
         not isinstance(self.chunk_attn_window_size, int) or self.chunk_attn_window_size <= 0
     ):
@@ -2259,9 +2275,7 @@ class MaxTextConfig(
           f"`python3 -m MaxText.muon_utils {self.model_name} True`"
       )
     if self.force_q_layout and not self.use_jax_splash:
-      raise ValueError(
-          "`force_q_layout` can only be true if `use_jax_splash` is also true."
-      )
+      raise ValueError("`force_q_layout` can only be true if `use_jax_splash` is also true.")
 
     # I. FINAL TYPE CONVERSIONS AND DERIVED LISTS
     # Create the ici_parallelism and dcn_parallelism lists for legacy compatibility.
--- a/src/MaxText/layers/attention_mla.py
+++ b/src/MaxText/layers/attention_mla.py
@@ -16,6 +16,8 @@
 
 import math
 from typing import Any, Optional, Tuple
+import copy
+
 import jax
 from jax.ad_checkpoint import checkpoint_name
 from jax.experimental import layout
@@ -57,6 +59,7 @@ from MaxText.common_types import (
     PREFILL_KV_BATCH,
     PREFILL_LENGTH,
     AttentionType,
+    DEFAULT_MASK_VALUE,
 )
 from MaxText.inference import kvcache
 from MaxText.inference import page_manager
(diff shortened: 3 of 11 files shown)


src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+    self.softmax_scale = self.head_dim**-0.5
+    self.rotary_embedding = rotary_embedding
+    assert config.index_topk < config.max_target_length
+
+    # Projection: Latent Query (qr) -> Indexer Heads
+    self.wq_b = nnx.Linear(
+        in_features=self.q_lora_rank,
+        out_features=self.n_heads * self.head_dim,
+        use_bias=False,
+        dtype=jnp.bfloat16,
```
RissyRan: let's use self.dtype, similar for others. May have exception for self.weights_proj

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+  """
+  DeepSeek V3.2 Sparse Attention Indexer.
+  Selects Top-K relevant tokens using a lightweight MQA mechanism.
+  """
+
+  def __init__(self, config: Any, rngs: nnx.Rngs, rotary_embedding):
+    self.n_heads = config.index_n_heads
+    self.head_dim = config.index_head_dim
+    self.index_topk = config.index_topk
+    self.dim = config.emb_dim
```
RissyRan: Let's keep alignment with MaxText term? I know reference called `dim`

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+        out_features=self.n_heads,
+        use_bias=False,
+        dtype=jnp.float32,
+        param_dtype=jnp.float32,
+        rngs=rngs,
+    )
+
+    # Internal Indexer Cache (distinct from main MLA KV Cache)
+    # Shape: [Batch, MaxLen, HeadDim]
+    self.k_cache = nnx.Variable(jnp.zeros((config.max_target_length, self.head_dim), dtype=jnp.bfloat16))
```
RissyRan: Why this is 2D, while note with [Batch, MaxLen, HeadDim]? We could remove this if not needed for training so far.

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+    k = self.k_norm(k)
+
+    # k = self._apply_partial_rope(k, freqs_cis)
+    # [CHANGE] Reshape K to have a "Heads" dimension for YarnRotaryEmbedding
+    # Input: [B, S, D] -> [B, S, 1, D]
+    k = k[:, :, None, :]
+    k = self._apply_partial_rope(k, positions)
+    k = k.squeeze(2)  # Back to [B, S, D]
+
+    # 3. Cache Update (Functional NNX update)
```
RissyRan: Shall we also apply `rotate_activation` from reference? It seems related to quantization strategy, and zero accuracy change without it.

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+    current_cache = self.k_cache.value
+    # Expand cache to batch if needed (simplified assumption)
+    if current_cache.ndim == 2:
+      current_cache = jnp.broadcast_to(current_cache, (bsz, current_cache.shape[0], current_cache.shape[1]))
+
+    updated_cache = jax.lax.dynamic_update_slice(current_cache, k.astype(current_cache.dtype), (0, start_pos, 0))
+    self.k_cache.value = updated_cache
+
+    # Active Keys: [B, TotalLen, D]
+    k_active = jax.lax.dynamic_slice(updated_cache, (0, 0, 0), (bsz, end_pos, self.head_dim))
```
RissyRan: Is this same with `start_pos:end_pos` for `end_pos`?

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +211,196 @@ def mla_as_linen(
...
+
+    # 6. Create Sparse Mask
+    bias_mask = jnp.full(index_score.shape, -1e10, dtype=x.dtype)
+
+    # Scatter 0.0 at topk indices
+    batch_idx = jnp.arange(bsz)[:, None, None]
+    seq_idx = jnp.arange(seqlen)[None, :, None]
+
+    # JAX scatter update
+    bias_mask = bias_mask.at[batch_idx, seq_idx, topk_indices].set(0.0)
```
RissyRan: I recall this `set` will be very inefficient. Will need to consider other operations later, leveraging matmul.

src/MaxText/layers/attention_mla.py
```diff
@@ -363,6 +555,15 @@ def __init__(
         rngs=rngs,
     )
 
+    # [CHANGE 1] Initialize Indexer
+    # We check a config flag to see if we are in Sparse/DeepSeek3.2 mode
+    self.use_sparse_indexer = getattr(config, "use_sparse_indexer", False)
```
RissyRan: Do you think it's better we pass this value with default as False?

src/MaxText/layers/attention_mla.py
```diff
@@ -363,6 +555,15 @@ def __init__(
...
+    if self.use_sparse_indexer:
+      indexer_rope = copy.copy(self.rotary_embedding)
```
RissyRan, line 609: Why do we need copy of rotary_embedding?

src/MaxText/layers/attention_mla.py
```diff
@@ -737,23 +950,83 @@ def __call__(
...
+    # jax_mask_bias = jnp.where(jax_mask, 0.0, -1e30).astype(jnp.bfloat16)
+
+    if self.use_sparse_indexer and low_rank_q is not None:
+      # Determine start_pos for cache updates (simplified logic)
+      start_pos = 0
+      if inputs_positions is not None:
+        start_pos = inputs_positions[0, 0]  # Assuming [B, L] or similar
+      # Run Indexer
+      # inputs_q is 'x', low_rank_q is 'qr'
+      sparse_bias, _, _ = self.indexer(
```
RissyRan: Let's rename it to index_mask instead of `bias`? bias is like adjustment values, and this is mask for sparsity.

src/MaxText/layers/attention_mla.py
```diff
@@ -737,23 +950,83 @@ def __call__(
...
+      if sparse_bias:
+        raise NotImplementedError
       unnormalized_out, _, exp_sum = self.ds_paged_attention_op(
           query, key, value, decoder_segment_ids, model_mode, previous_chunk, slot=slot, page_state=page_state
       )
       unnormalized_out = unnormalized_out[..., : self.v_head_dim]
       out = unnormalized_out / (exp_sum + 1e-9) if exp_sum is not None else unnormalized_out
     else:
-      out = self.attention_op(query, key, value, decoder_segment_ids, model_mode, cached_values)
+      # ds3.2, MHA mode for train / prefill, TODO: MQA model for decode (mathematically equivalent but speed faster)?
```
RissyRan: In MQA, the size of the Key-Value (KV) cache is reduced by a factor equal to the number of heads.

You may have this diff already, but just FYI. V3 MLA vs. V3.2: https://diff.googleplex.com/#key=3JSmf20vQG8U

src/MaxText/layers/attention_mla.py
shuningjin: Yes, I set it as `use_sparse_indexer: False` in base.yml, remove this line.

src/MaxText/layers/attention_mla.py
shuningjin: Sounds good. Renamed to index_mask.

src/MaxText/layers/attention_mla.py
shuningjin: I replaced these nnx.Linear with MaxText's DenseGeneral, and passed in config.dtype/weight/matmul_precision. self.weights_proj is kept as float32.

src/MaxText/layers/attention_mla.py
shuningjin: renamed to emb_dim

src/MaxText/layers/attention_mla.py
shuningjin: Both MLA and Indexer uses RoPE with YaRN extension, using the same YaRN frequency. They differs in the input layout: MLA yarn uses `interleave=true` and indexer uses `interleave=false`. Making a copy to keep the two process isolated. 

To clarify the use of YaRN, I have added comments in `attention_mla` and `YarnRotaryEmbedding`.

src/MaxText/layers/attention_mla.py
```diff
@@ -209,6 +213,193 @@ def mla_as_linen(
...
+    # Select TopK Indices (Values, Indices) after masking
+    _, topk_indices = jax.lax.top_k(index_score, k=self.index_topk)
+
+    # 6. Create Sparse Mask: 0 and large negatives
+    index_mask = jnp.full(index_score.shape, DEFAULT_MASK_VALUE, dtype=x.dtype)  # [B, S, T]
+    # Scatter 0.0 at topk indices
+    batch_idx = jnp.arange(bsz)[:, None, None]
+    seq_idx = jnp.arange(seqlen)[None, :, None]
+    # JAX scatter update
+    index_mask = index_mask.at[batch_idx, seq_idx, topk_indices].set(0.0)
```
RissyRan: Later - We could update it to jnp.where in this case. `set()` is a scatter operation, often less efficient.

```
# Assuming:
# topk_indices shape: [B, S, K]
# T is the target dimension size (index_score.shape[-1])

def get_mask_efficient(topk_indices, T, default_value):
    # 1. Create a range [0, 1, ..., T-1]
    # 2. Broadcast compare against [B, S, K] to get [B, S, K, T]
    # 3. Use .any() to see if a T-index is present in any of the K slots
    is_topk = (jnp.arange(T) == topk_indices[..., None]).any(axis=-2)
    
    # 4. Use where to select between 0.0 and the mask value
    return jnp.where(is_topk, 0.0, default_value)
    ```

Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: To generate mask from top-k indices, I test with jax.set (v1) and jax.where (v2) functions. On cpu, v1 is better. On tpu, v2 is better for both speed and memory.  Here is [result](https://b.corp.google.com/issues/475925910#comment2). So I will go with v2. Thanks for your great suggestion!

File: src/MaxText/layers/attention_mla.py (hunk 4 above)
Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: What `rotate_activation` does: hadamard transform
- `rotate_activation(q) @ rotate_activation(k).T` gives the same result as `q @ k.T`  (see below)
-  in their original implementation, they use `rotate_activation` before fp8-quantization for stability. 

We don't need it
- we are not using quantization for now, and `rotate_activation` does not affect result, will omit this from our implementation to avoid additional computation
- unit test can pass: maxtext w/o rotate_activation MATCH reference w/ rotate_activation

----
Equivalence:

Property of Hadamard matrix: For $n\times n$ Hadamard matrix $H$, $H H^\top = n I_n$. (By transposing both sides,  we also have $H^\top H = n I_n$)
- Let $\tilde{H} = H/ \sqrt{n}$. We have $\tilde{H} \tilde{H}^\top = \tilde{H}^\top \tilde{H} = I_n$. That is, the scaled Hadamard matrix $\tilde{H}$ is orthogonal.

Property of Hadamard Transform: mathematically, it is equivalent to multiplying input vector by a Hadamard matrix with some scale
- $(\tilde{H} k)^\top (\tilde{H} q) =k^\top \tilde{H}^\top \tilde{H} q = k^\top I_n q = k^\top q$. That is, Hadamard Transform does not change qk product.


Code example: product of scaled Hadamard matrix is identity
```
import torch, scipy
dim = 4
# Hadamard matrix, dim x dim
H = torch.tensor(scipy.linalg.hadamard(dim), dtype=torch.float32) 
print(H)
# Scaled Hadamard matrix, orthogonal matrix
scale = dim ** -0.5  # unitary scale
H_scaled = H * scale 
# product is identity
print(H_scaled @ H_scaled.T)
print(H_scaled.T @ H_scaled)
```
Code example: qk product with or without rotate_activation (hadamard transform) are the same
```
def rotate_activation(x: torch.Tensor) -> torch.Tensor:
  hidden_size = x.size(-1)
  return F.linear(x, torch.tensor(scipy.linalg.hadamard(hidden_size), dtype=x.dtype)) * hidden_size**-0.5

b, s, h, d = 1, 2, 1, 4
q = torch.randn(b, s, h, d)
k = torch.randn(b, s, d)
out1 = torch.einsum("bshd, btd -> bsth", q, k)
out2 = torch.einsum("bshd, btd -> bsth", rotate_activation(q), rotate_activation(k))
torch.testing.assert_close(out1, out2)
```

Reference:
- https://en.wikipedia.org/wiki/Hadamard_matrix
- https://en.wikipedia.org/wiki/Hadamard_transform
- https://github.com/Dao-AILab/fast-hadamard-transform: `hadamard_transform(x, scale=1.0)` Equivalent to `F.linear(x, torch.tensor(scipy.linalg.hadamard(dim))) * scale`.

File: src/MaxText/layers/attention_mla.py (hunk 6 above)
Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: changed from `set` to `jnp.where` as you suggested in the other comment. Can explore other operations in the future.

File: src/MaxText/layers/attention_mla.py (hunk 3 above)
Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: I will remove the k_cache as focusing on training for now.

File: src/MaxText/layers/attention_mla.py (hunk 5 above)
Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: I will remove the k_cache as focusing on training for now.

File: src/MaxText/layers/attention_mla.py (hunk 10 above)
Comment (shuningjin) [reply src/MaxText/layers/attention_mla.py]: Thank you for the diff, this is really helpful! 

I will leave inference-related code to future efforts.

File: src/MaxText/layers/attention_mla.py (hunk 4 above)
Comment (RissyRan) [reply src/MaxText/layers/attention_mla.py]: Sounds good. Let's focus on bf16 first, and adding this operation will potentially hit perf.

File: src/MaxText/configs/models/deepseek3.2-671b.yml [hunk 12]
```diff
@@ -0,0 +1,59 @@
+# Copyright 2023–2025 Google LLC
```
RissyRan: Nit with 2026

src/MaxText/configs/types.py
```diff
@@ -486,6 +487,15 @@ class MlaAttention(BaseModel):
   v_head_dim: NonNegativeInt = Field(128, description="Dimension of V heads in MLA.")
 
 
+class AttentionIndexer(BaseModel):
```
RissyRan, line 506: Could you add those 4 configs into MoE config readme doc? Could be a follow-up PR

src/MaxText/configs/types.py
```diff
@@ -486,6 +487,15 @@ class MlaAttention(BaseModel):
...
+  """Configuration for DeepSeek Sparse Attention (DSA): DeepSeek3.2-style MLA with indexer."""
+
+  use_sparse_indexer: bool = Field(False, description="If True, enables sparse indexer.")
+  index_head_dim: NonNegativeInt = Field(128, description="head dim for indexer")
```
RissyRan: Nit: Capitalize first letter in description to align with others. Similar comments for following.

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
 from MaxText.layers.quantizations import AqtQuantization as Quant
 
 
+class Indexer(nnx.Module):
+  """
+  Indexer for DeepSeek Sparse Attention (DSA).
+  Introduced by DeepSeek V3.2: https://arxiv.org/pdf/2512.02556.
```
RissyRan: Do you think we could also attach the reference implementation here along with paper?

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+    self.q_lora_rank = config.q_lora_rank
+    self.softmax_scale = self.head_dim**-0.5
+
+    # Projection: Latent Query -> Indexer Query
+    self.wq_b = DenseGeneral(
+        in_features_shape=self.q_lora_rank,
+        out_features_shape=(self.n_heads, self.head_dim),
+        axis=-1,
+        kernel_init=self.kernel_init,
+        # TODO(shuningjin): double check kernel axes
```
RissyRan: Do you have some concerns? I see it aligned with MLA: https://github.com/AI-Hypercomputer/maxtext/blob/d4a259dbbc3a7bf2194ea0a62bd210465e054be6/src/MaxText/layers/attention_mla.py#L425

We could start with this.

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+    )
+
+    # Projection: Input -> Importance Weights for Heads (with high precision)
+    self.weights_proj = DenseGeneral(
+        in_features_shape=self.emb_dim,
+        out_features_shape=self.n_heads,
+        axis=-1,
+        kernel_init=self.kernel_init,
+        kernel_axes=("embed", "q_heads"),
+        # Enforce FP32 as requested for stability in importance scoring
```
RissyRan: Shall we update the comment mentioning DS v3.2 using FP32? This indexer could be applied to other models in the future.

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+        dtype=jnp.float32,
+        weight_dtype=jnp.float32,
+        quant=None,  # Typically we don't quantize the importance selector head
```
RissyRan: Similar comment above.

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+    self.k_norm = nnx.LayerNorm(num_features=self.head_dim, use_bias=True, dtype=self.weight_dtype, rngs=rngs)
+
+  def apply_partial_rope(
+      self,
+      inputs: Array,
+      inputs_positions: Optional[Array | None] = None,
+  ):
+    """Applies partial RoPE to the indexer query or key
+
+    The Indexer's RoPE implementation differs from MLA's in two key aspects:
```
RissyRan, line 172: Thanks for the comment!

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+      d: Indexer Head Dimension (index_head_dim)
+    """
+    # NOTE: If sequence length <= topk, indexer always selects all tokens.
+    if self.config.max_target_length <= self.index_topk:
+      return None, None, None
+
+    bsz, seqlen = inputs_positions.shape  # s = t = seqlen
+
+    # Query Processing: Project from Latent low_rank_q
+    # TODO(shuningjin): why the shape is [b, t, h * d] rather than [b, t, h, d]?
```
RissyRan: Is this a comment for explanation? or related to implementation change?

src/MaxText/layers/attention_mla.py
```diff
@@ -737,23 +979,45 @@ def __call__(
...
     )
-
     query = checkpoint_name(query, "query_proj")
     key = checkpoint_name(key, "key_proj")
     value = checkpoint_name(value, "value_proj")
 
+    # Indexer Logic
+    index_mask = None
+    if self.use_sparse_indexer:
+      if self.q_lora_rank == 0:
```
RissyRan: Can we shift this logic earlier? I think validations are in type.py file. Similar comment, if you validate model_mode

src/MaxText/layers/attention_op.py
```diff
@@ -833,6 +836,9 @@ def apply_attention(
     length = query.shape[-3]
     target_hardware = self.mesh.devices[(0,) * self.mesh.devices.ndim].platform
 
+    if index_mask is not None and self.attention_kernel != "dot_product":
```
RissyRan: Similar comment with sparse attention enabled with other attention type.

tests/train_compile_test.py
```diff
@@ -769,3 +769,30 @@ def test_qwen3_next(self):
...
+            "compile_topology=v5p-256",
+            "use_iota_embed=true",
+            "compile_topology_num_slices=1",
+            "model_name=deepseek3.2-671b",
+            # megablox
+            "sparse_matmul=True",
+            "megablox=True",
+            "per_device_batch_size=1",
+            "max_target_length=1024",
+            "attention=dot_product",  # only support dot product now
```
RissyRan: You could write a TODO: update to flash attention when it's available.

tests/unit/deepseek32_vs_reference.py
```diff
@@ -0,0 +1,1092 @@
...
+  rope_truncate = True
+  rope_attention_scaling = False
+  # indexer
+  use_sparse_indexer = True
+  index_n_heads = 64
+  index_head_dim = 128  # > qk_rope_head_dim
+  index_topk = 4
+
+
+SEQ_LEN = 8
```
RissyRan: Could we also try SEQ_LEN < index_topk to ensure end-to-end working fine?

tests/unit/deepseek32_vs_reference.py
```diff
@@ -0,0 +1,1092 @@
...
+        inputs_q=to_jax(self.x),
+        inputs_kv=to_jax(self.x),
+        inputs_positions=decoder_positions,
+        decoder_segment_ids=decoder_segment_ids,
+        model_mode=MODEL_MODE_TRAIN,
+    )
+
+    print("torch out", pt_out)
+    print("jax out", jax_out)
+    # np.testing.assert_allclose(to_jax(pt_out / pt_out.sum()), jax_out / jax_out.sum(), rtol=1e-3, atol=1e-2)
```
RissyRan: It seems this is like a normalization? Suggest to remove it if not needed, or keep it in the code (not in comment)

RissyRan:
Also, don't forget to squash commits :)

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,224 @@
...
+        topk_indices: [b, t, k] int - The indices to keep.
+        s: int - The total size to select from.
+
+    Returns:
+        mask: [b, t, s] - `0.0` at topk_indices, `DEFAULT_MASK_VALUE` (large negative) elsewhere.
+    """
+    # 1. Create a range [0, 1, ..., s-1]
+    # 2. Broadcast compare against [b, t, k] to get [b, t, k, s]
+    # 3. Use .any() to see if a s-index is present in any of the k slots
+    is_topk = (jnp.arange(s) == topk_indices[..., None]).any(axis=-2)
```
gagika, line 204: isn't this really large Tensor `[b, t, k, s]` that can cause OOM?

I think you can use
`jnp.put_along_axis` (or `jax.lax.scatter`) to construct the mask directly without materializing the `[b, t, k, s]` tensor.

src/MaxText/configs/models/deepseek3.2-671b.yml
shuningjin: Updated to 2026 for newly added files.

src/MaxText/layers/attention_mla.py
shuningjin: Yes. I added: https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/model.py

tests/train_compile_test.py
shuningjin: Added TODO.

src/MaxText/layers/attention_mla.py
shuningjin: Sounds good.

src/MaxText/configs/types.py
shuningjin: Updated description with capitalization.

src/MaxText/configs/types.py
shuningjin: Will add in follow-up PR.

src/MaxText/layers/attention_mla.py
shuningjin: Thanks for the suggestion! It seems `jnp.where` is the best on tpu for both speed and memory, and `lax.scatter` is the best on cpu. A small benchmark: https://b.corp.google.com/issues/475925910#comment7.

gagika: thanks for benchmarking, seams like your solution works best for both memory & time considerations, seams xla doesn't materialize the large tensor.

parambole: Hey @shuningjin - Can we please elaborate on the benchmarking scale ? Also on an estimate on how this would look like for larger context lengths (e.g 128k ) ?

src/MaxText/layers/attention_mla.py
```diff
@@ -64,6 +67,229 @@
...
+    self.dtype = config.dtype
+    self.weight_dtype = config.weight_dtype
+
+    self.n_heads = config.index_n_heads
+    self.head_dim = config.index_head_dim
+    self.index_topk = config.index_topk
+    self.emb_dim = config.emb_dim
+    self.rope_head_dim = config.qk_rope_head_dim
+    self.q_lora_rank = config.q_lora_rank
+    self.softmax_scale = self.head_dim**-0.5
```
parambole, line 114: Can you please elaborate on why this scale is required ?

shuningjin: Thanks for noting this!
- The paper does not have this scale in [eq (1)](https://screenshot.googleplex.com/94zSE2EBzsDXyfa) for indexer.
- The reference implementation multiply the head_weights with `self.n_heads ** -0.5 * self.softmax_scale`
https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/87e509a2e5a100d221c97df52c6e8be7835f0057/inference/model.py#L478-L480
- Mathematically, this does affect top-k selection indices:`arg-topk(index_score) = arg-topk(index_score * some_scale)`. The scaling is most likely for numerical stability. 
- I keep the scaling to align with the reference, while adding a comment to clarify.

src/MaxText/layers/attention_op.py
shuningjin: moved the attention_type validation to type

src/MaxText/layers/attention_mla.py
shuningjin: moved the `q_lora_rank` validation to type

keeping `model_mode` validation here, as it is a run time attribute rather than config

src/MaxText/layers/attention_mla.py
shuningjin: updated comment

src/MaxText/layers/attention_mla.py
shuningjin: updated comment

tests/unit/deepseek32_vs_reference.py
shuningjin: I made MLA test parameterized, testing SEQ_LEN=2 < index_topk=4 (along with SEQ_LEN=8 > index_topk=4)

tests/unit/deepseek32_vs_reference.py
shuningjin: removed extra comment

src/MaxText/layers/attention_mla.py
shuningjin: removed extra comment

src/MaxText/layers/attention_mla.py
shuningjin: We can follow up on mask generation, as part of optimization @RissyRan

RissyRan: Yeah, let's put all optimization in later stage, and merge the functionality first.

shuningjin:
> LGTM! Please remove `tests/unit/yarn_vs_reference_test.py` if no changes there.

I am renaming `tests/unit/mla_vs_reference_test.py` to `tests/unit/yarn_vs_reference_test.py`, as the test is actually testing  yarn rope embedding rather than mla

src/MaxText/layers/attention_mla.py
shuningjin: For future reference, jax test with 4k len

```python
import scipy.linalg
import jax.numpy as jnp
import jax
import numpy as np

seed = 0
key = jax.random.key(seed)
# batch, seq len, num head, head dim
b, s, h, d = 2, 4096, 2, 2048
precision="float32"
dtype="float32"

def rotate_activation(q, d):
  hadamard_matrix = jnp.array(scipy.linalg.hadamard(d), dtype=q.dtype)
  return jnp.dot(q, hadamard_matrix, precision=precision) * (d ** -0.5)

q = jax.random.normal(key, shape=(b, s, h, d), dtype=dtype)
k = jax.random.normal(key, shape=(b, s, d), dtype=dtype)

out1 = jnp.einsum("bshd, btd -> bsth", q, k, precision=precision)
out2 = jnp.einsum("bshd, btd -> bsth", rotate_activation(q, d), rotate_activation(k, d), precision=precision)
np.testing.assert_allclose(out1, out2, atol=1e-4, rtol=1e-4)
```

PR #2953, opened 2026-01-15
Add option to skip first token during logits comparison

gagika:
# Description
Adds `--skip_first_token` flag to `tests/forward_pass_logit_checker.py`. This ignores the first token during logit comparison, preventing logit comparison failures due to high entropy at the first token which is sensitivity to numeric accuracy especially for MoE models with many experts.

# Tests
Verified locally on CPU. The script successfully ignored the initial token mismatch and passed the test criteria for the subsequent tokens using the new flag.

# Checklist

### DIFF
--- a/tests/utils/forward_pass_logit_checker.py
+++ b/tests/utils/forward_pass_logit_checker.py
@@ -302,12 +302,16 @@ def main(config, test_args):  # pylint: disable=W0621
             "Comparing up to the smaller vocab size."
         )
       min_vocab_size = min(full_train_logits.shape[-1], golden_logits.shape[-1])
+
+      start_index = 1 if test_args.skip_first_token else 0
       # shape [seq_len, vocab_size]
-      train_logits_slice = full_train_logits[0, :token_size, :min_vocab_size]
-      golden_logits_slice = golden_logits[:token_size, :min_vocab_size]
-      max_logging.log("\n[logits: token 2]")
-      max_logging.log(f"{golden_logits_slice[2]=}")
-      max_logging.log(f"{train_logits_slice[2]=}")
+      train_logits_slice = full_train_logits[0, start_index:token_size, :min_vocab_size]
+      golden_logits_slice = golden_logits[start_index:token_size, :min_vocab_size]
+
+      if train_logits_slice.shape[0] > 2:
+        max_logging.log(f"\n[logits: token {start_index + 2}]")
+        max_logging.log(f"{golden_logits_slice[2]=}")
+        max_logging.log(f"{train_logits_slice[2]=}")
 
       # Calculate absolute and relative differences for detailed reporting
       abs_diff = jnp.abs(train_logits_slice - golden_logits_slice)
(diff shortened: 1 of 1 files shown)


RissyRan:
Wondering if the 1st token doesn't match, will this affect following tokens?

gagika:
> Wondering if the 1st token doesn't match, will this affect following tokens?

at first token entropy is quite high (a lot of plausible options), so large (and MoE) models can be sensitive to numerical accuracy issues (e.g. routing to different expert).

2nd, 3rd, and upcoming tokens are all conditioned on the previous tokens (including first), and have smaller entropy, (less sensitive to numerical issues).

tests/forward_pass_logit_checker.py
```diff
@@ -297,12 +297,16 @@ def main(config, test_args):  # pylint: disable=W0621
...
-      golden_logits_slice = golden_logits[:token_size, :min_vocab_size]
+      train_logits_slice = full_train_logits[0, start_index:token_size, :min_vocab_size]
+      golden_logits_slice = golden_logits[start_index:token_size, :min_vocab_size]
+
       max_logging.log("\n[logits: token 2]")
-      max_logging.log(f"{golden_logits_slice[2]=}")
-      max_logging.log(f"{train_logits_slice[2]=}")
+      if train_logits_slice.shape[0] > 2:
+        max_logging.log(f"{golden_logits_slice[2]=}")
+        max_logging.log(f"{train_logits_slice[2]=}")
```
shuningjin, line 314: nit: could you move this log inside if and update the index
```
if train_logits_slice.shape[0] > 2:
  max_logging.log(f"\n[logits: token {start_index+2}]")
```
same for `max_logging.log("\n[probability: token 1]")` in the below

tests/utils/forward_pass_logit_checker.py
```diff
@@ -499,6 +509,13 @@ def main(config, test_args):  # pylint: disable=W0621
   parser.add_argument("--output_logits_path", type=str, required=False, default="")
   parser.add_argument("--gcs_output_logits_path", type=str, required=False, default="")
   parser.add_argument("--clip_logits_epsilon", type=float, required=False, default=None)
+  parser.add_argument(
+      "--skip_first_token",
```
shuningjin, line 518: "--skip_first_token" is only effective when we compare against pre-generated logit file, but not when run hf model on the fly (`run_hf_model=False`). 
- For clarity, might be good to raise not implemented error if `run_hf_model=false` and skip_first_token=true`
- same for "--clip_logits_epsilon"

shuningjin (approved):
Thanks for identifying the noise in the initial token prediction and adding the option to skip! LGTM.

RissyRan:
> > Wondering if the 1st token doesn't match, will this affect following tokens?
> 
> at first token entropy is quite high (a lot of plausible options), so large (and MoE) models can be sensitive to numerical accuracy issues (e.g. routing to different expert).
> 
> 2nd, 3rd, and upcoming tokens are all conditioned on the previous tokens (including first), and have smaller entropy, (less sensitive to numerical issues).

Thanks for the explanation! Potentially you have some data points for Kimi large model in logits verification. My mental model is: the first token is the most volatile due to high entropy. However, because each token depends on the previous one, that initial "tweak" usually causes the 2nd and 3rd tokens to diverge further from the original path, not settle back down. Right?

RissyRan:
> > > Wondering if the 1st token doesn't match, will this affect following tokens?
> > 
> > 
> > at first token entropy is quite high (a lot of plausible options), so large (and MoE) models can be sensitive to numerical accuracy issues (e.g. routing to different expert).
> > 2nd, 3rd, and upcoming tokens are all conditioned on the previous tokens (including first), and have smaller entropy, (less sensitive to numerical issues).
> 
> Thanks for the explanation! Potentially you have some data points for Kimi large model in logits verification. My mental model is: the first token is the most volatile due to high entropy. However, because each token depends on the previous one, that initial "tweak" usually causes the 2nd and 3rd tokens to diverge further from the original path, not settle back down. Right?

Discussed offline with @gagika! It makes more sense now. This could be an issue for large MoE models (like Kimi-K2), potentially related with numerics, even the tokens IDs and weights are fixed.

tests/utils/forward_pass_logit_checker.py
gagika: `--skip_first_token` should work for both paths pre-generated and hf model.

I am raising error for --clip_logits_epsilon on hf run.

tests/forward_pass_logit_checker.py
gagika: done, good catch.

PR #2968, opened 2026-01-20
Implement QK-Clip (Muon-Clip) for MLA attention

gagika:
# Description

Implements QK-Clip, a training stabilization technique for MLA attention models, as described in the Kimi K2 Technical Report.

**Changes:**
* **Core Logic:** Added `src/MaxText/utils/qk_clip_utils.py` containing `apply_qk_clip` and `calculate_max_logit_metric`.
* **Layers:** Updated `AttentionOp` to `sow` max logits statistics and `AttentionMLA` to enable this when configured.
* **Training:** Integrated the clipping step and `max_logits` metric reporting into `src/MaxText/train.py`.
* **Tests:** Added `tests/qk_clip_test.py`.

**Context:**
QK-Clip mitigates training instability by preventing attention logits from growing excessively. This implementation:
1.  Calculates global max logit ($S_{max}$) using GSPMD-compatible `jnp.max`.
2.  Computes per-head scaling factor $\gamma = \min(1, \tau / S_{max})$.
3.  Scales $W_q$ and $W_k$ while explicitly leaving shared rotary keys ($k^R$) and values ($W_v$) untouched.
4.  Leverages Flax `sow` to pass statistics from layers to the training loop efficiently.

# Tests

* **Unit Tests:** Ran `python3 tests/qk_clip_test.py`. Verified:
    * Correct scaling of $W_q$ and $W_k$.
    * Heads below threshold are not clipped.
    * Shared keys and values remain untouched.
    * Global `max_logits` metric calculation.
    * Error handling for non-MLA attention types.
* **Integration:** Verified `train_step` executes without shape mismatches or runtime errors.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -355,6 +355,10 @@ qk_nope_head_dim: 128
 qk_rope_head_dim: 64
 v_head_dim: 128
 
+# QK-Clip (Muon Clip) Configuration
+use_qk_clip: False  # Enable QK-Clip (supported in MLA with DotProduct or Tokamax Splash)
+qk_clip_threshold: 100.0  # Threshold for clipping (tau in the paper)
+
 # Combine matmuls for QKV and MLP
 fused_qkv: False
 fused_mlp: False
--- a/src/maxtext/layers/attention_mla.py
+++ b/src/maxtext/layers/attention_mla.py
@@ -1030,6 +1030,9 @@ class MLA(Attention):
           attention_mask=attention_mask,
       )
 
+    # Check if we need QK Clip stats
+    use_qk_clip = self.model_mode == MODEL_MODE_TRAIN and self.config.use_qk_clip
+
     if self.config.attention == "paged" and model_mode != MODEL_MODE_TRAIN:
       unnormalized_out, _, exp_sum = self.ds_paged_attention_op(
           query, key, value, decoder_segment_ids, model_mode, previous_chunk, slot=slot, page_state=page_state
@@ -1037,8 +1040,16 @@ class MLA(Attention):
       unnormalized_out = unnormalized_out[..., : self.v_head_dim]
       out = unnormalized_out / (exp_sum + 1e-9) if exp_sum is not None else unnormalized_out
     else:
-      # Pass the index_mask to the Attention Op
-      out = self.attention_op(query, key, value, decoder_segment_ids, model_mode, cached_values, index_mask=index_mask)
+      out = self.attention_op(
+          query,
+          key,
+          value,
+          decoder_segment_ids,
+          model_mode,
+          cached_values,
+          index_mask=index_mask,
+          record_max_logits=use_qk_clip,
+      )
 
     out = jax.ad_checkpoint.checkpoint_name(out, "attention_out")
     if model_mode == MODEL_MODE_TRAIN and self.config.expert_shard_attention_option == EP_AS_CONTEXT:
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -146,11 +146,15 @@ def apply_mask_to_logits(logits: Array, mask: Array):
   )
 
 
-def validate_flash_attention_with_sinks_on_gpu(sinks: Array | None) -> None:
-  """Helper function to check for sinks with flash attention on GPU."""
+def validate_gpu_flash_attention(sinks: Array | None, record_max_logits: bool) -> None:
+  """Helper function to check for unsupported features with flash attention on GPU."""
   if sinks is not None:
     raise ValueError(
         "The flash attention with sinks is not supported on GPU yet."
+    )
+  if record_max_logits:
+    raise NotImplementedError(
+        "record_max_logits (QK-Clip) is not supported for GPU flash attention kernels yet."
     )
 
 
@@ -940,6 +944,7 @@ class AttentionOp(nnx.Module):
       bidirectional_mask: Any = None,
       sinks: Array | None = None,
       index_mask: Array | None = None,
+      record_max_logits: bool = False,
       *,
       qk_product_einsum: Callable[..., Array],
       wv_product_einsum: Callable[..., Array],
@@ -959,7 +964,11 @@ class AttentionOp(nnx.Module):
         impl = self.gpu_ragged_attention
       else:
         raise NotImplementedError(target_hardware)
-      return impl(query, key, value, lengths, self.ragged_block_size)
+
+      local_out, local_max, local_sum = impl(query, key, value, lengths, self.ragged_block_size)
+      if record_max_logits:
+        self.sow("intermediates", "max_logits", local_max)
+      return local_out, local_max, local_sum
 
     # 'vllm_rpa' uses the same dot-attention wrapper but routes to the vLLM
     # ragged paged attention kernel in `Attention.__call__`.
@@ -983,6 +992,7 @@ class AttentionOp(nnx.Module):
           bidirectional_mask=bidirectional_mask,
           sinks=sinks,
           index_mask=index_mask,
+          record_max_logits=record_max_logits,
           qk_product_einsum=qk_product_einsum,
           wv_product_einsum=wv_product_einsum,
       )
@@ -996,21 +1006,21 @@ class AttentionOp(nnx.Module):
         if model_mode == MODEL_MODE_AUTOREGRESSIVE:
           raise ValueError("""Decode not supported with flash attention.
                               Use `dot_product` instead.""")
-        return (
-            self.tpu_flash_attention(
-                query,
-                key,
-                value,
-                decoder_segment_ids,
-                self.attn_logits_soft_cap,
-                sinks,
-                index_mask,
-            ),
-            None,
-            None,
+
+        out, max_logits = self.tpu_flash_attention(
+            query,
+            key,
+            value,
+            decoder_segment_ids,
+            self.attn_logits_soft_cap,
+            sinks,
+            record_max_logits=record_max_logits,
         )
+        if max_logits is not None:
+          self.sow("intermediates", "max_logits", max_logits)
+        return out, None, None
+
       else:
-        validate_flash_attention_with_sinks_on_gpu(sinks)
         if model_mode == MODEL_MODE_AUTOREGRESSIVE:
           # fallback to dot_product as pallas gpu flash attention doesn't support decode stage
           return self.apply_attention_dot(
@@ -1020,10 +1030,12 @@ class AttentionOp(nnx.Module):
               decoder_segment_ids,
               model_mode,
               bidirectional_mask=bidirectional_mask,
+              record_max_logits=record_max_logits,
               qk_product_einsum=qk_product_einsum,
               wv_product_einsum=wv_product_einsum,
           )
         else:
+          validate_gpu_flash_attention(sinks, record_max_logits)
           head_axis = -2
           num_query_heads = query.shape[head_axis]
           num_kv_heads = key.shape[head_axis]
@@ -1043,12 +1055,13 @@ class AttentionOp(nnx.Module):
             value = jnp.repeat(
                 value, q_heads_per_kv_head, axis=head_axis
             )  # value shape [batch_size, kv_seq_len, num_kv_heads, head_dim]
+
           out = gpu_pallas_attention.mha(
               query, key, value, decoder_segment_ids, sm_scale=1.0, causal=True
           )
           return out, None, None
     elif self.attention_kernel == "cudnn_flash_te":
-      validate_flash_attention_with_sinks_on_gpu(sinks)
+      validate_gpu_flash_attention(sinks, record_max_logits)
       if isinstance(key, KVTensor):
         key = key.dequant()
       if isinstance(value, KVTensor):
@@ -1064,7 +1077,7 @@ class AttentionOp(nnx.Module):
           None,
       )
     elif self.attention_kernel == "cudnn_flash_jax":
-      validate_flash_attention_with_sinks_on_gpu(sinks)
+      validate_gpu_flash_attention(sinks, record_max_logits)
       if isinstance(key, KVTensor):
         key = key.dequant()
       if isinstance(value, KVTensor):
@@ -1190,7 +1203,8 @@ class AttentionOp(nnx.Module):
       attn_logits_soft_cap: float | None = None,
       sinks: Array | None = None,
       index_mask: Array | None = None,
-  ) -> Array:
+      record_max_logits: bool = False,
+  ) -> tuple[Array, Array]:
     """TPU Flash Attention."""
 
     cp_size = self.config.context_parallel_size
@@ -1454,6 +1468,21 @@ class AttentionOp(nnx.Module):
     # 'segment_axis_names_q' maps to ['activation_q_length', ['context']] meaning that q is sharded over the context axis
     #  'segment_axis_names_kv' maps to ['activation_kv_length', []] meaning that K and V are not sharded
     # splash_kernel is sharded over (HEAD, LENGTH)
+
+    if record_max_logits:
+      # max_logits will share similar sharding as query but last dim is unrelated to model
+      # Using None for the last dimension of max_logits sharding
+      if isinstance(axis_names_q, jax.sharding.PartitionSpec):
+        # max_logits is Rank 3 (batch, heads, seq), so drop the last dimension (d_kv)
+        max_logits_spec = jax.sharding.PartitionSpec(*axis_names_q[:-1])
+      else:
+        # Fallback if axis_names_q is not a PartitionSpec (unlikely in this context)
+        max_logits_spec = axis_names_q[:-1]
+      out_specs = (axis_names_q, max_logits_spec)
+    else:
+      # Safely pad the specs with None to match the (attention_output, None) return type
+      out_specs = (axis_names_q, None)
+
     @functools.partial(
         jax.shard_map,
         mesh=self.mesh,
@@ -1470,7 +1499,7 @@ class AttentionOp(nnx.Module):
             sink_axis_names,  # sharding align with query heads
             index_mask_axis_names,
         ),
-        out_specs=axis_names_q,
+        out_specs=out_specs,
         check_vma=False,
     )
     def wrap_flash_attention(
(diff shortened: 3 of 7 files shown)


RissyRan (commented):
@gemini-cli /review

src/maxtext/configs/base.yml
```diff
@@ -342,6 +342,10 @@ qk_nope_head_dim: 128
 qk_rope_head_dim: 64
 v_head_dim: 128
 
+# QK-Clip (Muon Clip) Configuration
+use_qk_clip: False  # Enable QK-Clip (supported in MLA with DotProduct or Tokamax Splash)
```
RissyRan, line 359: Shall we add assertion in types.py? Something like

```
if use_qk_clip:
    assert attention_type == "mla", (
        f"QK-Clip is only supported when attention='mla', "
        f"but found attention='{attention_type}'."
    )
  ```

File: src/maxtext/layers/attention_mla.py [hunk 2]
```diff
@@ -1028,15 +1028,26 @@ def __call__(
       if index_mask is not None:
         index_mask = index_mask[:, None, None, :, :]  # [b, 1, 1, q_len, kv_len]
 
+    # Check if we need QK Clip stats
```
RissyRan, line 1033: In types.py file, potentially we could check the mode? i.e. use_qk_clip only supported in training mode, so that we could remove this check in attention_mla.py.

src/maxtext/layers/attention_op.py
```diff
@@ -826,6 +826,7 @@ def apply_attention(
       bidirectional_mask: Any = None,
       sinks: Array | None = None,
       index_mask: Array | None = None,
+      record_max_logits: bool = False,
```
RissyRan, line 947: With assertion in types.py, I think we could directly call it using self.config.use_qk_clip in attention op?

src/MaxText/layers/attention_op.py
```diff
@@ -881,8 +887,28 @@ def apply_attention(
...
           )
+        if record_max_logits:
+          out, max_logits = self.tpu_flash_attention(
+              query,
+              key,
+              value,
+              decoder_segment_ids,
+              self.attn_logits_soft_cap,
+              sinks,
+              record_max_logits=record_max_logits,
```
RissyRan: Similar to use config.use_qk_clip here?

src/MaxText/layers/attention_op.py
```diff
@@ -881,8 +887,28 @@ def apply_attention(
               """Decode not supported with flash attention.
                               Use `dot_product` instead."""
           )
+        if record_max_logits:
```
RissyRan: Do you think we could combine this if/else by checking:

```
# not None
if max_logits:
    self.sow("intermediates", "max_logits", max_logits)
```

src/MaxText/layers/attention_op.py
```diff
@@ -1297,9 +1340,24 @@ def wrap_flash_attention(
...
+
+          attention_output, max_logits = jax.vmap(kernel_fn, in_axes=(0, 0, 0, 0, None))(
+              query, key, value, decoder_segment_ids_tuple, sinks
+          )
+          return attention_output, max_logits
+        else:
+          attention_output = jax.vmap(lambda q, k, v, d, s: kernel(q, k, v, d, sinks=s), in_axes=(0, 0, 0, 0, None))(
+              query, key, value, decoder_segment_ids_tuple, sinks
+          )
+          return attention_output
```
RissyRan: Building on the previous comment: should we return `attention_output, None` here to align the two return paths? This would prevent potential compilation issues tied to the output_stats_flag logic

src/maxtext/layers/attention_op.py
```diff
@@ -1312,10 +1370,16 @@ def wrap_flash_attention(
             mask=materialized_mask,
             mask_value=DEFAULT_MASK_VALUE,
         )
+        if output_stats_flag:
```
RissyRan: Add early check in types.py?

src/maxtext/layers/attention_op.py
```diff
@@ -1348,9 +1412,22 @@ def _maybe_shard_with_pspec(inputs, pspec: jax.sharding.PartitionSpec | None):
...
+      x, max_logits = ret
+      x = jnp.transpose(x, axes=(0, 2, 1, 3))
+
+      # Max over sequence length (dim 2 of max_logits)
+      # max_logits from kernel is (batch, heads, q_len)
+      # output needs to be (batch, heads)
+      # Note: q_len is sharded. We first reduce locally.
+      max_logits_local = jnp.max(max_logits, axis=2)
+
+      return x, max_logits_local
```
RissyRan, line 1675: Similar comment. When a function returns a single value in one branch and a tuple in another, it forces the caller to know the internal state of your flags. Potentially something like:

```
x_transformed = jnp.transpose(x, axes=(0, 2, 1, 3))
    
    if record_max_logits:
        max_logits_local = jnp.max(max_logits, axis=2)
        return x_transformed, max_logits_local

    return x_transformed, None
```

src/MaxText/layers/attention_op.py
```diff
@@ -1850,7 +1957,8 @@ def __call__(
       else:
         return self.normalize_attention(unnormalized_outputs, exponentials_maxes, exponentials_sums)
     else:
-      return prefill_unnormalized_output / prefill_exponentials_sum
+      out = prefill_unnormalized_output / prefill_exponentials_sum
```
RissyRan: Nit: unnecessary change?

src/maxtext/layers/attention_op.py
```diff
@@ -1838,6 +1943,8 @@ def __call__(
         wv_product_einsum=self.AqtEinsum_3,
     )
 
+    ar_unnormalized_output, ar_exponentials_max, ar_exponentials_sum = ar_ret  # AR doesn't return stats
```
RissyRan: nit: unnecessary change?

src/maxtext/layers/attention_op.py
```diff
@@ -1814,18 +1909,28 @@ def __call__(
         bidirectional_mask=bidirectional_mask,
         sinks=sinks,
         index_mask=index_mask,
+        record_max_logits=record_max_logits,
         qk_product_einsum=self.AqtEinsum_0,
         wv_product_einsum=self.AqtEinsum_1,
     )
 
+    # Unpack consistently regardless of record_max_logits
+    prefill_unnormalized_output, prefill_exponentials_max, prefill_exponentials_sum = ret
```
RissyRan: Unnecessary?

tests/unit/qk_clip_test.py
```diff
@@ -0,0 +1,503 @@
...
+  def setUp(self):
+    """Initializes the configuration for each test using pyconfig."""
+    super().setUp()
+    if not is_decoupled():
+      jax.config.update("jax_remove_size_one_mesh_axis_from_type", True)
+
+    args = [sys.argv[0], get_test_config_path()]
+    self.cfg = pyconfig.initialize(args, **self.config_arguments)
+
+  def _run_mla_pass(self, config_overrides=None):
```
RissyRan, line 269: Do you think we could add into train_compile_test instead if no correctness check, including dot_product and flash integration.

src/maxtext/utils/qk_clip_utils.py
```diff
@@ -0,0 +1,165 @@
...
+  jax.tree_util.tree_map_with_path(extract_logits, intermediate_outputs)
+
+  if not all_max_logits:
+    return None
+
+  # Reduce each layer individually first to avoid
+  # creating a massive concatenated array of all logits.
+  layer_maxes = [jnp.max(m) for m in all_max_logits]
+  global_max_logit = jnp.max(jnp.stack(layer_maxes))
+  return global_max_logit
```
RissyRan: Gemini suggested: `return jnp.max(jnp.stack(all_max_logits))`. Try?

src/maxtext/utils/qk_clip_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+  tau = float(config.qk_clip_threshold)
+
+  def clip_mla_weights(path, param):
+    # path matches keys in state.params, e.g. ('decoder', 'layers_0', 'self_attention', 'wq_b', 'kernel')
+
+    # 1. Attempt to locate corresponding max_logits in intermediate_outputs
+    curr = intermediate_outputs
+    try:
```
RissyRan: Gemini suggested:

```
def _get_key(node):
    """Unwraps JAX DictKey or SequenceKey to a raw string/int."""
    return getattr(node, "key", getattr(node, "idx", node))

path_keys = [_get_key(p) for p in path]
```

src/maxtext/utils/qk_clip_utils.py
```diff
@@ -0,0 +1,165 @@
...
+        w_kc = param[..., :qk_nope]
+        w_v = param[..., qk_nope:]
+
+        scale_b = scale[None, :, None]
+        w_kc_new = w_kc * jnp.sqrt(scale_b)
+        return jnp.concatenate([w_kc_new, w_v], axis=-1)
+
+    except Exception as e:  # pylint: disable=broad-except
+      # Fail safe: return param as is if anything goes wrong
+      max_logging.log(f"QK-Clip exception for param at path {path}: {e}")
```
RissyRan: Curious, what could be the cause of this exception, and why it's safe failure? Thanks!

src/maxtext/utils/qk_clip_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    try:
+      # Traverse down to the layer level
+      for node in path[:-2]:
+        # Handle JAX tree_util Key objects if present
+        if hasattr(node, "key"):
+          key = node.key
+        elif hasattr(node, "idx"):
+          key = node.idx
+        else:
+          key = node
```
shuningjin: nit: perhaps resue `_get_key_name`

tests/unit/qk_clip_test.py
```diff
@@ -0,0 +1,503 @@
+# Copyright 2023–2026 Google LLC
```
shuningjin, line 1: move this file to inside `tests/unit`?

tests/utils/qk_clip_test.py
```diff
@@ -0,0 +1,503 @@
...
+    intermediates = {}
+
+    with self.assertRaisesRegex(ValueError, "QK-Clip is only supported for MLA attention"):
+      apply_qk_clip(state, intermediates, config)
+
+  def test_apply_qk_clip_logic(self):
+    """Tests QK Clip math and application logic on CPU."""
+    # 1. Setup Mock Data
+    wq_b = jnp.ones((2, 2, 6))  # [rank, heads, dim]
+    wkv_b = jnp.ones((2, 2, 6))
```
shuningjin: nit: could you try random weights (e.g., jax.random.normal) to ensure?

tests/unit/qk_clip_test.py
```diff
@@ -0,0 +1,503 @@
...
+    # Intermediates dict is empty
+    intermediates = {}
+
+    # Should not crash, should return original params
+    new_state = apply_qk_clip(state, intermediates, config)
+    new_wq = new_state.params["decoder"]["layers_0"]["self_attention"]["wq_b"]["kernel"]
+
+    self.assertTrue(jnp.array_equal(new_wq, wq_b))
+
+  def test_dot_product_max_logits(self):
```
shuningjin, line 178: Maybe move these max_logits test to a new class?

src/maxtext/layers/attention_mla.py
shuningjin: Currently `model_mode` is not part of config, but set by the calling script.
- If we call `train.py`, it use train
- If we call `decode.py`, it use prefill -> autoregressive

Not sure if it can be easily checked via type.py

src/maxtext/layers/attention_mla.py
```diff
@@ -1028,15 +1028,26 @@ def __call__(
...
-      out = self.attention_op(query, key, value, decoder_segment_ids, model_mode, cached_values, index_mask=index_mask)
+      out = self.attention_op(
+          query,
+          key,
+          value,
+          decoder_segment_ids,
+          model_mode,
+          cached_values,
+          index_mask=index_mask,
+          record_max_logits=use_qk_clip,
```
shuningjin, line 1051: 1. Would it be valuable to record max logit, even if not use_qk_clip, as a metric of training stability? (If so, we can have a separate flag)
2. Curious: How is it different from [config.use_max_logit_estimate](https://github.com/AI-Hypercomputer/maxtext/blob/f62ee44b1f3f0bdf6a78469dbc7b9f9998d83fb3/src/MaxText/configs/base.yml#L917C1-L917C23)?

src/maxtext/layers/attention_op.py
```diff
@@ -1649,7 +1727,23 @@ def apply_attention_dot(
       attn_mask = partitioning.with_sharding_constraint(attn_mask, (BATCH, HEAD, None, PREFILL_LENGTH, KV_LENGTH))
     if attn_mask is not None:
       attn_weights = apply_mask_to_logits(attn_weights, attn_mask)
-    return self.compute_local_attention(attn_weights, value, q_seq_len, model_mode, wv_product_einsum, sinks)
+
+    # We record max logits AFTER soft-capping and masking to match Flash/Splash attention behavior.
```
shuningjin, line 2065: I think the purpose of qk-clip is to suppress max logit before soft-capping?

(1) If we don't use qk-clip along with soft-cap, it doesn't matter. In deepseek3/kimi-k2, I think it does not use soft-cap.

(2) But if we are to use them together in other cases, may need max-logit BEFORE soft-cap . From Kimi-K2 paper: https://screenshot.googleplex.com/62yuZCVTHqkDcLn

> For instance, logit soft-cap [70] directly clips the attention logits, but the dot products between queries and keys can still grow excessively before capping is applied.

(Can be follow-up. For now, perhaps have a type check to say qk-clip is not fully supported along with soft-cap?)

src/maxtext/configs/base.yml
gagika: done, thanks

src/maxtext/layers/attention_mla.py
gagika: agree with Shuning, `model_mode` is injected dynamically at runtime by the executing script (train.py vs decode.py), so we can't reliably validate it inside the static types.py schema config. I'll leave the runtime check here to ensure we only collect these stats when actively training.

tests/unit/qk_clip_test.py
gagika: done

tests/utils/qk_clip_test.py
gagika: sure, done

tests/unit/qk_clip_test.py
gagika: Done. I extracted the max logits tests into a separate MaxLogitsMetricTest class within the same file to keep them logically separated from the core clipping math tests.

src/maxtext/layers/attention_op.py
gagika: done

src/maxtext/layers/attention_op.py
gagika: You are absolutely right, great catch! As the Kimi-K2 paper mentions, soft-capping hides the exploding dot products, which is the exact problem QK-Clip is trying to measure and fix. If we record max_logits after soft-capping, the values will be artificially bounded, and the QK-Clip threshold would never be breached.

Following your suggestion, I have added a cross-field validation check in types.py that raises a ValueError if a user tries to enable both use_qk_clip and attn_logits_soft_cap at the same time. If we ever need to use both simultaneously in the future, we can do a follow-up to refactor the kernels to return pre-capped, post-masked stats!"

src/maxtext/utils/qk_clip_utils.py
gagika: Done

src/maxtext/layers/attention_op.py
gagika: done

src/maxtext/layers/attention_mla.py
gagika: Yes, it would definitely be valuable as a general training stability metric.
We can easily add a separate flag (e.g., config.record_max_logit_metric) in a follow-up PR to enable monitoring.

Regarding your second question: in MaxText config.use_max_logit_estimate is a static estimate for MaxLogit that tokamax uses for an optimization.
There are two options we could consider when qk_clip is enabled:
1. qk_clip_threshold is 100.0, passing use_max_logit_estimate: 110.0 (10% buffer) 
2. using actual max logit from step N-1 to step N dynamically (seams tokamax supports that).

I think 1st is preferable.

src/maxtext/layers/attention_op.py
gagika: I wanted to keep it separate from use_qk_clip as there is a benefit for logging record_max_logits metric in Tensorboard when  use_qk_clip is False (through a separate flag) for convergence debugging.

tests/unit/qk_clip_test.py
gagika: Sure, I can add it in a follow up small PR.

src/maxtext/utils/qk_clip_utils.py
gagika: I think it was throwing exception for some non attention weights and 
It was 'safe' because catching the error and returning the unmodified param simply meant we skipped the QK-clip logic for all those irrelevant, non-attention weights.

I have updated the code not to rely on exception, e.g. added if len(path) < 2:) and dictionary key checks.

src/maxtext/utils/qk_clip_utils.py
gagika: yes updated to it return jnp.max(jnp.stack(all_max_logits)), 
my previous implementation also Gemini suggested, but was over optimized uneasily.

src/MaxText/layers/attention_op.py
gagika: done

src/MaxText/layers/attention_op.py
gagika: in terms of seperations of concerns, attention op records max_logits, training loop applies qk_clipping. 
attention_op doesn't need to know about config.use_qk_clip, and  record_max_logits later can be used for logging even if qk clip is not enabled.

src/maxtext/configs/base.yml
shuningjin: It seems not all attention kernel are supported.
https://github.com/AI-Hypercomputer/maxtext/blob/7f0028b6210453fa8be2c4165fbb06efbc52e576/src/maxtext/configs/base.yml#L326
- attention="dot_product" [yes]
- attention="flash" and use_tokamax_splash=true  [yes]
- attention="flash" and use_tokamax_splash=fase  [no] (either use_jax_splash or not)
- attention="cudnn_flash_te" [no] 

1. Do you think we should add to types.py: for early check and readability? like
https://github.com/AI-Hypercomputer/maxtext/blob/7f0028b6210453fa8be2c4165fbb06efbc52e576/src/maxtext/configs/types.py#L2232-L2237

2. When attention="autoselected", the kernel is dynamically determined. So these lines will still be needed.
https://github.com/AI-Hypercomputer/maxtext/blob/ab896ea1d11852601e574700cdc412e33422d8d7/src/maxtext/layers/attention_op.py#L1606-L1615

src/maxtext/layers/attention_op.py
shuningjin, line 1063: Maybe add some safe-guard against these gpu kernels? For gpu, it seems returning None for max logits, and qk clip is silently bypassed.

(I deleted the last comment about type check.)

gagika: I added a check with validate_gpu_flash_attention

```
def validate_gpu_flash_attention(sinks: Array | None, record_max_logits: bool) -> None:
  """Helper function to check for unsupported features with flash attention on GPU."""
  if sinks is not None:
    raise ValueError(
        "The flash attention with sinks is not supported on GPU yet."
    )
  if record_max_logits:
    raise NotImplementedError(
        "record_max_logits (QK-Clip) is not supported for GPU flash attention kernels yet."
    )
```
please have a look

shuningjin: Thanks, LGTM!

src/maxtext/layers/attention_mla.py
RissyRan: Yeah, `use_max_logit_estimate` is perf optimization strategy, and sometimes causes None loss if doesn't set it well.

PR #2970, opened 2026-01-20
Support system role in hf data processing

ChingTsai:
# Description
- Supporting `system` role in hf data processing
- Add test case for Qwen in `sft_data_processing_test.py`
- Refactoring `sft_data_processing_test.py` to support testing multiple tokenizer
- Fix incorrect left shift in targets_segmentation in `shift_and_refine`
   - see more details [here](https://github.com/AI-Hypercomputer/maxtext/pull/2970/files#r2707486614) 

FIXES: b/471740714

# Tests
### Unit test
```
MAXTEXT_PKG_DIR=<xxx>/maxtext/src/MaxText pytest tests/sft_data_processing_test.py
```
### E2E sft
```
python3 -m MaxText.sft.sft_trainer \
   src/MaxText/configs/sft.yml \
    run_name=$(date +%Y-%m-%d-%H-%M-%S) \
    base_output_directory=<xxx> \
    model_name=qwen3-4b \
    load_parameters_path=<xxx>/0/items \
    tokenizer_path=Qwen/Qwen3-4B \
    steps=53 \
    profiler=xplane \
    hf_path=arrow \
    dataset_type=hf \
    train_split=train \
    hf_train_files=<xxx>.arrow \
    hf_eval_files=<xxx>.arrow \
    per_device_batch_size=16 \
    max_target_length=1024 \
    learning_rate=5e-6 \
    warmup_steps_fraction=0.05 \
   data_shuffle_seed=42 \
    gradient_clipping_threshold=1 \
    weight_dtype=bfloat16
```

#### logs
[unpatched gpaste](https://paste.googleplex.com/4921812016300032)
[patched gpaste](https://paste.googleplex.com/4715041083490304)

Step 1 loss drop from `6.100826 ` to `5.839514 `

# Checklist

### DIFF
--- a/src/MaxText/input_pipeline/_input_pipeline_utils.py
+++ b/src/MaxText/input_pipeline/_input_pipeline_utils.py
@@ -170,30 +170,37 @@ def apply_chat_template(example, tokenizer_model, data_column_name):
       - The `data_column_name` column will be updated to a list of
         messages, each formatted according to the tokenizer's chat template.
       - A new column named "is_prompt" will be added, where `True`
-        indicates a user message (prompt) and `False` indicates an assistant
+        indicates a system message or a user message (prompt) and `False` indicates an assistant
         message (completion).
   """
   messages = []
   is_prompt = []
-  prompt = None
+  round_msgs = []
   try:
-    for message in example[data_column_name]:
-      if message["role"] == "user":
-        prompt = message
+    for idx, message in enumerate(example[data_column_name]):
+      if message["role"] == "system":
+        if idx != 0:
+          raise ValueError(f"System message found at index {idx}. System messages must be at index 0.")
+        round_msgs.append(message)
+      elif message["role"] == "user":
+        round_msgs.append(message)
         prompt_in_chat_template = tokenizer_model.apply_chat_template(
-            [prompt], add_generation_prompt=False, tokenize=False
+            round_msgs, add_generation_prompt=False, tokenize=False
         )
         messages.append(prompt_in_chat_template)
         is_prompt.append(True)
       elif message["role"] == "assistant":
+        round_msgs.append(message)
         prompt_completion_tokens = tokenizer_model.apply_chat_template(
-            [prompt, message], add_generation_prompt=False, tokenize=True
+            round_msgs, add_generation_prompt=False, tokenize=True
         )
-        prompt_tokens = tokenizer_model.apply_chat_template([prompt], add_generation_prompt=False, tokenize=True)
+        prompt_tokens = tokenizer_model.apply_chat_template(round_msgs[:-1], add_generation_prompt=False, tokenize=True)
         completion_tokens = prompt_completion_tokens[len(prompt_tokens) :]
         completion_in_chat_template = tokenizer_model.decode(completion_tokens, skip_special_tokens=False)
         messages.append(completion_in_chat_template)
         is_prompt.append(False)
+        # Round ended, clearing the buffer.
+        round_msgs.clear()
   except ValueError as e:
     max_logging.log(f"Unable to apply chat template: {e}")
     raise e
@@ -688,6 +695,7 @@ def shift_left(x, pad_id, axis=1):
 def shift_and_refine(x, ignored_ids, axis=1):
   """Shift inputs, set segmentation to 0 when target element is in ignored_ids if provided"""
   x["targets"] = shift_left(x["targets"], ignored_ids[0], axis=axis)
+  x["targets_segmentation"] = shift_left(x["targets_segmentation"], 0, axis=axis)
   for ignore_id in ignored_ids:
     x["targets_segmentation"] = np.where(x["targets"] != ignore_id, x["targets_segmentation"], 0)
 
--- a/src/MaxText/input_pipeline/_hf_data_processing.py
+++ b/src/MaxText/input_pipeline/_hf_data_processing.py
@@ -31,6 +31,16 @@ from MaxText.input_pipeline import instruction_data_processing
 from MaxText import multihost_dataloading
 
 
+def _get_pad_id(tokenizer):
+  if tokenizer.pad_token_id is not None:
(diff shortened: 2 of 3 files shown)


src/MaxText/input_pipeline/_input_pipeline_utils.py
```diff
@@ -688,6 +695,7 @@ def shift_left(x, pad_id, axis=1):
 def shift_and_refine(x, ignored_ids, axis=1):
   """Shift inputs, set segmentation to 0 when target element is in ignored_ids if provided"""
   x["targets"] = shift_left(x["targets"], ignored_ids[0], axis=axis)
+  x["targets_segmentation"] = shift_left(x["targets_segmentation"], 0, axis=axis)
```
ChingTsai, line 698: For Llama 2, since token_id `0` is `unknown_id`, the segmentation for pad_ids (added during packing) is updated correctly in lines 699-700. However, for Qwen3, token_id `0` is `!`,  therefore, we need to apply an explicit left shift to the segmentation here.

SurbhiJainUSC: Looks like this shifting is model specific. How do we ensure that it works for other models?

SurbhiJainUSC:
Please rebase to fix Documentation Build CI.

src/MaxText/input_pipeline/_input_pipeline_utils.py
aireenmei: Not sure I understand here. target_segmentation is a mask used here (https://github.com/AI-Hypercomputer/maxtext/blob/f9b1af066049a1a99252af92e4ffb1939ac632ec/src/MaxText/train.py#L152). 0 means masked token, unrelated to the token_id `0`

ChingTsai: Hi @SurbhiJainUSC and @aireenmei,

This change fixes an existing shift-left operation, so it  applies only to models using the `shift=True` in their processing pipeline.

As for this fix there, Please correct me if I'm wrong: Is `shift_and_refine` supposed to 1) shift targets left by one position, append a pad_token to the right, 2) synchronize `targets_segmentation ` by identifying all `ignore_id` positions in the new `targets` ?
In the original implementation, `targets_segmentation` did not shift alongside `targets`. Consequently, the first padding token added during packing was incorrectly assigned a segmentation value of `1`.

As shown in the example below, we currently see `!XXXXX` instead of the expected six X tokens at the end of the sequence. This occurs because the 0 (pad_token) still has its segmentation value set to `1`.
The proposed changes are intended to correct this behavior. Please let me know if this is the intended logic.

Many Thanks!

```python
import transformers
import numpy as np
from MaxText.input_pipeline._input_pipeline_utils import shift_left
def prepare_x():
    return {
        'inputs': np.array([[151644, 872, 198, 7841, 151645, 198, 151644, 77091, 198, 151667, 271, 151668, 271, 9217, 151645, 198, 0, 0, 0, 0]]),
        'inputs_segmentation': np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0]]),
        'inputs_position': np.array([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 0, 0, 0]]),
        'targets': np.array([[151643, 151643, 151643, 151643, 151643, 151643, 151644, 77091, 198, 151667, 271, 151668, 271, 9217, 151645, 198, 0, 0, 0, 0]]),
        'targets_segmentation': np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0]]),
        'targets_position': np.array([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 0, 0, 0]])
    }
qwen3_tok = transformers.AutoTokenizer.from_pretrained("Qwen/Qwen3-4B",
        add_bos_token=False,
        add_eos_token=False,
        legacy=False,)

qwen3_x = prepare_x()
print("Before Shift:" + repr(qwen3_tok.decode(np.where(qwen3_x["targets_segmentation"][0] > 0, qwen3_x["targets"][0], qwen3_tok.encode("X")[0]))))
shift_and_refine(qwen3_x, ignored_ids=[151643, None])
print("After Shift:" + repr(qwen3_tok.decode(np.where(qwen3_x["targets_segmentation"][0] > 0, qwen3_x["targets"][0], qwen3_tok.encode("X")[0]))))
```
```
Before Shift:'<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|im_start|>assistant\n<think>\n\n</think>\n\nanswer<|im_end|>\nXXXX'
After Shift:'XXXXX<|im_start|>assistant\n<think>\n\n</think>\n\nanswer<|im_end|>\n!XXXX'
```

aireenmei: Thanks for the example. I understand it now. I see the problem is more on the packing side, we should use the padding token defined by the tokenizer instead of using 0 to pad packed sequence. In this example, if packing use 151643 to pad "inputs" and "targets", then in the shift function, the step "2) synchronize targets_segmentation  by identifying all ignore_id positions in the new targets" would convert that last token's targets_segmentation to 0. But since this hf pipeline is using a legacy packing Op that doesn't support custom padding token, your change of shifting targets_segmentation is the easier fix for this. This will fix for all the models that use a tokenizer where token_id 0 maps to meaningful token, and won't impact when the tokenizer actually use 0 as pad_id.

ChingTsai: Got it, that makes sense. Thanks for clarifying!

SurbhiJainUSC (approved):

PR #2971, opened 2026-01-20
Add caching for GatedDeltaNetCache (linear attention models)

Rohan-Bierneni:
### Problem
The Qwen3-Next architecture utilizes a hybrid approach consisting of standard Self-Attention layers and Gated Delta Network (GDN) layers (a form of Linear Attention).
While MaxText’s existing KVCache handles standard attention (growing Key/Value history), it does not support the state management required for GDN. GDN layers require:

  * **A Fixed-size Recurrent State**: Compresses history into a constant-size matrix, rather than growing linearly with sequence length.

  * **A Convolution State**: A sliding window buffer for the short 1D convolution preceding the delta rule.

Without these, Qwen3-Next could not perform efficient autoregressive decoding; it would either need to recompute the entire history at every step or fail to capture temporal dependencies correctly. 

This PR adds caching support to the Gated Delta Net implementation for Qwen3-Next. This is the first non-standard attention caching support in MaxText and required changes to both the MaxEngine code and the Qwen3-Next implementation.

Note: Decoding is currently ONLY supported in unscanned checkpoint format (scan_layers=False and stack_prefill_result_cache=False).

### Solution
This PR implements specific caching infrastructure for Qwen3-Next's linear attention mechanism and integrates it natively into the MaxEngine workflow.

#### 1. New Cache Implementation (src/MaxText/inference/kvcache.py)

  * Introduced GatedDeltaNetCache. Unlike the standard KV cache, this stores fixed-size states:
    * recurrent_state: Shape (B, H, K, V)
    * conv_state: Shape (B, Kernel-1, Dim)
  * Both variables are wrapped in nn.with_logical_constraint to ensure MaxText can automatically infer proper shardings. 
  * Created a BaseCache class to abstract commonalities between standard KV and GDN caches.

#### 2. Layer Logic & State Management (src/MaxText/layers/qwen3.py)

  * Conv & Recurrent Updates: Modified Qwen3NextGatedDeltaNet to properly slide the convolution window and update the recurrent state during MODEL_MODE_AUTOREGRESSIVE.

  * Chunked Prefill Support: Re-wrote MODEL_MODE_PREFILL logic to read from self.cache rather than defaulting to zeros. This allows the GDN state to seamlessly carry over between chunks for long prompts.

  * Dynamic Batch Alignment: Added jnp.broadcast_to and [:batch] slicing during cache retrieval. This dynamically bridges the gap between MaxEngine's global cache allocation (e.g., batch=64) and the active compute shape (e.g., batch=1 during prefill), preventing concatenation shape errors during XLA dummy-tracing passes.

#### 3. MaxEngine Integration (src/MaxText/maxengine.py)

  * Non-sequence-growing Insertions: Updated _insert_jit and bulk_insert to handle fixed-size states. For GDN variables (recurrent_state, conv_state), the prefill step produces a "final state" which is directly copied into the decode cache slot (bypassing the sequence-length slicing logic used for standard KV caches).

  * Packed Prefill Guard: Added a NotImplementedError in insert_partial to block packed prefill for GDN states, preventing sequential memory from bleeding across independent packed prompts.

  * Logical Axes Fix: Updated init_decode_state to properly check for both .logical_axes and .names when unpacking LogicallyPartitioned variables, fixing a bug where empty tuples caused batch index crashes.

FIXES: b/448407748

# Tests

Decode.py tests:
  * On v5p-64 & pdb=1, global_batch_size=32: https://paste.googleplex.com/5916785586077696
  * On CPU vm with pdb=1: https://paste.googleplex.com/5608748317016064

# Checklist

### DIFF
--- a/src/maxtext/models/qwen3.py
+++ b/src/maxtext/models/qwen3.py
@@ -42,8 +42,10 @@ from maxtext.layers.attentions import Attention
 from maxtext.layers.linears import DenseGeneral, MlpBlock
 from maxtext.layers.moe import RoutedMoE
 from maxtext.layers.initializers import nd_dense_init, variable_to_logically_partitioned
-from maxtext.inference import page_manager
+
 from maxtext.utils import max_utils
+from maxtext.inference import page_manager, kvcache
+
 
 # -----------------------------------------
 # Qwen3-Next Layer Implementations
@@ -379,7 +381,7 @@ class Qwen3NextGatedDeltaNet(nnx.Module):
   2. output = Linear_out(y)
   """
 
-  def __init__(self, config: Config, *, rngs: nnx.Rngs):
+  def __init__(self, config: Config, dtype: DType = jnp.float32, model_mode: str = MODEL_MODE_TRAIN, *, rngs: nnx.Rngs):
     """
     Args:
       config: MaxText configuration object.
@@ -398,6 +400,17 @@ class Qwen3NextGatedDeltaNet(nnx.Module):
     conv_dim = self.key_dim * 2 + self.value_dim
     conv_kernel_size = cfg.gdn_conv_kernel_dim
     self.v_heads_per_k_head = self.num_v_heads // self.num_k_heads
+
+    if model_mode != MODEL_MODE_TRAIN:
+      self.cache = kvcache.GatedDeltaNetCache(
+          batch=config.per_device_batch_size,
+          num_heads=self.num_v_heads,
+          k_head_dim=self.head_k_dim,
+          v_head_dim=self.head_v_dim,
+          conv_kernel_size=self.config.gdn_conv_kernel_dim,
+          conv_dim=conv_dim,
+          dtype=dtype,
+      )
 
     # Submodule instantiations
     self.in_proj_qkvz = DenseGeneral(
@@ -458,7 +471,14 @@ class Qwen3NextGatedDeltaNet(nnx.Module):
         rngs=rngs,
     )
 
-  def __call__(self, hidden_states: Array) -> Array:
+  def __call__(
+      self,
+      hidden_states: Array,
+      model_mode: str = MODEL_MODE_TRAIN,
+      kv_cache=None,
+      decoder_segment_ids: None | Array = None,
+      **kwargs,
+  ) -> tuple[Array, dict[str, Array]]:
     # hidden_states: (B, S, E)
     cfg = self.config
     batch, seq_len, _ = hidden_states.shape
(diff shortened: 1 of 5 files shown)


src/MaxText/layers/decoders.py
```diff
@@ -895,8 +915,20 @@ def __call__(
...
+                is_full_attention_layer = (lyr + 1) % cfg.inhomogeneous_layer_cycle_interval == 0
+                if is_full_attention_layer:
+                  kv_caches["key_cache"][lyr] = returned_cache[0]
+                  kv_caches["value_cache"][lyr] = returned_cache[1]
+                else:
+                  # gdn_cache = returned_cache.get("gdn_cache")
+                  # if gdn_cache:
+                  #   kv_caches["conv_states"][lyr] = gdn_cache["conv_state"]
+                  #   kv_caches["recurrent_states"][lyr] = gdn_cache["recurrent_state"]
+                  pass
```
parambole: Please remove the commented section.

src/MaxText/layers/qwen3.py
```diff
@@ -326,6 +327,17 @@ def __init__(self, config: Config, dtype: DType = jnp.float32, *, rngs: nnx.Rngs
     conv_kernel_size = cfg.gdn_conv_kernel_dim
     self.v_heads_per_k_head = self.num_v_heads // self.num_k_heads
 
+    if model_mode != MODEL_MODE_TRAIN:
+      self.cache = kvcache.GatedDeltaNetCache(
+          batch=config.per_device_batch_size, # Or appropriate batch dim
```
parambole: Please remove inline comment here and below if not necessary.

src/MaxText/layers/qwen3.py
```diff
@@ -496,10 +532,31 @@ def __call__(self, hidden_states: Array) -> Array:
...
+        initial_state=recurrent_state,
+        use_qk_norm_in_gdn=cfg.use_qk_norm_in_gdn,
     )
+    
+    if model_mode != MODEL_MODE_TRAIN:
+        # Update self.cache in place for both prefill and decode
+        self.cache.recurrent_state.value = recurrent_state_out
+        
+    # Construct return dictionary for compatibility with decoders.py (optional but safer)
+    new_cache = {
```
parambole: Is this necessary I don't see it being used?

src/MaxText/layers/decoders.py
Rohan-Bierneni: Resolved.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Resolved.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: Yes, thanks for catching this I removed it and refactored the function signature.

RissyRan (commented):
For this one, let's add one unit test like https://github.com/AI-Hypercomputer/maxtext/blob/ec712bf3c7e490bcb0171c6e93e5ac1ac3bdeaa6/tests/unit/attention_test.py#L1195-L1241 to protect any future changes break your feature.

tests/unit/attention_test.py
```diff
@@ -1504,5 +1505,91 @@ def test_tpu_flash_attention_context_parallel(
...
+    for idx in range(prefill_length, decode_total_length):
+      lnx_idx = lnx[:, idx : idx + 1, :]
+
+      gdn_idx = gdn(
+          lnx_idx,
+          model_mode=MODEL_MODE_AUTOREGRESSIVE,
+      )
+
+      gdn_full_this_idx = gdn_full[:, idx : idx + 1, :]
+      self.assertEqual(gdn_full_this_idx.shape, gdn_idx.shape)
```
RissyRan, line 1589: Is the value matching? The example is from MLA, and I tested locally it's matching. If not, you may want to try dtype=f32.

Rohan-Bierneni: Yes, we are checking the shapes and the values themselves via jax.numpy.allclose(). The test passed locally in TPU vm and should have passed as part of tpu_unit_tests in this PR.

src/maxtext/models/qwen3.py
```diff
@@ -529,15 +545,42 @@ def __call__(self, hidden_states: Array) -> Array:
...
+    conv_state = None
+    if model_mode != MODEL_MODE_TRAIN:
+      # Retrieve state from self.cache
+      conv_state = self.cache.conv_state.value
+
+      if conv_state.shape[0] != batch:
+        if conv_state.shape[0] == 1:
+          conv_state = jnp.broadcast_to(conv_state, (batch,) + conv_state.shape[1:])
+        else:
+          conv_state = conv_state[:batch]
```
parambole, line 565: It assumes the cache (when shape[0] == 1) is zero-initialized and safe to duplicate.

Just to illustrate the edge case: If a test were to process the word 'Hello' with a batch size of 1, the `conv_state` would update to reflect 'Hello'. If the test then immediately scaled up to a batch size of 2, this logic would broadcast that 'Hello' state into the second slot, meaning both batch items would now share the exact same history.

Could we add a quick inline comment here (e.g., # Assumes zero-initialized state for testing) just to clarify the intent for anyone reading this later?

PR #2979, opened 2026-01-21
Add flops calculation for DeepSeek v3.2

RissyRan:
# Description

Add flops calculation for DeepSeek v3.2, and this PR depends on [this change](https://github.com/AI-Hypercomputer/maxtext/pull/2933)

* Add indexer flops helper function
* Add option to combine indexer flops inside of MLA
* Added a unit test to **rough** estimate tflops

# Tests

* Tested DS v2-16b, no impact

```
# Before the change
Per train step:
 Total TFLOPs: 593.27 
 split as 81.24% learnable weight flops and 18.76% attention flops
before the change: 593.272422531072

# After the change
Per train step:
 Total TFLOPs: 593.27 
 split as 81.24% learnable weight flops and 18.76% attention flops
after change: 593.272422531072
```

* Tested DS v3.2, with indexer (expected)
```
# Enable this feature
Per train step:
 Total TFLOPs: 4162.71 
 split as 88.50% learnable weight flops and 11.50% attention flops
enable use_sparse_indexer: 4162.71

# Disable this feature like v3
Per train step:
 Total TFLOPs: 4103.37 
 split as 87.74% learnable weight flops and 12.26% attention flops
disable use_sparse_indexer: 4103.370952409088
```

# Checklist

### DIFF
--- a/src/MaxText/maxtext_utils.py
+++ b/src/MaxText/maxtext_utils.py
@@ -319,11 +319,86 @@ def calculate_llama4_attention_tflops(config):
   return attention_tflops
 
 
+def calculate_indexer_mask_ratio(index_topk, max_target_length):
+  """
+  Calculates the sparse-to-dense ratio for Indexer TFLOPs.
+
+  The indexer evaluates all previous tokens in a causal manner until it hits
+  the Top-K limit.
+
+  Visual Representation (T=8, K=4):
+  Key (S) ->
+  Q1 [X . . . . . . .]  <- 1 token scored
+  Q2 [X X . . . . . .]  <- 2 tokens scored
+  Q3 [X X X . . . . .]  <- 3 tokens scored
+  Q4 [X X X X . . . .]  <- 4 tokens scored (K limit reached)
+  Q5 [X X X . X . . .]  <- 4 tokens scored
+  Q6 [X X . X . X . .]  <- 4 tokens scored
+  Q7 [X . X X . . X .]  <- 4 tokens scored
+  Q8 [X X . X . . . X]  <- 4 tokens scored
+
+  For MFU calculation:
+
+  Visual Representation (T=8, K=4):
+  Key (S) ->
+  Q1 [X . . . . . . .]  <- 1 token scored
+  Q2 [X X . . . . . .]  <- 2 tokens scored
+  Q3 [X X X . . . . .]  <- 3 tokens scored
+  Q4 [X X X X . . . .]  <- 4 tokens scored (K limit reached)
+  Q5 [X X X X . . . .]  <- 4 tokens scored
+  Q6 [X X X X . . . .]  <- 4 tokens scored
+  Q7 [X X X X . . . .]  <- 4 tokens scored
+  Q8 [X X X X . . . .]  <- 4 tokens scored
+
+  Mathematical Calculation:
+  - Triangle (Phase 1: 1 to K): K^2 / 2
+  - Rectangle (Phase 2: K+1 to T): (T - K) * K
+  - Total Active Area = TK - K^2 / 2
+  - Dense Area = T^2
+
+  Ratio = (TK - 0.5*K^2) / T^2  =>  (K/T) - 0.5*(K/T)^2
+  """
+
+  T = float(max_target_length)
+  K = float(index_topk)
+
+  ratio = K / T
+  mask_multiplier = ratio - (0.5 * ratio**2)
+  return mask_multiplier
+
+
+def calculate_indexer_tflops_per_device(config):
+  """Calculates TFLOPs for the DeepSeek Lightning Indexer (handles causal reduction)."""
+  batch_len = config.per_device_batch_size * config.max_target_length
+
+  # 1. Calculate projections flops
+  # Query: [batch, seq, q_lora_rank] @ [q_lora_rank, index_n_heads, index_head_dim]
+  q_flops = 2 * batch_len * config.q_lora_rank * config.index_n_heads * config.index_head_dim
+  # Key: [batch, seq, emb_dim] @ [emb_dim, index_head_dim]
+  k_flops = 2 * batch_len * config.emb_dim * config.index_head_dim
+  # Head weight: [batch, seq, emb_dim] @ [emb_dim, index_n_heads]
+  head_weight_flops = 2 * batch_len * config.emb_dim * config.index_n_heads
+  proj_flops = q_flops + k_flops + head_weight_flops
+
+  # 2. Calculate index score flops
+  # QK product [batch, seq, index_n_heads, index_head_dim] @ [batch, seq, index_head_dim]
+  # --> [batch, seq, seq, index_n_heads]
+  qk_product_flops = 2 * batch_len * config.max_target_length * config.index_n_heads * config.index_head_dim
+  # Aggregate heads [batch, seq, seq, index_n_heads] @ [batch, seq, index_n_heads]
+  head_reduction_flops = 2 * batch_len * config.max_target_length * config.index_n_heads
+  # Apply causal mask: Divide by 2 to account for triangular interactions
+  # The mask restricts the indexer's search space prior to Top-K filtering
+  scoring_flops = (qk_product_flops + head_reduction_flops) / 2
+
+  return proj_flops, scoring_flops
+
+
 def calculate_mla_tflops_per_device(config):
-  """Calculate Multi-Head Latent Attention TFLOP"""
+  """Calculate Multi-Head Latent Attention TFLOP (handles causal reduction)"""
   batch_len = config.per_device_batch_size * config.max_target_length
   qk_head_dim_sum = config.qk_nope_head_dim + config.qk_rope_head_dim
-  # calculate mla query projection
+
+  # 1. calculate mla query projection
   if config.q_lora_rank == 0:
     q_flops = 2 * batch_len * config.emb_dim * config.num_query_heads * qk_head_dim_sum
   else:
@@ -333,7 +408,8 @@ def calculate_mla_tflops_per_device(config):
         * batch_len
         * (config.emb_dim * config.q_lora_rank + config.q_lora_rank * config.num_query_heads * qk_head_dim_sum)
     )
-  # calculate mla kv projection with down and up flops
+
+  # 2. calculate mla kv projection
   kv_flops = (
       2
       * batch_len
@@ -344,9 +420,31 @@ def calculate_mla_tflops_per_device(config):
   )
   qkv_flops = q_flops + kv_flops
 
-  attention_flops = (
-      2 * batch_len * config.max_target_length * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
-  )
+  # 3. calculate attention
+  if config.use_sparse_indexer and config.max_target_length > config.index_topk:
+    # get indexer flops
+    indexer_proj_flops, indexer_scoring_flops = calculate_indexer_tflops_per_device(config)
+    qkv_flops += indexer_proj_flops
+
+    # calculate the proportion of the T x T causal matrix that the Indexer actually explores
+    # this follows the area: (TK - 0.5*K^2) / T^2 (T: max_target_length, K: index_topk)
+    multiplier = calculate_indexer_mask_ratio(config.index_topk, config.max_target_length)
+    attention_flops = (
+        2
+        * batch_len
+        * config.max_target_length
+        * config.num_query_heads
+        * (qk_head_dim_sum + config.v_head_dim)
+        * multiplier
+    )
+    attention_flops += indexer_scoring_flops
+  else:
+    # standard MLA & max_target_length <= index_topk in sparse indexer
+    # in both cases, the indexer is bypassed as the causal mask remains the efficient representation
+    attention_flops = (
+        2 * batch_len * config.max_target_length * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
+    )
+    attention_flops = attention_flops / 2
   projection_flops = 2 * batch_len * config.emb_dim * config.num_query_heads * config.v_head_dim
   return qkv_flops, attention_flops, projection_flops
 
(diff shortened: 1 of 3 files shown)


src/MaxText/maxtext_utils.py
```diff
@@ -314,11 +314,35 @@ def calculate_llama4_attention_tflops(config):
...
+  # Key: [batch, seq, emb_dim] @ [emb_dim, index_head_dim]
+  # and [batch, seq, emb_dim] @ [emb_dim, index_n_heads]
+  k_flops = 2 * batch_len * config.emb_dim * (config.index_head_dim + config.index_n_heads)
+  qk_flops = q_flops + k_flops
+
+  # QK product [batch, seq, index_n_heads, index_head_dim] @ [batch, seq, index_head_dim]
+  # --> [batch, seq, seq, index_n_heads]
+  qk_product_flops = 2 * batch_len * config.max_target_length * config.index_n_heads * config.index_head_dim
+  # Apply causal mask (theoretical needed)
+  qk_product_flops = qk_product_flops / 2
```
shuningjin: Interesting to know that qk by itself also divide 2

src/MaxText/maxtext_utils.py
```diff
@@ -314,11 +314,35 @@ def calculate_llama4_attention_tflops(config):
...
+
+  # Weight sum over heads [batch, seq, seq, index_n_heads] @ [batch, seq, index_n_heads]
+  weighted_sum_flops = batch_len * config.max_target_length * config.index_n_heads
```
shuningjin: why not multiply 2? is it another causal reduction due to logits: `Score = Sum_head(Logits * Head_Weights)`? if so, would be good to update comment for clarification

src/MaxText/maxtext_utils.py
```diff
@@ -314,11 +314,35 @@ def calculate_llama4_attention_tflops(config):
...
 
+def calculate_indexer_tflops_per_device(config):
+  """Calculates TFLOPs for the DeepSeek Lightning Indexer (handles causal reduction)."""
+  batch_len = config.per_device_batch_size * config.max_target_length
+
+  # Query: [batch, seq, q_lora_rank] @ [q_lora_rank, index_n_heads, index_head_dim]
+  q_flops = 2 * batch_len * config.q_lora_rank * config.index_n_heads * config.index_head_dim
+  # Key: [batch, seq, emb_dim] @ [emb_dim, index_head_dim]
+  # and [batch, seq, emb_dim] @ [emb_dim, index_n_heads]
+  k_flops = 2 * batch_len * config.emb_dim * (config.index_head_dim + config.index_n_heads)
```
shuningjin: This looks correct. The name `k_flops` is somewhat confusing to me. Could you make the comment/code clearer:
- key projection: `Wk @ X` -- 2 * batch_len * config.emb_dim * (config.index_head_dim
- Head_Weights calculation: `W_proj @ X` -- 2 * batch_len * config.emb_dim * config.index_n_heads

src/MaxText/maxtext_utils.py
```diff
@@ -339,9 +364,25 @@ def calculate_mla_tflops_per_device(config):
...
   qkv_flops = q_flops + kv_flops
 
-  attention_flops = (
-      2 * batch_len * config.max_target_length * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
-  )
+  # 3. calculate attention (sparse indexer and standard MLA)
+  if config.use_sparse_indexer:
+    # get indexer flops
+    indexer_qk_flops, indexer_scoring_flops = calculate_indexer_tflops_per_device(config)
+    qkv_flops += indexer_qk_flops
```
shuningjin: what's the criteria to distinguish `qkv_flops` and `attention_flops`? The indexer computation make these blurry. Is it mostly distinguished by linear flops O(T) and quadratic flops O(T^2), where T is sequence length?

src/MaxText/maxtext_utils.py
```diff
@@ -339,9 +364,25 @@ def calculate_mla_tflops_per_device(config):
...
+
+    is_causal_sparse = config.max_target_length > config.index_topk
+    effective_seq = config.index_topk if is_causal_sparse else config.max_target_length
+
+    # core attention flops
+    attention_flops = 2 * batch_len * effective_seq * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
+    attention_flops = attention_flops if is_causal_sparse else (attention_flops / 2)
+    attention_flops += indexer_scoring_flops
```
shuningjin, line 440: Does flop calculation depends on the implementation? Currently, we skip the indexer computation entirely **when max_target_length <= index_topk**. should it be
-  (1) indexer flop + regular attention flop
- or (2) just regular attention flop

src/MaxText/maxtext_utils.py
```diff
@@ -339,9 +364,25 @@ def calculate_mla_tflops_per_device(config):
...
+    # get indexer flops
+    indexer_qk_flops, indexer_scoring_flops = calculate_indexer_tflops_per_device(config)
+    qkv_flops += indexer_qk_flops
+
+    is_causal_sparse = config.max_target_length > config.index_topk
+    effective_seq = config.index_topk if is_causal_sparse else config.max_target_length
+
+    # core attention flops
+    attention_flops = 2 * batch_len * effective_seq * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)
+    attention_flops = attention_flops if is_causal_sparse else (attention_flops / 2)
```
shuningjin: when **max_target_length (T) > index_topk (K)**, should it be
- (1) currently, it is TK* C
- or (2)  (K^2/2 + (T-K) K) * C = (TK - K^2/2) * C? 

where C= 2 * batch * config.num_query_heads * (qk_head_dim_sum + config.v_head_dim)

also note If $T \gg K$, we have (1) $\approx$ (2)

src/MaxText/maxtext_utils.py
```diff
@@ -314,11 +314,35 @@ def calculate_llama4_attention_tflops(config):
...
+  # Apply causal mask (theoretical needed)
+  qk_product_flops = qk_product_flops / 2
+
+  # Weight sum over heads [batch, seq, seq, index_n_heads] @ [batch, seq, index_n_heads]
+  weighted_sum_flops = batch_len * config.max_target_length * config.index_n_heads
+  scoring_flops = qk_product_flops + weighted_sum_flops
+  return qk_flops, scoring_flops
+
+
 def calculate_mla_tflops_per_device(config):
```
shuningjin, line 396: On a high level, do you think putting "index_flop inside mla_flop" is better than "index_flop + mla_flop"?

src/MaxText/maxtext_utils.py
RissyRan: No this is not about causal mask. 

When we multiply a matrix of size M x N by a matrix of size N x K, the output matrix has M x N total elements.

The total flops is 2 x M x N x K. One flop for multiplication, and one for addition. Here is the sum, which is addition directly (even it writes into a einsum, like a matmul)

src/MaxText/maxtext_utils.py
RissyRan: Sounds good. Let me update a little bit to be clear.

src/MaxText/maxtext_utils.py
RissyRan: Yeah, this is theoretical value for model utilization right? Ideally only values inside of triangle are useful. Do you think this is helpful? https://github.com/AI-Hypercomputer/maxtext/blob/55278a2b8894008400023162c38989102dcfe25c/src/MaxText/maxtext_utils.py#L571-L575

src/MaxText/maxtext_utils.py
RissyRan: Good point! I don't think Indexer is used anywhere yet. Let's move inside of MLA.

src/MaxText/maxtext_utils.py
RissyRan: You could think of QKV is projection with learnable weights, and attention_flops has pure matmul based on (Q@K) @ V

src/MaxText/maxtext_utils.py
shuningjin: Sorry, I still didn't get it. 
- The weighted sum has two parts, "weighted" means multiplication, and "sum" means addition. 
- "btsh, bth -> bts" looks like `2 B T^2 H` flops to me

https://github.com/AI-Hypercomputer/maxtext/blob/18eee32ff9c7a6ec8769329ca60a6f0931588aa3/src/MaxText/layers/attention_mla.py#L284-L285

src/MaxText/maxtext_utils.py
RissyRan: That's a good point. If we'd like to be more precise, we may need to consider how the mask applied to Indexer part. Ideally, it only applies to https://screenshot.googleplex.com/4M4TpfMrmGbiooc. Let me think about it. If this is small, I think it's ok to be ignored.

src/MaxText/maxtext_utils.py
RissyRan: Discussed offline. Yes, this is not a sum operation, but a matmul. Do you mind I updating the comment a little bit from `# Weighted sum over head: sum_h(logits * weights) ` to something like `# logits @ weights` or others? Same to https://github.com/AI-Hypercomputer/maxtext/blob/18eee32ff9c7a6ec8769329ca60a6f0931588aa3/src/MaxText/layers/attention_mla.py#L229

https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/87e509a2e5a100d221c97df52c6e8be7835f0057/inference/kernel.py#L272

shuningjin: Yes, feel free to update the comments

src/MaxText/maxtext_utils.py
shuningjin: Yes, this would be helpful

src/MaxText/maxtext_utils.py
shuningjin: In theory, if T is much larger than K,  we can omit $K^2/2$ from $TK - K^2/2$. 

In practice, the ratio $\displaystyle \frac{K^2/ 2}{TK} = \frac{K}{2T}$
- If K=2048, T=128_000 (ds3.2 setting): then ratio is 0.8%
- If K=2048, T=4096: then ratio is 25%

I am not sure whether to ignore or not. I will leave it up to you.

src/MaxText/maxtext_utils.py
RissyRan: I moved the logic into calculate_mla_tflops_per_device(), but the function is now ~100 lines long. I'm leaning toward keeping it outside to improve readability—what do you think?

shuningjin: Seems good to keep the function outside! 

I originally meant 
- "inside": what you are doing now, calling `calculate_indexer_tflops_per_device` inside `calculate_mla_tflops_per_device`
- "outside": `calculate_mla_tflops_per_device() + calculate_indexer_tflops_per_device()`

Sorry for the confusion. I am fine either way. Up to you!

src/MaxText/maxtext_utils.py
RissyRan: I created a helper function to verify end-to-end tflops. Please cross check.

```
def calculate_indexer_mask_ratio(index_topk, max_target_length):
  """
    Calculates the sparse-to-dense ratio for Indexer TFLOPs.
    
    The indexer evaluates all previous tokens in a causal manner until it hits 
    the Top-K limit.

    Visual Representation (T=8, K=4):
    Key (S) ->
    Q1 [X . . . . . . .]  <- 1 token scored
    Q2 [X X . . . . . .]  <- 2 tokens scored
    Q3 [X X X . . . . .]  <- 3 tokens scored
    Q4 [X X X X . . . .]  <- 4 tokens scored (K limit reached)
    Q5 [X X X . X . . .]  <- 4 tokens scored
    Q6 [X X . X . X . .]  <- 4 tokens scored
    Q7 [X . X X . . X .]  <- 4 tokens scored
    Q8 [X X . X . . . X]  <- 4 tokens scored

    For model utilization perspective:
    
    Visual Representation (T=8, K=4):
    Key (S) ->
    Q1 [X . . . . . . .]  <- 1 token scored
    Q2 [X X . . . . . .]  <- 2 tokens scored
    Q3 [X X X . . . . .]  <- 3 tokens scored
    Q4 [X X X X . . . .]  <- 4 tokens scored (K limit reached)
    Q5 [X X X X . . . .]  <- 4 tokens scored
    Q6 [X X X X . . . .]  <- 4 tokens scored
    Q7 [X X X X . . . .]  <- 4 tokens scored
    Q8 [X X X X . . . .]  <- 4 tokens scored
    
    Area Calculation:
    - Triangle (Phase 1: 1 to K): K^2 / 2
    - Rectangle (Phase 2: K+1 to T): (T - K) * K
    - Total Active Area = TK - K^2 / 2
    - Dense Area = T^2
    """

  T = float(max_target_length)
  K = float(index_topk)

  ratio = K / T
  mask_multiplier = ratio - (0.5 * ratio**2)
  return mask_multiplier
```

Let's take a look at the total flops:

batch=4, seq=4096 (4162.71/4036.96=3.1%)

```
IS_PRECISE 1: True
IS_PRECISE 2: True
Per train step:
 Total TFLOPs: 4162.71 (current change)
 (one version before: Total TFLOPs: 4137.36)

Turn off 1: False
Turn off 2: False
Per train step:
 Total TFLOPs: 4036.96 
```


batch=36, seq=4096 (37464.41/36332.60=3.1%)
```
IS_PRECISE 1: True
IS_PRECISE 2: True
ratio 2: 0.5
Per train step:
 Total TFLOPs: 37464.41 
 (one version before: Total TFLOPs: 37236.28 )

Turn off 1: False
Turn off 2: False
Per train step:
 Total TFLOPs: 36332.60 
```

batch=4, seq=128000 (229735.36/222001.32=3.5% diff)

```
Updated:
IS_PRECISE 1: True
IS_PRECISE 2: True
Per train step:
 Total TFLOPs: 229735.36 (current change)
 (one version before: Total TFLOPs: 133863.55)

Turn off 1: False
Turn off 2: False
Per train step:
 Total TFLOPs: 222001.32 
```

I will update this PR first, and let's chat more tomorrow.

src/MaxText/maxtext_utils.py
RissyRan: Thanks! Let's consolidate into one thread above.

src/MaxText/maxtext_utils.py
RissyRan: I see. Thanks for explanation! If you don't have strong preference, let's keep it as it is? As indexer is part of MLA, and will affect `attention_flops` calculation.

PR #2999, opened 2026-01-22
Update tflops calc for qwen3 next & GatedDeltaNet

Rohan-Bierneni:
# Description

### Description
This PR implements the TFLOPs calculation logic for the Qwen3-Next architecture. Qwen3-Next utilizes a hybrid design containing both standard full attention layers and linear attention (Gated Delta Net) layers, alongside a Mixture-of-Experts (MoE) structure. This update ensures training TFLOPs are accurately reported for this model family.

### Key Changes

**1. TFLOPs Calculation Logic (`src/MaxText/maxtext_utils.py`)**
* **Added `calculate_gated_delta_net_flops_per_device`**: Implemented logic to calculate FLOPs for the Linear Attention layers, breaking down operations into:
    * **Projections**: QKVZ, BA, and Output projections.
    * **Convolution**: Depthwise convolutions on the key/value/gate states.
    * **Core Attention**: Intra-chunk and inter-chunk recurrent state operations.
* **Updated `calculate_tflops_training_per_device`**:
    * Added a specific branch for `DecoderBlockType.QWEN3_NEXT`.
    * Logic now calculates the number of "full attention" layers vs. "linear attention" layers based on the `inhomogeneous_layer_cycle_interval`.
    * Combines FLOPs from embeddings, MoE FFNs (routed + shared), full causal attention, and linear attention.
* **Updated FFN Helpers**: Verified `calculate_routed_and_shared_ffn_tflops_per_device` and `get_dense_moe_layers` explicitly support `QWEN3_NEXT`.

**2. Unit Tests (`tests/unit/flop_calculation_test.py`)**
* **Added `test_qwen3_next_flops`**: A new unit test that verifies the implementation against a "golden" manual calculation for the 80B model configuration.
* **Added `compute_qwen3_next_attention_flops_per_device`**: A helper function to compute the expected attention-specific FLOPs for the test assertion.

**3. Config Updates (`src/MaxText/configs/models/qwen3-next-80b-a3b.yml`)**
* Added `shared_experts: 1` to the model configuration to ensure correct parameter counting in the FFN FLOPs calculation.

FIXES: b/477291633

# Tests

# Checklist

### DIFF
--- a/src/maxtext/utils/maxtext_utils.py
+++ b/src/maxtext/utils/maxtext_utils.py
@@ -488,10 +488,80 @@ def get_dense_moe_layers(config):
   elif config.decoder_block == DecoderBlockType.LLAMA4:
     num_moe_layers = config.num_decoder_layers // config.interleave_moe_layer_step
     num_dense_layers = config.num_decoder_layers - num_moe_layers
+  elif config.decoder_block == DecoderBlockType.QWEN3_NEXT:
+    num_moe_layers = config.num_decoder_layers
+    num_dense_layers = 0
   else:
-    raise ValueError("Currently we only support DeepSeek and Llama4 calculation.")
+    raise ValueError("Currently we only support DeepSeek, Llama4, and Qwen3-Next calculation.")
 
   return num_dense_layers, num_moe_layers
+
+
+def calculate_gated_delta_net_flops_per_device(config):
+  """Calculates the FLOPs for a single Gated Delta Net (Linear Attention) layer."""
+  B = config.per_device_batch_size
+  S = config.max_target_length
+  E = config.emb_dim
+
+  H_k = config.gdn_num_key_heads
+  H_v = config.gdn_num_value_heads
+  D_k = config.gdn_key_head_dim
+  D_v = config.gdn_value_head_dim
+  C = config.gdn_chunk_size
+  K_conv = config.gdn_conv_kernel_dim
+
+  K_dim = H_k * D_k
+  V_dim = H_v * D_v
+
+  # 1. Projections (Learnable Weights)
+  # in_proj_qkvz: E -> 2*K_dim + 2*V_dim
+  flops_qkvz = 2 * B * S * E * (2 * K_dim + 2 * V_dim)
+  # in_proj_ba: E -> 2*H_v
+  flops_ba = 2 * B * S * E * (2 * H_v)
+  # out_proj: V_dim -> E
+  flops_out = 2 * B * S * V_dim * E
+
+  flops_projections = flops_qkvz + flops_ba + flops_out
+
+  # 2. Convolution (Learnable Weights)
+  # Depthwise conv on dim (2*K_dim + V_dim)
+  # 2 * B * S * Channels * Kernel
+  flops_conv = 2 * B * S * (2 * K_dim + V_dim) * K_conv
+
+  # 3. Core Gated Delta Net (Attention-like operations)
+  # Assumptions:
+  # H = H_v (broadcasting K to V heads if H_v > H_k)
+  # N = num_chunks & N * C ~ S
+  #
+  # Query (Q): [B, S, H_v, D_k]
+  # Keys (K): [B, S, H_v, D_k]
+  # Values (V): [B, S, H_v, D_v]
+  # Intra-Chunk Attention (A): [B, N, H_v, C, C]
+  # Recurrent State (S): [B, N, H_v, D_k, D_v]
+
+  # - Intra-chunk terms (per chunk C):
+  #   - attn (K*K): 2 * B * S * H_v * C * D_k
+  #   - val_intra (A*V): 2 * B * S * H_v * C * D_v
+  #   - k_cum (A*K): 2 * B * S * H_v * C * D_k
+  #   - inner_attn_body loop (iterative refinement): ≈ (C - 1) * B * H * N * C^2 ≈ B * H * S * C^2
+  flops_intra = 2 * B * S * H_v * C * (2 * D_k + D_v) + (B * H_v * S * C**2)
+
+  # - Inter-chunk terms (Recurrent State D_k * D_v):
+  #   - attn_i (Q*K): 2 * B * S * H_v * C * D_k
+  #   - v_prime (K*S): 2 * B * S * H_v * D_k * D_v
+  #   - attn_inter (Q*S): 2 * B * S * H_v * D_k * D_v
+  #   - core_out (A*V): 2 * B * S * H_v * C * D_v
+  #   - update (K*V): 2 * B * S * H_v * D_k * D_v
+  flops_inter = (2 * B * S * H_v * C * (D_k + D_v)) + (6 * B * S * H_v * D_k * D_v)
+
+  flops_core = flops_intra + flops_inter
+
+  # Weights part: Projections + Conv
+  gdn_weight_flops = flops_projections + flops_conv
+  # Attention part: Core
+  gdn_attn_flops = flops_core
+
+  return gdn_weight_flops, gdn_attn_flops
 
 
 def calculate_gemma3_vision_layers_tflops_per_device(config):
@@ -634,7 +704,7 @@ def calculate_tflops_training_per_device(config, log=True):
   # MLP flops
   if config.num_experts > 1:
     # calculation based on dropless implementation
-    if config.decoder_block in (DecoderBlockType.DEEPSEEK, DecoderBlockType.LLAMA4):
+    if config.decoder_block in (DecoderBlockType.DEEPSEEK, DecoderBlockType.LLAMA4, DecoderBlockType.QWEN3_NEXT):
       total_ffn_flops = calculate_routed_and_shared_ffn_tflops_per_device(config)
     else:
       gate_flops = 2 * config.per_device_batch_size * config.max_target_length * config.emb_dim * config.num_experts
@@ -702,6 +772,24 @@ def calculate_tflops_training_per_device(config, log=True):
         (total_ffn_flops + (qkv_flops + projection_flops) * config.num_decoder_layers + embedding_flops) * 3 / 10**12
     )
     attention_tflops = causal_attention_flops * config.num_decoder_layers * 3 / 10**12
+  elif config.decoder_block == DecoderBlockType.QWEN3_NEXT:
+    gdn_weight_flops_per_layer, gdn_attn_flops_per_layer = calculate_gated_delta_net_flops_per_device(config)
+    cycle_interval = config.inhomogeneous_layer_cycle_interval
+    num_full_attn_layers = config.num_decoder_layers // cycle_interval
+    num_linear_attn_layers = config.num_decoder_layers - num_full_attn_layers
+
+    # Weights TFLOPs:
+    total_weights = (
+        total_ffn_flops
+        + embedding_flops
+        + (qkv_flops + projection_flops) * num_full_attn_layers
+        + gdn_weight_flops_per_layer * num_linear_attn_layers
+    )
+    learnable_weight_tflops = total_weights * 3 / 10**12
+
+    # Attention TFLOPs:
+    total_attn = (causal_attention_flops * num_full_attn_layers) + (gdn_attn_flops_per_layer * num_linear_attn_layers)
+    attention_tflops = total_attn * 3 / 10**12
   else:
     # multiply by 3 for both feed forward and back propagation flops
     learnable_weight_tflops = (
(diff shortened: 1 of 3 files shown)


src/MaxText/maxtext_utils.py
```diff
@@ -389,6 +389,77 @@ def get_dense_moe_layers(config):
   return num_dense_layers, num_moe_layers
 
 
+def calculate_qwen3_next_ffn_tflops_per_device(config):
```
RissyRan: Do you think we could directly use calculate_routed_and_shared_ffn_tflops_per_device()? the dense layer=0 in your case, similar to Llama4.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -389,6 +389,77 @@ def get_dense_moe_layers(config):
...
+
+  # - Inter-chunk terms (Recurrent State D_k * D_v):
+  #   - attn_i (Q*K): 2 * B * S * H_v * C * D_k
+  #   - v_prime (K*S): 2 * B * S * H_v * D_k * D_v
+  #   - attn_inter (Q*S): 2 * B * S * H_v * D_k * D_v
+  #   - core_out (A*V): 2 * B * S * H_v * C * D_v
+  #   - update (K*V): 2 * B * S * H_v * D_k * D_v
+  flops_inter = (2 * B * S * H_v * C * (D_k + D_v)) + (6 * B * S * H_v * D_k * D_v)
+
+  flops_core = flops_intra + flops_inter
```
RissyRan, line 557: Do you think we should apply causal mask similar [here](https://github.com/AI-Hypercomputer/maxtext/blob/4bcee99601ae152ca8a9b642f1d597b17d769547/src/MaxText/maxtext_utils.py#L566-L570) or anywhere inside of `config.decoder_block == DecoderBlockType.QWEN3_NEXT`?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -389,6 +389,77 @@ def get_dense_moe_layers(config):
...
+  flops_out = 2 * B * S * V_dim * E
+
+  flops_projections = flops_qkvz + flops_ba + flops_out
+
+  # 2. Convolution (Learnable Weights)
+  # Depthwise conv on dim (2*K_dim + V_dim)
+  # 2 * B * S * Channels * Kernel
+  flops_conv = 2 * B * S * (2 * K_dim + V_dim) * K_conv
+
+  # 3. Core Gated Delta Net (Attention-like operations)
```
RissyRan, line 531: Thanks for the explanation! Could you help specify the shape of K, A, V, Q?

RissyRan: Also take a look at @parambole 's roof line analysis, it seems this part is dis-aligned?

https://screenshot.googleplex.com/BXZiCaUngeTe5rT

src/MaxText/maxtext_utils.py
Rohan-Bierneni: Sure, I changed the code to use calculate_routed_and_shared_ffn_tflops_per_device and removed calculate_qwen3_next_ffn_tflops_per_device.

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: I don't think we should apply the causal mask factor here. Unlike standard attention where we conceptually calculate an 2-D matrix and mask half of it, here we are performing a fixed-size state update sequentially for every chunk. Since we execute this operation fully for every step to maintain the recurrent state, these FLOPs are 'dense' and shouldn't be halved. 

However, note that in calculate_tflops_training_per_device, we do apply the causal mask reduction to the Full Attention layers

Rohan-Bierneni:
> Thanks for the update!
> 
> We have some unit [tests](tests/flop_calculation_test.py) for those tflops, if you could come up with a test, that would be great, but not mandatory for this PR. Also, could you also have some sanity check with other MoE model (no change impact)? Similar to #2979

I have added a unit test in flop_calculation_test.py. Also will run the sanity checks and update the pr description with results.

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: Sure, the shapes are as follow:

Query (Q): [Batch, Seq_Len, H_v, D_k]
Keys (K): [Batch, Seq_Len, H_v, D_k]
Values (V): [Batch, Seq_Len, H_v, D_v]
Intra-Chunk Attention (A): [Batch, Num_Chunks, H_v, Chunk_Size, Chunk_Size]
Recurrent State (S): [Batch, Num_Chunks, H_v, D_k, D_v]

Rohan-Bierneni: > Also take a look at @parambole 's roof line analysis, it seems this part is dis-aligned?
> 
> https://screenshot.googleplex.com/BXZiCaUngeTe5rT

The roofline analysis accounts for the State Expansion & State Retrieval but not for the State Reconstruction step as part of the delta rule: https://github.com/AI-Hypercomputer/maxtext/blob/c63692493712e20c7b1a4b0c945115e1737177c6/src/MaxText/layers/qwen3.py#L234

This would only increase the tflops in the roofline analysis for core linear attention compute from 0.004 to 0.006 which is a pretty negligible amount when compared to tflops for the projections at 0.206 tflops.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -380,15 +380,74 @@ def get_dense_moe_layers(config):
...
+    num_moe_layers = config.num_decoder_layers // interleave_step
     num_dense_layers = config.num_decoder_layers - num_moe_layers
   else:
-    raise ValueError("Currently we only support DeepSeek and Llama4 calculation.")
+    raise ValueError("Currently we only support DeepSeek, Llama4, ad Qwen3-Next calculation.")
 
   return num_dense_layers, num_moe_layers
 
 
+def calculate_gated_delta_net_flops_per_device(config):
```
parambole, line 500: The current `flops_intra` calculation:

`flops_intra = 2 * B * S * H_v * C * (2 * D_k + D_v)`

This correctly captures the FLOPs from the three main `jnp.matmul` operations using the *final* `attn` matrix:
1.  Initial `attn` matrix: `jnp.matmul(k_beta_c, key_c^T)`
2.  `value_intra = jnp.matmul(attn, v_beta_c)`
3.  `k_cumdecay = jnp.matmul(attn, k_beta_c * ...)`

However, this calculation is **missing the significant computational cost of the `inner_attn_body` `fori_loop`** within the `jax_chunk_gated_delta_rule` function. This loop iteratively refines the `attn` matrix *before* the `value_intra` and `k_cumdecay` matrix multiplications.

The dominant operation within each of the `C-1` iterations of the loop is `jnp.sum(term, axis=-2)`, where `term` is `(B, H, N, C, C)`. The FLOPs for this loop are approximately:

`F_inner_loop ≈ (C - 1) * B * H * N * C^2 ≈ B * H * S * C^2`

This cost should be added to `flops_intra`. The more accurate `flops_intra` would be:

`flops_intra = 2 * B * S * H_v * C * (2 * D_k + D_v) + B * H_v * S * C^2`

Rohan-Bierneni: Thank you for catching this! I have updated the flops calculation to the right flops_intra formula

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: The core attention flops should be correct now, and I will update the roofline analysis as well to align on both tflops calculations.

src/MaxText/maxtext_utils.py
```diff
@@ -380,15 +380,75 @@ def get_dense_moe_layers(config):
     num_dense_layers = config.first_num_dense_layers
     num_moe_layers = config.num_decoder_layers - config.first_num_dense_layers
     return num_dense_layers, num_moe_layers
-  elif config.decoder_block == DecoderBlockType.LLAMA4:
-    num_moe_layers = config.num_decoder_layers // config.interleave_moe_layer_step
+  elif config.decoder_block in (DecoderBlockType.LLAMA4, DecoderBlockType.QWEN3_NEXT):
```
RissyRan: Could you help split those 2 blocks, add something like 

```
elif config.decoder_block == DecoderBlockType.QWEN3_NEXT:
  num_moe_layers = config.num_decoder_layers # all moe layers
  num_dense_layers = 0
```

src/MaxText/maxtext_utils.py
```diff
@@ -380,15 +380,75 @@ def get_dense_moe_layers(config):
...
+  flops_projections = flops_qkvz + flops_ba + flops_out
+
+  # 2. Convolution (Learnable Weights)
+  # Depthwise conv on dim (2*K_dim + V_dim)
+  # 2 * B * S * Channels * Kernel
+  flops_conv = 2 * B * S * (2 * K_dim + V_dim) * K_conv
+
+  # 3. Core Gated Delta Net (Attention-like operations)
+  # Assumptions:
+  # - H = H_v (broadcasting K to V heads if H_v > H_k)
```
RissyRan: This is really helpful comment for readers, could you help add here?

```
Query (Q): [Batch, Seq_Len, H_v, D_k]
Keys (K): [Batch, Seq_Len, H_v, D_k]
Values (V): [Batch, Seq_Len, H_v, D_v]
Intra-Chunk Attention (A): [Batch, Num_Chunks, H_v, Chunk_Size, Chunk_Size]
Recurrent State (S): [Batch, Num_Chunks, H_v, D_k, D_v]
```

src/maxtext/utils/maxtext_utils.py
```diff
@@ -380,15 +380,75 @@ def get_dense_moe_layers(config):
...
+
+  # 3. Core Gated Delta Net (Attention-like operations)
+  # Assumptions:
+  # - H = H_v (broadcasting K to V heads if H_v > H_k)
+  # - Intra-chunk terms (per chunk C):
+  #   - attn (K*K): 2 * B * S * H_v * C * D_k
+  #   - val_intra (A*V): 2 * B * S * H_v * C * D_v
+  #   - k_cum (A*K): 2 * B * S * H_v * C * D_k
+  #   - inner_attn_body loop (iterative refinement): ≈ (C - 1) * B * H * N * C^2 ≈ B * H * S * C^2
+  flops_intra = 2 * B * S * H_v * C * (2 * D_k + D_v) + (B * H_v * S * C**2)
```
RissyRan, line 547: Sorry I didn't follow here when comparing with Param's doc. I understand the values are similar, but could we align the formula as well? I sent a quick sync-up to move it faster. Thank you!

```
     State Expansion (K^T x V): For every token, we compute the outer product.
       Math: [16, 4096, 128, 1] @ [16, 4096, 1, 128]
       FLOPs: 2 x 16 x 4096 x 128 x 128 / 10^12 = 0.002 TFLOPS
     State Retrieval (Q x S): For every token, we query the state.
       Math: [16, 4096, 1, 128] @ [16, 4096, 128, 128]
       FLOPs: 2 x 16 x 4096 x 128 x 128 / 10^12 = 0.002 TFLOPS
     Total: 0.004 TFLOPS

```

RissyRan (commented):
Thanks for the change! I sent a quick sync-up, hopefully this could iterate faster.

src/MaxText/maxtext_utils.py
Rohan-Bierneni: Resolved.

src/MaxText/maxtext_utils.py
Rohan-Bierneni: I have added these to the comments for easier reading.

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: Based on conversation from internal sync, we will continue with the formulas in this PR, as @parambole's doc highlights the main compute being done, while the formulas in this PR is more specific/accurate to the GatedDeltaNet implementation. However, note that the delta between the 2 Tflop calculations differ only by 0.1% since input projections dominate the overall Tflops.

src/MaxText/maxtext_utils.py
```diff
@@ -383,12 +383,82 @@ def get_dense_moe_layers(config):
   elif config.decoder_block == DecoderBlockType.LLAMA4:
     num_moe_layers = config.num_decoder_layers // config.interleave_moe_layer_step
     num_dense_layers = config.num_decoder_layers - num_moe_layers
+  elif config.decoder_block == DecoderBlockType.QWEN3_NEXT:
+    num_moe_layers = config.num_decoder_layers
+    num_dense_layers = 0
   else:
-    raise ValueError("Currently we only support DeepSeek and Llama4 calculation.")
+    raise ValueError("Currently we only support DeepSeek, Llama4, ad Qwen3-Next calculation.")
```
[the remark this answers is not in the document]
Rohan-Bierneni: Resolved.

Rohan-Bierneni:
Manually adding pull_ready tag since all tests passed and 2 approvals recieved

PR #3010, opened 2026-01-26
Add DeepSeek Engram layer

shuningjin:
# Description

## Background
- Paper: https://arxiv.org/pdf/2601.07372
- Reference Code: https://github.com/deepseek-ai/Engram/blob/main/engram_demo_v1.py

## What this PR does

Add Engram layer: `engram.py`
- **`NgramHashMapping`** (non-parametric): `CompressedTokenizer` + **hashing logic**, convert "input_id" to "ngram hash_token_id"
  - **`CompressedTokenizer`** (non-parametric): convert "input_id" to "compresed_input_id"
- **`Engram`** (multi-branch): inputs are "ngram hash_token_id" and "transformer state", `MultiHeadEmbedding` (lookup embedding using hash id as static memory) + **context-aware gating** (dot product static memory with contextual state) + `ShortConv` (temporal smoothing)
  - **`MultiHeadEmbedding`**: convert ngram hash_token_id to ngram embedding vector
  - **`ShortConv`** (multi-branch): depthwise (mix time steps, not mix channel), causal, short means kernel size is small

Add unit test: `tests.unit.engram_vs_reference_test`
- for each component, verify the output matches that from reference code

## Implementation Notes

Placement of: `NgramHashMapping`
- `NgramHashMapping` converts vanilla token-ids to hashed ngram token-ids, which  `Engram` consumes for embedding lookup 
- Future: I would like to `NgramHashMapping` and hash_input_ids generation be put in data input pipeline, which is CPU intensive. Just like how we put tokenizer and input_ids generation in pipeline.
  - This is the code skeleton: https://screenshot.googleplex.com/7YYxr4z7UqvBkpN
  - Also: b/478294696#comment5

Multi-branch 
- `Engram` and `ShortConv` handles multi-branch input and multi-branch output (if `mhc_expansion_rate > 1`), using nnx.vmap for independent norm per branch
-  Future: to be integrated into multi-branch backbone like mHC.

# Tests

unit test against reference
```
python3 -m pytest -v --pyargs tests.unit.engram_vs_reference_test -rP -s
```
log: https://paste.googleplex.com/5905570101067776

# Checklist

### DIFF
--- a/src/MaxText/layers/engram.py
+++ b/src/MaxText/layers/engram.py
@@ -1,0 +1,685 @@
+# Copyright 2023–2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+DeepSeek-AI, `Conditional Memory via Scalable Lookup: A New Axis of Sparsity for Large Language Models
+  <https://arxiv.org/pdf/2601.07372>`_, 2026
+  
+Reference implementation: https://github.com/deepseek-ai/Engram/blob/main/engram_demo_v1.py
+"""
+
+from typing import List, Optional
+import numpy as np
+from sympy import isprime
+from tokenizers import normalizers, Regex
+
+import jax
+import jax.numpy as jnp
+from jax.sharding import Mesh
+from flax import nnx
+
+from MaxText.tokenizer import HFTokenizer
+from MaxText.common_types import MODEL_MODE_TRAIN, Array, Config
+from MaxText.layers.embeddings import Embed
+from MaxText.layers.initializers import nd_dense_init, NdInitializer
+from MaxText.layers.linears import DenseGeneral
+from MaxText.layers.normalizations import RMSNorm
+from MaxText.layers.quantizations import AqtQuantization as Quant
+
+
+class CompressedTokenizer:
+  """
+  A canonicalizing wrapper that reduces vocabulary sparsity for n-gram lookup.
+
+  This class maps semantically equivalent tokens (e.g., "Apple", " apple", "APPLE")
+  to a single unified ID. This many-to-one mapping significantly reduces the
+  combinatorial size of the n-gram space.
+
+  Attributes:
+    lookup_table: Array mapping `original_id` -> `compressed_id`.
+    num_new_token: Size of the compressed vocabulary.
+  """
+
+  def __init__(self, tokenizer: HFTokenizer):
+    normalizer = self._build_normalizer()
+    self.lookup_table, self.num_new_token = self._build_lookup_table(tokenizer, normalizer)
+
+  def __len__(self) -> int:
+    return self.num_new_token
+
+  def _build_normalizer(self) -> normalizers.Sequence:
+    """
+    Builds the normalization pipeline for text processing.
+    """
+    # Private use Unicode character to protect single spaces during stripping
+    SENTINEL = "\uE000"
+
+    # Normalization pipeline: ensures variations like "Café" and "cafe" map to the same ID
+    normalizer = normalizers.Sequence(
+        [
+            # Compatibility decomposition (e.g., ½ -> 1/2)
+            normalizers.NFKC(),
+            # Canonical decomposition (e.g., é -> e + ´)
+            normalizers.NFD(),
+            # Strip diacritics (e.g., e + ´ -> e)
+            normalizers.StripAccents(),
+            # Lowercase conversion ("The" -> "the")
+            normalizers.Lowercase(),
+            # Collapse all whitespace variations to a single space
+            normalizers.Replace(Regex(r"[ \t\r\n]+"), " "),
+            # Protect standalone spaces from subsequent stripping
+            normalizers.Replace(Regex(r"^ $"), SENTINEL),
+            # Remove leading/trailing whitespace
+            normalizers.Strip(),
+            # Restore protected spaces
+            normalizers.Replace(SENTINEL, " "),
+        ]
+    )
+    return normalizer
+
+  def _build_lookup_table(self, tokenizer: HFTokenizer, normalizer: normalizers.Sequence) -> tuple[np.ndarray, int]:
+    """
+    Builds the mapping from the original vocabulary to the compressed vocabulary.
+    """
+    vocab_size = len(tokenizer)
+    # Mapping: original_tid -> compressed_nid (Many-to-One)
+    old2new = np.empty(vocab_size, dtype=np.int64)
+    # Mapping: normalized_string -> compressed_nid (One-to-One)
+    key2new = {}
+
+    # Batch decode token to raw text
+    texts = tokenizer.batch_decode([[tid] for tid in range(vocab_size)], skip_special_tokens=False)
+
+    for tid, text in zip(range(vocab_size), texts):
+      if "\ufffd" in text:
+        # Handle invalid UTF-8 (replacement char �). Use raw token instead.
+        key = tokenizer.convert_ids_to_tokens(tid)
+      else:
+        # Normalize text (e.g., "  APPLE" -> "apple")
+        normalized_text = normalizer.normalize_str(text)
+        # Fallback to raw text if normalization creates an empty string
+        key = normalized_text if normalized_text else text
+
+      # Assign compressed ID
+      nid = key2new.get(key)
+      if nid is None:
+        nid = len(key2new)
+        key2new[key] = nid
+
+      old2new[tid] = nid
+
+    return old2new, len(key2new)
+
+  def __call__(self, input_ids) -> np.ndarray:
+    """
+    Maps original token IDs to compressed IDs.
+    """
+    input_ids = np.asarray(input_ids, dtype=np.int64)
+
+    # Identify valid tokens (ignore padding/masks usually marked with negative IDs)
+    valid_mask = input_ids >= 0
+    valid_ids = input_ids[valid_mask]
+
+    # Vectorized lookup: O(1) per token
+    output_ids = input_ids.copy()
+    output_ids[valid_mask] = self.lookup_table[valid_ids]
+    return output_ids
+
+
+class NgramHashMapping:
+  """
+  Deterministically maps token indices to n-gram hash indices for embedding lookups.
+
+  This class implements Multi-Head Hashing to bypass the combinatorial memory requirements
+  of explicit n-gram vocabularies. Specifically, it applies multiplicative-XOR hashing
+  to each n-gram window.
+
+  Key Mechanisms for Collision Mitigation:
+  - Multi-Head Factorization: Uses K distinct hash heads per n-gram order to increase
+    effective capacity within fixed memory constraints.
+  - Unique Prime Moduli: Assigns a unique prime vocabulary size to each head to
+    minimize simultaneous collisions.
+  """
+
+  def __init__(
+      self,
+      engram_vocab_bases: List[int],
+      max_ngram_size: int,
+      engram_num_heads: int,
+      layer_ids: List[int],
+      tokenizer: HFTokenizer,
+      pad_id: int,
+      seed: int,
+  ):
+    """
+    Args:
+      engram_vocab_bases: List of minimum head vocab sizes for each n-gram order.
+      max_ngram_size: Max n-gram size to track (e.g., 3 tracks 2-grams and 3-grams).
+      engram_num_heads: Number of parallel heads per n-gram order.
+      layer_ids: List of layer indices using Engram.
+      tokenizer: Base Hugging Face tokenizer.
+      pad_id: Padding token ID.
+      seed: Random seed for hash multiplier generation.
+    """
+    self.min_head_vocab_size_per_ngram = engram_vocab_bases
+    self.max_ngram_size = max_ngram_size
+    self.n_head_per_ngram = engram_num_heads
+    self.layer_ids = layer_ids
+
+    # Initialize compressed tokenizer
+    self.compressed_tokenizer = CompressedTokenizer(tokenizer)
+    self.tokenizer_vocab_size = len(self.compressed_tokenizer)
+    if pad_id is not None:
+      self.pad_id = int(self.compressed_tokenizer.lookup_table[pad_id])
+
+    # Pre-calculate odd multipliers for hashing: {layer_id: multipliers}
+    self.layer_multipliers = self._calculate_multipliers_across_layers(seed)
+
+    # Pre-calculate unique prime vocab sizes for every head
+    # Structure: {layer_id: [[2gram_head1, ..., 2gram_headH], ..., [Ngram_head1, ..., Ngram_headH]]}
+    self.vocab_size_across_layers = self._calculate_vocab_size_across_layers()
+
+  def _calculate_multipliers_across_layers(self, seed: int) -> dict[int, np.ndarray]:
+    """
+    Pre-calculates random odd multipliers for each layer and n-gram position.
+
+    Returns:
+      A dictionary mapping layer_id to a list of `max_ngram_size` multipliers.
+    """
+    # Pre-calculate bounds for random generation
+    max_long = np.iinfo(np.int64).max
+    m_max = int(max_long // self.tokenizer_vocab_size)
+    half_bound = max(1, m_max // 2)
+    # Hard-code prime number to align with reference
+    LAYER_PRIME_OFFSET = 10007
+
+    layer_multipliers = {}
+    for layer_id in self.layer_ids:
+      # Offset seed to decorrelate layers
+      layer_seed = int(seed + LAYER_PRIME_OFFSET * int(layer_id))
+      np_rng = np.random.default_rng(layer_seed)
+      # Generate random odd integers
+      random_value = np_rng.integers(low=0, high=half_bound, size=(self.max_ngram_size,), dtype=np.int64)
+      multipliers = random_value * 2 + 1
+      layer_multipliers[layer_id] = multipliers
+    return layer_multipliers
+
+  def _calculate_vocab_size_across_layers(self) -> dict[int, List[List[int]]]:
+    """
+    Calculates unique prime vocabulary sizes for every head in every layer.
+    Using unique primes minimizes the probability of simultaneous collisions across heads.
+    """
+
+    def find_next_unseen_prime(start: int, seen_primes: set) -> int:
+      candidate = start + 1
+      while candidate in seen_primes or not isprime(candidate):
+        candidate += 1
+      return candidate
+
+    seen_primes = set()
+    vocab_size_across_layers = {}
+
+    for layer_id in self.layer_ids:
+      all_ngram_vocab_sizes = []
+
+      for n in range(2, self.max_ngram_size + 1):
+        current_ngram_heads_sizes = []
+        # Start search from the configured minimum size
+        vocab_size = self.min_head_vocab_size_per_ngram[n - 2]
+        current_prime_search_start = vocab_size - 1
+        # Find unique primes for each head
+        num_heads = self.n_head_per_ngram
+        for _ in range(num_heads):
+          found_prime = find_next_unseen_prime(current_prime_search_start, seen_primes)
+          seen_primes.add(found_prime)
+          current_ngram_heads_sizes.append(found_prime)
+          current_prime_search_start = found_prime
+        all_ngram_vocab_sizes.append(current_ngram_heads_sizes)
+
+      vocab_size_across_layers[layer_id] = all_ngram_vocab_sizes
+
+    return vocab_size_across_layers
+
+  def get_vocab_sizes(self, layer_id: int) -> List[int]:
+    """
+    Returns a flattened list of prime vocabulary sizes for a specific layer.
+    """
+    return [head_size for ngram_size in self.vocab_size_across_layers[layer_id] for head_size in ngram_size]
+
+  def _get_ngram_hashes(self, compressed_ids: np.ndarray, layer_id: int) -> np.ndarray:
+    """
+    Computes hash indices for all n-grams in the input batch.
+
+    Args:
+      compressed_ids: [B, S] input token IDs.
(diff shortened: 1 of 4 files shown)


shuningjin:
@gemini-cli /review

RissyRan (commented):
@gemini-cli /review

src/MaxText/pyconfig.py
```diff
@@ -33,7 +33,7 @@
 from maxtext.utils import max_utils
 
 logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
RissyRan: +1

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+DeepSeek-AI, `Conditional Memory via Scalable Lookup: A New Axis of Sparsity for Large Language Models
+  <https://arxiv.org/pdf/2601.07372>`_, 2026
```
RissyRan, line 17: is this extra "_" on purpose?

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+  combinatorial size of the n-gram space.
+
+  Attributes:
+    tokenizer: Base Hugging Face tokenizer.
+    lookup_table: Array mapping `original_id` -> `compressed_id`.
+    num_new_token: Size of the compressed vocabulary.
+  """
+
+  def __init__(self, tokenizer: HFTokenizer):
+    # TODO(shuningjin): maybe don't need to hold tokenizer, if we only use the lookup table as bridge
```
RissyRan: What's the consequence if we remove it here?

tests/unit/engram_vs_reference_test.py
```diff
@@ -0,0 +1,783 @@
...
+  # Engram
+  engram_max_ngram_size: int = 3  # max_ngram_size, use 2...N
+  # List of minimum head vocab sizes for each n-gram order
+  engram_vocab_bases: List[int] = field(default_factory=lambda: [129280 * 5, 129280 * 5])
+  engram_layer_ids: List[int] = field(default_factory=lambda: [1, 15])
+  engram_kernel_size: int = 4  # conv kernel size
+  engram_head_dim: int = 32
+  engram_num_heads: int = 8  # num heads per n-gram
+  # Hashing
+  engram_pad_id: int = 2  # TODO(shuningjin): not the same as tokenizer.pad_id?
```
RissyRan: Does this need to be defined by users?

RissyRan (commented):
I reviewed the test and CompressedTokenizer. Will continue to review the rest part tomorrow.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+
+class NgramHashMapping:
+  """
+  Maps n-gram sequences to hash-based indices for memory lookup.
+
+  This class implements the Engram hashing mechanism. It converts variable-length
+  n-grams into fixed integer IDs. To handle the large combinatorial space, it uses:
+  1. Unique Prime Vocabularies: Per-head prime moduli to minimize collision overlap.
+  2. Sliding Window: Efficient shifting to generate n-gram views.
+  3. Lightweight Hashing: A multiplicative-XOR function (Rabin-Karp variant).
```
RissyRan: Nice! I may miss this part in reference implementation. Did you add this optimization?

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+
+  def _calculate_multipliers_across_layers(self, seed: int):
+    """
+    Pre-calculates random odd multipliers for each layer and n-gram position.
+
+    Returns:
+      A dictionary mapping layer_id to a list of `max_ngram_size` multipliers.
+    """
+    # Pre-calculate bounds for random generation
+    max_long = np.iinfo(np.int64).max
```
RissyRan, line 201: Could you cross check if we could update all np to jnp?

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+    m_max = int(max_long // self.tokenizer_vocab_size)
+    half_bound = max(1, m_max // 2)
+    LAYER_PRIME_OFFSET = 10007
+
+    layer_multipliers = {}
+    for layer_id in self.layer_ids:
```
RissyRan, line 208: Do you think we could update this block using vectorized operation? dim will depends on len(layer_ids). It's fixed at compile time.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+
+  def __init__(
+      self,
+      rngs: nnx.Rngs,
+      config: Config,
+      mesh: Mesh,
+      quant: Optional[Quant] = None,
+      kernel_init: NdInitializer = nd_dense_init(1.0, "fan_in", "normal"),
+      *,
+      hc_mult: int = 4,
```
RissyRan: shall we put params with default value at the very end?

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+    @nnx.split_rngs(splits=hc_mult)
+    @nnx.vmap(in_axes=0, out_axes=0)
+    def create_key_projs(r):
+      return DenseGeneral(
+          in_features_shape=self.engram_dim,
+          out_features_shape=config.base_emb_dim,
+          axis=-1,
+          kernel_init=self.kernel_init,
+          # TODO(shuningjin): this needs to be actual logical axis? @reviewer
+          kernel_axes=("engram_dim", "embed"),
```
RissyRan: You could add the sharding constraint into base.yml. https://github.com/AI-Hypercomputer/maxtext/blob/352dd58bf1cb4f36cd414c561bf97cb9ade86dc8/src/MaxText/configs/base.yml#L402

I see it is smaller dim compared to emb, we could shard on tensor as a starting point. I see embed usually sharding on fsdp, sequence, context etc.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+      hash_input_ids: Hashed token IDs. Shape: (B, S, H_total).
+        Produced by `hash_mapping.hash(input_ids)[layer_id]`.
+
+    Returns:
+      Shape: (B, S, G, D)
+
+    Shape annotation:
+      B: Batch Size
+      S: Sequence Length
+      G: hc_mult, Number of Branches
```
RissyRan: Do you plan to separate this config or treat it same as mhc_expansion_rate?

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+          rngs=r,
+          use_bias=True,
+      )
+
+    self.key_projs = create_key_projs(rngs)
+
+    # Norms (vectorized)
+    # Independent weights per branch, Branched input
+    @nnx.split_rngs(splits=hc_mult)
+    @nnx.vmap(in_axes=0, out_axes=0)
```
RissyRan, line 568: Are you sharding on batch dimension? Why is that? Similar comment for other in_axes=0 vmap op.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+    # (B, S, H_total) -> (B, S, H_total, D_head)
+    embeddings = self.multi_head_embedding(hash_input_ids)
+    # (B, S, H_total, D_head) -> (B, S, D_en)
+    embeddings = embeddings.reshape(B, S, -1)
+
+    # 2. Static Memory as Key
+    # Vectorized broadcast: apply each of the G key_projs to the SAME embeddings.
+    # in_axes: (0, None) -> 0 splits the Dense layers, None broadcasts embeddings
+    # out_axes: 2        -> Stack the results at axis 2 to get (B, S, G, D)
+    @nnx.vmap(in_axes=(0, None), out_axes=2)
```
RissyRan: Do you know if this in_axes working properly? I see your unit test has b=2 setup. When I integrated flash attn with sparse attn, I have to change the unit test to from b=2 to b=4 when sharding on fsdp, otherwise, it will fail on v5p-8 local machine.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,679 @@
...
+  3. Lightweight Hashing: A multiplicative-XOR function (Rabin-Karp variant).
+  """
+
+  def __init__(
+      self,
+      engram_vocab_bases: List[int],
+      max_ngram_size: int,
+      engram_num_heads: int,
+      layer_ids: List[int],
+      tokenizer: HFTokenizer,
```
RissyRan, line 161: When you saying would like to put the look up table into data pipeline. Is this structure or performance beneficial? When we call the engram from decoder layer, we need to pass this tokenizer. So you are thinking, this engram module will call/depend on data pipeline to look up?

tests/unit/engram_vs_reference_test.py
```diff
@@ -0,0 +1,783 @@
...
+class NgramHashMappingTest(parameterized.TestCase):
+
+  def test_hash_match(self):
+    np.random.seed(42)
+    self.config = Config()
+    self.engram_cfg = EngramConfig(self.config)
+    self.backbone_config = BackBoneConfig(self.config)
+    tokenizer = AutoTokenizer.from_pretrained(self.config.tokenizer_path, trust_remote_code=True)
+    # input
+    batch, seq_len = 2, 3
```
RissyRan: Could we set up a longer sequence, like 8, so test overlap of 2/3-grams?

RissyRan (commented):
Thanks for the change! I left some initial comments, and may need to go over multihead embedding and conv parts. It should be quick.

src/MaxText/layers/engram.py
shuningjin: Thanks for the suggestion! Added `['engram_dim', ['tensor']]`.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,682 @@
...
+      return DenseGeneral(
+          in_features_shape=self.engram_dim,
+          out_features_shape=config.base_emb_dim,
+          axis=-1,
+          kernel_init=self.kernel_init,
+          # TODO(shuningjin): this needs to be actual logical axis? @reviewer
+          kernel_axes=("engram_dim", "embed"),
+          dtype=self.dtype,
+          weight_dtype=self.weight_dtype,
+          quant=self.quant,
```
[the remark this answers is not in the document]
shuningjin: done

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,682 @@
...
+
+    # Key Normalization
+    self.k_norms = create_norms(rngs)
+    # Query Normalization
+    self.q_norms = create_norms(rngs)
+
+    # Value Projection (shared): Retrieved memory -> Value
+    self.value_proj = DenseGeneral(
+        in_features_shape=self.engram_dim,
+        out_features_shape=config.base_emb_dim,
```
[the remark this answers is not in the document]
shuningjin: done

src/MaxText/layers/engram.py
shuningjin: I would like to put `NgramHashMapping` and `hash_input_ids` generation in data input pipeline, which is CPU intensive. Just like how we put `tokenizer` and `input_ids` generation in pipeline.

This is the overall structure: https://screenshot.googleplex.com/7YYxr4z7UqvBkpN

Also: b/478294696#comment5

src/MaxText/layers/engram.py
shuningjin: If this is to be in data input pipeline, maybe should align with that. Does it use `np` or `jnp`?

tests/unit/engram_vs_reference_test.py
shuningjin: In the reference code, it is a hyper-parameter. I am not sure why.
- https://github.com/deepseek-ai/Engram/blob/fb7f84a21f91223715394a33a1dc24bbfb7f788e/engram_demo_v1.py#L46C5-L46C11
- https://github.com/deepseek-ai/Engram/blob/fb7f84a21f91223715394a33a1dc24bbfb7f788e/engram_demo_v1.py#L197-L212

Maybe leave it for now, and revisit during integration?

src/MaxText/layers/engram.py
shuningjin: removed

tests/unit/engram_vs_reference_test.py
shuningjin: Thanks for noticing that. I change all tests to `seq_len=32`. Tests still pass.

src/MaxText/layers/engram.py
shuningjin: I will reuse `mhc_expansion_rate`. Have replaced all `hc_mult`.

src/MaxText/layers/engram.py
shuningjin: Thanks for the suggestions. I have updated the argument order.

src/MaxText/layers/engram.py
shuningjin: They are vectorizing over branch (G). I have updated comments for all vmap op to clarify.

src/MaxText/layers/engram.py
shuningjin: Similar as above: They are vectorizing over branch. The unit test is working. I haven't added sharding constraints inside engram.

src/MaxText/pyconfig.py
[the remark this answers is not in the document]
shuningjin: I will restore this before merge.

src/MaxText/layers/engram.py
shuningjin: I have updated the docstring of NgramHashMapping to be more accurate.

src/MaxText/layers/engram.py
shuningjin: I am mimicking the [optax](https://screenshot.googleplex.com/9fRXhsVSPY9EmzR) / [sphinx](https://screenshot.googleplex.com/AKTYixdJWpWkv6n) style.

RissyRan: I see. Since we are using Read the Docs, I am totally fine to use it here.

src/MaxText/pyconfig.py
```diff
@@ -34,7 +34,7 @@
...
-logger.setLevel(os.environ.get("LOGLEVEL", "INFO"))
+# logger.setLevel(os.environ.get("LOGLEVEL", "INFO"))
```
gagika: is this change intentianal?

src/MaxText/layers/engram.py
RissyRan: Looking at the [paper](https://screenshot.googleplex.com/3AKExHQxPsJDP5K), it appears they shard across GPUs during training but offload to the host during inference. While GPU-to-CPU bandwidth is usually a bottleneck, the [specs](https://screenshot.googleplex.com/7F9kgxgpyXnBK6Q) suggest their platform has bandwidth comparable to TPUs. I suspect the decision to keep it on-device during training depends on the number of n-gram layers—essentially a trade-off between the memory cost of the tables and the communication overhead. 

In this case, I am fine to keep it inside of engram layer. Or only move the `NgramHashMapping` out (not the hash_input_ids based on layer index). Potentially there could be some internal optimization using scan.

What do you think?

src/MaxText/layers/engram.py
RissyRan: Sounds good. Don't forget to update those if only on TPUs.

src/MaxText/layers/engram.py
RissyRan: Is this functional or perf needed for over batch dimension?

src/MaxText/layers/engram.py
RissyRan: This is not a blocker to merge the PR.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,697 @@
...
+    """
+    vocab_size = len(tokenizer)
+    # Mapping: original_tid -> compressed_nid (Many-to-One)
+    old2new = np.empty(vocab_size, dtype=np.int64)
+    # Mapping: normalized_string -> compressed_nid (One-to-One)
+    key2new = {}
+
+    for tid in range(vocab_size):
+      # Decode token to raw text
+      text = tokenizer.decode([tid], skip_special_tokens=False)
```
gagika: is this very slow for a vocab size of 100-200k tokens? 
if so paralelizing can help.

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,697 @@
...
+        kernel_size=(kernel_size,),
+        feature_group_count=total_channels,
+        kernel_dilation=(dilation,),
+        padding="CAUSAL",
+        use_bias=False,
+        dtype=config.dtype,
+        param_dtype=config.weight_dtype,
+        precision=config.matmul_precision,
+        rngs=rngs,
+    )
```
gagika, line 447: I think this needs to be intialized with zeros as the paper mentions:

`the convolution parameters are initialized to zero to strictly preserve the identity mapping at the start of training.`

src/MaxText/layers/engram.py
shuningjin: I am not able to fully reproduce the random seed generation behavior (from reference) with vectorization.
https://github.com/AI-Hypercomputer/maxtext/blob/4e7559c2821cc53d8da8d3a884953da6e283e1a7/src/MaxText/layers/engram.py#L207-L212

src/MaxText/layers/engram.py
shuningjin: Discussed offline to clarify branch vs. batch.

src/MaxText/layers/engram.py
aireenmei: If we want to run with data pipeline on CPU, then use np. If on TPU, use jnp

src/MaxText/layers/engram.py
shuningjin: Discussed offline. Will keep the current code, and defer changes to integration.

src/MaxText/layers/engram.py
shuningjin: Thanks for pointing out! I updated to one-time batch_decode
```
texts = tokenizer.batch_decode([[tid] for tid in range(vocab_size)], skip_special_tokens=False)
```

src/MaxText/layers/engram.py
shuningjin: Nice finding! I added the initialization for conv with comments
```
        # convolution parameters are initialized to zero
        # to strictly preserve the identity mapping at the start of training
        kernel_init=nnx.initializers.zeros,
```

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,699 @@
...
+"""
+DeepSeek-AI, `Conditional Memory via Scalable Lookup: A New Axis of Sparsity for Large Language Models
+  <https://arxiv.org/pdf/2601.07372>`_, 2026
+  
+Reference implementation: https://github.com/deepseek-ai/Engram/blob/main/engram_demo_v1.py
+"""
+
+from typing import List, Optional
+import numpy as np
+from sympy import isprime
```
aireenmei, line 24: Looks like a new dependency. Could we use existing dependencies? Otherwise we need to update requirements.txt

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,699 @@
...
+    Pre-calculates random odd multipliers for each layer and n-gram position.
+
+    Returns:
+      A dictionary mapping layer_id to a list of `max_ngram_size` multipliers.
+    """
+    # Pre-calculate bounds for random generation
+    max_long = np.iinfo(np.int64).max
+    m_max = int(max_long // self.tokenizer_vocab_size)
+    half_bound = max(1, m_max // 2)
+    LAYER_PRIME_OFFSET = 10007
```
aireenmei, line 205: I think this come from the paper, let's add a comment to clarify the hard coding

src/MaxText/layers/engram.py
```diff
@@ -0,0 +1,699 @@
...
+        rngs=self.rngs,
+    )
+
+    # Short Convolution (Vectorized Internally)
+    # Applies depthwise causal convolution to smooth the retrieved memory over time.
+    self.short_conv = ShortConv(
+        config=config,
+        hidden_size=config.base_emb_dim,
+        kernel_size=self.conv_kernel_size,
+        dilation=self.max_ngram_size,
```
aireenmei, line 606: In this case, the paper uses max_ngram_size as dilation, but I wonder if it's better to pass dilation to the module to make it configurable?

src/MaxText/layers/engram.py
shuningjin: Thanks for noticing this! I just run a sanity check: sympy can be imported within current MaxText environment, without updating requirements.

Verification steps: create a new env and import
```
cd maxtext
VENV=~/maxtext_venv2
# pip install uv
uv venv --python 3.12 --seed $VENV
source $VENV/bin/activate
uv pip install maxtext --resolution=lowest

>>> from sympy import isprime
```

log: https://paste.googleplex.com/5039186747981824

src/MaxText/layers/engram.py
shuningjin: Thanks for clarifying! I will assume `NgramHashMapping` is running on CPU and keep `np`. 

Can change later If we see problems during integration.

src/MaxText/pyconfig.py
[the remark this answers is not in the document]
shuningjin: Restored.

src/MaxText/layers/engram.py
shuningjin: I have updated key_projection to be a single linear rather than vmap, by adding branch dim to out_features_shape
```
        in_features_shape=self.engram_dim,
        out_features_shape=(mhc_expansion_rate, config.base_emb_dim),
```

src/MaxText/pyconfig.py
shuningjin: I have reverted the change.

src/MaxText/layers/engram.py
shuningjin: Thanks for pointing out! I have update the comment.

src/MaxText/layers/engram.py
shuningjin: Good point on the flexibility. For now, I’d prefer to keep this consistent with the reference implementation to ensure parity. We can revisit this if flexibility is needed for some variant.

PR #3025, opened 2026-01-28
Trim unit test time

RissyRan:
# Description

Trim CPU & TPU unit tests time:
* In moe_test.py, update the test seq from default 2048 to 128
* In train compile test, update to a smaller topology if not specified in test name or restricted by sharding (v6e-256 --> v5p-128/64, etc)
* In train compile test, update to seq len to 1024, if not testing CP
* In train compile test, use DS v3-test version if not restricted to some configs ([diff](https://diff.googleplex.com/#key=iuH5LF23sEBM))
* Remove one duplicate test

# Tests

* Expect tests are green
* CPU tests, saved: [7-8 min](https://screenshot.googleplex.com/Bdfk3JZ6GN5n4Zw)
* TPU tests saved: [half time for MoE tests](https://screenshot.googleplex.com/95JxyCF6BgRVrkN)

# Checklist

### DIFF
--- a/tests/unit/moe_test.py
+++ b/tests/unit/moe_test.py
@@ -1,4 +1,4 @@
-# Copyright 2023–2025 Google LLC
+# Copyright 2023–2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -460,6 +460,7 @@ class RoutedMoeTest(unittest.TestCase):
         megablox=True,
         sparse_matmul=True,
         per_device_batch_size=1,
+        max_target_length=128,
     )
 
     rng = jax.random.PRNGKey(1234)
@@ -488,6 +489,7 @@ class RoutedMoeTest(unittest.TestCase):
         megablox=False,
         sparse_matmul=True,
         per_device_batch_size=1,
+        max_target_length=128,
     )
 
     rng = jax.random.PRNGKey(1234)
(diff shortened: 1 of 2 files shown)


tests/unit/train_compile_test.py
```diff
@@ -304,8 +304,8 @@ def test_custom_64x4_mesh(self):
             "compile_topology_num_slices=1",
             "ici_sequence_parallelism=4",
             "global_parameter_scale=32",
-            "per_device_batch_size=0.25",
-            "max_target_length=65536",
+            "per_device_batch_size=1",
```
NuojCheng: why change pdb to 1?

tests/unit/moe_test.py
NuojCheng: nit: we can update to 2023-2026 now!

tests/unit/train_compile_test.py
RissyRan: This is not CP test, so update the per_device_batch_size=1 to align with other tests.

tests/unit/moe_test.py
RissyRan: Sure, will update.

tests/unit/train_compile_test.py
NuojCheng: But this test using SP=4, which is similar to CP. Using pdb=0.25 helps check fractional batch size support using SP>1.

RissyRan: Good catch! Let me update it back!

RissyRan: However, I don't think test custom mesh needs long context with SP: test_custom_64x4_mesh

Let me update it back first.

PR #3040, opened 2026-01-29
Fix gradient accumulation in post training sft

ChingTsai:
# Description

FIXES: b/478823561

This PR resolves discrepancies in loss calculation and training step counts when running SFT with gradient accumulation enabled. Currently, `MaxText.sft.sft_trainer` (which uses the Tunix trainer) exhibits breaking behavior when gradient accumulation is turned on, diverging significantly from the native implementation in `MaxText.sft_trainer`. This change aligns the Tunix-based SFT logic to match the native behavior.

## Problem Statement

- Loss Disparity: When GA is enabled, MaxText-Tunix observed a massive loss scale disparity compared to the native implementation. This occurs because MaxText-Tunix uses the same loss_fn as the native implementation. The native logic explicitly skips dividing by total_weights inside the function (deferring normalization to a later stage [here](https://github.com/AI-Hypercomputer/maxtext/pull/3040/changes#diff-4956f1016541609a6ae02b50cc212b3ba30cffc67e40d929e85dffaa0ce82baeR174-R178)). Consequently, Tunix inherited this behavior and failed to normalize the loss, resulting in broken calculations and inflated values. 

pr_fix_vs_original_vs_native

- Step Count Mismatch: While MaxText-Native handles micro-batching internally by reshaping the full global batch, Tunix relies on the input pipeline to provide pre-sized micro-batches. Without this adjustment, Tunix was ingesting full global batches at every step, resulting in incorrect epoch calculations and causing the run to terminate prematurely compared to the native implementation.

## Changes
- Introduced a new configuration `use_tunix`
  -  Determine whether gradient accumulation (GA) is performed through the Tunix implementation.

- Modified the loss function
  - Aligned Tunix loss calculation with native behavior. these changes are backward-compatible and will not affect existing native GA implementations.

- Pre-sliced batches in data preprocessing
  - Ensures that when using Tunix GA, the chunk size is correctly calculated during the data preparation stage.

- Add Tunix SFT integration test for gradient accumulation loss verification [here](https://github.com/AI-Hypercomputer/maxtext/pull/3040/changes#diff-7173b581a58bb895b89d15af0c2474889d2f969215b014053b0d5f2bd475019fR159-R224)

# Tests
## End2End
```
python3 -m MaxText.sft.sft_trainer \
    src/MaxText/configs/sft.yml \
    run_name=$RUN_NAME \
    base_output_directory=..../qwen3-4b \
    model_name=qwen3-4b \
    load_parameters_path=..../qwen3-4b/0/items \
    tokenizer_path=Qwen/qwen3-4b \
    steps=$train_step \
    profiler=xplane \
    hf_path=arrow \
    dataset_type=hf \
    train_split=train \
    hf_train_files=..../data-00000-of-00001.arrow \
    hf_eval_files=..../data-00000-of-00001.arrow \
    per_device_batch_size=4 \
    gradient_accumulation_steps=4 \
    max_target_length=1024 \
    learning_rate=1.3e-5 \
    warmup_steps_fraction=0.05 \
    data_shuffle_seed=42 \
    gradient_clipping_threshold=1 \
    learning_rate_final_fraction=0 \
    weight_dtype=bfloat16
```
After applying the changes, the loss graphs of both versions are now almost identical.

graph_2026-01-29_15-52-15

## Integration Test
```
python -m unittest tests.integration.gradient_accumulation_test.GradientAccumulationTest.test_tunix_sft_grad_accumulate_same_loss
```

# Checklist

### DIFF
--- a/src/MaxText/configs/types.py
+++ b/src/MaxText/configs/types.py
@@ -1056,6 +1056,10 @@ class Optimizer(BaseModel):
   gradient_accumulation_steps: PositiveInt = Field(
       1, description="Number of steps to accumulate gradients before updating."
   )
+  use_tunix_gradient_accumulation: bool = Field(
+      False,
+      description="Whether to use the Tunix implementation for gradient accumulation.",
+  )
   gradient_clipping_threshold: NonNegativeFloat = Field(
       1.0, description="The threshold for gradient clipping. 0 disables clipping."
   )
--- a/src/MaxText/input_pipeline/_hf_data_processing.py
+++ b/src/MaxText/input_pipeline/_hf_data_processing.py
@@ -54,7 +54,13 @@ def vision_sft_preprocessing_pipeline(
   """pipeline for multimodal SFT with HF dataset"""
 
   assert len(text_columns) == 2, f"Need two text_columns for query and response, received {text_columns=}"
-  batch_size = global_batch_size // jax.process_count()
+  # Tunix GA requires per-micro-batch slicing at the data level,
+  # whereas Native GA processes the full batch and splits it internally.
+  if config.use_tunix_gradient_accumulation:
+    batch_size = global_batch_size // jax.process_count() // config.gradient_accumulation_steps
+  else:
+    batch_size = global_batch_size // jax.process_count()
+
   if config.enable_data_shuffling:
     dataset = dataset.shuffle(seed=config.data_shuffle_seed)
 
@@ -195,6 +201,8 @@ def preprocessing_pipeline(
     generate_padding_batch=False,
     use_dpo=None,
     use_sft=None,
+    use_tunix_gradient_accumulation=False,
+    num_microbatches=1,
     sft_train_on_completion_only=True,
     grain_worker_count=1,  # only support 0 or 1
     max_segments_per_seq=None,
@@ -202,6 +210,12 @@ def preprocessing_pipeline(
   """pipeline for preprocessing HF dataset"""
 
   assert global_batch_size % global_mesh.size == 0, "Batch size should be divisible by number of global devices."
+  # Tunix GA requires per-micro-batch slicing at the data level,
+  # whereas Native GA processes the full batch and splits it internally.
+  if use_tunix_gradient_accumulation:
+    batch_size = global_batch_size // jax.process_count() // num_microbatches
+  else:
+    batch_size = global_batch_size // jax.process_count()
 
   if shuffle:
     dataset = dataset.shuffle(seed=data_shuffle_seed)
@@ -303,7 +317,7 @@ def preprocessing_pipeline(
       max_segments = None
     operations.append(
         grain.experimental.PackAndBatchOperation(
-            batch_size=global_batch_size // jax.process_count(),
+            batch_size=batch_size,
             length_struct=length_struct,
             max_sequences_per_bin=max_segments,
         )
@@ -311,7 +325,7 @@ def preprocessing_pipeline(
     operations.append(_input_pipeline_utils.ReformatPacking(data_column_names))
   else:
     operations.append(_input_pipeline_utils.PadOrTrimToMaxLength(max_target_length, pad_id))
-    operations.append(grain.Batch(batch_size=global_batch_size // jax.process_count(), drop_remainder=drop_remainder))
+    operations.append(grain.Batch(batch_size=batch_size, drop_remainder=drop_remainder))
 
   if shift and not use_dpo:
     operations.append(_input_pipeline_utils.ShiftData(ignored_ids=[pad_id, tokenizer.bos_token_id], axis=1))
@@ -390,6 +404,8 @@ def make_hf_train_iterator(
         generate_padding_batch=config.generate_padding_batch_train,
         use_dpo=config.use_dpo,
         use_sft=config.use_sft,
+        use_tunix_gradient_accumulation=config.use_tunix_gradient_accumulation,
+        num_microbatches=config.gradient_accumulation_steps,
         sft_train_on_completion_only=config.sft_train_on_completion_only,
         chat_template_path=config.chat_template_path,
         max_segments_per_seq=config.max_segments_per_seq,
@@ -443,6 +459,7 @@ def make_hf_eval_iterator(
         generate_padding_batch=config.generate_padding_batch_eval,
         use_dpo=config.use_dpo,
         use_sft=config.use_sft,
+        num_microbatches=config.gradient_accumulation_steps,
         sft_train_on_completion_only=config.sft_train_on_completion_only,
         chat_template_path=config.chat_template_path,
         max_segments_per_seq=config.max_segments_per_seq,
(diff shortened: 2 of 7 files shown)


src/maxtext/trainers/post_train/sft/train_sft.py
```diff
@@ -194,7 +194,7 @@ def main(argv: Sequence[str]) -> None:
   pathwaysutils.initialize()
   os.environ["TF_CPP_MIN_LOG_LEVEL"] = "0"
 
-  mt_config = pyconfig.initialize(argv)
+  mt_config = pyconfig.initialize(argv, use_tunix_ga=True)
```
SurbhiJainUSC: Can `use_tunix_ga` be part of SFT configs: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/configs/sft.yml? It can be set to True by default. Also, we can rename it to `use_tunix`.

src/MaxText/sft_trainer.py
```diff
@@ -165,7 +165,7 @@ def main(argv: Sequence[str]) -> None:
     os.environ["LIBTPU_INIT_ARGS"] = (
         os.environ.get("LIBTPU_INIT_ARGS", "") + " --xla_tpu_spmd_rng_bit_generator_unsafe=true"
     )
-  config = pyconfig.initialize(argv)
+  config = pyconfig.initialize(argv, use_tunix=False)
```
ChingTsai: Add an override here to maintain backward compatibility with the native `sft_trainer`.

src/maxtext/trainers/post_train/sft/train_sft.py
ChingTsai: Fixed.

src/MaxText/configs/types.py
```diff
@@ -1055,6 +1055,10 @@ class Optimizer(BaseModel):
   gradient_accumulation_steps: PositiveInt = Field(
       1, description="Number of steps to accumulate gradients before updating."
   )
+  use_tunix: bool = Field(
+      False,
+      description="Whether to use the Tunix implementation for gradient accumulation.",
+  )
```
NuojCheng, line 1062: I think the name `use_tunix` is too broad. If we are only interested in cases for sft + tunix + GA, please use a more specific name, e.g. `use_tunix_gradient_accumulation`.

src/MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -54,7 +54,9 @@ def vision_sft_preprocessing_pipeline(
   """pipeline for multimodal SFT with HF dataset"""
 
   assert len(text_columns) == 2, f"Need two text_columns for query and response, received {text_columns=}"
-  batch_size = global_batch_size // jax.process_count()
+  # Tunix GA requires per-micro-batch slicing at the data level,
+  # whereas Native GA processes the full batch and splits it internally.
+  batch_size = global_batch_size // jax.process_count() // (config.gradient_accumulation_steps if config.use_tunix else 1)
```
NuojCheng: nit: I would perfer
```
if config.use_tunix:
   batch_size = ()
else:
  batch_size = ()
```

src/MaxText/input_pipeline/_hf_data_processing.py
```diff
@@ -195,13 +197,18 @@ def preprocessing_pipeline(
...
     sft_train_on_completion_only=True,
     grain_worker_count=1,  # only support 0 or 1
     max_segments_per_seq=None,
 ):
   """pipeline for preprocessing HF dataset"""
 
   assert global_batch_size % global_mesh.size == 0, "Batch size should be divisible by number of global devices."
+  # Tunix GA requires per-micro-batch slicing at the data level,
+  # whereas Native GA processes the full batch and splits it internally.
+  batch_size = global_batch_size // jax.process_count() // (num_microbatches if use_tunix else 1)
```
NuojCheng: same with comment above

src/MaxText/train.py
```diff
@@ -179,7 +179,7 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
   # Zero1+GA to reduce communication overhead.
   # EPS was used to avoid division by zero, but it's not needed when gradient
   # accumulation is enabled since there's no division.
-  if config.gradient_accumulation_steps > 1:
+  if config.gradient_accumulation_steps > 1 and not config.use_tunix:
```
NuojCheng: please add comments explaining why we don't follow standard GA path when `use_tunix=true`

NuojCheng (commented):
It is a bit concerning to me why this PR fails tests/integration/train_tests.py::TrainTests::test_tpu_zero1_gradient_accumulation with OOM. The changes made by this PR should not affect regular GA performance in `train.py`. We will need to take a look on xprofs to figure out why.

src/MaxText/input_pipeline/_hf_data_processing.py
richjames0: Agree. I love expressions over statements but this is a bit complex for the former

ChingTsai:
> It is a bit concerning to me why this PR fails tests/integration/train_tests.py::TrainTests::test_tpu_zero1_gradient_accumulation with OOM. The changes made by this PR should not affect regular GA performance in `train.py`. We will need to take a look on xprofs to figure out why.

Hi @NuojCheng and @richjames0 ,

Thanks for the review! I have addressed your comments in the latest push.

Regarding the failing `tests/integration/train_tests.py::TrainTests::test_tpu_zero1_gradient_accumulation` OOM test. Is it possible the shared runner is competing for resources with other jobs? Or could some JAX arrays be leaking (not being garbage collected) during continuous testing? I’ve seen this test pass intermittently, for example in [here](https://github.com/AI-Hypercomputer/maxtext/actions/runs/21623645517/job/62318366296#step:7:57).
I'm happy to investigate further if you think it's related to the current changes.

ChingTsai:
hi @NuojCheng, could you help to merge this PR? Thanks!
cc @chishuen

ChingTsai:
Hi @NuojCheng ,

Although all integration tests are passed, I noticed that the TPU integration test is now taking significantly longer (> 70 min).
After investigating, It seems like the Tunix test might be leaving the testing process in a state that slows down subsequent tests. If I move that specific test to the end of the suite, the runtime returns to normal, as seen in [this run](https://github.com/AI-Hypercomputer/maxtext/actions/runs/21663656103/job/62453809768#step:7:90).

Should I pull this test from this PR and reintroduce it in a separate PR once the root cause is fixed?

NuojCheng:
> Hi @NuojCheng ,
> 
> Although all integration tests are passed, I noticed that the TPU integration test is now taking significantly longer (> 70 min). After investigating, It seems like the Tunix test might be leaving the testing process in a state that slows down subsequent tests. If I move that specific test to the end of the suite, the runtime returns to normal, as seen in [this run](https://github.com/AI-Hypercomputer/maxtext/actions/runs/21663656103/job/62453809768#step:7:90).
> 
> Should I pull this test from this PR and reintroduce it in a separate PR once the root cause is fixed?

Thank you for the investigation! Please create a separate PR.

I noticed the [pylint test fails](https://github.com/AI-Hypercomputer/maxtext/actions/runs/21631296934/job/62344455697?pr=3040). Please use `pre-commit` to fix it. Check `.pre-commit-config.yaml` for details.

PR #3067, opened 2026-02-02
Allow specifying trainer tensor parallelism in multislice RL

xuefgu:
# Description

Allow specifying trainer tensor parallelism in multislice RL

FIXES: b/480979614

# Tests

Manually tested.

# Checklist

### DIFF
--- a/src/MaxText/rl/train_rl.py
+++ b/src/MaxText/rl/train_rl.py
@@ -191,15 +191,32 @@ def setup_configs_and_devices(argv: list[str]):
     for i in range(config.num_trainer_slices, config.num_trainer_slices + config.num_samplers_slices):
       sampler_devices.extend(devices_by_slice[slice_indices[i]])
 
+    trainer_devices_per_slice = len(trainer_devices) // config.num_trainer_slices
+    trainer_fsdp = trainer_devices_per_slice
+    tp = config.ici_tensor_parallelism
+    if tp > 1:
+      if trainer_devices_per_slice % tp != 0:
+        raise ValueError(
+            f"trainer_devices_per_slice ({trainer_devices_per_slice}) must be divisible by tensor parallelism ({tp})"
+        )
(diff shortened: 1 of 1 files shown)


src/MaxText/rl/train_rl.py
```diff
@@ -191,15 +191,32 @@ def setup_configs_and_devices(argv: list[str]):
     for i in range(config.num_trainer_slices, config.num_trainer_slices + config.num_samplers_slices):
       sampler_devices.extend(devices_by_slice[slice_indices[i]])
 
+    trainer_devices_per_slice = len(trainer_devices) // config.num_trainer_slices
+    trainer_fsdp = trainer_devices_per_slice
+    tp = config.ici_tensor_parallelism
+    if tp > 1:
```
khatwanimohit, line 197: for my understanding, we usually have these kind of checks in types.py when running train.py, Is there something different about train_rl.py mesh creation which forces us to these checks here, also could we think about moving these checks to keep train_rl.py lean

xuefgu: Very good point. This brings out a key point in RL mesh creation: the generic parallelism config checks are based on one single mesh while we oftentimes have two meshes in RL. The RL meshes have to be treated differently. That's why the tensor parallelism didn't work before this PR. Also note that this PR wouldn't work without specifically setting sampler TP to `-1` in line 219.

That said, I will remove lines 198-201 but keep lines 202-206 to keep TP configs specific. We will gradually need to add other parallelism support in the trainer. At that time, we may want to think about how to make it more streamlined. But for this PR, it's for quickly adding TP support.

xuefgu: Actually, give `trainer_fsdp = trainer_devices_per_slice // tp`  in line 207 (to explicitly pass fsdp to `trainer_update` later), I prefer leaving lines 198 through 201 as is. Let's think about moving this logic out of train_rl later.

PR #3078, opened 2026-02-04
Add Colocated Python Checkpointing

SujeethJinesh:
# Description

Enable Experimental Colocated Python Checkpointing experience for Pathways on Cloud.

This feature enables using Orbax's Colocated Python Dispatchers as an alternative to Persistence API. It also enables users to save and restore checkpoints using zarr3 and ocdbt tensorstore formats. Currently in the process of performance tuning.

Further integration/unit testing will be done in a future PR once Colocated Python testing is incorporated.

This relies on Orbax version >=0.11.33.

FIXES: b/388583223

# Tests

Tested manually across a test matrix at [go/colocated-python-checkpointing-in-maxtext](https://goto.google.com/colocated-python-checkpointing-in-maxtext). Integration tests will need to be added once building a colocated python sidecar is supported in MaxText (WIP).

# Checklist


dependencies/requirements/generated_requirements/cuda12-requirements.txt
```diff
@@ -245,6 +245,7 @@ tzdata>=2025.2
 uritemplate>=4.2.0
 urllib3>=2.5.0
 uvicorn>=0.38.0
+uvloop>=0.19.0
```
hengtaoguo, line 248: Hi we usually don't recommend directly editing dependencies under `generated_requirements` folder. These two txt files are generated from `base_requirements` as in this [guide](https://maxtext.readthedocs.io/en/latest/install_maxtext.html#update-maxtext-dependencies). You need to edit [base requirements](https://github.com/AI-Hypercomputer/maxtext/tree/main/dependencies/requirements/base_requirements), run seed-env to generated a new set of generated requirements.

Your current patch can work temporarily, but if someone else generated new requirement files, your current change will be lost without notice.

SujeethJinesh: I created b/486268025 to try to figure out the issue, but it looks like when I follow that procedure, we get into a severe dependency hell with cloud-tpu-diagnostics and some other subsequent libraries.

I think for the purposes of my checkin, the only thing strictly needed actually is just that orbax be upgraded to version 0.11.33 or greater. Uvloop comes from Orbax.

hengtaoguo (approved):
cc @bvandermoon Hi this approval is to temporarily unblock this PR, since the generated dependency is too complicated to resolve shortly.

dependencies/requirements/generated_requirements/cuda12-requirements.txt
```diff
@@ -155,7 +155,7 @@ opt-einsum>=3.4.0
 optax>=0.2.6
 optree>=0.18.0
 optype>=0.14.0
-orbax-checkpoint>=0.11.28
+orbax-checkpoint>=0.11.33
```
sadikneipp, line 158: do we need to add this? if it's automated please disregard

benchmarks/recipes/user_configs.py
```diff
@@ -124,4 +128,7 @@ def __post_init__(self):
     selected_model_framework=["pathways"],
     selected_model_names=["llama3_1_8b_8192"],
     priority="medium",
+    base_output_directory=None,  # GCS Bucket path
```
sadikneipp, line 128: Should we have a good default here? Just because I see other fields have good defaults.

src/maxtext/common/checkpointing.py
```diff
@@ -269,6 +272,17 @@ def create_orbax_checkpoint_manager(
       logger=orbax_logger,
   )
 
+  # Use Colocated Python checkpointing optimization (Single Controller only).
+  if enable_single_controller and colocated_python_checkpointing:
```
sadikneipp, line 276: consider adding an early failure if `colocated_python_checkpointing` is true but `enable_single_controller` is false

dependencies/requirements/generated_requirements/cuda12-requirements.txt
SujeethJinesh: This is manually updated because the automated way is currently failing (b/486268025). It should be picked up in future generated requirements though.

benchmarks/recipes/user_configs.py
SujeethJinesh: We construct a default bucket path for the user and error out if it doesn't exist. I expect most people using this script will update this parameter though. So no need to update.

src/maxtext/common/checkpointing.py
SujeethJinesh: Thanks! Added to types.py validation.

src/maxtext/common/checkpointing.py
```diff
@@ -269,6 +272,17 @@ def create_orbax_checkpoint_manager(
       logger=orbax_logger,
   )
 
+  # Use Colocated Python checkpointing optimization (Single Controller only).
```
shauryagup, line 275: What makes the colocated python checkpointing implementation work only for pathways? It should work on both McJAX and Pathways and if it doesn't, a bug should be filed to fix it.

SujeethJinesh: This comment is only there to say that it's an optimization for Pathways. I haven't explicitly tested it with mcjax. But I think from the implementation of it, getting the cpu device mesh may mean a different behavior. I've filed a verification bug that I can pick up as I also verify performance - b/488077918

src/maxtext/configs/types.py
```diff
@@ -314,6 +314,7 @@ class Checkpointing(BaseModel):
       True, description="If True, saves a final checkpoint upon training completion."
   )
   enable_continuous_checkpointing: bool = Field(False, description="If True, enables continuous checkpointing.")
+  colocated_python_checkpointing: bool = Field(False, description="If True, enables colocated python checkpointing.")
```
bvandermoon: Nit: I wonder if this type of a feature would benefit from more of a description here. It's hard to understand what the current set of flags are all responsible for in MaxText. Not trying to solve that in this PR, but I wonder if there is some external documentation that could be linked here or somethign?

SujeethJinesh: Good point. It looks like the descriptions are short here so I've made a tiny update to make it a bit clearer but at the expense of it being a bit more verbose. Thanks!

SujeethJinesh:
I'm seeing the following issue with my formatting:

```
pyink....................................................................Failed
- hook id: pyink
- exit code: 1

would reformat benchmarks/recipes/user_configs.py
would reformat src/maxtext/configs/types.py

Oh no! 💥 💔 💥
2 files would be reformatted, 2 files would be left unchanged.

mdformat.............................................(no files to check)Skipped
```

I've gotten 3 approvals already and will just push a commit to fix the linting with `pyink benchmarks/recipes/user_configs.py src/maxtext/configs/types.py`. This results in a large number of whitespace changes.

Confirmed offline with @bvandermoon this is ok to do.

PR #3084, opened 2026-02-04
Fix: Added sharding Constraints for MTP block (b/481469708)

parambole:
# Description

The fix involves adding a single, targeted sharding annotation to `target_token_embedding` within the `MultiTokenPredictionLayer.__call__` method using `sharding.maybe_shard_with_logical`. This ensures that sharding is correctly propagated through the MTP layer's internal operations, including the projection and the subsequent transformer layer.

FIXES: b/481469708

# Tests

Verified with `train_compile.py` on the TPU backend and validated on ironwood run, where the model now trains without erros.

# Checklist

### DIFF
--- a/src/MaxText/layers/multi_token_prediction.py
+++ b/src/MaxText/layers/multi_token_prediction.py
@@ -16,20 +16,21 @@
 
 from typing import Type
 
+from flax import linen as nn
+from flax import nnx
 import jax
 import jax.numpy as jnp
 from jax.sharding import Mesh
 
-from flax import linen as nn
-from flax import nnx
-
+from MaxText import sharding
 from MaxText.common_types import Config, MODEL_MODE_TRAIN
+from MaxText.globals import EPS
+from MaxText.layers import nnx_wrappers
+from MaxText.layers.decoders import DecoderLayer
+from MaxText.layers.initializers import variable_to_logically_partitioned
 from MaxText.layers.linears import DenseGeneral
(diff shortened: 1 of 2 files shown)


NuojCheng (approved):

src/MaxText/layers/multi_token_prediction.py
```diff
@@ -118,10 +119,11 @@ def __init__(
     self.transformer_layer = nnx_wrappers.ToNNX(mtp_transformer_layer, rngs=rngs)
 
     # ToNNX requires explicit initialization with sample inputs for proper parameter setup.
+    batch_size, seq_len = max_utils.get_batch_seq_len_for_mode(config=cfg, model_mode=MODEL_MODE_TRAIN)
     self.transformer_layer.lazy_init(
-        inputs=jnp.zeros((1, 1, cfg.base_emb_dim), dtype=cfg.dtype),
+        inputs=jnp.zeros((batch_size, seq_len, self.config.base_emb_dim), dtype=self.config.dtype),
```
RissyRan: Usually we use `self.config.emb_dim` instead of base_emb_dim. Potentially leverage the global scaling: https://github.com/AI-Hypercomputer/maxtext/blob/55b57ffe957c9a36c42f4ac730ca35849115e9ab/src/maxtext/configs/base.yml#L144

parambole: Thank you for the suggestion. Done!

PR #3096, opened 2026-02-05
Qwen3-Omni Checkpoint Mapping

entrpn:
Original author @eitanporat in https://github.com/AI-Hypercomputer/maxtext/pull/2728
# Description

Checkpoint conversion mapping from hf

# Tests

- Converted HF weights to orbax checkpoint. 
- Tested regressions on other multimodal models by running SFT using tests/end_to_end/tpu/gemma3/4b/test_gemma3_multimodal_sft.sh (did not run jetstream/generation).
- Logs: https://paste.googleplex.com/5213472762757120

# Checklist

### DIFF
--- a/src/MaxText/layers/encoders.py
+++ b/src/MaxText/layers/encoders.py
@@ -64,7 +64,14 @@ class VisionEncoder(nnx.Module):
   def __call__(self, input_images, deterministic=False):
     # vision encoder output, frozen params in many cases
     encoder = getattr(self, self.encoder_name)
-    embeddings = encoder(input_images, deterministic=deterministic)
+    encoder_output = encoder(input_images, deterministic=deterministic)
+
+    deep_feats = None
+    if isinstance(encoder_output, tuple):
+      embeddings = encoder_output[0]
+      deep_feats = encoder_output[1]
+    else:
+      embeddings = encoder_output
 
     if self.config.freeze_vision_encoder_params:
       embeddings = jax.lax.stop_gradient(embeddings)
@@ -73,7 +80,7 @@ class VisionEncoder(nnx.Module):
     projector = getattr(self, self.projector_name)
     embeddings = projector(embeddings)
 
-    return embeddings
+    return embeddings, deep_feats
 
 
 class AudioEncoder(nnx.Module):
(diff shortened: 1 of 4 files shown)


src/MaxText/layers/models.py
```diff
@@ -154,7 +154,8 @@ def __call__(
     audio_embeddings = None
 
     if self.config.use_multimodal and encoder_images is not None:
-      image_embeddings = self.vision_encoder(input_images=encoder_images, deterministic=not enable_dropout)
+      # qwen3-omni-30b-a3b returns deep features from the vision encoder.
+      image_embeddings, _ = self.vision_encoder(input_images=encoder_images, deterministic=not enable_dropout)
```
hengtaoguo, line 158: Since you changed the output format of encoder, could you verify if all other references have been updated too?

src/MaxText/layers/encoders.py
```diff
@@ -64,7 +64,14 @@ def _setup_vision_encoder_layers(self):
   def __call__(self, input_images, deterministic=False):
     # vision encoder output, frozen params in many cases
     encoder = getattr(self, self.encoder_name)
-    embeddings = encoder(input_images, deterministic=deterministic)
+    encoder_output = encoder(input_images, deterministic=deterministic)
+
+    deep_feats = None
```
entrpn, line 69: @hengtaoguo instead of modifying all references, this is how its done to return a encoder_output and deep_feats which will be None for the other models.

src/MaxText/layers/models.py
aireenmei: Should we update return var number where vision_encoder is called, such as https://github.com/AI-Hypercomputer/maxtext/blob/main/src/MaxText/layers/models.py#L462

aireenmei: Also the returned deep_feats should not be _ ?

entrpn: @aireenmei it cannot be used right now, it requires the next PR in our list: https://github.com/AI-Hypercomputer/maxtext/pull/2729 thus I'm conforming to the order.

PR #3109, opened 2026-02-07
Support FP8 in batch split config

BirdsOfAFthr:
# Description

This update enables FP8 quantization support for DeepSeek batch split configurations.

When quantization is active, the following changes apply:

* Kernel Quantization: gmm kernels allow FP8 recipes (defined via the MaxText command line) in both forward and backward passes.

* `gmm` Weight All-Gathers: gmm weight all-gathers are now quantized along non-expert, non-tensor, and non-tensor_transpose axes.

Note: Quantization for dispatch all-gathers is currently out of scope and will be implemented in a follow-up PR.

# Tests

* Verification: Validated via end-to-end (e2e) perf and convergence benchmarks.

* Coverage: Unit tests will be added in a subsequent update.

# Checklist

### DIFF
--- a/src/maxtext/kernels/megablox/ops.py
+++ b/src/maxtext/kernels/megablox/ops.py
@@ -43,6 +43,7 @@ def gmm(
     use_tokamax_backend: bool = False,
     weight_gather_axes: List[Tuple[str, int]] | None = None,
     input_buffer_count: tuple[int, int, int] = (2, 2, 2),
+    combine_scopes: bool = False,
 ):
   """Grouped matrix multiplication operation."""
   quantization_rule = None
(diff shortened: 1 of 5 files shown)


src/maxtext/kernels/megablox/ops.py
```diff
@@ -212,14 +214,12 @@ def _gmm_bwd(
         quantization_rule.bwd_qtype,
         channelwise_axes=[] if quantization_rule.disable_channelwise_axes else [0],
         calibration_method=quantization_rule.bwd_calibration_method,
-        scale_dtype=jnp.float32,
     )
```
suexu1025, line 217: do we change scale_dtype here?

BirdsOfAFthr: removing this bc we can leverage bf16 scales, which will use native vpu ops

PR #3118, opened 2026-02-10
Add the option to avoid loading the checkpoint twice

xuefgu:
# Description

Add the option to avoid loading the checkpoint twice.

It's protected with the `load_checkpoint_only_once` flag.

# Tests

Manually tested.

# Checklist

### DIFF
--- a/src/MaxText/rl/train_rl.py
+++ b/src/MaxText/rl/train_rl.py
@@ -358,10 +358,18 @@ def rl_train(trainer_config, sampler_config, trainer_devices, sampler_devices):
     )
 
   # TODO: @mazumdera: change this to use lora
-  # TODO: @xfgu: instead of restoring a second time from GCS, can we just copy reference_model
-  # Load policy model
-  max_logging.log("Creating policy model with same config as reference model on trainer mesh")
-  actor_model, actor_mesh = get_maxtext_model(trainer_config, trainer_devices)
+  if trainer_config.load_checkpoint_only_once:
+    max_logging.log("Creating policy model by copying reference model instead of restoring from checkpoint again.")
+    with reference_mesh:
+      actor_base_model = nnx.clone(reference_model.base)
+      use_no_op_mappings = "maxtext_config" in trainer_config.vllm_additional_config
+      actor_model = TunixMaxTextAdapter(base_model=actor_base_model, use_no_op_mappings=use_no_op_mappings)
+      actor_model.config = None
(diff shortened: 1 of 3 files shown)


src/maxtext/configs/types.py
```diff
@@ -288,6 +288,7 @@ class Checkpointing(BaseModel):
   lora_input_adapters_path: PathStr = Field("", description="Input GCS path for LoRA adapters.")
   load_full_state_path: PathStr = Field("", description="Loads the complete training state from a checkpoint path.")
   enable_checkpointing: bool = Field(True, description="If True, enables saving checkpoints during training.")
+  load_checkpoint_only_once: bool = Field(False, description="If True, deep copy the reference model to the actor model.")
```
bvandermoon, line 291: What was the reason for loading this twice previously?

xuefgu: The reference model has to stay the same while the actor model is being updated during training. They can't share the same parameters.

src/MaxText/rl/train_rl.py
```diff
@@ -530,11 +538,23 @@ def rl_train(trainer_config, sampler_config, trainer_devices, sampler_devices):
...
+    max_logging.log("Capturing reference model state before training.")
+    ref_state_before = nnx.to_pure_dict(nnx.state(reference_model.base, nnx.Param))
+
   max_logging.warning("Starting RL training...")
 
   with reference_mesh, nn_partitioning.axis_rules(trainer_config.logical_axis_rules):
     rl_trainer.train(train_dataset)
 
+  if trainer_config.load_checkpoint_only_once:
+    max_logging.log("Checking if reference model state changed during training.")
```
khatwanimohit, line 551: can you share how much time it takes to do this check for our Qwen model. This seems to be an expensive check to do, if it is then can we hide it behind a flag

xuefgu: Llama70B takes ~6 secs to check in my manual testing. I would let folks routinely running Qwen provide the numbers for it :-)

That said, the whole block is a temporary safety check before the `load_checkpoint_only_once` flag is removed. A temporary 20 sec delay for a 235B model may not be worth adding complexities/mental load that comes with another flag.

khatwanimohit: agreed, thanks for the number for Llama70B, its pretty cheap.

khatwanimohit (approved):

PR #3130, opened 2026-02-12
Do not store mtp_losses/mtp_acceptance in state params

bvandermoon:
# Description

* Do not store `mtp_losses` and `mtp_acceptance` as part of params. This was causing a checkpoint loading error because they are not stored as part of the checkpoints.
* Update self attribute names to match pre-NNX checkpoints
* NOTE: Looks like pyink wants to update the format for the whole file. I'll do this in a follow-up PR to avoid losing the diff of this change

FIXES: b/483723849

# Tests

Successfully ran the XPK workload in the linked bug with checkpoint loading.

# Checklist

### DIFF
--- a/src/MaxText/layers/multi_token_prediction.py
+++ b/src/MaxText/layers/multi_token_prediction.py
@@ -128,6 +128,40 @@ class MultiTokenPredictionLayer(nnx.Module):
         model_mode=MODEL_MODE_TRAIN,
     )
 
+
+  @property
+  def embedding_norm(self):
+    return getattr(self, f"mtp_{self.layer_number}_embedding_norm")
+
+  @embedding_norm.setter
+  def embedding_norm(self, module):
+    setattr(self, f"mtp_{self.layer_number}_embedding_norm", module)
+
+  @property
+  def hidden_state_norm(self):
+    return getattr(self, f"mtp_{self.layer_number}_hidden_state_norm")
+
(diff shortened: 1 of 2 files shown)


src/MaxText/layers/multi_token_prediction.py
```diff
@@ -199,6 +199,8 @@ def __init__(
     self.mtp_preds = mtp_acceptance(jnp.zeros((1,), dtype=jnp.float32))
     self.mtp_mask = mtp_acceptance(jnp.zeros((1,), dtype=jnp.float32))
 
+    nnx.pop(self, (mtp_losses, mtp_acceptance))
```
cgarciae: Its a bit weird to `pop` on `__init__`, specially because you just created some Variables above, these would be lost.

src/MaxText/layers/multi_token_prediction.py
cgarciae: isn't it a bit easier just to delete these lines instead of assigning them and then using `pop`?

bvandermoon: Yes good point, thanks @cgarciae

src/MaxText/layers/multi_token_prediction.py
bvandermoon: Yes good point, thanks @cgarciae. Updated the PR

PR #3133, opened 2026-02-13
Optimize Qwen3-Next Gated Delta Network (GDN) Implementation

Rohan-Bierneni:
# Description

This PR introduces significant optimizations to the JAX implementation of the Gated Delta Network (GDN) layer for Qwen3-Next in src/MaxText/layers/qwen3.py. The new implementation transitions from an iterative refinement approach to a direct matrix inversion method using triangular solves, implements a mixed-precision strategy for better TPU utilization, and improves numerical stability. The TFLOPs calculation in maxtext_utils.py has also been updated to reflect the algorithmic changes.

1. Optimized GDN Algorithm (WY Representation)
 * Old Implementation: 
   * Used a jax.lax.fori_loop with iterative refinement to approximate the inverse of the intra-chunk attention matrix $(I - A)^{-1}$.
 * New Implementation: Replaces the loop with jax.scipy.linalg.solve_triangular. 
   * The algorithm now computes the $S$ matrix ($K \cdot K^T$), solves for $A = (I + S)^{-1}$, and pre-computes $W$ and $U$ factors ($u = A \cdot V$, $w = A \cdot K$).
 * **Impact**: This reduces control flow overhead and leverages optimized jax operations kernels on TPU, improving throughput for the recurrent calculations.

3. Mixed Precision Strategy
 * Old Implementation: Upcast all inputs (query, key, value, beta, g) to float32 at the very beginning of the function.
 * New Implementation: Adopts a mixed-precision approach:
   * Inputs: query, key, value, and beta are cast to compute_dtype (typically bfloat16) for heavy matrix multiplications.
   * Recurrence: The gates g, cumulative sums, and the recurrent state h are strictly forced to float32 to ensure numerical precision for the state update.
    * Accumulation: Matmuls accumulate in float32 via precision=jax.lax.Precision.HIGHEST.

4. Numerical Stability Improvements
 * Critical Fix: Applied masking to g_diff before exponentiation (jnp.where(mask, g_diff, -1e30)).
 * Reasoning: When testing with initial implementation of this optimized GDN, we applied the mask after exp(). In cases where g_diff contained large positive values in the masked-out upper triangle, exp() could produce inf, which resulted in NaN gradients even if zeroed out later. This change prevents inf generation entirely.

5. Recurrent State Update (scan)Update: 
 * The scan body has been rewritten to use the pre-computed $W$ and $U$ chunks.
   * State Update: $h_{new} = h \cdot \text{decay} + W^T \cdot U$
   * Output: Computed as the sum of the inter-chunk term ($q \cdot \exp(g) \cdot h$) and the intra-chunk attention term.

6. TFLOPs Calculation Update:
 * Updated calculate_gated_delta_net_flops_per_device in src/maxtext/utils/maxtext_utils.py. The cost model now accurately reflects the new "WY" representation complexity:
 * Intra-chunk: 
   * Accounts for $S = KK^T$, Triangular Solve ($\approx C^2$), and $W/U$ generation.
 * Inter-chunk: 
   * Accounts for the $W^T U$ state update and output projections. 
   * Formula: roughly $H_{eff} \times (6 C D_k + 4 C D_v + 4 D_k D_v + C^2)$ per token.

FIXES: b/123456

# Tests

Ran the script benchmark_gdn_optimization.py which is included as part of the pr. The script comares the new GDN impl against the baseline GDN which is currently in main. The script compares forward pass speed, backward pass speed, memory used, and tests stability for no NaN gradients. It also saves a profile of the GDN to local filesystem.

This is the output of running the script: https://paste.googleplex.com/6559852970246144
**tldr:**
* Forward pass speedup: **1.45x**
* Train step speedup: **3.33x**
* Memory Reduction: **77%**

(In Progress) Will also run other tests to verify optimized GDN fits within end-to-end model:
* Forward pass logit test: https://paste.googleplex.com/5438633436905472 
* Sample train run on synthetic data: https://screenshot.googleplex.com/AKrwzS3wzfst7Gb
  * https://pantheon.corp.google.com/kubernetes/service/europe-west4/mlperf-v5p/default/rbierneni-q3next-2026-02-17-23-09-30/logs?e=13802955&mods=allow_workbench_image_override&project=cloud-tpu-multipod-dev
  * **MFU: ~5%**

# Checklist

### DIFF
--- a/src/MaxText/layers/qwen3.py
+++ b/src/MaxText/layers/qwen3.py
@@ -21,6 +21,7 @@ import math
 
 import jax
 import jax.nn
+from jax import lax
 from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


src/MaxText/layers/qwen3.py
```diff
@@ -21,6 +21,7 @@
 
 import jax
 import jax.nn
+from jax import lax
```
parambole, line 24: Do we also need to import `jax.scipy` ?

src/MaxText/layers/qwen3.py
```diff
@@ -50,236 +51,189 @@
...
+  chunk_decay_all = jnp.exp(g_cumsum[..., -1]).transpose(1, 0, 2)
+
+  if initial_state is None:
+    h_init = jnp.zeros((B, H, K_dim, V_dim), dtype=jnp.float32)
+  else:
+    h_init = initial_state.astype(jnp.float32)
+
+  xs = (w_scan, u_scan, q_scan, k_scan, v_scan, g_scan, beta_scan, chunk_decay_all)
+
+  def scan_body(h, args):
```
parambole, line 296: Maybe a future improvement but one IMO would give a good perf boost 

Calculate Math for ALL Chunks at once (Parallel)
1. Start Loop:Take Pre-calculated Math for Chunk 1 + State -> Update State.
2. Take Pre-calculated Math for Chunk 2 + State -> Update State.Done.

currently 

```
def scan_body(prev_state, x):
    q, k, v = x
    # HEAVY MATH HAPPENS HERE (SEQUENTIALLY)
    attn = jnp.matmul(q, k) 
    intra_out = jnp.matmul(attn, v)
    
    # STATE UPDATE
    new_state = prev_state * decay + intra_out
    return new_state, intra_out

jax.lax.scan(scan_body, init, xs)
```

Proposed 

```
# 1. HEAVY MATH HAPPENS HERE (PARALLEL - ALL CHUNKS AT ONCE)
all_attn = jnp.matmul(qs, ks)
all_intra_outs = jnp.matmul(all_attn, vs)

def scan_body(prev_state, intra_out):
    # 2. ONLY STATE UPDATE HAPPENS HERE (VERY FAST)
    new_state = prev_state * decay + intra_out
    return new_state, intra_out

jax.lax.scan(scan_body, init, all_intra_outs)
```

src/MaxText/layers/qwen3.py
```diff
@@ -50,236 +51,189 @@
 
 
 def jax_chunk_gated_delta_rule(
-    query: Array,
-    key: Array,
-    value: Array,
-    g: Array,
-    beta: Array,
+    query: jax.Array,
```
RissyRan: Did you meet issues when using `Array` from common type?

src/MaxText/layers/qwen3.py
```diff
@@ -50,236 +51,189 @@
...
-
-  Returns:
-    Output tensor. Shape (B, S, H, D_v)
-    Final recurrent state. Shape (B, H, D_k, D_v) or None
+  Optimized JAX implementation of Gated Delta Rule (Mixed Precision + Stability Fix).
+  
+  Precision Strategy:
+  - Inputs (q, k, v, beta): 'compute_dtype' (e.g. bfloat16) for tensor core usage.
+  - Gates (g) & State (h): float32 (forced) for numerical stability.
+  - Matmuls: compute_dtype inputs -> float32 accumulation/output.
```
RissyRan: In MaxText, for Matmuls, we usually do:

activation (bf16) @ weights (cast f32 to bf16) --> outputs, but params are initialized in f32 for master weights in bwd pass. Shall we try here as well and see if also stable? So output will be bf16 in this case.

One example: [initialization](https://github.com/AI-Hypercomputer/maxtext/blob/98fb5cf62678efd09356caa05ae3252e8c55cdbf/src/MaxText/layers/moe.py#L423-L429), [cast](https://github.com/AI-Hypercomputer/maxtext/blob/98fb5cf62678efd09356caa05ae3252e8c55cdbf/src/MaxText/layers/moe.py#L1993-L1995)

src/maxtext/scratch_code/benchmark_gdn_optimization.py
```diff
@@ -0,0 +1,533 @@
...
+from typing import Any, cast
+
+# Import common dependencies
+from MaxText import common_types
+from MaxText.layers import normalizations
+from MaxText.layers.linears import DenseGeneral
+from MaxText.layers import qwen3 
+
+# ==============================================================================
+# SECTION 1: THE MIRROR (BASELINE IMPLEMENTATION)
```
RissyRan: Wondering the purpose of this script in the main branch?

tests/unit/train_compile_test.py
```diff
@@ -794,3 +795,33 @@ def test_olmo3_7b(self):
             "max_target_length=1024",
         )
     )
+
+  @pytest.mark.cpu_only
+  def test_qwen3_next_tokamax(self):
+    """AOT test for Qwen3-Next with Tokamax GMM on v5p-128"""
```
shuningjin: I didn't see the flag `use_tokamax_gmm=True`, despite the comment says Tokamax GMM

I see another test. Do we need both?
https://github.com/AI-Hypercomputer/maxtext/blob/3c56dd341096985f150a0a91fa26b429fc8c644e/tests/unit/train_compile_test.py#L736-L750

tests/unit/qwen3_next_vs_reference_test.py
```diff
@@ -766,6 +766,105 @@ def test_l2norm(self):
     )
     print("test_l2norm passed!")
 
+  def test_chunk_gated_delta_rule_logic_kl(self):
+    """
+    Directly tests the `jax_chunk_gated_delta_rule` against the PyTorch reference
+    using KL Divergence instead of strict atol/rtol.
+    """
```
shuningjin: Could you explain why this test is need? I see we already have `test_chunk_gated_delta_rule_logic`

https://github.com/AI-Hypercomputer/maxtext/blob/0b42732e72411639034dcdc6734f890611a3b81f/tests/unit/qwen3_next_vs_reference_test.py#L868-L871

src/maxtext/utils/maxtext_utils.py
```diff
@@ -533,33 +533,28 @@ def calculate_gated_delta_net_flops_per_device(config):
...
-  # - Inter-chunk terms (Recurrent State D_k * D_v):
-  #   - attn_i (Q*K): 2 * B * S * H_v * C * D_k
-  #   - v_prime (K*S): 2 * B * S * H_v * D_k * D_v
-  #   - attn_inter (Q*S): 2 * B * S * H_v * D_k * D_v
-  #   - core_out (A*V): 2 * B * S * H_v * C * D_v
-  #   - update (K*V): 2 * B * S * H_v * D_k * D_v
-  flops_inter = (2 * B * S * H_v * C * (D_k + D_v)) + (6 * B * S * H_v * D_k * D_v)
-
-  flops_core = flops_intra + flops_inter
+  # 3. Core Gated Delta Net (Optimized WY Representation)
```
RissyRan: Taking another look. It is clear that we are calculating the hardware utilization here.

It seems from recurrent view, which gives 7 * D^2.

```
# --- 3. CORE DELTA INTERACTION (Recurrent View) ---
    # This is the "True" Model work. For each token:
    # 1. Predict: v_hat = S @ k       -> 2 * D_k * D_v
    # 2. Update:  S = S - beta*v_err@k -> 3 * D_k * D_v (outer product + sub + scale)
    # 3. Read:    out = S @ q         -> 2 * D_k * D_v
```

From [Megatron](https://github.com/NVIDIA/Megatron-LM/blob/8f1c2f8ae53b4e3f32c0ae7f397d8b38a675eaa2/megatron/training/training.py#L532-L534), it seems from Chunk-parallel view, which gives 4 * D^2, and `fma_expansion_factor` seems as 2.


```
                        + num_v_heads
                        * (v_head_dim ** 2)
                        * 4  # KK^T, VK^T, S(a(I-bKK^T)), and SQ
```

To ensure consistent benchmarking with industry-standard baselines, (after cross-checking) let's align our calculations with the Megatron methodology?

cc @Rohan-Bierneni @parambole @shuningjin

src/MaxText/layers/qwen3.py
Rohan-Bierneni: We dont have to since we import jax at the top of the file. To call it we just use:

```
jax.scipy.linalg.solve_triangular
```

src/MaxText/layers/qwen3.py
Rohan-Bierneni: This could be a good improvement. We can revisit this in a follow up pr for optimizing the ops done in the GDN.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: I did not have any issues. I was able to run jit compile tests on both cpu and tpu vm, as well as forward pass logit check. Although I should use maxtext common_types Array instead for consistency with other files

src/maxtext/scratch_code/benchmark_gdn_optimization.py
Rohan-Bierneni: I initially had it for testing purposes, but will remove it now since tests are verified. Resolved.

tests/unit/train_compile_test.py
Rohan-Bierneni: Yes I had this for testing purposes to verify different configs would compile or not before triggering an xlml run. I previously did test with the tokamax flags but will now remove this testcase since it does not need to be in main. Resolved.

tests/unit/qwen3_next_vs_reference_test.py
Rohan-Bierneni: This added test was just for testing purposes, as I wanted to see the kl divergence instead of atol/rtol when comparing logit check for pytorch vs optimized jax code. Do not need thi stest anymore and will remove it.

src/MaxText/layers/qwen3.py
Rohan-Bierneni: I have modified the code to follow this. The q, k, v, beta, and gate in jax_chunk_gated_delta_rule are computed in cfg.dtype (bf16. The output of this layer should be cfg.dtype as well. 

Also, changed the A_log and dt_bias to this:
```
A_log = jnp.asarray(self.A_log[...], dtype=cfg.dtype) 
dt_bias = jnp.asarray(self.dt_bias[...], dtype=cfg.dtype)
```

However, for stability purposes and since we are working with a fixed state, some ops are forced to stay in highest precision. Specifically the ops related to the recurrent state and chunk operations.

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: Thanks for proposing this, following megatron then I will change the core_flops calculation to this:
```
Total Core FLOPs = 2 (FMA) * 4 (Ops) * H * D^2 = 8 * H * D^2 per token
```

src/maxtext/utils/maxtext_utils.py
```diff
@@ -503,7 +503,10 @@ def get_dense_moe_layers(config):
 
 
 def calculate_gated_delta_net_flops_per_device(config):
-  """Calculates the FLOPs for a single Gated Delta Net (Linear Attention) layer."""
+  """
+  - Calculates the FLOPs for a single Gated Delta Net (Linear Attention) layer.
+  - Core complexity is based on the recurrent state update view (4 ops * 2 FLOPs = 8).
+  """
```
shuningjin, line 511: might be good to add the megatron reference: https://github.com/NVIDIA/Megatron-LM/blob/8f1c2f8ae53b4e3f32c0ae7f397d8b38a675eaa2/megatron/training/training.py#L513

shuningjin (approved):
Thanks!

src/maxtext/utils/maxtext_utils.py
Rohan-Bierneni: Resolved.

Rohan-Bierneni:
Manually adding pull_ready tag since all tests pass and 3 LGTMs.

PR #3183, opened 2026-02-18
Integrate Engram into custom model

RissyRan:
# Description

Integrate Engram feature into a custom model
* Add configs into `base.yml` and `type.py`, integrates with `deepseek-custom` model
* This PR supports unscan version of Engram, and scan version will be next PR.
* Tried to initialize the hash map to model level for one time initialization, but met various JAX initialization error (see more in b/478294699 and this [PR](https://github.com/AI-Hypercomputer/maxtext/pull/3125)). Also added a comment there.
* Currently, to make it work, you will see mix of `jnp` and `np` as some operations running on CPU to avoid ConcretizationTypeError and other issues. 

# Tests

* Expect github runners to pass
* Unit tests still passing for Engram: [link](https://paste.googleplex.com/4533027603611648)
* End-to-end unscan training for custom model: [link](https://paste.googleplex.com/6288392850636800)
* Sanity check for DS v2 (expect no impact)

```
Before change:

I0218 19:13:51.139775 139970093350464 max_utils.py:697] 	Using (GB) 43.98 / 95.74 (45.936912%) on TPU_1(process=0,(1,0,0,0))
I0218 19:15:18.498541 139970093350464 metric_logger.py:181] completed step: 19, seconds: 4.366, TFLOP/s/device: 123.145, Tokens/s/device: 7505.830, total_weights: 131072, loss: 8.135

After change:

I0218 19:05:35.490278 140385131265600 max_utils.py:697] 	Using (GB) 43.99 / 95.74 (45.947357%) on TPU_0(process=0,(0,0,0,0))
I0218 19:07:02.849351 140385131265600 metric_logger.py:181] completed step: 19, seconds: 4.366, TFLOP/s/device: 123.144, Tokens/s/device: 7505.806, total_weights: 131072, loss: 8.135
```

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -315,6 +315,7 @@ out_proj: 'remat'
 mla_q: 'remat'
 mla_kv: 'remat'
 attention_out: 'remat'
+engram: 'remat'
 
 optimizer_memory_host_offload: False
 parameter_memory_host_offload: False
@@ -1102,3 +1103,20 @@ force_q_layout: false
 mhc_expansion_rate: 1
 # The number of iterations for the Sinkhorn-Knopp algorithm.
 sinkhorn_iterations: 20
+
+################################## DeepSeek Engram ##################################
+# Indices of transformer layers where Engram are integrated; leave empty [] to disable.
+# Example: [1, 4] attaches to the 2nd and 5th layer. 
+engram_layers: []
+# The max 'n' in N-gram. Example: n=3 means it covers both 2-grams and 3-grams.
+engram_max_ngram_size: 3
+# Number of heads dedicated to the Engram.
+engram_num_heads: 8
+# Head dimension for heads.
+engram_head_dim: 1280
+# List of minimum head vocab sizes for each n-gram order.
+engram_vocab_bases: []
+# Temporal window size for Engram convolution.
+engram_kernel_size: 4
+# The seed for Engram hash mapping.
+engram_seed: 0
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -897,6 +897,7 @@ class RematAndOffload(BaseModel):
       RematLocation.REMAT,
       description="Remat policy for the attention output.",
   )
+  engram: RematLocation = Field(RematLocation.REMAT, description="Remat policy for the engram output.")
 
   optimizer_memory_host_offload: bool = Field(False, description="Offload optimizer state to host memory.")
   parameter_memory_host_offload: bool = Field(False, description="Offload parameters to host memory.")
@@ -1630,6 +1631,23 @@ class SpecialTokens(BaseModel):
   solution_end_token: str = Field("</answer>", description="Token to mark the end of a solution section.")
 
 
+class Engram(BaseModel):
+  """Configuration for DeepSeek Engram (https://www.arxiv.org/pdf/2601.07372)."""
+
+  engram_layers: list[int] = Field(
+      default_factory=list,
+      description="Indices of transformer layers where Engram are integrated.",
+  )
+  engram_num_heads: int = Field(8, description="Number of heads dedicated to the Engram.")
+  engram_head_dim: int = Field(1280, description="Head dimension for heads.")
+  engram_vocab_bases: list[int] = Field(
+      default_factory=list, description="List of minimum head vocab sizes for each n-gram order."
+  )
+  engram_max_ngram_size: int = Field(3, description="The max 'n' in N-gram.")
+  engram_kernel_size: int = Field(4, description="Temporal window size for Engram convolution.")
+  engram_seed: int = Field(0, description="The seed for Engram hash mapping.")
+
+
 class DerivedValues(BaseModel):
   """Holds all fields that are derived from other config values for perfect legacy compatibility."""
 
@@ -1782,6 +1800,7 @@ class MaxTextConfig(
     Quantization,
     # Core Model Architecture
     ModelArchitecture,
+    Engram,
     MTP,
     Logits,
     # Attention Mechanisms
@@ -2262,6 +2281,18 @@ class MaxTextConfig(
         and self.gradient_accumulation_steps > 1
     ):
       raise ValueError("FP8 quantization is not compatible with gradient accumulation.")
+    if self.engram_layers:
+      if not self.hf_access_token or not self.tokenizer_path:
+        raise ValueError(
+            "Engram requires both 'hf_access_token' and 'tokenizer_path' " "to load the Hugging Face tokenizer."
+        )
+      if self.scan_layers:
+        raise NotImplementedError("Currently Engram only supports unscanned version. Please set scan_layers=False.")
+      if len(self.engram_vocab_bases) != (self.engram_max_ngram_size - 1):
+        raise ValueError(
+            f"Engram vocab size mismatch: expected {self.engram_max_ngram_size - 1} (max_ngram_size - 1), "
+            f"but got {self.engram_vocab_bases}."
+        )
     if self.num_experts > 1:
       is_fully_moe = (
           self.interleave_moe_layer_step == 1
--- a/src/maxtext/layers/engram.py
+++ b/src/maxtext/layers/engram.py
@@ -35,6 +35,7 @@ import numpy as np
 from sympy import isprime
 from tokenizers import Regex, normalizers
 
+
 class CompressedTokenizer:
   """
   A canonicalizing wrapper that reduces vocabulary sparsity for n-gram lookup.
@@ -50,7 +51,8 @@ class CompressedTokenizer:
 
   def __init__(self, tokenizer: HFTokenizer):
     normalizer = self._build_normalizer()
-    self.lookup_table, self.num_new_token = self._build_lookup_table(tokenizer, normalizer)
+    self.lookup_table_np, self.num_new_token = self._build_lookup_table(tokenizer, normalizer)
+    self.lookup_table = jnp.array(self.lookup_table_np, dtype=jnp.int64)
 
   def __len__(self) -> int:
     return self.num_new_token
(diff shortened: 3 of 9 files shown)


src/maxtext/models/deepseek.py
```diff
@@ -91,6 +97,43 @@ def __init__(
...
+          rngs=rngs,
+      )
+      tokenizer = transformers.AutoTokenizer.from_pretrained(config.tokenizer_path, token=config.hf_access_token)
+      # TODO(ranran): Refactor NgramHashMapping to initialize once globally or at the model level.
+      # Moving this to decoders.py currently causes JAX initialization errors.
+      self.ngram_hash_mapping = NgramHashMapping(
+          engram_vocab_bases=config.engram_vocab_sizes,
+          max_ngram_size=config.engram_max_ngram_size,
+          engram_num_heads=config.engram_num_heads,
+          layer_ids=[layer_idx],
```
shuningjin: This should be initialized with global layer_ids, as in unit test. Otherwise hashing behavior is different.
https://github.com/AI-Hypercomputer/maxtext/blob/52e0ceafa5f65f8e99c44cf19acca90439f33d3a/tests/unit/engram_vs_reference_test.py#L762

src/maxtext/configs/base.yml
```diff
@@ -1093,3 +1094,20 @@ force_q_layout: false
...
+################################## DeepSeek Engram ##################################
+# Indices of transformer layers where Engram are integrated; leave empty [] to disable.
+# Example: [1, 4] attaches to the 2nd and 5th layer. 
+engram_layers: []
+# Number of heads dedicated to the Engram.
+engram_num_heads: 8
+# Head dimension for heads.
+engram_head_dim: 1280
+# List of vocab sizes for each n-gram order.
+engram_vocab_sizes: []
```
shuningjin: (1) This config can be confusing: This is not the actual size, but the search start for prime sizes.  

For accuracy, I suggest: `engram_vocab_bases`, and `# List of minimum head vocab sizes for each n-gram order in 2...engram_max_ngram_size`.

For example, if the `max_ngram = 2, num_head=3, engram_vocab_bases = [4]`
- ngram2, head0, find_next_unseen_prime > 4-1  -- size = 5
- ngram2, head1, find_next_unseen_prime > 5 --- size = 7
- ngram2, head2, find_next_unseen_prime > 7 --- size = 11

(2) to reflect the hierarchy (engram -> 2...max ngram -> num_head per ngram), the order could be:
```
engram_max_ngram_size 
engram_num_heads, engram_head_dim
engram_vocab_sizes
```

src/maxtext/configs/types.py
```diff
@@ -2247,6 +2264,13 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
         and self.gradient_accumulation_steps > 1
     ):
       raise ValueError("FP8 quantization is not compatible with gradient accumulation.")
+    if self.engram_layers:
```
shuningjin, line 2284: Thanks for adding the checks! Might be good to check `len(engram_vocab_sizes) == (engram_max_ngram_size -1)`

src/MaxText/layers/engram.py
```diff
@@ -181,10 +182,14 @@ def __init__(
     self.compressed_tokenizer = CompressedTokenizer(tokenizer)
     self.tokenizer_vocab_size = len(self.compressed_tokenizer)
     if pad_id is not None:
```
shuningjin: When pad_id is not none, we init self.pad_id, which is required by _get_ngram_hashes for padding.

When it is None, it is unclear from reference
https://github.com/deepseek-ai/Engram/blob/fb7f84a21f91223715394a33a1dc24bbfb7f788e/engram_demo_v1.py#L211
Maybe raise error, and add a todo in comment?

src/maxtext/models/deepseek.py
```diff
@@ -91,6 +97,43 @@ def __init__(
...
+          num_features=self.dummy_inputs_shape[-1],
+          dtype=self.config.dtype,
+          weight_dtype=self.config.weight_dtype,
+          kernel_axes=("norm",),
+          epsilon=self.config.normalization_layer_epsilon,
+          rngs=rngs,
+      )
+      tokenizer = transformers.AutoTokenizer.from_pretrained(config.tokenizer_path, token=config.hf_access_token)
+      # TODO(ranran): Refactor NgramHashMapping to initialize once globally or at the model level.
+      # Moving this to decoders.py currently causes JAX initialization errors.
```
shuningjin, line 112: Alternatively, can we initialize it in data pipeline? treating it as a special tokenizer, operating with np and cpu, similar to existing hf_tokenizer

RissyRan: Yeah, I tried both in this draft [PR](https://github.com/AI-Hypercomputer/maxtext/pull/3125):

1) put in the data pipeline: https://screenshot.googleplex.com/5dhowKp7EjzMuud - not working.
2) put in the model level: https://screenshot.googleplex.com/5FW7nxAXu9fFDCA - not working.

Met different issues here and there. I think if bandwidth allows, we could dive deep more.

src/maxtext/models/deepseek.py
RissyRan: Yes, this is global version: https://screenshot.googleplex.com/4pEmyDgFWhiXF6c

src/MaxText/layers/engram.py
RissyRan: OK, let me raise am error here.

src/maxtext/configs/base.yml
RissyRan: Yeah, it seems this is the start value for the current prime: https://github.com/deepseek-ai/Engram/blob/fb7f84a21f91223715394a33a1dc24bbfb7f788e/engram_demo_v1.py#L246.

Let me update a little bit.

RissyRan:
> Thank you! I have some minor comment. As a follow up, we could try placing ngram-hash-mapping globally for efficiency: either (1) at model level or (2) at data pipeline level, treat it as a special tokenizer.

Thanks for the review! As noted in the PR description and our previous [discussion](https://screenshot.googleplex.com/6zPforq3i3npFrq), I’d like to prioritize getting the core functionality landed first. We can then tackle those trickier edge cases in a follow-up. How does that sound?

src/maxtext/models/deepseek.py
shuningjin: To clarify, if `self.config.engram_layer_ids = [1, 15]`, both layer should have `layer_ids=[1, 15]`, rather than `layer_ids=[1]` and `layer_ids[15]` respectively. 

[Vocab size search](https://github.com/deepseek-ai/Engram/blob/fb7f84a21f91223715394a33a1dc24bbfb7f788e/engram_demo_v1.py#L236-L239) needs to know all layers at once. It affects the `seen_prime`: if layer1 takes prime 3, layer 15 cannot use that size. See this small example: https://paste.googleplex.com/6291527186907136.

RissyRan: Sounds good, updated.

tests/unit/engram_vs_reference_test.py
shuningjin, line 792: Could you update these `self.config` to `cfg` in class initialization to align unit test with our e2e? 
- For example, `self.config.engram_layer_ids` -> `cfg.engram_layers`. 
- May need to change `Config` in the top to align names. 
- Same for other tests in this file.

similar to 
https://github.com/AI-Hypercomputer/maxtext/blob/d8c8862f9eff45296e7d358a5c9c7bb062ff35de/tests/unit/deepseek32_vs_reference_test.py#L972-L989

RissyRan: Sure, I see mainly is `engram_layers` change. Updated and tested.

```
.
----------------------------------------------------------------------
Ran 8 tests in 51.458s

OK
```

src/maxtext/configs/types.py
```diff
@@ -2250,6 +2269,18 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+      if not self.hf_access_token or not self.tokenizer_path:
+        assert ValueError(
```
gagika: assert -> raise

src/maxtext/configs/types.py
```diff
@@ -2250,6 +2269,18 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+            "Engram requires both 'hf_access_token' and 'tokenizer_path' " "to load the Hugging Face tokenizer."
+        )
+      if self.scan_layers:
+        assert NotImplementedError("Currently Engram only supports unscanned version. Please set scan_layers=False.")
```
gagika: assert -> raise

src/maxtext/layers/engram.py
```diff
@@ -118,19 +120,19 @@ def _build_lookup_table(self, tokenizer: HFTokenizer, normalizer: normalizers.Se
 
     return old2new, len(key2new)
 
-  def __call__(self, input_ids) -> np.ndarray:
+  def __call__(self, input_ids) -> Array:
     """
     Maps original token IDs to compressed IDs.
     """
-    input_ids = np.asarray(input_ids, dtype=np.int64)
+    input_ids = jnp.asarray(input_ids, dtype=jnp.int64)
```
gagika, line 127: is jnp.int64 really needed? or `jnp.int32` is good enough for table sizes we will need?

src/maxtext/layers/engram.py
```diff
@@ -118,19 +120,19 @@ def _build_lookup_table(self, tokenizer: HFTokenizer, normalizer: normalizers.Se
...
 
     # Identify valid tokens (ignore padding/masks usually marked with negative IDs)
-    valid_mask = input_ids >= 0
-    valid_ids = input_ids[valid_mask]
+    # In JAX, we map negative IDs to 0 (or a valid index) for lookup, then mask output back.
```
gagika: should the comment be updated?

src/maxtext/configs/types.py
RissyRan: Sounds good!

src/maxtext/layers/engram.py
RissyRan: Good point. My initial attempt with jnp.int32 failed due to a mismatch with PyTorch's int64, but we could try unifying both to int32. The range (2^31-1) is should be sufficient for our vocab and hash IDs. @shuningjin, any concerns?

src/maxtext/layers/engram.py
RissyRan: I think so? I also update the code a little bit bellow.

src/maxtext/layers/engram.py
RissyRan: Also discussing offline for test failures by converting int64 to int32.

src/maxtext/layers/engram.py
RissyRan: I updated a little bit. Could you explain more if haven't addressed it yet.

gagika: looks good now, previous comment was refering to refactored code.

PR #3188, opened 2026-02-18
Support NNX ckpt in forward_pass_logit_checker

hengtaoguo:
# Description
Support NNX checkpoint in forward_pass_logit_checker.py:
- [New] Add a flag `--ckpt_type=["linen", "nnx"]` to support nnx ckpt. Default to linen.
- Implement `setup_decode_state_from_nnx()`, which now calls `from_pretrained()` to natively load an NNX model and shard properly across devices (instead of loading to CPU or duplicating across TPUs). This function will return a linen `maxtext_state` for decoding purposes.

When loading an NNX checkpoint that was saved via `TunixMaxTextAdapter` (e.g. an RL training checkpoint), the on-disk pytree can contain extra top-level keys such as to_nnx__rngs that are present in the checkpoint but absent from the target model's parameter tree. Passing both trees directly to `jax.tree.map` requires them to have identical structure, so the extra checkpoint keys cause a key-mismatch error. 
- `_filter_to_model_keys` prunes the restored checkpoint dict down to only the keys present in `model_arrays`, ensuring the two trees are structurally identical before the `jax.tree.map(_expand_checkpoint_to_model_shapes, ...)` call, without silently skipping any weights the model actually needs.

# Tests

Added unit test `test_returns_linen_train_state_and_annotations` which verifies that `setup_decode_state_from_nnx`:
1. Calls `from_pretrained` with `model_mode=MODEL_MODE_AUTOREGRESSIVE` (never touches a real checkpoint in CI)
2. Returns a linen `TrainState` with concrete `jax.Array` params nested under `state.params["params"]`
3. Extracts param values faithfully — each leaf is numerically identical to the source NNX model's `nnx.Param` values

End-to-end runs:
- NNX-RL with `base`, `--ckpt_type=nnx`: [example1](https://paste.googleplex.com/5791541516959744), [example2](https://paste.googleplex.com/5402751547867136)
- NNX-SFT without `base`, `--ckpt_type=nnx`: [example3](https://paste.googleplex.com/6319774511529984)
- Linen ckpt unimpacted: [example4](https://paste.googleplex.com/5626151453196288)

# Checklist

### DIFF
--- a/tests/utils/forward_pass_logit_checker.py
+++ b/tests/utils/forward_pass_logit_checker.py
@@ -53,6 +53,7 @@ from maxtext.layers import quantizations
 from maxtext.models import models
 from maxtext.utils import max_logging
 from maxtext.utils import maxtext_utils
+from maxtext.utils import model_creation_utils
 import numpy as np
 import torch
 import torch.nn.functional as F
@@ -447,7 +448,10 @@ def main(config, test_args):  # pylint: disable=W0621
     else:
       maxtext_model = models.transformer_as_linen(config, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
       init_state_fn = functools.partial(maxtext_utils.init_initial_state, maxtext_model, None, config, False, rng1)
-    maxtext_state, _ = maxtext_utils.setup_decode_state(config, mesh, None, init_state_fn)
+    if test_args.ckpt_type == "linen":
+      maxtext_state, _ = maxtext_utils.setup_decode_state(config, mesh, None, init_state_fn)
+    else:
+      maxtext_state, _ = model_creation_utils.setup_decode_state_from_nnx(maxtext_model, config, rng1, mesh)
 
     prompts = ["I love to", "Today is a", "What is the"]
     all_data_to_save = []
@@ -553,6 +557,14 @@ if __name__ == "__main__":
       required=False,
       default=False,
       help="Skip the first token during comparison to ignore BOS/init mismatches.",
+  )
+  parser.add_argument(
+      "--ckpt_type",
+      type=str,
+      required=False,
+      default="linen",
+      choices=["linen", "nnx"],
+      help="Checkpoint format to load: 'linen' (default) or 'nnx'.",
   )
 
   # Parse known args returns the namespace AND the list of remaining arguments
--- a/src/maxtext/utils/model_creation_utils.py
+++ b/src/maxtext/utils/model_creation_utils.py
@@ -31,7 +31,7 @@ import jax.numpy as jnp
 import numpy as np
 from jax.sharding import Mesh
 from maxtext.configs import pyconfig
-from maxtext.common.common_types import MODEL_MODE_TRAIN
+from maxtext.common.common_types import MODEL_MODE_AUTOREGRESSIVE, MODEL_MODE_TRAIN
 from maxtext.layers import quantizations
 from maxtext.models import models
 from maxtext.utils import max_logging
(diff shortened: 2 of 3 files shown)


src/maxtext/utils/maxtext_utils.py
```diff
@@ -1087,6 +1087,53 @@ def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
...
+      filtered_tree = {k: v for k, v in tree.items() if "to_nnx__rngs" not in k}
+      if (
+          len(filtered_tree) == 1
+          and "value" in filtered_tree
+          and isinstance(filtered_tree["value"], (jax.Array, np.ndarray))
+      ):
+        return filtered_tree["value"]
+      return {k: unwrap_nnx_values(v) for k, v in filtered_tree.items()}
+    return tree
+
```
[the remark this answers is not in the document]
hengtaoguo: Refactored to use `create_nnx_model` instead, which creates a properly sharded nnx model.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1087,6 +1087,53 @@ def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
...
+  _, state_mesh_annotations, state_mesh_shardings = get_abstract_state(model, None, config, rng, mesh, False)
+
+  loaded_params = load_orbax_checkpoint(config)
+  loaded_params = unwrap_nnx_values(loaded_params)
+  # NNX-RL checkpoints may have an extra 'base' nesting level, while NNX-SFT may not
+  raw_params = loaded_params["base"] if "base" in loaded_params else loaded_params
```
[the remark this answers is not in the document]
hengtaoguo: Obsolete

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1087,6 +1087,52 @@ def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
...
+
+  # Extract only Param values from the NNX model (filters out dropout/RNG state)
+  # nnx.state() returns a State object (custom pytree), which jax.device_put can't match
+  # against a plain-dict sharding spec. Convert to a plain nested dict first.
+  def _state_to_dict(tree):
+    if hasattr(tree, "items") and not isinstance(tree, jax.Array):
+      return {k: _state_to_dict(v) for k, v in tree.items()}
+    return tree
+
+  nnx_param_state = nnx.state(nnx_model, nnx.Param)
```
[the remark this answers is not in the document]
hengtaoguo: Done

tests/utils/forward_pass_logit_checker.py
```diff
@@ -425,7 +425,10 @@ def main(config, test_args):  # pylint: disable=W0621
     mesh = jax.sharding.Mesh(devices_array, config.mesh_axes)
     quant = quantizations.configure_quantization(config)
     maxtext_model = models.transformer_as_linen(config, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
-    maxtext_state, _ = maxtext_utils.setup_decode_state(maxtext_model, config, rng1, mesh, None)
+    if test_args.ckpt_type == "linen":
```
entrpn, line 451: Can we use the already existing enable_nnx flag instead of a creating a new one ckpt_type?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1087,6 +1087,54 @@ def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
...
+    state: state with decode params loaded from the NNX checkpoint
+    state_mesh_annotations: the mesh annotations for the state
+  """
+  from flax import nnx  # pylint: disable=import-outside-toplevel
+  from maxtext.utils.model_creation_utils import create_nnx_model  # pylint: disable=import-outside-toplevel
+
+  _, state_mesh_annotations, state_mesh_shardings = get_abstract_state(model, None, config, rng, mesh, False)
+
+  # Load the NNX model with checkpoint (handles NNX, NNX-RL, and linen checkpoint formats)
+  nnx_model, _ = create_nnx_model(config, mesh=mesh, model_mode=MODEL_MODE_AUTOREGRESSIVE)
```
entrpn: why are we manually forcing sharding via device_put instead of letting the model creation utility do it using `with mesh, nn_partitioning.axis_rules(cfg.logical_axis_rules):`

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1087,6 +1087,54 @@ def setup_decode_state(model, config, rng, mesh, checkpoint_manager):
...
+    config: config object
+    rng: jax.prng key
+    mesh: jax.devices() mesh
+
+  Returns:
+    state: state with decode params loaded from the NNX checkpoint
+    state_mesh_annotations: the mesh annotations for the state
+  """
+  from flax import nnx  # pylint: disable=import-outside-toplevel
+  from maxtext.utils.model_creation_utils import create_nnx_model  # pylint: disable=import-outside-toplevel
```
shuningjin: nit: wondering why not import at top level?

shuningjin (approved):
Thanks, LGTM!

tests/utils/forward_pass_logit_checker.py
shuningjin: My impression is that `enable_nnx=true`([src1](https://github.com/AI-Hypercomputer/maxtext/blob/3971206a2a8cedd052c930195aef4cccbe425b4e/src/maxtext/models/models.py#L247)) might be different from `create_nnx_model` ([src2](https://github.com/AI-Hypercomputer/maxtext/blob/3971206a2a8cedd052c930195aef4cccbe425b4e/src/maxtext/utils/model_creation_utils.py#L115)). Overriding the meaning might cause confusion.

src/maxtext/utils/maxtext_utils.py
hengtaoguo: Thanks for the suggestion! Removed this `device_put` and now the function `from_pretrained` (formerly create_nnx_model) will handle the sharding by default.

src/maxtext/utils/maxtext_utils.py
hengtaoguo: Thanks for pointing out! I feel these two dependencies are only used within the function `setup_decode_state_from_nnx`. And the AI suggests: 

> If `maxtext_utils` were to import model_creation_utils at file level in return, Python would hit a circular import error at module load time. The function-level import sidesteps this because it only runs when setup_decode_state_from_nnx is actually called, by which point both modules are fully initialized.

hengtaoguo: Actually although I could run this new feature on VM, g3 is complaining about cycle graph imports. So I moved this function into model_creation_utils (nnx related). Now all the imports are top-level, thanks!

PR #3193, opened 2026-02-19
Workflow to save top-k teacher logits in GCS to use in distillaiton

ajkv-google:
# Description

This PR introduces a standalone script (`save_top_k_teacher_logits.py`) to generate, extract, and save top-K logits and their corresponding vocabulary indices from a teacher model directly to a GCS bucket for offline distillation. In the current form of distillation, both the teacher and student models are loaded together into memory, and the process of distillation is done in an "online" fashion. This can lead to slower training times due to computing the teacher's probability distribution on the fly and training the student model simultaneously. In this change, we are saving the top-k teacher outputs in a GCS bucket (instead of the full probability distribution), so that this can be used to train the student model more efficiently without having to use extra memory to compute teacher logits on the fly. 

# Tests

Distillation YAML used to run this script: [https://paste.googleplex.com/5736358690816000](https://paste.googleplex.com/5736358690816000)

Ran the following command from the head of the Maxtext directory:
`python3 src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py src/maxtext/configs/post_train/distillation.yml`

Terminal output for the above command (truncated due to size): [https://paste.googleplex.com/4609623547052032](https://paste.googleplex.com/4609623547052032)

Verified that the ArrayRecord file has been written to a GS bucket in the correct path.  

# Checklist

### DIFF
--- a/src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
+++ b/src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
@@ -1,3 +1,17 @@
+# Copyright 2023–2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 This module provides functionality to save top-k teacher logits
 for distillation purposes in MaxText.
--- a/src/maxtext/trainers/post_train/distillation/verify_saved_logits.py
+++ b/src/maxtext/trainers/post_train/distillation/verify_saved_logits.py
@@ -1,3 +1,17 @@
+# Copyright 2023–2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 Verification script to check the correctness of saved top-k teacher logits.
 


src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,107 @@
...
+        tokens = batch["inputs"]
+        
+        logits = teacher_model(
+            decoder_input_tokens=tokens,
+            decoder_positions=batch["inputs_position"],
+            enable_dropout=False 
+        )
+        
+        # determine top-k size and extract top-k logits and indices
+        k_val = config.decode_sampling_top_k if config.decode_sampling_top_k > 0 else 128
```
gagika: Could you make k_val a command line argument for this script similar to `--lazy_load_tensors` for src/maxtext/checkpoint_conversion/to_maxtext.py

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,107 @@
...
+def generate_and_save_data(config):
+    """Generates top-k logits from the teacher model and saves them to an ArrayRecord file"""
+    devices = jax.devices()
+    devices_array = maxtext_utils.create_device_mesh(config, devices)
+    mesh = jax.sharding.Mesh(devices_array, config.mesh_axes)
+    
+    # Loading teacher model and dataset iterator
+    max_logging.log(f"Loading Teacher Model from {config.load_parameters_path}...")
+    teacher_model, _ = model_creation_utils.create_nnx_model(config, mesh=mesh)
+    train_iter, _ = input_pipeline_interface.create_data_iterator(config, mesh)
```
gagika: For long running jobs, we will want to checkpoint this iterator not to start over. It can be a follow up PR.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,107 @@
...
+        gathered_tokens = multihost_utils.process_allgather(tokens, tiled=True)
+        
+        if process_index == 0:
+            # Writing the gathered tokens, top-k logits, and top-k indices to the ArrayRecord file
+            record_dict = {
+                "tokens": np.array(gathered_tokens),
+                "top_k_logits": np.array(gathered_vals),
+                "top_k_indices": np.array(gathered_idx)
+            }
+            writer.write(pickle.dumps(record_dict))
```
gagika: For this data to work downstream in train_distill.py, especially on packed or Supervised Fine-Tuning (SFT) datasets, we need to save the packing and padding metadata. Without targets_segmentation, the student model will incorrectly calculate loss on padding and SFT prompts. Without inputs_segmentation, packed documents will cross-attend.

Could we update the dictionary construction to safely include these keys if they exist in the batch?

Something like this:

```
    optional_keys = [
        "inputs_position",     # Needed for positional embeddings
        "inputs_segmentation", # Needed for block-diagonal attention in packed sequences
        "targets_segmentation",# Needed to mask SFT prompts and padding during loss calculation
        "targets"              # Needed for standard shifted labels
    ]
    
    for key in optional_keys:
        if key in batch:
            gathered_meta = multihost_utils.process_allgather(batch[key], tiled=True)
            record_dict[key] = np.array(gathered_meta)
            
    #  Write to ArrayRecord
    writer.write(pickle.dumps(record_dict))
```

please check if any other fields might be missing / needed?

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,127 @@
...
+  local_temp_file = "/tmp/teacher_top_k.array_record"
+
+  writer = None
+  if process_index == 0:
+    max_logging.log(f"Opening local ArrayRecordWriter at {local_temp_file}")
+    writer = array_record_module.ArrayRecordWriter(local_temp_file, "group_size:1000")
+
+  max_logging.log(f"Starting Top-K generation loop for {config.steps} steps...")
+  for step, batch in enumerate(train_iter):
+    if step >= config.steps:
```
vlad-karp: it is better to use enumerate(islice(train_iter, config.steps)) and avoid "if" entirely

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # determine top-k size and extract top-k logits and indices
+    k_val = config.decode_sampling_top_k if config.decode_sampling_top_k > 0 else 128
+    top_k_vals, top_k_idx = get_top_k_logits(logits, k=k_val)
+
+    gathered_vals = multihost_utils.process_allgather(top_k_vals, tiled=True)
+    gathered_idx = multihost_utils.process_allgather(top_k_idx, tiled=True)
+    gathered_tokens = multihost_utils.process_allgather(tokens, tiled=True)
+
+    optional_keys = ["inputs_position", "inputs_segmentation", "targets_segmentation", "targets"]
```
vlad-karp: hardcode, let's put it into config?

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,127 @@
...
+
+    gathered_optionals = {}
+    for key in optional_keys:
+      if key in batch:
+        gathered_optionals[key] = multihost_utils.process_allgather(batch[key], tiled=True)
```
vlad-karp: gathered_optionals = { key: multihost_utils.process_allgather(batch[key], tiled=True) for key in optional_keys if key in batch}

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: I added top_k as a command line arg

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: I agree, sure we can make this a follow up PR for this task.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: This is a good point, we need to save the packing and padding metadata as well. I've implemented this in the code change, and I believe the fields that you provided are good.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: I made this change, it looks to be cleaner and avoid the if-statement.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: I believe it was mentioned that the config was getting too big. I could put it in the config to avoid hard-coding or change it to be a command line arg. @gagika can you weigh in on this?

vlad-karp: big config is still a way better that hardcoded strings list 
at the very least you can move it to some static module-level variables

gagika: is the reference of using those as string list?

`optional_keys = ["inputs_position", "inputs_segmentation", "targets_segmentation", "targets"]`

I don't think those are config base, and used in maxtext as strings literals.

Later we could create string constants and use those instead as a refactoring, but they are used throughout the codebase.

ajkv-google: For now, I moved it into the config so that the code doesn't contain any hardcoded lists of strings. This makes it easier to work with (simple modifications in the config rather than changing the code).

src/maxtext/configs/types.py
```diff
@@ -1809,7 +1815,6 @@ class MaxTextConfig(
     # Reinforcement Learning
     RLHardware,
     VLLM,
-    RL,
```
gagika, line 1894: was this intentional?

src/maxtext/configs/types.py
```diff
@@ -1053,6 +1053,12 @@ class Distillation(BaseModel):
   distill_alpha: float = Field(0.5, description="Weight for the distillation loss component.")
   distill_temperature: float = Field(1.0, description="Temperature for distillation softening.")
 
+  # --- Teacher topk distillation ---
+  teacher_logits_optional_keys: list[str] = Field(
+      default=["inputs_position", "inputs_segmentation", "targets_segmentation", "targets"],
+      description="Optional keys to save from teacher logits",
+  )
```
gagika: as those are only used in save_top_k_teacher_logits.py script (for now), should we just keep it arguments / or constants in save_top_k_teacher_logits.py script or under distillation folder?

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,131 @@
...
+      record_dict = {
+          "tokens": np.array(gathered_tokens),
+          "top_k_logits": np.array(gathered_vals),
+          "top_k_indices": np.array(gathered_idx),
+      }
+
+      for key, gathered_val in gathered_optionals.items():
+        record_dict[key] = np.array(gathered_val)
+
+      writer.write(pickle.dumps(record_dict))
```
gagika, line 114: do we need to use pickle.dump here?

@aireenmei for writing in array record format what's best practice?

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,131 @@
+"""
+This module provides functionality to save top-k teacher logits
+for distillation purposes in MaxText.
+"""
```
gagika, line 24: Could you add an example command here?

python3 src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py src/maxtext/configs/post_train/distillation.yml max_target_length=8192 tokenizer_type=huggingface tokenizer=...  --top_k=128

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
```diff
@@ -0,0 +1,131 @@
...
+def generate_and_save_data(config, k_val):
+  """Generates top-k logits from the teacher model and saves them to an ArrayRecord file"""
+  devices = jax.devices()
+  devices_array = maxtext_utils.create_device_mesh(config, devices)
+  mesh = jax.sharding.Mesh(devices_array, config.mesh_axes)
+
+  # Loading teacher model and dataset iterator
+  max_logging.log(f"Loading Teacher Model from {config.load_parameters_path}...")
+  teacher_model, _ = model_creation_utils.create_nnx_model(config, mesh=mesh)
+  train_iter, _ = input_pipeline_interface.create_data_iterator(config, mesh)
```
gagika, line 69: if train_iter is grain based, let's add a train iterator checkpointing / restore logic too
This can be a follow up PR.

src/maxtext/configs/types.py
ajkv-google: No, that was not. I restored this back to the original state.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: Sure, i added the example command that I used to run this workflow.

src/maxtext/configs/types.py
ajkv-google: Yes, since we are just using these optional keys in save_top_k_teacher_logits.py, I moved them out of types.py and made them a cmd argument.  It will default to ["inputs_position", "inputs_segmentation", "targets_segmentation", "targets"], but we can override it in the command line with anything else that we want to add.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: Yes, that is a good point. This can be a follow up PR so that this PR is not too heavy. I will work to do that once this PR gets merged.

src/maxtext/trainers/post_train/distillation/save_top_k_teacher_logits.py
ajkv-google: @aireenmei @gagika just following up on this

PR #3195, opened 2026-02-19
[Deprecated] Enable Indexer cache for DS v3.2 decoding

RissyRan:
# Description

Enable Indexer cache for DS v3.2 decoding, to unblock the eval benchmark for DS v3.2 model with sparse attention bringup.
* DS reference implementation for Indexer is [here](https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/model.py#L435-L495)
* Add `init_indexer_cache` & `update_indexer_cache` for indexer cache
* Other small changes, including formatting to pass the test (needed for `src/maxtext/layers/attention_op.py`) - sorry I have to rebase to resolve the conflict

# Tests

* All runners are green
* Training end-to-end (no impact) with a smaller model version: [link](https://paste.googleplex.com/4878875365605376)
* Test against reference implementation still green: [link](https://paste.googleplex.com/5361978873872384)
* Decoding gives reasonable 
  * small seq len to skip the indexer: [link](https://screenshot.googleplex.com/6hPKqZmCz47jUNZ)
  * large seq len to process the indexer `max_prefill_predict_length=3072 max_target_length=4096`: [link](https://screenshot.googleplex.com/5H9uqCovQ5d8vMy)

# Checklist

### DIFF
--- a/src/maxtext/layers/attention_mla.py
+++ b/src/maxtext/layers/attention_mla.py
@@ -38,6 +38,11 @@ from maxtext.common.common_types import (
     AxisNames,
     BATCH,
     BATCH_NO_EXP,
+    CACHE_BATCH,
+    CACHE_BATCH_PREFILL,
+    CACHE_SEQUENCE,
+    CACHE_HEADS_NONE,
+    CACHE_KV,
     Config,
     DECODE_BATCH,
     DECODE_LENGTH,
@@ -75,6 +80,9 @@ from maxtext.inference.kvcache import KVQuant
 from maxtext.utils.sharding import create_sharding
 
 
+PLACEHOLDER_SEQ_LEN = 1
+
+
 class Indexer(nnx.Module):
   """Indexer for DeepSeek Sparse Attention (DSA).
 
@@ -108,6 +116,7 @@ class Indexer(nnx.Module):
     self.rngs = rngs
     self.dtype = config.dtype
     self.weight_dtype = config.weight_dtype
+    self.max_target_length = config.max_target_length
 
     self.n_heads = config.index_n_heads
     self.head_dim = config.index_head_dim
@@ -167,6 +176,31 @@ class Indexer(nnx.Module):
         rngs=self.rngs,
     )
 
+  def update_indexer_cache(self, kv_cache, k, decoder_segment_ids, model_mode, previous_chunk):
+    """Updates Indexer buffers by processing KV cache results."""
+    k_expanded = k[:, :, jnp.newaxis, :]
+    p_res, a_res = kv_cache(
+        key=k_expanded,
+        value=k_expanded,
+        decoder_segment_ids=decoder_segment_ids,
+        model_mode=model_mode,
+        use_ragged_attention=self.config.use_ragged_attention,
+        previous_chunk=previous_chunk,
+    )
+
+    # Filter out None values to handle PREFILL vs AR modes uniformly
+    active_results = [res for res in [p_res, a_res] if res is not None]
+
+    if not active_results:
+      return None, None
+
+    # Extract keys (index 0) and segment IDs (index 2)
+    keys = jnp.concatenate([res[0] for res in active_results], axis=1)
+    segs = jnp.concatenate([res[2] for res in active_results], axis=1)
+
+    # squeeze(2) removes the jnp.newaxis added above
+    return keys.squeeze(2), segs
+
   def apply_partial_rope(
       self,
       inputs: Array,
@@ -220,6 +254,10 @@ class Indexer(nnx.Module):
       inputs_kv: Array,
       inputs_positions: Optional[Array | None] = None,
       attention_mask: Optional[Array | None] = None,
+      decoder_segment_ids: Optional[Array | None] = None,
+      previous_chunk: Any = None,
+      kv_cache: Any = None,
+      model_mode: str = MODEL_MODE_TRAIN,
   ):
     """Computes the index score to determine the top-k relevant tokens.
 
@@ -244,6 +282,10 @@ class Indexer(nnx.Module):
         `DEFAULT_MASK_VALUE` (a large negative number) prevent it.
         Returns `None` if no masking is determined to be necessary based on
         the inputs and configuration.
+      decoder_segment_ids: Segment IDs for decoder masking.
+      previous_chunk: Previous chunk info for prefill.
+      kv_cache: Key-value cache used when serving models.
+      model_mode: "train", "prefill", or "autoregressive".
 
     Returns:
       index_mask: A sparse mask [b, t, s] with 0.0 for top-k selected tokens
@@ -258,10 +300,6 @@ class Indexer(nnx.Module):
       h: Number of Indexer Heads (index_n_heads)
       d: Indexer Head Dimension (index_head_dim)
     """
-    # NOTE: If sequence length <= topk, indexer always selects all tokens.
-    if self.config.max_target_length <= self.index_topk:
-      return None, None, None
-
     bsz, seqlen, _ = inputs_q.shape  # s = t = seqlen
 
     # Query Processing: Project from Latent low_rank_q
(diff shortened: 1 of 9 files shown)


shuningjin (commented):
Thank you! There were some attention autoregressive/decoding tests:
- mha: https://github.com/AI-Hypercomputer/maxtext/blob/3fbe3be9830943857b6ded2f031588c85afaaea5/tests/unit/attention_test.py#L371
- mla: https://github.com/AI-Hypercomputer/maxtext/blob/3fbe3be9830943857b6ded2f031588c85afaaea5/tests/unit/attention_test.py#L1195

Shall we add a similar one to test indexer cache in mla? [example from gemini](https://paste.googleplex.com/5902350570094592)

RissyRan:
> Thank you! There were some attention autoregressive/decoding tests:
> 
> * mha: https://github.com/AI-Hypercomputer/maxtext/blob/3fbe3be9830943857b6ded2f031588c85afaaea5/tests/unit/attention_test.py#L371
> * mla: https://github.com/AI-Hypercomputer/maxtext/blob/3fbe3be9830943857b6ded2f031588c85afaaea5/tests/unit/attention_test.py#L1195
> 
> Shall we add a similar one to test indexer cache in mla? [example from gemini](https://paste.googleplex.com/5902350570094592)

Sounds good. Also enabled the mla assertion, and did some sanity check for decoding: [long seq](https://screenshot.googleplex.com/3JNUxh32MXmBboW), [short seq](https://screenshot.googleplex.com/4evUZuChVnZJmJf)

src/maxtext/layers/attention_mla.py
```diff
@@ -167,6 +176,31 @@ def __init__(
...
     )
 
+  def update_indexer_cache(self, kv_cache, k, decoder_segment_ids, model_mode, previous_chunk):
+    """Updates Indexer buffers by processing KV cache results."""
+    k_expanded = k[:, :, jnp.newaxis, :]
+    p_res, a_res = kv_cache(
+        key=k_expanded,
+        value=k_expanded,
+        decoder_segment_ids=decoder_segment_ids,
+        model_mode=model_mode,
```
Rohan-Bierneni, line 186: IIUC, a property of the indexer is that it only requires storage for the keys. 

However, we are initializing the indexer cache via KVCache, which stores keys and values. Passing `k_expanded` as the value doubles the HBM footprint. Is it possible to pass in a dummy value/tensor. Maybe of shape [B, S, 1, 1] for the value. 

Moreover, to accomodate the usage of KVCache for indexer class, we artificially unsqueeze `k` to 4D to satisfy the generic API, only to squeeze it back to 3D upon retrieval. This operation is redundant but would still work functionally.

Maybe a comment we can put here as a TODO is if a dedicated SparseKVCache class were written explicitly for the Indexer, we can drop the head dimension entirely and simplify the computation graph.

src/maxtext/layers/attention_mla.py
```diff
@@ -615,16 +672,47 @@ def __init__(
...
     # Module attribute names must match names previously passed to Linen for checkpointing
     self.MlaKVCache_0 = self.init_mla_kv_caches(inputs_kv_shape) if model_mode != MODEL_MODE_TRAIN else None
 
+  def init_indexer_cache(self, inputs_kv_shape: Tuple):
+    """Initializes Indexer Cache."""
+    batch_size, _, _ = inputs_kv_shape
+    # Use standard KVCache to store keys. Values are unused but required by KVCache API.
+    # KVCache expects key_heads and value_heads. Since k is shared (MQA-like for Indexer),
+    # we use key_heads=1, value_heads=1.
+    return kvcache.KVCache(
```
Rohan-Bierneni, line 695: Looking at `MLA.__init__`, it sets up a `PagedAttentionOp` for the main query, key, and value computations if paged attention is enabled. However, the Indexer cache explicitly relies on the standard, dense kvcache.KVCache which would be incompatible for paged attention.

We should add a comment that paged attention is not supported with Sparse Attention.

src/maxtext/layers/attention_mla.py
```diff
@@ -276,6 +314,16 @@ def __call__(
...
     k = k.squeeze(2)  # [b, s, 1, d] -> [b, s, d]
 
+    # Update and retrieve from cache if not training
+    cached_s = None
+    if model_mode != MODEL_MODE_TRAIN:
+      k_cached, cached_s = self.update_indexer_cache(kv_cache, k, decoder_segment_ids, model_mode, previous_chunk)
+      k = k_cached if k_cached is not None else k
+
+    # NOTE: If the total available sequence length <= topk, indexer always selects all tokens.
+    if k.shape[1] <= self.index_topk:
```
Rohan-Bierneni, line 324: For this check, doesn't k.shape[1] always equal to max_prefill_predict_length. Since JAX uses static shapes unlike Pytorch, then this condition would evaluate to False anytime max_prefill_predict_length > top_k value. 

However, the ideal behavior would be as decoding runs, the condition should check against the current token number that increases by 1 every decode step. 

I asked gemini to generate jax code for this check and gave this:
```
# 1. Dynamically count active tokens
# cached_s has > 0 for valid tokens, 0 for padding
active_token_count = jnp.sum(cached_s > 0, axis=-1) 

# 2. Define the heavy compute path
def compute_sparse_mask():
    logits = jnp.einsum("bthd, bsd -> btsh", q, k)
    # ... (rest of the heavy einsum and top_k math) ...
    return index_mask, topk_indices, index_score

# 3. Define the bypass path (Dense Attention)
def bypass_sparse_mask():
    # If we have fewer tokens than top-k, we just want standard dense attention.
    # Dense attention is achieved by simply returning the padding mask!
    index_mask = jnp.where(cached_s > 0, 0.0, DEFAULT_MASK_VALUE)
    
    # Return dummy tensors for the unused outputs to keep JAX happy
    dummy_topk = jnp.zeros((bsz, seqlen, self.index_topk), dtype=jnp.int32)
    dummy_score = jnp.zeros((bsz, seqlen, k.shape[1]), dtype=self.dtype)
    return index_mask, dummy_topk, dummy_score

# 4. Use jax.lax.cond to execute only ONE of the paths
index_mask, topk_indices, index_score = jax.lax.cond(
    (active_token_count > self.index_topk).any(), # Condition
    compute_sparse_mask,                          # Runs if True
    bypass_sparse_mask                            # Runs if False
)
```

Rohan-Bierneni (commented):
Thank you for this PR! I have added a few comments PTAL.

src/maxtext/layers/attention_mla.py
shuningjin: We could focus on functionality now, and optimize as a followup. Currently, the logic seems correct to me
```
train, k.shape[1] = max_target_length
prefill, k.shape[1] = max_prefill_length
autoregressive, k.shape[1] = max_target_length
```

e.g., I tried
```
python3 -m maxtext.inference.decode src/maxtext/configs/base.yml \
model_name=deepseek3.2-671b \
override_model_config=true base_num_decoder_layers=4 first_num_dense_layers=2 \
attention=dot_product scan_layers=false sparse_matmul=false \
dtype=bfloat16 weight_dtype=bfloat16 \
per_device_batch_size=1 max_prefill_predict_length=512 max_target_length=1024 mla_naive_kvcache=false 'prompt=An attention function can be described as mapping a query and a set of key-value pairs to an output, where the query, keys, values, and outputs are all vectors. The output is '
```
log shows `train: 1024, prefill: 512, autoregressive: 1024`

src/maxtext/layers/attention_mla.py
shuningjin: We could focus on functionality now, and optimize as a followup, either via maxtext or vLLM.

Similar KV cache optimization: In the other PR which set value = key, we could similarly reduce KV cache to K cache only.  Their solution is to defer the decoding optimization to vLLM: https://github.com/AI-Hypercomputer/maxtext/pull/3283#issuecomment-3994572628

src/maxtext/layers/attention_mla.py
RissyRan: Yeah, I think this is included in this code block? For paged attention, we need to set `attention=paged` to enable it.

```
    if self.use_indexer:
      if self.q_lora_rank == 0:
        raise NotImplementedError("Sparse indexer has not implemented for q_lora_rank = 0.")
      supports_dot_product = self.attention == "dot_product"
      supports_flash_splash = self.attention == "flash" and self.use_tokamax_splash
      if not (supports_dot_product or supports_flash_splash):
        raise NotImplementedError(
            "Sparse indexer is only supported dot_product attention or flash attention with tokamax splash."
        )
```

src/maxtext/layers/attention_mla.py
RissyRan: Yeah, let's make it functional first. We could collaborate with inference team and make it more efficient.

src/maxtext/layers/attention_mla.py
RissyRan: Please let me know if still has questions or concerns.

PR #3235, opened 2026-02-24
Fix Pathways Elastic Training Goodput Gaps

dipannita08:
# Description

When Pathways elastic training catches a JAX error and triggers a restart, a `job_end_time` entry was being incorrectly written to the goodput log. Occurance of a `job_end_time` prior to the restart is a condition that the library doesn't support (similar to re-using a workload_id for a new unrelated run). This results in undetermined behavior, sometimes resulting in inability to calculate metrics and skipping uploads., which skews goodput and badput metrics for the whole workload.

Changes:
- Job start and end times are now recorded explicitly instead of automatically by a context manager. The `job_end_time` entry is only written when training genuinely completes. Any other exit, including elastic restarts, leaves no end-time log.
- This fix is applied consistently across all trainer entry points: pre-training, SFT (both variants), and GRPO.
- The goodput context manager used for exisiting events (device initialization, data loading, training preparation, step timing) is unchanged.
- Unit tests added to cover graceful completion, elastic restart, and constant correctness.
- Named constants (`RECORD_JOB_START_TIME`, `RECORD_JOB_END_TIME`) replace raw strings so the recorder method names are defined in one place.

FIXES: b/486232512, b/489779448

# Tests
- E2E MaxText run with simulated restarts
- https://screenshot.googleplex.com/B6iZe3bULyfH4Jc
- 
image

- Unit Tests
- Github CI

# Checklist


igorts-git (approved):
Thanks for adding me to the CODEOWNERS!

PR #3247, opened 2026-02-25
Fix broken links and formatting for documentation

melissawm:
# Description

Also adds the API documentation pages to the Table of Contents so it's more discoverable.

# Tests

Docs are built successfully localy.

# Checklist

- [N/A] I have necessary comments in my code, particularly in hard-to-understand areas.


melissawm:
I'll rebase in a second.

melissawm:
Should be good to go now!

docs/development.md
```diff
@@ -9,22 +9,22 @@ The MaxText documentation website is built using [Sphinx](https://www.sphinx-doc
...
-pip install -r src/dependencies/requirements/requirements_docs.txt
+uv pip install -r src/dependencies/requirements/requirements_docs.txt
 ```
 
-Once the dependencies are installed, you can navigate to the `docs/` folder and run:
+Once the dependencies are installed and your `maxtext_venv` virtual environment is activated, you can navigate to the `docs/` folder and run:
 
 ```bash
-sphinx-build -b html . _build/html
+uv run sphinx-build -b html . _build/html
```
igorts-git: I don't think that `uv run` works with `seed_env` and `uv pip install` flow. It requires a uv-managed project created using `uv init`, which we currently don't instruct our users to do.
Just recently I attempted to support `uv init` in https://github.com/AI-Hypercomputer/maxtext/pull/3474. 
I think that the correct thing to do here is to revert to calling `sphinx-build` command directly.

With that said, I am not an expert on `uv` and might be wrong.

melissawm:
Yes - these are internal docs that were not included in the toctrees (like the API documentation) or that had been moved. I believe you mean the linkcheck here (https://github.com/AI-Hypercomputer/maxtext/blob/09309ff63762933a991aa2008a111d23dcc63823/.github/workflows/check_docs_build.yml#L58) which checks for broken *external* links.

docs/development.md
```diff
@@ -9,22 +9,22 @@ The MaxText documentation website is built using [Sphinx](https://www.sphinx-doc
...
 ```
 
 This will generate the documentation in the `docs/_build/html` directory. These files can be opened in a web browser directly, or you can use a simple HTTP server to serve the files. For example, you can run:
 
 ```bash
-python -m http.server -d _build/html
+uv run python -m http.server -d _build/html
```
igorts-git: ditto

igorts-git (approved):

docs/development.md
melissawm: You are right, I am building locally without the `uv run` part. Let me fix.

PR #3254, opened 2026-02-26
Fix `KeyError` by providing fallback for USER environment variable

RUEI4341:
# Description

This PR addresses a `KeyError: 'USER'` that occurs when Python scripts run in environments (such as certain Kubernetes configurations) where the `USER` environment variable is not explicitly defined.

# Changes:
 - Updated `generate_xpk_workload_cmd` and `xpk_benchmark_runner` to provide a resilient default value for the user argument.
 - Ensures the workload command generation does not fail if the environment shell is restricted.

FIXES: b/486699931

# Tests

https://paste.googleplex.com/4652104816918528

# Checklist

### DIFF
--- a/benchmarks/maxtext_xpk_runner.py
+++ b/benchmarks/maxtext_xpk_runner.py
@@ -586,12 +586,13 @@ def generate_xpk_workload_cmd(
     cluster_config: XpkClusterConfig,
     wl_config: WorkloadConfig,
     workload_name=None,
-    user=os.environ["USER"],
+    user=None,
     temp_key=None,
     exp_name=None,
 ):
   """Generates a command to run a maxtext model on XPK."""
 
+  user = user or os.environ.get("USER", "user")
   is_pathways_enabled = wl_config.pathways_config is not None
   is_pathways_headless_enabled = wl_config.pathways_config and wl_config.pathways_config.headless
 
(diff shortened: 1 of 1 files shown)


benchmarks/maxtext_xpk_runner.py
```diff
@@ -586,12 +586,18 @@ def generate_xpk_workload_cmd(
...
-    user=os.environ["USER"],
+    user=None,
     temp_key=None,
     exp_name=None,
 ):
   """Generates a command to run a maxtext model on XPK."""
 
+  if user is None:
+    user = os.environ["USER"]
+    if user is None:
```
igorts-git: Can this condition ever trigger?
If $USER is not set, then the call to `os.environ["USER"]` will raise KeyError.
I would write these lines as:

```
user = user or os.environ.get("USER", "user")
```

benchmarks/maxtext_xpk_runner.py
```diff
@@ -738,6 +744,12 @@ def xpk_benchmark_runner(
   Returns:
     The DisruptionManager instance, potentially updated with new workloads.
   """
+  if user is None:
+    user = os.environ["USER"]
+    if user is None:
```
igorts-git: Same here

igorts-git:
I am not convinced that the paste in the "Tests" section: https://paste.googleplex.com/4652104816918528 actually tested this case, because there is no printout in the logs "WARNING: USER environment variable not set".

PR #3257, opened 2026-02-26
Create initial tpu-post-train installation option

bvandermoon:
# Description

* Create initial installation option for installing `tpu-post-train` dependencies
* To get started, the generated requirements currently match the base tpu ones exactly. We will update them to the correct post-train dependencies after this

# Tests

```
uv pip install -e .[tpu-post-train] --resolution=lowest
```

This installation option is not included in PR tests at this point. Just getting the initial setup, it is not used anywhere so it shouldn't impact other places

# Checklist


dependencies/requirements/generated_requirements/tpu-post-train-requirements.txt
```diff
@@ -0,0 +1,250 @@
...
+opt-einsum>=3.4.0
+optax>=0.2.6
+optree>=0.18.0
+optype>=0.14.0
+orbax-checkpoint>=0.11.28
+packaging>=25.0
+pandas>=2.3.3
+parameterized>=0.9.0
+pathspec>=0.12.1
+pathwaysutils>=0.1.3
```
NicoGrande, line 157: Can we bump this up to 0.1.4 which we need for resharding? Or will this be upgraded in a later step?

bvandermoon: Yes, we will bump it at a later step. Keeping everything the same for this PR to just get it setup. Let me know if you have any concerns with this

bvandermoon: My understanding is that we actually cannot just bump the version here. The current setup has a [build_hook](https://github.com/AI-Hypercomputer/maxtext/blob/main/build_hooks.py) that inherits from the base tpu requirements. So we need to bump both or turn that feature off (we will likely do both of these)

NicoGrande (approved):

PR #3275, opened 2026-02-27
Add a harder testcase for logit check

JamesDeng42:
# Description

This introduces a more complex test case to compare logits between two models during the forward pass. Previous tests were too weak and could potentially mask subtle differences. This stricter test ensures any discrepancies are reliably exposed.

# Tests
Tested locally.

# Checklist

### DIFF
--- a/tests/utils/forward_pass_logit_checker.py
+++ b/tests/utils/forward_pass_logit_checker.py
@@ -427,7 +427,13 @@ def main(config, test_args):  # pylint: disable=W0621
     maxtext_model = models.transformer_as_linen(config, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
     maxtext_state, _ = maxtext_utils.setup_decode_state(maxtext_model, config, rng1, mesh, None)
 
-    prompts = ["I love to", "Today is a", "What is the"]
+    prompts = [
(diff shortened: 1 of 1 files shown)


vlad-karp:
@JamesDeng42 , what about a sequence of length which is close to the limit of some popular default model ?

JamesDeng42:
> @JamesDeng42 , what about a sequence of length which is close to the limit of some popular default model ?

JamesDeng42:
> @JamesDeng42 , what about a sequence of length which is close to the limit of some popular default model ?

I think we need a better hard case especially for DS v3 and Qwen3 which are in our target. Let me think about a better way to do this. Maybe a hard case for 8B model and even harder case for Qwen3-480B-MoE if an additional flag is turn on. what do you think?

vlad-karp:
> > @JamesDeng42 , what about a sequence of length which is close to the limit of some popular default model ?
> 
> I think we need a better hard case especially for DS v3 and Qwen3 which are in our target. Let me think about a better way to do this. Maybe a hard case for 8B model and even harder case for Qwen3-480B-MoE if an additional flag is turn on. what do you think?
The test is using a specific model, let's use and example of a maximum (or even exceeding) length of that model

PR #3283, opened 2026-03-02
Enable key and value projection matrix sharing

gagika:
# Description

Adds a `share_kv_projections` configuration option to the Attention layer. When enabled, it forces the Key and Value projections to share the exact same weight matrix. This reduces the model's overall parameter count and shrink the KV cache footprint (trading off some representational capacity).

**Implementation Details:**
* **Config:** Added `share_kv_projections` to `base.yml` and `types.py` with validation safeguards (incompatible with `fused_qkv`, `attention_type='mla'`).
* **Attention Layer:** Skips initialization of the `value` projection when enabled. Reuses the evaluated `key` tensor (`value = key`) during the forward pass to avoid duplicating the dense matrix multiplication.

# Tests
* **Unit Tests:** Added `test_share_kv_projections` to verify that the `value` parameter is successfully omitted from the NNX state tree. 
* **Equivalence:** Asserted mathematical equivalence (`jax.numpy.allclose`) between the shared layer and an unshared layer using copied weights.
* **Kernels:** Extended `test_tpu_kernel_attention_mha` and `test_tpu_kernel_attention_gqa` to ensure compatibility with TPU flash attention.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -331,6 +331,7 @@ param_scan_axis: 1
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+share_kv_projections: False # Note: Not compatible with attention_type='mla'
 attention_bias: False # If True, adds a learnable bias to the query, key, and value projections
 attention_sink: False
 sliding_window_size: 0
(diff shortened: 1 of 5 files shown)


gagika:
> LGTM! Minor comments (non-blocking):
> 
> 1. since `value = key`, may want to test sharding. At a glance, original key sharding and value sharding seems identical, probably fine.
> 
> > This reduces the model's overall parameter count and shrink the KV cache footprint
> 
> 2. kv cache is probably not reduced yet. may need additional change

1. yes, I had also verified first

2. kv cache management is done in vLLM (tpu-inference), so probably we don't need to do anything here.

cc @NicoGrande

PR #3262, opened 2026-02-26
fix completion token extraction when using chat templates

vlad-karp:
# Description

Whenever apply_chat_template() is called in the input_pipeline (SFT/RL scenario), it was assumed that the tokenizer_model.apply_chat_template() returns a single list of tokens. 
But starting from some HF version (~4.40), it already returns BatchEncoding object if you use it with tokenization enabled.

Then the logic of extracting the completion part of a conversation round appears to be broken.

This PR changes the extraction logic by handling multiple types of tokenizer_model.apply_chat_template() outputs - BatchEncoding, dict and lists(old approach).

This change is required to correctly use chat templates preprocessing for SFT training scenarios when a relatively new HF transformer's library tokenizers is used.

# Tests
python3 -m tests.integration.sft_trainer_correctness_test

This is the supposed test to check this logic, but it is disabled currently. 
I had to regenerate the gold tokens & their logits since the old HF llama2 tokenizer is not correctly applying its chat template.

If I force execution of sft_trainer_correctness_test test by adding
"""
if __name__ == "__main__":
  unittest.main()
"""
To sft_trainer_correctness_test, it passes.
If I run it via pytest it still fails for some reason so I keep it disabled.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #3288, opened 2026-03-02
fix completion token extraction when using chat templates

vlad-karp:
Whenever apply_chat_template() is called in the input_pipeline (SFT/RL scenario), it was assumed that the tokenizer_model.apply_chat_template() returns a single list of tokens.
But starting from some HF version (~4.40), it already returns BatchEncoding object if you use it with tokenization enabled.

Then the logic of extracting the completion part of a conversation round appears to be broken.

A fresh re-do of https://github.com/AI-Hypercomputer/maxtext/pull/3262

This PR changes the extraction logic by handling multiple types of tokenizer_model.apply_chat_template() outputs - BatchEncoding, dict and lists(old approach).

This change is required to correctly use chat templates preprocessing for SFT training scenarios when a relatively new HF transformer's library tokenizers is used.

Tests
python3 -m tests.integration.sft_trainer_correctness_test

This is the supposed test to check this logic, but it is disabled currently.
I had to regenerate the gold tokens & their logits since the old HF llama2 tokenizer is not correctly applying its chat template.

If I force execution of sft_trainer_correctness_test test by adding
"""
if __name__ == "__main__":
unittest.main()
"""
To sft_trainer_correctness_test, it passes.
If I run it via pytest it still fails for some reason so I keep it disabled.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ See our guide on running MaxText in decoupled mode, without any GCP dependencies
 
 ## 🔥 Latest news 🔥
 
+* \[February 27, 2026\] New MaxText structure! MaxText has been restructured according to [RESTRUCTURE.md](https://github.com/AI-Hypercomputer/maxtext/blob/1b9e38aa0a19b6018feb3aed757406126b6953a1/RESTRUCTURE.md). Please feel free to share your thoughts and feedback.
 * \[December 22, 2025\] [Muon optimizer](https://kellerjordan.github.io/posts/muon) is now supported.
 * \[December 10, 2025\] DeepSeek V3.1 is now supported. Use existing configs for [DeepSeek V3 671B](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/configs/models/deepseek3-671b.yml) and load in V3.1 checkpoint to use model.
 * \[December 9, 2025\] [New RL and SFT Notebook tutorials](https://github.com/AI-Hypercomputer/maxtext/tree/main/src/maxtext/examples) are available.
--- a/docs/reference/core_concepts/batch_size.md
+++ b/docs/reference/core_concepts/batch_size.md
@@ -1,0 +1,85 @@
+<!--
+ Copyright 2026 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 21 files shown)


entrpn (approved):

tests/assets/golden_logits/golden_data_sft_default.jsonl
```diff
@@ -1 +1 @@
...
+{"data": {"messages": [{"role": "user", "content": "Hello, what is your name?"}, {"role": "assistant", "content": "I am a chatbot. How can I help?"}]}, "tokens": [1, 29961, 25580, 29962, 15043, 29892, 825, 338, 596, 1024, 29973, 518, 29914, 25580, 29962, 306, 626, 263, 13563, 7451, 29889, 1128, 508, 306, 1371, 29973, 29871, 2], "attention_mask": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1], "token_log_probs": [-10.360702514648438, -11.012994766235352, -10.751636505126953, -9.73588752746582, -11.174783706665039, -11.906787872314453, -10.50442123413086, -11.422593116760254, -12.447595596313477, -10.885910034179688, -11.982933044433594, -10.058539390563965, -10.950790405273438, -12.060896873474121, -10.68459701538086, -11.916288375854492, -12.050270080566406, -9.983818054199219, -10.710721015930176, -9.216376304626465, -11.008810043334961, -9.728713989257812, -12.391929626464844, -11.235883712768555, -9.664995193481445, -11.548173904418945, -10.014203071594238]}
```
aireenmei, line 1: Why is there a change in token id?

vlad-karp: 518 is a whitespace which should not be there according to the chat jinja template.

tests/assets/logits_generation/generate_sft_golden_data.py tokenizer defined via meta-llama/Llama-2-7b-chat-hf returns this incorrect token while the one from MAXTEXT_ASSETS_ROOT works correctly.

gagika (approved):
thanks

PR #3316, opened 2026-03-04
run teacher/student in same jit function for performance.

entrpn:
# Description

Runs the teacher/student's forward pass in the same jitted function, removing the d2h and h2d activation transfers and thus improving performance.

# Tests

Updated tests under train_distill_test

# Checklist

### DIFF
--- a/src/maxtext/trainers/post_train/distillation/train_distill.py
+++ b/src/maxtext/trainers/post_train/distillation/train_distill.py
@@ -54,7 +54,7 @@ from maxtext.utils import maxtext_utils
 from maxtext.utils import model_creation_utils
 
 # Tunix Imports
-from tunix.distillation import distillation_trainer
+from tunix.sft import peft_trainer
 from tunix.sft import metrics_logger
 from tunix.sft import profiler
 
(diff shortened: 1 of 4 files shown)


src/maxtext/trainers/post_train/distillation/train_distill.py
```diff
@@ -173,13 +173,82 @@ def _log_config_details(config: pyconfig.HyperParameters, label: str) -> None:
...
-class MaxTextDistillationTrainer(distillation_trainer.DistillationTrainer):
+class MaxTextDistillationTrainer(peft_trainer.PeftTrainer):
   """Custom Trainer to preserve MaxText fields and log Teacher metrics.
 
   This class overrides `_prepare_inputs` to ensure MaxText-specific fields
   (positions, segment_ids) are passed to the model.
   """
 
+  def __init__(self, student_model, teacher_model, strategy, optimizer, training_config, **kwargs):
+    super().__init__(model=(student_model, teacher_model), optimizer=optimizer, training_config=training_config, **kwargs)
```
vlad-karp: can a tuple be a cause of an error in the parent PeftTrainer class when the model attribute is handled like a model? Since it is typed like "model: nnx.Module," and we send a tuple here.

vlad-karp:
In general, will this change make it easy to adopt the offline-teacher mode?

src/maxtext/trainers/post_train/distillation/train_distill.py
entrpn: This is a really good point. Jax is treating it like a pytree and that's why its not crashing, but I agree its not a good way to implement this. Let me see if I can update the tunix side to support this better.

entrpn:
> In general, will this change make it easy to adopt the offline-teacher mode?

yes it would. We would just not call the forward fn of the teacher if the teacher's outputs already exist.

src/maxtext/trainers/post_train/distillation/train_distill.py
entrpn: After taking a look at this, there is no current easier way to implement this in tunix. For now, I suggest we keep the implementation as is and revisit in the future. Because we override most methods from PeftTrainer, this solution stays viable.

PR #3323, opened 2026-03-05
Make sure that maxtext_venv does not get packaged into docker images.

igorts-git:
# Description

For customers who have the "maxtext_venv" under their "maxtext" directory we add "maxtext_venv" to ".dockerignore". This makes sure that `docker_build_dependency_image.sh` does not produce unnecessarily large docker images (5.9GB vs 18.8GB).

Also updated the file checks in the `docker_upload_runner.sh` script to respect the `.dockerignore` file and to exclude these files/dirs from the checks.

For reference, the Maxtext installation instructions (from source option) are instructing our users to create the `maxtext_venv` under the "maxtext" directory: https://maxtext.readthedocs.io/en/latest/install_maxtext.html#from-source

We could potentially save about 60MB extra space by removing unit tests, documentation, and some other files from the docker image. However, I don't think it would make a meaningful difference.

# Tests

Invoked  `bash dependencies/scripts/docker_build_dependency_image.sh DEVICE=tpu MODE=stable` on my VM to confirm that it runs much faster and the produced image size is significantly smaller.

Invoked `bash dependencies/scripts/docker_upload_runner.sh CLOUD_IMAGE_NAME=${CLOUD_IMAGE_NAME}` command to validate that its checks respect the `.dockerignore` file.

# Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 .git
+maxtext_venv
--- a/dependencies/scripts/docker_upload_runner.sh
+++ b/dependencies/scripts/docker_upload_runner.sh
@@ -56,8 +56,30 @@ if [[ ! -v CLOUD_IMAGE_NAME ]]; then
   exit 1
 fi
 
-# Check for dangling symbolic links (target does not exist).
-DANGLING_LINKS=$(find -L . -type l)
+# In the following sections we will want to check that the files that we are
+# packaging in the Docker image don't have outside symbolic links.
+# However, we want to exclude files and dirs listed in `.dockerignore` from this check
+
+# Build find exclusion arguments
+EXCLUDE_PATHS=()
+if [ -f .dockerignore ]; then
+  while IFS= read -r pattern; do
+    # Ignore empty lines and comments
+    if [[ -n "$pattern" && ! "$pattern" =~ ^# ]]; then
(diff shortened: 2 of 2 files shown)


docs/install_maxtext.md
```diff
@@ -19,25 +19,28 @@
...
 
 ```bash
 # 1. Install uv, a fast Python package installer
 pip install uv
 
 # 2. Create virtual environment
-uv venv --python 3.12 --seed maxtext_venv
-source maxtext_venv/bin/activate
+uv venv --python 3.12 --seed ~/maxtext_venv
+source ~/maxtext_venv/bin/activate
```
bvandermoon: @igorts-git thanks for this idea. We have strong guidance to not re-use the same MaxText venv across multiple workspaces. This will lead to unintended behavior for users pip installing from source with `pip install -e .` (e.g. running code from a previously installed MaxText workspace instead of the intended one) I like the idea of not packaging this into the Docker file to save space. But we still need a way to tie this to a specific MaxText workspace

igorts-git: OK, I removed the changes to the installation instructions to keep the 'maxtext_venv` under the "maxtext" directory. I also modified the `docker_upload_runner.sh` file to respect `.dockerignore`.

richjames0 (approved):
lgtm

PR #3330, opened 2026-03-06
update for group_size for tokamax ragged_dot

suexu1025:
# Description
[from tokamax team]
Use Tokamax RaggedDotGroupSizes with representative values in MoE.

This change updates the tokamax.ragged_dot calls within MaxText's MoE implementation to utilize tokamax.RaggedDotGroupSizes. This allows providing representative group sizes, which can aid in autotuning. A helper function generate_representative_group_sizes is added to create synthetic group size distributions.

TEST though internal runs.

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -891,6 +891,10 @@ class RoutedMoE(nnx.Module):
     def gmm(
         inputs, kernel, tiling, group_sizes, expert_assignments, weight_gather_axes, input_buffer_count, combine_scopes
     ):
+      tokamax_group_sizes = tokamax.RaggedDotGroupSizes(
+          group_sizes,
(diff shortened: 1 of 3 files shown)


src/maxtext/layers/moe.py
```diff
@@ -888,9 +888,22 @@ def sparse_matmul(
   ):
     """Perform sparse matrix multiplication of inputs and Experts."""
 
+    def generate_representative_group_sizes(target_m: int, g: int) -> tuple[int, ...]:
```
khatwanimohit: could we move this function to utils.py so moe.py and deepseek_batchsplit.py can just use it directly instead of writing it in both places

suexu1025: updated.

khatwanimohit (approved):

PR #3184, opened 2026-02-18
checkpoint utility: optimize to_maxtext, add deepseek

shuningjin:
# Description

**Optimize `to_maxtext` loading and saving**
- The main goal is to optimize the `to_maxtext` eager loading and saving pipelines. **By controlling data type footprint, we reduces memory and delivers speedup.**
- The optimization **unblocks large-scale model** with checkpoint conversion utility. Consequently, we are able to **onboard DeepSeek model family to utility** at 671B scale (`deepseek3.2-671b`, `deepseek3-671b`, `deepseek2-16b`).
- See this document for detailed logic and tests (http://shortn/_KlHIRwUxvI).
- Fix: memory optimization (b/452391831), speed optimization (b/477316979), ds3.2 (b/469550012, b/469550011), ds2 (b/459536844), ds3 (b/457820372, b/457820735), ds (b/452392346)

**Problem**

Previously, eager load defaulted to `transformers_class.from_pretrained(...)`, which loaded, converted, and saved checkpoints in `float32`. 
- Latency (Loading Bottleneck): The previous implementation performed an implicit conversion to float32 during the load/save cycle. Since most source tensors (e.g., in MaxText) are natively torch.bfloat16, this conversion added significant overhead to startup times.
- Memory Inefficiency: Forcing float32 doubles the memory footprint compared to bfloat16, leading to higher peak memory usage and potential OOM (Out of Memory) issues during the loading phase on constrained hardware.
- Version Incompatibility (Transformers 5.0+): While the current dependency is pinned to 4.57.3, the latest 5.0.0 release introduces a breaking change: NumPy's lack of native bfloat16 support causes end-to-end (e2e) failures when the library attempts to bridge tensor data.

**What Changed**

This PR introduces two optimized eager loading methods and adds the ability to save in `bfloat16`:
*   **Eager load Method 1:** `transformers_class.from_pretrained(..., dtype="auto")` to load the original tensor type.
*   **Eager load  Method 2:** `safetensors.safe_open(..., framework="pt")` to load natively from safetensors. Similar to Method 1, this can either process remote repo or local path.
*   **Save Dtype:** Added `bfloat16` as the recommended save option (with `float32` retained as a backup). **This works for both eager load and lazy load.**

```
# to_maxtext flag
--eager_load_method=<transformers (default) | safetensors>
--save_dtype=<bfloat16 (default) | float32>
```

**Why It Matters (Impact & Benefits)**
[...]

PR #3347, opened 2026-03-07
Fix BFloat16 conversion error in eager checkpoint loading

phu0ngng:
# Description

`to_maxtext.py` crashes with `TypeError: Got unsupported ScalarType BFloat16` when converting bf16 HuggingFace models (e.g. Qwen3, Llama 3) in eager mode. PyTorch's `.numpy()` doesn't support bfloat16 tensors. This fix casts bf16 tensors to float32 before the numpy conversion. 
The lazy loading path is unaffected (safetensors handles this internally).

# Checklist


src/maxtext/checkpoint_conversion/to_maxtext.py
```diff
@@ -595,9 +595,14 @@ def main(
     hf_config_obj = AutoConfig.from_pretrained(model_id, token=hf_token, revision=revision)
     hf_model = get_hf_model(model_id, token=hf_token, revision=revision)
     hf_state_dict_numpy = hf_model.state_dict()
-    # Convert all to numpy immediately in eager mode
+    # Convert all to numpy immediately in eager mode.
+    # torch.Tensor.numpy() does not support bfloat16, so cast to float32 first.
+    import torch  # pylint: disable=g-import-not-at-top
```
shuningjin, line 600: move this import to the top?

shuningjin (commented):
Thanks for the fix!

I believe this error is dependent on transformer version.
- Using older transformer version, `from_pretrained` by default loads bf16 ckpt in fp32. so it can pass with `v.numpy()`
- using new transformer version, it loads bf16 ckpt as is, and hence error with `v.numpy()`. Yes, your fix `v.float().numpy()` will help.

Lastly, we are working more on loading & type conversion for `to_maxtext` in https://github.com/AI-Hypercomputer/maxtext/pull/3184

src/maxtext/checkpoint_conversion/to_maxtext.py
phu0ngng: We can, but then it will require torch in the lazy loading path as well, even though torch is not needed there. I think we should go with the current implementation to avoid torch requirements for lazy load.

phu0ngng: @shuningjin what do you think?

phu0ngng: @shuningjin, friendly reminder.

shuningjin:
Thanks for your patience! We re-vamped the loading/conversion/save pipeline for `to_maxtext` in [PR 3184](https://github.com/AI-Hypercomputer/maxtext/pull/3184), with careful consideration of speed and memory.

The current hf -> maxtext behavior would be:
- eager loading (auto dtype as specified in config.json) -> conversion -> save with specified dtype (e.g., bf16 or f32, with bf16 as the recommended type to save memory). 
- A typical conversion path: `torch.bfloat16 -> torch.float32 -> numpy.float32 -> ml_dtypes.bfloat16`.

Could you verify if the latest code solve your original issue? Thanks!

PR #3351, opened 2026-03-09
Fix ring of experts using TP

NuojCheng:
# Description

When `use_ring_of_experts=True` and TP is enabled, a sharding shape error appears. This PR fixes it by adding constraint on the unpermute shapes.

FIXES: b/490177598

# Tests

CI tests

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1318,7 +1318,7 @@ class RoutedMoE(nnx.Module):
         )
 
         # Sum up the partial outputs across the expert shards.
-        output = jnp.reshape(output, (-1, sequence_length, self.config.emb_dim))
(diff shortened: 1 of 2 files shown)


src/maxtext/layers/moe.py
```diff
@@ -1318,7 +1318,7 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
         )
 
         # Sum up the partial outputs across the expert shards.
-        output = jnp.reshape(output, (-1, sequence_length, self.config.emb_dim))
+        output = jnp.reshape(output, (-1, sequence_length, self.config.emb_dim // self.get_tensor_parallelism_size()))
```
suexu1025, line 1321: Do we need manually rs gradient output?

NuojCheng: If we rely on JAX autodiff then no. If custom vjp then probably yes.

PR #3364, opened 2026-03-10
Add Indexer training loss

RissyRan:
# Description

Add Indexer training loss for DeepSeek V3.2:
* Add loss for Indexer module based on ref [paper](https://screenshot.googleplex.com/B6P6jMmC3YG2rSp) (another reference: Megatron LM - [link](https://github.com/NVIDIA/Megatron-LM/blob/e8fe06865cd49f80597cadb182d59a7a9b73de99/megatron/core/transformer/experimental_attention_variant/dsa.py#L161-L252))
* Update all `Index` to `Indexer` to align in MaxText
* For fully integration, we will need [3 stages](https://b.corp.google.com/issues/474722268#comment3), and this is the 1st one.
* Add unit tests for sanity check.

# Tests

* Expect all tests are green.
* Verify `deepseek32_vs_reference_test` same as main branch (b/491486716)
* End to end training test is functional - [link](https://paste.googleplex.com/6139527849705472)
* Tensorboard to check loss is recorded properly - [link](https://screenshot.googleplex.com/BzZGdV8vcsNJGUj)

Run cmd

```
export BASE_OUTPUT_PATH=gs://runner-maxtext-logs/no-mhc-change/$(date +%Y-%m-%d)
export RUN_NAME=ds-loss-verify

python3 -m MaxText.train maxtext/configs/base.yml base_output_directory=${BASE_OUTPUT_PATH} run_name=${RUN_NAME} per_device_batch_size=8 enable_checkpointing=false model_name=deepseek-custom ici_fsdp_parallelism=4 steps=100 max_target_length=1024 async_checkpointing=false dtype=bfloat16 weight_dtype=bfloat16 scan_layers=True attention=flash train_split=train dataset_type=hf hf_path='HuggingFaceFW/fineweb-edu' hf_name=default hf_access_token=<> enable_tensorboard=true tokenizer_type=huggingface tokenizer_path=deepseek-ai/DeepSeek-V3.2 mhc_expansion_rate=4 data_shuffle_seed=1234 scan_layers=True use_tokamax_splash=True indexer_loss_scaling_factor=0.01 log_period=5 enable_tensorboard=True
```

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -359,6 +359,13 @@ use_sparse_indexer: False
 index_head_dim: 128
 index_n_heads: 64
 index_topk: 2048
+# Determines the token selection strategy for indexer loss: 
+# - False: Uses all tokens (Dense Warm-up).
+# - True: Uses only top-k tokens (Sparse Training).
+# Note: This is only active when `indexer_loss_scaling_factor` > 0.
+sparse_indexer_loss: False
+# Multiplier for the indexer KL divergence loss
+indexer_loss_scaling_factor: 0.0
 
 # MLA parameters
 q_lora_rank: 0
--- a/src/maxtext/layers/attention_mla.py
+++ b/src/maxtext/layers/attention_mla.py
@@ -73,6 +73,7 @@ from maxtext.inference import page_manager
 from maxtext.inference import paged_attention
 from maxtext.inference.kvcache import KVQuant
 from maxtext.utils.sharding import create_sharding
+from maxtext.utils.globals import EPS
 
 
 class Indexer(nnx.Module):
@@ -246,10 +247,10 @@ class Indexer(nnx.Module):
         the inputs and configuration.
 
     Returns:
-      index_mask: A sparse mask [b, t, s] with 0.0 for top-k selected tokens
+      indexer_mask: A sparse mask [b, t, s] with 0.0 for top-k selected tokens
         and large negative values otherwise.
       topk_indices: Indices of the top-k selected tokens [b, t, k].
-      index_score: The computed relevance scores [b, t, s].
+      indexer_score: The computed relevance scores [b, t, s].
 
     Notation:
       b: Batch size
@@ -283,27 +284,27 @@ class Indexer(nnx.Module):
     logits = jax.nn.relu(logits)
     # Compute head weights: project from input, [b, t, embed_dim] -> [b, t, h]
     weights = self.weights_proj(inputs_q)
-    # Weights scaling affect index_score, but does not affect topk_indices. Keep scaling for numerical stability.
+    # Weights scaling affect indexer_score, but does not affect topk_indices. Keep scaling for numerical stability.
     # https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/87e509a2e5a100d221c97df52c6e8be7835f0057/inference/model.py#L478-L480
     weights = weights * (self.n_heads**-0.5) * self.softmax_scale
     # Aggregate head-wise logits: logits @ weights
-    index_score = jnp.einsum("btsh, bth -> bts", logits, weights, precision=self.config.matmul_precision)  # [b, t, s]
+    indexer_score = jnp.einsum("btsh, bth -> bts", logits, weights, precision=self.config.matmul_precision)  # [b, t, s]
 
     # Apply attention mask before TopK
     if attention_mask is not None:
-      index_score += attention_mask
+      indexer_score += attention_mask
 
     # TopK selection based on index score
-    _, topk_indices = jax.lax.top_k(index_score, k=self.index_topk)  # topk_indices [b, t, k]
+    _, topk_indices = jax.lax.top_k(indexer_score, k=self.index_topk)  # topk_indices [b, t, k]
 
     # Create Sparse Index Mask: 0 and large negatives
-    index_mask = self.generate_mask(topk_indices, seqlen)  # [b, t, s]
+    indexer_mask = self.generate_mask(topk_indices, seqlen)  # [b, t, s]
 
     # Re-apply attention mask after TopK: in case number of unmasked tokens < TopK
     if attention_mask is not None:
-      index_mask += attention_mask
+      indexer_mask += attention_mask
 
-    return index_mask, topk_indices, index_score
+    return indexer_mask, topk_indices, indexer_score
 
 
 def mla_as_linen(
@@ -951,6 +952,71 @@ class MLA(Attention):
 
     return key, value, cached_values
 
+  def calculate_indexer_loss(
+      self,
+      indexer_score: Array,
+      query: Array,
+      key: Array,
+      attention_mask: Optional[Array | None],
+      indexer_mask: Array,
+      sparse_loss: bool,
+      scaling_factor: float,
+  ) -> Array:
+    """Calculates the indexer KL divergence loss.
+
+    This loss trains the indexer to predict which tokens are important by matching
+    the distribution of true attention scores from the main model.
+
+    The target distribution is derived through the following steps:
+    1. Compute raw attention scores via Q @ K^T.
+    2. Aggregate scores by summing across all attention heads.
+    3. Apply L1-normalization across the sequence dimension.
+
+    target_distribution = L1_Normalize(Sum_h(Softmax(Q @ K^T)))
+
+    Reference:
+    DeepSeek-V3.2 - https://arxiv.org/pdf/2512.02556
+
+    Args:
+      indexer_score: Scores predicted by indexer [batch, q_len, kv_len].
+      query: Query tensor from main model [batch, q_len, heads, dim].
+      key: Key tensor from main model [batch, kv_len, heads, dim].
+      attention_mask: Attention mask [batch, q_len, kv_len] or None.
+      indexer_mask: Indexer mask [batch, q_len, kv_len].
+      sparse_loss: Whether to use sparse loss.
+      scaling_factor: The scaling factor for the loss.
+
+    Returns:
+      The computed KL divergence loss.
+    """
+    # Compute attention scores: [b, t, h, d] @ [b, s, h, d] -> [b, h, t, s]
+    attention_scores = jnp.einsum("bthd, bshd -> bhts", query, key, precision=self.config.matmul_precision)
+
+    if sparse_loss:
+      # indexer_mask is already pre-filtered with the attention_mask if any
+      attention_scores = attention_scores + indexer_mask[:, None, :, :]
+      indexer_score = indexer_score + indexer_mask
+    elif attention_mask is not None:
+      # indexer_score already applies attention_mask; updating attention_scores only
+      attention_scores = attention_scores + attention_mask[:, None, :, :]
+
+    # Use float32 for softmax numerical stability.
+    attention_probs = jax.nn.softmax(attention_scores.astype(jnp.float32), axis=-1)
+    indexer_probs = jax.nn.softmax(indexer_score.astype(jnp.float32), axis=-1)
+
+    # Aggregate heads: [b, h, t, s] -> [b, t, s]
+    attention_probs = jnp.sum(attention_probs, axis=1)
+    # L1 normalize aggregated target distribution
+    attention_probs = attention_probs / (jnp.sum(attention_probs, axis=-1, keepdims=True) + EPS)
+
+    # KL Divergence: KL(attention || indexer)
+    log_attention_probs = jnp.log(attention_probs + EPS)
+    log_indexer_probs = jnp.log(indexer_probs + EPS)
+    kl_per_token = attention_probs * (log_attention_probs - log_indexer_probs)
+    indexer_loss = jnp.mean(jnp.sum(kl_per_token, axis=-1))
+
+    return indexer_loss * scaling_factor
+
   def __call__(
       self,
       inputs_q: Array,
@@ -1013,22 +1079,36 @@ class MLA(Attention):
     value = checkpoint_name(value, "value_proj")
 
     # Indexer Logic
-    index_mask = None
+    indexer_mask = None
     if self.use_sparse_indexer:
       if model_mode != MODEL_MODE_TRAIN:
         raise NotImplementedError("Sparse indexer has not implemented for inference yet.")
       # generate mask: with 0 and large negative, [b, 1, 1, q_len, kv_len] -> [b, q_len, kv_len]
       attention_mask = self.attention_op.generate_attention_mask(
           query, key, decoder_segment_ids, model_mode, previous_chunk, bidirectional_mask
-      ).squeeze(axis=(1, 2))
-      # apply indexer, index_mask [b, q_len, kv_len]
-      index_mask, _, _ = self.indexer(
+      )
+      if attention_mask is not None:
+        attention_mask = attention_mask.squeeze(axis=(1, 2))
+      # apply indexer, indexer_mask [b, q_len, kv_len]
+      indexer_mask, _, indexer_score = self.indexer(
           inputs_q=inputs_q,
           low_rank_q=low_rank_q,
           inputs_kv=inputs_kv,
           inputs_positions=inputs_positions,
           attention_mask=attention_mask,
       )
+
+      if self.config.indexer_loss_scaling_factor > 0.0:
+        indexer_loss = self.calculate_indexer_loss(
+            indexer_score=indexer_score,
+            query=query,
+            key=key,
+            attention_mask=attention_mask,
+            indexer_mask=indexer_mask,
+            sparse_loss=self.config.sparse_indexer_loss,
+            scaling_factor=self.config.indexer_loss_scaling_factor,
+        )
+        self.sow(nnx.Intermediate, "indexer_loss", indexer_loss)
 
     # Check if we need QK Clip stats
     use_qk_clip = self.model_mode == MODEL_MODE_TRAIN and self.config.use_qk_clip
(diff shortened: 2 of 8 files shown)


src/maxtext/layers/attention_mla.py
```diff
@@ -951,6 +952,61 @@ def mla_kv_projection(self, inputs: Array, inputs_positions: Array, decoder_segm
...
+
+    Returns:
+      The computed KL divergence loss.
+    """
+    # Compute attention scores: [b, t, h, d] @ [b, s, h, d] -> [b, h, t, s]
+    attention_scores = jnp.einsum("bthd, bshd -> bhts", query, key, precision=self.config.matmul_precision)
+    if attention_mask is not None:
+      attention_scores = attention_scores + attention_mask[:, None, :, :]
+      # indexer_score already has attention_mask added in Indexer.__call__
+
```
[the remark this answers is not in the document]
RissyRan: Thanks for indicating the concern! I updated a little bit on if/else branch.

src/maxtext/configs/base.yml
```diff
@@ -359,6 +359,8 @@ use_sparse_indexer: False
 index_head_dim: 128
 index_n_heads: 64
 index_topk: 2048
+sparse_indexer_loss: False # Indicate if use sparse loss for indexer training.
+indexer_loss_scaling_factor: 0.0 # Scaling factor for the indexer KL divergence loss
```
shuningjin: Could you add more explanation here? For `sparse_indexer_loss`, what happens when false / true, my first impression is that false is not training indexer. 

For instance, "If false, use all tokens to calculate indexer loss (dense warm-up stage). If true, only use top-k tokens  to calculate indexer loss (sparse training stage). Only relevent when indexer_loss_scaling_factor > 0."

tests/unit/deepseek32_vs_reference_test.py
```diff
@@ -541,7 +541,7 @@ def forward(  # pylint: disable=missing-function-docstring
     # fp8_index is defined by: https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/87e509a2e5a100d221c97df52c6e8be7835f0057/inference/kernel.py#L254 # pylint: disable=line-too-long
     # Replace fp8_index with standard PyTorch: Sum_h( ReLU(Q @ K.T) * Weights
     # OLD
-    # index_score = fp8_index(
+    # indexer_score = fp8_index(
```
shuningjin: Might be good to not change variable names in the torch reference code? similar for other places above this line https://github.com/AI-Hypercomputer/maxtext/blob/0b401fa619866c0caed27c3a51db6542b058775e/tests/unit/deepseek32_vs_reference_test.py#L692

tests/unit/attention_test.py
```diff
@@ -1591,6 +1592,131 @@ def test_tpu_flash_attention_context_parallel(
...
+        key=key,
+        attention_mask=attention_mask,
+        indexer_mask=indexer_mask,
+        sparse_loss=False,
+        scaling_factor=scaling_factor,
+    )
+
+    self.assertTrue(jnp.isfinite(loss))
+
+  def test_indexer_loss_sparse(self):
```
shuningjin: could combine dense and sparse as one test function using parameterized cases, as the test logic looks the same

src/maxtext/layers/attention_mla.py
```diff
@@ -951,6 +952,62 @@ def mla_kv_projection(self, inputs: Array, inputs_positions: Array, decoder_segm
...
+      key: Array,
+      attention_mask: Optional[Array | None],
+      indexer_mask: Array,
+      sparse_loss: bool,
+      scaling_factor: float,
+  ) -> Array:
+    """Calculates the indexer KL divergence loss.
+
+    This loss trains the indexer to predict which tokens are important by matching
+    the distribution of true attention scores from the main model.
```
shuningjin, line 968: Could you add a summary how target distribution is calculated, e.g.,
```
target distribution = L1_Normalize(Sum_h(Softmax(Q @ K^T)))
```

src/maxtext/layers/attention_mla.py
```diff
@@ -951,6 +952,62 @@ def mla_kv_projection(self, inputs: Array, inputs_positions: Array, decoder_segm
...
+    indexer_probs = jax.nn.softmax(indexer_score.astype(jnp.float32), axis=-1)
+
+    # Aggregate heads: [b, h, t, s] -> [b, t, s]
+    attention_probs = jnp.sum(attention_probs, axis=1)
+    # L1 normalize aggregated target distribution
+    attention_probs = attention_probs / (jnp.sum(attention_probs, axis=-1, keepdims=True) + EPS)
+
+    # KL Divergence: KL(attention || indexer)
+    kl_per_token = jax.scipy.special.kl_div(attention_probs + EPS, indexer_probs + EPS)
+    # kl_per_token = attention_probs * (jnp.log(attention_probs + EPS) - jnp.log(indexer_probs + EPS))
```
shuningjin: maybe this could be more robust
```
log_indexer = jax.nn.log_softmax(indexer_score.astype(jnp.float32), axis=-1)
log_attention = jnp.log(attention_probs + EPS)
kl_per_token = attention_probs * (log_attention - log_indexer)
```

tests/unit/attention_test.py
```diff
@@ -1591,6 +1592,131 @@ def test_tpu_flash_attention_context_parallel(
...
+        indexer_score=indexer_score,
+        query=query,
+        key=key,
+        attention_mask=attention_mask,
+        indexer_mask=indexer_mask,
+        sparse_loss=False,
+        scaling_factor=scaling_factor,
+    )
+
+    self.assertTrue(jnp.isfinite(loss))
```
shuningjin: maybe also test > 0 as KL divergence should always positive theoretically

src/maxtext/trainers/pre_train/train.py
```diff
@@ -218,6 +218,27 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
...
+    # uses mixed scan/unscan logic, which groups layers into variable-named
+    # blocks like 'moe_layers_x_y' (where x and y denote the layer range).
+    # Example structure: ('intermediates', 'decoder', 'moe_layers_2_3', 'self_attention', 'indexer_loss').
+    decoder_intermediates = intermediate_outputs.get("intermediates", {}).get("decoder", {})
+    for layer_intermediates in decoder_intermediates.values():
+      if isinstance(layer_intermediates, dict) and "self_attention" in layer_intermediates:
+        if "indexer_loss" in layer_intermediates["self_attention"]:
+          # flax sow returns a tuple, take the first element and flatten it
+          val = jnp.ravel(layer_intermediates["self_attention"]["indexer_loss"][0])
+          indexer_losses.append(val)
```
shuningjin: (optional) Can we make loss extraction independent of path? I see moe_lb_loss also uses paths. Would a similar logic like max-logit extraction help these cases? can be follow-up
https://github.com/AI-Hypercomputer/maxtext/blob/05c5083d177097059c33647b115bd574b3465413/src/maxtext/utils/qk_clip_utils.py#L21-L57

src/maxtext/trainers/pre_train/train.py
```diff
@@ -218,6 +218,27 @@ def loss_fn(model, config, data, dropout_rng, params, is_train=True):
...
+
+    if indexer_losses:
```
shuningjin, line 233: shouldn't this always be non-empty, as we are inside `if config.use_sparse_indexer and config.indexer_loss_scaling_factor > 0.0`?

tests/unit/deepseek32_vs_reference_test.py
shuningjin: Might be good to not change variable names in the torch reference code? similar for other places above the following line
https://github.com/AI-Hypercomputer/maxtext/blob/0b401fa619866c0caed27c3a51db6542b058775e/tests/unit/deepseek32_vs_reference_test.py#L692

shuningjin (commented):
Thanks for the thoughtful implementation of indexer loss calculation, extraction, and testing! Overall looks good to me. Left some minor comments.

tests/unit/deepseek32_vs_reference_test.py
RissyRan: Sounds good. I will convert them back for alignment.

tests/unit/attention_test.py
RissyRan: Sure, I can combine two tests as one.

src/maxtext/trainers/pre_train/train.py
RissyRan: Theoretically yes, but not always. Imagine if any new checkpoint path is different to `('intermediates', 'decoder', any, 'self_attention', 'indexer_loss')`. Let me add a logging there.

src/maxtext/trainers/pre_train/train.py
RissyRan: The logic appears to validate the path hierarchy by checking if the parent key (path[-2]) matches 'max_logits'. However, this implementation seems to rely on a simple key search, which introduces a relatively loose constraint on the structure. This could lead to ambiguity if 'max_logits' appears elsewhere within the MLP layers

https://github.com/AI-Hypercomputer/maxtext/blob/05c5083d177097059c33647b115bd574b3465413/src/maxtext/utils/qk_clip_utils.py#L45-L49

src/maxtext/layers/attention_mla.py
RissyRan: Thanks! Could you provide reasons behind it?

src/maxtext/layers/attention_mla.py
RissyRan: Sure, I will add it.

src/maxtext/configs/base.yml
RissyRan: Sure, let me add more details.

src/maxtext/trainers/pre_train/train.py
RissyRan: Added. Please let me know your thoughts!

src/maxtext/layers/attention_mla.py
Rohan-Bierneni: +1

Currently, we are using jax.nn.softmax followed by kl_div which is mathematically correct, but it can sometimes lead to NaN grads during training. If the indexer_score logits become very negative, the softmax can underflow to 0.0.

@shuningjin mention of using jax.nn.log_softmax() uses the log-sum-exp trick under the hood to preserve precision, eliminating the risk of division-by-zero or log(0) errors during backprop.

src/maxtext/trainers/pre_train/train.py
Rohan-Bierneni: Maybe instead of matching just the key we can match the last 2 strings of the path tuple? 

So the code can look something like this:
```
indexer_loss = 0.0
if config.use_sparse_indexer and config.indexer_loss_scaling_factor > 0.0:
    flat_intermediates = flatten_dict(intermediate_outputs)
    indexer_losses = []
    
    for path_tuple, value_tuple in flat_intermediates.items():
        # Ensures it came specifically from the attention block, not an arbitrary MLP.
        if len(path_tuple) >= 2 and path_tuple[-2:] == ("self_attention", "indexer_loss"):
            # flax sow returns a tuple, take the first element and flatten it
            val = jnp.ravel(value_tuple[0])
            indexer_losses.append(val)

    if indexer_losses:
        indexer_loss = jnp.mean(jnp.concatenate(indexer_losses))
        loss += indexer_loss
```

RissyRan: Sure let's update it to check 2 keys. So everyone could be happier :)

src/maxtext/layers/attention_mla.py
RissyRan: Hmm... I didn't get it. In the original code, it's already added EPS? Let's discuss offline.

```
kl_per_token = jax.scipy.special.kl_div(attention_probs + EPS, indexer_probs + EPS)
```

RissyRan: Discussed offline. Let's align with [Megatron LM](https://github.com/NVIDIA/Megatron-LM/blob/e8fe06865cd49f80597cadb182d59a7a9b73de99/megatron/core/transformer/experimental_attention_variant/dsa.py#L241-L242) as well. I will update it.

src/maxtext/trainers/pre_train/train.py
RissyRan: Updated, and also tested that loss is captured:

```
path_keys: ('intermediates', 'decoder', 'dense_layers_0_0', 'self_attention', 'indexer_loss')
loss is found
jnp.ravel(val): LinearizeTracer<float32[1]>
path_keys: ('intermediates', 'decoder', 'moe_layers_2_3', 'self_attention', 'indexer_loss')
loss is found
jnp.ravel(val): LinearizeTracer<float32[2]>
path_keys: ('intermediates', 'decoder', 'moe_layers_5_5', 'self_attention', 'indexer_loss')
loss is found
jnp.ravel(val): LinearizeTracer<float32[1]>
path_keys: ('intermediates', 'decoder', 'moe_layers_engram_1', 'self_attention', 'indexer_loss')
loss is found
jnp.ravel(val): LinearizeTracer<float32[1]>
path_keys: ('intermediates', 'decoder', 'moe_layers_engram_4', 'self_attention', 'indexer_loss')
loss is found
jnp.ravel(val): LinearizeTracer<float32[1]>
```

src/maxtext/layers/attention_mla.py
RissyRan: Due to the Nan issue, we will have to switch back to jax.nn.softmax instead of jax.nn.log_softmax. Also discussed offline.

Repost:

Currently, the [indexer_mask](https://github.com/AI-Hypercomputer/maxtext/blob/6b6dbc27b9a683804189fa9641e4a77486646257/src/maxtext/layers/attention_mla.py#L302-L305) is added with attention_mask. For some tokens, there are double added with [DEFAULT_MASK_VALUE](https://github.com/AI-Hypercomputer/maxtext/blob/6b6dbc27b9a683804189fa9641e4a77486646257/src/maxtext/common/common_types.py#L81) (will underflow). attention_probs = jax.nn.softmax(attention_scores) converts the -inf scores to exactly 0.0, and log_indexer_probs = jax.nn.log_softmax(indexer_score) converts the -inf scores to -inf. Then the kl_div is calculated to 0.0 * (log(0.0 + EPS) - (-inf)) -> 0.0 * (-13.8 + inf) -> 0.0 * inf, which gives NaN.

src/maxtext/layers/attention_mla.py
```diff
@@ -951,6 +952,71 @@ def mla_kv_projection(self, inputs: Array, inputs_positions: Array, decoder_segm
...
+      # indexer_mask is already pre-filtered with the attention_mask if any
+      attention_scores = attention_scores + indexer_mask[:, None, :, :]
+      indexer_score = indexer_score + indexer_mask
+    elif attention_mask is not None:
+      # indexer_score already applies attention_mask; updating attention_scores only
+      attention_scores = attention_scores + attention_mask[:, None, :, :]
+
+    # Use float32 for softmax numerical stability.
+    attention_probs = jax.nn.softmax(attention_scores.astype(jnp.float32), axis=-1)
+    indexer_probs = jax.nn.softmax(attention_scores.astype(jnp.float32), axis=-1)
```
shuningjin: for `indexer_probs`: `attention_scores` -> `indexer_score`

tests/unit/attention_test.py
```diff
@@ -1591,6 +1592,107 @@ def test_tpu_flash_attention_context_parallel(
...
+    """Test that KL divergence is 0 when target and pred distributions match exactly."""
+    mla_config_args = self.config_arguments.copy()
+    mla_config_args["use_sparse_indexer"] = True
+    mla_config_args["attention"] = "dot_product"
+    _, mla = self.init_mla(mla_config_args, rope_type="default")
+
+    batch_size = 2
+    q_len = 3
+    kv_len = 4
+    num_heads = 2
```
shuningjin: nit: num_heads = 2 ->  3 to not overlap with batch, may help catch error like above

Rohan-Bierneni (approved):
LGTM. Thank you for the change!

shuningjin (approved):
Thank you! LGTM

src/maxtext/layers/attention_mla.py
```diff
@@ -951,6 +952,71 @@ def mla_kv_projection(self, inputs: Array, inputs_positions: Array, decoder_segm
...
+  def calculate_indexer_loss(
+      self,
+      indexer_score: Array,
+      query: Array,
+      key: Array,
+      attention_mask: Optional[Array | None],
+      indexer_mask: Array,
+      sparse_loss: bool,
+      scaling_factor: float,
+  ) -> Array:
```
NuojCheng, line 964: sort of random thought but I wander why this functions shows a property of module. Why not moving it to max_utils.py? The only place `self` get used in this function is `precision=self.config.matmul_precision` but it can be set as an additional input

shuningjin: I see that moe load balance loss is in a similar situation. The only place self being used is `self.num_experts` and `self.config.load_balance_loss_weight`.
https://github.com/AI-Hypercomputer/maxtext/blob/63f5ca9cb29beba89c7d04e78a86e25f78409c64/src/maxtext/layers/moe.py#L1622-L1631

RissyRan: In my opinion, put it in the attention_mla layer itself makes more sense. One example we should refactor is MFU calculation by moving the logic from maxtext_utils.py into the individual decoder layers. Implementing a layer.get_mfu() method follows a cleaner design pattern, allowing each model (DeepSeek, Mixtral, etc.) to manage its own performance metrics.

PR #3367, opened 2026-03-10
Make tokenizer_path to be non-mandatory

A9isha:
# Description

Simplify the parameters for running on MaxText by removing `tokenizer_path` as a required argument.

FIXES: b/490520651

# Tests

Ran locally on a v5p-8 the following commands:
```
# With tokenizer path
python3 -m src.maxtext.trainers.post_train.rl.train_rl src/maxtext/configs/post_train/rl.yml   model_name=llama3.1-8b-Instruct   tokenizer_path=meta-llama/Llama-3.1-8B-Instruct   load_parameters_path=/path/to/checkpoint   run_name=maz-8b-$RANDOM   base_output_directory=/path/to/gcs/bucket   hf_access_token=<HF_TOKEN> dataset_name=gsm8k steps=4 rollout_tensor_parallelism=-1 rollout_data_parallelism=1 rollout_expert_parallelism=1

# Without tokenizer path
python3 -m src.maxtext.trainers.post_train.rl.train_rl src/maxtext/configs/post_train/rl.yml   model_name=llama3.1-8b-Instruct   tokenizer_path=meta-llama/Llama-3.1-8B-Instruct   load_parameters_path=/path/to/checkpoint   run_name=maz-8b-$RANDOM   base_output_directory=/path/to/gcs/bucket   hf_access_token=<HF_TOKEN> dataset_name=gsm8k steps=4 rollout_tensor_parallelism=-1 rollout_data_parallelism=1 rollout_expert_parallelism=1
```
# Checklist

### DIFF
--- a/src/maxtext/configs/pyconfig.py
+++ b/src/maxtext/configs/pyconfig.py
@@ -29,12 +29,13 @@ import jax.numpy as jnp
 import omegaconf
 
 from maxtext.configs import pyconfig_deprecated
-from maxtext.utils.globals import MAXTEXT_CONFIGS_DIR
+from maxtext.utils.globals import MAXTEXT_CONFIGS_DIR, MAXTEXT_ASSETS_ROOT, HF_IDS
 from maxtext.common.common_types import DecoderBlockType, ShardMode
 from maxtext.configs import types
 from maxtext.configs.types import MaxTextConfig
 from maxtext.inference.inference_utils import str2bool
 from maxtext.utils import max_utils
+from maxtext.utils import max_logging
 
 logger = logging.getLogger(__name__)
 logger.setLevel(os.environ.get("LOGLEVEL", "INFO"))
@@ -62,7 +63,7 @@ def resolve_config_path(param: str) -> str:
   # For pip-installed packages, strip the src prefix and resolve against
   # the installed configs directory (MAXTEXT_CONFIGS_DIR).
   if param.startswith("src/maxtext/configs/"):
-    candidate = os.path.join(MAXTEXT_CONFIGS_DIR, param[len("src/maxtext/configs/"):])
+    candidate = os.path.join(MAXTEXT_CONFIGS_DIR, param[len("src/maxtext/configs/") :])
     if os.path.isfile(candidate):
       return candidate
   return os.path.join("src", param)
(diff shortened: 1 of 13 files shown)


SurbhiJainUSC:
Can we do the similar change for SFT?

A9isha:
> Can we do the similar change for SFT?

Yes done now

src/maxtext/configs/pyconfig.py
```diff
@@ -142,14 +137,26 @@ def _prepare_for_pydantic(raw_keys: dict[str, Any]) -> dict[str, Any]:
...
+        )
         and new_value == ""
     ):
       new_value = None
 
     if key == "run_name" and new_value is None:
       new_value = ""
 
+    if key == "tokenizer_path" and new_value is None:
+      new_value = HF_IDS.get(raw_keys["model_name"])
```
SurbhiJainUSC: What if `MODEL_NAME` is not present in `HF_IDS`?

src/maxtext/configs/types.py
```diff
@@ -205,7 +204,9 @@ class ProfilerType(str, Enum):
     "llama2-13b",
     "llama2-70b",
     "llama3-8b",
+    "llama3.1-8b-Instruct",
```
SurbhiJainUSC, line 208: Instead of explicitly adding `Instruct` model name to this list, can we somehow derive it from the base model? In future, we might need `Instruct` model for gemma and then we would have to update this list.

SurbhiJainUSC: You have added `qwen3-omni-30b-a3b-Instruct` to `HF_IDS`, but it is not present in `ModelName` list.

A9isha: Yes, I think the idea for this PR is to allow on the fly tokenizer_path calculation for the popular models used in the tutorials to make it easier for first time users. We still have the option of passing `tokenizer_path`. So, for the models which are not supported for this on the fly calculations we will default to using the previous `maxtext/assets/tokenizers/tokenizer.llama2` default and if the user intended to use something else they would pass `tokenizer_path`

Great point about `qwen3-omni-30b-a3b-Instruct`, I will remove that change.

src/maxtext/configs/pyconfig.py
A9isha: Please see my response [here](https://github.com/AI-Hypercomputer/maxtext/pull/3367#discussion_r2914868826)

PR #3373, opened 2026-03-11
Add support to infer config with no config file

dipannita08:
# Description

Currently users need to pass in the file path to config file when running MaxText workloads (`src/maxtext/configs/post_train/rl.yml`)

This change adds support to resolve or infer the config without requiring the user to specify the config path, while previous functionality is still supported.

FIXES: b/491604657

# Tests

- CI
-  Run a workload without config:
```
python3 -m maxtext.trainers.pre_train.train \
  run_name=$RUN_NAME \
  base_output_directory=$OUTPUT_PATH \
  dataset_type=synthetic \
  steps=$STEPS
```
-  Run a workload with config:
```
  run_name=$RUN_NAME \
  base_output_directory=$OUTPUT_PATH \
  dataset_type=synthetic \
  steps=$STEPS
```
- Checkpoint conversion test results: https://paste.googleplex.com/4850770987057152

# Checklist

### DIFF
--- a/src/maxtext/configs/pyconfig.py
+++ b/src/maxtext/configs/pyconfig.py
@@ -29,7 +29,7 @@ import jax.numpy as jnp
 import omegaconf
 
 from maxtext.configs import pyconfig_deprecated
-from maxtext.utils.globals import MAXTEXT_CONFIGS_DIR
+from maxtext.utils.globals import MAXTEXT_CONFIGS_DIR, MAXTEXT_PKG_DIR
 from maxtext.common.common_types import DecoderBlockType, ShardMode
 from maxtext.configs import types
 from maxtext.configs.types import MaxTextConfig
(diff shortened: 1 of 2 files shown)


src/maxtext/configs/pyconfig.py
```diff
@@ -227,7 +243,16 @@ def initialize_pydantic(argv: list[str], **kwargs) -> MaxTextConfig:
...
+  config_arg = argv[1]
+  if config_arg.startswith("config="):
+    name = config_arg[len("config="):]
+    if name not in _CONFIG_FILE_MAPPING:
+      raise ValueError(
+          f"Cannot find a config file for config type '{name}'."
+      )
+    config_path = os.path.join(MAXTEXT_CONFIGS_DIR, _CONFIG_FILE_MAPPING[name])
+  else:
+    config_path = resolve_config_path(config_arg)
```
shralex: Instead of the changes above, can we add the following line to resolve_configs_path:

  if param in _CONFIG_FILE_MAPPING: return os.path.join(MAXTEXT_CONFIGS_DIR, _CONFIG_FILE_MAPPING[param])

Here is why I like this more: the changes above require "config=..." to always be the second argument of the command, the key and value need to be parsed using this new logic rather than one of the standard parsing libraries we're using. Instead what I propose is a minimal change to user experience and seems a bit more robust.

shralex (changes requested):

src/maxtext/configs/pyconfig.py
dipannita08: Got rid of the `config=`parameter all together. The config file is inferred if the user does not want to provide the config file path.

src/maxtext/configs/pyconfig.py
```diff
@@ -46,6 +46,47 @@
 # Don't log the following keys.
 KEYS_NO_LOGGING = ("hf_access_token",)
 
+# Module paths to their default config file (relative to MAXTEXT_CONFIGS_DIR).
+_CONFIG_FILE_MAPPING: dict[str, str] = {
```
igorts-git, line 50: do we need to include our checkpoint conversion script `to_maxtext` here as well?

dipannita08: Yes included `to_maxtext` and `to_huggingface`, and validated by @hengtaoguo.

src/maxtext/configs/pyconfig.py
```diff
@@ -46,6 +46,49 @@
...
+    "maxtext.trainers.pre_train.train": "base.yml",
+    "maxtext.trainers.pre_train.train_compile": "base.yml",
+    "maxtext.trainers.post_train.distillation.train_distill": "post_train/distillation.yml",
+    "maxtext.trainers.post_train.rl.train_rl": "post_train/rl.yml",
+    "maxtext.trainers.post_train.sft.train_sft": "post_train/sft.yml",
+    "maxtext.trainers.post_train.sft.train_sft_deprecated": "post_train/sft.yml",
+    "maxtext.inference.decode": "inference/inference.yml",
```
bvandermoon: I thought some of these like decode still mainly used `base.yml`?

bvandermoon: Yes it looks `decode` uses `base.yml`: https://github.com/search?q=repo%3AAI-Hypercomputer%2Fmaxtext%20%22python3%20-m%20maxtext.inference.decode%22&type=code

Mind double checking if some of the others do too?

dipannita08: Thanks, updated.

PR #3382, opened 2026-03-11
Implementation for soft offline distillation using saved top-k teacher logits

ajkv-google:
# Description

This PR introduces an end-to-end offline distillation training pipeline. Previously, the distillation loop executed in an "online" mode, which required both the frozen Teacher model and the learning Student model to be loaded and executed simultaneously during training. This change allows the trainer to load pre-computed, top-K Teacher logits from .array_record files, which allows us to bybass the forward pass for the teacher model during the training loop.

# Tests

Tested this code change by running the following command (using [Yaml File for Offline Distillation](https://paste.googleplex.com/4523353181388800))

`python3 src/maxtext/trainers/post_train/distillation/train_distill.py   src/maxtext/configs/post_train/distillation.yml   steps=100   tokenizer_path="/mnt/ajkv/disks/codebase/maxtext/src/maxtext/assets/tokenizers/tokenizer_llama3.tiktoken"`

Truncated output showing the successful run: https://paste.googleplex.com/4879271282737152.

Verified that the training happened successfully and finished the distillation run. 

# Checklist

### DIFF
--- a/tests/post_training/unit/train_distill_test.py
+++ b/tests/post_training/unit/train_distill_test.py
@@ -709,6 +709,8 @@ class TrainDistillTest(unittest.TestCase):
     mock_student_cfg.checkpoint_period = 10
     mock_student_cfg.gradient_clipping_threshold = 0.0
     mock_student_cfg.eval_interval = -1
+    mock_student_cfg.gradient_accumulation_steps = 1
+    mock_student_cfg.global_batch_size = 8
 
     # Add dummy numbers for strategy math/logic
     mock_student_cfg.distill_temperature = 1.0
@@ -786,6 +788,8 @@ class TrainDistillTest(unittest.TestCase):
     mock_student_cfg.checkpoint_period = 10
     mock_student_cfg.gradient_clipping_threshold = 0.0
     mock_student_cfg.eval_interval = -1
+    mock_student_cfg.gradient_accumulation_steps = 1
+    mock_student_cfg.global_batch_size = 8
 
     # Add dummy numbers for strategy math/logic
     mock_student_cfg.distill_temperature = 1.0


src/maxtext/trainers/post_train/distillation/distillation_utils.py
```diff
@@ -63,13 +68,64 @@ class MaxTextTrainingInput(peft_trainer.TrainingInput):
...
 # -----------------------------------------------------------------------------
 
 
+class OfflineArrayRecordIterator:
+  """Reads the pre-generated global top-k logits file."""
+
+  def __init__(self, data_dir: str, epochs: int = 100):
+    # Check if the user passed a directory or a direct file path
+    if tf.io.gfile.isdir(data_dir):
+      self.filepath = os.path.join(data_dir, "teacher_top_k_global.array_record")
```
entrpn: is it ok to hardcode this file as teacher_top_k_global.array_record?

src/maxtext/trainers/post_train/distillation/train_distill.py
```diff
@@ -605,8 +637,26 @@ def main(argv: Sequence[str]) -> None:
   teacher_config = pyconfig.initialize(teacher_argv, **teacher_overrides)
 
   # 4. Run Training
-  train_distill(student_config, teacher_config)
+  train_distill(student_config, teacher_config, local_args.offline_distillation, local_args.offline_data_dir)
 
 
 if __name__ == "__main__":
-  app.run(main)
+  parser = argparse.ArgumentParser()
```
entrpn: I think these should go inside types.py to add them as part of the config.

ajkv-google: That makes sense, it would make the command less complex and make things more organized if it is in the config. I moved these to types.py and verified the training ran successfully after the change.

src/maxtext/trainers/post_train/distillation/distillation_utils.py
ajkv-google: In the save_top_k_teacher logits file (from this [PR](https://github.com/AI-Hypercomputer/maxtext/pull/3193)), we are writing a single arrayrecord file from one host rather than having multiple hosts write their chunks of data. So, I just named the file as "teacher_top_k_global.arrayrecord". But, I believe not everyone running offline distillation will use the same file to save top-k teacher logits, so I will add this as a field to the config so that users can specify the filename of the saved top-k teacher logits to have it be dynamic.

src/maxtext/configs/types.py
```diff
@@ -1071,6 +1071,14 @@ class Distillation(BaseModel):
       description="Overrides specific to the Teacher model (e.g., {'num_query_heads': 64}).",
   )
 
+  # --- Offline Distillation Fields ---
+  offline_distillation: bool = Field(
```
vlad-karp: looks redundant 
if  you specify offline_data_dir parameter, that can be a direct sign of switching to the offline processing

src/maxtext/trainers/post_train/distillation/train_distill.py
```diff
@@ -513,18 +527,35 @@ def labels_fn(targets, targets_segmentation=None, **kwargs):
...
+    }
+
+    # If we are in online mode then we exit
+    if getattr(batch, "top_k_logits", None) is None:
+      return inputs_dict
+
+    # Scatter the offline arrays into a dense tensor of -10000s
+    dense_shape = batch.input_tokens.shape + (student_config.vocab_size,)
+    dense_logits = jnp.full(dense_shape, -10000.0, dtype=jnp.float32)
+    dense_logits = jnp.put_along_axis(dense_logits, batch.top_k_indices, batch.top_k_logits, axis=-1, inplace=False)
```
vlad-karp, line 562: why inplace=False?

ajkv-google: Jax arrays are immutable, so when you use `jnp.put_along_axis`, it will create a new array with the updated values. In standard numpy, you can modify arrays in place, but cannot with jax. So, you need to set inplace=False so that jax knows to return the newly created array instead of throwing an error for trying to modify the original one. Here is the documentation explaining how this arg must be set to False: https://docs.jax.dev/en/latest/_autosummary/jax.numpy.put_along_axis.html

src/maxtext/configs/types.py
ajkv-google: That's true, I removed the offline_distillation field from the config and relied on offline_data_dir in the train_distill.py file to check when to perform offline/online distillation

src/maxtext/trainers/post_train/distillation/train_distill.py
vlad-karp: yes, but is there a specific need to use a jax initial array instead of numpy one?

ajkv-google: yes, the code runs inside of a jitted function, and numpy functions cannot handle the jax tracers. If we use numpy, the code will crash and not trace properly, so we have to stick to jnp.

vlad-karp: wondering if it will be faster to do this logic on cpu in-place in parallel with train steps execution on tpus

ajkv-google:
Added unit tests to make sure that in offline mode, only the student model is loaded, while in online mode, both the student and teacher models are loaded. Below are the commands I used to run each of the unit tests:

Test offline distillation: 

- `pytest tests/post_training/unit/train_distill_test.py -k "test_main_offline_mode_skips_teacher_loading"`

Test online distillation: 

- `pytest tests/post_training/unit/train_distill_test.py -k "test_main_online_mode_loads_teacher"`

vlad-karp (approved):

PR #3409, opened 2026-03-13
Adding single-host integration test for GRPO and GSPO.

NicoGrande:
# Description

Adds an end-to-end integration test for both GRPO and GSPO workflows on qwen3-0.6b (which should work e2e on a single v6e-4 TPU VM). 

FIXES: b/490410666, b/490411100, b/490409949, b/490410743

# Tests

Tests are skipped for now pending debugging on v6e-4

# Checklist

### DIFF
--- a/src/maxtext/configs/inference/vllm.yml
+++ b/src/maxtext/configs/inference/vllm.yml
@@ -14,6 +14,8 @@
 
 base_config: "base.yml"
 attention: "vllm_rpa"
+model_call_mode: "inference"
+
 # NNX required for vLLM integration
 enable_nnx: True
 # Avoid re-initializing JAX distributed system when using vLLM
--- a/tests/integration/single_host_train_rl_test.py
+++ b/tests/integration/single_host_train_rl_test.py
@@ -1,0 +1,221 @@
+# Copyright 2023–2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Integration test for train RL GRPO and GSPO-token workflows with actual dataset/checkpoint.
+Tests the workflow in src/maxtext/trainers/post_train/rl/train_rl.py
+"""
+
+import os
+import logging
+import unittest
+import pytest
+import jax
+import jax.numpy as jnp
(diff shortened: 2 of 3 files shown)


src/maxtext/configs/inference/vllm.yml
```diff
@@ -14,6 +14,8 @@
 
 base_config: "base.yml"
 attention: "vllm_rpa"
+model_call_mode: "inference"
```
xuefgu, line 17: Where is this new config used below?

tests/integration/single_host_train_rl_test.py
```diff
@@ -0,0 +1,201 @@
...
+        f"chips_per_vm={jax.device_count()}",
+        "scan_layers=True",
+        "hbm_utilization_vllm=0.4",
+        "rollout_data_parallelism=1",
+        "rollout_tensor_parallelism=-1",
+        "rl.num_generations=16",
+        "load_parameters_path=gs://maxtext-model-checkpoints/qwen3-0.6b/2025-10-27/scanned/0/items",
+    ]
+
+  def _run_rl_workflow_end_to_end(self, extra_argv):
```
xuefgu, line 86: Why do we need this vs. directly calling `rl_train()` for various configs?

xuefgu (commented):
Many thanks Nico! Just a couple of quick clarifications.

src/maxtext/configs/inference/vllm.yml
NicoGrande: This is for MoE! Its unrelated but I wanted to squeeze it in to this PR since we should always be setting this in the vLLM codepath.

tests/integration/single_host_train_rl_test.py
NicoGrande: The biggest benefit is to be able to access some of the internals of `rl_train()`. For instance, this allows us to assert that the actor model params are being updated.

xuefgu: The concern is the gradual/eventual divergence of logic in `rl_train()` and the tests. Any way we can mitigate that?

src/maxtext/configs/inference/vllm.yml
dipannita08: Is the entry point for this through train_rl?

xuefgu (approved):
Deferring to you on how to handle the divergence between tests and application logic. Thanks again for the PR!

src/maxtext/configs/inference/vllm.yml
NicoGrande: No, vllm.yml is only used in the MaxText on vLLM adapter class here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py#L73

tests/integration/single_host_train_rl_test.py
NicoGrande: Its a good point - I could add a couple of test cases that just calls `rl_train()` directly to test that API also.

PR #3423, opened 2026-03-16
Restrict supported Python versions to >=3.11

psanal35:
Downstream dependencies, such as `pathwaysutils`, require Python >= 3.11.
The current requires-python >=3.10 causes dependency resolution failures.

This PR updates the project metadata to reflect the supported Python versions. The error:
```
Updating https://github.com/AI-Hypercomputer/pathways-utils.git (HEAD)
    Updated https://github.com/mlperf/logging.git (34903058ee63d3a7c3bcc1a6ca04e2aa5ed6af9c)
    Updated https://github.com/AI-Hypercomputer/pathways-utils.git (2c8593c7ac6e183f9208be1940dac123373bf9bb)
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version == '3.10.*'):
  ╰─▶ Because the requested Python version (>=3.10) does not satisfy
      Python>=3.11 and jax>=0.7.2 depends on Python>=3.11, we can conclude
      that jax>=0.7.2 cannot be used.
      And because only the following versions of jax are available:
          jax<=0.7.2
          jax==0.8.0
          jax==0.8.1
          jax==0.8.2
          jax==0.8.3
          jax==0.9.0
          jax==0.9.0.1
          jax==0.9.1
      and pathwaysutils==0.1.5 depends on jax>=0.7.2, we can conclude that
      pathwaysutils==0.1.5 cannot be used.
      And because only pathwaysutils==0.1.5 is available and your project
      depends on pathwaysutils, we can conclude that your project's
      requirements are unsatisfiable.
      hint: While the active Python version is 3.12, the resolution failed for
      other Python versions supported by your project. Consider limiting your
      project's supported Python versions using `requires-python`.
      hint: The `requires-python` value (>=3.10) includes Python versions that
      are not supported by your dependencies (e.g., jax>=0.7.2 only supports
      >=3.11). Consider using a more restrictive `requires-python` value (like
      >=3.11).
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ name = "MaxText"
 version = "0.1.0"
 description = "MaxText is a high performance, highly scalable, open-source LLM written in pure Python/Jax and targeting Google Cloud TPUs and GPUs for training and inference"
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 license = {file = "LICENSE"}
 dynamic = ["dependencies"]
 


psanal35:
Opened to the wrong branch --closing.

PR #3437, opened 2026-03-18
Added SFT Pre-Processing for Grain Input Pipeline

ajkv-google:
# Description

This PR introduces SFT support to the Grain input pipeline by adding a separate `sft_preprocessing_pipeline` function. Rather than cluttering the existing pretrain code, it uses simple conditionals inside the train and eval iterators to route to this new SFT logic. I followed the existing Hugging Face SFT implementation and adapted its logic to be compatible with Grain's element-wise datasets. 

# Tests

I added a unit test to verify end-to-end functionality to make sure the Grain SFT pipeline formats the data and outputs correctly. Ran this command to execute the unit test: 

- `pytest tests/unit/grain_data_processing_test.py::GrainSFTParquetProcessingTest -v`

This is the output of the test: [Test Passed Output](https://paste.googleplex.com/4548555512610816)

Also, ran the training pipeline in Maxtext with sft enabled using a grain dataset with this command:
- `python3 -m maxtext.trainers.post_train.sft.train_sft     src/maxtext/configs/post_train/sft.yml     run_name=test_grain_sft     dataset_type=grain     grain_file_type=parquet     grain_train_files=gs://maxtext-dataset/hf/ultrachat_200k/train_sft-*.parquet     steps=10     tokenizer_type=huggingface     tokenizer_path=HuggingFaceH4/zephyr-7b-beta`

Verified that the sft processing changes worked and trained successfully: [Logs](https://paste.googleplex.com/4895584239616000)

# Checklist

### DIFF
--- a/src/maxtext/input_pipeline/data_processing_utils.py
+++ b/src/maxtext/input_pipeline/data_processing_utils.py
@@ -26,7 +26,7 @@ from maxtext.input_pipeline import tokenizer
 
 def parse_and_keep_features(dataset, config, data_columns, tokenize):
   """Parse arrayrecord features or keep specified columns for other formats."""
-  if config.grain_file_type == "arrayrecord":
+  if config.grain_file_type in ("arrayrecord", "tfrecord"):
     dataset = dataset.map(input_pipeline_utils.ParseFeatures(data_columns, tokenize))
     dataset = dataset.map(input_pipeline_utils.NormalizeFeatures(data_columns, tokenize))
   else:


src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+def sft_preprocessing_pipeline(
+    dataset,
+    config,
+    data_columns,
+    tokenize,
+    grain_worker_count,
+    grain_per_worker_buffer_size,
+):
+  """Use grain pipeline to pre-process the dataset and return iterators for sft fine-tuning"""
+  if config.grain_file_type == "arrayrecord":
```
vlad-karp: there is exactly this block in the pretrain_preprocessing_pipeline(). Almost identical code there suggests it would be great to reuse it

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+    dataset = dataset.map(input_pipeline_utils.ParseFeatures(data_columns, tokenize))
+    dataset = dataset.map(input_pipeline_utils.NormalizeFeatures(data_columns, tokenize))
+  else:
+    dataset = dataset.map(input_pipeline_utils.KeepFeatures(feature_names=data_columns))
+
+  tokenizer_model = tokenizer.build_tokenizer(
```
vlad-karp: same for the tokenizer

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+    pad_id = tokenizer_model.unk_id
+  else:
+    pad_id = -1
+
+  tokenizer_model = tokenizer_model.tokenizer
+
+  if getattr(config, "chat_template", None) and hasattr(tokenizer_model, "chat_template"):
+    tokenizer_model.chat_template = config.chat_template
+
+  supported_columns = [["prompt", "completion"], ["messages"], ["question", "answer"]]
```
vlad-karp: This is again repeating logic from the hf pipeline, one can extract utility routines to avoid code repetition

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+def _format_chat_template_grain(element, data_columns, tokenizer_model):
+  """Grain-compatible mapping function to format raw columns into conversational messages."""
+  # Convert raw columns to conversational messages
+  if "messages" in data_columns:
+    messages = element["messages"]
+  elif set(data_columns) == {"prompt", "completion"}:
+    messages = [{"role": "user", "content": element["prompt"]}, {"role": "assistant", "content": element["completion"]}]
+  elif set(data_columns) == {"question", "answer"}:
+    messages = [{"role": "user", "content": element["question"]}, {"role": "assistant", "content": element["answer"]}]
+  else:
```
vlad-karp, line 278: HF pipeline asserts sft is running on a conversational format

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+  if tokenize:
+    dataset = dataset.map(
+        functools.partial(
+            _tokenize_sft_chunks,
+            text_column_name=data_columns[0],
+            tokenizer_model=tokenizer_model,
+        )
+    )
+
+  dataset = dataset.map(
```
vlad-karp, line 334: any pros/cons of doing it via Grain operations like in the hf pipeline?

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+  )
+  data_columns = ("inputs", "targets")
+
+  # Pack and batch examples
+  batch_size = config.global_batch_size_to_load // jax.process_count()
+  if config.expansion_factor_real_data > 1:
+    # global_batch_size_to_load has been expanded in pyconfig.py when expansion_factor_real_data > 1.
+    batch_size = int(batch_size // config.expansion_factor_real_data)
+
+  if config.packing:
```
vlad-karp: this block is again identical to one in pretrain_preprocessing_pipeline().
Also, would be great to check if need some sft related modifications

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+    }
+    dataset = dataset.map(input_pipeline_utils.Rekey(rekey_dict))
+  else:
+    dataset = dataset.map(input_pipeline_utils.PadOrTrimToMaxLength(config.max_target_length, pad_id))
+
+  batch_fn = functools.partial(grain.experimental.batch_and_pad, batch_size=batch_size, pad_value=pad_id)
+  dataset = dataset.batch(batch_size, batch_fn=batch_fn)
+
+  # Shift inputs for teacher-forced training
+  dataset = dataset.map(
```
vlad-karp: should it alway be executed in a generic sft_preprocessing_pipeline() ?

src/maxtext/input_pipeline/grain_data_processing.py
aireenmei: dataset.map is the newer and recommended way of using Grain

src/maxtext/input_pipeline/grain_data_processing.py
aireenmei: Yes, there are many common operations between pretrain and sft. I think it's a good idea to extract the common pattern into util functions

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -352,6 +352,166 @@ def dpo_preprocessing_pipeline(
...
+  if getattr(config, "chat_template", None) and hasattr(tokenizer_model, "chat_template"):
+    tokenizer_model.chat_template = config.chat_template
+
+  supported_columns = [["prompt", "completion"], ["messages"], ["question", "answer"]]
+  assert any(
+      set(data_columns) == set(supported) for supported in supported_columns
+  ), f"Dataset column names mismatch. Expected columns to match one of {supported_columns}, but got {data_columns}"
+
+  dataset = dataset.map(
+      functools.partial(_format_chat_template_grain, data_columns=data_columns, tokenizer_model=tokenizer_model)
```
aireenmei, line 322: The hf pipeline calls instruction_data_processing.convert_to_conversational_format, do we support the same conversion here? https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/input_pipeline/hf_data_processing.py#L261

src/maxtext/input_pipeline/grain_data_processing.py
ajkv-google: I extracted the shifting logic into a generic shift_dataset() helper in data_processing_utils.py and applied it uniformly across both the SFT and Pretrain pipelines.

src/maxtext/input_pipeline/grain_data_processing.py
ajkv-google: Yes, we do support the conversion. Because Grain's `.map()` processes row-by-row (unlike HF dataset operations), I implemented the conversion inside `_format_chat_template_grain()` above this line. It handles both prompt/completion and question/answer pairs.

src/maxtext/input_pipeline/data_processing_utils.py
```diff
@@ -0,0 +1,146 @@
...
+      config.add_bos,
+      config.add_eos,
+      config.hf_access_token,
+  )
+  if tokenizer_model.pad_id is not None:
+    pad_id = tokenizer_model.pad_id
+  elif tokenizer_model.unk_id is not None:
+    pad_id = tokenizer_model.unk_id
+  else:
+    pad_id = -1
```
vlad-karp: I think 0 as the deafult is better

src/maxtext/input_pipeline/data_processing_utils.py
```diff
@@ -0,0 +1,146 @@
...
+  batch_size = config.global_batch_size_to_load // jax.process_count()
+  if config.expansion_factor_real_data > 1:
+    # global_batch_size_to_load has been expanded in pyconfig.py when expansion_factor_real_data > 1.
+    # But when using Grain, we want to keep the batch_size consistent with that in the checkpoint.
+    # We revert the batch_size expansion here, but load multiple batches per step in multihost_dataloading.py.
+    batch_size = int(batch_size // config.expansion_factor_real_data)
+  return batch_size
+
+
+def pack_or_pad_and_batch_dataset(dataset, config, batch_size, pad_id, data_columns, tokenizer_model):
```
vlad-karp: some simpler name?

src/maxtext/input_pipeline/grain_data_processing.py
vlad-karp: my concern was that the comment suggests it is distillation only logic but it is applied always.
What is the meaning of that shift operation? should it only be applied in distillation pipeline?

ajkv-google: I see, yeah that comment is a bit misleading. I just meant standard next-token prediction, not distillation. From my understanding, the 1-token shift is required for all autoregressive training such as pretraining and sft to align inputs and targets, and it can be applied in distillation as well.

src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -335,20 +243,96 @@ def dpo_preprocessing_pipeline(
...
+    data_columns,
+    tokenize,
+    grain_worker_count,
+    grain_per_worker_buffer_size,
+):
+  """Use grain pipeline to pre-process the dataset and return iterators for sft fine-tuning"""
+  dataset = data_processing_utils.parse_and_keep_features(dataset, config, data_columns, tokenize)
+
+  tokenizer_model, pad_id = data_processing_utils.get_tokenizer_and_pad_id(config)
+  base_tokenizer_model = tokenizer_model
```
JamesDeng42, line 313: So for base_tokenizer_model, we assume it is a huggingFace tokenizer? if otherwise, this line 298 might throw AttributeError.

ajkv-google: That's a good point, thanks for pointing this out. I just added a change to make sure there is a "tokenizer" attribute

PR #3438, opened 2026-03-18
Add custom mesh and logical rule support

NuojCheng:
# Description

This pull request introduces the ability to define and use customized physical meshes and logical axis rules in MaxText via external YAML configuration files. This enhances flexibility for complex parallelism strategies, such as those required for large scale MoE setups.

### Configuration Enhancements
- **Custom Config Loading**: Added a `custom_mesh_and_rule` field to `base.yml`.
- **Dynamic YAML Parsing**: Implemented logic in `src/maxtext/configs/types.py` to automatically load `mesh_axes`, `logical_axis_rules`, and `data_sharding` from specified YAML files located in `src/maxtext/configs/custom_mesh_and_rule/`.
- **Example Configs**: Provided initial custom configurations for `pure-fsdp.yml` and a complex `ds3-large-pp.yml`.

### Robust Utility Functions
- Added helper functions in `src/maxtext/utils/max_utils.py` for safer access to mesh properties:
    - `get_mesh_axes_size`: Returns the size of a physical axis or 1 if it doesn't exist.
    - `get_logical_rule_contents`: Safely retrieves physical axes corresponding to a logical name.

### Code Refactoring for Mesh Flexibility
- Updated `src/maxtext/layers/attention_op.py`, `src/maxtext/utils/sharding.py`, and `src/maxtext/utils/train_utils.py` to use the new utility functions. This ensures the model can compile even when standard axes (like `data` or `fsdp`) are absent or redefined in a custom mesh.

# Tests

Added `tests/unit/custom_mesh_and_rule_test.py` to verify successful compilation using the new custom mesh flags for both FSDP and multi-axis configurations.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -431,6 +431,7 @@ internal_compile_num_devices: -1 # You must specify the number of devices when u
 
 # Parallelism
 shard_mode: "auto" # can be either auto or explicit
+custom_mesh_and_rule: "" # replace default mesh and logical rule by specifying yml name under config/mesh_and_rule/.
 mesh_axes: ['diloco', 'data', 'stage', 'fsdp', 'fsdp_transpose', 'sequence', 'context', 'context_autoregressive', 'tensor', 'tensor_transpose', 'tensor_sequence', 'expert', 'autoregressive']
 logical_axis_rules: [
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
--- a/src/maxtext/configs/custom_mesh_and_rule/pipeline-large-moe.yml
+++ b/src/maxtext/configs/custom_mesh_and_rule/pipeline-large-moe.yml
@@ -1,0 +1,68 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This logical rule is designed to optimize pipeline parallelism for large-scale jobs. 
+# Key changes include removing expert weight sharding on the `q_lora` dimension, which 
+# is relatively small (e.g., 512 for DeepSeek), and limiting sharding strategies when 
+# EP x FSDP > 512. 
+#
+# The `data` axis is preserved for two reasons: first, the pipeline stage acts as a 
+# data parallel (DP) domain externally, making the `data` axis a necessary reference; 
+# second, it may be required for DCN communication. 
+#
+# Finally, the `tensor` axis is used to shard weights when `pipeline_fsdp_ag_once` or 
+# `pipeline_fsdp_ag_per_repeat` is enabled, ensuring we have sufficient memory to 
+# store prefetched weights.
+mesh_axes: ['data', 'stage', 'fsdp', 'tensor', 'expert']
+data_sharding: [['data', 'stage', 'fsdp', 'tensor', 'expert']]
+logical_axis_rules: [
+                      ['activation_batch', ['data', 'fsdp', 'expert']],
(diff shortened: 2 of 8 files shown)


src/maxtext/configs/custom_mesh_and_rule/pipeline-large-moe.yml
gobbleturk, line 1: Should ds3 and pp be included in this filename? Does this generalize to any large model we need to use 2d fsdp with? Would folks not using pp not want to use this - what is specific about pipelining to these rules?

src/maxtext/configs/custom_mesh_and_rule/pure-fsdp.yml
gobbleturk, line 1: this is kinda beautiful tho

src/maxtext/configs/base.yml
```diff
@@ -431,6 +431,7 @@ internal_compile_num_devices: -1 # You must specify the number of devices when u
 
 # Parallelism
 shard_mode: "auto" # can be either auto or explicit
+custom_mesh_and_rule: ""
```
gobbleturk: for better or worse we have our best documentation in this file - can you include a comment or a perhaps a link to documentation here for what custom_mesh_and_rule is?

src/maxtext/utils/max_utils.py
```diff
@@ -50,6 +50,21 @@
 # pylint: disable=too-many-positional-arguments
 
 
+def get_mesh_axes_size(mesh, axes_name):
```
gobbleturk: perhaps these should go in sharding.py?

src/maxtext/utils/max_utils.py
```diff
@@ -50,6 +50,21 @@
...
+  """Get physical axes size in given mesh."""
+  if axes_name in mesh.shape.keys():
+    return mesh.shape[axes_name]
+  return 1
+
+
+def get_logical_rule_contents(rule, logical_name):
```
gobbleturk: I realize this mainly a refactor of existing code but this looks odd to me - many of our rules have duplicate keys - it seems this code will somewhat arbitrarily choose the top rule in a list (not sure how dict(rule) works if duplicate keys in rule)

tests/unit/custom_mesh_and_rule_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pytest
+
+from maxtext.trainers.pre_train.train_compile import main as train_compile_main
+from tests.utils.test_helpers import get_test_config_path
+
+
+@pytest.mark.tpu_backend
+class CustomMeshAndRuleTest(unittest.TestCase):
+  """Tests for custom_mesh functionality in train_compile.py"""
```
gobbleturk, line 32: not clear to me this is an effective test of custom_mesh - even if we pass empty logical axis rules I think this test would pass. Can we instead perhaps assert that the model weights are sharded as we expect or something that actually checks the rules?

src/maxtext/configs/custom_mesh_and_rule/pipeline-large-moe.yml
NuojCheng: I added some comments in each new yml file

src/maxtext/configs/custom_mesh_and_rule/pure-fsdp.yml
NuojCheng: agree! And this rule is the one we use for 90% of time!

src/maxtext/configs/base.yml
NuojCheng: added!

src/maxtext/utils/max_utils.py
NuojCheng: both functions are removed now

src/maxtext/utils/max_utils.py
NuojCheng: I took a closer look at the `attention_op.py` and they are poorly coded unfortunately. I totally agree with you. I used `nn.logical_to_mesh_axes` to replace the original statement but it probably does not solve your concern. We need a more JAXy refactoring of `attention_op.py`, but not this PR!

tests/unit/custom_mesh_and_rule_test.py
NuojCheng: Ok for this test I want to make sure custom rules bug-free. If in future PR we added `mesh.shape["context"]` somewhere it will bug by this test since for some custom rules we did not define "context" axes. This test is not targeted for proper functionality of custom rule, though we could have one, but implementing them is non-trivial ; (

PR #3440, opened 2026-03-18
Add `abort_on_nan_loss` and `abort_on_inf_loss` options

Steboss:
# Description

We experienced from time to time the following problem: 
```bash 
0: I0122 21:20:55.494193 140737350402816 metric_logger.py:180] completed step: 30, seconds: 0.449, TFLOP/s/device: 1877.176, Tokens/s/device: 36470.971, total_weights: 131072, loss: nan
```
While a model training was carried on, the loss was `nan`. 
This should be flagged as an error, and the training job should be stopped. 

For this reason, here we are introducing `abort_on_nan_loss` and `abort_on_inf_loss` for checking NaN/Inf in training jobs loss. The options are boolean, the check is done after all the metrics have been written with `write_metrics` function. Two unit tests have been added, one in `tests/unit/configs_value_test.py` to check CLI overrides of values, and one in `tests/unit/metric_logger_abort_test.py` to check if the implementation works fine.

Future improvements: 
- These two options could be lists, so that we can monitor NaN or Inf on `grad` or `params` or other variables.

# Tests

The changes were tested through the created unit tests in `tests/unit/configs_value_test.py` and `tests/unit/metric_logger_abort_test.py`

# Checklist

### DIFF
--- a/tests/unit/configs_value_test.py
+++ b/tests/unit/configs_value_test.py
@@ -126,26 +126,6 @@ class ConfigTest(unittest.TestCase):
     config = pyconfig.initialize(argv)
     self.assertEqual(config.tokenizer_type, "tiktoken")
 
-  def test_abort_on_nan_loss_defaults_and_cli_override(self):
-    """Tests abort_on_nan_loss default from base.yml and CLI override parsing."""
-    default_cfg = pyconfig.initialize(["", _BASE_CONFIG_PATH, "run_name=test_default_abort_nan"])
-    self.assertEqual(default_cfg.abort_on_nan_loss, True)
-
-    cli_cfg = pyconfig.initialize(
-        ["", _BASE_CONFIG_PATH, "run_name=test_cli_abort_nan", "abort_on_nan_loss=False"]
-    )
-    self.assertEqual(cli_cfg.abort_on_nan_loss, False)
-
-  def test_abort_on_inf_loss_defaults_and_cli_override(self):
-    """Tests abort_on_inf_loss default from base.yml and CLI override parsing."""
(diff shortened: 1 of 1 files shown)


tests/unit/configs_value_test.py
```diff
@@ -126,6 +126,26 @@ def test_llama3_tokenizer_correction(self):
     config = pyconfig.initialize(argv)
     self.assertEqual(config.tokenizer_type, "tiktoken")
 
+  def test_abort_on_nan_loss_defaults_and_cli_override(self):
```
gobbleturk: I don't think these tests are necessary - this is generic to our config setup, not specific to these values, we don't need individual tests for every value (the other tests provide great coverage in your new file!)

src/maxtext/configs/base.yml
```diff
@@ -852,6 +852,8 @@ decode_sampling_temperature: 1.
 eval_interval: -1  # the specific number of train step between eval_step
 eval_steps: -1  # run this number of steps for eval, recommend setting this to prevent error due to running out of evel data
 target_eval_loss: 0.  # early stop once reaching target eval_loss
+abort_on_nan_loss: True # Check for NaN and abort if found in training loss
```
gobbleturk, line 855: will check with team about these defaults - we may prefer default to false

gobbleturk: fail fast - leave as True it is!

tests/unit/configs_value_test.py
Steboss: Removed :)

Steboss:
@gobbleturk  thanks for reviewing this. I removed the tests from `configs_value_test.py`. Please, let me know if you need other modifications :)

PR #3427, opened 2026-03-16
NNX migration prep (1/N): pure_nnx flag and init_state_fn scaffolding

ecnal-cienet:
# NNX Migration Route Map
1. :white_check_mark: Add NNX support function and utils / pure_nnx flag=False. Won't affect current Linen workflow.
2. :x: NNX fully supported. pure_nnx flag=False, but user can do NNX runs/tests.
3. :x: NNX unit tests and performance tests are completed and verified. pure_nnx flag=True.
4. :x: Remove Linen code related NNX flags.

# Description

> **Note:** This is the first in a series of NNX migration PRs. Pure NNX training is **not yet implemented** — all NNX code paths currently raise `NotImplementedError`. This PR only introduces the structural scaffolding needed for subsequent patches to plug in NNX logic without modifying shared infrastructure.

This PR introduces two abstractions that enable incremental NNX migration while keeping existing Linen code fully functional.

## `pure_nnx` config flag
A boolean (`configs/base.yml`, `configs/types.py`) that will route all major code paths — training, compilation, inference, RL, and utilities — to pure-NNX logic when `True`, falling back to Linen otherwise. Defaults to `False` so all existing behaviour is unchanged.

## `init_state_fn`
A pluggable callable for initializing the model training state, threaded through `create_checkpoint_manager`, `setup_training_state`, `setup_decode_state`, and `get_abstract_state`. This decouples state
initialization from shared infrastructure so future NNX and Linen paths can provide their own implementations without forking utilities.

## Other structural changes
- `create_training_tools` is split into `create_training_optimizer` and `create_checkpoint_manager` for cleaner separation of concerns.
- `jit_train_step` gains a `mesh` parameter to accommodate NNX callers where `model` is a `GraphDef` with no `.mesh` attribute.
- `get_shaped_inputs` in `train_compile.py` adds a `pure_nnx` branch that omits `example_rng`, matching the future NNX `train_step(state, batch)` signature.
- `get_first_step` restored to the two-argument `(model, state)` form to support both Linen `TrainState` and NNX `TrainStateNNX` step retrieval.
- All entry points (`grpo_trainer`, `maxengine`, `generate_param_only_checkpoint`, `layerwise_quantization`, `lora_utils`, `standalone_checkpointer`, integration tests) updated to accept and pass `init_state_fn`.

## New files
[...]

PR #3442, opened 2026-03-18
NNX train

charlesli640:
# Description

Implement pre-train using NNX style

# Tests
```
python3 src/maxtext/trainers/pre_train/nnx_train.py src/maxtext/configs/base.yml \
run_name="run_llama2_7b" \
model_name="llama2-7b" \
dataset_type=synthetic \
steps=10 \
scan_layers=True \
debug_sharding=True \
async_checkpointing=False \
remat_policy=full \
checkpoint_storage_use_zarr3=false \
enable_checkpointing=false \
enable_nnx=true \
pure_nnx_decoder=true
```
# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/compare_linen_nnx_checkpoint.py
+++ b/src/maxtext/checkpoint_conversion/compare_linen_nnx_checkpoint.py
@@ -1,0 +1,609 @@
+# Copyright 2023-2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


bvandermoon (commented):
Thank you @charlesli640. Is it possible to migrate `train.py` directly instead of forking this logic? I am concerned these two files will get out of sync before they are merged, this setup will skip running unit tests, and it makes the code a bit more complicated/harder to follow

charlesli640:
> Thank you @charlesli640. Is it possible to migrate `train.py` directly instead of forking this logic? I am concerned these two files will get out of sync before they are merged, this setup will skip running unit tests, and it makes the code a bit more complicated/harder to follow

Definitely we can move the logic to train.py and make it controlled by enable_nnx config. Actually this is one of experimental solutions I am doing internally - try to create brand-new pre-train using pure NNX style, leaving old linen style pre-train untouched/co-existing. 

Another solution is submitted on PR [#3427](https://github.com/AI-Hypercomputer/maxtext/pull/3427). This solution tries to keep/re-use current linen style train loop as much as possible. It created TrainStateNNX class and make existing linen functions compatible to both linen model and nnx model. Please also review the PR3427. We can discuss more on which direction we are going.

PR #3451, opened 2026-03-18
[TE] Fix TE permute silent fallback issue and fix integration of TE permute

jberchtold-nvidia:
# Description

Currently, if `te_permutation_impl=true`, it will try TE permute but if it isn't available (e.g. `triton` isn't installed), it will silently fall back to the MaxText permute. Importantly, MaxText permute does not respet `te_permutation_align_size`, meaning silently are groups will not meet alignment requirements, which can lead to IMA or ILLEGAL_INSTRUCTION from the TE GMM.

This PR fixes this so that trying `te_permutation_impl=true` without TE permute available will raise an error instead of silently falling back.

Additionally, there was a small integration issue with TE permute and the global padding of the group buffer (not intra-group padding) where we get an error due to TE permute's buffer output exceeding the expected amount. This is actually expected because to pad individual groups, we also need to increase the total buffer size to the worst case padding. I've restructured where the TE grouped GEMM is called to skip this MaxText padding as with TE permute we handle this padding already.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -928,10 +928,11 @@ class RoutedMoE(nnx.Module):
       - lb_loss: Load balance loss (or None).
       - bias_updates: Bias updates (or None).
     """
-    use_te = getattr(self.config, "te_permutation_impl", False) and te_permutation.TE_PERMUTATION_AVAILABLE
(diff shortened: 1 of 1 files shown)


jberchtold-nvidia:
Closing, meant to target a fork instead of the official upstream as these changes are still experimental

PR #3493, opened 2026-03-24
[DRAFT] Paged Stashing for ring-of-experts MoE Activation Memory

abhinavgoel95:
## Problem

  When `use_ring_of_experts=True` with expert parallelism EP=4, the ring-all-gather inflates the token buffer on each device to `batch × EP × seq × top_k` tokens. For DeepSeek V3 671B with `batch=2, seq=4096, top_k=8, EP=4`, this is **262,144 tokens/device**.                                                                                 
   
  The `moe_mlpwo` GMM output has shape `(262144, 7168)` = **3.5 GB per layer**. With `scan_layers=True` over 60 MoE layers and host offloading, this totals **~210 GB** of host memory — exceeding typical limits (180 GB).                                                                                                                         
                                                                                                          
  ## Solution: Paged Stashing                                                                                                                                                                                                                                                                                                                       
                                                                                                          
  Inspired by [Megatron-LM PR #2690](https://github.com/NVIDIA/Megatron-LM/pull/2690), this PR introduces a **paged stash** mechanism that decouples the *compute buffer* (static worst-case shape, required by XLA) from the *storage buffer* (actual tokens only).                                                                                
   
  ### Key idea                                                                                                                                                                                                                                                                                                                                      
                                                                                                          
  - A single **shared static buffer** is allocated once for all MoE layers: `total_capacity = num_moe_layers × expected_tokens_per_layer + max_chunk_slack`                                                                                                                                                                                         
  - Each layer writes only its **actual** routed tokens into this buffer at a tracked offset using `lax.dynamic_update_slice` (static size, dynamic start index — XLA-compatible)
  - On backward pass, each layer reads back its compact slice and expands to the full compute shape                                                                                                                                                                                                                                                 
  - Expected actual tokens ≈ `worst_case / EP = 262144 / 4 = 65536`, so the shared buffer is ~4× smaller than per-layer worst-case storage
                                                                                                                                                                                                                                                                                                                                                    
  ### Memory comparison (DeepSeek V3 671B, EP=4, 60 MoE layers)                                           
                                                                                                                                                                                                                                                                                                                                                    
  | Approach | Storage for `moe_mlpwo` | Notes |                                                                                                                                                                                                                                                                                                    
  |---|---|---|
  | Baseline (offload per layer) | ~210 GB | 262k × 7168 × 60 layers |                                                                                                                                                                                                                                                                              
  | Paged stash (this PR) | ~52 GB | 65k × 7168 × 60 layers |                                             
                                                                                                                                                                                                                                                                                                                                                    
  ## Implementation
                                                                                                                                                                                                                                                                                                                                                    
  ### New file: `src/MaxText/layers/paged_stash.py`                                                       

  - `make_stash_fns(max_chunk, hidden)` — returns `(stash_fn, restore_fn)` pair with `jax.custom_vjp` so gradients flow correctly through the compact/expand operations                                                                                                                                                                             
  - `stash_buffer_size(num_moe_layers, expected_per_layer, max_chunk)` — buffer sizing helper
  - `expected_tokens_per_layer(batch, ep, seq, top_k)` — computes expected load (= worst_case / EP)                                                                                                                                                                                                                                                 
                                                                                                          
  ### Modified: `src/MaxText/layers/moe.py`                                                                                                                                                                                                                                                                                                         
                                                                                                          
  When `config.ring_paged_stash=True`, replaces the `checkpoint_name("moe_mlpwo")` call with `stash_fn`/`restore_fn` calls that pack actual tokens into the shared buffer.                                                                                                                                                                          
   
  ### Modified: `src/maxtext/configs/types.py`                                                                                                                                                                                                                                                                                                      
                                                                                                          
  Adds two new config fields:                                                                                                                                                                                                                                                                                                                       
  - `ring_paged_stash: bool = False`                                                                      
  - `ring_paged_stash_safety_margin: float = 1.5`

  ## TODOs / Open Questions

  - [ ] **Decoder scan carry wiring**: The shared `stash_buf` and `write_ptr` need to be threaded through the `lax.scan` carry in `decoders.py` / `deepseek.py`. Currently the per-layer integration is sketched but the cross-layer buffer threading is not yet wired up. Looking for guidance on the best pattern here.                           
  - [ ] **`restore_fn` backward**: The gradient accumulation in `restore_fn_bwd` needs review — specifically the scatter from compact → full shape.
  - [ ] **Tests**: Unit tests for stash/restore round-trip and gradient correctness.                                                                                                                                                                                                                                                                
  - [ ] **`moe_mlpwi_0`/`moe_mlpwi_1`**: Same technique could apply to wi outputs if they are offloaded.

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -629,6 +629,24 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  ring_paged_stash: bool = Field(
+      False,
+      description=(
+          "Enable paged stashing for ring-of-experts MoE layers. "
+          "Instead of checkpointing GMM activations at worst-case buffer size "
+          "(batch*EP*seq*top_k), compactly stores only the actual routed tokens "
+          "in a shared static buffer, reducing host-offload memory ~4x for EP=4. "
+          "See layers/paged_stash.py for details."
+      ),
+  )
+  ring_paged_stash_safety_margin: float = Field(
+      1.5,
+      description=(
+          "Safety margin multiplier on the expected-per-layer token count used to "
+          "size each layer's stash chunk (max_chunk = expected * margin). "
+          "1.0 = no slack; 1.5 = tolerate 50%% per-layer imbalance without dropping."
+      ),
+  )
   use_random_routing: bool = Field(False, description="Whether to use random routing for debugging.")
   interleave_moe_layer_step: int = Field(1, description="Frequency of MoE layers, e.g., 2 means every 2nd layer is MoE.")
   expert_shard_attention_option: Literal["fsdp", "context"] = Field(
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -25,6 +25,7 @@ from aqt.jax.v2 import aqt_tensor as aqt
 from flax import nnx
 import jax
 from jax import ad_checkpoint as adc
+from MaxText.layers import paged_stash as ps
 from jax.experimental import xla_metadata
 from jax.sharding import NamedSharding, Mesh
 from jax.sharding import PartitionSpec as P
@@ -1305,7 +1306,40 @@ class RoutedMoE(nnx.Module):
         )
       if self.config.mlp_bias:
         intermediate_output = intermediate_output + wo_bias
-      intermediate_output = adc.checkpoint_name(intermediate_output, "mlpwo")
+      if self.config.use_ring_of_experts and self.config.ring_paged_stash:
+        # -------------------------------------------------------------------
+        # Paged stashing: instead of checkpointing intermediate_output at
+        # worst-case size (batch*EP*seq*top_k, hidden), store only the actual
+        # routed tokens in a compact shared buffer passed through the scan
+        # carry.  This reduces host-offload memory ~4x for EP=4.
+        #
+        # stash_buf and write_ptr are threaded through the decoder scan carry.
+        # TODO: wire (stash_buf, write_ptr, layer_sizes) into the decoder
+        # __call__ signatures in decoders.py / deepseek.py.  For now this
+        # block shows the moe.py side of the integration.
+        #
+        # See layers/paged_stash.py for full documentation.
+        # -------------------------------------------------------------------
+        actual_tokens = jnp.sum(group_sizes)
+        expected = ps.expected_tokens_per_layer(
+            batch_size, num_expert_parallelism, sequence_length,
+            self.config.num_experts_per_tok,
+        )
+        max_chunk = int(expected * self.config.ring_paged_stash_safety_margin)
+        stash_fn, restore_fn = ps.make_stash_fns(max_chunk, self.config.emb_dim)
+
+        # Stash: compact intermediate_output into the shared buffer.
+        # stash_buf / write_ptr come from the scan carry (see TODO above).
+        stash_buf, write_ptr = stash_fn(
+            stash_buf, write_ptr, intermediate_output, actual_tokens
+        )
+        # intermediate_output is no longer needed; the backward will restore it.
+        intermediate_output = restore_fn(
+            stash_buf, write_ptr - actual_tokens, actual_tokens,
+            intermediate_output.shape[0],
+        )
+      else:
+        intermediate_output = adc.checkpoint_name(intermediate_output, "mlpwo")
 
       if self.config.use_ring_of_experts:
         # Set the outputs of tokens which were not processed to 0.
--- a/src/maxtext/layers/paged_stash.py
+++ b/src/maxtext/layers/paged_stash.py
@@ -1,0 +1,222 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


src/maxtext/configs/types.py
```diff
@@ -629,6 +629,24 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  ring_paged_stash: bool = Field(
+      False,
+      description=(
+          "Enable paged stashing for ring-of-experts MoE layers. "
+          "Instead of checkpointing GMM activations at worst-case buffer size "
+          "(batch*EP*seq*top_k), compactly stores only the actual routed tokens "
```
RissyRan, line 637: for the formula, could you align it with https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/src/maxtext/layers/moe.py#L1170?

src/maxtext/configs/types.py
```diff
@@ -629,6 +629,24 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  ring_paged_stash: bool = Field(
```
RissyRan, line 632: Shall we also add those into base.yml & this [doc](https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/docs/reference/core_concepts/moe_configuration.md#L4) for alignment? Similar comments for other occurrences.

RissyRan: Shall we call it ring_of_experts_paged_stash to distinguish ring attetnion?

src/maxtext/configs/types.py
```diff
@@ -629,6 +629,24 @@ class MoEGeneral(BaseModel):
...
+          "in a shared static buffer, reducing host-offload memory ~4x for EP=4. "
+          "See layers/paged_stash.py for details."
+      ),
+  )
+  ring_paged_stash_safety_margin: float = Field(
+      1.5,
+      description=(
+          "Safety margin multiplier on the expected-per-layer token count used to "
+          "size each layer's stash chunk (max_chunk = expected * margin). "
+          "1.0 = no slack; 1.5 = tolerate 50%% per-layer imbalance without dropping."
```
RissyRan, line 647: Will this strategy introduce dropping? If I understand correctly, no matter how we set margin value, it will iterate till the last chunk, right? If so, shall we avoid dropping word here?

src/maxtext/layers/moe.py
```diff
@@ -1305,7 +1306,40 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
...
+        #
+        # stash_buf and write_ptr are threaded through the decoder scan carry.
+        # TODO: wire (stash_buf, write_ptr, layer_sizes) into the decoder
+        # __call__ signatures in decoders.py / deepseek.py.  For now this
+        # block shows the moe.py side of the integration.
+        #
+        # See layers/paged_stash.py for full documentation.
+        # -------------------------------------------------------------------
+        actual_tokens = jnp.sum(group_sizes)
+        expected = ps.expected_tokens_per_layer(
```
RissyRan, line 1324: We found this math will give smaller range: https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/src/maxtext/layers/moe.py#L1170?

src/maxtext/layers/paged_stash.py
```diff
@@ -0,0 +1,222 @@
+# Copyright 2025 Google LLC
```
RissyRan, line 1: nit: 2026

src/maxtext/layers/paged_stash.py
```diff
@@ -0,0 +1,222 @@
...
+
+However, with load-balance loss the *actual* tokens routed to each EP shard is
+roughly:
+
+    expected = worst_case / EP              (e.g. 65,536)
+
+Naively checkpointing the GMM outputs (moe_mlpwi_0, moe_mlpwi_1, moe_mlpwo) at
+worst-case size balloons host-offload memory to ~210 GB for a 60-layer model.
+
+Idea (inspired by Megatron-LM PR #2690 "Paged Stashing")
```
RissyRan, line 33: Shall we include a link?

src/maxtext/layers/paged_stash.py
```diff
@@ -0,0 +1,222 @@
...
+
+  - On the backward scan the process is reversed using the stored per-layer
+    offsets and sizes.
+
+Memory comparison (bs=2, EP=4, seq=4096, top_k=8, hidden=7168, 60 MoE layers):
+
+  Strategy                  | tokens/layer  | Host memory (moe_mlpwo only)
+  --------------------------|---------------|-----------------------------
+  No cap (baseline)         | 262,144       | ~210 GB   ❌
+  50% static cap            | 131,072       | ~105 GB   ✅
```
RissyRan, line 70: We didn't see the trend of dropping strategy due to model quality. Shall we remove this option if Paged stash is obvious better here?

src/maxtext/layers/paged_stash.py
```diff
@@ -0,0 +1,222 @@
...
+import jax
+import jax.numpy as jnp
+from jax import lax
+
+
+# ---------------------------------------------------------------------------
+# Core stash / restore primitives
+# ---------------------------------------------------------------------------
+
+def make_stash_fns(max_chunk: int, hidden: int):
```
RissyRan, line 104: Shall we write a unit test in moe_test.py to ensure the correctness with expert sharding? One example: https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/tests/unit/moe_test.py#L542

RissyRan (commented):
Thank you for the change! Have a few initial comments. Also, it seems unit tests are quite red for this PR. Could you please check?

PR #3498, opened 2026-03-25
Specifically install maxtext 0.2.1 in documentation

bvandermoon:
# Description

Pin `maxtext==0.2.1` in ReadTheDocs installation guides. Our commands have been defaulting to install `maxtext==0.1.0`, which we figured out is caused by using the `--resolution=lowest` flag to keep transitive dependencies locked.

Pinning the latest version in the docs for now, but we will want to find a better long-term solution.

Next steps: Can we somehow cherry-pick this change into the ReadTheDocs `0.2.1` tag?

# Tests

```
uv pip install maxtext[tpu]==0.2.1 --resolution=lowest
pip show maxtext # gives 0.2.1
```

# Checklist

### DIFF
--- a/docs/build_maxtext.md
+++ b/docs/build_maxtext.md
@@ -61,7 +61,7 @@ source ${VENV_NAME?}/bin/activate
 
 # Install MaxText with the [runner] extra
 # This enables Docker image building and workload scheduling via XPK
-uv pip install maxtext[runner] --resolution=lowest
+uv pip install maxtext[runner]==0.2.1 --resolution=lowest
 ```
 
 > **Note:** The `maxtext[runner]` extra includes all necessary dependencies for building MaxText Docker images and running workloads through XPK. It automatically installs XPK, so you do not need to install it separately to manage your clusters and workloads.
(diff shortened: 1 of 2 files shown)


docs/build_maxtext.md
```diff
@@ -61,7 +61,7 @@ source ${VENV_NAME?}/bin/activate
 
 # Install MaxText with the [runner] extra
 # This enables Docker image building and workload scheduling via XPK
-uv pip install maxtext[runner] --resolution=lowest
+uv pip install maxtext[runner]==0.2.1 --resolution=lowest
```
SurbhiJainUSC, line 64: Should we do `uv pip install maxtext[runner]>=0.2.1 --resolution=lowest` instead?

bvandermoon: This silently selects `0.1.0` for me unless I add quotes around `uv pip install "maxtext[runner]>=0.2.1" resolution==lowest`. I also think it will generally still select `0.2.1` even after we release later versions due to the `--resolution=lowest` flag. Thoughts?

SurbhiJainUSC: hmm, I guess you are right. We need to sort this out before the next release.

PR #3633, opened 2026-04-10
Fix DS v3.2 decoding issue

RissyRan:
# Description

Fix DeepSeek V3.2 model decoding issue, see more details in b/500088962.

* We should enable the Dense warm up only when `indexer_loss_scaling_factor > 0` like mentioned [here](https://github.com/AI-Hypercomputer/maxtext/blob/b83ff02ca64b3a26feeaf9efe79670d7483cfa36/src/maxtext/configs/base.yml#L385).
* We should use `model_mode` instead of `self.model_mode`. Updated other branches as well.

# Tests

* Before the change: random outputs
* After the change: reasonable decoding [outputs](https://screenshot.googleplex.com/9TdgJTW46b4UihA)

# Checklist

### DIFF
--- a/src/maxtext/layers/decoders.py
+++ b/src/maxtext/layers/decoders.py
@@ -1102,11 +1102,14 @@ class Decoder(nn.Module):
       logits = None
     # When in the Indexer Dense Warm-up stage, skip the expensive output head projection
     # for efficiency, as the main model is frozen and the LM loss is not needed.
-    elif (cfg.use_indexer and not cfg.indexer_sparse_training) and self.model_mode == MODEL_MODE_TRAIN:
+    # TODO(b/501446870): Investigate model_mode as train at beginning for decoding stage
+    elif (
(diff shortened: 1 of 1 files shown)


shuningjin (approved):
Thanks for the investigation and fix! Could you also add a comment on the model mode for future reference?
https://github.com/AI-Hypercomputer/maxtext/blob/b83ff02ca64b3a26feeaf9efe79670d7483cfa36/src/maxtext/layers/decoders.py#L1117

RissyRan:
> Thanks for the investigation and fix! Could you also add a comment on the model mode for future reference?
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/b83ff02ca64b3a26feeaf9efe79670d7483cfa36/src/maxtext/layers/decoders.py#L1117

Added.

Rohan-Bierneni (approved):
LGTM! Thanks for the fix.

PR #3643, opened 2026-04-10
Add back is_batch_shard_by_expert

NuojCheng:
# Description

This PR recover the `is_batch_shard_by_expert` logic fixing previous full EP decoding issue b/501537579. We introduced a `decode_batch_moe` logical rule to specifically deal with the decoding case in moe component when batch size is one.

# Tests

- Performance test: https://paste.googleplex.com/4690855089799168. We get the same training performance (~81 TFLOP/s/dev compared to #3645)

- Inference vllm test: https://paste.googleplex.com/5221471321456640

- Inference decode test: https://paste.googleplex.com/6557626352664576

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1027,7 +1027,13 @@ class RoutedMoE(nnx.Module):
         output = output[: orig_inputs_shape[0]]
       return output
 
-    batch_logical_axis = "activation_batch"
+    # The batch is sharded by expert, except during inference decoding (where batch size == 1).
+    # In the decoding case, the expert axis is instead replicated along the tensor's batch dimension.
+    is_batch_sharded_by_expert = inputs.shape[0] > 1
+    if is_batch_sharded_by_expert:
+      batch_logical_axis = "activation_batch"
+    else:
+      batch_logical_axis = "decode_batch_moe"
 
     if self.get_tensor_transpose_parallelism_size() > 1:
       input_partition_pspec = self._logical_to_mesh_axes(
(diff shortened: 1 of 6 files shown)


src/maxtext/layers/moe.py
```diff
@@ -1027,7 +1027,31 @@ def gmm(
         output = output[: orig_inputs_shape[0]]
       return output
 
-    batch_logical_axis = "activation_batch"
+    # Currently, we support data, tensor, and expert parallelism with Megablox.
```
gobbleturk: longer term this logic should be separated into its own function following go/small-functions, but for now this is probably safer as a rollback and I am currently working on refactoring/cleaning up this file

RissyRan (commented):
Thanks Nuojin for the change! I was comparing the before/after, and noticed this change. Could you help cross check?

* https://screenshot.googleplex.com/XuuC5bCYnWpgzd9
* Diff: https://diff.googleplex.com/#key=B3n5CBidn5Wc

NuojCheng:
> Thanks Nuojin for the change! I was comparing the before/after, and noticed this change. Could you help cross check?
> 
> * https://screenshot.googleplex.com/XuuC5bCYnWpgzd9
> * Diff: https://diff.googleplex.com/#key=B3n5CBidn5Wc

The previous hint `("activation_batch_no_exp", "activation_length_no_exp", None)` was not correct. It always makes the output replicated along expert physical axis. The new implementation is a more memory-efficient way. 

This change of logic only impacts performance under explicit shard mode,  which has not been included in benchmarking yet. One exception is batch split, which has their own code path `deepseek_batchsplit.py` and should not get impacted.

RissyRan (approved):

tests/integration/decode_tests.py
```diff
@@ -96,6 +96,38 @@ class DecodeTests(unittest.TestCase):
...
+          "tokenizer_path=deepseek-ai/DeepSeek-V3.2-Exp",
+          "scan_layers=False",
+          "attention=dot_product",
+          "sparse_matmul=True",
+          "use_tokamax_gmm=False",
+          "dtype=bfloat16",
+          "weight_dtype=bfloat16",
+          "per_device_batch_size=1",
+          "max_prefill_predict_length=3072",
+          "max_target_length=4096",
```
khatwanimohit: how long does this test take to run ? probably can reduce the max_target_length

khatwanimohit (approved):

tests/integration/decode_tests.py
NuojCheng: Good idea! Done!

PR #3662, opened 2026-04-14
Track next token prediction loss in training workload

Rohan-Bierneni:
# Description

We want to also track the immediate next token prediction loss when MTP is enabled. We can get this by doing total_loss - mtp_loss. The benefit of this is we can see if having mtp enabled is helping the model's loss converge at a faster rate.

# Tests

Unit tests should pass.

# Checklist

### DIFF
--- a/src/maxtext/trainers/pre_train/train.py
+++ b/src/maxtext/trainers/pre_train/train.py
@@ -430,6 +430,7 @@ def train_step(model, config, state_mesh_shardings, params_shardings, state, dat
       "learning/indexer_loss": indexer_loss,
       "learning/mtp_loss": mtp_loss,
       "learning/total_weights": total_weights,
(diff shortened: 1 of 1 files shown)


src/maxtext/trainers/pre_train/train.py
```diff
@@ -430,6 +430,7 @@ def move(path, value):
       "learning/indexer_loss": indexer_loss,
       "learning/mtp_loss": mtp_loss,
       "learning/total_weights": total_weights,
+      "learning/loss_without_mtp": loss - mtp_loss,
```
parambole, line 433: @Rohan-Bierneni We would like to track the main model loss without any auxiliary loss, as well as the loss including auxiliary and other losses. However, this approach might not scale or hold true for other auxiliary losses.

PR #3692, opened 2026-04-17
Add enable_expert_parallel flag in vLLM RL

hengtaoguo:
# Description

Add an explicit flag `enable_expert_parallel` in vLLM sampler config. Previously maxtext doesn't have this flag "enable_expert_parallel", and we were hacking through rollout_expert_parallelism>1. 

For the usage, in the vllm weight sync logic, later we may need this flag to determine whether to use rollout_tensor_parallelism as the actual EP value for conversion.

# Tests

N/A

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -1731,6 +1731,7 @@ class VLLM(BaseModel):
   hbm_utilization_vllm: float = Field(0.72, description="Target HBM utilization for vLLM.")
   swap_space_vllm_gb: int = Field(2, description="Swap space in GB for vLLM.")
   enable_dp_attention: bool = Field(False, description="Enable the attn_dp mesh axis in vLLM.")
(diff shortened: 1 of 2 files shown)


NicoGrande (approved):
LGTM!

khatwanimohit (approved):

bvandermoon (approved):

bvandermoon (commented):
Could you add this flag to `base.yml` also? Seems we need a path to get rid of `base.yml` and cutover to `types.py`

hengtaoguo:
> Could you add this flag to `base.yml` also? Seems we need a path to get rid of `base.yml` and cutover to `types.py`

I do see many flags in `types.py` not included in base.yml, such as `rollout_data_parallelism` and `train_micro_batch_size`. Those flags are very specific to RL so may not be necessary to base.yml? But agree that we should consolidate to types.py to avoid confusion.

PR #3699, opened 2026-04-19
Fix masking error when using mlp_bias=True

snehalv2002:
# Description

Bug fix for b/497864549: [XL ML] NaN training loss for GPT-OSS SFT.

GPT-OSS is currently experiencing NaN training loss after one step when using `expert_parallelism > 1`. The bug was caused by lack of masking in the mlp bias inside of expert computation. Since no other models use the mlp bias they didn't experience this issue. 

## Issue
Currently, when `expert_parallelism > 1` we introduce a buffer to store the output of `ragged_all_to_all`, which may contain padding along the token axis. If padding values are not masked after adding the MLP bias, JAX will include them in the gradient computation. `jnp.where` allows us to disconnect the padding values from the backwards graph during bias gradient calculation.

# Tests

1. Ran `pre_train.train` on `GPT_OSS` with `expert_parallelism > 1`: https://paste.googleplex.com/6279941382602752
2. logits matching against golden logits: https://paste.googleplex.com/6165440129794048

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1237,6 +1237,7 @@ class RoutedMoE(nnx.Module):
                 global_sorted_experts=selected_experts,
                 use_custom_sort_vjp=self.config.use_custom_sort_vjp,
             )
+        mask = jnp.arange(x.shape[0]) < jnp.sum(group_sizes)
 
       if self.config.mlp_bias:
         w0_bias, w1_bias, wo_bias = self.transform_bias(selected_experts, w0_bias, w1_bias, wo_bias)
@@ -1300,6 +1301,7 @@ class RoutedMoE(nnx.Module):
         layer_w0 = jax.lax.psum(layer_w0, "tensor_transpose")
       if self.config.mlp_bias:
         layer_w0 = layer_w0 + w0_bias
+        layer_w0 = jnp.where(mask[:, None], layer_w0, 0)
       layer_w0 = adc.checkpoint_name(layer_w0, "moe_mlpwi_0")
 
       layer_w1 = gmm_fn(
@@ -1312,6 +1314,7 @@ class RoutedMoE(nnx.Module):
         layer_w1 = jax.lax.psum(layer_w1, "tensor_transpose")
       if self.config.mlp_bias:
         layer_w1 = layer_w1 + w1_bias
+        layer_w1 = jnp.where(mask[:, None], layer_w1, 0)
       layer_w1 = adc.checkpoint_name(layer_w1, "moe_mlpwi_1")
       intermediate_layer = self.apply_ffn_activation(layer_w0, layer_w1)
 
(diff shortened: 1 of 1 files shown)


RissyRan (commented):
Thanks! Could you have a logits correctness check with EP>1 using GPT-OSS?

src/maxtext/layers/moe.py
```diff
@@ -1304,6 +1301,7 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
         layer_w0 = jax.lax.psum(layer_w0, "tensor_transpose")
       if self.config.mlp_bias:
         layer_w0 = layer_w0 + w0_bias
+        layer_w0 = jnp.where(mask[:, None], layer_w0, 0)
```
RissyRan, line 1304: Is default mask defined without EP sharding?

snehalv2002: Yes the mask is defined even with no EP. [`mask`](https://github.com/AI-Hypercomputer/maxtext/pull/3699/changes#diff-e080eae53990c3bcfd6cd1f60db63f57d80988dfe8715b4ffbc09fbcabb5d8f3R1240) is defined outside the EP > 1 block

RissyRan (approved):
Thanks for the fix!

src/maxtext/layers/moe.py
```diff
@@ -2138,12 +2146,13 @@ def _moe_body(
         )
       else:
         x, routing, route_metadata = route(x, logits, pre_bias_logits, rngs, input_ids=sharded_input_ids)
+        mask = jnp.arange(x.shape[0]) < jnp.sum(routing.group_sizes)
```
gagika: One thing before merge:

`mask` assumes valid rows = `sum(routing.group_sizes)`, which only holds when ring-of-experts is off. With ring, `get_gmm_for_local_experts` uses `routing.local_group_sizes` (L1942) or a `group_offset`-ed slice (L1950), while `routing.group_sizes` stays the **global** bincount — so `sum(...) >= x.shape[0]`, the mask is all-True, and `use_ring_of_experts=True` + `mlp_bias=True` still NaNs. It doesn't fail loudly, it just quietly does nothing.

Suggestion: compute the valid-row count next to `get_gmm_for_local_experts`, off the same branches, so the two can't drift apart:

```python
def valid_token_count(x, routing, route_metadata):
  """Rows of `x` the gmm actually computes; the rest is ragged-buffer padding."""
  num_ep = self.get_expert_parallelism_size()
  num_experts_per_shard = self.config.num_experts // num_ep
  if self.config.use_ring_of_experts and x.shape[0] < routing.sorted_selected_experts.shape[0]:
    return jnp.sum(routing.local_group_sizes)
  if self.config.use_ragged_sort and self.config.use_ring_of_experts:
    experts_start = route_metadata.expert_shard_id * num_experts_per_shard
    return jnp.sum(
        jax.lax.dynamic_slice_in_dim(routing.group_sizes, experts_start, num_experts_per_shard)
    )
  return jnp.sum(routing.group_sizes)
```

snehalv2002: Done

gagika (approved):

PR #3724, opened 2026-04-22
Track lower-case maxtext package in coverage to unblock Codecov

ecnal-cienet:
## Summary

- Add `maxtext` (lower-case) to `[run] source` in `.coveragerc` alongside `MaxText` (upper-case).

## Why

`coverage.py` uses `[run] source` to decide which packages to trace. The config previously listed only `MaxText`, so files imported via the lower-case `maxtext` package (the NNX-migrated path at `src/maxtext/`) were never instrumented. Lines covered by real unit tests therefore showed up on Codecov as `null` rather than `hit`, pushing patch coverage artificially low.

The existing `[paths]` aliasing only merges already-collected measurements across filesystems — it does not control what is traced in the first place.

# Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 # .coveragerc
 
 [run]
-source = MaxText
+source = maxtext
 branch = True
 omit =
     tests/*


.coveragerc
```diff
@@ -2,6 +2,7 @@
 
 [run]
 source = MaxText
+         maxtext
```
bvandermoon: These are pointing to `src/MaxText` and `src/maxtext`, right? The upper-case `MaxText` can be removed if so. There is only `src/MaxText/__init__` now: https://github.com/AI-Hypercomputer/maxtext/tree/532c8b3d8a1e299fd9b05a521d4f7f08253c2141/src/MaxText

bvandermoon (approved):
LGTM ,

.coveragerc
ecnal-cienet: Done — since `src/MaxText/` only contains `__init__.py` (a shim), tracking it in [run] source doesn't add coverage. Dropped the `MaxText` entry, leaving just `maxtext`.

SurbhiJainUSC (approved):

PR #3727, opened 2026-04-22
fix dtype mismatch when using vocab_tiling

aireenmei:
# Description
Encountered [this error](https://paste.googleplex.com/6595822797062144) when testing gemma4 with bfloat16 and `num_vocab_tiling=8`.

The root cause is the incompatibility of weight_dtype=bfloat16 and vocab_tiling. In my test, I was using `weight_dtype=float32 dtype=bfloat16`,  but gemma4 has a bug that uses config.dtype (should use config.weight_dtype) to initiate layer scalar, resulting in a bfloat16 weight that causes this error.

This PR fix:
* add dtype cast to allow using weight_dtype=bfloat16 with vocab_tiling. Although weight_dtype=bfloat16 is not common, I see some test scripts use it, so it's still good to support
* gemma4 layer scalar initiate with config.weight_dtype

# Tests
* Added unit test and verified that test fail without the fix and pass after.
* [test script](https://paste.googleplex.com/5252416829259776) 
* [log with error](https://cloudlogging.app.goo.gl/R3YseuHvduWy6cbe6)
* [log after fix](https://cloudlogging.app.goo.gl/r46gq5pKdcZmaoYV8)

# Checklist

### DIFF
--- a/src/maxtext/utils/vocabulary_tiling.py
+++ b/src/maxtext/utils/vocabulary_tiling.py
@@ -224,10 +224,10 @@ def vocab_tiling_linen_loss(
         _bwd_scan_body, initial_grad_params_acc, (reshaped_hidden_states, reshaped_labels, reshaped_segmentation)
     )
     grad_reshaped_hidden_states = _maybe_shard_with_name(grad_reshaped_hidden_states, reshaped_hidden_spec)
-    # TODO (chengnuojin): we may want to convert grad_params to bf16 to save memory
-    # grad_params = jax.tree_util.tree_map(lambda x, y: y.astype(x.dtype), gathered_params, grad_params)
     # Chain-rule to accumulate gradients
     grad_params = jax.tree_util.tree_map(lambda g: g * loss_cotangent, grad_params)
+    # Cast cotangents back to each primal's dtype; custom_vjp requires dtype match.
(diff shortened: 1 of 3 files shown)


src/maxtext/utils/vocabulary_tiling.py
```diff
@@ -228,6 +228,8 @@ def _bwd_scan_body(grad_params_acc, chunk_data):
     # grad_params = jax.tree_util.tree_map(lambda x, y: y.astype(x.dtype), gathered_params, grad_params)
```
NuojCheng: could you remove the comments in line 227 and 228?

NuojCheng (approved):
thank you for the fix!

gobbleturk (commented):
Is there a compile test using vocab tiling that should have caught this? Can we add some test protecting this?

src/maxtext/utils/vocabulary_tiling.py
aireenmei: Cleaned the comments as discussed offline

aireenmei:
> Is there a compile test using vocab tiling that should have caught this? Can we add some test protecting this?

Added and updated PR decription

gobbleturk (approved):
Awesome, thank you for the test!

PR #3733, opened 2026-04-23
make AddLabel workflow work as expected

charlesli640:
# Description

Make Add Label workflow work - adding Label "pull ready" when PR got approved.

 1.Make AddLabel workflow in chain of "MaxText Package Tests".
 2.Make AddLabel workflow to fail by `core.setFailed + return;` when condition is not met
 3.Need >=2 approvals (with different approvers) to set the "pull ready" label

# Tests

CI/CD

# Checklist

### DIFF
--- a/.github/workflows/AddLabel.yml
+++ b/.github/workflows/AddLabel.yml
@@ -16,9 +16,8 @@ name: Add Label
 
 on:
   workflow_run:
-    workflows: [Tests, CodeQL]
-    types:
-      - completed
+    workflows: ["MaxText Package Tests"]
+    types: [completed]
   pull_request_review:
   pull_request_review_comment:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/AddLabel.yml
```diff
@@ -16,7 +16,7 @@ name: Add Label
 
 on:
   workflow_run:
-    workflows: [Tests, CodeQL]
```
charlesli640, line 19: Only need chain CI Test workflow "MaxText Package Tests"

.github/workflows/AddLabel.yml
```diff
@@ -68,15 +69,15 @@ jobs:
...
             }
             let is_approved = false
             if (decision_result.repository.pullRequest.reviewDecision === "APPROVED") {
               is_approved = true
             }
             if (!is_approved) {
               console.log("Not adding pull ready because the PR is not approved yet by sufficient code owners.")
-              process.exit()
+              process.exit(1)
             }
```
MichaelHudgins: Would all of this logic be better handled by making this a condition of the job executing at all.

```
if: github.event.review.state == 'approved'
```

See https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

.github/workflows/AddLabel.yml
```diff
@@ -85,11 +86,11 @@ jobs:
...
               per_page: 100,
             })
-            // Check that the number of commits in the PR is 1.
-            if (commits.data.length !== 1) {
-              console.log("Not adding pull ready because the PR has more than one commit. Please squash your commits.")
-              process.exit(1)
-            }
+            // Check that the number of commits in the PR is 1. This is not enforced in practice.
+            // if (commits.data.length !== 1) {
+            //  console.log("Not adding pull ready because the PR has more than one commit. Please squash your commits.")
```
MichaelHudgins: I recommend leaving instructions or a link on how to do this for users.  

https://docs.jax.dev/en/latest/contributing.html#single-change-commits-and-pull-requests is how jax lists it

.github/workflows/AddLabel.yml
```diff
@@ -85,11 +86,11 @@ jobs:
               pull_number,
               per_page: 100,
             })
-            // Check that the number of commits in the PR is 1.
-            if (commits.data.length !== 1) {
-              console.log("Not adding pull ready because the PR has more than one commit. Please squash your commits.")
-              process.exit(1)
-            }
+            // Check that the number of commits in the PR is 1. This is not enforced in practice.
```
MichaelHudgins: I also recommend adding a bypass label for checks like this.  Allow the reviewer to basically flag this PR is ok to have multi commits.

.github/workflows/AddLabel.yml
```diff
@@ -98,13 +99,13 @@ jobs:
...
             }
             for (const checkRun of checkRuns.data.check_runs) {
               if (checkRun.name.endsWith(context.job)) continue
               if (checkRun.conclusion !== "success") {
                 console.log("Not adding pull ready because " + checkRun.name + " has not passed yet: " + checkRun.html_url)
-                process.exit()
+                process.exit(1)
               }
             }
             console.log("Adding pull ready label because the PR is approved AND all the check runs have passed")
```
MichaelHudgins, line 121: This job seems like it should be doing some extra verification on who the approver is before adding the label.

.github/workflows/AddLabel.yml
charlesli640: Agree. That logic is more concise.

.github/workflows/AddLabel.yml
charlesli640: Actually not all people can approve PRs. And the approvers are kind of dynamic as people may join/leave the project. I am afraid it is not necessary to check who approved the PR. 
Even we check who approved PR, what is the next step? What is the criteria to give it a green light or red light?

.github/workflows/AddLabel.yml
charlesli640: If forcing single-commit is the rule, let's follow the rule.

.github/workflows/AddLabel.yml
charlesli640: Actually don't want to make it too complicated for this simple workflow.
If multi-commit is acceptable in special case, approver can manually add "pull ready" label to specially treat that PR.

.github/workflows/AddLabel.yml
charlesli640: The error github.event.review not found (or being empty) will occur when workflow is triggered by an event that does not include a review object in its payload. Cannot use github.event.review.state directly without checking.

.github/workflows/AddLabel.yml
```diff
@@ -16,9 +16,8 @@ name: Add Label
 
 on:
   workflow_run:
-    workflows: [Tests, CodeQL]
-    types:
-      - completed
+    workflows: ["MaxText Package Tests"]
```
charlesli640, line 19: Comments test

shralex (approved):

shralex (approved):

.github/workflows/AddLabel.yml
charlesli640: comments trigger test.

khatwanimohit (approved):

PR #3736, opened 2026-04-23
Add fused_moe_mlp: fuse wi_0 and wi_1 into one grouped GEMM for MoE FFN1

abhinavgoel95:
Analogous to the existing fused_mlp flag for dense models. When enabled, concatenates the wi_0 and wi_1 expert weight matrices along the output dimension ([G,K,N] + [G,K,N] -> [G,K,2N]), issues a single grouped GEMM call, then splits the result. This halves FFN1 kernel launches and reads the input activations from HBM once instead of twice.

Works with all grouped GEMM backends (megablox, tokamax, ragged_dot). Off by default; requires sparse_matmul=True.

# Description

Adds a `fused_moe_mlp` boolean config flag (default `false`) for MoE models, analogous to the existing `fused_mlp` flag for dense models. When enabled, the two FFN1 input projections (`wi_0` gate and `wi_1` up) are fused into a single grouped GEMM call instead of two separate   
  ones.                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                         
  **Why this change:**                                      
  In a gated MoE FFN, `wi_0` and `wi_1` share the same input `x` but are currently dispatched as two independent grouped GEMMs. This means `x` is loaded from HBM twice and two kernel launches are issued back-to-back for every MoE layer.
                                                                                                                                                                                                                                                                                         
  **How it works:**
  The two expert weight matrices (`[G, K, N]` each) are concatenated along the output dimension to form a single fused weight `[G, K, 2N]`. One grouped GEMM call produces `[M, 2N]`, which is then split into `layer_w0` and `layer_w1`. This is the same approach `fused_mlp` uses for 
  dense models (stacking `wi_0`/`wi_1` into a single `dot_general`), extended to the grouped GEMM case.                                                                                                                                                                                  
   
  **Benefits:**                                                                                                                                                                                                                                                                          
  - Halves FFN1 kernel launches (2 → 1) per MoE layer       
  - Reads input activations `x` from HBM once instead of twice
  - Presents the accelerator with a wider `N = 2 × mlp_dim` GEMM, which typically achieves better hardware utilization than two narrower back-to-back GEMMs                                                                                                                              
  - Backend-agnostic: works with megablox, tokamax, and `jax.lax.ragged_dot`
                                                                                                                                                                                                                                                                                         
  **Implementation:**                                       
  The change is entirely in `sparse_matmul` in `moe.py` — a conditional branch on `config.fused_moe_mlp` wraps the existing `gmm_fn` calls. The unfused path is unchanged. A config-load-time validation ensures `sparse_matmul=True` (the dense `einsum` path does not use `gmm_fn` and 
  is unaffected). Off by default (`fused_moe_mlp: false`).                                                                                                                                                                                                                               
   
  # Tests                                                                                                                                                                                                                                                                                
                                                            
  Forward-pass correctness: run with `fused_moe_mlp=false` and `fused_moe_mlp=true` on identical synthetic inputs and verify `intermediate_layer` outputs match numerically. Gradient correctness: verify gradients w.r.t. `wi_0` and `wi_1` match the unfused path (JAX AD traces       
  cleanly through `jnp.concatenate` + slicing with no custom VJP needed). Validation: confirm `fused_moe_mlp=true` with `sparse_matmul=false` raises `ValueError` at config load.
              
# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -496,6 +496,12 @@ class RoutedMoE(nnx.Module):
 
     kernel_in_axis = np.arange(1)
     kernel_out_axis = np.arange(1, 2)
+    # Pad the MoE input weight kernels for GMM_v2 execution when configured.
+    moe_intermediate_dim = (
+        self.config.padded_base_moe_mlp_dim
+        if self.config.padded_base_moe_mlp_dim is not None
+        else self.intermediate_dim
+    )
 
     if quantizations.in_serve_mode(self.quant):
       # During aqt convert state we delete kernel weight from params to save
@@ -504,13 +510,7 @@ class RoutedMoE(nnx.Module):
       self.wi_0 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
       self.wi_1 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
       self.wo = jnp.zeros((num_experts, intermediate_dim, self.moe_expert_input_dim))
-    elif self.config.prefuse_moe_weights and self.config.attention == "vllm_rpa":
-      # Pad model dimension in Fused MoE weight kernels for GMM_v2 execution.
-      moe_intermediate_dim = (
-          self.config.padded_base_moe_mlp_dim
-          if self.config.padded_base_moe_mlp_dim is not None
-          else self.intermediate_dim
-      )
+    elif self.config.prefuse_moe_weights:
       self.wi = nnx.Param(
           self.kernel_init(
               self.rngs.params(),
@@ -532,12 +532,6 @@ class RoutedMoE(nnx.Module):
           out_sharding=self.wo_kernel_axes,
       )
     else:
-      # Pad model dimension in Unfused MoE weight kernels for GMM_v2 execution.
-      moe_intermediate_dim = (
-          self.config.padded_base_moe_mlp_dim
-          if self.config.padded_base_moe_mlp_dim is not None
-          else self.intermediate_dim
-      )
       self.wi_0 = nnx.Param(
           self.kernel_init(
               self.rngs.params(),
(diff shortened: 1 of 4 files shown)


src/maxtext/layers/moe.py
```diff
@@ -1318,29 +1318,44 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
...
-          weight_gather_axes=wi_gather_axes,
-      )
-      if self.get_tensor_transpose_parallelism_size() > 1:
-        layer_w1 = jax.lax.psum(layer_w1, "tensor_transpose")
-      if self.config.mlp_bias:
-        layer_w1 = layer_w1 + w1_bias
-      layer_w1 = adc.checkpoint_name(layer_w1, "moe_mlpwi_1")
+      if self.config.fused_mlp:
+        # Fuse wi_0 and wi_1: [G,K,N] + [G,K,N] -> [G,K,2N], one GEMM, split result.
+        w_fused = jnp.concatenate([w0, w1], axis=-1)
```
gobbleturk, line 1664: we may want an implementation that initializes the weight in concatted shape (G, K,2N), which guarantees there are no expensive weight HBM movements to concat. This is the format assumed by vllm-tpu for concatted weights which gets higher performance by avoiding this HBM concat - may be less of an important optimization for training. However this would be more consistent with the dense implementation as well https://github.com/AI-Hypercomputer/maxtext/blob/b117f50cf0489853ea6a9ce81ad2785a0fde3410/src/maxtext/layers/linears.py#L413 - which actually does a (K,2,N) format...

RissyRan (commented):
Thanks for the feature! Could you help add a test to compare output are the same with/without this flag? https://github.com/AI-Hypercomputer/maxtext/blob/b88ea632974fe49c42c721d0bf2891926dae0ca8/tests/unit/moe_test.py#L4

Also, could you help add this config in https://github.com/AI-Hypercomputer/maxtext/blob/b88ea632974fe49c42c721d0bf2891926dae0ca8/docs/reference/core_concepts/moe_configuration.md#L4

src/maxtext/layers/moe.py
RissyRan: That's a good point. The checkpoint conversion in the future will need to update a little bit accordingly if fine tuning workloads. I will help create a bug to track this feature and put in the backlog.

abhinavgoel95: Done

RissyRan (approved):
LGTM! Thanks! I see a few unit tests are failing. Could you try to rebase and see if it works? You may also need to work on format, which could be addressed by pre-commit.

gobbleturk (approved):

src/maxtext/layers/moe.py
```diff
@@ -410,7 +409,7 @@ def __init__(
       self.wi_0 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
       self.wi_1 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
       self.wo = jnp.zeros((num_experts, intermediate_dim, self.moe_expert_input_dim))
-    elif self.config.prefuse_moe_weights and self.config.attention == "vllm_rpa":
+    elif (self.config.prefuse_moe_weights and self.config.attention == "vllm_rpa") or self.config.fused_mlp:
```
NicoGrande: Should we consolidate `prefuse_moe_weights` and `fused_mlp` or are they doing different things?

gobbleturk: good catch! they are doing the same thing, altho perhaps we want separate flags for dense vs sparse MLP (some models have both)

gobbleturk: I think this PR should use prefuse_moe_weights instead of fused_mlp - we can keep both - the former for moe weights the latter for non moe-weights @abhinavgoel95 

in the future we can consider renaming these to be more similar

abhinavgoel95: Done.

src/maxtext/layers/moe.py
```diff
@@ -1412,29 +1406,44 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
...
-      if self.get_tensor_transpose_parallelism_size() > 1:
-        layer_w1 = jax.lax.psum(layer_w1, "tensor_transpose")
-      if self.config.mlp_bias:
-        layer_w1 = layer_w1 + w1_bias
-      layer_w1 = adc.checkpoint_name(layer_w1, "moe_mlpwi_1")
+      if self.config.prefuse_moe_weights:
+        # Weights are stored as (G,K,2N); w0/w1 are adjacent slices so XLA elides this concat.
+        w_fused = jnp.concatenate([w0, w1], axis=-1)
+        out = gmm_fn(x, w_fused, tiling=wi_tile_size, weight_gather_axes=wi_gather_axes)
+        n = w0.shape[-1]
```
CaptainO5: `w0.shape[-1]` is not always equal to `out.shape[-1] // 2`, which is causing shape mismatch errors downstream. This specifically occurs when running FP8 workloads and is due to an all gather operation in [ops.py#L169](https://github.com/AI-Hypercomputer/maxtext/blob/41a5d98f8501a9a5a44ada428d08032182f56bb8/src/maxtext/kernels/megablox/ops.py#L169) happening before the `ragged_dot`.

PR #3746, opened 2026-04-25
fix gemma4 dtype mismatch

aireenmei:
# Description

When fixing another bug in #3727 , I changed to use config.weight_dtype from config.dtype to initialize layer_scalar because it's a weight, this requires casting layer_scalar back to dtype during use time, or it causes error `The input   
  carry component c[1] has type bfloat16[2048,4096,2816] but the corresponding output carry component has type float32[2048,4096,2816], so the dtypes do not match.` when `weight_dtype=float32, dtype=bfloat16`

# Tests
[log with error](https://cloudlogging.app.goo.gl/uUPxUg7qFrHu5eGC7)
[working log](https://cloudlogging.app.goo.gl/88g2Ebn3FKU1jPiP9)

# Checklist

### DIFF
--- a/src/maxtext/models/gemma4.py
+++ b/src/maxtext/models/gemma4.py
@@ -370,7 +370,7 @@ class Gemma4DecoderLayer(nnx.Module):
 
     next_layer_addition = mlp_lnx + residual
     layer_output = next_layer_addition
-    layer_output = layer_output * self.layer_scalar.value
+    layer_output = layer_output * jnp.asarray(self.layer_scalar.value, cfg.dtype)
 
     layer_output = nn.with_logical_constraint(layer_output, self.activation_axis_names)
 


NuojCheng (approved):

shralex (approved):
typo in title dtye

src/maxtext/models/gemma4.py
```diff
@@ -370,7 +370,7 @@ def __call__(
 
     next_layer_addition = mlp_lnx + residual
     layer_output = next_layer_addition
-    layer_output = layer_output * self.layer_scalar.value
+    layer_output = layer_output * jnp.asarray(self.layer_scalar.value, cfg.dtype)
```
gagika, line 373: nit:  self.layer_scalar.value.astype(cfg.dtype)  probably matches better maxtext style.

gagika (approved):

PR #3749, opened 2026-04-26
Add dataset type olmo_grain for AI2 OLMo numpy pretrain mixes

gagika:
# Description

- Adds `dataset_type=olmo_grain`, a Grain-based input pipeline for AI2's
  pre-tokenized OLMo numpy mixes (e.g. [`OLMo-mix-0925-official.txt`](https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/data/mixes/OLMo-mix-0925-official.txt)).
  Reads headerless `.npy` token streams from a gcsfuse mount, applies
  OLMo-core's repeated-n-gram filter, and yields the shapes the MaxText
  pretrain trainer expects.
- Stateless sampler: record at step k is a pure function of
  `(seed, shard, k)`. Resume reads the latest step from
  `config.checkpoint_dir` and shifts the sampler — no Grain iterator
  state in the checkpoint.
- Ships two data tools (`download_olmo_data_to_gcs.py` with HTTP-Range
  resume; `build_olmo_npy_index.py` for header-scan indexing) and two
  launchers (`run_olmo3_7b_grain_smoke.sh`,
  `run_olmo3_7b_grain_resume_test.sh`).

# Tests
- Unit tests pass (`tests/unit/input_pipeline/olmo_*`)
- Smoke train: 50 steps, loss 11.99 → 8.93 on v4-8 (4-layer bf16)
- Resume test: Run B picks up at step 50 with loss continuity 8.931 → 8.930

# Checklist

authored-by: @aireenmei

### DIFF
--- a/docs/guides/data_input_pipeline/olmo_grain.md
+++ b/docs/guides/data_input_pipeline/olmo_grain.md
@@ -1,0 +1,76 @@
+# OLMo numpy pipeline (`dataset_type=olmo_grain`)
+
+Grain-based input pipeline for AI2's pre-tokenized OLMo data mixes (e.g.
+`OLMo-mix-0925-official.txt`). Reads headerless flat `.npy` token streams
+from a gcsfuse mount, shards across hosts, optionally masks repeated-n-gram
+instances, and yields the shapes the MaxText pretrain trainer expects.
+
+## Quick start
+
+1. **Download the data** to a GCS bucket. `--mix-file` is a local AI2 manifest listing relative npy paths to fetch from AI2's public bucket (e.g. `OLMo-mix-0925-official.txt` for the 6T pretrain mix or `OLMo-midtraining-mix-0625-100B.txt` for the 100B midtraining mix).
+
+   ```bash
+   python tools/data_generation/download_olmo_data_to_gcs.py \
+       --mix-file ./OLMo-mix-0925-official.txt \
+       --gcs-dest gs://my-bucket/dataset/ \
+       --staging-dir /mnt/local-ssd/olmo-staging \
+       --workers 16
+   ```
+
+2. **Mount it read-only** with gcsfuse (`np.memmap` needs a local path):
+
+   ```bash
+   gcsfuse --implicit-dirs --o ro my-bucket /mnt/olmo-readonly
+   ```
+
+3. **Build the index**:
+
+   ```bash
+   python tools/data_generation/build_olmo_npy_index.py \
+       --mix-file /path/to/OLMo-mix-0925-official.txt \
+       --gcs-base gs://my-bucket/dataset/ \
+       --tokenizer allenai/dolma3-tokenizer \
+       --sequence-length 8192 \
+       --output /path/to/olmo_index_seq8192.json
+   ```
+
+4. **Configure + run** the trainer:
+
+   ```yaml
+   dataset_type: olmo_grain
+   olmo_index_path: /path/to/olmo_index_seq8192.json
+   olmo_path_remap_from: "gs://my-bucket/"
+   olmo_path_remap_to:   "/mnt/olmo-readonly/"
+   max_target_length: 8192        # must equal index sequence_length
+   tokenizer_type: huggingface
+   tokenizer_path: allenai/Olmo-3-7B-Instruct
+   ```
+
+   See `scripts/run_olmo3_7b_grain_smoke.sh` for a runnable smoke launcher.
+
+## Resume
+
+Stateless sampler: record at step *k* is a pure function of `(seed, shard, k)`. On startup, the trainer adapter reads the latest step from
+`config.checkpoint_dir` and shifts the sampler so the data stream picks
+up where it left off — no Grain-iterator-state in the checkpoint.
+
+`scripts/run_olmo3_7b_grain_resume_test.sh` validates this end-to-end.
+
+## Notes
+
+- Files are headerless raw uint32 by default (matches AI2's published
+  format). The numpy `.npy` extension is misleading.
+- Documents may span instance boundaries; this matches OLMo-core.
+- `olmo_apply_ngram_filter: True` (default) zeroes loss on instances with
+  ≥ 32 repetitions of any 1–13-gram, per OLMo-core.
+- For mixing pretraining + midtraining, build a combined index by
+  concatenating the two .txt mix files.
+
+## Troubleshooting
+
+| Symptom                                                       | Fix                                                                                                       |
+| ------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| `OLMo index sequence_length=N but config.max_target_length=M` | Rebuild the index with `--sequence-length M`.                                                             |
+| `q_block_size=512 should divide q_seq_len=…`                  | Set `max_target_length` to a multiple of 512.                                                             |
+| OOM during compile on a small TPU                             | Shrink with `override_model_config=True base_num_decoder_layers=N`, use `weight_dtype=bfloat16`.          |
+| Resume restarts at step 0                                     | Iterator log should print `resumed_step=N initial_step=…`; if both 0, `checkpoint_dir` is empty or wrong. |
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -177,6 +177,7 @@ class DatasetType(str, Enum):
   GRAIN = "grain"
   TFDS = "tfds"
   C4MLPERF = "c4_mlperf"
+  OLMO_GRAIN = "olmo_grain"
 
 
 class SamplingStrategy(str, Enum):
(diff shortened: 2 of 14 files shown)


tests/unit/input_pipeline/olmo_data_grain_test.py
```diff
@@ -0,0 +1,353 @@
...
+          idx,
+          seed=0,
+          batch_size=4,
+          shard_index=0,
+          shard_count=1,
+          apply_ngram_filter=False,
+          # Disable shift so we can read the raw ``instance_id`` field for
+          # this audit-style test. (Production runs always use shift=True.)
+          shift_to_inputs_targets=False,
+          grain_worker_count=0,
```
aireenmei: Does this test aim to test grain_worker_count=2 ?

src/maxtext/input_pipeline/olmo_data_grain.py
```diff
@@ -0,0 +1,487 @@
...
+
+import hashlib
+import threading
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+
+# Grain is structured so that the public Python API lives under
+# ``grain.python``. Tests can monkeypatch the ``_grain`` module attribute.
+import grain.python as _grain  # type: ignore[import-not-found]
```
aireenmei: Can just be `import grain`

src/maxtext/input_pipeline/olmo_data_grain.py
```diff
@@ -0,0 +1,487 @@
...
+  if shift_to_inputs_targets:
+    ops.append(ShiftToInputsTargets())
+  ops.append(_grain.Batch(batch_size=batch_size, drop_remainder=True))
+
+  # Grain >=0.2.16 expects ``shard_options`` on the DataLoader (sharding used
+  # to live on the Sampler). Even though our sampler already does the
+  # shard-by-rank slicing, Grain still requires this object to validate
+  # checkpoint compatibility.
+  shard_options = _grain.ShardOptions(shard_index=shard_index, shard_count=shard_count, drop_remainder=True)
+  return _grain.DataLoader(
```
aireenmei: grain.DataLoader with sampler and a list of operations is the old API. Now it's recommended to use the new MapDataSource API (https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/input_pipeline/grain_data_processing.py#L94), and use dataset.<method()> to apply transformations: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/input_pipeline/grain_data_processing.py#L49

src/maxtext/input_pipeline/olmo_data_grain.py
```diff
@@ -0,0 +1,487 @@
...
+      arr = np.memmap(path, dtype=dtype, mode="r")
+      self._mmaps[path] = arr
+      while len(self._mmaps) > self._max:
+        # Evict oldest.
+        oldest = next(iter(self._mmaps))
+        del self._mmaps[oldest]
+      return arr
+
+
+class OlmoIndexSampler:
```
aireenmei, line 153: Why do we need a new Sampler? Can we re-use grain's sampler?

tests/unit/input_pipeline/olmo_data_grain_test.py
gagika: Good catch — fixed, now passing grain_worker_count=2.

src/maxtext/input_pipeline/olmo_data_grain.py
gagika: Done. Switched to import grain and updated call sites to the public submodule paths   (grain.sources.RandomAccessDataSource, grain.transforms.Map, grain.transforms.Batch, grain.sharding.ShardOptions).

src/maxtext/input_pipeline/olmo_data_grain.py
gagika: Coupled with next comment. The custom sampler powers stateless initial_step resume — pure function of (seed, shard, train_step *  batch_size), no Grain iterator-state checkpoint needed (pinned by olmo_data_grain_resume_test.py).  MapDataset.shuffle().repeat() has no step-offset hook, so migrating would drop that. We can do a follow follow-up fix if you think it's needed.

src/maxtext/input_pipeline/olmo_data_grain.py
gagika: Same as above comment.

docs/guides/data_input_pipeline/olmo_grain.md
```diff
@@ -0,0 +1,76 @@
...
+from a gcsfuse mount, shards across hosts, optionally masks repeated-n-gram
+instances, and yields the shapes the MaxText pretrain trainer expects.
+
+## Quick start
+
+1. **Download the data** to a GCS bucket:
+
+   ```bash
+   python tools/data_generation/download_olmo_data_to_gcs.py \
+       --mix-file /path/to/OLMo-mix-0925-official.txt \
```
RissyRan: Will this path be at GCS or not necessary?

src/maxtext/configs/types.py
```diff
@@ -1158,6 +1129,26 @@ class GrainDataset(BaseModel):
...
+class OlmoGrainDataset(BaseModel):
+  """Configuration for the OLMo numpy fixed-seq-length input pipeline (dataset_type=olmo_grain).
+
+  Worker count, per-worker buffer size, and shuffle seed reuse the standard
+  grain flags (``grain_worker_count``, ``grain_per_worker_buffer_size``,
+  ``data_shuffle_seed``); only OLMo-specific fields are listed here.
+  """
+
+  olmo_index_path: PathStr = Field("", description="Path or gs:// URI to the JSON index from build_olmo_npy_index.py.")
+  olmo_path_remap_from: PathStr = Field(
```
RissyRan, line 1177: Maybe add some comments and indicate why we need this?

RissyRan (commented):
Thanks for the change! 

Have you had a chance to chat with @aireenmei about this yet? I'm wondering if we could design these features to directly leverage the existing `grain_data_processing.py`. For example, we could add things like n-gram filtering and pre-tokenized numpy mixes as features there to improve code reusability.

The main benefit I see is reducing maintenance overhead. We currently maintain both tfds and c4_tfds_mlperf, but the latter is rarely used and has some maintenance issues. Since Grain will be heavily used moving forward, it makes sense to build on top of it directly. Let me know your thoughts—happy to discuss!

src/maxtext/input_pipeline/olmo_data.py
```diff
@@ -0,0 +1,415 @@
...
+  true_locs = np.where(prog[:-1:] == prog[1::])[0]
+  return int(true_locs[0] + 1)
+
+
+def _find_start_last_consecutive_true(arr: np.ndarray) -> int:
+  """Start offset of the trailing run of True in ``arr``, or -1 if none."""
+  reverse = _find_end_first_consecutive_true(arr[::-1])
+  return len(arr) - reverse if reverse > 0 else -1
+
+
```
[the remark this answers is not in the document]
gagika: Done.

src/maxtext/input_pipeline/olmo_data.py
```diff
@@ -0,0 +1,415 @@
...
+) -> Generator[RepetitionTuple, None, None]:
+  """Yield :class:`RepetitionTuple` for periodic spans of length ≥ 3 in
+  ``arr``. ``mask_value`` must not appear in the array (used internally as
+  reshape padding)."""
+  if (arr == mask_value).sum() > 0:
+    raise ValueError("`mask_value` is in the array")
+
+  max_period = min(max_period, len(arr) // 3)
+
+  for period in range(min_period, max_period + 1):
```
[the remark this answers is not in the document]
gagika: Updated the docstring: under uint32, -1 is the max uint32 value, way above any realistic vocab. Callers with vocabs that hit that id should pass an out-of-vocab sentinel.

docs/guides/data_input_pipeline/olmo_grain.md
gagika: local path (updating the doc).

src/maxtext/configs/types.py
gagika: Done, added comments.

gagika:
> Thanks for the change!
> 
> Have you had a chance to chat with @aireenmei about this yet? I'm wondering if we could design these features to directly leverage the existing `grain_data_processing.py`. For example, we could add things like n-gram filtering and pre-tokenized numpy mixes as features there to improve code reusability.
> 
> The main benefit I see is reducing maintenance overhead. We currently maintain both tfds and c4_tfds_mlperf, but the latter is rarely used and has some maintenance issues. Since Grain will be heavily used moving forward, it makes sense to build on top of it directly. Let me know your thoughts—happy to discuss!

I haven't yet chatted with @aireenmei, I will follow up.

The main reason I kept this separate is the stateless-resume contract — record at step k must be a pure function of `(seed, shard, k)` so we recover the data offset from the checkpoint step alone,  with no Grain iterator state in the checkpoint. `MapDataset.shuffle().repeat()` has no step-offset  hook, so reusing it as-is would drop that property.

The general pieces, e.g.  the n-gram mask, probably can be lifted into  `grain_data_processing.py` as a follow-up. Happy to take that on after this lands. 
I will chat with Aireen about it.

src/maxtext/input_pipeline/olmo_data_grain.py
```diff
@@ -0,0 +1,497 @@
...
+
+  Checkpointing is trivial: the only mutable state is "which epoch's
+  permutation is currently cached" (a perf optimization). The user-visible
+  position is just the index passed to ``__getitem__``.
+
+  Args:
+    total_instances: ``index.total_instances`` from the OLMo index.
+    seed: Base seed for the shuffle.
+    shard_index: Zero-based index of this data-loading host. Typically
+      ``jax.process_index()``.
```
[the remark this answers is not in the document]
gagika: at this scale it's fine, we can follow up with Chunked / Philox-keyed shuffle as follow-up if we need larger.

aireenmei (approved):

RissyRan (approved):

PR #3754, opened 2026-04-27
Add remove size one func to eval dataloader

NuojCheng:
# Description

Since `remove_size_one_mesh_axis_from_type` is tuned as False in batch split implementation, some redundant physical axes, e.g. context, data, are included in eval data iterators. This PR additionally enforces size one mesh axes get ignored when the eval data get generated.

# Tests

Repro in v5p-8 vm: https://paste.googleplex.com/5394820295163904

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #3761, opened 2026-04-28
Add device put for eval batch data

NuojCheng:
# Description

PR #3754 addressed the eval batch data sharding issue, but the fix only applied when `dataset_type=synthetic`. 

Although we previously removed redundant size-one mesh axes from the real data pipeline ([link](https://github.com/AI-Hypercomputer/maxtext/blob/02ebebef6e8acc3440a128416fcf96d9cd180904/src/maxtext/input_pipeline/input_pipeline_interface.py#L40-L42)), these axes inadvertently persist in the eval batch data sharding. To resolve this, this PR introduces a targeted sharding hint in `train.py` to explicitly enforce the removal of these size-one axes.

Additionally, this PR adds a small training smoke test with an active evaluation step to prevent future regressions in eval data loading.

# Tests

Added tests should protect changes in this PR.

`dataset_type=tfds` test: https://paste.googleplex.com/5101082608009216

# Checklist

### DIFF
--- a/tests/integration/smoke/train_smoke_test.py
+++ b/tests/integration/smoke/train_smoke_test.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" Smoke test """
+"""Smoke test"""
 import os
 import unittest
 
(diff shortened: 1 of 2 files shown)


shuningjin (approved):

tests/integration/smoke/train_smoke_test.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" Smoke test """
```
aireenmei, line 15: Looks like this test is skipped: https://github.com/AI-Hypercomputer/maxtext/blob/main/pytest.ini#L11.

shuningjin: The comment says: `g3 only`. Perhaps it is run as g3 test? Need double check
https://github.com/AI-Hypercomputer/maxtext/blob/dddb0cf7b35b3ee7b0fce83e838594d01e0f4dfa/tests/integration/smoke/train_smoke_test.py#L28

aireenmei (approved):
Looks like the g3 vs github unit tests are confusing. LGTM the PR to unblock internal workloads. We can figure out the best test file later.

PR #3652, opened 2026-04-13
[NNX] NNX migration prep (4/N): sharding tools and post-training fixes

ecnal-cienet:
## NNX Migration Route Map

1. ✅ Add NNX scaffolding: `pure_nnx` flag, `init_state_fn`, `TrainStateNNX`, NNX utils. Linen workflow unchanged. (PR #3427)
2. ✅ NNX sharding utilities: `get_abstract_state_nnx`, `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, `get_mesh_from_config`. (PR #3470)
3. ✅ NNX fully supported end-to-end: `TrainStateNNX`, model creation, gradient accumulation, checkpointing, and training loop dispatch. (PR #3500)
4. 🔄 **[This PR]** Sharding diagnostics on NNX, plus post-training bugfixes that surfaced once the NNX path got exercised end-to-end. Originally bundled with the Linen↔NNX checkpoint utilities under PR4; on 2026-05-07 those were split out into PR4.5 (converter) + PR4.6 (comparator) so each ships independently.
4.5. ❌ Linen↔NNX checkpoint converter.
4.6. ❌ Linen↔NNX checkpoint comparator.
5. ❌ NNX correctness fixes, feature enablements, and vocab tiling on NNX.
6. ❌ NNX-native DPO.
7. ❌ NNX-native MaxEngine inference.
8. ❌ NNX-native LoRA + GRPO.
9. ❌ NNX-aware QK-Clip + remaining checkpoint utilities.
9.5. ❌ NNX + AQT in MaxEngine + serve-mode reload + gpt3 prefill fix.
10. ❌ Vocab tiling `custom_vjp` for NNX.
11. ❌ Set NNX defaults to `True`; regenerate sharding goldens; flip back integration-test `pure_nnx=False` annotations.
12. ❌ Delete Linen-specific code paths and NNX compatibility flags.

## Description

This PR adds NNX-aware sharding diagnostics and a handful of post-training bugfixes that surfaced once the NNX path got exercised end-to-end (the SFT / distillation / RL trainers and the `adam_pax` optimizer hit them on NNX-shaped state).

[...]

PR #3500, opened 2026-03-25
NNX migration prep (3/N): TrainState, model creation, and end-to-end training loop

ecnal-cienet:
# NNX Migration Route Map
1. ✅ Add NNX scaffolding: `pure_nnx` flag, `init_state_fn`, `TrainStateNNX`, NNX utils. Linen workflow unchanged. ([PR #3427](https://github.com/AI-Hypercomputer/maxtext/pull/3427))
2. ✅ NNX sharding utilities: `get_abstract_state_nnx`, `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, `get_mesh_from_config`. ([PR #3470](https://github.com/AI-Hypercomputer/maxtext/pull/3470))
3. 🔄 **[This PR]** NNX fully supported end-to-end: `TrainStateNNX`, model creation, gradient accumulation, checkpointing, and training loop dispatch. ([PR #3500](https://github.com/AI-Hypercomputer/maxtext/pull/3500))
4. ✅ NNX sharding diagnostics, bidirectional Linen↔NNX checkpoint conversion utilities, and post-training fixes. ([PR #3652](https://github.com/AI-Hypercomputer/maxtext/pull/3652))
5. ❌ Enable NNX by default; fix unit and integration test failures.
6. ❌ Remove Linen-specific code paths and NNX compatibility flags.

# Description
> **Note:** This is the third in a series of NNX migration PRs. With this PR, `pure_nnx=True` runs a complete NNX training loop — initialization, sharding, gradient accumulation, eval, and checkpointing — without hitting any `NotImplementedError`. The `pure_nnx` flag still defaults to `False`, preserving the existing Linen workflow unchanged.

## `TrainStateNNX` and unit tests

`src/maxtext/layers/train_state_nnx.py` implements the `TrainStateNNX` container, which holds an NNX model and its Optax optimizer as a single composable unit. Unit tests cover state creation, optimizer step, and Orbax checkpoint round-trip:

- `tests/unit/train_state_nnx_test.py`
- `tests/unit/train_state_nnx_checkpoint_test.py`

## Muon optimizer and model creation utilities

- `muon_utils.py` — updated to support NNX models alongside Linen.
- `model_creation_utils.py` — refactored to expose `create_nnx_abstract_model` and `from_config`, which create and initialize an NNX model from a config without running a full forward pass.

## End-to-end training loop (`train.py` + supporting modules)

The core training loop in `train.py` now dispatches on `pure_nnx` at every major decision point:

[...]

PR #3470, opened 2026-03-20
NNX migration prep (2/N): NNX utils and sharding utilities

ecnal-cienet:
# NNX Migration Route Map
1. ✅ Add NNX scaffolding: `pure_nnx` flag, `init_state_fn`, `TrainStateNNX`, NNX utils. Linen workflow unchanged. ([PR #3427](https://github.com/AI-Hypercomputer/maxtext/pull/3427))
2. ✅ **[This PR]** NNX sharding utilities: `get_abstract_state_nnx`, `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, `get_mesh_from_config`. ([PR #3470](https://github.com/AI-Hypercomputer/maxtext/pull/3470))
3. ✅ NNX fully supported end-to-end: `TrainStateNNX`, model creation, gradient accumulation, checkpointing, and training loop dispatch. ([PR #3500](https://github.com/AI-Hypercomputer/maxtext/pull/3500))
4. ✅ NNX sharding diagnostics and bidirectional Linen↔NNX checkpoint conversion utilities. ([PR #3525](https://github.com/AI-Hypercomputer/maxtext/pull/3525))
5. ❌ NNX post-training fixes: MultimodalInput unpacking, scalar LR guard, nested NNX transform workaround.
6. ❌ Enable NNX by default; fix unit and integration test failures.
7. ❌ Remove Linen-specific code paths and NNX compatibility flags.

# Description
> **Note:** This is the second in a series of NNX migration PRs. Pure NNX training is **not yet implemented** — all NNX code paths currently raise `NotImplementedError`. This PR only introduces the structural scaffolding needed for subsequent patches to plug in NNX logic without modifying shared infrastructure.

- **NNX sharding utilities** (`maxtext_utils_nnx.py`) — Functions to manipulate NNX model shardings using abstract model state: `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, and memory movement helpers (`move_memory_to_host` / `move_memory_to_device`).
- **`get_abstract_state` NNX path** — Added `get_abstract_state_nnx` to `maxtext_utils.py`, which uses `nnx.get_abstract_model` to return a flat `nnx.State` (rather than a full `TrainStateNNX`), and updated `get_abstract_state` to dispatch to it when `pure_nnx=True`.
- **`maxtext_utils.get_mesh_from_config()`** — Extracted mesh creation into a standalone function with unit tests.
[...]

PR #3114, opened 2026-02-09
Feat/Decoder migration: DeepSeek/Gemma3/Gemma4/Llama4

hsuan-lun-chiang:
# Description

Implement and update the following models in NNX decoder that were are not supported in previous [PR 2831](https://github.com/AI-Hypercomputer/maxtext/pull/2831):

- DeepSeek
- Gemma 3, 4
- Llama4

# Tests

Test with different model and compare with Linen training. Details in the [GDoc file](https://docs.google.com/document/d/1NbUP3g5glgbC6bMyt44pwM_vQA1NR7U2rBUzfbTDwSs/edit?pli=1&resourcekey=0-9EUahtzL-hCycdu7l0grhQ&tab=t.htq5367h8au0)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #3766, opened 2026-04-28
[NNX] NNX migration prep (5/N): correctness fixes and feature enablements

ecnal-cienet:
## NNX Migration Route Map

1. ✅ Add NNX scaffolding: `pure_nnx` flag, `init_state_fn`, `TrainStateNNX`, NNX utils. Linen workflow unchanged. (PR #3427)
2. ✅ NNX sharding utilities: `get_abstract_state_nnx`, `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, `get_mesh_from_config`. (PR #3470)
3. ✅ NNX fully supported end-to-end: `TrainStateNNX`, model creation, gradient accumulation, checkpointing, and training loop dispatch. (PR #3500)
4. ✅ Sharding diagnostics on NNX, plus post-training bugfixes that surfaced once the NNX path got exercised end-to-end. (PR #3652)
4.5. ✅ Linen↔NNX checkpoint converter. (PR #3843)
4.6. ❌ Linen↔NNX checkpoint comparator (sibling branch on PR4.5; file-disjoint with this PR).
5. 🔄 **[This PR]** NNX correctness fixes, feature enablements, and vocab tiling on NNX. All "Pure NNX support has not been implemented yet" `NotImplementedError` sites cleared (17 → 0); vocab tiling on NNX is now a real feature, not a `NotImplementedError`. Stacks on PR4.5.
6. ❌ NNX-native DPO.
7. ❌ NNX-native MaxEngine inference.
8. ❌ NNX-native LoRA + GRPO.
9. ❌ NNX-aware QK-Clip + remaining checkpoint utilities.
9.5. ❌ NNX + AQT in MaxEngine + serve-mode reload + gpt3 prefill fix.
10. ❌ Vocab tiling `custom_vjp` for NNX.
11. ❌ Set NNX defaults to `True`; regenerate sharding goldens; flip back integration-test `pure_nnx=False` annotations.
12. ❌ Delete Linen-specific code paths and NNX compatibility flags.

# Description

The original plan was for PR5 to bundle the flag flip (`pure_nnx: True`) together with the correctness fixes and feature enablements that make NNX usable end-to-end. That bundle produced many UT failures and was hard to review. **This PR splits out the no-op-with-flags-False portion** so reviewers can evaluate bug fixes and feature enablements separately from the flag flip and sharding-golden regeneration that PR11 will own.

## Part 1: NNX correctness fixes

Real bug fixes that activate when `pure_nnx=True`. Each addresses a concrete crash or numerical bug observed during NNX trial runs.

| File | Fix |
|------|-----|
| `layers/nnx_wrappers.py` | Add `_refresh_variable_trace_state` + `is_linen_initializing`; called from `ToLinen` after `nnx.update`. Fixes `"Cannot extract graph node from different trace level"` when `jax.grad` tracers leak into `Variable._trace_state`. |
| `models/gpt_oss.py`, `models/olmo3.py` | Replace inline `nn.Dropout(...)()` with `self.dropout = linears.Dropout(...)` in `__init__`. Fixes `CallCompactUnboundModuleError`. |
| `layers/normalizations.py` | Accept `shard_mode`, `kernel_axes`, `parameter_memory_host_offload` kwargs on `Qwen3NextRMSNorm` for callsite parity with `RMSNorm`. |
| `models/models.py` | Build MTP block as `MultiTokenPredictionBlock(...)` directly (drop the `ToNNX(linen)` + `lazy_init` wrap); pass `multimodal_input` whole to `NNXDecoder` instead of unpacking 5 fields (undo the bad unpack from PR #3652). |
| `utils/gradient_accumulation.py` | ZeRO-1 + GA: defer the unreduced-→all-reduce annotation until after `lax.scan` (reduced/unreduced `PartitionSpec` is rejected inside scan carry). Use `nnx.merge(..., copy=True)` to avoid `Variable` reuse across re-traces. |
| `trainers/diloco/diloco.py` | NNX-aware state handling: `state.params` → `state.model.filter(nnx.Param)`, step counter at `state.optimizer.step`, `replace_nnx_model_params` helper for `jax.lax.cond` pytree-structure parity. |
| `trainers/pre_train/train_compile.py` | New `_collect_nnx_activation_shardings` helper (forward pass populates `_ACTIVATION_SHARDINGS_DUMP` — `get_abstract_state_nnx` only traces `__init__`). NNX path now passes 2-arg `shaped_train_args` (no rng); diloco path patched to handle the 2-vs-3 length difference. |
| `utils/muon_utils.py` | `get_model_mdn` default `pure_nnx=True`; wrap NNX result as `{"params": nnx.to_pure_dict(...)}` for parity with the Linen tree shape. |
| `layers/nnx_decoders.py` | **FP8 + NNX scan fix**: Linen FP8 ops (`fp8_nanoo`, `fp8_gpu`) retain JAX tracers in Linen scope across re-traces. Skip `jax.checkpoint` and use a Python for-loop instead of `jax.lax.scan` when `quantization in ("fp8_nanoo", "fp8_gpu")`. **Makes FP8 quantization usable on the NNX path.** |
| `trainers/pre_train/train.py` (NNX `train_step`) | Return `nnx.state(new_state, nnx.Not(nnx.Intermediate))` so sowed forward-pass artifacts (e.g. `max_logits` for QK-Clip) don't break leaf-count parity with `state_mesh_shardings`. |
| `models/llama2.py` | Pass `parameter_memory_host_offload` to `pre_self_attention_layer_norm` `RMSNorm` (was missing on this norm only — broke host-offload for Llama-2 NNX). |

## Part 2: NNX feature enablements (clear all `NotImplementedError` sites)

Clear all 17 `Pure NNX support has not been implemented yet` `NotImplementedError` sites by routing the Linen-coupled utilities to the Linen path internally. The on-disk format these utilities operate on is fundamentally Linen, so the `pure_nnx` flag (which affects training, not these tools) shouldn't gate them.

| File | Sites cleared | Why route to Linen |
|------|---------------|--------------------|
| `utils/layerwise_quantization.py` | 2 | Operates on Linen-format checkpoints via `DeepSeek*ToLinen` layer classes. |
| `utils/lora_utils.py` | 1 | Downstream `get_lora_abstract_state` expects Linen tree shape; LoRA adapters on disk are Linen. Logs a runtime warning when `pure_nnx=True` so users aren't silently downgraded. |
| `utils/standalone_checkpointer.py` | 2 | `add_entropy_to_checkpoint` accesses `state.opt_state[0]._replace(mu=..., nu=...)` — Linen-only. |
| `utils/generate_param_only_checkpoint.py` | 3 | `_possibly_unroll_params` and `_save_decode_checkpoint` use `state.params["params"]["decoder"]` — Linen tree. |
| `checkpoint_conversion/.../convert_gpt3_ckpt_from_paxml.py` | 2 | `keystr_map` targets Linen tree paths (`.params['params']`, `.opt_state.mu['params']`). |
| `inference/maxengine/maxengine.py` | 3 | Inference engine uses `transformer_as_linen` adapter so the surface stays Linen-shaped regardless of `pure_nnx`. |
| `experimental/rl/grpo_trainer.py` | 4 | RL trainer is end-to-end Linen-shaped (uses MaxEngine for sampling). Logs a runtime warning when `pure_nnx=True` since NNX-format checkpoints will fail at restore time. |

NNX-native versions of these utilities (DPO, MaxEngine, LoRA, GRPO, the remaining checkpoint utilities) are scheduled as follow-up PRs (PR6–PR9.5).

## Part 3: Vocab tiling on NNX (real implementation, not just routing)

| File | Change |
|------|--------|
| `models/models.py` | Add `Transformer.logits_from_hidden_states(hidden_states, deterministic, model_mode)` on the NNX `Transformer` class — wraps `NNXDecoder.apply_output_head(shared_embedding=self.token_embedder, ...)`. Mirrors `TransformerLinenPure.logits_from_hidden_states`. |
| `utils/vocabulary_tiling.py` | Add `vocab_tiling_nnx_loss(model, hidden_states, data, config, is_train)`. Chunks the vocab axis via `jax.lax.scan` and calls `model.logits_from_hidden_states(chunk)` per chunk. The NNX model carries its parameters internally so no explicit FSDP gather is needed (unlike the Linen `gathered_params` pattern). MVP uses default autograd; `custom_vjp` memory-savings optimization is scheduled as PR10. |
| `trainers/pre_train/train.py` (NNX `loss_fn`) | Replace `if config.num_vocab_tiling > 1: raise NotImplementedError` with `vocab_tiling_nnx_loss` dispatch using `hidden_states` from `nnx.Intermediate`. |
| `configs/pyconfig_deprecated.py`, `configs/types.py` | Drop the `num_vocab_tiling > 1 and enable_nnx` validation guards (no longer needed). |

## Deferred (intentionally out of scope)

- **DPO + NNX**: `train.py` `NotImplementedError` retained with a more informative message pointing users at the `pure_nnx=False` workaround. Full implementation requires a new TrainState shape carrying both policy and reference NNX models plus an NNX `dpo_loss_fn` — scheduled as PR6.
- **`nnx_decoders.py` larger structural rewrite** (`gemma4`, `deepseek_batchsplit_fp8`, `engram_layers`, `NNXSequentialPipelineStage`/`NNXScannedPipelineStage`): depends on [PR #3114](https://github.com/AI-Hypercomputer/maxtext/pull/3114). Reserved for a future PR.
- **`pipeline.py` 106 KB NNX pipeline rewrite + 4 pipeline-related `logical_axis_rules` in `base.yml`**: depends on [PR #2885](https://github.com/AI-Hypercomputer/maxtext/pull/2885). Reserved for PR11 (the flag-flip PR), where `use_nnx_pipeline=True` will land alongside.
- **`maxtext_utils.py` `PartitionSpec` → `P` alias rename**: pure cosmetic, no behavior change. Reserved for PR11.
- **`base.yml` flag flip (`pure_nnx: True`)**: PR11's whole purpose.

# Tests

- **Linen invariant verified**: `pure_nnx=False`, `enable_nnx=False`, `pure_nnx_decoder=False` still hold (no `base.yml` flag changes); `tests/unit/model_test.py` 3/3 still passes.
- **NNX path imports verified**: all 23 modified files import cleanly under `pure_nnx=True`. `Transformer.logits_from_hidden_states` and `vocab_tiling_nnx_loss` are exposed and reachable.
- **NotImplementedError audit**: `grep -c "Pure NNX support has not been implemented" src/` is now **0** (was 17).
- **Pre-existing failures left unchanged**: `sharding_compare_test::deepseek2-16b`, `optimizers_test::test_model_integration_kimi-k2-1t`, `diloco_test::two_slices` ×2 — all also fail on the parent branch HEAD without these changes (confirmed via `git stash`).
- **Stats**: 23 source files modified, +354 insertions, −128 deletions.

# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
+++ b/src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
@@ -87,11 +87,10 @@ def convert(paxml_ckpt_path, maxtext_model_name, base_output_directory, run_name
   devices_array = maxtext_utils.create_device_mesh(cfg)
   mesh = Mesh(devices_array, cfg.mesh_axes)
 
+  # Output is Linen-format (keystr_map below uses Linen tree paths). Route to
+  # Linen regardless of pure_nnx.
   quant = quantizations.configure_quantization(cfg)
-  if cfg.pure_nnx:
-    raise NotImplementedError("Pure NNX support has not been implemented yet.")
-  else:
-    model = transformer_as_linen(cfg, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
+  model = transformer_as_linen(cfg, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
   learning_rate_schedule = maxtext_utils.create_learning_rate_schedule(cfg)
   tx = optimizers.get_optimizer(cfg, learning_rate_schedule)
 
@@ -102,11 +101,7 @@ def convert(paxml_ckpt_path, maxtext_model_name, base_output_directory, run_name
       cfg.checkpoint_period,
   )
 
-  if cfg.pure_nnx:
-    # NNX has a different function to init the training state.
-    raise NotImplementedError("Pure NNX support has not been implemented yet.")
-  else:
-    init_state_fn = functools.partial(maxtext_utils.init_initial_state, model, tx, cfg, True, init_rng)
+  init_state_fn = functools.partial(maxtext_utils.init_initial_state, model, tx, cfg, True, init_rng)
   state, _, _, _ = maxtext_utils.setup_training_state(None, cfg, mesh, checkpoint_manager, init_state_fn)
   max_logging.log("start")
   max_utils.print_mem_stats("After params initialized")
--- a/src/maxtext/configs/pyconfig_deprecated.py
+++ b/src/maxtext/configs/pyconfig_deprecated.py
@@ -195,10 +195,9 @@ def validate_expert_shard_attention_option(expert_shard_attention_option: str) -
 
 
 def validate_vocab_tiling(num_vocab_tiling: int, per_device_batch_size: int, max_target_length: int, enable_nnx: bool):
+  del enable_nnx  # NNX vocab tiling supported via vocab_tiling_nnx_loss in vocabulary_tiling.py
   if (per_device_batch_size * max_target_length) % num_vocab_tiling != 0:
     raise ValueError("Per device batch size times sequence length should be divisible by the number of vocab tiles.")
-  if num_vocab_tiling > 1 and enable_nnx:  # TODO (chengnuojin) enable vocab tiling on NNX after NNX migration
-    raise ValueError("We currently don't support vocab tiling on NNX module.")
 
 
 def validate_rampup_batch_size(batch_size_start, batch_size_end, batch_size_increment, global_rampup_samples):
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -2959,8 +2959,6 @@ class MaxTextConfig(
         and (self.per_device_batch_size * self.max_target_length) % self.num_vocab_tiling != 0
     ):
       raise ValueError("Per device batch size times sequence length should be divisible by the number of vocab tiles.")
-    if self.num_vocab_tiling > 1 and self.enable_nnx:
-      raise ValueError("We currently don't support vocab tiling on NNX module.")
     if self.context_parallel_size > 1 and self.context_parallel_strategy.lower() == "ring":
       if "gpu" not in self.hardware:
         raise ValueError(
--- a/src/maxtext/inference/maxengine/maxengine.py
+++ b/src/maxtext/inference/maxengine/maxengine.py
@@ -111,12 +111,10 @@ class MaxEngine(_BaseEngine):
     devices_array = maxtext_utils.create_device_mesh(config=config, devices=devices)
     self._mesh = jax.sharding.Mesh(devices_array, config.mesh_axes)
 
-    # Model and Optimizer definition
+    # MaxEngine serves Linen-format inference checkpoints; the surface stays
+    # Linen-shaped via transformer_as_linen regardless of pure_nnx.
     quant = quantizations.configure_quantization(config)
-    if config.pure_nnx:
-      raise NotImplementedError("Pure NNX support has not been implemented yet.")
-    else:
-      self.model = models.transformer_as_linen(config, mesh=self._mesh, quant=quant, model_mode=MODEL_MODE_PREFILL)
+    self.model = models.transformer_as_linen(config, mesh=self._mesh, quant=quant, model_mode=MODEL_MODE_PREFILL)
     self.replicated_sharding = jax.sharding.NamedSharding(self._mesh, P(None))
 
     self.abstract_params = None
@@ -232,11 +230,7 @@ class MaxEngine(_BaseEngine):
     rng1, rng2, rng3 = jax.random.split(rng, 3)
     if params:
       print("Resharding given params")
-      if self.config.pure_nnx:
-        # NNX has a different function to init the training state.
-        raise NotImplementedError("Pure NNX support has not been implemented yet.")
-      else:
-        init_state_fn = functools.partial(maxtext_utils.init_initial_state, self.model, None, self.config, False, rng)
+      init_state_fn = functools.partial(maxtext_utils.init_initial_state, self.model, None, self.config, False, rng)
       _, self.state_mesh_annotations, state_mesh_shardings = maxtext_utils.get_abstract_state(
           self.config, self._mesh, init_state_fn, False
       )
@@ -245,11 +239,7 @@ class MaxEngine(_BaseEngine):
       state = maxtext_utils.init_decode_state(None, params)
       state = max_utils.unbox_logicallypartioned(state)
     else:
-      if self.config.pure_nnx:
-        # NNX has a different function to init the training state.
-        raise NotImplementedError("Pure NNX support has not been implemented yet.")
-      else:
-        init_state_fn = functools.partial(maxtext_utils.init_initial_state, self.model, None, self.config, False, rng1)
+      init_state_fn = functools.partial(maxtext_utils.init_initial_state, self.model, None, self.config, False, rng1)
       state, self.state_mesh_annotations = maxtext_utils.setup_decode_state(self.config, self._mesh, None, init_state_fn)
     # pylint: disable=isinstance-second-argument-not-valid-type
     self.abstract_params = jax.tree_util.tree_map(
--- a/src/maxtext/models/models.py
+++ b/src/maxtext/models/models.py
@@ -108,7 +108,7 @@ class TransformerLinenPure(nn.Module):
           rngs=self.make_rng("mtp_block"),
       )
 
-  def logits_from_hidden_states(self, hidden_states, deterministic, model_mode):
+  def logits_from_hidden_states_for_vocab_tiling(self, hidden_states, deterministic, model_mode):
     """
     Compute logits from hidden states (wrapping decoder.apply_output_head).
     This function is only used for vocabulary tiling.
@@ -397,6 +397,15 @@ class Transformer(nnx.Module):
   def no_op(self, *args, **kwargs):
     """A no-op method to allow the model to be used in a lazy context."""
     return
+
+  def logits_from_hidden_states_for_vocab_tiling(self, hidden_states, deterministic, model_mode):
+    """Computes logits from hidden states; used by vocabulary tiling."""
+    return self.decoder.apply_output_head(
+        shared_embedding=self.token_embedder,
+        y=hidden_states,
+        deterministic=deterministic,
+        model_mode=model_mode,
+    )
 
   def init_cache(self, cache_size: int, batch_size: int, dtype=jnp.float32):
     """Initializes the KV cache for the Transformer.
--- a/src/maxtext/trainers/pre_train/train_compile.py
+++ b/src/maxtext/trainers/pre_train/train_compile.py
@@ -29,6 +29,7 @@ from absl import app
 from flax import nnx
 from flax.linen import partitioning as nn_partitioning
 import jax
+import jax.numpy as jnp
 from jax.experimental.serialize_executable import serialize
 from jax.experimental.topologies import get_topology_desc
 from jax.sharding import AxisType, Mesh
@@ -91,6 +92,28 @@ def get_topology_mesh(config):
   return topology_mesh
 
 
+def _collect_nnx_activation_shardings(create_model_fn, config, mesh):
+  """Runs an abstract NNX forward pass to populate `_ACTIVATION_SHARDINGS_DUMP`.
+
+  `get_abstract_state_nnx` only traces `__init__`; activation shardings need
+  a forward pass to be collected.
+  """
+  input_shape = (config.micro_batch_size_to_train_on, config.max_target_length)
+  abstract_input = jax.ShapeDtypeStruct(input_shape, jnp.int32)
+
+  def _nnx_forward(decoder_input_tokens, decoder_positions, decoder_segment_ids):
+    model_instance = create_model_fn()
+    return model_instance(
+        decoder_input_tokens=decoder_input_tokens,
+        decoder_positions=decoder_positions,
+        decoder_segment_ids=decoder_segment_ids,
+        enable_dropout=False,
+    )
+
+  with jax.set_mesh(mesh), nn_partitioning.axis_rules(config.logical_axis_rules):
+    jax.eval_shape(_nnx_forward, abstract_input, abstract_input, abstract_input)
+
+
 def get_shaped_inputs(topology_mesh, config):
   """Get shaped abstractions of inputs to train_step: state, batch and rng"""
   # Construct the model and optimizer to get shaped versions of the state
@@ -128,7 +151,8 @@ def get_shaped_inputs(topology_mesh, config):
     # For NNX, get_functional_train_with_signature expects the graphdef (static structure),
     # not the raw model — mirroring how the training loop does nnx.split(train_state).
     with nn_partitioning.axis_rules(config.logical_axis_rules):
-      graphdef, _ = nnx.get_abstract_model(init_state_fn, topology_mesh)
+      abs_train_state = nnx.eval_shape(init_state_fn)
+      graphdef, _ = nnx.split(abs_train_state)
     model = graphdef
   else:
     # unsharded logical annotations
@@ -138,10 +162,16 @@ def get_shaped_inputs(topology_mesh, config):
   shaped_batch = maxtext_utils.get_shaped_batch(config)
 
   if config.pure_nnx:
-    shaped_train_args = (abstract_state, shaped_batch, None)  # NNX doesn't use dropout_rng
+    shaped_train_args = (abstract_state, shaped_batch)  # NNX doesn't use dropout_rng
   else:
     shaped_train_args = (abstract_state, shaped_batch, shaped_rng)
   shaped_train_kwargs = {}
+
+  # Collect NNX activation shardings via an abstract forward pass (must run
+  # after get_abstract_state, which only traces __init__).
+  if config.debug_sharding and config.pure_nnx:
+    _collect_nnx_activation_shardings(_create_model_partial, config, topology_mesh)
+
   return shaped_train_args, shaped_train_kwargs, state_mesh_shardings, logical_annotations, model
 
 
@@ -299,7 +329,9 @@ def main(argv: Sequence[str]) -> None:
     diloco_state, state_mesh_shardings, inner_state_shardings = diloco.build_abstract_diloco_state(
         config, abstract_state, state_mesh_shardings, topology_mesh
     )
-    shaped_train_args = (diloco_state, shaped_train_args[1], shaped_train_args[2])
+    # For NNX, shaped_train_args has 2 elements (state, batch) — no rng; pass None for prng.
+    shaped_rng_arg = shaped_train_args[2] if len(shaped_train_args) > 2 else None
+    shaped_train_args = (diloco_state, shaped_train_args[1], shaped_rng_arg)
 
     # Wrap train_step with diloco
     train_step_partial = functools.partial(train.train_step, model, config, inner_state_shardings, params_shardings)
--- a/src/maxtext/utils/gradient_accumulation.py
+++ b/src/maxtext/utils/gradient_accumulation.py
@@ -71,10 +71,18 @@ def gradient_accumulation_loss_and_grad(
 
   is_nnx = isinstance(model, nnx.Module)
 
-  # For more efficient DP/ZeRO-1 + GA
-  if config.shard_mode == ShardMode.EXPLICIT and config.ici_data_parallelism > 1:
-    ga_params_shardings = jax.tree.map(update_sharding_for_reduced, params_shardings)
-    grad_shardings = jax.tree.map(update_sharding_for_unreduced, params_shardings)
+  # For ZeRO-1 + GA, read the resolved "data" axis size from the mesh rather than
+  # config.ici_data_parallelism, which may be -1 (auto-fill) and resolves to 1 when
+  # FSDP already consumes every device — in which case data parallelism is not active.
+  param_mesh = jax.tree.leaves(params_shardings)[0].mesh
+  data_parallel_active = config.shard_mode == ShardMode.EXPLICIT and param_mesh.shape.get("data", 1) > 1
+  if data_parallel_active:
+    # reduced/unreduced PartitionSpecs are rejected inside a jax.lax.scan carry: scan
+    # traces its body against an AbstractMesh whose axis types are all Auto, and the
+    # annotations require Explicit axes. Keep plain params_shardings in the carry and
+    # apply the data-parallel all-reduce to the gradients after the scan instead.
+    ga_params_shardings = params_shardings
+    grad_shardings = params_shardings
   else:
     ga_params_shardings = grad_shardings = params_shardings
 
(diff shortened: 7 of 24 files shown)


src/maxtext/utils/vocabulary_tiling.py
```diff
@@ -247,3 +247,110 @@ def _bwd_scan_body(grad_params_acc, chunk_data):
...
+  Args:
+    model: The NNX model instance (must implement `logits_from_hidden_states`).
+    hidden_states: The final hidden states from the decoder.
+    data: A dictionary containing the input data, including 'targets' and 'targets_segmentation'.
+    config: The model and training configuration.
+    is_train: A boolean indicating if the model is in training mode.
+
+  Returns:
+    A tuple (total_loss, total_z_loss).
+  """
```
bvandermoon, line 272: @NuojCheng can you please review the vocab tiling changes?

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
```diff
@@ -87,11 +87,12 @@ def convert(paxml_ckpt_path, maxtext_model_name, base_output_directory, run_name
...
+  # This conversion script reads paxml-format weights and emits a Linen-format
+  # MaxText checkpoint (downstream uses `.params['params']`, `.opt_state.mu['params']`,
+  # `.opt_state.nu['params']` keystr paths; the keystr_map below targets the Linen
+  # tree shape). Use the Linen path regardless of pure_nnx.
   quant = quantizations.configure_quantization(cfg)
-  if cfg.pure_nnx:
-    raise NotImplementedError("Pure NNX support has not been implemented yet.")
-  else:
-    model = transformer_as_linen(cfg, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
+  model = transformer_as_linen(cfg, mesh, quant=quant, model_mode=MODEL_MODE_TRAIN)
```
bvandermoon: Since we decided to make the checkpoints equivalent between Linen and NNX, should we keep the NNX version now too? Should this script support NNX?

src/maxtext/configs/base.yml
```diff
@@ -559,6 +559,13 @@ logical_axis_rules: [
                       ['tokens_per_page', []],
                       ['paged_kv_head_dim_size', []],
                       # ==========================================
+                      # Pipeline Parallelism
+                      # ==========================================
+                      ['layers_outside_pipeline', []],
+                      ['layers_per_stage', []],
+                      ['num_activations', []],
+                      ['circular_repeats', []],
```
bvandermoon: Why are these being added here?

src/maxtext/configs/pyconfig_deprecated.py
```diff
@@ -195,10 +195,9 @@ def validate_expert_shard_attention_option(expert_shard_attention_option: str) -
 
 
 def validate_vocab_tiling(num_vocab_tiling: int, per_device_batch_size: int, max_target_length: int, enable_nnx: bool):
+  del enable_nnx  # NNX vocab tiling supported via vocab_tiling_nnx_loss in vocabulary_tiling.py
```
bvandermoon, line 198: Why is this being deleted?

src/maxtext/inference/maxengine/maxengine.py
```diff
@@ -111,12 +111,12 @@ def __init__(self, config: Any, devices: Any | None = None):
     devices_array = maxtext_utils.create_device_mesh(config=config, devices=devices)
     self._mesh = jax.sharding.Mesh(devices_array, config.mesh_axes)
 
-    # Model and Optimizer definition
+    # Model and Optimizer definition.
+    # MaxEngine uses Linen-shaped state (state.params, state_mesh_shardings.params,
+    # state.opt_state) and serves Linen-format inference checkpoints. Use Linen path
+    # regardless of pure_nnx — the flag affects training, not inference serving.
```
bvandermoon: Is the plan to not support NNX for MaxEngine?

src/maxtext/layers/attentions.py
```diff
@@ -525,14 +525,14 @@ def __init__(
     elif self.is_qwen3_hybrid:
       self.query_norm = Qwen3NextRMSNorm(
           num_features=self.config.head_dim,
-          eps=self.config.normalization_layer_epsilon,
+          epsilon=self.config.normalization_layer_epsilon,
```
bvandermoon: Why is this variable name being changed? I am fine with it, but is there a reason?

src/maxtext/layers/moe.py
```diff
@@ -2243,9 +2243,9 @@ def __call__(
       w0_kernel = jnp.asarray(self.wi_0[...], self.dtype)
       w1_kernel = jnp.asarray(self.wi_1[...], self.dtype)
 
-    # Only apply per expert scales if we have not fused with the out-projections at init time.
-    if self.per_expert_scale is not None and cfg.model_call_mode != "inference" and not cfg.fuse_expert_scales:
-      wo_kernel = wo_kernel * jnp.asarray(self.per_expert_scale[...], self.dtype)[:, None, None]
+      # Only apply per expert scales if we have not fused with the out-projections at init time.
+      if self.per_expert_scale is not None and cfg.model_call_mode != "inference" and not cfg.fuse_expert_scales:
+        wo_kernel = wo_kernel * jnp.asarray(self.per_expert_scale[...], self.dtype)[:, None, None]
```
bvandermoon: Is this indentation change intentional? Seems we aren't setting `wo_kernel` in places we were before

bvandermoon (commented):
Overall question: Since we decided to make the checkpoints equivalent between Linen and NNX, should we keep the `NotImplementedError` messages for the NNX flag now also?

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
ecnal-cienet: This script outputs Linen-format checkpoints (the keystr_map below uses Linen tree paths like `.params['params']['decoder']`). Route to Linen regardless of pure_nnx.

src/maxtext/configs/base.yml
ecnal-cienet: These rules are for the NNX pipeline rewrite and have no consumer in PR5. Removed.

src/maxtext/inference/maxengine/maxengine.py
ecnal-cienet: NNX is supported — MaxEngine uses `transformer_as_linen` so the surface stays Linen regardless of pure_nnx. This block just picks AQT quant mode: `serve` only when the checkpoint is already quantized, otherwise `train`.

src/maxtext/layers/attentions.py
ecnal-cienet: Reverted.

src/maxtext/layers/moe.py
ecnal-cienet: You're right, this was a format regression. Fixed.

src/maxtext/configs/pyconfig_deprecated.py
ecnal-cienet: This PR implements NNX vocab tiling via `vocab_tiling_nnx_loss` in vocabulary_tiling.py. The guard would block a working feature now.

ecnal-cienet:
> Overall question: Since we decided to make the checkpoints equivalent between Linen and NNX, should we keep the `NotImplementedError` messages for the NNX flag now also?

We need to examine them case by case:
- Checkpoint utilities (convert_gpt3, generate_param_only, layerwise_quantization, standalone_checkpointer): output is Linen by design, so route-to-Linen is correct. No NotImplementedError needed.
- maxengine: NNX supported via transformer_as_linen adapter. No NotImplementedError.
- grpo_trainer, lora_utils: end-to-end Linen-shaped today. grpo logs a warning, lora should too.
- DPO in train.py: kept as NotImplementedError — the TrainState shape isn't there yet, real work needed.

src/maxtext/configs/types.py
```diff
@@ -2901,8 +2901,7 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
         and (self.per_device_batch_size * self.max_target_length) % self.num_vocab_tiling != 0
     ):
       raise ValueError("Per device batch size times sequence length should be divisible by the number of vocab tiles.")
-    if self.num_vocab_tiling > 1 and self.enable_nnx:
-      raise ValueError("We currently don't support vocab tiling on NNX module.")
+    # Vocab tiling on NNX is now supported via vocab_tiling_nnx_loss in vocabulary_tiling.py.
```
NuojCheng: Let's remove this comment in types.py

src/maxtext/models/models.py
```diff
@@ -398,6 +398,15 @@ def no_op(self, *args, **kwargs):
     """A no-op method to allow the model to be used in a lazy context."""
     return
 
+  def logits_from_hidden_states(self, hidden_states, deterministic, model_mode):
```
NuojCheng: could you rename this function to `logits_from_hidden_states_for_vocab_tling`

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
bvandermoon: Thanks @ecnal-cienet. But based on our last discussion, aren't we trying to make the NNX checkpoints match the Linen format? Let me know if I am misunderstanding something here

src/maxtext/trainers/pre_train/train_compile.py
```diff
@@ -91,6 +92,27 @@ def get_topology_mesh(config):
...
+  def _nnx_forward():
+    model_instance = create_model_fn()
+    return model_instance(
+        decoder_input_tokens=jnp.ones(input_shape, dtype=jnp.int32),
+        decoder_positions=jnp.ones(input_shape, dtype=jnp.int32),
+        decoder_segment_ids=jnp.ones(input_shape, dtype=jnp.int32),
+        enable_dropout=False,
+    )
+
+  with nn_partitioning.axis_rules(config.logical_axis_rules):
```
NuojCheng: could you do 
```
with jax.set_mesh(mesh), nn_partitioning.axis_rules(config.logical_axis_rules):
```

src/maxtext/trainers/pre_train/train_compile.py
```diff
@@ -91,6 +92,27 @@ def get_topology_mesh(config):
...
+  a forward pass to be collected.
+  """
+  input_shape = (config.micro_batch_size_to_train_on, config.max_target_length)
+
+  def _nnx_forward():
+    model_instance = create_model_fn()
+    return model_instance(
+        decoder_input_tokens=jnp.ones(input_shape, dtype=jnp.int32),
+        decoder_positions=jnp.ones(input_shape, dtype=jnp.int32),
+        decoder_segment_ids=jnp.ones(input_shape, dtype=jnp.int32),
```
NuojCheng: I wonder if we could use `jax.ShapeDtypeStruct` instead of `jnp.ones()`

src/maxtext/trainers/pre_train/train_compile.py
```diff
@@ -138,10 +161,16 @@ def create_train_state_fn():
...
-    shaped_train_args = (abstract_state, shaped_batch, None)  # NNX doesn't use dropout_rng
+    shaped_train_args = (abstract_state, shaped_batch)  # NNX doesn't use dropout_rng
   else:
     shaped_train_args = (abstract_state, shaped_batch, shaped_rng)
   shaped_train_kwargs = {}
+
+  # Collect NNX activation shardings via an abstract forward pass (must run
+  # after get_abstract_state, which only traces __init__).
+  if config.debug_sharding and config.pure_nnx:
+    _collect_nnx_activation_shardings(_create_model_partial, config, topology_mesh)
```
NuojCheng, line 173: what if config.pure_nnx is false, is `_ACTIVATION_SHARDINGS_DUMP` gets populated without doing anything extra?

src/maxtext/utils/gradient_accumulation.py
```diff
@@ -71,10 +71,16 @@ def _maybe_shard_with_name(inputs, sharding_names):
 
   is_nnx = isinstance(model, nnx.Module)
 
-  # For more efficient DP/ZeRO-1 + GA
-  if config.shard_mode == ShardMode.EXPLICIT and config.ici_data_parallelism > 1:
-    ga_params_shardings = jax.tree.map(update_sharding_for_reduced, params_shardings)
-    grad_shardings = jax.tree.map(update_sharding_for_unreduced, params_shardings)
+  # For more efficient DP/ZeRO-1 + GA.
+  # config.ici_data_parallelism may be -1 (auto-fill: resolved at mesh creation time, but
+  # the config field remains -1). Treat any value != 1 as "data parallelism is active".
```
NuojCheng: it is also possible that num_device=8 and ici_fsdp_parallelism=8 and ici_data_parallelism=-1, in which case, DP is not active.

src/maxtext/utils/gradient_accumulation.py
```diff
@@ -71,10 +71,16 @@ def _maybe_shard_with_name(inputs, sharding_names):
...
+  if config.shard_mode == ShardMode.EXPLICIT and config.ici_data_parallelism != 1:
+    # jax.lax.scan traces its body with an AbstractMesh where all axis types are Auto,
```
NuojCheng: how this is related to lax.scan? I don't quite follow what this comment mean here...

src/maxtext/utils/gradient_accumulation.py
```diff
@@ -156,6 +162,11 @@ def reshape_to_microbatch_accumulations(batch_arr):
       + grad_and_loss["mtp_loss"] / config.gradient_accumulation_steps
   )
   raw_grads = grad_and_loss["grad"]
+  if config.shard_mode == ShardMode.EXPLICIT and config.ici_data_parallelism != 1:
+    # Apply unreduced annotation after the scan to trigger all-reduce across data-parallel
+    # devices (reduced/unreduced cannot be used inside jax.lax.scan carry tensors).
```
NuojCheng: could you elaborate more on 
> reduced/unreduced cannot be used inside jax.lax.scan carry tensors

It might be a JAX bug if that is the case

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
ecnal-cienet: Right, that's exactly what #3929 does — load-time interop across all 4 combos via `load_parameters_path`. The reason this script stays Linen-only is a bit different from "the formats aren't equivalent":

This is a converter that *writes* a checkpoint, and it writes a full-state checkpoint (it maps `.opt_state.mu` / `.opt_state.nu` / `count` from the Paxml ckpt). An NNX run can already load the *params* from this Linen output via `load_parameters_path` (#3929 handles the conversion). Loading the *full state* cross-framework is intentionally out of scope in #3929 — optax internal state doesn't translate 1:1 — so making this script emit NNX format wouldn't unlock anything; the full-state path wouldn't work either way.

The `keystr_map` here is also inherently Linen (`.params['params']`, `.opt_state.mu['params']`), so pure_nnx doesn't really apply to the converter itself.

src/maxtext/configs/types.py
ecnal-cienet: Done, removed.

src/maxtext/models/models.py
ecnal-cienet: Renamed.

src/maxtext/trainers/pre_train/train_compile.py
ecnal-cienet: Done

src/maxtext/trainers/pre_train/train_compile.py
ecnal-cienet: Done

src/maxtext/trainers/pre_train/train_compile.py
ecnal-cienet: Yes, Linen populates it with no extra work: `init_initial_state` runs `model.init` under `eval_shape`, which traces `__call__`, so `maybe_shard_with_name` fills the dump as a side effect. NNX's `get_abstract_state_nnx` uses `nnx.eval_shape` which only traces `__init__`, so the dump is empty without a forward pass — that's why this helper is gated on `pure_nnx`.

src/maxtext/utils/gradient_accumulation.py
ecnal-cienet: Fixed, thank you.

src/maxtext/utils/gradient_accumulation.py
ecnal-cienet: Reworded. The point: scan traces its body against an AbstractMesh with all-Auto axis types, which rejects reduced/unreduced PartitionSpecs in the carry. So we keep plain shardings in the carry and apply the DP all-reduce after the scan.

src/maxtext/utils/gradient_accumulation.py
ecnal-cienet: Not a bug, by design. reduced/unreduced needs Explicit axes, but scan traces its body with an all-Auto AbstractMesh, so they're rejected in the carry. We apply the unreduced("data") annotation after the scan to trigger the all-reduce. Reworded the comment to spell this out.

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
bvandermoon: What will happen once the Linen code is removed and there is no more `transformer_as_linen `?

src/maxtext/utils/gradient_accumulation.py
NuojCheng: Does it mean jax.lax.scan does not work under explicit shard mode + using reduced shard labels?

NuojCheng (approved):

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
ecnal-cienet: Thanks for raising this, it's a good point about the long-term direction.

I'd love to make sure I'm doing the right thing here: would you prefer this script to emit an NNX-format checkpoint now, or is it ok to handle it as a follow-up?

If you'd like it now, I'm happy to do it — I'd rewrite keystr_map to target the NNX param tree instead of the Linen one (the paxml->maxtext mapping stays the same, just the output shape changes). The only thing is it would grow this PR's scope a little, since the rest is interim routing.

If a follow-up works better, my plan is: when transformer_as_linen is removed, I'll switch this script to build the NNX model directly and retarget keystr_map as part of the Linen-removal work. NNX users wouldn't be blocked in the meantime, since #3929 can already load this Linen output.

Happy to go either way — just let me know what you prefer!

bvandermoon (approved):

src/maxtext/checkpoint_conversion/standalone_scripts/convert_gpt3_ckpt_from_paxml.py
bvandermoon: Sounds good, it can be switched to NNX later on. Thanks @ecnal-cienet

PR #3779, opened 2026-04-29
Support specifying tokamax gmm tile sizes in MaxText

darisoy:
### Description
This PR enables cleanly overriding the GMM tile sizes for both the forward and backward passes of Tokamax (`ragged_dot`) in MaxText using an elegant global heuristics monkey-patching approach.

**Proposed Changes**

* **Heuristics Monkey-Patching (`src/maxtext/layers/moe.py`)**: Instead of introducing complex custom `vjp` function wrappers or polluting model layers with low-level compilation code, this PR dynamically monkey-patches `PallasMosaicTpuRaggedDot._get_heuristics_config` globally during RoutedMoE initialization.
* **Dynamic Shapes & Pass Routing**: The monkey-patch dynamically inspects the shapes of the operands (`lhs`/`rhs`) and the JAX dimension numbers. This allows it to detect the active pass (**Forward**, **Backward DLHS**, or **Backward DRHS**) and layer type (**Weight Input `wi`** or **Weight Output `wo`**) on the fly, applying your exact manual tile configurations:
  * `wi_tile_fwd`, `wi_tile_dlhs`, `wi_tile_drhs` configs.
  * `wo_tile_fwd`, `wo_tile_dlhs`, `wo_tile_drhs` configs.

FIXES: b/506157856

### Tests

Verified compilation and trace collection via a 5-step synthetic pretraining run on a `v6e-4` VM using the `deepseek3-671b` MoE model configuration with your manual tiling overrides:

```bash
JAX_PLATFORMS=tpu,cpu PYTHONPATH=src python3 -m maxtext.trainers.pre_train.train \
    run_name=darisoy-moe-train \
    base_output_directory=gs://maxtext-experiments-multipod \
    model_name=deepseek3-671b \
    dataset_type=synthetic \
    steps=5 \
    enable_checkpointing=False \
    enable_goodput_recording=False \
    attention_type=mla \
    sparse_matmul=True \
    megablox=True \
    per_device_batch_size=2 \
    base_num_decoder_layers=1 \
    first_num_dense_layers=0 \
    shared_experts=1 \
    base_mlp_dim=2048 \
    profiler=xplane \
    use_tokamax_gmm=True \
    override_model_config=True \
    max_target_length=4096 \
    ici_expert_parallelism=4 \
    use_ring_of_experts=True \
    tokamax_gmm_autotune=True \
    weight_dtype=bfloat16 \
    opt_type=sgd \
    wi_tile_fwd_batch_seq=128 \
    wi_tile_fwd_embed_dim=128 \
    wi_tile_fwd_mlp_dim=128 \
    wi_tile_dlhs_batch_seq=256 \
    wi_tile_dlhs_embed_dim=256 \
    wi_tile_dlhs_mlp_dim=256 \
    wi_tile_drhs_batch_seq=512 \
    wi_tile_drhs_embed_dim=512 \
    wi_tile_drhs_mlp_dim=512 \
    wo_tile_fwd_batch_seq: 128 \
    wo_tile_fwd_embed_dim: 128 \
    wo_tile_fwd_mlp_dim: 128 \
    wo_tile_dlhs_batch_seq: 256 \
    wo_tile_dlhs_embed_dim: 256 \
    wo_tile_dlhs_mlp_dim: 256 \
    wo_tile_drhs_batch_seq: 512 \
    wo_tile_drhs_embed_dim: 512 \
    wo_tile_drhs_mlp_dim: 512
```

* **Execution Status**: Completed successfully with no shape or compilation errors.
* **Xprof Trace URL**: [https://xprof.corp.google.com/trace_viewer/darisoy-440103416108113419](https://xprof.corp.google.com/trace_viewer/darisoy-440103416108113419)

See details in http://b/506157856

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -48,6 +48,57 @@ import qwix
 from qwix.contrib.sparsity import sparsity_module
 import qwix.pallas as qpl
 import tokamax
+from tokamax._src.ops.ragged_dot.pallas_mosaic_tpu import (
+    PallasMosaicTpuRaggedDot,
+    Config,
+    DEFAULT_RAGGED_DOT_DIM_NUMS,
+    DLHS_RAGGED_DOT_DIM_NUMS,
+    DRHS_RAGGED_DOT_DIM_NUMS,
+)
+from tokamax._src.ops.ragged_dot import base
+import dataclasses
+
+
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass(frozen=True, kw_only=True, slots=True)
+class PallasMosaicTpuRaggedDotCustom(PallasMosaicTpuRaggedDot):
+  """A custom PallasMosaicTpuRaggedDot subclass that overrides _get_heuristics_config."""
+
+  config: Config | None = None
+  fwd_tile: tuple[int, int, int] = (128, 128, 128)
+  dlhs_tile: tuple[int, int, int] = (128, 128, 128)
+  drhs_tile: tuple[int, int, int] = (128, 128, 128)
+
+  def __post_init__(self):
+    qdtype = self.qdtype if self.qdtype is None else jnp.dtype(self.qdtype).name
+    if self.vjp is None:
+
+      def fn(*args, **kw):
+        # pylint: disable=unexpected-keyword-arg
+        return PallasMosaicTpuRaggedDotCustom(
+            qdtype=qdtype,
+            interpret=self.interpret,
+            fwd_tile=self.fwd_tile,
+            dlhs_tile=self.dlhs_tile,
+            drhs_tile=self.drhs_tile,
+        )(*args, **kw)
+
+      object.__setattr__(
+          self,
+          "vjp",
+          functools.partial(base.vjp, dlhs_ragged_dot=fn, drhs_ragged_dot=fn),
+      )
+
+  def _get_heuristics_config(self, ba) -> Config:
+    dims = ba.arguments.get("ragged_dot_dimension_numbers", DEFAULT_RAGGED_DOT_DIM_NUMS)
+    if dims == DEFAULT_RAGGED_DOT_DIM_NUMS:
+      return Config(tile_m=self.fwd_tile[0], tile_k=self.fwd_tile[1], tile_n=self.fwd_tile[2])
+    elif dims == DLHS_RAGGED_DOT_DIM_NUMS:
+      return Config(tile_m=self.dlhs_tile[0], tile_k=self.dlhs_tile[1], tile_n=self.dlhs_tile[2])
+    elif dims == DRHS_RAGGED_DOT_DIM_NUMS:
+      return Config(tile_m=self.drhs_tile[0], tile_k=self.drhs_tile[1], tile_n=self.drhs_tile[2])
+    return Config()
+
 
 set_xla_metadata = xla_metadata.set_xla_metadata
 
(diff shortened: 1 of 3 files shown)


gobbleturk:
elegant global heuristics monkey-patching approach - hmmm =p

src/maxtext/layers/moe.py
```diff
@@ -1137,9 +1147,10 @@ def get_tokamax_group_sizes(group_sizes, inputs, kernel):
       elif self.config.attention == "vllm_rpa":
         return group_sizes
       else:
+        ep = self.get_expert_parallelism_size()
         return tokamax.RaggedDotGroupSizes(
             group_sizes,
-            (inputs.shape[0] // kernel.shape[0],) * kernel.shape[0],
+            (inputs.shape[0] // kernel.shape[0] // ep,) * kernel.shape[0],
```
gobbleturk: I would consider wrapping this one line (line 1153) into its own function get_gmm_group_sizes() - which at the very least allow us to put a comment on it (potentially in the future makes it easy to support both random routing and heavily unbalanced routing too, as well as ragged buffer factor support..)

src/maxtext/layers/moe.py
```diff
@@ -2541,3 +2554,74 @@ def get_routed_and_shared_moe(
...
+  if _heuristics_patched and not force:
+    return
+
+  def custom_heuristics(self, ba: op.BoundArguments) -> Config:
+    lhs, rhs = ba.arguments["lhs"], ba.arguments["rhs"]
+    dims = ba.arguments.get("ragged_dot_dimension_numbers", DEFAULT_RAGGED_DOT_DIM_NUMS)
+
+    is_wo = False
+    if dims == DEFAULT_RAGGED_DOT_DIM_NUMS:
+      is_wo = rhs.shape[1] == config.base_mlp_dim
```
gobbleturk: unfortunately this logic is very brittle - it will fail if the mlp and/or embed dimension is sharded or chunked (or if the mlp and embed dimension are equal)

I am not sure of a good solution - my imagined solution was to monkey patch twice, once for gmm_ups and another for gmm_down

src/maxtext/layers/moe.py
```diff
@@ -2541,3 +2554,74 @@ def get_routed_and_shared_moe(
...
+            tile_n=config.wi_tile_drhs_mlp_dim,
+        )
+
+    return Config()
+
+  # Apply class-level monkey patch!
+  # pylint: disable=protected-access
+  PallasMosaicTpuRaggedDot._get_heuristics_config = custom_heuristics
+  _heuristics_patched = True
+  print("[TOKAMAX_PATCH] Successfully monkey-patched Tokamax GMM heuristics globally!")
```
gobbleturk: remove debugging statement?

src/maxtext/layers/moe.py
```diff
@@ -2541,3 +2554,74 @@ def get_routed_and_shared_moe(
...
+            tile_k=config.wi_tile_drhs_embed_dim,
+            tile_n=config.wi_tile_drhs_mlp_dim,
+        )
+
+    return Config()
+
+  # Apply class-level monkey patch!
+  # pylint: disable=protected-access
+  PallasMosaicTpuRaggedDot._get_heuristics_config = custom_heuristics
+  _heuristics_patched = True
```
gobbleturk: this variable _heuristics_patched is unused? just debugging can be removed?

src/maxtext/layers/moe.py
gobbleturk: If monkey patching twice proves too complicated we can keep this logic for now with a TODO(bug) to improve this.

@Shuwen-Fang is working on a collective matmul feature which will chunk up the embed dimension so this is of practical concern

src/maxtext/layers/moe.py
gobbleturk: nvm I see it is somehow used, global variables are confusing. Perhaps some comment somewhere - I guess this somehow ensures monkey patching only occurs once?

gobbleturk (commented):
Lets try to find a more robust way to differentiate wi and wo - my idea is to monkey patch twice

gmm_up = monkey_patch_up(PallasMosaicTpuRaggedDot)
gmm_down = monkey_patch_down(PallasMosaicTpuRaggedDot) - see also [comment](https://github.com/AI-Hypercomputer/maxtext/pull/3779/changes#r3336557341)

src/maxtext/layers/moe.py
```diff
@@ -1139,6 +1219,11 @@ def jax_ragged_dot_gmm(inputs, kernel, tiling, group_sizes, expert_assignments,
         output *= scales
       return output
 
+    def get_gmm_group_sizes(inputs, kernel, ep):
+      # Calculates perfectly balanced group sizes where each local expert receives an equal
```
gobbleturk, line 1195: can you add that this function assumes the inputs are ragged and padded to worst case size (generally a factor of EP larger than balanced)? This is why we must divide by EP (in fact dividing by EP is not always correct here, but for now it is good enough)

src/maxtext/models/deepseek_batchsplit_fp8.py
```diff
@@ -944,13 +948,93 @@ def unroute(
 def compute(x, w0, w1, wo, group_sizes, weights, *, config, mesh):
   """Processes routed tokens through the MLP."""
 
+  class PallasMosaicTpuRaggedDotWI(PallasMosaicTpuRaggedDot):
```
gobbleturk: can we not re-use the classes defined in moe.py? Is quantization different here?

cc @shuningjin

src/maxtext/layers/moe.py
darisoy: done

src/maxtext/layers/moe.py
darisoy: done

gobbleturk (approved):
Looks good, lets have the tokamax team take a look as well

gobbleturk (commented):
Note we currently don't support passing in input_buffer_count - perhaps this can be a separate add if we find significant lift. I think this defaults to 2 (with or without this change?)

src/maxtext/layers/moe.py
darisoy: no longer doing global monkey patching

src/maxtext/layers/moe.py
```diff
@@ -1084,6 +1133,22 @@ def sparse_matmul(
       wo_bias,
   ):
     """Perform sparse matrix multiplication of inputs and Experts."""
+    config = self.config
+
+    gmm_impl_wi = PallasMosaicTpuRaggedDotCustom(
+        qdtype=None,
```
gobbleturk: does this only support bf16?

src/maxtext/layers/moe.py
darisoy: done

PR #3803, opened 2026-05-03
Fix CI warnings and CodeCov

shralex:
Resolves observed DeprecationWarnings and UserWarnings across unit and integration tests. Fixed the Codecov CI pipeline.

Key Changes:

- Flax NNX Modernization: Replaced all deprecated .value accesses on nnx.Variable instances with the indexing syntax [...] or .get_value() throughout the maxtext models and test suites.
- Sharding Annotations: Swapped out the deprecated sharding keyword parameter for the newly supported out_sharding explicitly in nnx.Cache and nnx.Param instantiations to align with standard JAX naming conventions.
- Pytree State Refactor: Updated internal state tracking logic from _object__state to the modern _pytree__state.
- JAX UserWarnings: Mitigated JAX integer truncation warnings in the DeepSeek engram.py layer by enforcing explicit jnp.int32 definitions for token and hash index allocations.
- Repaired the CodeCov integration that recently ceased running to restore code coverage tracking.

# Tests

Tested on TPU VM

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -848,7 +848,7 @@ class HardwareAndMesh(BaseModel):
       description="Strategy for context parallelism ('all_gather' or 'ring').",
   )
   context_parallel_reorder_strategy: ReorderStrategy = Field(
-      "auto",
+      ReorderStrategy.AUTO,
       description="Reorder strategy for load-balanced context parallelism.",
   )
   custom_mesh: str = Field("", description="Available options: ['hybrid_ring_64x4', 'hybrid_ring_32x8']")
--- a/src/maxtext/layers/engram.py
+++ b/src/maxtext/layers/engram.py
@@ -15,7 +15,7 @@
 """
 DeepSeek-AI, `Conditional Memory via Scalable Lookup: A New Axis of Sparsity for Large Language Models
   <https://arxiv.org/pdf/2601.07372>`_, 2026
-  
+
 Reference implementation: https://github.com/deepseek-ai/Engram/blob/main/engram_demo_v1.py
 """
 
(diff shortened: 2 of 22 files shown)


igorts-git (approved):

src/maxtext/configs/types.py
```diff
@@ -847,7 +847,7 @@ class HardwareAndMesh(BaseModel):
       "all_gather",
       description="Strategy for context parallelism ('all_gather' or 'ring').",
   )
-  context_parallel_reorder_strategy: ReorderStrategy = Field(
+  context_parallel_reorder_strategy: str = Field(
```
ecnal-cienet: LGTM

Weakening the type from `ReorderStrategy` to `str` silently breaks the equality checks downstream. With `ConfigDict(...)` not setting `validate_default=True`, Pydantic only coerces strings → enum on explicit user input — so before this PR, `Config(f="striped").f` was `ReorderStrategy.STRIPED` and comparisons worked; after, it stays as the raw string and equality is False.

I think this minimun fix might work:
```python
context_parallel_reorder_strategy: ReorderStrategy = Field(
    ReorderStrategy.AUTO, description="..."
)
```

richjames0 (approved):
lgtm

src/maxtext/layers/engram.py
```diff
@@ -53,7 +53,7 @@ class CompressedTokenizer:
   def __init__(self, tokenizer: HFTokenizer):
     normalizer = self._build_normalizer()
     self.lookup_table_np, self.num_new_token = self._build_lookup_table(tokenizer, normalizer)
-    self.lookup_table = jnp.array(self.lookup_table_np, dtype=jnp.int64)
```
RissyRan, line 56: If I recall correctly for previous runs, the reference [test](https://github.com/AI-Hypercomputer/maxtext/blob/1adb45ce021c39103c0266854189f47f79c88386/tests/unit/engram_vs_reference_test.py#L173) will fail with different type. 

I am fine to update int64 to int32, as currently Engram is not used in OSS models. For future debugging purpose, could we create a bug and track this int64 changes in lines? So we could convert it back in the future

RissyRan (approved):
Thanks Alex for the fix! Just one comment about int64 change.

bvandermoon (approved):

PR #3813, opened 2026-05-05
Support to_huggingface checkpoint conversion for Gemma4

aireenmei:
# Description

hf to_maxtext checkpoint conversion is already supported. Adding support for the other direction. The following changes are included to reduce frictions during the conversion:
* Previously param mapping dict won't include multimodal keys, when `use_multimodal=False`, this cause error if the provided maxtext ckpt contains multimodal keys, while the user only want to convert the text-related weights to hf ckpt. I improved this logic to have param_mapping always load the full dict, but keep or skip multimodal keys later depends on `use_multimodal` flag
* Adjust moe tile size in the scripts for 26b, so it fits the memory of v5p-8
* Add hint when scan_layers config doesn't match the ckpt

b/508269624

# Tests

For gemma4-26b-a4b-it model:
* `26b/test_gemma4_to_hf.sh`: the script runs `to_huggingface.py` for maxtext -> hf ckpt conversion, then validate the conversion by `forward_pass_logit_checker.py`. [text-only results](https://paste.googleplex.com/4627989443772416)
* The above results are comparable to the results from `convert_gemma4.sh`, which runs `to_maxtext.py` for hf -> maxtext ckpt conversion, then validate by `forward_pass_logit_checker.py` [text-only results](https://paste.googleplex.com/4853267222888448)

For gemma4-31b:
* `31b/test_gemma_to_hf.sh`: [with image results](https://paste.googleplex.com/4895915426054144)

# Checklist

### DIFF
--- a/tests/end_to_end/tpu/gemma4/26b/test_gemma4_to_hf.sh
+++ b/tests/end_to_end/tpu/gemma4/26b/test_gemma4_to_hf.sh
@@ -1,0 +1,103 @@
+#!/bin/bash
+
+# This script is both an end-to-end test and documentation for converting a
+# Gemma4-26B MaxText checkpoint to Hugging Face format. Can be run on a v5p-8.
+
+# The flow of this script is as follows:
+# 1. Convert a MaxText checkpoint to a Hugging Face model checkpoint.
+# 2. Run a forward pass check to compare the logits and KL divergence between
+#    the converted checkpoint and the original HF model.
+
+# Pre-requisites:
+# 1. Set HF_TOKEN environment variable to your Hugging Face access token.
+#    export HF_TOKEN=<Hugging Face access token>
+# 2. Provide a MaxText-format Gemma4-26B checkpoint via CKPT_PATH.
+#    One can be produced with tests/end_to_end/tpu/gemma4/26b/convert_gemma4.sh.
+
+set -ex
+idx=$(date +%Y-%m-%d-%H-%M)
+MODEL_NAME='gemma4-26b'
+export MODEL_VARIATION='26b-it'
+# To convert the multimodal model, set USE_MULTIMODAL=true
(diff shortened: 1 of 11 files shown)


tests/end_to_end/tpu/gemma4/26b/test_gemma4_to_hf.sh
```diff
@@ -0,0 +1,103 @@
...
+        per_device_batch_size=1 \
+        dtype=float32 \
+        wi_tile_fwd_embed_dim=512 \
+        wi_tile_fwd_mlp_dim=512 \
+        wo_tile_fwd_embed_dim=512 \
+        wo_tile_fwd_mlp_dim=512 \
+        --max_kl_div=0.1 \
+        --run_hf_model=true \
+        --hf_model_path=${LOCAL_PATH}
+fi
```
hengtaoguo, line 103: Just learned that `-it` checkpoints should work with [chat_template](https://github.com/AI-Hypercomputer/maxtext/blob/7d6e1caca4f16da8c9eb6d77ead22706c77a1b1d/src/maxtext/inference/vllm_decode.py#L128) in order to have reasonable decode outputs. As a follow up, we could add some vllm_decode commands to this scripts later.

hengtaoguo (approved):

gagika (approved):
Thanks!

tests/end_to_end/tpu/gemma4/26b/test_gemma4_to_hf.sh
aireenmei: I actually tested with template, but [results](https://paste.googleplex.com/4733826950627328) doesn't look great either. I think it's because "-it" is for using question as prompt and our prompt is for sentence completion.

PR #3814, opened 2026-05-05
Onboard ragged sorting kernels from tpu inference to maxtext

NuojCheng:
# Description

This PR onboards two critical kernels: `ragged_gather` and `ragged_gather_reduce`. 

The key difference lies in their routing and output shapes:
*   **`ragged_gather`**: Performs simultaneous permutation and fan-out (`num_tokens x emb_dim` → `num_tokens x top_k x emb_dim`).
*   **`ragged_gather_reduce`**: Adds an accumulation step for fan-in (`num_tokens x top_k x emb_dim` → `num_tokens x emb_dim`).

During the forward pass, we use `ragged_gather` for dispatch and `ragged_gather_reduce` for combine. In the backward pass, these roles are swapped. 

Four APIs are onboarded in `ragged_sort.py`, namely `ring_ragged_sort`/`ring_ragged_unsort` and `a2a_ragged_sort`/`a2a_ragged_unsort`.  They are used for different flavors of EP (ring of experts or all2all) and equipped with fwd/bwd using these 2 kernels for training purpose.

FIXES: b/496676734

# Tests

We observe performance advantage when ragged sort is enabled, while also notice some regression issue between JAX==0.10.0 and 0.10.1.

Toy example v5p-8, EP=4: https://diff.googleplex.com/#key=iIgJcV78QY4B

More benchmark results in b/496676734.

# Checklist


src/maxtext/configs/base.yml
```diff
@@ -218,6 +218,9 @@ load_balance_loss_weight: 0.0 # weight for the load balance loss
 use_random_routing: false # whether to use random routing for debug/test purpose
 use_custom_sort_vjp: true # whether to use a custom VJP sort for efficient backward pass processing in sparse matmul
 use_ring_of_experts: false # whether to use ring of experts for sparse matmul expert parallelism
+use_ragged_sort: false # whether to use the Pallas ragged-sort kernels in the MoE permute path; valid both with and
```
gobbleturk, line 221: local-unpermute as opposed to un-local-permute?

src/maxtext/layers/moe.py
```diff
@@ -741,22 +742,35 @@ def permute(self, inputs, gate_logits, pre_bias_logits, use_custom_sort_vjp=True
...
+      topk_indices_2d = jnp.reshape(selected_experts, (bsz_times_seq_len, selected_experts.shape[2]))
+      sorted_inputs, group_size, sorted_selected_experts = gather_tokens_locally(
+          inputs_2d,
+          topk_indices_2d,
+          self.config.num_experts,
+          self.num_experts_per_tok,
+          self._expert_parallelism_name,
+          num_expert_parallelism,
+      )
+    else:
```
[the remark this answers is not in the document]
gobbleturk: hmm is the robot correct here? It does look like we need to use some information about expert_parallelism_index to be correct here (roll to expert id uses this information). Do you have correctness tests with this?

src/maxtext/layers/moe.py
```diff
@@ -883,7 +925,15 @@ def local_permute(
...
 
     sorted_indices = jnp.argsort(expert_indices)
-    sorted_inputs = _sort_activations(inputs, sorted_indices, use_custom_sort_vjp)
+    if use_ragged_sort:
+      # Only the first `valid_end` rows of `inputs` carry actual tokens for
+      # this shard (`local_group_size.sum()`), the remainder is padding from
+      # the worst-case ragged buffer. Restricting the gather to that prefix
+      # makes both forward and backward proportional to the routed token count.
+      valid_end = jnp.sum(local_group_size).astype(jnp.int32)
+      sorted_inputs = a2a_ragged_sort(inputs, sorted_indices, valid_end)
```
gobbleturk, line 935: amazingly simple, nice!

tests/unit/moe_test.py
```diff
@@ -603,8 +603,124 @@ def test_ring_of_expert_and_tensor_parallelism(self):
       actual_output, _, _ = self.get_moe_output(variables, hidden_states, cfg, mesh)
       self.assertTrue(jax.numpy.allclose(expected_output, actual_output, rtol=1e-02, atol=1e-02, equal_nan=False))
 
+  def _run_ragged_sort_loss_and_grad(self, use_ring_of_experts: bool):
```
gobbleturk, line 606: Amazing test, thank you!

tests/unit/moe_test.py
```diff
@@ -603,8 +603,124 @@ def test_ring_of_expert_and_tensor_parallelism(self):
...
+    and inputs. Both the scalar loss and the full pytree of parameter
+    gradients must match within bf16 tolerance.
+    """
+
+    def _build_cfg(use_ragged_sort: bool):
+      return pyconfig.initialize(
+          [None, get_test_config_path()],
+          run_name=(f"moe_block_use_ragged_sort_{use_ragged_sort}" f"_ring_{use_ring_of_experts}_test"),
+          enable_checkpointing=False,
+          model_name="mixtral-8x7b",
```
gobbleturk, line 620: this is a big model, should we use something smaller - how long does this test take?

src/maxtext/layers/moe.py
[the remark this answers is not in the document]
gobbleturk: it looks like there are correctness tests - can you explain in a comment somewhere where the expert_parallelism_index was used in this flow, its not clear to me (or maybe somehow its not needed?)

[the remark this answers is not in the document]
gobbleturk: oh its probably used within the kernel itself?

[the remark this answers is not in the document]
gobbleturk: ok kind of answered my own question poking around in the kernel - perhaps just a comment here something like the index information is used in the kernel itself

gobbleturk (approved):
Amazing work lets go! No more spurious data movement bottlenecking these great chips!

tests/unit/moe_test.py
NuojCheng: They do take long time 
```
68.16s call     tests/unit/attention_test.py::AttentionTest::test_dot_product_reshape_q
65.27s call     tests/unit/attention_test.py::AttentionTest::test_dot_product_cache_axis_order
53.27s call     tests/unit/moe_test.py::RoutedMoeTest::test_ragged_sort_loss_and_grad_no_ring_of_experts
38.18s call     tests/unit/moe_test.py::RoutedMoeTest::test_ragged_sort_loss_and_grad_ring_of_experts
```
I will make them smaller.

src/maxtext/layers/moe.py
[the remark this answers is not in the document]
NuojCheng: correct, expert id is called inside `_ring_ragged_sort_fwd`, see 
```
shard_idx = jax.lax.axis_index(ep_name)
```

src/maxtext/configs/base.yml
NuojCheng: done

src/maxtext/configs/types.py
```diff
@@ -702,6 +703,9 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  use_ragged_sort: bool = Field(
```
RissyRan, line 706: Shall we add a validation in types.py for `only valid when EP is enabled`?

src/maxtext/kernels/ragged/ragged_sort.py
```diff
@@ -0,0 +1,399 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Ragged token sorting operations with custom VJP."""
+
+import jax
```
RissyRan, line 17: Having a standard JAX implementation in the kernels folder is a bit confusing, as this acts more like a layer utility. I considered suggesting we move it to moe.py, but that file is already getting quite large. I'm on the fence about whether we should keep it here or extract it into a separate layer file. Let me know if you have a preference or a better idea!

RissyRan (approved):
Thanks for the improvement!  Just a few minor comments

src/maxtext/kernels/ragged/ragged_sort.py
NuojCheng: I agree we should keep only Pallas kernels in `kernels/` and relocate the pure JAX code. At the same time, `moe.py` is definitely getting bloated. We could move the MoE pure JAX helpers to `utils/moe_utils.py` (or `utils/moe/ragged_sort.py`), similar to our approach with `utils/pipeline_utils.py`.

Given there are a few other pure JAX files inside `kernels/`, it might be best to migrate all of them to `utils/` in a separate PR.

src/maxtext/configs/types.py
NuojCheng: It is a little bit tricky since during types.py there is no single static flags indicating the real EP rank. Using `ici_expert_parallelism` is one way but in some of my proposed custom mesh and rules, EP is practically effective even when `ici_expert_parallelism=1`, e.g. using `cp-as-ep`.

Alternatively, we have access to real EP ranks when initializing RoutedMoE module and using its `get_expert_parallelism_size()` method. I can add a check here, and the program will terminate during the JAX compilation.

NuojCheng: In fact ragged_sort should work for both EP=1 and EP>1. Let me update the description instead of disabling it when EP=1.

PR #3835, opened 2026-05-07
Fix decoupled tests, refactor ICI parallelism override removal

gulsumgudukbay:
# Description

Fix decoupled MaxText tests so they can run against the local decoupled test configuration without per-test ICI parallelism overrides.
Previously, many tests patched `ici_fsdp_parallelism` dynamically through a shared test helper. This made decoupled behavior harder to reason about and spread device-count-specific setup across unrelated tests. This PR moves that behavior into `decoupled_base_test.yml` by inheriting from `base.yml` and using config-level parallelism defaults (using @darisoy's changes in https://github.com/AI-Hypercomputer/maxtext/commit/dbbc69e17b87dc061f16226a2c889539b0be136c), then removes the ad-hoc helper from unit and integration tests.
The follow-up ROCm-specific fixes keep unsupported ROCm paths isolated: parameter host offload is skipped on ROCm, ROCm GPU architecture strings are not parsed as CUDA compute capabilities, and packed Transformer Engine attention passes the THD metadata required by newer TE versions while preserving compatibility with older versions.
This keeps the upstream-facing behavior focused on decoupled/ROCm test correctness rather than changing general test configs.

# Tests
```
export PYTHONPATH=$PWD/src
export DECOUPLE_GCLOUD=TRUE
pytest -m 'not cpu_only and not tpu_only and not post_training and decoupled' -v tests --ignore=tests/post_training --csv=decoupled-tests-report.csv --html=decoupled-tests-report.html --self-contained-html 2>&1 | tee maxtext_decoupled_UT.log
```

# Checklist


darisoy (approved):

src/maxtext/input_pipeline/tfds_data_processing.py
darisoy: duplicate line

shralex (approved):

PR #3837, opened 2026-05-07
[Test only] Maxtext side changes for parallel replica

lepan-google:


src/maxtext/common/checkpointing.py
```diff
@@ -256,6 +256,28 @@ def create_orbax_checkpoint_manager(
...
+      array_handler = ocp.type_handlers.SingleReplicaArrayHandler(
+          replica_axis_index=0,
+          broadcast_memory_limit_bytes=1024 * 1024 * 1000,  # 1000 MB limit
+          use_replica_parallel=False,
+      )
+      if checkpoint_use_replica_parallel is not None:
+        max_logging.log(
+            f"Overriding use_replica_parallel to {checkpoint_use_replica_parallel} for SingleReplicaArrayHandler based on checkpoint_use_replica_parallel flag."
+        )
+        array_handler.use_replica_parallel = checkpoint_use_replica_parallel
```
orbax-dev, line 269: Just specify it when creating the handler in the first place? I don't see why it should be overridden like this.

PR #3871, opened 2026-05-11
[DeepSeek-V4] Implement MoE routing primitives

parambole:
Description
===========
Implement Mixture of Experts (MoE) routing primitives and infrastructural updates required for DeepSeek-V4 integration into MaxText. 

-   **DeepSeek Routing Primitives:** Integrated token-based hash routing logic and DeepSeek-specific Top-K routing directly into `moe.py`'s `get_topk` function.
-   **Safe Layer Parameters:** Updated `RoutedMoE` and `RoutedAndSharedMoE` to accept `layer_idx` and `input_ids` via optional keyword arguments. This allows DeepSeek to pipe these variables down into the hashing logic, while safely defaulting to `0` and `None` so existing callers are entirely unimpacted.
-   **Hardware Optimization Bypasses:** Added boundary checks to bypass the TPU `fused_moe_matmul` (via the `vllm_rpa` block) specifically for hash-routed layers where it is unsupported.
-   **Activation Functions:** Registered the `sqrtsoftplus` custom activation function in `linears.py` required by the DeepSeek V4 MoE router.
-   **Unit test suite** (`tests/unit/deepseek_v4_vs_reference_test.py`): Validates MoE routing parity against PyTorch reference implementations.

Tests
=====
Tested on CPU. All DeepSeek V4 and core MoE unit tests pass successfully. 
```bash
pytest tests/unit/deepseek_v4_vs_reference_test.py tests/unit/moe_test.py
============================= test session starts ==============================
collecting ... 
collected 43 items
tests/unit/deepseek_v4_vs_reference_test.py ............                 [ 27%]
tests/unit/moe_test.py ...............................                   [100%]
================ 25 passed, 18 skipped, 123 warnings in 47.50s =================
```

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -264,6 +264,7 @@ routed_bias: false # a flag if a learnable bias is added for routing
 routed_bias_update_rate: 0.0 # a flag indicate the update rate applied to the router bias term
 mlp_bias: false # a flag if a learnable bias is added for MLP matmul, and originally implemented to support the GPT-OSS model architecture.
 n_routing_groups: -1 # number of groups for routing, disabled by default
+first_num_hash_layers: 0 # number of hash routing layers, used in DeepSeek V4 (0 means disabled)
 topk_routing_group: -1 # number of top groups to route inputs. For EP,
 # Splits the batch to allow for better scheduling when using expert parallelism by overlapping the
 # all-to-all communication with compute. Currently only implemented with DeepSeek sparse layers.
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -227,6 +227,7 @@ ModelName = Literal[
     "deepseek3-test",
     "deepseek3-tiny",
     "deepseek3.2-671b",
+    "deepseek4",
     "deepseek-custom",
     "kimi-k2-1t",
     "gemma-7b",
@@ -499,6 +500,7 @@ class ModelArchitecture(BaseModel):
       -1.0,
       description="Upper bound to clip the MLP activation values. -1.0 means no clipping.",
   )
+
   normalization_layer_epsilon: float = Field(1.0e-05, description="Epsilon value for normalization layers.")
   fused_qkv: bool = Field(False, description="If supported, fuse the Q, K, and V projections.")
   attention_bias: bool = Field(
@@ -836,6 +838,9 @@ class DeepSeekMoE(BaseModel):
       "and originally implemented to support the GPT-OSS model architecture",
   )
   n_routing_groups: int = Field(-1, description="Number of groups for routing, disabled by default.")
+  first_num_hash_layers: int = Field(
+      0, description="Number of hash routing layers, used in DeepSeek V4 (0 means disabled)."
+  )
   topk_routing_group: int = Field(-1, description="Number of top groups to route inputs to.")
   use_batch_split_schedule: bool = Field(
       False,
@@ -2972,6 +2977,8 @@ class MaxTextConfig(
         raise ValueError("GPT-OSS MoE only supports dropless (capacity_factor=-1) with dense matmul.")
       if self.routed_bias and self.routed_bias_update_rate > 0.0 and self.decoder_block != DecoderBlockType.DEEPSEEK:
         raise ValueError("Loss-free load balancing is only supported for the DeepSeek decoder block.")
+      if self.model_name.startswith("deepseek4") and self.first_num_hash_layers > 0 and self.use_ring_of_experts:
+        raise ValueError("DeepSeek V4 hash routing is currently not supported with ring of experts.")
       self.validate_ragged_buffer_factor()
 
     # Gemma 4 small (E2B / E4B) uses per-layer KV sharing, which is incompatible with nn.scan.
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -336,8 +336,10 @@ class GateLogit(nnx.Module):
 
     if self.score_func:
       output = linears._convert_to_activation_function(self.score_func)(output)
-      if self.model_name.startswith("deepseek3"):
-        pre_bias_logits = output
+
+    # NOTE: deepseek2 has a different pattern
+    if self.model_name.startswith(("deepseek3", "deepseek4")):
+      pre_bias_logits = output
 
     if self.use_bias:
       bias = jnp.asarray(self.bias[...], self.dtype)
@@ -361,6 +363,7 @@ class RoutedMoE(nnx.Module):
       weight_dtype: ctypes.DType = jnp.float32,
       dtype: ctypes.DType = jnp.float32,
       quant: Optional[quantizations.AqtQuantization] = None,
+      is_hash_routing: bool = False,
   ):
     """Initializes the RoutedMoE module.
 
@@ -376,6 +379,7 @@ class RoutedMoE(nnx.Module):
       weight_dtype: The data type of the kernel weights.
       dtype: The data type for the computation.
       quant: The quantization configuration. If None, no quantization is applied.
+      is_hash_routing: Whether this layer uses deterministic hash routing instead of top-K routing.
     """
     self.config = config
     self.num_experts = num_experts
@@ -388,6 +392,17 @@ class RoutedMoE(nnx.Module):
     self.dtype = dtype
     self.quant = quant
     self.rngs = rngs
+    self.is_hash_routing = is_hash_routing
+
+    # DeepSeek V4 Hash Routing
+    if self.is_hash_routing:
+      # Token-ID to Expert-ID lookup table for static routing
+      self.tid2eid = nnx.Variable(
+          jnp.zeros((self.config.vocab_size, self.num_experts_per_tok), dtype=jnp.int32),
+          out_sharding=None,  # Replicated across shards for local lookup
+      )
+    else:
+      self.tid2eid = None
 
     self.moe_expert_input_dim = (
         self.config.emb_dim if self.config.moe_expert_input_dim <= 0 else self.config.moe_expert_input_dim
@@ -427,7 +442,7 @@ class RoutedMoE(nnx.Module):
         quant=self.quant,
         kernel_init=self.kernel_init,
         kernel_axes=self.kernel_axes,
-        use_bias=self.config.routed_bias,
+        use_bias=self.config.routed_bias and not self.is_hash_routing,
         # tpu-inference applies the score function in the fused_moe_gmm kernel,
         # so we don't apply it here to avoid redundant computation.
         # See https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/fused_moe_gmm.py#L58.
@@ -628,10 +643,14 @@ class RoutedMoE(nnx.Module):
     return self.mesh.shape.get("context_autoregressive", 1)
 
   def should_update_load_balance(self):
-    """Determines if loss-free load balancing updates should be applied."""
-    return self.config.routed_bias and self.config.routed_bias_update_rate > 0.0
+    """Determines if loss-free load balancing updates should be applied.
 
-  def get_topk(self, gate_logits, pre_bias_logits, rngs=None):
+    The bias update logic is only applicable to Top-K router.
+    Hash router does not use routed bias.
+    """
+    return self.config.routed_bias and self.config.routed_bias_update_rate > 0.0 and not self.is_hash_routing
+
+  def get_topk(self, gate_logits, pre_bias_logits, rngs=None, input_ids=None):
     """get topk."""
     # shape of top_k_weights & top_k_indices:
     # (batch, sequence, num_experts_per_tok).
@@ -643,7 +662,11 @@ class RoutedMoE(nnx.Module):
       top_k_weights, top_k_indices = random_routing(rng, gate_logits, self.num_experts_per_tok)
       return top_k_weights, top_k_indices
 
-    if self.config.model_name.startswith("deepseek3"):
+    if self.is_hash_routing:
+      top_k_indices = self.tid2eid[input_ids]
+      top_k_weights = jnp.take_along_axis(pre_bias_logits, top_k_indices, axis=-1)
+    # NOTE: deepseek2 has a different pattern
+    elif self.config.model_name.startswith(("deepseek3", "deepseek4")):
       top_k_weights, top_k_indices = self.deepseek_routing(gate_logits, pre_bias_logits)
     elif self.config.decoder_block == ctypes.DecoderBlockType.GEMMA4:
       router_probs = jax.nn.softmax(gate_logits.astype(jnp.float32), axis=-1)
@@ -666,8 +689,8 @@ class RoutedMoE(nnx.Module):
   def deepseek_scale_weights(self, weights):
     """Scales weights according to DeepSeek's v3 reference implementation."""
     # https://github.com/deepseek-ai/DeepSeek-V3/blob/2f7b80eecebf3d1c84da5a0d465f6639ea175012/inference/model.py#L592-L594.
-    if self.config.routed_score_func == "sigmoid":
-      weights /= weights.sum(-1, keepdims=True)
+    if self.config.routed_score_func in ("sigmoid", "sqrtsoftplus"):
+      weights /= weights.sum(-1, keepdims=True) + 1e-20
     weights *= self.config.routed_scaling_factor
     return weights
 
@@ -746,20 +769,35 @@ class RoutedMoE(nnx.Module):
         layer_act = self.activation_fn(layer_w0 * 1.702)
         glu = jnp.multiply(layer_w0, layer_act)
         intermediate_layer = jnp.multiply(glu, (layer_w1 + 1))
+      elif self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK and self.config.mlp_activations_limit > 0.0:
+        # DeepSeek V4 uses bounds to clip the SwiGLU activations
+        layer_w0 = jnp.clip(layer_w0, min=None, max=self.config.mlp_activations_limit)
+        layer_w1 = jnp.clip(layer_w1, min=-self.config.mlp_activations_limit, max=self.config.mlp_activations_limit)
+        layer_act = self.activation_fn(layer_w0)
+        intermediate_layer = jnp.multiply(layer_act, layer_w1)
       else:
         layer_act = self.activation_fn(layer_w0)
         intermediate_layer = jnp.multiply(layer_act, layer_w1)
       return intermediate_layer.astype(self.dtype)
 
-  def permute(self, inputs, gate_logits, pre_bias_logits, use_custom_sort_vjp=True, rngs=None, roll_to_expert_id=None):
+  def permute(
+      self,
+      inputs,
+      gate_logits,
+      pre_bias_logits,
+      use_custom_sort_vjp=True,
+      rngs=None,
+      roll_to_expert_id=None,
+      input_ids=None,
+  ):
     """Permute tokens to group by expert to fit gmm call."""
     # reshape inputs (batch, sequence, emb) to (batch * sequence, emb)
     inputs_shape = inputs.shape
     bsz_times_seq_len = inputs_shape[0] * inputs_shape[1]
     inputs_2d = jnp.reshape(inputs, (bsz_times_seq_len, inputs_shape[2]))
-    weights, selected_experts = self.get_topk(gate_logits, pre_bias_logits, rngs)
+    weights, selected_experts = self.get_topk(gate_logits, pre_bias_logits, rngs, input_ids)
     lb_loss = None
-    if self.config.load_balance_loss_weight > 0.0:
+    if self.config.load_balance_loss_weight > 0.0 and not self.is_hash_routing:
       softmax_probs = jax.nn.softmax(gate_logits.astype(jnp.float32), axis=-1).astype(self.dtype)
       lb_loss = self.load_balance_loss(selected_experts, softmax_probs)
 
@@ -1122,6 +1160,7 @@ class RoutedMoE(nnx.Module):
       w0_bias,
       w1_bias,
       wo_bias,
+      input_ids=None,
   ):
     """Perform sparse matrix multiplication of inputs and Experts."""
 
@@ -1288,7 +1327,7 @@ class RoutedMoE(nnx.Module):
         wo_pspec = aqt.partition_spec(wo_pspec, (1,), wo_kernel.dtype, use_bias=False)
       return w0_pspec, w1_pspec, wo_pspec
 
-    def get_routed_moe_shardings(is_batch_sharded_by_expert):
+    def get_routed_moe_shardings(is_batch_sharded_by_expert, has_input_ids):
       if is_batch_sharded_by_expert:
         batch_logical_axis = "activation_batch"
       else:
@@ -1308,11 +1347,17 @@ class RoutedMoE(nnx.Module):
         wo_bias_pspec = self._logical_to_mesh_axes(("exp", "activation_embed"))
 
       gate_logits_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length", None))
-      if self.config.model_name.startswith("deepseek3"):
+      # NOTE: deepseek2 has a different pattern
+      if self.config.model_name.startswith(("deepseek3", "deepseek4")):
         pre_bias_logits_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length", None))
       else:
-        # pre_bias_logits is None for non-DeepSeek v3 models
+        # pre_bias_logits is None for non-deepseek3/4 models, including deepseek2
         pre_bias_logits_pspec = None
+
+      if has_input_ids:
+        decoder_tokens_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length"))
+      else:
+        decoder_tokens_pspec = None
 
       # w0, w1, wo needs to be un sharded on fsdp / fsdp_transpose axis, so use
       # mlp_no_fsdp axis
@@ -1344,6 +1389,7 @@ class RoutedMoE(nnx.Module):
           w0_bias_pspec,
           w1_bias_pspec,
           wo_bias_pspec,
+          decoder_tokens_pspec,
       )
 
     is_batch_sharded_by_expert = is_batch_sharded_by_ep(inputs)
@@ -1359,10 +1405,11 @@ class RoutedMoE(nnx.Module):
         w0_bias_pspec,
         w1_bias_pspec,
         wo_bias_pspec,
-    ) = get_routed_moe_shardings(is_batch_sharded_by_expert)
+        decoder_tokens_pspec,
+    ) = get_routed_moe_shardings(is_batch_sharded_by_expert, input_ids is not None)
     w0_pspec, w1_pspec, wo_pspec = maybe_aqt_partition(w0_kernel, w0_pspec, w1_kernel, w1_pspec, wo_kernel, wo_pspec)
 
-    def route(x, logits, pre_bias_logits, rngs):
+    def route(x, logits, pre_bias_logits, rngs, input_ids=None):
       """Performs both across device and within device token routing/sorting"""
       num_ep = self.get_expert_parallelism_size()
       expert_shard_id = jax.lax.axis_index(self._expert_parallelism_name) if num_ep > 1 else 0
@@ -1390,11 +1437,12 @@ class RoutedMoE(nnx.Module):
             self.config.use_custom_sort_vjp,
             roll_to_expert_id=num_experts_per_shard * expert_shard_id,
             rngs=rngs,
+            input_ids=input_ids,
         )
 
       else:
         x, sorted_selected_experts, weights, group_sizes, selected_experts, lb_loss, bias_updates = self.permute(
-            x, logits, pre_bias_logits, self.config.use_custom_sort_vjp, rngs
+            x, logits, pre_bias_logits, self.config.use_custom_sort_vjp, rngs=rngs, input_ids=input_ids
         )
 
         if num_ep > 1:
@@ -1548,6 +1596,7 @@ class RoutedMoE(nnx.Module):
             w0_bias_pspec,
             w1_bias_pspec,
             wo_bias_pspec,
+            decoder_tokens_pspec,
             P(),  # Replicate the input key
         ),
         out_specs=(
@@ -1557,9 +1606,9 @@ class RoutedMoE(nnx.Module):
         ),
         check_vma=self.config.check_vma,
     )
-    def wrapper(x, logits, pre_bias_logits, w0, w1, wo, w0_bias, w1_bias, wo_bias, rngs):
+    def wrapper(x, logits, pre_bias_logits, w0, w1, wo, w0_bias, w1_bias, wo_bias, sharded_input_ids, rngs):
       batch_size, sequence_length, _ = x.shape
-      x, routing, route_metadata = route(x, logits, pre_bias_logits, rngs)
+      x, routing, route_metadata = route(x, logits, pre_bias_logits, rngs, input_ids=sharded_input_ids)
 
       if self.config.mlp_bias:
         w0_bias, w1_bias, wo_bias = self.transform_bias(routing.selected_experts, w0_bias, w1_bias, wo_bias)
@@ -1712,7 +1761,8 @@ class RoutedMoE(nnx.Module):
       input_axes = (batch_logical_axis, "activation_norm_length", None)
 
     gate_logits_axes = (batch_logical_axis, "activation_norm_length", None)
-    if self.config.model_name.startswith("deepseek3"):
+    # NOTE: deepseek2 has a different pattern
+    if self.config.model_name.startswith(("deepseek3", "deepseek4")):
       pre_bias_logits_axes = (batch_logical_axis, "activation_norm_length", None)
     else:
       pre_bias_logits_axes = None
@@ -1732,7 +1782,17 @@ class RoutedMoE(nnx.Module):
       wo_bias = self._maybe_shard_with_pspec(wo_bias, wo_bias_pspec)
 
     return wrapper(
-        inputs, gate_logits, pre_bias_logits, w0_kernel, w1_kernel, wo_kernel, w0_bias, w1_bias, wo_bias, self.rngs
+        inputs,
+        gate_logits,
+        pre_bias_logits,
+        w0_kernel,
+        w1_kernel,
+        wo_kernel,
+        w0_bias,
+        w1_bias,
+        wo_bias,
+        input_ids,
+        self.rngs,
     )
 
   def reshape_and_update_weights(self, weights, indices):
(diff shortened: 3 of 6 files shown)


src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
     return output, pre_bias_logits
 
 
+def _sqrtsoftplus(x: jax.Array) -> jax.Array:
+  """Computes sqrtsoftplus activation: sqrt(softplus(x))."""
+  # [Any] -> [Any]
+  return jnp.sqrt(jax.nn.softplus(x))
+
+
+class DeepSeekV4TopKRouter(nnx.Module):
```
RissyRan: I see the logic is similar for GateLogic + topK 

Gate logic: https://github.com/AI-Hypercomputer/maxtext/blob/b5e53301e5e24bb4a0aad80e4aab7349d211f1f1/src/maxtext/layers/moe.py#L174

Topk: https://github.com/AI-Hypercomputer/maxtext/blob/b5e53301e5e24bb4a0aad80e4aab7349d211f1f1/src/maxtext/layers/moe.py#L599

Shall we leverage _sqrtsoftplus, and combine them to avoid some duplicate?

src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+    scaled_weights = weights * self.routed_scaling_factor
+
+    return (
+        logits.astype(self.config.dtype),
+        scaled_weights.astype(self.config.dtype),
+        indices,
+    )
+
+
+class DeepSeekV4HashRouter(nnx.Module):
```
RissyRan: Wondering if we should name it HashRouter directly? Do you know if it's specific for DS v4?

src/maxtext/layers/moe.py
```diff
@@ -704,7 +889,13 @@ def deepseek_routing(self, gate_logits: jax.Array, pre_bias_logits: jax.Array) -
   def apply_ffn_activation(self, layer_w0, layer_w1):
     """Applies FFN activation function."""
     with jax.named_scope("ffn_act"):
-      if self.config.decoder_block == ctypes.DecoderBlockType.GPT_OSS:
+      if self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK_V4:
+        limit = getattr(self.config, "swiglu_limit", 1.0)
```
RissyRan: Shall we reuse `self.config.mlp_activations_limit` config as bellow?

src/maxtext/layers/moe.py
```diff
@@ -1255,6 +1461,16 @@ def get_routed_moe_shardings(is_batch_sharded_by_expert):
     ) = get_routed_moe_shardings(is_batch_sharded_by_expert)
     w0_pspec, w1_pspec, wo_pspec = maybe_aqt_partition(w0_kernel, w0_pspec, w1_kernel, w1_pspec, wo_kernel, wo_pspec)
 
+    if gate_weights is not None:
+      gate_weights_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length", None))
```
RissyRan: Nit: could you help add shape info here? same comment for gate_indices

src/maxtext/layers/moe.py
```diff
@@ -2231,13 +2478,31 @@ def retrieve_quantized_weight(
...
+      gate_weights: jax.Array | None = None,
+      gate_indices: jax.Array | None = None,
   ) -> tuple[jax.Array, Optional[jax.Array], Optional[jax.Array]]:
     cfg = self.config
     inputs = inputs.astype(cfg.dtype)
     gate_dtype = jnp.float32 if cfg.float32_gate_logits else cfg.dtype
     routing_inputs = inputs if gate_inputs is None else gate_inputs.astype(gate_dtype)
-    gate_logits, pre_bias_logits = self.gate(routing_inputs)
+
+    if self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK_V4:
```
RissyRan: Could you help make some comments for the conditions? Thanks!

RissyRan (commented):
Thanks for the change! Have a few comments.

src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
     return output, pre_bias_logits
 
 
+def _sqrtsoftplus(x: jax.Array) -> jax.Array:
```
shuningjin: 1. Integrate this as `GateLogit(score_func="sqrtsoftplus")`

src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+    self.routed_scaling_factor = config.routed_scaling_factor
+
+    # Initialize gate weight matrix.
+    # Shape: [hidden_dim, num_experts]
+    kernel_init = nd_dense_init(1.0, "fan_in", "truncated_normal")
+    kernel_shape = (self.hidden_dim, self.num_experts)
+    kernel_in_axis = np.arange(1)
+    kernel_out_axis = np.arange(1, 2)
+
+    self.kernel = nnx.Param(
```
shuningjin: 2.1 Let DeepSeekV4HashRouter reuse with `GateLogit`, as it handles sharding and quantization
```
self.gate_projection = GateLogit(
    in_features_shape=self.hidden_dim,
    out_features_shape=self.num_experts,
    model_name=config.model_name,
    mesh=mesh,
    rngs=rngs,
    dtype=jnp.float32,
    weight_dtype=config.weight_dtype,
    kernel_axes=kernel_axes,
    use_bias=False, # no e_score_correction_bias
    score_func="sqrtsoftplus", # update activation
    quant=config.quant, 
)
```

src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+    flat = inputs.reshape(-1, self.hidden_dim)
+
+    # Compute raw logits in float32.
+    # [tokens, hidden_dim] x [hidden_dim, num_experts] -> [tokens, num_experts]
+    kernel_f32 = jnp.asarray(self.kernel[...], dtype=jnp.float32)
+    logits = jnp.matmul(flat.astype(jnp.float32), kernel_f32)
+
+    # Apply custom scoring function (sqrtsoftplus).
+    # [tokens, num_experts] -> [tokens, num_experts]
+    scores = _sqrtsoftplus(logits)
```
shuningjin: 2.2 Get raw logits and activated scores directly from GateLogit
```
# logits: the raw matmul output, scores: activated with sqrtsoftplus
scores, logits = self.gate_projection(hidden_states)
```

src/maxtext/layers/moe.py
shuningjin: 3. I agree that this could be duplicate logic. We may remove `DeepSeekV4TopKRouter`. Instead use existing `GateLogit` and `RoutedMoE`: see more comments in `RoutedMoE` below.

src/maxtext/layers/moe.py
```diff
@@ -381,25 +558,33 @@ def __init__(
...
-        matmul_precision=self.config.matmul_precision,
-        shard_mode=config.shard_mode,
-        rngs=self.rngs,
+    self.is_hash = self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK_V4 and 0 <= layer_idx < getattr(
+        config, "num_hash_layers", 3
     )
+    if self.is_hash:
+      self.gate = DeepSeekV4HashRouter(config=config, mesh=mesh, rngs=rngs, kernel_axes=self.kernel_axes)
+    elif self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK_V4:
+      self.gate = DeepSeekV4TopKRouter(config=config, mesh=mesh, rngs=rngs, kernel_axes=self.kernel_axes)
```
shuningjin: 3.1 replace DeepSeekV4TopKRouter: reuse GateLogit
```
self.gate = GateLogit(use_bias=True, score_func="sqrtsoftplus")
```

src/maxtext/layers/moe.py
shuningjin: 3.2 replace DeepSeekV4TopKRouter: reuse deepseek_routing
```
if self.config.model_name.startswith("deepseek3") or self.config.model_name.startswith("deepseek4")
```

shuningjin: 3.3 replace DeepSeekV4TopKRouter: reuse deepseek_scale_weights
```
# add v4 scaling
if self.config.routed_score_func == "sqrtsoftplus"
```

shuningjin (commented):
Thanks for the implementation! I would recommend reusing existing MoE logic. In particular, we can leverage `GateLogit` (well-optimized, e.g., sharding, quantization) and `RoutedMoE`. On the high level
1. `GateLogit`: extend with `score_func="sqrtsoftplus"`, for DSV4 activation
2. `DeepSeekV4HashRouter`: replace kernel with `GateLogit(use_bias=False, score_func="sqrtsoftplus")`
3. `DeepSeekV4TopKRouter`: remove entirely, reuse existing logic in `RoutedMoE`
  - `RoutedMoE.__init__`:`GateLogit(use_bias=True, score_func="sqrtsoftplus")`
  - `RoutedMoE.get_topk`: reuse `deepseek_routing`
  - `RoutedMoE.deepseek_scale_weights`: update with DSV4 scaling

See more details in comments.

src/maxtext/layers/moe.py
```diff
@@ -306,6 +308,202 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+            config.weight_dtype,
+            kernel_in_axis,
+            kernel_out_axis,
+        ),
+        out_sharding=kernel_axes,
+    )
+
+    # Load-balancing expert score correction bias.
+    # Shape: [num_experts]
+    self.e_score_correction_bias = nnx.Param(
```
dipakg-lang: According to the DeepSeek-V4 specification for auxiliary-loss-free load balancing, this bias is updated manually based on routing frequency,  please consider changing this to a non-trainable.

dipakg-lang: Also we will have to add respective bias update logic here, please add it in this PR or please feel free to add a TODO, 
DeepSeek-V4 target update speed is 0.001 as per the paper.


Also we will need to integrate that logic in the train step to update biases accordingly.

src/maxtext/layers/moe.py
```diff
@@ -306,6 +308,202 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+  def __call__(self, hidden_states: jax.Array) -> Tuple[jax.Array, jax.Array, jax.Array]:
+    # input hidden_states shape: [batch, seq_len, hidden_dim] or [tokens, hidden_dim]
+    inputs = jnp.asarray(hidden_states, dtype=self.config.dtype)
+    # [batch, seq_len, hidden_dim] -> [tokens, hidden_dim]
+    flat = inputs.reshape(-1, self.hidden_dim)
+
+    # Compute raw logits in float32.
+    # [tokens, hidden_dim] x [hidden_dim, num_experts] -> [tokens, num_experts]
+    kernel_f32 = jnp.asarray(self.kernel[...], dtype=jnp.float32)
+    logits = jnp.matmul(flat.astype(jnp.float32), kernel_f32)
```
dipakg-lang: minor : einsum?

dipakg-lang (commented):
..

src/maxtext/layers/moe.py
dipakg-lang: .

dipakg-lang: .

src/maxtext/layers/moe.py
dipakg-lang: .

src/maxtext/layers/moe.py
```diff
@@ -23,6 +23,8 @@
```
dipakg-lang, line 23: Also we need to add provision for calculation of the sequence wise balance loss, please feel to add a TODO or doing it in the same PR.


Note: When adding seq_balance_loss to the total model loss, note to multiply it by the loss weight of 0.0001 mentioned in the DeepSeek-V4 paper.

entrpn (approved):
LGTM. Please resolve other comments before merging.

dipakg-lang (commented):
Looks like this change wrongly has attention related changes instead of MOE routing related ones?

src/maxtext/layers/moe.py
parambole: Done! We completely removed the standalone DeepSeekV4TopKRouter. We have now integrated sqrtsoftplus directly into linears.py and are fully reusing the existing GateLogit class by passing score_func="sqrtsoftplus".

parambole: Addressed! We removed the DeepSeekV4HashRouter class entirely to avoid duplicating logic. The hash routing logic is now integrated directly into RoutedMoE.get_topk, reusing the existing MoE primitives!

src/maxtext/layers/moe.py
parambole: Addressed! We removed the DeepSeekV4HashRouter class entirely to avoid duplicating logic. The hash routing logic is now integrated directly into RoutedMoE.get_topk, reusing the existing MoE primitives!

src/maxtext/layers/moe.py
parambole: Good point. Instead of hardcoding 1.0 or reusing mlp_activations_limit, we will formally add swiglu_limit to MaxTextConfig (in types.py) to ensure clear discoverability and type safety for DeepSeek V4.

src/maxtext/layers/moe.py
parambole: Done. We've refactored these branch conditions significantly by pulling the logic into the unified get_topk and passing layer_idx. I will ensure clear inline comments are added explaining the hash vs. top-k routing branches.

src/maxtext/layers/moe.py
```diff
@@ -381,25 +558,33 @@ def __init__(
...
-        kernel_axes=self.kernel_axes,
-        use_bias=self.config.routed_bias,
-        # tpu-inference applies the score function in the fused_moe_gmm kernel,
-        # so we don't apply it here to avoid redundant computation.
-        # See https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/fused_moe_gmm.py#L58.
-        score_func="" if self.config.attention == "vllm_rpa" else self.config.routed_score_func,
-        matmul_precision=self.config.matmul_precision,
-        shard_mode=config.shard_mode,
-        rngs=self.rngs,
+    self.is_hash = self.config.decoder_block == ctypes.DecoderBlockType.DEEPSEEK_V4 and 0 <= layer_idx < getattr(
```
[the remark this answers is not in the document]
parambole: Resolved! The plumbing is now fully supported. We added layer_idx as an optional kwarg to RoutedAndSharedMoE and RoutedMoE. The upcoming deepseek_v4.py implementation will explicitly pass layer_idx=self.layer_idx down, while existing models safely fall back to the default of 0.

src/maxtext/layers/moe.py
```diff
@@ -381,25 +558,33 @@ def __init__(
...
+        config, "num_hash_layers", 3
```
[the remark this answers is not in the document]
parambole: Done! num_hash_layer has been formally added to types.py (with a default of 0 to safely disable it for non-DeepSeek models).

src/maxtext/layers/moe.py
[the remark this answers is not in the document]
parambole: Addressed! The plumbing for input_ids has been securely piped all the way through RoutedAndSharedMoE and RoutedMoE. The upcoming deepseek_v4.py model will correctly pass decoder_input_tokens into it.

src/maxtext/layers/moe.py
```diff
@@ -306,6 +306,181 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
...
+    )
+
+
+class DeepSeekV4HashRouter(nnx.Module):
+  """Hash Router for DeepSeek-V4 MoE routing.
+
+  Computes logits, static routing weights based on token IDs, and expert indices.
+  """
+
+  def __init__(
```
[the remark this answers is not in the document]
parambole: Agreed. I will formally add swiglu_limit to types.py to ensure type safety and discoverability.

src/maxtext/layers/moe.py
parambole: Done! We followed this exact architectural feedback. Both DeepSeekV4HashRouter and DeepSeekV4TopKRouter have been completely deleted. We are now fully reusing GateLogit with score_func='sqrtsoftplus', and we integrated the DeepSeek hash and top-k logic directly into RoutedMoE.get_topk.

src/maxtext/layers/moe.py
parambole: Done! We followed this exact architectural feedback. Both the DeepSeekV4HashRouter and DeepSeekV4TopKRouter classes have been completely deleted. We are now fully reusing GateLogit by passing score_func='sqrtsoftplus', and we integrated the DeepSeek hash and top-k logic directly into RoutedMoE.get_topk.

src/maxtext/layers/moe.py
parambole: Done. We removed the custom kernel_f32 matmul logic entirely. We are now directly using the output of self.gate(routing_inputs), which relies on the highly optimized GateLogit module to return both the logits and the activated scores cleanly.

src/maxtext/layers/moe.py
parambole: Done. DeepSeekV4TopKRouter was removed. We are now initializing self.gate = GateLogit(...) and passing score_func="sqrtsoftplus".

src/maxtext/layers/moe.py
parambole: Done. We are now routing DeepSeek V4 directly through the existing deepseek_routing code path in get_topk, completely reusing the shared top-k logic instead of writing a new top-k router.

src/maxtext/layers/moe.py
parambole: Good catch. Splitting the load balancing loss into a Phase 2 PR, I will add a TODO for this here. In the follow-up PR, we will convert e_score_correction_bias to a non-trainable state variable updated via exponential moving average (EMA) according to the V4 paper.

parambole: I've opted to add a TODO here. To keep this PR focused strictly on the MoE routing primitives, the bias update logic (using the 0.001 speed) and the training step integration will be introduced in the dedicated DeepSeek-V4 load balancing PR.

src/maxtext/layers/moe.py
parambole: Obsolete. We actually deleted this raw jnp.matmul block entirely based on Shuning's feedback! The logits are now computed securely inside GateLogit which uses the properly optimized dot_general.

parambole:
> Looks like this change wrongly has attention related changes instead of MOE routing related ones?

Ah! That was just an artifact of the base branch I selected. I recently changed the PR's base branch to deepseek_v4_compressed_attention, which should isolate the diff to show only the MoE routing changes going forward. Let me know if the diff still looks noisy!

src/maxtext/layers/moe.py
parambole: I'm adding a TODO for this. The sequence-wise balance loss (including the 0.0001 loss weight modifier) will be implemented alongside the auxiliary-loss-free bias updates in the upcoming load-balancing PR.

dipakg-lang: I don't see TODO added here, forgot to push the change may be?

src/maxtext/layers/moe.py
```diff
@@ -388,6 +393,16 @@ def __init__(
     self.dtype = dtype
     self.quant = quant
     self.rngs = rngs
+    self.layer_idx = layer_idx
+
+    # DeepSeek V4 Hash Routing
+    if self.layer_idx < self.config.num_hash_layer:
+      self.tid2eid = nnx.Param(
```
dipakg-lang: I know that we are going to add TODO here for correcting this, but declaring this as param might fail the model down the line in case if there any precision casting etc, should we just fix it right now?

src/maxtext/layers/moe.py
```diff
@@ -643,7 +658,10 @@ def get_topk(self, gate_logits, pre_bias_logits, rngs=None):
       top_k_weights, top_k_indices = random_routing(rng, gate_logits, self.num_experts_per_tok)
       return top_k_weights, top_k_indices
 
-    if self.config.model_name.startswith("deepseek3"):
+    if self.layer_idx < self.config.num_hash_layer:
+      top_k_indices = jax.lax.stop_gradient(self.tid2eid.value)[input_ids]
```
dipakg-lang: If we don't declare the tid2eid as variable then  we don't need to call stop_gradient

src/maxtext/configs/types.py
```diff
@@ -227,6 +227,8 @@ class ProfilerType(str, Enum):
     "deepseek3-test",
     "deepseek3-tiny",
     "deepseek3.2-671b",
+    "deepseek4",
+    "deepseekv4",
```
RissyRan: Why two here?

src/maxtext/configs/types.py
```diff
@@ -499,6 +501,10 @@ class ModelArchitecture(BaseModel):
       -1.0,
       description="Upper bound to clip the MLP activation values. -1.0 means no clipping.",
   )
+  swiglu_limit: float = Field(
```
RissyRan: What's the difference between this and mlp_activations_limit?

src/maxtext/layers/moe.py
```diff
@@ -336,8 +336,9 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
 
     if self.score_func:
       output = linears._convert_to_activation_function(self.score_func)(output)
-      if self.model_name.startswith("deepseek3"):
-        pre_bias_logits = output
+
+    if self.model_name.startswith("deepseek"):
```
RissyRan: Please also note, we have deepseek v2, which may not need this. 

Better with `self.model_name in ["deepseek3", "deepseek4"]`?

src/maxtext/layers/moe.py
RissyRan: Also, what's tid2eid? Could you help add a comment?

RissyRan (commented):
Thanks for the change! Please note that we have DS v2, and a few `if` condition may break the functionality.

src/maxtext/configs/base.yml
```diff
@@ -264,6 +265,7 @@ routed_bias: false # a flag if a learnable bias is added for routing
 routed_bias_update_rate: 0.0 # a flag indicate the update rate applied to the router bias term
 mlp_bias: false # a flag if a learnable bias is added for MLP matmul, and originally implemented to support the GPT-OSS model architecture.
 n_routing_groups: -1 # number of groups for routing, disabled by default
+num_hash_layer: 0 # number of hash routing layers, used in DeepSeek V4 (0 means disabled)
```
shuningjin: nit: `first_num_hash_layers`, similar to

https://github.com/AI-Hypercomputer/maxtext/blob/e05491747b1718d1f136468235e09d8d343e5703/src/maxtext/configs/models/deepseek3-671b.yml#L26

src/maxtext/layers/moe.py
shuningjin: +1. There are multiple places with this condition, should exclude `deepseek2`. Please update these instances to
```
self.model_name.startswith(("deepseek3", "deepseek4"))
```

src/maxtext/layers/moe.py
```diff
@@ -376,6 +378,9 @@ def __init__(
       weight_dtype: The data type of the kernel weights.
       dtype: The data type for the computation.
       quant: The quantization configuration. If None, no quantization is applied.
+      layer_idx: The integer index of the current layer within the model architecture.
+        Used to determine whether the layer should use hash routing (if layer_idx < num_hash_layer)
+        or top-K routing. Defaults to 0 for models that do not use layer-specific routing.
```
shuningjin: I prefer we use `is_hashing_routing` in MoE to indicate the hash routing, for clarity and generalization.

The external decoder layer can use condition like `layer_idx < num_hash_layer` to determine this value.

src/maxtext/layers/moe.py
```diff
@@ -388,6 +393,16 @@ def __init__(
     self.dtype = dtype
     self.quant = quant
     self.rngs = rngs
+    self.layer_idx = layer_idx
+
+    # DeepSeek V4 Hash Routing
+    if self.layer_idx < self.config.num_hash_layer:
```
shuningjin: Let's use `self.is_hashing_routing` instead to indicate hash routing. There are multiple occurrences.

```
self.layer_idx < self.config.num_hash_layer -> self.is_hashing_routing
self.layer_idx >= self.config.num_hash_layer -> not self.is_hashing_routing
```

src/maxtext/configs/types.py
shuningjin: +1. `mlp_activations_limit` was introduced for gpt-oss.

DeepSeek4 uses `swiglu_limit` almost same as how gpt-oss uses `mlp_activations_limit`
https://github.com/AI-Hypercomputer/maxtext/blob/0167a4e4de20ae4c8669b34ca500ce1a77622e2a/src/maxtext/layers/moe.py#L761-L772

- HF DSV4: https://github.com/huggingface/transformers/blob/e31443976ba8aca39f28c36c97f1805ba8326f43/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1019-L1026

- HF GPT-OSS: https://github.com/huggingface/transformers/blob/e31443976ba8aca39f28c36c97f1805ba8326f43/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L87-L93

src/maxtext/layers/moe.py
shuningjin: +1 add more explanation for hash_routing mechanism. I believe this is token-id to expert-id look-up table:

https://github.com/huggingface/transformers/blob/e31443976ba8aca39f28c36c97f1805ba8326f43/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1050-L1067

Torch uses `register_buffer("tid2eid")` rather than `nn.Parameter`. This means the variable should not receive gradient update.

I see you are using `nnx.Param` + `stop_gradient`, which might do the trick. Alternatively, would `nnx.Variable` be a better choice?

src/maxtext/layers/moe.py
```diff
@@ -2010,7 +2072,8 @@ def dense_matmul(
       lb_loss = None
 
     # Calculate routed bias updates (loss-free)
-    if self.should_update_load_balance():
+    # The bias update logic is only applicable to Top-K routed layers (layer_idx >= num_hash_layer)
+    if self.should_update_load_balance() and self.layer_idx >= self.config.num_hash_layer:
```
shuningjin: I see there are two such conditions (also below). Can you directly update `self.should_update_load_balance()`?

https://github.com/AI-Hypercomputer/maxtext/blob/0167a4e4de20ae4c8669b34ca500ce1a77622e2a/src/maxtext/layers/moe.py#L645-L647

to
```
  def should_update_load_balance(self):
    """Determines if loss-free load balancing updates should be applied."""
    return self.config.routed_bias and self.config.routed_bias_update_rate > 0.0 and not self.is_hash_routing
```

src/maxtext/layers/moe.py
shuningjin: for hashing_routing, we do not need "e_correction_bias"
https://github.com/huggingface/transformers/blob/e31443976ba8aca39f28c36c97f1805ba8326f43/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1038

Update to
```
use_bias=self.config.routed_bias and not is_hashing_routing 
```

shuningjin, line 287: While not introduced by this PR, it seems similar to tid2eid (i.e., register buffer instead nn.parameter in torch). Just a question for discussion.

I wonder if this bias (e_correction_bias) could be updated from `nnx.Param` to `nnx.Variable`
- https://github.com/huggingface/transformers/blob/e31443976ba8aca39f28c36c97f1805ba8326f43/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1038
- I see if it actually receives zero gradient, which is correct
- but maybe `nnx.Variable` can be further ignored optimizer.state and save a bit memory?

@RissyRan

tests/unit/deepseek_v4_vs_reference_test.py
```diff
@@ -724,5 +729,221 @@ def test_document_packing_masking(self):
...
+    self.pt_config = DeepseekV4Config(
+        hidden_size=self.hidden_dim,
+        num_local_experts=self.num_experts,
+        num_experts_per_tok=self.num_experts_per_tok,
+        routed_scaling_factor=2.0,
+        scoring_func="sqrtsoftplus",
+        vocab_size=self.vocab_size,
+    )
+
+    from maxtext.configs import pyconfig
```
shuningjin: nit: move imports to top

shuningjin (commented):
Thanks for refactoring and reusing MoE logic! Further comments:
- Perhaps we can use `is_hash_routing` instead `layer_idx` to indicate hash router for clarify. Decoder can use idx to determine.
- Perhaps can use `nnx.Variable` instead of `nnx.Param` for register_buffer counterpart

src/maxtext/layers/moe.py
parambole: >I see you are using nnx.Param + stop_gradient, which might do the trick. Alternatively, would nnx.Variable be a better choice?

In the initial implementation I had used `nnx.Variable` but there were some nit issues with respect to the int dtype and hence changes it to the nnx.Param + stop_gradient approach. 

But yes I agree on the memory point. I have swithced to nnx.Variable for now and will re-verify it during the `decoder` plumbing phase. 

Additionally, I have added comments around this variable.

src/maxtext/configs/types.py
parambole: Oops. I have corrected it.

src/maxtext/configs/types.py
parambole: That is a great point which I + agent + agent reviewer overlooked. I have corrected it now.

src/maxtext/layers/moe.py
parambole: Thanks for calling this out I wasn't aware of it. I have corrected it.

src/maxtext/configs/base.yml
parambole: Agreed I like `first_num_hash_layers` it better. Changed it everywhere.

tests/unit/deepseek_v4_vs_reference_test.py
parambole: Corrected it.

src/maxtext/layers/moe.py
parambole: Agreed. I have changed it back to `nnx.Variable` I had faced issues in the past with this around the dtype. I will verify it in the plumbing PR.

parambole: Done great suggestion

src/maxtext/layers/moe.py
parambole: Done.

src/maxtext/layers/moe.py
parambole: Done. This is a great suggestion.

src/maxtext/layers/moe.py
parambole: Agreed. This should be encapsulated in a layer above as this logic is specific to the model and not MoE. 

Thank you for the great suggestion @shuningjin

src/maxtext/layers/moe.py
parambole: Done I have made the changes.

src/maxtext/layers/moe.py
parambole: Added it [here](https://github.com/AI-Hypercomputer/maxtext/pull/3871/changes#diff-e080eae53990c3bcfd6cd1f60db63f57d80988dfe8715b4ffbc09fbcabb5d8f3R2070)

src/maxtext/configs/base.yml
```diff
@@ -179,6 +179,7 @@ global_head_dim: 0
 global_num_kv_heads: 0
 mlp_activations: ["silu", "linear"]
 mlp_activations_limit: -1.0
+
```
RissyRan: nit: remove extra line

src/maxtext/layers/moe.py
```diff
@@ -1374,6 +1414,9 @@ def route(x, logits, pre_bias_logits, rngs):
             jax.lax.all_gather(z, axis_name=self._expert_parallelism_name, tiled=True)
             for z in (x, logits, pre_bias_logits)
         )
+        if input_ids is not None:
+          # TODO(b/509929019): Gather input_ids for DeepSeek V4 Hash Routing when Ring-of-Experts.
```
RissyRan: Let's add an assertion in types.py to avoid the issue (v4 + ring-of-experts). We could remove the `if` here.

src/maxtext/layers/moe.py
```diff
@@ -2005,15 +2061,19 @@ def dense_matmul(
...
     # Calculate load balance loss
-    if self.config.model_call_mode != "inference":
+    # DeepSeek V4 uses Hash Routing for the first `first_num_hash_layers` layers.
+    # These layers route deterministically based on token IDs and do not generate auxiliary loss.
+    if self.config.model_call_mode != "inference" and not self.is_hash_routing:
       softmax_probs = jax.nn.softmax(gate_logits.astype(jnp.float32), axis=-1).astype(self.dtype)
       lb_loss = (
           self.load_balance_loss(top_k_indices, softmax_probs) if self.config.load_balance_loss_weight > 0.0 else None
       )
+      # TODO(dipakg-lang): Add sequence-wise balance loss * 0.0001
```
RissyRan: Can we attach a bug?

RissyRan (approved):
Thanks for the change! Please add an assertion in types.py.

src/maxtext/configs/base.yml
parambole: Removed.

src/maxtext/layers/moe.py
parambole: Done added a check

src/maxtext/layers/moe.py
parambole: Done added a bug number

src/maxtext/layers/moe.py
```diff
@@ -336,8 +336,9 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
 
     if self.score_func:
       output = linears._convert_to_activation_function(self.score_func)(output)
-      if self.model_name.startswith("deepseek3"):
-        pre_bias_logits = output
+
+    if self.model_name.startswith(("deepseek3", "deepseek4")):
```
shuningjin, line 341: For future reference, might be good to highlight
```
NOTE: deepseek2 has a different pattern
```

src/maxtext/layers/moe.py
```diff
@@ -1302,12 +1335,17 @@ def get_routed_moe_shardings(is_batch_sharded_by_expert):
         wo_bias_pspec = self._logical_to_mesh_axes(("exp", "activation_embed"))
 
       gate_logits_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length", None))
-      if self.config.model_name.startswith("deepseek3"):
+      if self.config.model_name.startswith(("deepseek3", "deepseek4")):
         pre_bias_logits_pspec = self._logical_to_mesh_axes((batch_logical_axis, "activation_norm_length", None))
       else:
-        # pre_bias_logits is None for non-DeepSeek v3 models
+        # pre_bias_logits is None for non-DeepSeek models
```
shuningjin: nit: update this comment to be accurate (same for the other occurence)
```
# pre_bias_logits is None for non-deepseek3/4 models, including deepseek2
```

src/maxtext/layers/moe.py
```diff
@@ -629,9 +643,9 @@ def get_context_autoregressive_parallelism_size(self):
 
   def should_update_load_balance(self):
     """Determines if loss-free load balancing updates should be applied."""
-    return self.config.routed_bias and self.config.routed_bias_update_rate > 0.0
+    return self.config.routed_bias and self.config.routed_bias_update_rate > 0.0 and not self.is_hash_routing
```
shuningjin, line 651: nit: can you incorporate this into comment
```
The bias update logic is only applicable to Top-K router.
Hash router does not use routed bias.
```

shuningjin (approved):
Great work!

shuningjin:
Note sure why `tests/integration/hlo_diff_test.py` [fails](https://github.com/AI-Hypercomputer/maxtext/actions/runs/27243485688/job/80452404494?pr=3871). Please double check.

src/maxtext/layers/moe.py
parambole: SG. Done

src/maxtext/layers/moe.py
parambole: SG. Done

src/maxtext/layers/moe.py
parambole: SG. Done

dipakg-lang (approved):

PR #3875, opened 2026-05-11
Improve performance of gemma4 MoE inference.

NicoGrande:
# Description

This PR introduces a small, inference specific optimization to reduce overall step time. Specifically, this PR moves the per-expert scale application to the model initialization instead of during the call method, given that these parameters are static during inference.

# Tests

`vllm_decode.py` before and after. 

Step time without this PR: 13.59 ms
Step time with this PR: 7.975 ms

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -539,6 +539,14 @@ class RoutedMoE(nnx.Module):
     else:
       self.per_expert_scale = None
 
+    # Scale the output projection ahead of time during inference for higher generation throughput.
+    if (
+        self.per_expert_scale is not None
+        and self.config.model_call_mode == "inference"
+        and self.config.fuse_expert_scales
+    ):
+      self.wo.value = self.wo.value * self.per_expert_scale.value[:, None, None]
+
   def _maybe_shard_with_logical(self, inputs, logical_name):
     return maybe_shard_with_logical(
         inputs,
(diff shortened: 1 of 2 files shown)


src/maxtext/layers/moe.py
```diff
@@ -538,6 +538,10 @@ def __init__(
       )
     else:
       self.per_expert_scale = None
+    
+    # Scale the output projection ahead of time during inference for higher generation throuhput.
+    if self.per_expert_scale is not None and self.config.model_call_mode == "inference":
```
gobbleturk: I think this assumes all inference callers have this pre-fused. Should we instead create a new config option "fused_expert_scale?" to support both cases? I guess the unfused case is always unperformant for inference but some folks might want it for quick correctness tests?

gobbleturk (changes requested):

src/maxtext/layers/moe.py
NicoGrande: This scale parameter is applied to the out-projection weights IIUC (`wo`). Since the out-projections are not part of the fused codepath, are we okay to leave it as is?

gobbleturk: fused is overloaded lol, this is an orthogonal kind of fusion

gobbleturk: you are fusing the gemma4 scales with wo IIUC

NicoGrande: Ahhhh lol I see what you are saying

NicoGrande: Done! Added one more flag to trigger this codepath

gobbleturk (approved):

igorts-git (approved):

PR #3890, opened 2026-05-12
mHC optimization for expansion rate 4 (flag controlled)

dandragona:
# Description

When expansion rate is equal to four we don't need to do sinkhorn as we can generate the doubly-stochastic matrix by taking a random convex combination of the permutation matrices (there are 4! = 24 of them when using expansion rate 4).

Based on https://arxiv.org/pdf/2601.05732.

* Flag-controlled for now, but once we establish a DeepSeek v4 baseline we can use it by default if it shows a performance improvement.
* Can potentially support this for expansion rates higher than 4 too before N! becomes too large (maybe it will increase training stability?).
* Did some refactoring on the mhc_test.py code.
* pylint refactors on some other files.

# Tests

Just tested with unit tests, adding the gemini-review now.

Waiting to performance test it once we get baseline DeepSeek v4 numbers.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -1224,6 +1224,15 @@ force_q_layout: false
 mhc_expansion_rate: 1
 # The number of iterations for the Sinkhorn-Knopp algorithm.
 sinkhorn_iterations: 20
+# mHC-lite: https://openreview.net/pdf?id=5IJX6kvOif
+# - Controls whether to generate the MHC doubly stochastic matrix via
+#   permutation-based convex combination rather than Sinkhorn-Knopp.
+# - Practical only for a small mhc_expansion_rate (e.g., k=4), as the number
+#   of permutation matrices scales factorially (k!).
+# - Ideally, this should be used whenever expansion rate is small as it removes
+#   the expensive sinkhorn iterations, the downside to this approach is that
+#   it is factorial in k.
+enable_mhc_lite: False
 
 ################################## DeepSeek Engram ##################################
 # Indices of transformer layers where Engram are integrated; leave empty [] to disable.
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -1407,6 +1407,14 @@ class ManifoldConstrainedHyperConnections(BaseModel):
 
   mhc_expansion_rate: PositiveInt = Field(1, description="The number of parallel streams in Hyper Connection.")
   sinkhorn_iterations: PositiveInt = Field(20, description="The number of iterations for the Sinkhorn-Knopp algorithm.")
+  enable_mhc_lite: bool = Field(
+      False,
+      description=(
+          "Whether to generate the MHC doubly stochastic matrix via "
+          "permutation-based convex combination rather than Sinkhorn-Knopp. "
+          "Practical only for a small mhc_expansion_rate (e.g., k=4)."
+      ),
+  )
 
 
 class DilocoParams(BaseModel):
--- a/src/maxtext/layers/mhc.py
+++ b/src/maxtext/layers/mhc.py
@@ -14,7 +14,10 @@
 
 """DeepSeek Manifold-Constrained Hyper Connections (mHC) Layer."""
 
+import itertools
+import math
 from typing import Callable
+
 from flax import nnx
 import jax
 import jax.numpy as jnp
@@ -23,6 +26,17 @@ from maxtext.common.common_types import Array, Config
 from maxtext.common.common_types import HyperConnectionType
 from maxtext.layers.initializers import default_bias_init, default_scalar_init, nd_dense_init
 from maxtext.layers.normalizations import RMSNorm
+
+
+def get_permutation_matrices(k: int) -> Array:
+  """Generates all permutation matrices of size k.
+
+  Reference: mHC-lite: https://openreview.net/pdf?id=5IJX6kvOif
+  Shape: (k!, k, k)
+  """
+  perms = list(itertools.permutations(range(k)))
+  perms_array = jnp.array(perms)
+  return jnp.eye(k)[perms_array]
 
 
 def get_functions(expansion_rate: int):
@@ -118,6 +132,17 @@ class ManifoldConstrainedHyperConnections(nnx.Module):
         out_sharding=(None,),
     )
 
+    if self.config.enable_mhc_lite:
+      num_perms = math.factorial(self.k)
+      res_out_dim = num_perms
+      res_beta_shape = (num_perms,)
+      res_beta_sharding = (None,)
+      self.permutation_matrices = get_permutation_matrices(self.k)
+    else:
+      res_out_dim = self.k * self.k
+      res_beta_shape = (self.k, self.k)
+      res_beta_sharding = (None, None)
+
     # Weight matrices
     scale_init = nd_dense_init(1.0, "fan_in", "normal")
     in_axis = 0
@@ -126,7 +151,7 @@ class ManifoldConstrainedHyperConnections(nnx.Module):
     self.res_alpha = nnx.Param(
         scale_init(
             self.rngs.params(),
-            (self.k * self.dim, self.k * self.k),
+            (self.k * self.dim, res_out_dim),
             self.weight_dtype,
             in_axis=in_axis,
             out_axis=out_axis,
@@ -156,8 +181,8 @@ class ManifoldConstrainedHyperConnections(nnx.Module):
 
     # Biases
     self.res_beta = nnx.Param(
-        default_bias_init(self.rngs.params(), (self.k, self.k), self.weight_dtype),
-        out_sharding=(None, None),
+        default_bias_init(self.rngs.params(), res_beta_shape, self.weight_dtype),
+        out_sharding=res_beta_sharding,
     )
     self.pre_beta = nnx.Param(
         default_bias_init(self.rngs.params(), (self.k,), self.weight_dtype),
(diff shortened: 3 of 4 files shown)


src/maxtext/configs/types.py
```diff
@@ -506,7 +506,10 @@ class ModelArchitecture(BaseModel):
       True,
       description="Whether to apply scale on query and key normalizations (default True).",
   )
-  v_norm_with_scale: bool = Field(True, description="Whether to apply scale on value normalization (default True).")
+  v_norm_with_scale: bool = Field(
+      True,
+      description="Whether to apply scale on value normalization (default True).",
+  )
```
shuningjin: Could you remove these unrelated formatting changes, same for other places?
- Seems this formatting is not necessary as it already complies with [maxtext code style](https://github.com/AI-Hypercomputer/maxtext/blob/fb79a9e8b2c465248874212202dc27a9e38363e4/.pre-commit-config.yaml#L47-L54). 
- If using vscode, could add [these](https://paste.googleplex.com/6254841977503744) to settings.json.

src/maxtext/configs/base.yml
```diff
@@ -1217,6 +1217,8 @@ force_q_layout: false
 mhc_expansion_rate: 1
 # The number of iterations for the Sinkhorn-Knopp algorithm.
 sinkhorn_iterations: 20
+# Whether to enable the permutation-based convex combination shortcut when mhc_expansion_rate is 4.
+enable_mhc_k4_shortcut: True
```
shuningjin: In theory, I think this is applicable for any k?
- In practice, when k is large, it becomes inefficient
- In the mHC-lite paper, they use k=4 as an example to be consistent as mHC paper
- Let's make k flexible

Update comment / name / default.
```
# mHC-lite: https://openreview.net/pdf?id=5IJX6kvOif
# - Controls whether to generate the MHC doubly stochastic matrix via
#   permutation-based convex combination rather than Sinkhorn-Knopp.
# - Practical only for a small mhc_expansion_rate (e.g., k=4), as the number 
#   of permutation matrices scales factorially (k!).
enable_mhc_lite: False
```

tests/unit/grain_data_processing_test.py
shuningjin: is this change intentional?

tests/unit/mhc_test.py
```diff
@@ -137,7 +142,14 @@ def setUp(self):
 
   # Skip GPU due to NotImplementedError: dynamic grid bounds not supported in the Triton backend
   @pytest.mark.tpu_only
-  def test_moe_layer_output_shape(self):
+  @parameterized.named_parameters(("Rate3", 3), ("Rate4", 4))
+  def test_moe_layer_output_shape(self, rate):
+    # Skip test if TPU hardware is not available
+    has_tpu = any(d.platform == "tpu" for d in jax.devices())
+    if not has_tpu:
+      self.skipTest("test_moe_layer_output_shape requires TPU hardware.")
```
shuningjin: I thought this is already handled by `@pytest.mark.tpu_only`?

tests/unit/mhc_test.py
```diff
@@ -86,19 +89,21 @@ def test_doubly_stochastic_property(self):
...
-  def setUp(self):
+  def _setup_mhc(self, rate):
+    """Sets up the common configurations and modules for MHC testing."""
     self.dim = 16
     extra_args = get_decoupled_parallelism_overrides()
     self.config = pyconfig.initialize(
         [None, get_test_config_path()],
         **extra_args,
-        run_name="test_mhc",
+        skip_jax_distributed_system=True,
```
shuningjin: do not skip, instead mark `@pytest.mark.tpu_only` for relavent tests

src/maxtext/layers/mhc.py
```diff
@@ -14,17 +14,29 @@
...
-from maxtext.layers.initializers import default_bias_init, default_scalar_init, nd_dense_init
+from maxtext.layers.initializers import (
+    default_bias_init,
+    default_scalar_init,
+    nd_dense_init,
+)
 from maxtext.layers.normalizations import RMSNorm
 
 
+def get_4x4_permutation_matrices():
```
shuningjin: update this
- add shape(k!, k, k) in comment
- add reference mHC-lite: https://openreview.net/pdf?id=5IJX6kvOif
- when using in mhc in the below, generate this once in init, avoid repeat generate forward

src/maxtext/layers/mhc.py
```diff
@@ -118,6 +130,15 @@ def __init__(
         out_sharding=(None,),
     )
 
+    if self.k == 4 and self.config.enable_mhc_k4_shortcut:
+      res_out_dim = 24
+      res_beta_shape = (24,)
+      res_beta_sharding = (None,)
```
shuningjin: remove `k==4` constraint, use `k!` for shape instead of hardcoding?

tests/unit/mhc_test.py
```diff
@@ -221,6 +241,72 @@ def test_attention_layer_output_shape(self):
...
+      # To compare with Sinkhorn, we pass log(res_mapping_out). Since Sinkhorn
+      # applies softmax, log will undo that. Thus, we should immediately exit
+      # Sinkhorn (since we started with a doubly-stochastic matrix).
+      epsilon = 1e-12
+      log_input = jnp.log(res_mapping_out + epsilon)
+      # Use 0 iterations to prove we are already converged.
+      sinkhorn_out = mhc.sinkhorn(log_input, 0)
+
+      # They should be close
+      np.testing.assert_allclose(res_mapping_out, sinkhorn_out, atol=1e-2, rtol=1e-2)
```
shuningjin: These steps seems to be checking

y := softmax(log(x)) == x, 
- where x =  res_mapping_out form mhc-lite
- and y = mhc.sinkhorn(log_input, 0)
- essentially verifying x is doubly stochastic 

Instead of sinkhorn(0) trick, maybe we can directly test the property for mhc-lite output, similar to `test_doubly_stochastic_property`

shuningjin (commented):
Thanks for implementing [mHC-lite](https://openreview.net/pdf?id=5IJX6kvOif)! The core logic looks solid.
- Your implementation aligns with Eqs (2) and (3), and the parameter shapes are correct.
- Generalize k: The shortcut is currently hardcoded for k=4. While the $k!$ permutation scaling limits large $k$ in practice, the underlying math applies to any $k$. Let's parameterize it for flexibility.
- Cleanup: I left a few minor inline comments. Please also revert unrelated formatting changes.

Also thanks for adding the unit tests! As a follow-up, you can run an end-to-end test to verify training stability and performance.

src/maxtext/configs/types.py
dandragona: done.

src/maxtext/configs/base.yml
dandragona: Good idea, implemented.

tests/unit/grain_data_processing_test.py
dandragona: Nope it was from some checks failing due to preexisting failures :( but I think they are fixed now so I removed this.

tests/unit/mhc_test.py
dandragona: Done.

tests/unit/mhc_test.py
dandragona: Done.

src/maxtext/layers/mhc.py
dandragona: Sounds good! Added this.

src/maxtext/layers/mhc.py
dandragona: Done.

tests/unit/mhc_test.py
dandragona: Works for me. Added this.

src/maxtext/layers/mhc.py
```diff
@@ -14,17 +14,35 @@
...
 import jax.numpy as jnp
 from jax.sharding import Mesh
 from maxtext.common.common_types import Array, Config
 from maxtext.common.common_types import HyperConnectionType
-from maxtext.layers.initializers import default_bias_init, default_scalar_init, nd_dense_init
+from maxtext.layers.initializers import (
+    default_bias_init,
+    default_scalar_init,
+    nd_dense_init,
+)
```
shuningjin: nit: unrelated change?

.gitignore
```diff
@@ -155,3 +155,4 @@ gha-creds-*.json
 
 # vscode workspace
 maxtext.code-workspace
+.jetskicli/
```
shuningjin: remove?

tests/unit/mhc_test.py
```diff
@@ -33,7 +33,7 @@
 from tests.utils.test_helpers import get_test_config_path
 
 
-class TestExpandReduce(unittest.TestCase):
+class TestExpandReduce(absltest.TestCase):
```
shuningjin: wonder why replace `unittest.TestCase` with `absltest.TestCase`

tests/unit/mhc_test.py
shuningjin: remove `skip_jax_distributed_system=True`?

tests/unit/mhc_test.py
```diff
@@ -86,17 +86,19 @@ def test_doubly_stochastic_property(self):
...
     self.dim = 16
     self.config = pyconfig.initialize(
         [None, get_test_config_path()],
-        run_name="test_mhc",
+        skip_jax_distributed_system=True,
+        run_name=f"test_mhc_k{rate}",
         enable_checkpointing=False,
         model_name="deepseek-custom",
-        per_device_batch_size=jax.device_count(),
+        per_device_batch_size=max(4, jax.device_count()),
```
shuningjin: is this change necessary?

shuningjin (approved):
LGTM! Minor comments.

.gitignore
dandragona: Done.

src/maxtext/layers/mhc.py
dandragona: Removed this.

tests/unit/mhc_test.py
dandragona: Done.

tests/unit/mhc_test.py
dandragona: Was trying to take advantage of the parametrized test thing.

src/maxtext/layers/mhc.py
```diff
@@ -174,13 +199,30 @@ def res_mapping(self, x: Array):
...
     res_alpha_scale = jnp.asarray(self.res_alpha_scale[...], self.dtype)
-    # Apply projection: (b, s, k*d) @ (k*d, k*k) -> (b, s, k*k)
-    h_res = jnp.einsum("bsm,mn -> bsn", x, res_alpha, precision=self.matmul_precision)
-    b, s, _ = h_res.shape
-    h_res = jnp.reshape(h_res, (b, s, self.k, self.k))
-    intermediate = res_alpha_scale * h_res + res_beta[None, None, :, :]
-    output = sinkhorn(intermediate, self.sinkhorn_iterations)
-    return output
+
+    if self.config.enable_mhc_lite:
```
parambole, line 203: Hey @dandragona, thanks for the change. Can you add some details about when lite should/can be used and what the limitations are?

Also, can we follow up in the PR description (now or later) on the perf difference between mhc and mhc-lite?

parambole (approved):
LGTM 🚀

src/maxtext/layers/mhc.py
dandragona: Added some more details on this to base.yml. I'll try to attain the perf difference in a follow up. If it proves useful we can then try to remove the need for the flag and just turn it on by default.

PR #3895, opened 2026-05-13
Add DPO and ORPO preference data preprocessing pipeline utils

igorts-git:
# Description

To simplify code review  I am splitting the Tunix-based DPO implementation into smaller PRs.
This one adds the data reading processing required by DPO.

The classic DPO inputs consist of three data columns: ["prompt", "chosen_response", "rejected_response"]. 
However, some DPO datasets use a two-column format where the prompt is the prefix to the choosen and rejected strings.
When a 2-column dataset is used our implementation extracts the common prefix into the "prompt" field that is then fed into the model separately.
The column names in the dataset can wary, for example ["input", chosen", "rejected"]. Our implementation allows the user to supply the dataset column names via the `train_data_columns` and `eval_data_columns` parameters.

Tunix requires left-padded prompt and right-padded responses. Our code implements this padding (and truncation if needed) it also provides Tunix with the corresponding masks.

NOTE: once this PR is merged the legacy DPO will stop working correctly. The follow up PRs will enable Tunix-based DPO.

Caveat: This PR only adds support for HuggingFace datasets, while the legacy DPO implementation supported HuggingFace, TFDS and Grain. This is on-par with our SFT implementation. We need to discuss the priority of supporting TFDS and Grain in post-training.

# Tests

Added unit tests. Ran DPO/ORPO and performed logits comparison against the legacy implementation.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -86,7 +86,7 @@ checkpoint_conversion_fn: none
 # optional checkpoint context to use for loading. options: "orbax", "safetensors"
 source_checkpoint_layout: "orbax"
 
-# Only applicable to Single Controller/Pathways on Cloud. Experimental feature, under testing 
+# Only applicable to Single Controller/Pathways on Cloud. Experimental feature, under testing
 colocated_python_checkpointing: False
 
 # enables autocheckpoint, which saves a checkpoint at the preemption step.
@@ -451,7 +451,7 @@ hardware: 'tpu' # Supported hardware types are 'tpu', 'gpu', 'gpu_multiprocess' 
 # internal_compile allows bypassing open-source topology name mappings when using internal topologies directly via get_topology_desc.
 internal_compile: False
 internal_compile_num_devices: -1 # You must specify the number of devices when using internal_compile.
-compile_xla_flags: "" # Compiler options e.g. compile_xla_flags="--xla_tpu_num_sparse_cores_for_gather_offloading=1 --xla_tpu_scoped_vmem_limit_kib=65536" 
+compile_xla_flags: "" # Compiler options e.g. compile_xla_flags="--xla_tpu_num_sparse_cores_for_gather_offloading=1 --xla_tpu_scoped_vmem_limit_kib=65536"
 
 # Parallelism
 shard_mode: "auto" # can be either auto or explicit
@@ -564,8 +564,8 @@ logical_axis_rules: [
                       # ==========================================
                       # Deprecated / Scheduled for Removal
                       # ==========================================
-                      ['mlp_no_fsdp', ['tensor', 'tensor_sequence', 'autoregressive']], 
-                      ['embed_tensor_transpose', ['tensor_transpose']],                 
+                      ['mlp_no_fsdp', ['tensor', 'tensor_sequence', 'autoregressive']],
+                      ['embed_tensor_transpose', ['tensor_transpose']],
                       ['exp_with_fsdp', 'fsdp'],
                   ]
 # Axes used for DCN must be earlier in this list than ICI, see (b/339009148) for details
@@ -680,8 +680,6 @@ global_rampup_samples: 500
 
 # direct preference optimization (DPO)
 use_dpo: False
-dpo_label_smoothing: 0.0
-dpo_beta: 0.1
 
 # Supervised Fine-Tuning (SFT)
 use_sft: False
@@ -1206,7 +1204,7 @@ use_jax_splash: false
 # Path to the HuggingFace-style config directory for the adapter (e.g. src/maxtext/integration/vllm/maxtext_vllm_adapter)
 vllm_hf_config_path: ""
 # A JSON string of overrides to apply to the HuggingFace-style config for the vLLM adapter.
-# This can be used to override specific settings without modifying the original config file. 
+# This can be used to override specific settings without modifying the original config file.
 vllm_hf_overrides: {}
 # JSON string containing additional configuration for the vLLM model (e.g. '{"maxtext_config": {...}}')
 vllm_additional_config: {}
@@ -1221,7 +1219,7 @@ sinkhorn_iterations: 20
 
 ################################## DeepSeek Engram ##################################
 # Indices of transformer layers where Engram are integrated; leave empty [] to disable.
-# Example: [1, 4] attaches to the 2nd and 5th layer. 
+# Example: [1, 4] attaches to the 2nd and 5th layer.
 engram_layers: []
 # The max 'n' in N-gram. Example: n=3 means it covers both 2-grams and 3-grams.
 engram_max_ngram_size: 3
--- a/src/maxtext/configs/post_train/dpo.yml
+++ b/src/maxtext/configs/post_train/dpo.yml
@@ -1,6 +1,12 @@
 base_config: "base.yml"
 
 use_dpo: true
+dpo:
+  algo: 'dpo'
+  orpo_lambda: 0.1
+  dpo_label_smoothing: 0.0
+  dpo_beta: 0.1
+  max_prompt_length: null
 packing: false
 train_data_columns: ['chosen', 'rejected']
 eval_data_columns: ['chosen', 'rejected']
@@ -24,8 +30,6 @@ hf_eval_split: 'test'
 
 gradient_clipping_threshold: 10.0
 learning_rate: 5.0e-7
-dpo_label_smoothing: 0.0
-dpo_beta: 0.1
 
 enable_goodput_recording: false
 monitor_goodput: false
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -1206,12 +1206,24 @@ class OlmoGrainDataset(BaseModel):
   olmo_apply_ngram_filter: bool = Field(True, description="Mask repetitive instances per OLMo-core's repetition filter.")
 
 
+class DPO(BaseModel):
+  """Configuration for DPO and ORPO preference optimization algorithms."""
+
+  algo: Literal["dpo", "orpo"] = Field("dpo", description="Alignment algorithm to use.")
+  dpo_beta: float = Field(0.1, description="Beta parameter for DPO.")
+  orpo_lambda: float = Field(0.1, description="Weight for preference loss in ORPO.")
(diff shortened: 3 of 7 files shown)


src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,96 @@
...
+      raise KeyError(
+          f"Column '{e.args[0]}' not found in the dataset. "
+          f"Expected columns: {self.data_column_names}. "
+          f"Available columns: {list(element.keys())}. "
+          "Please verify that 'train_data_columns' and 'eval_data_columns' match your dataset."
+      ) from e
+
+    # 2. Padding and Masking
+    max_prompt_length = self.max_target_length // 2
+    max_response_length = self.max_target_length - max_prompt_length
```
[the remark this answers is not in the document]
igorts-git: added "max_prompt_length" parameter

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Padding can be done on the left or right. Trimming is only on the right.
+    """
+    x = np.asarray(x)
+    pad_amount = max(length - x.shape[0], 0)
+    if left:
+      pad_width = ((pad_amount, 0),)
+    else:
+      pad_width = ((0, pad_amount),)
+    return np.pad(x[:length], pad_width, constant_values=self.pad_id).astype(np.int32)
```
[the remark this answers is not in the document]
igorts-git: applied the recommended patch (with a small modification)

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,96 @@
...
+        full_chosen = element[self.data_column_names[0]]
+        full_rejected = element[self.data_column_names[1]]
+
+        # Find common prefix length
+        i = 0
+        for i, (c, r) in enumerate(zip(full_chosen, full_rejected)):
+          if c != r:
+            break
+        input_ids = full_chosen[:i]
+        chosen_ids = full_chosen[i:]
```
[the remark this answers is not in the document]
igorts-git: applied the recommended patch

tests/post_training/unit/dpo_data_processing_test.py
```diff
@@ -0,0 +1,278 @@
...
+
+  def test_two_column_prefix_extraction(self):
+    """Verify common prefix extraction for 2-column datasets."""
+    # The column names will be remappend into "chosen" and "rejected"
+    prep = dpo_utils.DPOTunixPrep(pad_id=self.pad_id, max_target_length=21, data_column_names=("liked", "disliked"))
+    sample = {
+        "liked": np.array([1, 2, 3, 10, 11]),
+        "disliked": np.array([1, 2, 3, 20, 21, 22]),
+    }
+    output = prep.map(sample)
```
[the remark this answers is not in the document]
igorts-git: added a test case

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+import dataclasses
+import grain.python as grain
+import numpy as np
+
+
+@dataclasses.dataclass
+class DPOTunixPrep(grain.MapTransform):
+  """Prepares DPO data for Tunix.
+  Renames input columns, extracts common prefix if needed, generates masks,and performs
```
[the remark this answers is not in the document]
igorts-git: reformatted

src/maxtext/configs/types.py
```diff
@@ -1201,12 +1217,23 @@ class OlmoGrainDataset(BaseModel):
...
 
+class DPO(BaseModel):
+  """Configuration for DPO and ORPO preference optimization algorithms."""
+
+  algo: Literal["dpo", "orpo"] = Field("dpo", description="Alignment algorithm to use.")
+  dpo_beta: float = Field(0.1, description="Beta parameter for DPO.")
+  orpo_lambda: float = Field(0.1, description="Weight for preference loss in ORPO.")
+  dpo_label_smoothing: float = Field(0.0, ge=0.0, le=1.0, description="Label smoothing for DPO.")
+  max_prompt_length: int | None = Field(
+      None,
```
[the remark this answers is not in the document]
igorts-git: applied with "gt=0" instead of the suggested "ge=0"

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    """Pads or trims an array to a specific length.
+
+    When left=True (for prompts), trims from the left to keep the suffix (closest context).
+    When left=False (for responses), trims from the right to keep the prefix.
+    """
+    x = np.asarray(x)
+    pad_amount = max(length - x.shape[0], 0)
+    if left:
+      pad_width = ((pad_amount, 0),)
+      x_trimmed = x[-length:]
```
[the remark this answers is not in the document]
igorts-git: Instead added an assertion in a different place to make sure that length is never 0

src/maxtext/configs/base.yml
```diff
@@ -680,8 +680,12 @@ global_rampup_samples: 500
 
 # direct preference optimization (DPO)
 use_dpo: False
-dpo_label_smoothing: 0.0
-dpo_beta: 0.1
+dpo:
```
SurbhiJainUSC: Do we need to add these DPO configs to base.yml too?

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+"""DPO specific input pipeline utilities."""
+
+import dataclasses
+import grain.python as grain
+import numpy as np
+
+
+@dataclasses.dataclass
+class DPOTunixPrep(grain.MapTransform):
```
SurbhiJainUSC: Please rename it to `DPODataFormatting` or something similar

tests/post_training/unit/dpo_data_processing_test.py
```diff
@@ -0,0 +1,393 @@
...
+    )
+    # 'rejected' column is missing
+    sample = {"input": np.array([1, 2]), "chosen": np.array([3, 4])}
+    with self.assertRaisesRegex(KeyError, "Column 'rejected' not found in the dataset"):
+      prep.map(sample)
+
+
+@pytest.mark.external_training
+class TestDPOPipelineIntegration(unittest.TestCase):
+  """End-to-end DPO pipeline integration tests."""
```
SurbhiJainUSC: Is this really an integ test?

src/maxtext/configs/post_train/dpo.yml
```diff
@@ -24,8 +30,6 @@ hf_eval_split: 'test'
 
 gradient_clipping_threshold: 10.0
 learning_rate: 5.0e-7
-dpo_label_smoothing: 0.0
-dpo_beta: 0.1
```
gagika, line 28: will this break older DPO code, if so should we just leave as it is for now, perhpas comment that used for older DPO?

gagika (approved):

src/maxtext/configs/post_train/dpo.yml
igorts-git: Unfortunately, this PR breaks the older DPO code, not just due to the configs, but also in how the dataset is loaded. Once this and the next PR in the series is merged, I plan to follow up with a PR that completely deletes the legacy DPO implementation.

src/maxtext/configs/base.yml
igorts-git: Sorry, I don't fully understand the comment. Are you suggesting to remove these configs from base.yml?

src/maxtext/input_pipeline/dpo_utils.py
igorts-git: Renamed the class.

tests/post_training/unit/dpo_data_processing_test.py
igorts-git: you're right, it does not fall into the "integration test" category. rephrased.

src/maxtext/configs/types.py
```diff
@@ -2887,6 +2903,16 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
           raise ValueError("For multimodal SFT, `sft_train_on_completion_only` must be True.")
         if self.packing:
           raise ValueError("For multimodal SFT, `packing` is not yet supported.")
+    if self.use_dpo:
+      if self.packing:
+        raise ValueError("For DPO/ORPO, `packing` is not supported.")
+      if self.dpo.max_prompt_length is not None and self.dpo.max_prompt_length > self.max_target_length:
```
aireenmei: In the case of max_prompt_length == max_target_length, it will cause max_response_length=0 error in DPODataFormatting, should we guard it here?

src/maxtext/configs/base.yml
A9isha: most likely these `dpo` related parameters ended up being in `base.yml` due to historical reasons. if we have them in the `dpo.yaml` and we have separate yml files in the `configs/post_train` directory then it makes sense to remove them from `base.yml` since the ones in `base.yml` are meant to be shareable across multiple use-cases.

aireenmei (approved):
LGTM. But please note that both tfds and hf pipelines are planned for deprecation. There's already support for [SFT](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/input_pipeline/grain_data_processing.py#L316) and [DPO](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/input_pipeline/grain_data_processing.py#L258) in Grain pipeline. It would be great if we can have follow up changes to enable the same support in Grain pipeline

A9isha (approved):
thanks Igor

Approved barring the one comment

src/maxtext/configs/types.py
igorts-git: Updated the guard here. FYI, There is a slightly more comprehensive assertion in dpo_utils.py that can trigger in a few more edge cases.

src/maxtext/configs/base.yml
igorts-git: OK, thanks for the details. I removed the "dpo:" section from `base.yml`

PR #3904, opened 2026-05-14
Add Gemma 4 E2B / E4B (text) support to MaxText

gagika:
# Description

Adds Gemma 4 small variants — **E2B** and **E4B** (text-only) — to MaxText.

These are the smaller members of the Gemma 4 family. They share the
broader Gemma 4 attention / norm structure but introduce two new features
that drive their parameter efficiency:

- **Per-Layer Embeddings (PLE).** Each decoder layer consumes a per-layer
  slice of an extra embedding tensor injected by a new `Gemma4SmallPLE`
  block. Controlled by `hidden_size_per_layer_input` /
  `vocab_size_per_layer_input`.
- **KV sharing.** The last `num_kv_shared_layers` decoder layers reuse
  K / V from the most recent non-shared layer of the same attention type
  (sliding↔sliding, full↔full). E2B additionally widens the MLP on those
  shared layers (`use_double_wide_mlp: true`) to compensate for the
  missing parameters.

Both features carry per-layer state that is not expressible inside
`nn.scan`, so a new `GEMMA4_SMALL` `DecoderBlockType` is added with its
own non-scanned execution path (`Decoder._apply_gemma4_small_layers`).
The model validator enforces `scan_layers=False` for these variants.

### What's included

- New model file `src/maxtext/models/gemma4_small.py` (PLE + attention
  with optional KV sharing + decoder layer).
- New configs `configs/models/gemma4-e2b.yml` and `gemma4-e4b.yml`.
- HF round-trip: `hf_model_configs.py`, `hf_shape.py`,
  `param_mapping.py` updated to handle PLE params, KV-shared layers, and
  the (optional) double-wide MLP.
- TFLOP/MFU accounting: `calculate_gemma4_small_tflops_training_per_device`.
- Config plumbing: `DecoderBlockType.GEMMA4_SMALL`, four new
  `Attention` fields in `configs/types.py`, base.yml defaults, and
  validation that rejects `scan_layers=true` / `use_multimodal=true` for
  E2B / E4B.

### Out of scope

- **Multimodal.** E2B / E4B ship a vision tower in their HF configs, but
  MaxText support for the gemma4-small vision encoder (clipped linears
  in particular) is not in this PR. `use_multimodal=true` is rejected by
  the validator with a clear error.
- **Scanned layers.** Per-layer KV sharing isn't expressible under
  `nn.scan`; rejected by the validator.

# Tests

- New unit tests:
  - `tests/unit/gemma4_small_test.py` — attention-pattern dispatch,
    layer-type tuples, KV donor/shared-layer mapping for both variants.
  - `tests/unit/flop_calculation_test.py::test_calculate_gemma4_small_tflops_*` —
    closed-form TFLOP accounting matching the layer/donor structure.
  - `tests/unit/configs_test.py` — E2B / E4B yml configs are loaded by
    the existing config-instantiation sweep.
- End-to-end forward-pass logit checks:
  - `tests/end_to_end/tpu/gemma4/e2b/{convert_gemma4,convert_gemma4_pt}.sh`
  - `tests/end_to_end/tpu/gemma4/e4b/{convert_gemma4,convert_gemma4_pt}.sh`
  - Each converts the HF checkpoint with `to_maxtext`, then runs
    `forward_pass_logit_checker` against the HF model with
    `--max_kl_div=0.03`. This is the recommended smoke test after
    touching the model code, param map, or either YAML.

# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/utils/param_mapping.py
+++ b/src/maxtext/checkpoint_conversion/utils/param_mapping.py
@@ -2528,6 +2528,236 @@ def GEMMA4_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=False
   return mapping
 
 
+def GEMMA4_SMALL_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=False):
+  """MaxText↔HF weight-path map for Gemma 4 small (E2B / E4B).
+
+  The small variants thread per-layer state (PLE input + donor K/V) through
+  the layer loop, so scanned blocks are not supported.
+  """
+  if scan_layers:
+    raise NotImplementedError("Scan layers is not supported for the gemma4_small decoder block.")
+
+  tcfg = config.get("text_config", config)
+  nlayers = tcfg["num_hidden_layers"]
+  num_kv_shared = tcfg.get("num_kv_shared_layers", 0) or 0
+  first_shared = max(0, nlayers - num_kv_shared) if num_kv_shared > 0 else nlayers
+  vcfg = config.get("vision_config", {})
+  text_base = "model.language_model" if vcfg else "model"
+
+  mapping = {
+      "params-token_embedder-embedding": f"{text_base}.embed_tokens.weight",
+      "params-decoder-decoder_norm-scale": f"{text_base}.norm.weight",
+  }
+
+  ple_dim = tcfg.get("hidden_size_per_layer_input", 0) or 0
+  if ple_dim > 0:
+    mapping.update(
+        {
+            "params-decoder-per_layer_embedder-embed_tokens_per_layer": (f"{text_base}.embed_tokens_per_layer.weight"),
+            "params-decoder-per_layer_embedder-per_layer_model_projection-kernel": (
+                f"{text_base}.per_layer_model_projection.weight"
+            ),
+            "params-decoder-per_layer_embedder-per_layer_projection_norm-scale": (
+                f"{text_base}.per_layer_projection_norm.weight"
+            ),
+        }
+    )
+
+  for lyr in range(nlayers):
+    prefix = f"params-decoder-layers_{lyr}"
+    hf_prefix = f"{text_base}.layers.{lyr}"
+    is_shared = lyr >= first_shared > 0
+    mapping.update(
+        {
+            f"{prefix}-pre_self_attention_norm-scale": f"{hf_prefix}.input_layernorm.weight",
+            f"{prefix}-post_self_attention_norm-scale": f"{hf_prefix}.post_attention_layernorm.weight",
+            f"{prefix}-pre_ffw_norm-scale": f"{hf_prefix}.pre_feedforward_layernorm.weight",
+            f"{prefix}-post_ffw_norm-scale": f"{hf_prefix}.post_feedforward_layernorm.weight",
+            f"{prefix}-self_attention-query-kernel": f"{hf_prefix}.self_attn.q_proj.weight",
+            f"{prefix}-self_attention-query_norm-scale": f"{hf_prefix}.self_attn.q_norm.weight",
+            f"{prefix}-self_attention-out-kernel": f"{hf_prefix}.self_attn.o_proj.weight",
+            f"{prefix}-mlp-wi_0-kernel": f"{hf_prefix}.mlp.gate_proj.weight",
+            f"{prefix}-mlp-wi_1-kernel": f"{hf_prefix}.mlp.up_proj.weight",
+            f"{prefix}-mlp-wo-kernel": f"{hf_prefix}.mlp.down_proj.weight",
+            f"{prefix}-layer_scalar": f"{hf_prefix}.layer_scalar",
+        }
+    )
+    if not is_shared:
+      mapping.update(
+          {
+              f"{prefix}-self_attention-key-kernel": f"{hf_prefix}.self_attn.k_proj.weight",
+              f"{prefix}-self_attention-value-kernel": f"{hf_prefix}.self_attn.v_proj.weight",
+              f"{prefix}-self_attention-key_norm-scale": f"{hf_prefix}.self_attn.k_norm.weight",
+              f"{prefix}-self_attention-value_norm-scale": f"{hf_prefix}.self_attn.v_norm.weight"
+              if maxtext_config.v_norm_with_scale
+              else None,
+          }
+      )
+    if ple_dim > 0:
+      mapping.update(
+          {
+              f"{prefix}-per_layer_input_gate-kernel": f"{hf_prefix}.per_layer_input_gate.weight",
+              f"{prefix}-per_layer_projection-kernel": f"{hf_prefix}.per_layer_projection.weight",
+              f"{prefix}-post_per_layer_input_norm-scale": f"{hf_prefix}.post_per_layer_input_norm.weight",
+          }
+      )
+
+  # TODO: gemma4-small multimodal not yet supported — vision-encoder mappings below are dead.
+  if maxtext_config.use_multimodal and vcfg:
+    nvis = vcfg.get("num_hidden_layers", 0)
+    mapping.update(
+        {
+            "params-vision_encoder-Gemma4VisionEncoderLayer_0-vision_entry-input_projection-kernel": (
+                "model.vision_tower.patch_embedder.input_proj.weight"
+            ),
+            "params-vision_encoder-Gemma4VisionEncoderLayer_0-vision_entry-pos_emb_param": (
+                "model.vision_tower.patch_embedder.position_embedding_table"
+            ),
+            "params-vision_encoder-Gemma4VisionProjector_0-projection-kernel": (
+                "model.embed_vision.embedding_projection.weight"
+            ),
+        }
+    )
+    if vcfg.get("standardize", False):
+      mapping.update(
+          {
+              "params-vision_encoder-Gemma4VisionEncoderLayer_0-std_scale": "model.vision_tower.std_scale",
+              "params-vision_encoder-Gemma4VisionEncoderLayer_0-std_bias": "model.vision_tower.std_bias",
+          }
+      )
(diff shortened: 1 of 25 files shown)


src/maxtext/models/gemma4_small.py
```diff
@@ -0,0 +1,643 @@
...
+  cos_half = jnp.cos(sinusoid).astype(x.dtype)
+  sin = jnp.concatenate([sin_half, sin_half], axis=-1)
+  cos = jnp.concatenate([cos_half, cos_half], axis=-1)
+
+  x1, x2 = jnp.split(x, 2, axis=-1)
+  rotated = jnp.concatenate([-x2, x1], axis=-1)
+  return (x * cos) + (rotated * sin)
+
+
+class Gemma4SmallAttention(nnx.Module):
```
aireenmei: Should we integrate into maxtext.layers.attentions like other models?

src/maxtext/models/gemma4_small.py
```diff
@@ -0,0 +1,643 @@
...
+    return out.astype(cfg.dtype)
+
+
+PLEToLinen = nnx_wrappers.to_linen_class(
+    Gemma4SmallPLE,
+    base_metadata_fn=initializers.variable_to_logically_partitioned,
+)
+
+
+def _apply_gemma4_partial_rope(
```
aireenmei: Should we have the rope modules in layers/embeddings.py like other RoPE?

aireenmei (commented):
Thanks for the rapid implementation! I wonder if you have test results from forward_pass_logit_checker? Also do you add some unit tests for comparison with torch on the new modules such as Gemma4SmallPLE, Gemma4SmallAttention, Gemma4SmallDecoderLayer? https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/gemma4_layers_test.py This was added recently.

src/maxtext/models/gemma4_small.py
shuningjin: Can we reuse `Gemma4PartialRotaryEmbedding`?

https://github.com/AI-Hypercomputer/maxtext/blob/e835a6c582056dcda9c829d8f04f4f55342956cd/src/maxtext/layers/embeddings.py#L510

src/maxtext/models/gemma4_small.py
```diff
@@ -0,0 +1,643 @@
...
+      key = jnp.repeat(key, group, axis=-2)
+      value = jnp.repeat(value, group, axis=-2)
+
+    if self.config.float32_qk_product:
+      query = query.astype(jnp.float32)
+      key = key.astype(jnp.float32)
+
+    # HF Gemma 4 attention is unscaled (scaling=1.0); the trained weights
+    # absorb the 1/sqrt(d) factor.
+    logits = jnp.einsum("bqhd,bkhd->bhqk", query, key)
```
shuningjin: nit: jnp.einsum(..., precision=self.config.matmul_precision)

src/maxtext/models/gemma4_small.py
shuningjin: +1. Can we merge this into `Attention`, e.g., with optional call to `shared_key` and `shared_value`? This might be easier to reuse flash attention.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -2528,6 +2528,239 @@ def GEMMA4_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=False
...
+    return input_tensor.T.reshape(target_shape)
+
+  def scale_rmsnorm_layer(input_tensor, target_shape):
+    # Gemma 4 folds the +1 shift into the checkpoint scale, so no offset.
+    return input_tensor.reshape(target_shape)
+
+  hooks["params-token_embedder-embedding"] = pad_hf_embedding_layer
+  hooks["params-decoder-decoder_norm-scale"] = scale_rmsnorm_layer
+
+  def identity_hook(input_tensor, target_shape):
```
shuningjin: No need for explicit identity hook. By default, any param with unspecified hook will use identity.

https://github.com/AI-Hypercomputer/maxtext/blob/3e71df757698a8be6dab9bf471052df99c67a750/src/maxtext/checkpoint_conversion/utils/utils.py#L137-L141

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -2528,6 +2528,239 @@ def GEMMA4_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=False
...
+    if not vcfg.get("standardize", False):
+
+      def make_ones(_input, target_shape):
+        return np.ones(target_shape, dtype=np.float32)
+
+      def make_zeros(_input, target_shape):
+        return np.zeros(target_shape, dtype=np.float32)
+
+      hooks["params-vision_encoder-Gemma4VisionEncoderLayer_0-std_scale"] = make_ones
+      hooks["params-vision_encoder-Gemma4VisionEncoderLayer_0-std_bias"] = make_zeros
```
shuningjin: Could you explain what this do? Is it equivalent to skip execution in encoder?

Seems multimodal is not supported yet, maybe add some `#TODO` to indicate

tests/end_to_end/tpu/gemma4/e2b/convert_gemma4_base.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# End-to-end conversion + forward-pass logit check for Gemma 4 E2B (pre-trained).
```
shuningjin, line 3: nit: `convert_gemma4_pt.sh`‎ -> `convert_gemma4_base.sh`. I often think `pt` as pytorch.

tests/end_to_end/tpu/gemma4/Run_Gemma4.md
```diff
@@ -18,10 +18,19 @@
...
 
-Gemma 4 in MaxText is available in two sizes with Dense (31B) and Mixture-of-Experts (MoE) (26B A4B) architectures, and is well-suited for tasks like text generation, coding, and reasoning. The models are designed for enhanced performance and efficiency, capable of running on environments ranging from laptops and servers.
+Gemma 4 in MaxText is available in four sizes — two small variants (E2B, E4B) and the larger Dense (31B) / Mixture-of-Experts (MoE) (26B A4B) configurations — and is well-suited for tasks like text generation, coding, and reasoning. The models are designed for enhanced performance and efficiency, capable of running on environments ranging from laptops and servers.
 
 We provide examples for checkpoint conversion scripts at [tests/end_to_end/tpu/gemma4](https://github.com/AI-Hypercomputer/maxtext/tree/main/tests/end_to_end/tpu/gemma4).
 
+## E2B / E4B small variants
+
+The E2B and E4B configs (`src/maxtext/configs/models/gemma4-e2b.yml`, `gemma4-e4b.yml`) introduce two architecture features beyond the dense Gemma 4 path:
```
shuningjin: nit: use same path format. Could you add that `E` here stands for edge device, and both `E2B` and `E4B` are dense?

shuningjin (commented):
Thanks! I agree with @aireenmei that it would be good if we could reuse rope/attention, along with component-wise unit tests (potentially as follow-up). Some minor comments.

gagika:
> Thanks for the rapid implementation! I wonder if you have test results from forward_pass_logit_checker? Also do you add some unit tests for comparison with torch on the new modules such as Gemma4SmallPLE, Gemma4SmallAttention, Gemma4SmallDecoderLayer? https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/gemma4_layers_test.py This was added recently.

Added those unit test, PTL

for forward logits test, yes, I have done for both models.
I did again after addressing PR feedback:

https://paste.googleplex.com/5790253452492800
https://paste.googleplex.com/6349200758538240

src/maxtext/models/gemma4_small.py
gagika: Done, thanks for pointing it out.

src/maxtext/models/gemma4_small.py
gagika: Done.
Initially I didn't do it not to over-load already complex Attention class with even more flags and branching, but vLLM support (for RL) and Flash attention both are needed, so integrated with Attention.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
gagika: Done.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
gagika: It's not equivalent to skipping the encoder — when multimodal does eventually land, the encoder will run and consume std_scale / std_bias. The placeholder solves a checkpoint-loading problem: HF doesn't ship those two tensors for gemma4-E2B/E4B (their standardize=false), but our MaxText vision tower has the slots wired up. So we synthesize identity values (scale=1, bias=0) at load time, which mathematically  turns the standardize op into a no-op — matching HF's standardize=false semantics without diverging the  model architecture.

Added a short # TODO: gemma4-small multimodal not yet supported.

src/maxtext/models/gemma4_small.py
gagika: done

tests/end_to_end/tpu/gemma4/e2b/convert_gemma4_base.sh
gagika: done

tests/end_to_end/tpu/gemma4/Run_Gemma4.md
gagika: done, PTL.

shuningjin (approved):
Thanks for refactoring attention and adding component test! LGTM.

aireenmei (approved):
Thanks!

PR #3907, opened 2026-05-14
[Qwen3.5] Fix caching bug in GDN layer for autoregressive mode

Rohan-Bierneni:
# Description

Previously when we enabled caching in the GDN layer for Qwen3-Next model bringup, there was a bug in autoregressive mode when running inference with api_server, our benchmarking tool. There was an issue with how api_server was managing the GDN caches resulting in repeated token output in decoding. 

Now, I have fixed this bug by reusing the existing kvcache class in kvcache.py for storing the GDN recurrent_state and conv_state. It seems like the kvcache class is well integrated with api_server in terms of batching, padding, updating, etc. 

For new models with special cache structures, for the fastest way to functional decoding, it seems like reusing the kvcache class as much as possible is the best solution since it is already integrated within our inference framework.

Also, in the GDN layer I have rewritten the gated delta rule function for autoregressive mode. This is because the algorithm by default splits the token into chunk_sizes of 64. However, for autoregressive mode, seq_len=1. Our current gated delta rule did not account for this and we had wasted padding of 63 tokens!. Adding a new gated delta rule function specifically for this has led to performance boost in decoding speed allowing us to run benchmarks.

# Tests

I have run standard decode.py and have also run an inference benchmark on api_server for Qwen3.5, which uses the GDN

decode.py output: https://paste.googleplex.com/6084567757881344

For the MMLU-Pro run of Qwen3.5-35b, the details are in b/502303541

# Checklist

### DIFF
--- a/benchmarks/api_server/maxtext_server.py
+++ b/benchmarks/api_server/maxtext_server.py
@@ -279,13 +279,20 @@ def _collect_batched_requests():
   """Waits for and collects a batch of requests from the queue."""
   batched_items = []
   start_time = time.time()
-  while len(batched_items) < LLM.batch_size and (time.time() - start_time) < BATCH_TIMEOUT_S:
+
+  while len(batched_items) < LLM.batch_size:
+    elapsed = time.time() - start_time
+    remaining = BATCH_TIMEOUT_S - elapsed
+
+    if remaining <= 0:
+      break
+
     try:
-      item = request_queue.get(timeout=0.01)
+      item = request_queue.get(timeout=remaining)
       batched_items.append(item)
     except queue.Empty:
-      if batched_items:
-        break  # Process what we have if timeout is reached
+      break  # We hit the full BATCH_TIMEOUT_S
+
   return batched_items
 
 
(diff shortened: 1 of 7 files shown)


src/maxtext/configs/base.yml
```diff
@@ -489,7 +489,8 @@ logical_axis_rules: [
                       # Attention Activations
                       ['activation_batch_attn', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_heads', ['tensor', 'tensor_transpose', 'tensor_sequence', 'autoregressive']],
-                      ['activation_kv_heads', ['tensor', 'tensor_transpose', 'tensor_sequence']],
+                      # ['activation_kv_heads', ['tensor', 'tensor_transpose', 'tensor_sequence']],
```
entrpn: any reason to comment out instead of removing these?

Rohan-Bierneni: I initially had to make some changes to have TP=4. This is because the kv_heads for the model is 2. This change was only made for being able to run mmlu pro. I have reverted this and updated pr.

entrpn (approved):

parambole (approved):
LGTM! Great work

hengtaoguo (approved):

khatwanimohit (approved):

PR #3922, opened 2026-05-15
[ROCm] optimizations to reduce temp size prior to training starts

cj401-amd:
# Description

The purpose of this PR is to reduce temp (compile) memory consumption. The optimizations have been tested on ROCm (MI355) and CUDA (H100), from which temp size was reduced. However, we would like to see your help for testing them on TPU to check if they are applicable or not. 

 ## Checklist
  - [x] Code follows the project style guidelines (pyink formatting applied)

The PR might be a bit big, however, we can split into several PRs, if that helps to review them on your side. 

| Configuration | PP | EP | Temp (main) | Temp (cj/tmem-fixes-clean) | Reduction |
|---------------|----|----|------------:|------------------:|----------:|
| ds-proxy-se2-e256-h4096 | 8 | 1 | 36.5 GB | 20.4 GB | **−16.1 GB (−44%)** |
| ds-proxy-N1-ep2-pp4     | 4 | 2 | 44.4 GB | 22.9 GB | **−21.5 GB (−48%)** |

We also detail the files changed along with the reason behind them.

## 1. `src/maxtext/configs/base.yml` -- Default `float32_weight_sum` to false

**What changed:** `float32_weight_sum: true` --> `false`

**Why:** When MoE models sum expert weights, using fp32 precision requires promoting all expert weight tensors to float32, then summing, then casting back. This materializes ~2 GB of temporary f32 tensors per device. By defaulting to false, we keep the summation in the model's native dtype (bf16), which avoids the temporary buffers entirely.

**Trade-off:** Slightly reduced numerical precision in expert weight summation. For bf16 training this is unlikely to matter in practice, and the parameter remains configurable for users who need strict fp32 summation.

---

## 2. `src/maxtext/configs/types.py` -- Pydantic field default alignment

**What changed:** `float32_weight_sum` field default `True` --> `False`, description updated.

**Why:** Mirrors the base.yml change so the Pydantic config schema and the YAML config file agree on the default.

---

## 3. `src/maxtext/kernels/gather_reduce_sc.py` -- JIT decorator restructuring

**What changed:** `@jax.jit(static_argnames=[...])` --> `@functools.partial(jax.jit, static_argnames=[...])`

**Why:** The direct `@jax.jit(...)` decorator form can interact poorly with certain JAX versions when the decorated function is further wrapped or inspected. `functools.partial` preserves function metadata (name, docstring) and ensures the decorator is properly composed. This is a robustness fix that prevents subtle compilation issues.

---

## 4. `src/maxtext/utils/sharding.py` -- `skip_trivial_specs` + `remove_size_one_mesh_axis`

**What changed:** Two additions:

**(a) `skip_trivial_specs` parameter on `maybe_shard_with_logical()`:**
When `skip_trivial_specs=True`, the function checks if all axes in the sharding spec resolve to `None` or `()`. If so, it returns the input tensor unchanged -- no `with_sharding_constraint` call is emitted. This is the foundational optimization used by attentions.py, pipeline.py, and mixtral.py.

**Why:** `jax.lax.with_sharding_constraint` is not free even when the spec is trivial. XLA can materialize copy buffers to enforce a constraint that effectively does nothing. Skipping it when the spec is all-None avoids those temporaries.

**(b) `remove_size_one_mesh_axis()` function:**
Filters out mesh axes from a `PartitionSpec` where the corresponding mesh dimension has size 1. For example, on a single-node setup where `stage=1`, any sharding over `"stage"` is meaningless and can be dropped.

**Why:** Size-1 mesh axes create sharding overhead (XLA inserts identity reshards) without providing any actual parallelism. Removing them simplifies the HLO and eliminates associated temporary buffers. The existing `logical_to_mesh_axes()` was refactored to call this new function automatically.

---

## 5. `src/maxtext/layers/normalizations.py` -- Replace einsum with multiply in RMSNorm

**What changed:**
- `jnp.einsum("...k,k->...k", y, effective_scale, out_sharding=...)` --> `y * effective_scale` + separate `jax.lax.with_sharding_constraint`
- Added conditional: `effective_scale = scale + self.scale_offset if self.scale_offset != 0.0 else scale`

**Why:** The `einsum("...k,k->...k", ...)` operation is semantically just element-wise multiplication (broadcasting the 1D scale across the leading dims). However, XLA's einsum lowering can generate larger intermediate buffers than a direct multiply, especially inside `lax.scan` (pipeline parallelism). Inside a scan body, these intermediates get hoisted into the scan carry by XLA's `loop_broadcast_fusion` pass, permanently increasing the carry size.

By using direct multiplication, XLA generates a simpler `broadcast + multiply` HLO sequence with no intermediate allocation. The sharding constraint is decoupled and applied only when needed (`out_sharding is not None`).

The `scale_offset != 0.0` check avoids a pointless addition when offset is zero (the standard RMSNorm case, vs. RMSNorm with learnable offset).

---

## 6. `src/maxtext/layers/attentions.py` -- `skip_trivial_specs=True`

**What changed:** Added `skip_trivial_specs=True` to `maybe_shard_with_logical()` call in the Attention class.

**Why:** On mesh configurations where the attention output sharding resolves to all-None (e.g., no tensor parallelism), this avoids materializing a copy buffer. This is a minor but consistent optimization applied across the codebase.

---

## 7. `src/maxtext/layers/embeddings.py` -- 2 GB iota_embed guard + sharding constraints

**What changed:**
- **Iota guard:** Compute one-hot tensor size as `batch * seq * vocab * dtype_size`. If >2 GB, fall back to gather-based embedding instead of iota (one-hot) embedding.
- **Conditional sharding:** Only create `out_pspec`/`out_sharding` when `shard_mode == ShardMode.EXPLICIT`.
- **Logical constraints:** Added `nn.with_logical_constraint(output, ...)` after the embedding lookup.

**Why:** Iota embedding works by creating a one-hot matrix of shape `[batch, seq, vocab]` and multiplying by the embedding table. For large vocabularies (e.g., vocab=102400), this one-hot tensor can be enormous: `8 * 4096 * 102400 * 2 bytes = 6.7 GB` in bf16. XLA materializes this entire tensor even though only one element per row is non-zero. The gather-based path (`jnp.take`) uses constant memory regardless of vocabulary size.

The 2 GB threshold is conservative -- it covers most common model configs while still allowing small-vocab models to use the faster iota path.

The sharding spec was being computed unconditionally even when `shard_mode != EXPLICIT`. Moving it under a conditional avoids unnecessary mesh lookups.

---

## 8. `src/maxtext/layers/attention_op.py` -- Causal mask for synthetic data

**What changed:** For `dataset_type == "synthetic"`, skip materializing the attention mask tensors entirely. Use `mask_type="causal"` directly instead of computing mask arrays.

**Why:** With synthetic data, segment IDs are always all-ones (a single segment per sequence). The segment mask is therefore all-True, so the combined attention mask reduces to pure causal masking. By using `mask_type="causal"` natively, we avoid:
1. Materializing `f32/s32[batch, 1, seq, seq]` mask tensors (~5 GiB for seq=4096, batch=8)
2. These masks being hoisted into the scan carry by XLA's `loop_broadcast_fusion` pass

Also removed two dead parameters from the `DotProductAttention` call:
- `scale_factor=1.0` (no-op: default is already 1.0)
- `context_parallel_strategy=...` (not used in this non-CP code path)

---

## 9. `src/maxtext/layers/moe.py` -- Tile truncation + remove redundant astype

**What changed:**
**(a) Tile size tuples:** Reduced from 9-tuples to 3-tuples (forward-only).

**(b) Removed `.astype(self.dtype)`** from `return intermediate_layer`.

**Why (a):** The 9-tuple `wi_tile_size` and `wo_tile_size` contain tiling parameters for forward (fwd), dlhs (backward LHS), and drhs (backward RHS) passes of the Group Matrix Multiply (GMM). When `megablox=False` (the JAX `ragged_dot` path), only the forward pass tile values are used -- the backward-pass tile values are allocated in the tuple but never read. By truncating to 3-tuples, we avoid allocating and passing unused tile parameters.

**Why (b):** The `intermediate_layer` is already computed in the correct dtype from the matmul operations. The explicit `.astype(self.dtype)` creates a redundant copy that XLA cannot always elide (especially inside scan bodies). Removing it avoids one tensor-sized temporary.

---

## 10. `src/maxtext/layers/pipeline.py` -- Replace ppermute with slice/concat, reduce remat

**What changed:** Multiple optimizations in the `Pipeline` class (not `CircularPipeline`):

**(a) Replace `shard_map+ppermute` with `lax.slice`/`jnp.concatenate`/`jnp.pad`:**

The old `_rotate_right` used `@jax.shard_map` with `jax.lax.ppermute` for circular shifting across pipeline stages. The new version uses `lax.slice_in_dim` + `jnp.concatenate`:
```python
last = lax.slice_in_dim(arr, N-1, N, axis=0)
rest = lax.slice_in_dim(arr, 0, N-1, axis=0)
return jnp.concatenate([last, rest], axis=0)
```

The old `_shift_right` used `ppermute` + zero-fill. The new version uses `jnp.pad` + `lax.slice`:
```python
padded = jnp.pad(arr, [[1,0]] + [[0,0]]*(ndim-1))
return lax.slice(padded, [0]*ndim, arr.shape)
```

Similarly, `_update_state_io` was converted from `shard_map`+`ppermute`-based left-shift to `pad`+`slice_in_dim`-based left-shift.

**Why:** XLA's `loop_broadcast_fusion` pass hoists ppermute's collective communication buffers into the scan carry. For `pp=4`, this added **10-15 GB** of temporary memory. Slice-based operations don't use collective scratch space -- they operate on local arrays. The operations are mathematically identical (they perform the same circular/linear shift) but have very different memory footprints inside scan loops.

**(b) Remove unnecessary sharding constraints:**
- Removed sharding on `shift`, `first_stage_in`, `microbatches_processed` -- these are small tensors where sharding adds overhead without benefit.
- Removed `out_sharding` from `broadcasted_iota` call.
- Added `skip_trivial_specs=True` to the base sharding helper.

**(c) Remove `decoder_layer_input` from remat save list:**
Changed from `save_only_these_names("iteration_input", "decoder_layer_input")` to `save_only_these_names("iteration_input")`. The `iteration_input` checkpoint already contains the data needed for backward pass. Saving `decoder_layer_input` too created two redundant stacked checkpoint buffers (~3.3 GB).

**(d) Remove `meta.remove_axis` call:**
The axis removal logic for vmap weights was unnecessary and created intermediate metadata objects.

**(e) Replace indexed gather with `dynamic_slice_in_dim`:**
`x.at[idx].get(out_sharding=...)` creates intermediate index tuples. `jax.lax.dynamic_slice_in_dim(x, i, 1, dim)` is a single primitive with no intermediate allocation.

---

## 11. `src/maxtext/models/deepseek.py` -- Remove reshard calls

**What changed:**
- Added import: `from maxtext.utils.sharding import remove_size_one_mesh_axis`
- Changed MoE kernel init scale from `self.config.dense_init_scale` to `1.0`
- Replaced explicit `jax.reshard()` in/out pair with `remove_size_one_mesh_axis()`-based PartitionSpec
- Expanded PartitionSpec to include `"fsdp_transpose"` and `"context"` axes

**Why:** The original code captured the input sharding, resharded inputs to a specific activation PartitionSpec, ran the MoE block, then resharded outputs back to the original input sharding. Each `jax.reshard()` creates a full-activation-sized temporary buffer. By using `remove_size_one_mesh_axis()` to simplify the PartitionSpec (dropping axes that have size 1 on the current mesh), we get the same effective sharding without the explicit reshard calls. This eliminates two activation-sized temporaries.

The PartitionSpec was also expanded to include `fsdp_transpose` and `context` axes for correctness on meshes that use those dimensions.

---

## 12. `src/maxtext/models/mixtral.py` -- NNX to Linen conversion

**What changed:** Complete rewrite from `nnx.Module` to `nn.Module` (Linen).

- Removed `from flax import nnx` and `nnx_wrappers` imports
- Added `attention_as_linen`, `rms_norm`, `maybe_shard_with_logical` imports
- Changed class from `nnx.Module` (with `__init__`) to `nn.Module` (with `@nn.compact` `__call__`)
- All layer instantiation moved from `__init__` into `__call__`
- Added local `shard()` helper with `skip_trivial_specs=True`
- Removed `MixtralDecoderLayerToLinen = nnx_wrappers.to_linen_class(...)` wrapper -- the class IS Linen now

**Why:** The NNX-to-Linen wrapper (`nnx_wrappers.to_linen_class`) creates an NNX module graph and then converts it to Linen form. This process materializes intermediate module objects and parameter trees that consume memory during `eval_shape` and `jit` tracing. By writing the layer directly in Linen, we bypass this overhead entirely.

The Linen `@nn.compact` style also defers all parameter creation to call time, avoiding eager allocation of layer parameters during module construction. This is particularly impactful for MoE models where the number of parameters is large.

Weight names are preserved through explicit `name=` arguments to ensure checkpoint compatibility.

---

## 13. `src/maxtext/trainers/pre_train/train.py` -- Flax config + grad dtype guard

**What changed:**
**(a)** Added `import flax` + `flax.config.update("flax_always_shard_variable", False)` at module level, guarded by `try/except LookupError`.

**(b)** Wrapped the gradient dtype tree_map in `if config.grad_dtype != jnp.float32:`.

**Why (a):** When `flax_always_shard_variable` is True (the default in some Flax versions), Flax inserts `with_sharding_constraint` on every variable during creation. This generates intermediate sharding copies even for variables that don't need explicit sharding. Setting it to False means sharding is only applied where the model code explicitly requests it.

**Why (b):** The tree_map that casts gradients from fp32 to `config.grad_dtype` is a no-op when `grad_dtype` is already fp32. But `tree_map` still walks the entire gradient tree and calls the lambda on every leaf, which can trigger unnecessary tensor copies if XLA doesn't fully elide the identity cast. Skipping the tree_map entirely when it's a no-op avoids this overhead.

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -186,7 +186,7 @@ logits_dot_in_fp32: false  # whether to use fp32 in logits_dense or shared_embed
 cast_logits_to_fp32: true # whether to cast the logits to fp32. the higher precision is generally beneficial, but it can vary slightly.
 float32_qk_product: false # in dot_product attention, whether to cast to fp32 the inputs to qk product
 float32_logits: false # in dot_product attention, whether to cast to fp32 the inputs to softmax
-float32_weight_sum: true # whether to use full fp32 precision to sum expert weights for numerical stability
+float32_weight_sum: false # whether to use fp32 for MoE expert weight summation; true adds ~2 GB f32 temp per device
 float32_gate_logits: false # whether to cast inputs to fp32 to compute MoE gate logits for numerical stability
 
 # multi-token prediction configs
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -319,6 +319,7 @@ scan_pipeline_repeats: false
 scan_layers_per_stage: false
 set_remat_policy_on_pipeline_iterations: true
 set_remat_policy_on_layers_per_stage: false
+pipeline_save_decoder_layer_input: true  # set to false to reduce pipeline tmem at cost of recomputing decoder layer inputs in backward pass
 
 
 # Choose 'remat_policy' between 'minimal_with_context', 'minimal', 'save_dot_with_context_except_mlp', 'save_dot_except_mlpwi', 'save_dot_except_mlp',
--- a/src/maxtext/kernels/gather_reduce_sc.py
+++ b/src/maxtext/kernels/gather_reduce_sc.py
@@ -55,7 +55,8 @@ _F32 = VectorTypeHelper(ir.F32Type.get)
 _BF16 = VectorTypeHelper(ir.BF16Type.get)
 
 
-@jax.jit(
+@functools.partial(
+    jax.jit,
     static_argnames=[
         "reduce_group_size",
         "single_sc",
--- a/src/maxtext/kernels/gather_reduce_sc.py
+++ b/src/maxtext/kernels/gather_reduce_sc.py
@@ -55,8 +55,8 @@ _F32 = VectorTypeHelper(ir.F32Type.get)
 _BF16 = VectorTypeHelper(ir.BF16Type.get)
 
 
-@functools.partial(
-    jax.jit,
+# fmt: off
+@jax.jit(
     static_argnames=[
         "reduce_group_size",
         "single_sc",
@@ -70,6 +70,7 @@ _BF16 = VectorTypeHelper(ir.BF16Type.get)
         "topk_wgt_zero_nan",
     ],
 )
+# fmt: on
 def sc_gather_reduce(
     op: jax.Array,
     idx: jax.Array,
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -1588,13 +1588,22 @@ class AttentionOp(nnx.Module):
       dummy_attn_mask = None
       mask_type = "causal"
     else:
-      # Default case: no packing, no context parallelism
-      dummy_attn_mask = jnp.zeros(
-          (1, 1, 1, self.max_target_length, self.max_target_length),
-          dtype=jnp.uint8,
-      )
-      attn_mask = self.generate_attention_mask(query, key, decoder_segment_ids, model_mode)
-      attn_mask = jnp.where((attn_mask >= DEFAULT_MASK_VALUE * 0.5), 0, 1).astype(jnp.uint8)
+      # Default case: no packing, no context parallelism.
+      # For synthetic data, segment IDs are always all-ones (one segment per sequence), so
+      # the segment mask is all-True and the combined mask reduces to pure causal masking.
+      # Use mask_type="causal" directly to avoid materializing f32/s32[seq,seq] tensors that
+      # XLA loop_broadcast_fusion hoists into the pipeline scan carry (+5 GiB temp memory).
+      if self.config.dataset_type == "synthetic":
+        attn_mask = None
+        dummy_attn_mask = None
+        mask_type = "causal"
+      else:
+        dummy_attn_mask = jnp.zeros(
+            (1, 1, 1, self.max_target_length, self.max_target_length),
+            dtype=jnp.uint8,
+        )
+        attn_mask = self.generate_attention_mask(query, key, decoder_segment_ids, model_mode)
+        attn_mask = jnp.where((attn_mask >= DEFAULT_MASK_VALUE * 0.5), 0, 1).astype(jnp.uint8)
 
     dpa_layer = DotProductAttention(
         head_dim=head_dim,
@@ -1607,12 +1616,10 @@ class AttentionOp(nnx.Module):
         dtype=self.dtype,
         float32_logits=self.float32_logits,
         qkv_layout=qkv_layout,
-        scale_factor=1.0,
         transpose_batch_sequence=False,
         window_size=sliding_window_size,
         context_parallel_causal_load_balanced=self.config.context_parallel_load_balance,
         context_parallel_axis=self.config.context_sharding,
-        context_parallel_strategy=self.config.context_parallel_strategy,
         max_segments_per_seq=max_segments_per_seq,
     )
 
--- a/src/maxtext/layers/normalizations.py
+++ b/src/maxtext/layers/normalizations.py
@@ -88,8 +88,11 @@ class RMSNorm(nnx.Module):
       scale = jax.device_put(scale, max_utils.device_space())
 
     scale = jnp.asarray(scale, self.dtype)
-    effective_scale = scale + self.scale_offset
-    return jnp.einsum("...k,k->...k", y, effective_scale, out_sharding=out_sharding)
+    effective_scale = scale + self.scale_offset if self.scale_offset != 0.0 else scale
+    y = y * effective_scale
+    if out_sharding is not None:
+      y = jax.lax.with_sharding_constraint(y, out_sharding)
+    return y
 
 
 class GlobalRMSNorm(RMSNorm):
--- a/src/maxtext/layers/normalizations.py
+++ b/src/maxtext/layers/normalizations.py
@@ -22,7 +22,7 @@ from flax.linen import initializers as linen_initializers
 import jax
 from jax import lax
 import jax.numpy as jnp
-from jax.sharding import NamedSharding
+from jax.sharding import NamedSharding, reshard
 from maxtext.common.common_types import Array, DType, ShardMode
 from maxtext.layers import nnx_wrappers
 from maxtext.layers.initializers import Initializer, variable_to_logically_partitioned
@@ -78,7 +78,10 @@ class RMSNorm(nnx.Module):
 
     if not self.with_scale:
       if out_sharding is not None:
-        y = jax.lax.with_sharding_constraint(y, out_sharding)
+        if self.shard_mode == ShardMode.EXPLICIT:
+          y = reshard(y, out_sharding)
+        else:
+          y = jax.lax.with_sharding_constraint(y, out_sharding)
       return y
 
     scale = self.scale.get_value()
@@ -91,7 +94,10 @@ class RMSNorm(nnx.Module):
     effective_scale = scale + self.scale_offset if self.scale_offset != 0.0 else scale
     y = y * effective_scale
     if out_sharding is not None:
-      y = jax.lax.with_sharding_constraint(y, out_sharding)
+      if self.shard_mode == ShardMode.EXPLICIT:
+        y = reshard(y, out_sharding)
+      else:
+        y = jax.lax.with_sharding_constraint(y, out_sharding)
     return y
 
 
--- a/src/maxtext/layers/pipeline.py
+++ b/src/maxtext/layers/pipeline.py
@@ -118,6 +118,7 @@ class PipelineBase(nn.Module):
         rules=self.config.logical_axis_rules,
         debug_sharding=self.config.debug_sharding,
         extra_stack_level=1,
+        skip_trivial_specs=True,
     )
 
   def _maybe_shard_with_name(self, inputs, sharding_name):
@@ -139,7 +140,6 @@ class PipelineBase(nn.Module):
     # Setup potential input from state_io, which has a rotating microbatch index (size of microbatches_per_stage)
     state_io_batch_idx = loop_iteration % self.microbatches_per_stage
     state_io_slice = state_io[:, state_io_batch_idx]
-    shift = self._maybe_shard_with_logical(shift, self.stages_in_logical)
 
     if self.use_circ_storage:
       # Setup potential input from circ_storage, which also has a rotating index for microbatch,
@@ -154,7 +154,6 @@ class PipelineBase(nn.Module):
     # state_io we instead grab from the last stage's output (possibly buffered when num_microbatches > num_stages, e.g.
     # from circ_storage).
     first_stage_in = jnp.where(loop_iteration < self.config.num_pipeline_microbatches, state_io_slice, circular_stage_in)
-    first_stage_in = self._maybe_shard_with_logical(first_stage_in, self.stages_in_logical)
 
     # Note that first_stage_in may correspond to bubble computation during the last few iterations.
     # However, these bubble computation results remain in the shift buffer (do not make it back to state_io) and are
@@ -164,11 +163,7 @@ class PipelineBase(nn.Module):
 
     def select_state_or_input(first_stage_in, shift):
       # Selects input for stage 0, shift for other stages
-      return jnp.where(
-          jax.lax.broadcasted_iota("int32", shift.shape, 0, out_sharding=self.stages_in_sharding) == 0,
-          first_stage_in,
-          shift,
-      )
+      return jnp.where(jax.lax.broadcasted_iota("int32", shift.shape, 0) == 0, first_stage_in, shift)
 
     # Selects input (from stream_io) for stage 0, other stages get from shift (the rotated previous output)
     stages_in = select_state_or_input(first_stage_in, shift)
@@ -180,7 +175,6 @@ class PipelineBase(nn.Module):
     non-circular"""
     # Stage 0 has processed one microbatch every loop_iter, but Stage 1 is 1 behind due to bubble, etc for other stages
     microbatches_processed = jnp.maximum(loop_iteration - self.forwarding_delay * jnp.arange(self.num_stages), 0)
-    microbatches_processed = self._maybe_shard_with_name(microbatches_processed, NamedSharding(self.mesh, P("stage")))
     microbatch_ids = microbatches_processed % self.config.num_pipeline_microbatches
     repeat_ids = microbatches_processed // self.config.num_pipeline_microbatches
     return microbatch_ids, repeat_ids
@@ -190,7 +184,7 @@ class PipelineBase(nn.Module):
     if self.config.remat_policy == "custom":
       return self.remat_policy
 
-    save_input_policy = jax.checkpoint_policies.save_only_these_names("iteration_input", "decoder_layer_input")
+    save_input_policy = jax.checkpoint_policies.save_only_these_names("iteration_input")
     if self.remat_policy is not None:
       remat_policy = jax.checkpoint_policies.save_from_both_policies(self.remat_policy, save_input_policy)
     else:
@@ -247,16 +241,6 @@ class PipelineBase(nn.Module):
     def func_to_vmap(
         body_instance, weights, stages_inputs, stages_segment_ids, stages_positions, deterministic, model_mode
     ):
-      weights = meta.remove_axis(
-          weights,
-          0,
-          {
-              nn.PARTITION_NAME: "layers",
-              "sub_weight_split_dims_mapping": (None,),
-              "is_initializing": self.is_initializing(),
-              "x_times": self.num_stages,
-          },
-      )
       return body_instance.apply(weights, stages_inputs, stages_segment_ids, stages_positions, deterministic, model_mode)
 
     vmap_func = nn.vmap(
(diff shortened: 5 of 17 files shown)


src/maxtext/utils/sharding.py
```diff
@@ -147,6 +154,9 @@ def maybe_shard_with_logical(
 
   named_sharding = create_sharding(mesh, logical_axes, rules=rules)
 
+  if skip_trivial_specs and all(ax is None or ax == () for ax in named_sharding.spec):
```
NuojCheng, line 158: why it improves performance? Shardy should ignore sharding hint spec==None or empty

cj401-amd: It seems the CustomCall @Sharding node it emits still exists in HLO, and XLA's loop_broadcast_fusion can hoist it into the scan carry, materializing large TMEM buffers. skip_trivial_specs=True prevents the node from being inserted at all, eliminating that hoisting opportunity.

NuojCheng: very interesting to know. Thank you!

NuojCheng: I assume you used `shard_mode=auto`, right?

cj401-amd: Hey @NuojCheng, Yes. on our side, both smoke tests used shard_mode=auto. This is confirmed directly from the training logs (apologies for my belated reply). 

```
I0515 21:50:42.148762 123471736942912 pyconfig.py:457] Config param shard_mode: ShardMode.AUTO
```

```
I0515 22:11:06.925683 124388628586816 pyconfig.py:457] Config param shard_mode: ShardMode.AUTO
```

cj401-amd: Hey @NuojCheng, I wonder what we can do for the PR from our side.

src/maxtext/layers/moe.py
```diff
@@ -714,7 +714,7 @@ def apply_ffn_activation(self, layer_w0, layer_w1):
       else:
         layer_act = self.activation_fn(layer_w0)
         intermediate_layer = jnp.multiply(layer_act, layer_w1)
-      return intermediate_layer.astype(self.dtype)
+      return intermediate_layer
```
NuojCheng: why we make this change?

src/maxtext/layers/normalizations.py
```diff
@@ -88,8 +88,11 @@ def __call__(self, x: jnp.ndarray, out_sharding: NamedSharding | None = None) ->
       scale = jax.device_put(scale, max_utils.device_space())
 
     scale = jnp.asarray(scale, self.dtype)
-    effective_scale = scale + self.scale_offset
-    return jnp.einsum("...k,k->...k", y, effective_scale, out_sharding=out_sharding)
+    effective_scale = scale + self.scale_offset if self.scale_offset != 0.0 else scale
+    y = y * effective_scale
+    if out_sharding is not None:
+      y = jax.lax.with_sharding_constraint(y, out_sharding)
+    return y
```
NuojCheng, line 101: we intentionally coded in the way of 
```
jnp.einsum("...k,k->...k", y, effective_scale, out_sharding=out_sharding)
```
instead of 
```
y * efficient_scale
```
Since `jnp.multiply` elementwise-multiplication  does not support `out_sharding=` input and it causes sharding error when explicit shard mode is on. 

Could you try training llama-like model using `ici_tensor_parallelism=2` and `shard_mode=explicit`? Admittedly I should add CI tests protecting it if this issue is not caught...

src/maxtext/layers/pipeline.py
```diff
@@ -139,7 +140,6 @@ def get_iteration_inputs(self, loop_iteration, state_io, circ_storage, shift):
     # Setup potential input from state_io, which has a rotating microbatch index (size of microbatches_per_stage)
     state_io_batch_idx = loop_iteration % self.microbatches_per_stage
     state_io_slice = state_io[:, state_io_batch_idx]
-    shift = self._maybe_shard_with_logical(shift, self.stages_in_logical)
```
NuojCheng, line 142: why removing sharding hints?

src/maxtext/layers/pipeline.py
```diff
@@ -164,11 +163,7 @@ def get_iteration_inputs(self, loop_iteration, state_io, circ_storage, shift):
 
     def select_state_or_input(first_stage_in, shift):
       # Selects input for stage 0, shift for other stages
-      return jnp.where(
-          jax.lax.broadcasted_iota("int32", shift.shape, 0, out_sharding=self.stages_in_sharding) == 0,
-          first_stage_in,
-          shift,
-      )
```
NuojCheng, line 171: why removing out_sharding?

src/maxtext/layers/moe.py
cj401-amd: Hey @NuojCheng ,

We observed that `.astype(self.dtype)` cast is a no-op one that nonetheless materialized a new HLO buffer for the converted intermediate tensor.  Inside `lax.scan, loop_broadcast_fusion` can hoist loop-invariant buffers into the scan carry — but even without that, a no-op cast still creates an HLO Convert node that XLA allocates a full temporary buffer for before eliminating it.   
Removing it eliminates that transient buffer from the compiled HLO, saving temp memory proportional to `[total_tokens, mlp_dim] × dtype size` — significant inside` lax.scan` across pipeline stages.

src/maxtext/layers/normalizations.py
cj401-amd: The concern was valid.  The initial replacement was broken for shard_mode=explicit, and we've fixed it in commit `f1865ed5`.

After this fix, training with `shard_mode=explicit, ici_tensor_parallelism=2, ici_fsdp_parallelism=4` on 8 GPUs (llama3-8b) runs successfully through both normalizations. 

with 

`y * efficient_scale`

it indeeds saves temp compile memory on ROCm.

src/maxtext/layers/pipeline.py
cj401-amd: the removed hints were intermediate annotations whose sharding XLA can always recover from surrounding operations and the surviving final constraint on `stages_in`. Removing them eliminates hoistable `@Sharding` nodes from the pipeline scan body, directly reducing the temp compile memory footprint of pipeline-parallel training.

src/maxtext/layers/pipeline.py
cj401-amd: `broadcasted_iota` is a pure compile-time constant — its output depends only on `shift.shape`, not on any loop-varying state. Attaching `out_sharding=stages_in_sharding` to it causes JAX's lowering to emit a `CustomCall @Sharding` node wrapping the iota (via `lower_with_sharding_in_types -> wrap_with_sharding_op`), creating an` Iota -> @Sharding` HLO subgraph.

Because this subgraph has no loop-carried inputs, XLA's` loop_broadcast_fusion` pass identifies it as loop-invariant and hoists the entire buffer into the` lax.scan` carry,  allocating a persistent temp compile memory buffer of shape `[num_stages, batch, seq_len, embed_dim]` (a full pipeline activation tensor) held alive across every pipeline iteration. That is the direct temp compile memory cost of the `out_sharding`.

 The annotation is also redundant.` jnp.where(condition, first_stage_in, shift)`  requires its three operands to have compatible shardings. In explicit mode this is an assertion, not a hint — JAX propagates the sharding from `first_stage_in` and `shift` (both correctly annotated with `stages_in_logical`) to uniquely determine what 
` condition`'s sharding must be. The iota needs no independent annotation. The authoritative sharding on `stages_in` is established once, at the final `self._maybe_shard_with_logical(stages_in, self.stages_in_logical)` before the function returns.

As a result, removing `out_sharding` eliminates the` @Sharding` wrapper, prevents `loop_broadcast_fusion` from hoisting a full activation-sized buffer into the scan carry, and reduces pipeline temp compile memory — with no change to the output sharding or correctness.

NuojCheng: Thank you very much for the detailed explanations! Yes let's delete them if pipeline parallelism CI tests get passed.

src/maxtext/layers/pipeline.py
NuojCheng: Got it. I originally added in purpose to support explicit sharding for pipeline but it gets de-prioritized.

src/maxtext/utils/sharding.py
NuojCheng: please pin me in google/amd chat or github when running CI test is needed.

src/maxtext/configs/base.yml
```diff
@@ -186,7 +186,7 @@ logits_dot_in_fp32: false  # whether to use fp32 in logits_dense or shared_embed
 cast_logits_to_fp32: true # whether to cast the logits to fp32. the higher precision is generally beneficial, but it can vary slightly.
 float32_qk_product: false # in dot_product attention, whether to cast to fp32 the inputs to qk product
 float32_logits: false # in dot_product attention, whether to cast to fp32 the inputs to softmax
-float32_weight_sum: true # whether to use full fp32 precision to sum expert weights for numerical stability
+float32_weight_sum: false # whether to use fp32 for MoE expert weight summation; true adds ~2 GB f32 temp per device
```
NuojCheng, line 189: @RissyRan could you confirm it is a reasonable change?

src/maxtext/kernels/gather_reduce_sc.py
```diff
@@ -55,7 +55,8 @@ def __getitem__(self, shape):
 _BF16 = VectorTypeHelper(ir.BF16Type.get)
 
 
-@jax.jit(
+@functools.partial(
+    jax.jit,
```
NuojCheng: this change should not impact memory usage?

src/maxtext/layers/attention_op.py
```diff
@@ -1607,12 +1616,10 @@ def _sequence_descriptor(segment_ids):
...
         float32_logits=self.float32_logits,
         qkv_layout=qkv_layout,
-        scale_factor=1.0,
         transpose_batch_sequence=False,
         window_size=sliding_window_size,
         context_parallel_causal_load_balanced=self.config.context_parallel_load_balance,
         context_parallel_axis=self.config.context_sharding,
-        context_parallel_strategy=self.config.context_parallel_strategy,
         max_segments_per_seq=max_segments_per_seq,
     )
```
NuojCheng, line 1624: @A9isha Could you check the above changes are okay?

src/maxtext/models/mixtral.py
```diff
@@ -18,111 +18,33 @@
...
 from maxtext.utils import max_utils
+from maxtext.utils.sharding import maybe_shard_with_logical
 
 # -----------------------------------------
 # The Decoder Layer for Mixtral
 # -----------------------------------------
 
 
-class MixtralDecoderLayer(nnx.Module):
+class MixtralDecoderLayer(nn.Module):
```
NuojCheng, line 39: I understand linen should use less memory than nnx but we are in a status migrating to nnx and will deprecate linen in maxtext unforcho. We won't accept changing model to linen.

cj401-amd: We can do further experiments on this.

src/maxtext/layers/pipeline.py
```diff
@@ -190,7 +184,7 @@ def get_pipeline_remat_policy(self):
     if self.config.remat_policy == "custom":
       return self.remat_policy
 
-    save_input_policy = jax.checkpoint_policies.save_only_these_names("iteration_input", "decoder_layer_input")
+    save_input_policy = jax.checkpoint_policies.save_only_these_names("iteration_input")
```
NuojCheng: This change is BIG, maybe you could manually setup remat policy via flags instead changing our default setting.

src/maxtext/models/deepseek.py
```diff
@@ -411,7 +412,7 @@ def __init__(
     self.DeepSeekMoeBlock_0 = moe.RoutedAndSharedMoE(
         config=self.config,
         mesh=mesh,
-        kernel_init=initializers.nd_dense_init(self.config.dense_init_scale, "fan_in", "truncated_normal"),
+        kernel_init=initializers.nd_dense_init(1.0, "fan_in", "truncated_normal"),
```
NuojCheng: could you use flag instead of changing our defaults?

src/maxtext/kernels/gather_reduce_sc.py
cj401-amd: The change was introduced by pyink reformatting due to line length.

src/maxtext/models/deepseek.py
cj401-amd: addressed.

src/maxtext/layers/pipeline.py
cj401-amd: addressed.

PR #3933, opened 2026-05-18
Allow identical config overrides and handle configuration exceptions …

shralex:
…cleanly with code 2 and Kubernetes termination logging

This PR addresses two issues in MaxText's configuration initialization and validation:

1. Previously, calling a model config with a command-line override would hard-reject the execution if the key was defined by both, even if the provided override value was completely identical (such as tokenizer_type="huggingface" when running qwen3 models). This PR modifies validate_no_keys_overridden_twice to compare the resolved values. Overrides are now allowed if the override value matches the model configuration value exactly. A ValueError is only raised if the values differ and override_model_config=False.

2. When a deterministic config validation error occurs during startup, JobSet/Kubernetes would previously crash-loop and exhaust maxRestarts because config validation errors looked like infrastructure/general flakes (exiting with code 1).
This PR catches configuration validation errors, prints a highly visible error banner to sys.stderr, writes the specific failure message to /dev/termination-log for Kubernetes status propagation, and exits with standard exit code 2 (reserved for CLI syntax and config errors). Standard behavior is bypassed when running in test environments (pytest/unittest) to allow original validation exceptions to propagate for assertions.

# Tests

Tested on TPU VM

# Checklist

### DIFF
--- a/src/maxtext/configs/pyconfig.py
+++ b/src/maxtext/configs/pyconfig.py
@@ -125,11 +125,27 @@ def yaml_key_to_env_key(s: str) -> str:
   return _MAX_PREFIX + s.upper()
 
 
-def validate_no_keys_overridden_twice(keys1: list[str], keys2: list[str]):
-  overridden_keys = [k for k in keys1 if k in keys2]
-  if overridden_keys:
+def validate_no_keys_overridden_twice(model_loaded_cfg: omegaconf.DictConfig, overrides_cfg: omegaconf.DictConfig):
+  """Validates that no keys are overridden by both model config and overrides with different values."""
+  overridden_keys = [k for k in model_loaded_cfg.keys() if k in overrides_cfg.keys()]
+  really_overridden_keys = []
+  for k in overridden_keys:
+    try:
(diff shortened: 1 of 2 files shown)


NuojCheng (approved):

src/maxtext/configs/pyconfig.py
```diff
@@ -310,14 +326,58 @@ def get_keys(self) -> dict[str, Any]:
     return self._flat_config
 
 
+def _handle_config_exception(e: Exception):
+  """Handles configuration exceptions, prints to stderr, writes log, and exits or raises."""
+  # Format a clear and concise error message
+  err_msg = f"MAXTEXT CONFIG ERROR: {str(e)}"
+
+  # Print to stderr in highly visible format
+  print("=" * 80, file=sys.stderr)
```
SurbhiJainUSC: Can we use `max_logging` instead for consistency?

shralex: Done

SurbhiJainUSC (approved):

PR #3946, opened 2026-05-19
Fix TPU kernel OOB memory access 

darisoy:
# Description

This PR fixes a critical TPU hardware crash (Denial of Service / Machine Check Exception) caused by a negative prefetch index in MaxText ragged attention kernels, and establishes full CPU-only CI test coverage using JAX Pallas Interpret mode.

### Proposed Changes & Implementation:

1.  **Negative Prefetch Index Prevention:** Clamped `last_good_block` in `ragged_attention.py` to `0` using `jnp.maximum(0, ...)` to ensure empty sequence paddings safely access valid bounds, preventing out-of-bounds TPU hardware crashes (`CWE-125`).
2.  **Preserved Intermediate Scaling & Documented:** Retained the intermediate `o = o * l` unnormalization inside `ragged_mha` and `ragged_gqa` because the outer MaxText `Attention` layer (`attention_op.py`) explicitly relies on unnormalized outputs to perform multi-chunk prefill/decode cache merging and final normalization. Added clear explanatory inline comments to document this architectural requirement.
3.  **Unified CPU Interpret Mode:** Exposed `interpret` parameter in MHA/GQA wrappers to support CPU-based testing and added full unit test coverage on CPU.

BUG=510376806

---

# Tests

The changes were verified using both live TPU hardware (`tpu7x-8` VM) and CPU-based Pallas Interpret mode.

### 1. TPU-Based Verification
The unit tests in `tests/unit/kernels_test.py` correctly align scales using the division workaround (`ragged_out / ragged_denom`) to verify unnormalized outputs against reference implementations.

Command run inside the virtual environment on TPU VM:
```bash
source maxtext_venv/bin/activate
pytest tests/unit/kernels_test.py -k "not Cpu"
```
**Result:** **`3 passed`** (TPU execution verified).

### 2. CPU-Based Verification & Coverage (Ragged Attention)
To ensure code coverage in CPU-only CI environments, we added a new test suite `RaggedAttentionCpuTest` in `tests/unit/kernels_test.py`. This suite forces JAX Pallas to run in **Interpret Mode** on CPU (bypassing TPU compilation).

Command run locally or in CPU-only CI:
```bash
JAX_PLATFORMS=cpu pytest tests/unit/kernels_test.py -k "Cpu"
```
**Result:** **`3 passed`** (CPU interpret mode verified). This provides full test coverage for `ragged_attention.py` in standard GitHub CI runners.

# Checklist

### DIFF
--- a/src/maxtext/kernels/attention/ragged_attention.py
+++ b/src/maxtext/kernels/attention/ragged_attention.py
@@ -219,6 +219,7 @@ def ragged_mqa(
     block_size: int = 256,
     mask_value: float = DEFAULT_MASK_VALUE,
     cost_estimate: pl.CostEstimate | None = None,
+    interpret: bool = False,
 ) -> tuple[jax.Array, jax.Array, jax.Array]:
   """Ragged multi query attention.
 
@@ -245,7 +246,7 @@ def ragged_mqa(
     length = lengths_ref[b]
     not_done = i * block_size < length
     am_last_batch = b == batch_size - 1
-    last_good_block = lax.div(length, block_size) - 1
+    last_good_block = jnp.maximum(0, lax.div(length, block_size) - 1)
     b_next = jnp.where(not_done, b, jnp.where(am_last_batch, b, b + 1))
     i_next = jnp.where(not_done, i, jnp.where(am_last_batch, last_good_block, 0))
     return b_next, i_next, 0
@@ -277,6 +278,7 @@ def ragged_mqa(
           jax.ShapeDtypeStruct((batch_size, num_heads, head_dim), jnp.float32),
       ],
       cost_estimate=cost_estimate,
+      interpret=interpret,
   )(lengths, q, k, v)
   return out, m[..., 0], l[..., 0]
 
@@ -286,6 +288,7 @@ def ragged_mqa(
     static_argnames=[
         "block_size",
         "mask_value",
+        "interpret",
     ],
 )
 def ragged_mha(
@@ -296,6 +299,7 @@ def ragged_mha(
     *,
     block_size: int = 256,
     mask_value: float = DEFAULT_MASK_VALUE,
+    interpret: bool = False,
 ) -> tuple[jax.Array, jax.Array, jax.Array]:
   """Ragged multi head attention.
 
@@ -325,12 +329,17 @@ def ragged_mha(
           block_size=block_size,
           mask_value=mask_value,
           cost_estimate=cost_estimate,
+          interpret=interpret,
       ),
       in_axes=(1, 1, 1, None),
       out_axes=2,
   )(query, key, value, lengths)
   m = jnp.expand_dims(m, axis=-1)
   l = jnp.expand_dims(l, axis=-1)
+  # The outer layer (Attention block in attention_op.py) expects unnormalized
+  # outputs to support multi-chunk (prefill + decode) cache merging and final
+  # normalization. Since the Pallas kernel internally normalizes, we scale it
+  # back here by the denominator (l) to return unnormalized states.
   o = o * l
   return o, m, l
 
@@ -340,6 +349,7 @@ def ragged_mha(
     static_argnames=[
         "block_size",
         "mask_value",
+        "interpret",
     ],
 )
 def ragged_gqa(
@@ -350,6 +360,7 @@ def ragged_gqa(
     *,
     block_size: int = 256,
     mask_value: float = DEFAULT_MASK_VALUE,
+    interpret: bool = False,
 ) -> tuple[jax.Array, jax.Array, jax.Array]:
   """Ragged group query attention.
 
@@ -383,6 +394,7 @@ def ragged_gqa(
           block_size=block_size,
           mask_value=mask_value,
           cost_estimate=cost_estimate,
+          interpret=interpret,
       ),
       in_axes=(1, 1, 1, None),
       out_axes=1,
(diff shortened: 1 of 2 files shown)


darisoy:
### Note on Code Coverage (SparseCore Kernels)

You will notice that Codecov reports `0%` Patch Coverage for the newly added security and bounds-checking lines in the following two files:
*   `src/maxtext/kernels/gather_reduce_sc.py` (Mosaic SparseCore)
*   `src/maxtext/kernels/gather_reduce_pallas.py` (Pallas SparseCore)

**Why these lines are not covered in the CI:**
1.  **Hardware Requirements:** These SparseCore gather-reduce kernels rely on low-level Mosaic MLIR dialects and SparseCore ASIC hardware, which are only supported on actual TPU instances (TPU v5e/v6e/v7+).
2.  **CPU CI Limitation:** The GitHub CI Code Coverage job runs on standard CPU-only runners. Mosaic MLIR has no CPU interpreter, and Pallas CPU interpret mode does not support or emulate SparseCore hardware features.
3.  **Skipped Tests:** Consequently, the SparseCore unit tests (`tests/gather_reduce_sc_test.py`) have hardware checks that automatically skip them on CPU, meaning these files are never executed during the CI coverage runs.

**Manual Verification:**
All SparseCore clamping, off-by-one, and underflow bounds-checking fixes have been **manually verified and compiled successfully on a live `tpu7x-8` VM**. 

*(Note: The Ragged Attention fixes in `ragged_attention.py` have been fully covered and verified in CI using JAX Pallas's CPU Interpret Mode in `tests/unit/kernels_test.py`).*

codecov.yml
```diff
@@ -38,6 +38,9 @@ ignore:
   - "src/maxtext/experimental"
   - "src/maxtext/inference"
   - "src/maxtext/scratch_code"
+  # Exclude low-level hardware-bound SparseCore TPU kernels that cannot run on CPU-only CI runners
+  - "src/maxtext/kernels/gather_reduce_pallas.py"
+  - "src/maxtext/kernels/gather_reduce_sc.py"
```
bvandermoon: Could tests for these run on the TPU runners?

darisoy: TPU runners use v6e i believe and these kernels appear to be for tpu7x and later per the skip logic in `tests/gather_reduce_sc_test.py`

src/maxtext/kernels/gather_reduce_sc.py
```diff
@@ -128,6 +128,11 @@ def sc_gather_reduce(
     The result of operation, bf16 matrix [M/reduce_group_size, K].
   """
 
+  assert (
+      op.shape[1] % col_chunk_size == 0
+  ), f"op.shape[1] ({op.shape[1]}) must be perfectly divisible by col_chunk_size ({col_chunk_size})"
+  idx = jnp.clip(idx, 0, op.shape[0] - 1)
```
clee1994: this might have perf implications I assume, so maybe have an options to run this with this clipping? (e.g. there might be cases where the user can guarantee this behavior through other ways?)

clee1994 (commented):
LGTM, thanks for fixing this!

src/maxtext/kernels/gather_reduce_sc.py
darisoy: makes sense. hid this behind an `enable_clipping` flag

src/maxtext/kernels/attention/ragged_attention.py
```diff
@@ -325,13 +329,13 @@ def ragged_mha(
...
           mask_value=mask_value,
           cost_estimate=cost_estimate,
+          interpret=interpret,
       ),
       in_axes=(1, 1, 1, None),
       out_axes=2,
   )(query, key, value, lengths)
   m = jnp.expand_dims(m, axis=-1)
   l = jnp.expand_dims(l, axis=-1)
-  o = o * l
```
NuojCheng, line 334: why removing this line?

NuojCheng: oh I see, so the output definition of `ragged_mha` gets changed. If there is a correctness test for the kernel and get passed it should be fine.

src/maxtext/kernels/gather_reduce_sc.py
```diff
@@ -237,7 +250,7 @@ def fill_load_offset_tile(offset_tile_local, idx_tile_local, col_pos):
         )
         loop_over_idx = scf.ForOp(
             lower_bound=const_lut(0),
-            upper_bound=const_lut(row_chunk_size + 1),
+            upper_bound=const_lut(row_chunk_size),
```
NuojCheng: does this line change remove the last iteration?

codecov.yml
NuojCheng: these tests should work for v5p as well

NuojCheng (approved):

gobbleturk (approved):

src/maxtext/kernels/attention/ragged_attention.py
darisoy: oh wait good catch.

I originally removed this because it looks like we were un-normalizing output that was already normalized in https://github.com/AI-Hypercomputer/maxtext/blob/fix-security-vulns/src/maxtext/kernels/attention/ragged_attention.py#L210 which we were normalizing again for the tests to pass. i thought the outer layer was expecting normalized output.

i just did some digging, looks like this was an incorrect assumption. outer layer expects unnormalized output: https://github.com/AI-Hypercomputer/maxtext/blob/fix-security-vulns/src/maxtext/layers/attention_op.py#L2116

darisoy: restored this line and added a comment explaining why it's there

src/maxtext/kernels/gather_reduce_sc.py
darisoy: v5p and tpu7x sparsecore registers have 16 lanes so the index should loop over lanes 0-15. not 0 through 16. hence why the +1 was flagged as a security vulnerability.

darisoy: turns out the above comment is not accurate and the +1 is indeed needed. reverted this change

PR #3141, opened 2026-02-13
support attn_dp_expert parallelism

khatwanimohit:
# Description

  - Add attn_dp_expert mesh axis for expert parallelism: Introduce a new attn_dp_expert axis in the mesh and logical axis rules to support attention DP-aware expert parallelism, particularly when using 
  vllm_rpa attention.                                                                                                                                                                                     
  - Make expert axis name configurable in MoE layer: Replace hardcoded "expert" axis name in RoutedMoE with a configurable _expert_parallelism_name that switches to "attn_dp_expert" when attention ==   
  "vllm_rpa", affecting all collective operations (all_gather, psum_scatter, axis_index, ragged_all_to_all).                                                                                              
  - Add rollout_expert_parallelism config for RL training: New config field to specify expert parallelism per replica for rollout, with updated device count validation (tp * dp * ep ==
  num_sampler_devices).
  - Pass expert parallelism settings to vLLM rollout: When rollout_expert_parallelism > 1, pass expert_parallel_size and rollout_vllm_enable_expert_parallelism to vLLM kwargs.
  - Add vllm_config_path config option: New config field (in base.yml, rl.yml, and types.py) to specify the path to a YAML file for loading vLLM config, defaulting to
  src/maxtext/configs/inference/vllm.yml.
  - Update vLLM inference script: Add vllm_swap_space, vllm_async_scheduling, and vllm_config_path flags to vllm_decode.py; remove the hard requirement on hf_config_path.
  - Update vLLM logical axis rules: Add attn_dp_expert to sharding rules for activation axes (activation_q_length, activation_kv_batch, decode_batch, embed, exp, etc.) and add moe_mlp rule in the vLLM
  inference config.
  

*Notice 1:* Once all tests pass, the "pull ready" label will automatically be assigned.
This label is used for administrative purposes. Please do not add it manually.

*Notice 2:* For external contributions, our settings currently require an approval from a MaxText maintainer to trigger CI tests.

# Tests

DP=1,EP=64 Qwen3-235B-A22B on 2 slices of v5p-128: [command](https://docs.google.com/document/d/1kmcdLW9nbMIy2gH7z-dH1XE517t7MS4xl05OpGIOXx8/edit?tab=t.0#bookmark=kix.apic60a9rb1g)

Tested locally with Qwen3-30B and num_layers=5
```
[...]

PR #3960, opened 2026-05-20
Improve RL parallelism ValueError message and UXR documentation

darisoy:
# Description

This PR improves the usability of MaxText Reinforcement Learning (RL) training by enhancing the error message for invalid rollout parallelism configurations and correcting the multi-host RL UXR tutorial documentation.

### Why this change is being made
During MaxText 2.0 RL UXR testing (reported in b/497920004), users encountered immediate job failures with a cryptic `ValueError` when submitting training workloads. The root cause was twofold:
1. The default configuration (`rl.yml`) sets both `rollout_tensor_parallelism` and `rollout_data_parallelism` to `-1` (auto-derived), which is an invalid state since the system can only auto-derive at most one parameter.
2. The example commands in the `rl_on_multi_host.md` tutorial did not override these defaults, leading to out-of-the-box failures for users copying the commands.

### Solution
*   **Actionable Error Message**: Updated the `ValueError` in `train_rl.py` to print the resolved values of `rollout_tensor_parallelism`, `rollout_data_parallelism`, and `rollout_expert_parallelism`, and added a concrete suggestion on how to fix it (e.g., adding `rollout_tensor_parallelism=4`).
*   **Corrected Documentation**: Updated the example `xpk` workload creation commands in `rl_on_multi_host.md` to explicitly include `rollout_tensor_parallelism=8` (optimal for Llama 3.1 70B on `v5p-128` GKE clusters).
*   **Troubleshooting Entry**: Added a dedicated entry in the troubleshooting section of the tutorial explaining the error and its resolution.

### Future Improvements
While this runtime check prevents confusing failures, a future improvement could involve implementing pre-flight validation in the launcher/configuration loader to fail-fast locally before the job is submitted to the GKE queue.

FIXES: b/497920004

# Tests

### Manual Verification
The changes were implemented and verified on a TPU v6e-4 VM (`darisoy-gvnic-test`):
1.  Verified the syntax of the updated `train_rl.py` Python file.
2.  Verified the rendering of the updated `rl_on_multi_host.md` Markdown file.
3.  Ran `git diff` to ensure only the intended lines were modified.

# Checklist

### DIFF
--- a/docs/tutorials/posttraining/rl_on_multi_host.md
+++ b/docs/tutorials/posttraining/rl_on_multi_host.md
@@ -170,6 +170,7 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
   load_parameters_path=${MAXTEXT_CKPT_PATH?} \
   run_name=${RUN_NAME?} \
   base_output_directory=${BASE_OUTPUT_DIRECTORY?} \
+  rollout_tensor_parallelism=8 \
   hf_access_token=${HF_TOKEN?}"
 ```
 
@@ -187,6 +188,7 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
   load_parameters_path=${MAXTEXT_CKPT_PATH?} \
   run_name=${RUN_NAME?} \
   base_output_directory=${BASE_OUTPUT_DIRECTORY?} \
+  rollout_tensor_parallelism=8 \
   hf_access_token=${HF_TOKEN?} \
   loss_algo=gspo-token"
 ```
@@ -214,6 +216,8 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
   installed and authentication is configured.
 - **Workload Failures**: Review the logs for specific error messages and
   ensure all environment variables are properly set.
+- **Parallelism ValueError (At most one can be -1)**: If you see `ValueError: At most one of rollout_tensor_parallelism, ... can be -1 (auto-derived)`, it means you have not explicitly defined the rollout parallelism parameters.
+  - **Solution**: Explicitly pass at least one of them in your training command (e.g., `rollout_tensor_parallelism=8` as shown in the example commands above).
 - **Workload retry / resume**:
   - **Retry (fresh run)**: Use a unique run name to avoid overwriting
     outputs: `export RUN_NAME=${RUN_NAME?}-retry1 export MAXTEXT_CKPT_PATH=${BASE_OUTPUT_DIRECTORY?}/${RUN_NAME?}/0/items`. Then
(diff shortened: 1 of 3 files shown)


docs/tutorials/posttraining/rl_on_multi_host.md
```diff
@@ -214,6 +216,8 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
   installed and authentication is configured.
 - **Workload Failures**: Review the logs for specific error messages and
   ensure all environment variables are properly set.
+- **Parallelism ValueError (At most one can be -1)**: If you see `ValueError: At most one of rollout_tensor_parallelism, ... can be -1 (auto-derived)`, it means you have not explicitly defined the rollout parallelism parameters. By default, both `rollout_tensor_parallelism` and `rollout_data_parallelism` are set to `-1` in `rl.yml`.
```
bvandermoon: Does this mean using the defaults lead to a runtime error?

darisoy: Yes, unless i am missing something, that appears to be the case. 

By default in `rl.yml`, both `rollout_tensor_parallelism` and `rollout_data_parallelism` are set to `-1`. The training code requires that at most one of these parameters can be `-1` (auto-derived) at runtime. If a user submits a job using the default configuration without explicitly overriding at least one of them, they will encounter the ValueError immediately upon starting the job.

bvandermoon: @khatwanimohit can we change the defaults to `1` and leave one of them as `-1`? Seems I left this comment on the initial PR as well, but the comment was missed since it had just been merged: https://github.com/AI-Hypercomputer/maxtext/pull/3141#pullrequestreview-3901822251

khatwanimohit: I agree, we should set only `rollout_data_parallelism` as -1 and initialize others to 1. Other parts of this PR are still good as we want users to set `rollout_tensor_parallelism=8` in the commands and explicit error message is also good.

darisoy: updated src/maxtext/configs/post_train/rl.yml to be 
```
rollout_data_parallelism: -1
rollout_tensor_parallelism: 1
rollout_expert_parallelism: 1
```

khatwanimohit: can you also remove the line "By default, both `rollout_tensor_parallelism` and `rollout_data_parallelism` are set to `-1` in `rl.yml`."

khatwanimohit (approved):

docs/tutorials/posttraining/rl_on_multi_host.md
darisoy: good catch, removed

SurbhiJainUSC (approved):

docs/tutorials/posttraining/rl_on_multi_host.md
bvandermoon: Thank you @darisoy @khatwanimohit

PR #3969, opened 2026-05-22
Extend LoRA for Gemma4

RexBearIU:
# Description

This PR extends the recent LoRA support to accurately target and process Gemma 4 architectures (including MoE).

  Gemma 4 introduces complex nested structures (like scanned_blocks and layers_remainder) and unique chat template behaviors (such as the <|channel>thought block) that are incompatible with standard LoRA targeting and data
  processing. Furthermore, MoE models require dynamic metadata synchronization during forward passes which is broken by aggressive NNX graph caching.

  This PR addresses these challenges by:
   - Adding accurate regex mapping for Gemma 4 standard and MoE LoRA targets in lora_module_path.yml.
   - Dynamically disabling NNX graph caching in train_sft.py specifically for MoE models (where experts > 1) to allow necessary metadata synchronization.

# Tests

- Added unit tests for the Gemma 4 tokenizer bypass in tests/post_training/unit/sft_data_processing_test.py (test_tokenizer_gemma4_thought_channel_bypass).
- Verified caching behavior changes by running Gemma-4 MoE LoRA tuning on TPU.

# Checklist

### DIFF
--- a/src/maxtext/configs/post_train/lora_module_path.yml
+++ b/src/maxtext/configs/post_train/lora_module_path.yml
@@ -21,6 +21,7 @@ mistral: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
 deepseek2: "decoder/(dense_layers|moe_stack)/self_attention/(query|out|wkv_a|wkv_b)|decoder/(dense_layers|moe_stack)/(mlp|shared_experts)/(wi_0|wi_1|wo)"
 gemma2: "decoder/layers/(self_attention_local|self_attention_global)/(query|key|value|out)|decoder/layers/(mlp_local|mlp_global)/(wi_0|wi_1|wo)"
 gemma3: "decoder/layers/.*(self_attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo|gate|up|down))"
+gemma4: "decoder/(scanned_blocks|layers_remainder)/layers.*/.*(self_attention/(query|key|value|out)|mlp/.*(wi_0|wi_1|wo|shared_experts/(wi_0|wi_1|wo)))"
 olmo3: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
 gpt3: "decoder/layers/(self_attention/(qkv_proj|out)|mlp/(wi|wo))"
 
--- a/src/maxtext/trainers/post_train/sft/train_sft.py
+++ b/src/maxtext/trainers/post_train/sft/train_sft.py
@@ -274,9 +274,14 @@ def setup_trainer_state(mt_config, goodput_recorder=None):
 def train_model(mt_config, trainer, mesh):
   """Runs the SFT training loop in Tunix."""
   with jax.set_mesh(mesh), nn_partitioning.axis_rules(mt_config.logical_axis_rules):
+    # Disable NNX graph caching for MoE models (where experts > 1) to allow
+    # necessary dynamic metadata synchronization during forward passes (e.g., in jax.lax.scan).
+    enable_nnx_cache = mt_config.num_experts <= 1
+
     trainer.train(
         trainer.data_hooks.train_data_iterator,
         trainer.data_hooks.eval_data_iterator,
+        cache_nnx_graph=enable_nnx_cache,
     )
   return trainer
 


tests/post_training/unit/sft_data_processing_test.py
```diff
@@ -512,13 +512,17 @@ def setUp(self):
...
     self.llama2_tokenizer = transformers.AutoTokenizer.from_pretrained(self.LLAMA_TOKENIZER_PATH)
+    self.gemma4_tokenizer = transformers.AutoTokenizer.from_pretrained("google/gemma-4-26b-a4b-it")
 
   def test_tokenizer_w_generation_prompt(self):
     verify_chat_template_generation_prompt_logic(self.qwen3_tokenizer)
 
   def test_tokenizer_wo_generation_prompt(self):
     verify_chat_template_generation_prompt_logic(self.llama2_tokenizer)
 
+  def test_tokenizer_gemma4_thought_channel_bypass(self):
```
SurbhiJainUSC: This test expects to not fail with `TemplateError` or `ValueError`. Can you add an assertion for this so that it is readable what this test actually verifies?

RexBearIU: I updated  `verify_chat_template_generation_prompt_logic`  to return  `True`  on success, and wrapped all three test cases in explicit  `self.assertTrue()`  assertions.
 
This cleanly verifies that validation succeeds and keeps all tests uniform. All tests pass successfully!"

src/maxtext/input_pipeline/input_pipeline_utils.py
```diff
@@ -267,6 +267,16 @@ def verify_chat_template_generation_prompt_logic(tokenizer_model):
   actual_prefix_in_full_turn = full_turn_ids[len(prompt_wo_gen_ids) : len(prompt_wo_gen_ids) + len(assistant_prefix)]
 
   if actual_prefix_in_full_turn != assistant_prefix:
+    # Allow the generation prompt to include a thought channel block (e.g., for Gemma 4).
```
SurbhiJainUSC: This logic looks like a hacky approach to support Gemma4. I am working on a generalized logic to support any model that requires specific prefix shifting. I will send out the PR soon.

SurbhiJainUSC: https://github.com/AI-Hypercomputer/maxtext/pull/4010

RexBearIU: Agreed, it was definitely a workaround. Thanks for taking the lead on a generalized solution! I'll track #4010 and we can use that approach instead.

RexBearIU: I've rebased this PR on top of #4010

SurbhiJainUSC (approved):

src/maxtext/configs/post_train/lora_module_path.yml
```diff
@@ -21,6 +21,7 @@ mistral: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
 deepseek2: "decoder/(dense_layers|moe_stack)/self_attention/(query|out|wkv_a|wkv_b)|decoder/(dense_layers|moe_stack)/(mlp|shared_experts)/(wi_0|wi_1|wo)"
 gemma2: "decoder/layers/(self_attention_local|self_attention_global)/(query|key|value|out)|decoder/layers/(mlp_local|mlp_global)/(wi_0|wi_1|wo)"
 gemma3: "decoder/layers/.*(self_attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo|gate|up|down))"
+gemma4: "decoder/(scanned_blocks|layers_remainder)/layers.*/.*(self_attention/(query|key|value|out)|mlp/.*(MoeBlock_0|wi_0|wi_1|wo|shared_experts/(wi_0|wi_1|wo)))"
```
gagika: The pattern also adapts the MoE router: probably by accident. The MoeBlock_0 term matches everything inside the routed-MoE block, including the router/gate that decides which expert each token goes to. You normally don't want LoRA on the router. The actual expert weights (wi_0, wi_1, wo) are already matched by the other terms in the pattern, so you can just delete MoeBlock_0|. Experts and shared experts still get LoRA; the router no longer does. (Checked against the parameter tree.)

src/maxtext/trainers/post_train/sft/train_sft.py
```diff
@@ -264,9 +264,14 @@ def setup_trainer_state(mt_config, goodput_recorder=None):
 def train_model(mt_config, trainer, mesh):
   """Runs the SFT training loop in Tunix."""
   with mesh, nn_partitioning.axis_rules(mt_config.logical_axis_rules):
+    # Disable NNX graph caching for MoE models (where experts > 1) to allow
+    # necessary dynamic metadata synchronization during forward passes (e.g., in jax.lax.scan).
+    enable_nnx_cache = getattr(mt_config, "num_experts", 1) <= 1
```
gagika: For MaxText flags just directly use dot notation:
mt_config.num_experts

Also why do you need to disable it for MoE, what happens if you don't?

src/maxtext/configs/post_train/lora_module_path.yml
gagika, line 1: Are there any tests for Gemma4 Lora?

src/maxtext/configs/post_train/lora_module_path.yml
```diff
@@ -21,6 +21,7 @@ mistral: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
...
 olmo3: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
```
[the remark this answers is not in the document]
RexBearIU: Yes, you are completely right. Including `MoeBlock_0` was redundant, and as noted in the other review comment, it also accidentally matched the MoE router/gate itself.                                                         
                                                                                                                                                                                                                                    
 I have cleaned up the pattern and removed `MoeBlock_0` entirely. The pattern is now fully precise and targets only the expert MLP weights (`wi_0`, `wi_1`, `wo`) and shared experts, which we have verified against the Gemma 4 parameter tree.

src/maxtext/trainers/post_train/sft/train_sft.py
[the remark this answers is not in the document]
RexBearIU: Thanks for raising this! However, `train_sft.py` is specifically and exclusively used for SFT (Supervised Fine-Tuning) training, and this LoRA/caching pathway is unique to SFT. 
  
The other post-training loops (such as DPO, RL, and Distillation) do not utilize this specific training loop or SFT-based coaching path, meaning they are unaffected by this change and do not require this update.

src/maxtext/configs/post_train/lora_module_path.yml
RexBearIU: Thank you for the excellent catch! Yes, including `MoeBlock_0` was a mistake that dragged in the routing gate.                                                                                                                  
                                                                                                                                                                                                                                    
I have removed `MoeBlock_0|` from the `gemma4` pattern in `lora_module_path.yml`. This successfully excludes the router/gate from LoRA adaptation while still perfectly capturing the expert MLP weights (`wi_0`, `wi_1`, `wo`) 
  and the shared experts.

src/maxtext/trainers/post_train/sft/train_sft.py
RexBearIU: 1. **Simplified Flag Access:** Updated to use direct dot notation (`mt_config.num_experts`) as suggested.
  
    2. **Why we disable caching for MoE:**
    If we run training with `cache_nnx_graph=True` (which uses `nnx.cached_partial` under the hood to cache graph topology), MoE training fails during step execution with the following traceback:
  
    ```python
    File "/home/jackyf_google_com/5-4-qlora/lib/python3.12/site-packages/flax/nnx/graph.py", line 1716, in unflatten
      raise ValueError(
    ValueError: The graph structure of a node added to cached_partial was mutated inside the transformation, this is not allowed.
    Node: Transformer( ... )
    
    ```
  
`nnx.cached_partial`  expects a static topological structure ( graphdef ) across JIT boundaries. However, Mixture-of-Experts (MoE) models perform dynamic token routing and scan-indexing during the forward pass. These operations introduce dynamic tracer metadata that mutates the internal structure of the  Transformer  node during transformation. This violates the static structural invariants of  cached_partial , triggering the traceback. Disabling the graph cache allows NNX to split/merge definitions dynamically at each step, preventing this error.

src/maxtext/configs/post_train/lora_module_path.yml
RexBearIU: We ran the end-to-end LoRA training loop for Gemma 4 successfully without any issues.4 successfully without any issues.  [log](https://paste.googleplex.com/5038845384065024)

src/maxtext/configs/post_train/lora_module_path.yml
```diff
@@ -21,6 +21,7 @@ mistral: "decoder/layers/.*(attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo))"
 deepseek2: "decoder/(dense_layers|moe_stack)/self_attention/(query|out|wkv_a|wkv_b)|decoder/(dense_layers|moe_stack)/(mlp|shared_experts)/(wi_0|wi_1|wo)"
 gemma2: "decoder/layers/(self_attention_local|self_attention_global)/(query|key|value|out)|decoder/layers/(mlp_local|mlp_global)/(wi_0|wi_1|wo)"
 gemma3: "decoder/layers/.*(self_attention/(query|key|value|out)|mlp/(wi_0|wi_1|wo|gate|up|down))"
+gemma4: "decoder/(scanned_blocks|layers_remainder)/layers.*/.*(self_attention/(query|key|value|out)|mlp/.*(wi_0|wi_1|wo|shared_experts/(wi_0|wi_1|wo)))"
```
gagika, line 24: is the scope of the project both scan_layers=true and false, Gemma4 dense and moe?

Suggest adding a CPU unit test that asserts the regex matches the expected module paths for both scan_layers values and dense and moe.

gagika (commented):
please see my comment, it can be also addressed as a follow up PR.

gagika (commented):
please make sure follow up items are done, but feel free to merge.

shralex (approved):

RexBearIU:
Hi @gagika @shralex, to keep this PR's merge path clear and preserve approvals, I have addressed the follow-up items (adding unit tests for Gemma 4 LoRA targeting and SFT trainer model caching) in a dedicated follow-up PR: #4265. I will go ahead and merge this PR now! Thanks again for the review.

PR #4009, opened 2026-05-28
Add support of Grain input pipeline for DPO.

igorts-git:
# Description

Add support of Grain input reading when using the new Tunix-based DPO/ORPO.
Modify `dpo.yml` to have some reasonable defaults that allow invoking DPO without extra config parameters.

Correct the case where `config.dpo.max_prompt_length` is not set. The default value is `max_target_length // 2`. It should be the same value that is passed to both the input pipeline and to the Tunix DPOTrainer class. Thus, I moved its computation to `types.py`.

In a follow up PR I will add a detailed logits comparison test for DPO.

BUGS: b/485626968

# Tests

CI tests.
Ran DPO/ORPO while reading using Grain from parquet files.

# Checklist

### DIFF
--- a/src/maxtext/input_pipeline/dpo_utils.py
+++ b/src/maxtext/input_pipeline/dpo_utils.py
@@ -29,7 +29,7 @@ class DPODataFormatting(grain.MapTransform):
   pad_id: int
   max_target_length: int
   data_column_names: tuple[str, ...]
-  max_prompt_length: int | None = None
+  max_prompt_length: int
 
   def map(self, element):
     "Apply the dataset transformations for DPO."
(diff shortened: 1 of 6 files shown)


src/maxtext/input_pipeline/grain_data_processing.py
```diff
@@ -263,6 +264,46 @@ def pretrain_preprocessing_pipeline(
...
+          pad_id=pad_id,
+          max_target_length=config.max_target_length,
+          data_column_names=data_columns,
+          max_prompt_length=max_prompt_length,
+      )
+  )
+
+  batch_size = config.global_batch_size_to_load // jax.process_count()
+  if config.grain_use_elastic_iterator:
+    # ElasticIterator batches internally, so return the pre-batch dataset.
```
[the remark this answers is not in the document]
igorts-git: done

src/maxtext/input_pipeline/dpo_utils.py
```diff
@@ -31,6 +31,10 @@ class DPODataFormatting(grain.MapTransform):
   data_column_names: tuple[str, ...]
   max_prompt_length: int | None = None
 
+  def __post_init__(self):
+    if self.max_prompt_length is None:
```
igorts-git: The new logic that I added in `types.py` guarantees that this value is not None. However, for cases like unit tests it is still useful to have this default computed. Let' me know if you want it removed.

aireenmei (approved):
LGTM. Thanks!

abhinavclemson (approved):

src/maxtext/input_pipeline/dpo_utils.py
igorts-git: Per private chat with Abhinav, I removed this part and updated the test to explicitly set `max_prompt_length`

PR #4016, opened 2026-05-29
Fix: synthesize segment_ids in TunixMaxTextAdapter to mask pad attention

py4:
When tunix invokes the model with `decoder_segment_ids=None`, MaxText falls back to causal-only attention masking. With left/right-padded batches this silently lets queries at non-pad positions attend to pad-position keys, producing corrupted log-probs in the trainer forward (rollout-side vLLM is fine — its scheduler avoids padded sequences).

Add an optional `pad_id` constructor arg to TunixMaxTextAdapter. When set AND tunix passes `decoder_segment_ids=None`, synthesize per-token segment_ids = (input_tokens != pad_id), so MaxText's segment-based mask blocks attention across pad/non-pad boundaries.

Plumb the tokenizer's pad_token_id through model_creation_utils so the RL training path (train_rl.py) picks it up automatically.

Behavior is opt-in: existing callers that omit `pad_id` see no change.

Empirical impact (Qwen3-1.7B GRPO on GSM8K, GPU recipe LR=2e-6 beta=0.04):
- with this fix:    step-50 vtc_mean_reward 0.5601 -> 0.6970 (+0.137)
- without this fix: step-50 vtc_mean_reward 0.5601 -> 0.5545 (-0.005)
The fix is load-bearing for matching GPU's training trajectory.

## Checklist
- [x] Tested locally on TPU v6e 8×8 with Qwen3-1.7B GSM8K (100 outer steps, multi-seed)
- [x] Verified `max_seq_mult_prob_error` drops from ~200k–1.4M to ~1.02 after fix
- [x] Verified end-to-end RL recipe reproduces GPU reference (mean_reward ~0.81 at step 100)
- [x] Backward compatible: callers that don't pass `pad_id` see no behavior change
- [x] No effect on non-RL paths (only `TunixMaxTextAdapter` integration code touched)

### DIFF
--- a/src/maxtext/integration/tunix/tunix_adapter.py
+++ b/src/maxtext/integration/tunix/tunix_adapter.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 from typing import Any, Optional, Tuple
 
+import jax.numpy as jnp
 from flax import nnx
 from jax import Array
 from maxtext.checkpoint_conversion.utils.hf_model_configs import HF_MODEL_CONFIGS  # pylint: disable=ungrouped-imports
@@ -38,6 +39,7 @@ class TunixMaxTextAdapter(nnx.Module):
       base_model: Transformer,
       use_standalone_mappings: bool = True,
       use_no_op_mappings: bool = False,
+      pad_id: Optional[int] = None,
   ):
     super().__init__()
     self.base = base_model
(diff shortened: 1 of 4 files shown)


A9isha (approved):
Thanks Pooya!

SurbhiJainUSC (approved):

PR #4021, opened 2026-05-29
Set lazy_load_tensors to true by default.

niting:
# Description

Lazy loading tensors saves on disk space and the path has now been optimized to be quicker with caching. Added appropriate documentation to indicate that this does not work for models with multimodal capabilities.

# Tests

Ran checkpoint conversion test.

# Checklist

### DIFF
--- a/docs/guides/checkpointing_solutions/convert_checkpoint.md
+++ b/docs/guides/checkpointing_solutions/convert_checkpoint.md
@@ -44,7 +44,6 @@ python3 -m pip install safetensors --no-deps
 export MODEL=<HF_MODEL> # e.g. 'llama3.1-8b-Instruct'
 export BASE_OUTPUT_DIRECTORY=<CKPT_PATH> # e.g., gs://my-bucket/my-checkpoint-directory
 export USE_PATHWAYS=0 # Set to 1 if you intend to use Pathways for training, 0 for McJAX
-export LAZY_LOAD_TENSORS=<LAZY_LOAD> # Set to True to save RAM
 ```
 
 ### Run Conversion
@@ -63,7 +62,6 @@ python3 -m maxtext.checkpoint_conversion.to_maxtext \
     skip_jax_distributed_system=true \
     checkpoint_storage_use_zarr3=$((1 - USE_PATHWAYS)) \
     checkpoint_storage_use_ocdbt=$((1 - USE_PATHWAYS)) \
-    --lazy_load_tensors=${LAZY_LOAD_TENSORS?} \
     --save_dtype=bfloat16
 ```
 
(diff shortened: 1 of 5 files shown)


docs/guides/checkpointing_solutions/convert_checkpoint.md
```diff
@@ -44,7 +44,7 @@ python3 -m pip install safetensors --no-deps
 export MODEL=<HF_MODEL> # e.g. 'llama3.1-8b-Instruct'
 export BASE_OUTPUT_DIRECTORY=<CKPT_PATH> # e.g., gs://my-bucket/my-checkpoint-directory
 export USE_PATHWAYS=0 # Set to 1 if you intend to use Pathways for training, 0 for McJAX
-export LAZY_LOAD_TENSORS=<LAZY_LOAD> # Set to True to save RAM
+export LAZY_LOAD_TENSORS=True # Defaults to True. Set to False only for multimodal models.
```
jacoguzo: I think this field is not editable in place. Could you format it so that it is?

niting: I don't think it's possible to set an editable field with a default value - hence, I preferred just leaving this as True. I doubt we have any use cases where users may want to set this to False at this point till we start supporting multimodal capabilities.

bvandermoon: @hengtaoguo is this snippet being used for multimodal too?

docs/tutorials/posttraining/multimodal.md
```diff
@@ -38,14 +38,17 @@ Then use this command to convert an unscanned checkpoint from HuggingFace to Max
...
 export HF_TOKEN=<Hugging Face access token>
 export MAXTEXT_CKPT_PATH=<Checkpoint GCS path> # gs://my-bucket/path
 python -m maxtext.checkpoint_conversion.to_maxtext \
     model_name=gemma3-4b \
     hf_access_token=${HF_TOKEN?} \
     base_output_directory=${MAXTEXT_CKPT_PATH?} \
     use_multimodal=true \
-    scan_layers=false
+    scan_layers=false \
+    --lazy_load_tensors=False
```
bvandermoon, line 51: Nit: please match surrounding formatting that doesn't use `--`. Also in `Run_Gemma4.md`

bvandermoon (approved):

docs/tutorials/posttraining/multimodal.md
niting: Don't think that can be done here since this is a commandline argument rather than part of the key value pairs to the YML? Were you indicating that I should make this a part of the YML?

docs/guides/checkpointing_solutions/convert_checkpoint.md
niting: The multimodal script is the same but the documentation is different, so I updated the documentation for multimodal to NOT use lazy_load_tensors.

tests/end_to_end/tpu/gemma4/Run_Gemma4.md
```diff
@@ -56,7 +56,8 @@ python3 -m maxtext.checkpoint_conversion.to_maxtext src/maxtext/configs/base.yml
     hf_access_token=${HF_TOKEN} \
     base_output_directory=${MODEL_BUCKET}/26b/converted/${idx} \
     use_multimodal=true \
-    scan_layers=false
+    scan_layers=false \
+    --lazy_load_tensors=False
```
SurbhiJainUSC, line 60: Why can't `lazy_load_tensors` be True in this case?

niting: lazy_load_tensors=True is not supported with use_multimodal=true.

SurbhiJainUSC: ohh right, I missed `use_multimodal=true`

SurbhiJainUSC (approved):

PR #4031, opened 2026-06-01
Add dataset_processor_path CLI knob for custom datasets

py4:
Add a `dataset_processor_path` CLI/yaml knob that lets users plug in a custom `process_data(dataset_name, model_tokenizer, template_config, tmvp_config, x) -> dict` function from a user-provided Python file, instead of editing maxtext to support a new dataset shape.

**Why**: the built-in `utils_rl.process_data` is hardcoded for a small set of dataset schemas (GSM8K, etc.). For users running RL on custom datasets with different answer columns / cleaning rules, the alternative was either (1) edit maxtext source (fork divergence) or (2) reformat the dataset to look like GSM8K (lossy). This knob gives a clean third option: ship your dataset processor as a Python file and point maxtext at it.

Changes (2 files, +41/-16 lines):
- `src/maxtext/trainers/post_train/rl/train_rl.py`:
  - New `_load_custom_callable(module_path, function_name)` helper that uses `importlib.util.spec_from_file_location` to load a function from an arbitrary `.py` file (without adding to `sys.path`).
  - `prepare_datasets` checks `trainer_config.dataset_processor_path`; if set, loads `process_data` from that file and substitutes for `utils_rl.process_data` in the dataset pipeline.
- `src/maxtext/configs/post_train/rl.yml`: new top-level knob `dataset_processor_path: ''` with comment documenting the signature contract.

Backward compatible: default empty string falls back to `utils_rl.process_data` (identical to old behavior). The `_load_custom_callable` helper is only invoked when the user explicitly sets the path.

User-facing contract:
```python
# user_process_data.py
def process_data(dataset_name, model_tokenizer, template_config, tmvp_config, x):
    return {"prompts": ..., "question": ..., "answer": ...}
```
```bash
python3 -m maxtext.trainers.post_train.rl.train_rl rl.yml \
  dataset_processor_path=/path/to/user_process_data.py \
  ...
```

## Checklist
- [x] Tested locally with a custom processor file (VTC-style raw-text prompt template); produced expected outputs
- [x] Backward compatible: default empty string preserves `utils_rl.process_data` behavior
- [x] No effect on non-RL paths (only `prepare_datasets` in the RL trainer touched)
- [x] `_load_custom_callable` doesn't pollute `sys.path` (uses `spec_from_file_location`)

### DIFF
--- a/src/maxtext/trainers/post_train/rl/train_rl.py
+++ b/src/maxtext/trainers/post_train/rl/train_rl.py
@@ -254,6 +254,17 @@ def prepare_datasets(
         f"Chat template is required for processing dataset but failed to load from {trainer_config.chat_template_path}"
     )
 
+  # Optional user-provided `process_data(dataset_name, tokenizer, template, config, x) -> dict`.
+  # When `dataset_processor_path` is set in config, load that file's `process_data`
+  # and use it instead of the built-in utils_rl.process_data. Lets users adapt
+  # custom datasets (with non-standard answer columns / cleaning) without editing maxtext.
+  _custom_processor_path = getattr(trainer_config, "dataset_processor_path", "") or ""
+  if _custom_processor_path:
+    _process_data = utils_rl.load_custom_callable(_custom_processor_path, "process_data")
+    max_logging.log(f"prepare_datasets: using custom process_data from {_custom_processor_path}")
+  else:
+    _process_data = utils_rl.process_data
+
   # Prepare train and test data from training data for certain datasets
   eval_dataset_name = getattr(trainer_config, "eval_dataset_name", None)
   test_dataset = None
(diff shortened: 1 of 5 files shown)


tests/post_training/unit/load_custom_callable_test.py
```diff
@@ -0,0 +1,130 @@
...
+  @pytest.mark.cpu_only
+  def test_raises_when_function_not_defined(self):
+    """File exists but doesn't define the named function -> ValueError."""
+    with tempfile.TemporaryDirectory() as tmpdir:
+      user_file = _write_user_file(tmpdir)
+      with self.assertRaises(ValueError):
+        _load_custom_callable(user_file, "no_such_function")
+
+  @pytest.mark.cpu_only
+  def test_does_not_pollute_sys_path(self):
```
khatwanimohit, line 109: for my understanding, why do we want to make sure the file does not append its directory to the sys.path. Is it because it is not meant to use any helper functions from maxtext ?

py4: The user's file CAN still import maxtext.utils.max_logging or any other maxtext helper. those work because maxtext is already on sys.path independently. What we're guarding against is the user's directory (and its sibling files) leaking into the global import namespace

khatwanimohit (approved):
LGTM

xuefgu (approved):

src/maxtext/trainers/post_train/rl/train_rl.py
```diff
@@ -243,6 +244,27 @@ def prepare_train_and_eval_dataset(
   }
 
 
+def _load_custom_callable(module_path: str, function_name: str) -> Callable:
```
A9isha: we should move this to utils_rl.py, the goal is to keep train_rl.py as minimal as possible

A9isha (commented):
Just one request for the refactor - the rest look good, thank you Pooya!

src/maxtext/trainers/post_train/rl/train_rl.py
py4: Tnx! done! ptal

PR #4040, opened 2026-06-02
Support Qwix quantization on NNX

hsuan-lun-chiang:
# Description

This PR enables full support for Qwix quantization algorithms on NNX models and restores support for Linen models. Because the underlying Transformer architecture has been migrated to an NNX module, the existing Qwix quantization integration for Linen models was broken. This PR fixes the Linen integration and extends full feature parity to the NNX backend.

# Tests

Gpt3-6b
[Qwix Quantization Test - Linen](https://paste.googleplex.com/5058255918333952)
[Qwix Quantization Test - NNX](https://paste.googleplex.com/6613645239910400)

Gemma4-26b
[Qwix Quantization Test - Linen](https://paste.googleplex.com/5045455372288000)
[Qwix Quantization Test - NNX](https://paste.googleplex.com/5499316176289792)

# Checklist

### DIFF
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -25,7 +25,7 @@ import jax
 import jax.numpy as jnp
 from flax import linen as nn
 from flax import nnx
-from flax.nnx import wrappers as nnx_wrappers
+from maxtext.layers import nnx_wrappers
 from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
 
@@ -716,17 +716,8 @@ class NNXDecoder(nnx.Module):
       out = merged_layer(y_in, **kwargs)
       return out, nnx.state(merged_layer)
 
-    # Linen FP8 ops keep amax_history in mutable Linen scope; jax.checkpoint
-    # re-traces and hits UnexpectedTracerError. Skip remat for FP8.
-    uses_linen_fp8_mutable_state = self.config.quantization in {
-        "fp8_nanoo",
-        "fp8_gpu",
-    }
-    if uses_linen_fp8_mutable_state:
-      out, new_state = pure_layer_fn(state, y)
-    else:
-      checkpointed_fn = jax.checkpoint(pure_layer_fn, policy=policy, prevent_cse=prevent_cse)
-      out, new_state = checkpointed_fn(state, y)
+    checkpointed_fn = jax.checkpoint(pure_layer_fn, policy=policy, prevent_cse=prevent_cse)
+    out, new_state = checkpointed_fn(state, y)
     nnx.update(layer, new_state)
 
     return out
@@ -854,26 +845,7 @@ class NNXDecoder(nnx.Module):
       params = nnx_ensure_scan_leading_axis(params, length)
       state = nnx_ensure_scan_leading_axis(state, length)
 
-      # Linen FP8 ops keep amax_history in mutable Linen scope; jax.lax.scan
-      # leaks the tracer and hits UnexpectedTracerError. Use a Python for-loop
-      # for FP8 instead.
-      uses_linen_fp8_mutable_state = self.config.quantization in {
-          "fp8_nanoo",
-          "fp8_gpu",
-      }
-      if uses_linen_fp8_mutable_state:
-        carry = x_in
-        per_layer_states = []
-        for i in range(length):
-          current_params = jax.tree.map(lambda x, i=i: x[i], params)
-          current_state = jax.tree.map(lambda x, i=i: x[i], state)
-          carry, new_state_i = layer_fn(carry, (current_params, current_state))
-          per_layer_states.append(new_state_i)
-        final_carry = carry
-        # pylint: disable-next=no-value-for-parameter  (*per_layer_states supplies the `tree` arg)
-        scanned_state = jax.tree.map(lambda *xs: jnp.stack(list(xs)), *per_layer_states)
-      else:
-        final_carry, scanned_state = jax.lax.scan(layer_fn_wrapped, x_in, (params, state))
+      final_carry, scanned_state = jax.lax.scan(layer_fn_wrapped, x_in, (params, state))
       returned_kv_stacked = None
 
     if scan_axis != 0:
--- a/src/maxtext/layers/nnx_wrappers.py
+++ b/src/maxtext/layers/nnx_wrappers.py
@@ -285,14 +285,18 @@ class ToNNX(Module):
       # Get `mutable` from top level bridge.Module context if any
       if mutable is not None:
         pass
-      elif (m := bdg_module.current_module()) is not None:
+      elif getattr(bdg_module.MODULE_CONTEXT, "module_stack", None) and (m := bdg_module.current_module()) is not None:
         assert m.scope is not None
         mutable = m.scope.mutable
       elif (m := current_linen_module()) is not None:
         assert m.scope is not None
         mutable = m.scope.mutable
       else:
-        mutable = False
+        # Safe fallback mutability: when running functionally isolated inside standard JAX transforms,
+        # we determine which collections (such as "stats" or "amax_history") are present and mark them mutable.
+        mutable = [k for k in variables.keys() if k != "params"]
+        if not mutable:
+          mutable = False
 
       out = self.to_nnx__module.apply(variables, *args, rngs=_rngs, method=method, mutable=mutable, **kwargs)
 
@@ -509,7 +513,31 @@ class ToLinen(linen.Module):
       for path, _ in unknown_state_flat.items():
         paths_str += f"\n  - {'/'.join(map(str, path))}"
 
-      warnings.warn(f"Found unknown module paths in incoming state:{paths_str}")
+        # Dynamically reconstruct the unknown variables
+        curr = module
+        for p in path[:-1]:
+          if isinstance(curr, dict):
+            if p not in curr:
+              curr[p] = nnx.Module()
+            curr = curr[p]
+          elif isinstance(curr, list):
+            if not isinstance(p, int):
+              raise TypeError(f"Expected int index for list, got {type(p)}: {p}")
+            while len(curr) <= p:
+              curr.append(nnx.Module())
+            curr = curr[p]
+          elif isinstance(curr, tuple):
+            raise ValueError(f"Cannot dynamically reconstruct elements within a tuple at path {path}.")
+          else:
+            if not isinstance(p, str):
+              p = str(p)
+            if not hasattr(curr, p):
+              setattr(curr, p, nnx.Module())
+            curr = getattr(curr, p)
+
+      warnings.warn(
+          f"Found unknown module paths in incoming state:{paths_str}. Intermediate modules have been reconstructed."
+      )
 
     _fix_for_qwix_quantization(module)
     nnx.update(module, new_state)
--- a/src/maxtext/layers/quantizations.py
+++ b/src/maxtext/layers/quantizations.py
@@ -38,6 +38,9 @@ from jax.tree_util import tree_flatten_with_path, tree_unflatten
 from flax.linen import fp8_ops
 from flax.linen import initializers as flax_initializers
 import flax.linen as nn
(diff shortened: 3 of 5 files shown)


src/maxtext/layers/nnx_wrappers.py
```diff
@@ -499,7 +499,16 @@ def maybe_unbox(x):
...
         paths_str += f"\n  - {'/'.join(map(str, path))}"
 
-      warnings.warn(f"Found unknown module paths in incoming state:{paths_str}")
+        # Dynamically reconstruct the unknown variables
+        curr = module
+        for p in path[:-1]:
+          if not hasattr(curr, p):
+            setattr(curr, p, nnx.Module())
+          curr = getattr(curr, p)
+
```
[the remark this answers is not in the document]
hsuan-lun-chiang: Updated the dynamic reconstruction logic in ToLinen. The loop now explicitly checks isinstance(curr, dict), isinstance(curr, list), and isinstance(curr, tuple) to construct the missing paths safely without throwing  hasattr  errors on integer indices.

src/maxtext/trainers/pre_train/train.py
```diff
@@ -371,15 +374,18 @@ def train_step(model, config, state_mesh_shardings, params_shardings, state, dat
...
+        _, _, new_overwrite_vars, new_rest = nnx.split(local_model, nnx.Param, OverwriteWithGradient, ...)
+        return loss, (aux, new_overwrite_vars, new_rest)
 
-      grad_func = jax.value_and_grad(diff_wrapper, argnums=0, has_aux=True)
-      (loss, (aux, new_rest)), raw_grads = grad_func(curr_params, rest, config, data)
+      grad_func = jax.value_and_grad(diff_wrapper, argnums=(0, 1), has_aux=True)
+      (loss, (aux, _, new_rest)), (raw_grads, overwrite_grads) = grad_func(
+          curr_params, overwrite_vars, rest, config, data
+      )
       nnx.update(state.model, new_rest)
```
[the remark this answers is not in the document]
hsuan-lun-chiang: By updating with overwrite_grads, we hook into JAX's existing gradient accumulation and distributed all-reduce infrastructure. This ensures that statistics like AMAX are automatically and correctly aggregated across all data-parallel workers, rather than having to manually handle the all-reduce logic for new_overwrite_vars. I've added a comment in train_step right above the nnx.update call to clarify this behaviour.

src/maxtext/layers/quantizations.py
```diff
@@ -707,6 +711,28 @@ def configure_kv_quant(config):
...
+  """Applies a Linen module within an NNX context."""
+  try:
+    parent = flax_util.get_current_module()
+    is_nnx = isinstance(parent, nnx.Module)
+  except Exception:  # pylint: disable=broad-exception-caught
+    is_nnx = False
+
+  if is_nnx:
+    attr_name = f"_qwix_fp8_gpu_{op_id}"
+    if not hasattr(parent, attr_name):
```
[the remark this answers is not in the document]
hsuan-lun-chiang: Updated the logic to check if  parent.rngs  exists (which is the standard NNX convention). If it does, we now call  parent.rngs.fork()  to cleanly branch the RNG stream for the quantization layer. If no parent rngs object is found, it safely falls back to nnx.Rngs(0) .

src/maxtext/layers/quantizations.py
```diff
@@ -812,7 +838,21 @@ def maybe_quantize_model(model, config):
   if config.use_qwix_quantization and not config.use_batch_split_schedule:
     quantization_provider = get_qt_provider(config)
     if quantization_provider:
-      model = qwix.quantize_model(model, quantization_provider)
+      if config.pure_nnx:
+        input_shape = (config.micro_batch_size_to_train_on, config.max_target_length)
+        dummy_tokens = jnp.ones(input_shape, dtype=jnp.int32)
+        dummy_positions = jnp.ones(input_shape, dtype=jnp.int32)
+        dummy_segment_ids = jnp.ones(input_shape, dtype=jnp.int32)
+        model = qwix.quantize_model(
```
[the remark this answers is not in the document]
hsuan-lun-chiang: They are purely dummy values for tracing shapes, so the jnp.ones array of the appropriate token shape is sufficient.

src/maxtext/utils/model_creation_utils.py
```diff
@@ -534,7 +534,10 @@ def get_nnx_create_model_fn(config, mesh=None, devices=None, model_mode=MODEL_MO
 
   def _create_model():
     rngs = maxtext_utils_nnx.create_nnx_rngs(config, model_mode=model_mode, rng_key=rng_key)
-    return from_config(config, devices, mesh, rngs=rngs, model_mode=model_mode)
+    model = from_config(config, devices, mesh, rngs=rngs, model_mode=model_mode)
+    if config.pure_nnx:
+      _ = nnx.pop(model, nnx.Intermediate)
```
bvandermoon: Why are we popping the intermediates here?

src/maxtext/layers/quantizations.py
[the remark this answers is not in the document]
bvandermoon: If they are just for tracing shapes, why not use `eval_shape`?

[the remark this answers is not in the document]
hsuan-lun-chiang: Because qwix.quantize_model actually executes a forward pass to trace the model and initialize the quantization states, we must provide actual numerical arrays (dummy variables). If we use eval_shape or  jax.ShapeDtypeStruct, it fails because it only provides shape metadata and cannot be used in actual tensor operations during the forward pass.

src/maxtext/utils/model_creation_utils.py
hsuan-lun-chiang: When `create_nnx_abstract_model` traces the model shapes, the forward pass naturally generates these intermediates. We must pop them here so the abstract state perfectly matches the clean, persistent parameter tree that is saved to and loaded from checkpoints. If we didn't pop them here, the system would mistakenly expect them to exist during checkpoint restoration, leading to PyTree mismatch errors.

bvandermoon (commented):
What is the `NaN` in the last line of https://paste.googleplex.com/5058255918333952?

RexBearIU:
> What is the `NaN` in the last line of https://paste.googleplex.com/5058255918333952?

 The warning `NaN or Inf found in input tensor` in `x2num.py` is a benign TensorBoard logging warning caused by the perplexity metric overflowing, not by any actual NaN/Inf in model weights or gradients.
  
 Since perplexity is calculated as `exp(loss)`, and at the start of training the loss is extremely high (above 88.7), `exp(loss)` overflows to positive infinity (`inf`) in standard float32 precision. When TensorBoard's       
  summary writer attempts to log this infinity value, its internal validation check in `x2num.py` raises the warning.
  
 This is completely normal and occurs identically in unquantized pre-training. Here is a log snippet from a standard, unquantized baseline pre-training run showing the exact same loss progression and warning:
           
```text
    completed step: 0, loss: 445.785, lm_loss: 445.785, perplexity: inf 
    W0610 08:12:16.985936 x2num.py:13] NaN or Inf found in input tensor.
  
    completed step: 1, loss: 445.785, lm_loss: 445.785, perplexity: inf 
    W0610 08:12:18.443920 x2num.py:13] NaN or Inf found in input tensor.
  
    completed step: 2, loss: 217.815, lm_loss: 217.815, perplexity: inf 
    W0610 08:12:19.904166 x2num.py:13] NaN or Inf found in input tensor.
```

src/maxtext/layers/nnx_decoders.py
```diff
@@ -536,22 +536,22 @@ def _update_leaf(leaf):
...
 
-    graphdef, state = nnx.split(layer)
-
-    def pure_layer_fn(state_in, y_in):
-      merged_layer = nnx.merge(graphdef, state_in)
-      out = merged_layer(y_in, **kwargs)
-      return out, nnx.state(merged_layer)
-
     # Linen FP8 ops keep amax_history in mutable Linen scope; jax.checkpoint
     # re-traces and hits UnexpectedTracerError. Skip remat for FP8.
```
bvandermoon: Why are we skipping remat for `FP8`? This is different from Linen, right?

RexBearIU: Under Flax Linen, `nn.remat` is a custom wrapper provided by Flax that has specialized handling for Flax's internal variable collections (such as the `amax_history` mutable collection used by FP8 operations) during rematerialization.

Under the NNX decoder when bridging NNX to Linen, the `_apply_layer_with_remat` helper splits the NNX module state and wraps the pure functionalized pass in standard `jax.checkpoint`. Since standard `jax.checkpoint` does not possess Flax Linen's specialized state-tracking mechanisms, the re-tracing process intercepts Flax's mutable tracers (which are captured inside Flax's `Scope` object), resulting in `UnexpectedTracerError`.

Therefore, we must skip the standard `jax.checkpoint` rematerialization when using Linen-based FP8 quantization configurations (`fp8_nanoo`, `fp8_gpu`) inside the NNX framework to avoid this tracing issue. This is indeed a difference between pure Linen and NNX decoder checkpointing.

bvandermoon: @cgarciae can you please take a look at this? Is there a way to make this work from the NNX persepctive? I don't think we will be able to complete the migration without FP8 remat working

src/maxtext/layers/nnx_wrappers.py
```diff
@@ -161,6 +161,33 @@ def lazy_init(fn: Module | tp.Callable[..., tp.Any], *args, **kwargs):
   return fn
 
 
+def isolate_linen_stacks(fn: tp.Callable[..., tp.Any], *args, **kwargs):
```
cgarciae: what is this doing?

RexBearIU: **What `isolate_linen_stacks` does and why it's needed:**
  
During JAX compilation (`.init` and `.apply`), Flax Linen uses a thread-local stack (`linen.module._context.module_stack`) to resolve the hierarchy and scopes of active Linen modules.
  
However, on our **Flax NNX** path, we dynamically execute wrapped Linen operators (via `ToNNX`) inside an outer NNX model context. If the outer  context or JAX transforms have active module contexts, the thread-local Linen stack can retain stale/unexpected elements, leading to compiled tracer leaks and JAX `UnexpectedTracerError`.
  
`isolate_linen_stacks` solves this by temporarily shadowing and clearing the thread-local Linen and Bridge module stacks during the inner `ToNNX`  execution, and restoring them immediately after.

cgarciae: `init_with_output` already does this isolation though...

RexBearIU: You're totally right, good catch. I overlooked that `init_with_output` already handles that isolation under the hood. I've removed the redundant wrapper.

src/maxtext/layers/quantizations.py
```diff
@@ -710,6 +713,32 @@ def configure_kv_quant(config):
   return None if not config.quantize_kvcache else KVQuant(config)
 
 
+def _apply_linen_module_in_nnx(linen_module_cls, op_id, *args, **kwargs):
+  """Applies a Linen module within an NNX context."""
+  try:
+    parent = flax_util.get_current_module()
+    is_nnx = isinstance(parent, nnx.Module)
+  except Exception:  # pylint: disable=broad-exception-caught
+    is_nnx = False
```
bvandermoon: What is the exception being caught here? Is it `ValueError` from Qwix? Mind just catching the specific error if so?

bvandermoon (approved):

src/maxtext/layers/quantizations.py
RexBearIU: Thanks to point this out! When called outside of a module context, `flax_util.get_current_module()` raises a `ValueError` (`"Current module is not known."`). I have updated the block to catch `ValueError` explicitly.

NuojCheng (approved):

khatwanimohit (approved):

PR #4043, opened 2026-06-02
Add elastic training (MaxText + Pathways on GKE) example

inardini:
This PR adds a runnable elastic training example: train Qwen3 0.6B across 3 TPU v5e slices on GKE with Pathways, kill a slice mid-run, and watch training recover in-process from the last checkpoint without restarting the job.

## Motivation

When a TPU slice fails mid-run (hardware fault, preemption, network blip), the default outcome is the whole job crashes and restarts from scratch, losing XLA compilation time plus everything since the last checkpoint. **Elastic training** keeps the process alive: Pathways detects the dead slice, MaxText's `elastic_retry` catches the error in-process, Orbax discards any half-written checkpoint, and training rewinds to the last good step. The head pod never restarts; recovery takes ~40s end to end. There is no runnable example in `examples/` that shows this end to end — this PR adds one.

## What's added

A self-contained example under `src/maxtext/examples/elastic_training/`:

- **`terraform/`** — GKE cluster, TPU + CPU node pools, GCS bucket, Artifact Registry, a dedicated Cloud Build SA, IAM, and HF-token wiring.
- **`deploy/`** — Cloud Build pipeline, the training JobSet + `train.sh`, and a data-prep job.
- **`Makefile`** — terminal driver (`make infra` / `deploy` / `logs` / `fail` / `demo` / `destroy`).
- **`elastic_qwen3_demo.ipynb`** — equivalent cell-by-cell run with narration and "what you should see" checkpoints; runs raw `terraform` / `gcloud` / `kubectl`, cleared outputs, no emojis.
- **`README.md`** — narrative, mermaid recovery diagram, flags appendix.

### DIFF
--- a/src/maxtext/examples/elastic_training/.env.example
+++ b/src/maxtext/examples/elastic_training/.env.example
@@ -1,0 +1,29 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Required
+PROJECT_ID=your-project-id
+HF_TOKEN=hf_...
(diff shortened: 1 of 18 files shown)


inardini:
Superseded by a reworked version per maintainer feedback: the example now lives under docs/ as an xpk-based guide (xpk workload create-pathways with --elastic-slices / --max-slice-restarts), is framed as demonstration-only, and disables profiling (an elastic event during a profile is unsupported). The reworked commands were validated end to end on 48 v5e chips. Replacement PR follows.

PR #4044, opened 2026-06-02
Add intermediate eval hook: fire evaluate() every eval_interval outer steps

py4:
## Summary

`eval_interval` was a silently-dead config knob: even when plumbed into tunix's `RLTrainingConfig.eval_every_n_steps`, tunix's `_run_eval` is a no-op unless an `eval_ds` is passed to `trainer.train()`. And even if you do pass one, tunix's default GRPO eval re-runs the full sampled rollout (`num_generations` responses per prompt), which is ~3hr/eval and impractical for trajectory monitoring.

This PR installs a `tunix.sft.hooks.TrainingHooks` subclass that hooks `on_train_step_end`, checks `rl_cluster.global_steps % eval_interval`, and calls maxtext's own `evaluate(...)` (greedy decode + the configured scoring pipeline). Gives matched-step PRE / step_N / POST trajectory logging at near-zero cost beyond the eval itself (which is already fast when `eval_batch_size` is set per #4030).

## Behavior

- No-op when `eval_interval <= 0` or `num_test_batches <= 0`.
- Soft-skips with a warning if `tunix.sft.hooks` isn't importable, so the launcher still works against a stock-only tunix.
- Default behavior (no `eval_interval` set) is unchanged.

## Checklist
- [x] PR scoped to one commit, no unrelated changes
- [x] Backward compatible — default-off via `eval_interval <= 0`
- [x] No effect on non-RL training paths

### DIFF
--- a/src/maxtext/trainers/post_train/rl/utils_rl.py
+++ b/src/maxtext/trainers/post_train/rl/utils_rl.py
@@ -760,3 +760,43 @@ class MaxTextChatParser(agentic_chat_template_parser.DefaultChatTemplateParser):
     return super().parse(
         messages=formatted_messages, add_generation_prompt=add_generation_prompt, is_first_msg=is_first_msg
     )
+
+
+def install_training_hooks(
+    rl_cluster: Any,
+    trainer_config: Any,
+    test_dataset: Any,
+) -> None:
+  """Install maxtext's `RLTrainingHooks` on the actor trainer.
+
+  No-op if `eval_interval <= 0` or `num_test_batches <= 0` or tunix's hooks
+  module is unavailable.
+  """
+  if trainer_config.num_test_batches <= 0:
+    return
+  eval_interval = int(getattr(trainer_config, "eval_interval", 0))
+  if eval_interval <= 0:
+    return
+  try:
+    # `hooks` hard-imports `tunix.sft.hooks`. If that's missing (stock-only
+    # tunix without the SFT hooks API), the import below raises and we
+    # soft-skip rather than crash the launcher.
+    from maxtext.trainers.post_train.rl.hooks import RLTrainingHooks  # pylint: disable=import-outside-toplevel
+  except ImportError:
+    max_logging.warning("[intermediate-eval] tunix.sft.hooks not importable; skipping hook install.")
+    return
+
+  # PeftTrainer composes a single training_hooks; install if free, else warn.
+  try:
+    actor = rl_cluster.actor_trainer
+    if getattr(actor, "training_hooks", None) is None:
+      actor.training_hooks = RLTrainingHooks(rl_cluster, trainer_config, test_dataset, eval_interval)
+      max_logging.warning(
+          f"[intermediate-eval] hook installed: evaluate(...) will fire every {eval_interval} outer steps."
+      )
+    else:
+      max_logging.warning(
+          "[intermediate-eval] actor.training_hooks already set; skipping install (chain manually if you need both)."
+      )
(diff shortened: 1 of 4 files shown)


A9isha (commented):
This is definitely very helpful. We should move the functionality to `utils_rl.py` such that `train_rl.py` stays clean and minimal.

I will let @SurbhiJainUSC to take a look too at this since she has worked on the hooks in the past for SFT.

Also, I think the hook structure will go away in medium future.

py4:
@A9isha addressed in 4db581ef — moved `_install_intermediate_eval_hook` (95 lines) out of `train_rl.py` and into `utils_rl.py` as the public `install_intermediate_eval_hook`. `train_rl.py` now just has the one-line call. Used a lazy import of `evaluate` inside the function to avoid the `utils_rl` ↔ `evaluate_rl` circular import (`evaluate_rl` imports `utils_rl` at module load time). Fine to noop later if the hook structure goes away upstream.

A9isha (approved):
Thank you Pooya

src/maxtext/trainers/post_train/rl/utils_rl.py
```diff
@@ -773,3 +760,103 @@ def parse(
...
+  except ImportError:
+    max_logging.warning("[intermediate-eval] tunix.sft.hooks not importable; skipping hook install.")
+    return
+  # Lazy import to avoid the utils_rl <-> evaluate_rl circular import
+  # (evaluate_rl imports utils_rl at module load time).
+  from maxtext.trainers.post_train.rl.evaluate_rl import evaluate  # pylint: disable=import-outside-toplevel
+
+  state: dict = {"last_step_evaluated": -1}
+
+  class _IntermediateEvalHook(_hk.TrainingHooks):  # type: ignore[name-defined]
```
SurbhiJainUSC: Can you move this Hook class out of `utils_rl.py` and create a separate module, similar to how other trainers in MaxText: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/trainers/post_train/dpo/hooks.py

py4: Done

py4:
Thanks @SurbhiJainUSC. moved the class out into its own module at `src/maxtext/trainers/post_train/rl/hooks.py` per the DPO layout, named it `RLTrainingHooks` matching `DPOTrainingHooks`, and kept the install function as `utils_rl.install_training_hooks(...)`. Pushed in 22c4b047.

One design choice I want to flag rather than guess on: I subclassed `tunix.sft.hooks.TrainingHooks` directly rather than `BaseTrainingHooks`. `BaseTrainingHooks` is built around the SFT/DPO per-step metric logging pipeline. it writes `learning/loss`, `learning/perplexity`, `learning/grad_norm`, runs the sharding asserts in `on_train_start`, records goodput events, and raises `StopTraining` against `target_eval_loss`. The RL path's training metrics already flow through tunix's own `metrics_logger`/wandb, the `get_total_weights(batch)` token count doesn't translate to GRPO's trajectory data, and the `__init__` signature `(config, mesh, learning_rate_schedule, goodput_recorder)` is shaped for the SFT/DPO use case. our hook needs `(rl_cluster, trainer_config, test_dataset, eval_interval)`. So inheriting `BaseTrainingHooks` would either force-fit those arguments or stub out most of the parent's lifecycle methods.

Happy to revisit if you intended the deeper inheritance and there's a clean way to reconcile the lifecycle differences  WDYT?

SurbhiJainUSC (approved):

PR #4045, opened 2026-06-02
Add reward_functions_path + reward_functions CLI knobs for custom rewards

py4:
> **Stacked on top of #4031** (`pr/dataset-processor-path`). This PR reuses the `_load_custom_callable` helper added there. Will rebase onto `main` after #4031 merges.

## Summary

The post-train RL reward stack in `create_rl_components` is hardcoded to a 3-function list:

```python
reward_fns = [
    make_reward_fn(utils_rl.match_format_exactly),
    make_reward_fn(utils_rl.match_format_approximately),
    make_reward_fn(utils_rl.check_numbers),
]
```

Replacing it requires editing `train_rl.py`. This is fine for the GSM8K out-of-the-box recipe but blocks any user-provided reward (VTC-style partial credit, `math_verify`-based scorer, etc.) without forking maxtext.

This PR adds two new config knobs:
- `reward_functions_path` (str, default `""`) — filesystem path to a user Python file
- `reward_functions` (str, default `""`) — comma-separated list of function names to import from that file

When **both** are set, the built-in reward list is REPLACED entirely by the user-provided functions. Each user function must accept `(prompts, completions, tmvp_config, **kwargs)` and return `list[float]`. Default behavior is unchanged.

Field declarations land in the same `RLDataset` mixin as `dataset_processor_path` (#4031), consistent with the file's other RL-only knobs.

## Checklist
- [x] Pydantic schema updated in `src/maxtext/configs/types.py` (`RLDataset` mixin)
- [x] Backward compatible — default-off when either knob is empty
- [x] No effect on non-RL training paths
- [x] Reuses `_load_custom_callable` from #4031 (covered by `load_custom_callable_test.py` there)

### DIFF
--- a/src/maxtext/configs/post_train/rl.yml
+++ b/src/maxtext/configs/post_train/rl.yml
@@ -232,6 +232,13 @@ eval_dataset_name: 'openai/gsm8k'
 # tmvp_config, x) -> dict with keys {prompts, question, answer}. When empty
 # (default), the built-in utils_rl.process_data is used.
 dataset_processor_path: ''
+# Optional: path to a user-provided Python file with custom reward functions,
+# AND a comma-separated list of function names to import from that file.
+# Each function signature: fn(prompts, completions, tmvp_config, **kwargs) -> list[float].
+# When both are set, the built-in [match_format_exactly, match_format_approximately,
+# check_numbers] reward list is REPLACED entirely.
+reward_functions_path: ''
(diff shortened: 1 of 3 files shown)


py4:
Closing for now — will reopen after #4031 (`pr/dataset-processor-path`) merges so this can target `main` cleanly instead of being stacked.

PR #4051, opened 2026-06-03
Support of gdn kernel from tpu-inference

khatwanimohit:
# Description

- This PR is inspired from @NicoGrande's PR #3917. ( Thank you @NicoGrande, we miss you! 😊 )
- GDN Integration and Sharding Support: Integrated Gated Delta Net (GDN) logic from  nicogrande/add-gdn-support .
- GDN Expert Replication: Added experimental  self._gdn_replicate_expert  option (triggered via the  MAXTEXT_GDN_REPLICATE_EXPERT  environment variable) to control whether  attn_head  is set to ShardingAxisName.MODEL  or  ShardingAxisName.ATTN_HEAD.
- Unit Test Fixes: Resolved initialization errors in  attention_test.py  by ensuring  self.mesh  is properly initialized
- Adds support for profiling in vllm_decode.py

BUGS: b/517158881

# Tests

```
NEW_MODEL_DESIGN=1 \
    PYTHONPATH=/home/mohitkhatwani_google_com/workspace/maxtext/src \
    MAXTEXT_GDN_REPLICATE_EXPERT=true \
    /home/mohitkhatwani_google_com/workspace/max_venv/bin/python3 \
      -m maxtext.inference.vllm_decode src/maxtext/configs/base.yml \
      base_output_directory=gs://runner-maxtext-logs \
      run_name=mohit-qwen3.5-maxtext-bench-$RANDOM \
      model_name=qwen3.5-35b-a3b \
      tokenizer_path=Qwen/Qwen3.5-35B-A3B \
      vllm_hf_overrides='{architectures: ["MaxTextForCausalLM"]}' \
      load_parameters_path=gs://maxtext-model-checkpoints/qwen3.5-35b-a3b/unscanned/0/items \
      ici_tensor_parallelism=4 \
      hbm_utilization_vllm=0.5 \
      prompt="Tell me three fun facts about Buenos Aires." \
      decode_sampling_temperature=0.0 decode_sampling_nucleus_p=1.0 decode_sampling_top_k=0.0 \
      pure_nnx_decoder=True use_chat_template=True max_num_seqs=1 \
      enable_dp_attention=False prefuse_moe_weights=True max_target_length=64 debug_sharding=False \
      scan_layers=false profiler=xplane
 ```
 
 Profile: https://xprof.corp.google.com/trace_viewer/mohitkhatwani-11546393038070241787?view_start=71184.720&view_end=71204.219
 
 Logs: https://paste.googleplex.com/5464600022745088
 
 Decode performance: 25ms

# Checklist

### DIFF
--- a/src/maxtext/inference/vllm_decode.py
+++ b/src/maxtext/inference/vllm_decode.py
@@ -42,6 +42,7 @@ from maxtext.utils import model_creation_utils
 from maxtext.utils import max_logging
 from maxtext.utils.globals import MAXTEXT_CONFIGS_DIR
 from maxtext.common.common_types import Config
+from maxtext.common import profiler
 from maxtext.integration.tunix.tunix_adapter import TunixMaxTextAdapter
 from tunix.rl.rollout import base_rollout
 from tunix.rl.rollout.vllm_rollout import VllmRollout
@@ -50,6 +51,13 @@ from vllm.sampling_params import SamplingParams
 from maxtext.configs import pyconfig
 import maxtext.integration.vllm.maxtext_vllm_adapter as adapter
 
+# Force uses_mrope to False to disable 3D multimodal position IDs in text-only runs.
+# TODO(b/520142315): Class-level monkey patching is required here because ModelConfig.uses_mrope
+# is evaluated during the internal initialization of the LLM engine, making instance-level
+# configuration impossible. Check if a cleaner configuration option can be added in upstream vLLM.
+from vllm.config import ModelConfig
+
+ModelConfig.uses_mrope = property(lambda _: False)
 
 # --- DEFINE FLAGS GLOBALLY ---
 FLAGS = flags.FLAGS
@@ -110,6 +118,11 @@ def decode_with_vllm(config: Config) -> None:
     vllm_args["additional_config"]["sharding"]["sharding_strategy"]["expert_parallelism"] = config.ici_expert_parallelism
     vllm_args["enable_expert_parallel"] = enable_expert_parallel
 
+  if config.max_num_batched_tokens is not None:
+    vllm_args["max_num_batched_tokens"] = config.max_num_batched_tokens
+  if config.max_num_seqs is not None:
+    vllm_args["max_num_seqs"] = config.max_num_seqs
+
   max_logging.log(
       f"Initializing LLM with DP={config.ici_data_parallelism}, TP={config.ici_tensor_parallelism} "
       f"and EP={config.ici_expert_parallelism if enable_expert_parallel else 1}..."
@@ -157,7 +170,10 @@ def decode_with_vllm(config: Config) -> None:
       top_p=top_p,
   )
 
+  prof = profiler.Profiler(config)
+  prof.activate()
   outputs = llm.generate(prompts, sampling_params)
+  prof.deactivate()
 
   # max_logging.log Outputs
   for output in outputs:
--- a/src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
+++ b/src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
@@ -40,6 +40,11 @@ except ImportError:
 
 from vllm.config import VllmConfig
 
+# Threshold to determine if the ratio of attention to mamba layers is highly imbalanced.
+# If max_count / min_count >= this threshold, we group KV cache allocations by the
+# smaller count to prevent excessive memory padding for the minority layer type.
+_HYBRID_LAYER_IMBALANCE_THRESHOLD = 1.5
+
 
 def next_power_of_two(x: int) -> int:
   """Finds the smallest power of 2 >= x using bit manipulation.
@@ -56,7 +61,7 @@ def next_power_of_two(x: int) -> int:
   return 1 << (x - 1).bit_length()
 
 
-def generate_maxtext_config(vllm_config: VllmConfig, mesh: Mesh) -> pyconfig.HyperParameters:
+def generate_maxtext_config(vllm_config: VllmConfig) -> pyconfig.HyperParameters:
   """Generates a MaxText configuration from a vLLM configuration.
 
   This function takes a vLLM configuration object and translates relevant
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: src/maxtext/models/qwen3.py [hunk 1]
```diff
@@ -381,13 +385,25 @@ class Qwen3NextGatedDeltaNet(nnx.Module):
   2. output = Linear_out(y)
   """
 
-  def __init__(self, config: Config, dtype: DType = jnp.float32, model_mode: str = MODEL_MODE_TRAIN, *, rngs: nnx.Rngs):
+  def __init__(
+      self,
+      config: Config,
+      mesh=None,
```
xuefgu, line 457: The constructor signature makes `mesh` optional, but the forward pass unconditionally uses `self.mesh` in `jax.shard_map` (via `shard_mapped_delta_rule`) for the standard (non-paged) path. If `mesh` is not provided, this will crash during the forward pass.

We should either make `mesh` a required argument in `__init__` (and update docstrings), or add a fallback path in `__call__` that bypasses `shard_map` when `mesh` is `None` (which is useful for single-device CPU/GPU debugging).

src/maxtext/inference/vllm_decode.py
```diff
@@ -52,6 +53,11 @@
 
 adapter.register()
 
+# Force uses_mrope to False to disable 3D multimodal position IDs in text-only runs.
+from vllm.config import ModelConfig
+
+ModelConfig.uses_mrope = property(lambda _: False)
```
xuefgu, line 60: Globally monkey-patching `ModelConfig.uses_mrope` on the class level is fragile. Is there a way to configure this at the engine initialization level (e.g., via `vllm_args` or `ModelConfig` instance arguments) instead of patching the class? If this is a temporary workaround for a vLLM limitation, please add a TODO linking to the tracking issue.

src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -324,3 +341,159 @@ def load_weights(self, rng_key: jax.Array) -> None:
...
+      value_dim = head_v_dim * num_v_heads
+      conv_dim = key_dim * 2 + value_dim
+
+      conv_state_shape = (conv_kernel_size - 1, conv_dim)
+      recurrent_state_shape = (num_v_heads, head_k_dim, head_v_dim)
+
+      mamba_shapes = (conv_state_shape, recurrent_state_shape)
+
+      torch_dtype = torch.bfloat16
+      if str(cfg.dtype) == "float32":
```
xuefgu, line 423: Can the dtype be `float16`? If so, we want to handle that explicitly.

src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -324,3 +341,159 @@ def load_weights(self, rng_key: jax.Array) -> None:
...
+      num_layers = cfg.base_num_decoder_layers
+      num_attn = num_layers // interval
+      num_mamba = num_layers - num_attn
+
+      min_count = min(num_attn, num_mamba)
+      max_count = max(num_attn, num_mamba)
+      if max_count < min_count * 1.5:
+        group_size = max_count
+      else:
+        group_size = min_count
```
xuefgu: The logic to `calculate group_size` and the 1.5 multiplier threshold is not immediately obvious.

Could you add a brief comment explaining the design choice behind this grouping logic and how it relates to uniform page size allocation in `tpu_inference`? This will help future maintainers understand the math.

xuefgu (changes requested):

src/maxtext/models/qwen3.py
khatwanimohit: Thanks for catching this. I've updated the forward pass to include a fallback path. If  self.mesh  is  None  (e.g., during single-device CPU/GPU debugging), we bypass  jax.shard_map  completely and directly call  jax_chunk_gated_delta_rule

src/maxtext/inference/vllm_decode.py
khatwanimohit: Class-level monkey-patching is unfortunately necessary here because  ModelConfig.uses_mrope  is evaluated internally inside the  LLM(...)  constructor when the engine initializes (which happens before the  LLM  instance is returned, making instance-level configuration impossible). I've added a TODO linking to the tracking issue to see if we can expose this configuration parameter in upstream vLLM.

src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
khatwanimohit: Good point. I've updated the mapping logic to handle  float16  explicitly so that  torch_dtype  maps correctly to  torch.float16 , and configured the  dtype_size  to be set to 4 bytes for  float32  and 2 bytes for 16-bit float types.

src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
khatwanimohit: I've added comments explaining the design decisions behind the grouping math and the  1.5  multiplier threshold.

src/maxtext/integration/vllm/maxtext_vllm_adapter/adapter.py
```diff
@@ -324,3 +341,168 @@ def load_weights(self, rng_key: jax.Array) -> None:
...
+      # To allocate memory uniformly for a hybrid model's KV/recurrent cache page table,
+      # we group layers together. The uniform page size must support both attention and
+      # mamba layers.
+      # If the ratio of attention to mamba layers is relatively balanced (less than 1.5x),
+      # we use the larger count as the group size to minimize the total number of groups.
+      # If they are highly imbalanced (>= 1.5x), we group by the smaller count to prevent
+      # the page size from being inflated by excessive padding for the minority layer type.
+      min_count = min(num_attn, num_mamba)
+      max_count = max(num_attn, num_mamba)
+      if max_count < min_count * 1.5:
```
xuefgu: Can we make this 1.5 a constant with comment?

xuefgu (approved):
A nit but otherwise LGTM. Thanks for the PR!

A9isha (approved):
Thanks Mohit!

PR #4060, opened 2026-06-03
Measure routing mismatch for Qwen models

xuefgu:
# Description
This PR implements MoE routing mismatch measurement during training. It compares the model's natural routing decisions with forced_routed_experts passed to the model call, and logs the average mismatch rate.

Crucially, this PR only measures the mismatch and does not force the routing. The model's natural routing is still used for the actual computation.

The changes are only scoped to Qwen3 and Qwen3.5. More model support should be added at a later time.

# Tests

Unit tests.

# Checklist

### DIFF
--- a/src/maxtext/layers/decoders.py
+++ b/src/maxtext/layers/decoders.py
@@ -856,6 +856,7 @@ class Decoder(nn.Module):
       kv_caches: list[jax.Array] | None = None,
       attention_metadata=None,
       deepstack_visual_embeds: None | list[jnp.ndarray] = None,
+      forced_routed_experts: jax.Array | None = None,
   ):
     cfg = self.config
     mesh = self.mesh
@@ -877,6 +878,11 @@ class Decoder(nn.Module):
       y = mhc_expand(y)
 
     policy = self.get_remat_policy()
+    supports_forced_routing = cfg.decoder_block in (
+        DecoderBlockType.QWEN3_MOE,
+        DecoderBlockType.QWEN3_NEXT,
+        DecoderBlockType.QWEN3_5,
+    )
     RemattedBlockLayers = self.set_remat_policy(self.decoder_layer, policy)
     # scan does not support kwargs in layer call, passing broadcast_args as positional arg
     broadcast_args = (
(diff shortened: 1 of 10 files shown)


src/maxtext/layers/decoders.py
```diff
@@ -1150,6 +1194,25 @@ def __call__(
             layer = RemattedBlockLayer(
                 config=cfg, mesh=mesh, name=f"layers_{lyr}", quant=self.quant, model_mode=self.model_mode, **layer_kwargs
             )
+            is_moe = False
```
igorts-git: I am not sure that `is_moe` is the correct name here. Do only the three QWEN3 models listed below support MoE? Could we reuse `supports_forced_routing` defined above?

src/maxtext/layers/moe.py
```diff
@@ -2578,13 +2666,35 @@ def __call__(
...
+          w0_kernel,
+          w1_kernel,
+          wo_kernel,
+          w0_bias,
+          w1_bias,
+          wo_bias,
+          input_ids=input_ids,
+          forced_routed_experts=forced_routed_experts,
       )
+    if mismatch_rate is not None:
```
igorts-git, line 2696: should we set `self.mismatch_rate=None` in the `else`-clause? I think that otherwise we will get an exception in line 1875 (if I am tracing this file correctly)

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1568,7 +1635,22 @@ def pure_layer_fn(graphdef, state_in, y_in, kv_in):
         if input_tokens is not None:
           layer_kwargs["decoder_input_tokens"] = input_tokens
 
-        y, kv_cache, new_state = checkpointed_fn(graphdef, state, y, kv_cache)
+        is_moe = False
```
igorts-git: Similar comment about the name `is_moe` vs `supports_forced_routing`

igorts-git (approved):

src/maxtext/layers/decoders.py
xuefgu: Done.

src/maxtext/layers/moe.py
xuefgu: Done

src/maxtext/layers/nnx_decoders.py
xuefgu: Done

PR #4084, opened 2026-06-05
Add colocated python Dockerfile and build upload script

ycchenzheng:
# Description

This PR is to add colocated python image related Dockerfile and build&upload script.
1. `src/dependencies/dockerfiles/colocated_python.Dockerfile` is added to install all necessary dependencies for base image `us-docker.pkg.dev/cloud-tpu-v2-images/pathways-colocated-python/sidecar`
1. `src/dependencies/scripts/build_upload_colocated_python_image.sh` will automatically pick base docker image from `us-docker.pkg.dev/cloud-tpu-v2-images/pathways-colocated-python/sidecar` that matches `jax` version in `src/dependencies/requirements/generated_requirements/tpu-requirements.txt`

Example usage:
```
# Example 1: Use defaults for PROJECT and LOCAL_IMAGE_NAME
# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev

# Example 2: Specify a custom project and local name
# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev LOCAL_IMAGE_NAME=maxtext-cp

# Example 3: Provide a complete IMAGE_LOCATION, overriding defaults
# bash src/dependencies/scripts/build_upload_colocated_python_image.sh IMAGE_LOCATION=us-docker.pkg.dev/my-artifact-repo/images/colocated-python:stable

# Example 4: Specify a custom JAX version
# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev JAX_VERSION=0.10.0
```

FIXES: b/515481092

# Tests

Tested colocated python images built by this change with different TPU versions & topologies
[Example logs](https://cloudlogging.app.goo.gl/yVdb2BefDmEvqHoq9)

# Checklist

### DIFF
--- a/src/dependencies/scripts/build_upload_colocated_python_image.sh
+++ b/src/dependencies/scripts/build_upload_colocated_python_image.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+
+# Copyright 2023–2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Script to build and upload the MaxText colocated Python Docker image.
+# This script should be run from the root directory of the MaxText repository.
+#
+# Arguments can be provided as KEY=VALUE pairs, e.g.:
+#   bash build_upload_colocated_python_image.sh PROJECT=my-gcp LOCAL_IMAGE_NAME=my-colocated
+#
+# Supported KEYs:
+#   PROJECT: GCP project for gcr.io. 
+#   LOCAL_IMAGE_NAME: Overrides the local tag used during docker build. Defaults to 'maxtext-colocated-python'.
+#   IMAGE_LOCATION: The full URL for the final image in the registry.
+#                   Defaults to gcr.io/${PROJECT}/${USER}_${LOCAL_IMAGE_NAME}:latest.
+#                   Setting this variable fully overrides PROJECT and LOCAL_IMAGE_NAME.
+
+# Example 1: Provide PROJECT
+# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev
+
+# Example 2: Specify a custom local image name
+# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev LOCAL_IMAGE_NAME=maxtext-cp
+
+# Example 3: Provide a complete IMAGE_LOCATION
+# bash src/dependencies/scripts/build_upload_colocated_python_image.sh IMAGE_LOCATION=us-docker.pkg.dev/my-artifact-repo/images/colocated-python:stable
+
+# Example 4: Specify a custom JAX version
+# bash src/dependencies/scripts/build_upload_colocated_python_image.sh PROJECT=my-tpu-dev JAX_VERSION=0.10.0
+
+for ARGUMENT in "$@"; do
+    if [[ "$ARGUMENT" == *"="* ]]; then
+        IFS='=' read -r RAW_KEY VALUE <<< "$ARGUMENT"
+        KEY=$(echo "$RAW_KEY" | tr '[:lower:]' '[:upper:]')
+        export "$KEY"="$VALUE"
+        echo "  Parsed: $KEY=$VALUE"
+    else
+        echo "Warning: Ignoring argument '$ARGUMENT'. Arguments should be in KEY=VALUE format (e.g., PROJECT=my-proj)."
+    fi
+done
+
+# Set default values using parameter expansion.
+: "${LOCAL_IMAGE_NAME:=maxtext-colocated-python}"
+if [[ -z "${IMAGE_LOCATION}" ]]; then
+  # IMAGE_LOCATION was not provided, so PROJECT is required for the default.
+  if [[ -z "${PROJECT}" ]]; then
+    echo "Error: Either PROJECT or IMAGE_LOCATION must be provided." >&2
+    echo "  - To use a default location in GCR, set PROJECT=your-gcp-project-id." >&2
+    echo "  - To specify a full custom location, set IMAGE_LOCATION=your/registry/path/image:tag." >&2
+    exit 1
+  fi
+  IMAGE_LOCATION="gcr.io/${PROJECT}/${USER}_${LOCAL_IMAGE_NAME}:latest"
+fi
+
+DOCKERFILE_TEMPLATE="src/dependencies/dockerfiles/colocated_python.Dockerfile.template"
+CONTEXT="maxtext/"
+
+echo "$(date): Building and pushing MaxText Colocated Python image..."
+echo "  PROJECT: ${PROJECT}"
+echo "  LOCAL_IMAGE_NAME: ${LOCAL_IMAGE_NAME}"
+echo "  IMAGE_LOCATION: ${IMAGE_LOCATION}"
+echo "  Dockerfile Template: ${DOCKERFILE_TEMPLATE}"
+echo "  Build Context: ${CONTEXT}"
+
+# --- Step 1: Handle JAX Version ---
+REQ_FILE="src/dependencies/requirements/generated_requirements/tpu-requirements.txt"
+if [[ -z "${JAX_VERSION}" ]]; then
+    echo "$(date): Extracting JAX version from requirements..."
+    if [[ ! -f "${REQ_FILE}" ]]; then
+      echo "Error: Requirements file not found: ${REQ_FILE}" >&2
+      exit 1
+    fi
+    # Extracts version like "0.10.0" from lines like "jax>=0.10.0" or "jax==0.10.0"
+    JAX_VERSION=$(grep -E "^jax(>=|==)" "${REQ_FILE}" | head -1 | sed -E 's/^jax(>=|==)([0-9.]+).*/\2/')
+    if [[ -z "${JAX_VERSION}" ]]; then
+      echo "Error: Could not extract jax version from ${REQ_FILE}. Ensure it's in the format 'jax>=X.Y.Z' or 'jax==X.Y.Z'." >&2
+      exit 1
+    fi
+    echo "  Detected required JAX version: ${JAX_VERSION}"
+else
+    TEMP_REQ_FILE=$(mktemp)
+    echo "$(date): Generating temporary requirements file: ${TEMP_REQ_FILE} for custom JAX version ${JAX_VERSION}"
+    trap 'rm -f "${TEMP_REQ_FILE}"' EXIT
+    sed "s/^jax\(>=\|==\).*$/jax==${JAX_VERSION}/" "${REQ_FILE}" > "${TEMP_REQ_FILE}"
+    REQ_FILE="${TEMP_REQ_FILE}"
+fi
+
+# --- Step 2: Find the Latest Compatible Base Image Tag ---
+BASE_IMAGE_REPO="us-docker.pkg.dev/cloud-tpu-v2-images/pathways-colocated-python/sidecar"
+TARGET_JAX_TAG_PART="jax_${JAX_VERSION}"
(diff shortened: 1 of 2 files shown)


src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,137 @@
...
+        KEY=$(echo "$RAW_KEY" | tr '[:lower:]' '[:upper:]')
+        export "$KEY"="$VALUE"
+        echo "  Parsed: $KEY=$VALUE"
+    else
+        echo "Warning: Ignoring argument '$ARGUMENT'. Arguments should be in KEY=VALUE format (e.g., PROJECT=my-proj)."
+    fi
+done
+
+# Set default values using parameter expansion.
+: "${PROJECT:=cloud-tpu-multipod-dev}"
```
lukebaumann: Do not set a default for the project

src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,137 @@
...
+# limitations under the License.
+
+# Script to build and upload the MaxText colocated Python Docker image.
+# This script should be run from the root directory of the MaxText repository.
+#
+# Arguments can be provided as KEY=VALUE pairs, e.g.:
+#   bash build_upload_colocated_python_image.sh PROJECT=my-gcp LOCAL_IMAGE_NAME=my-colocated
+#
+# Supported KEYs:
+#   PROJECT: Overrides the default GCP project for gcr.io. Defaults to 'cloud-tpu-multipod-dev'.
```
lukebaumann: do not use a default for project

src/dependencies/dockerfiles/colocated_python.Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM us-docker.pkg.dev/cloud-tpu-v2-images/pathways-colocated-python/sidecar:20260423-python_3.12-jax_0.10.0
```
lukebaumann: Can you make this a placeholder and rename the dockerfile to colocated_python.Dockerfile.template

src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,137 @@
...
+
+# Extract JAX Version from requirements.txt
+echo "$(date): Extracting JAX version from requirements..."
+REQ_FILE="src/dependencies/requirements/generated_requirements/tpu-requirements.txt"
+if [[ ! -f "${REQ_FILE}" ]]; then
+  echo "Error: Requirements file not found: ${REQ_FILE}" >&2
+  exit 1
+fi
+# Extracts version like "0.10.0" from lines like "jax==0.10.0"
+JAX_VERSION=$(grep "^jax>=" "${REQ_FILE}" | head -1 | sed -E 's/.*>=([0-9.]+).*/\1/')
```
lukebaumann: This seems like a brittle way of extracting the JAX version. Is there another more robust way of getting the actual JAX version that would be resolved by pip/uv?

lukebaumann (changes requested):

src/dependencies/scripts/build_upload_colocated_python_image.sh
ycchenzheng: @sadikneipp can you please give some suggestions on this one?

src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,134 @@
...
+# Arguments can be provided as KEY=VALUE pairs, e.g.:
+#   bash build_upload_colocated_python_image.sh PROJECT=my-gcp LOCAL_IMAGE_NAME=my-colocated
+#
+# Supported KEYs:
+#   PROJECT: GCP project for gcr.io. 
+#   LOCAL_IMAGE_NAME: Overrides the local tag used during docker build. Defaults to 'maxtext-colocated-python'.
+#   IMAGE_LOCATION: The full URL for the final image in the registry.
+#                   Defaults to gcr.io/${PROJECT}/${USER}_${LOCAL_IMAGE_NAME}:latest.
+#                   Setting this variable fully overrides the default construction using PROJECT and LOCAL_IMAGE_NAME.
+# For instructions on building the MaxText Docker image, please refer to the https://maxtext.readthedocs.io/en/latest/build_maxtext.html.
```
sadikneipp: I don't think this is relevant here since the intent of this image is different than the linked intent

src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,134 @@
...
+    exit 1
+  fi
+  IMAGE_LOCATION="gcr.io/${PROJECT}/${USER}_${LOCAL_IMAGE_NAME}:latest"
+fi
+
+echo "$(date): Building and pushing MaxText Colocated Python image..."
+echo "  PROJECT: ${PROJECT}"
+echo "  LOCAL_IMAGE_NAME: ${LOCAL_IMAGE_NAME}"
+echo "  IMAGE_LOCATION: ${IMAGE_LOCATION}"
+echo "  Dockerfile Template: src/dependencies/dockerfiles/colocated_python.Dockerfile.template"
```
sadikneipp: Consider not hard-coding these.  This echo can become misleading if locations change

src/dependencies/scripts/build_upload_colocated_python_image.sh
```diff
@@ -0,0 +1,134 @@
...
+echo "  Build Context: maxtext/"
```
sadikneipp: same as above, consider not hard-coding

src/dependencies/scripts/build_upload_colocated_python_image.sh
sadikneipp: We'll use whatever we parse here to query AR below, which does not cover many JAX versions. A simple enhancement here is to make it work for a pinned (==) JAX too. A more robust pip/uv interop would be great but I think might bloat

ycchenzheng: Added for checking version with pinned (==) JAX

src/dependencies/scripts/build_upload_colocated_python_image.sh
ycchenzheng: Removed the unrelated link and description

src/dependencies/scripts/build_upload_colocated_python_image.sh
ycchenzheng: Added a var for the Dockerfile path

src/dependencies/scripts/build_upload_colocated_python_image.sh
sadikneipp: We have discussed offline and we believe it's better to expose the JAX version as an argument of this script or the dockerfile instead of something we parse from JAX. It means the invocation command will change but it's better to have the user pick the version instead of a loose requirement from maxtext (which might make the sidecar build with jax 0.10.0 and maxtext with 0.10.2, for instance).

sadikneipp (approved):

lukebaumann (approved):

src/dependencies/scripts/build_upload_colocated_python_image.sh
ycchenzheng: @sadikneipp I updated the build script with the following and tested again
```
# --- Step 1: Handle JAX Version ---
REQ_FILE="src/dependencies/requirements/generated_requirements/tpu-requirements.txt"
if [[ -z "${JAX_VERSION}" ]]; then
    echo "$(date): Extracting JAX version from requirements..."
    if [[ ! -f "${REQ_FILE}" ]]; then
      echo "Error: Requirements file not found: ${REQ_FILE}" >&2
      exit 1
    fi
    # Extracts version like "0.10.0" from lines like "jax>=0.10.0" or "jax==0.10.0"
    JAX_VERSION=$(grep -E "^jax(>=|==)" "${REQ_FILE}" | head -1 | sed -E 's/^jax(>=|==)([0-9.]+).*/\2/')
    if [[ -z "${JAX_VERSION}" ]]; then
      echo "Error: Could not extract jax version from ${REQ_FILE}. Ensure it's in the format 'jax>=X.Y.Z' or 'jax==X.Y.Z'." >&2
      exit 1
    fi
    echo "  Detected required JAX version: ${JAX_VERSION}"
else
    TEMP_REQ_FILE=$(mktemp)
    echo "$(date): Generating temporary requirements file: ${TEMP_REQ_FILE} for custom JAX version ${JAX_VERSION}"
    trap 'rm -f "${TEMP_REQ_FILE}"' EXIT
    sed "s/^jax\(>=\|==\).*$/jax==${JAX_VERSION}/" "${REQ_FILE}" > "${TEMP_REQ_FILE}"
    REQ_FILE="${TEMP_REQ_FILE}"
fi
```
and got the following error if I set `JAX_VERSION` to `0.9.2` (`0.10.0` works fine)
```
 => ERROR [6/7] RUN uv pip install -r maxtext/src/dependencies/requirements/generated_requirements/tpu-requirements.txt -c /opt/ven  5.2s 
------                                                                                                                                    
 > [6/7] RUN uv pip install -r maxtext/src/dependencies/requirements/generated_requirements/tpu-requirements.txt -c /opt/venv/server_constraints.txt:                                                                                                                               
0.721 Using Python 3.12.13 environment at: /opt/venv                                                                                      
5.102   נNo solution found when resolving dependencies:                                                                                  
5.102   \u2570\u2500\u25b6 Because only flax<=0.12.7 is available and flax==0.12.7 depends on                                                            
5.102       jax>=0.10.0, we can conclude that flax>=0.12.7 depends on jax>=0.10.0.
5.102       And because you require flax>=0.12.7 and jax==0.9.2, we can conclude
5.102       that your requirements are unsatisfiable.
------

 2 warnings found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE} results in empty or invalid base image name (line 3)
 - UndefinedVar: Usage of undefined variable '$PYTHONPATH' (line 18)
colocated_python.Dockerfile.template:15
```
In other words, the build script installs all deps from `tpu-requirements.txt`, if we lower the jax version, there might be conflicts with other deps since the req file sets the lowest versions that do not support the jax version user wants. 
Do you have any suggestions about this or keep it as is to at least support current maxtext?

sadikneipp: If there's conflicts and they are clearly raised to the user that's fine. A user in this scenario will know 0.10.0 works and 0.9.2 would need dependency changes in maxtext.

ycchenzheng: Got it. Will push the change above

ycchenzheng:
@bvandermoon Can you please review this PR?

SurbhiJainUSC:
Synced with @ycchenzheng offline and here are the next steps that we discussed:
This PR restricts users to run this script only by cloning Git repository. However, MaxText is only Pip installable, so we suggest:
1. Add a new flag and call `src/dependencies/scripts/build_upload_colocated_python_image.sh` behind that flag here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/dependencies/scripts/build_maxtext_docker_image.py#L27. This way, users can simply run `build_maxtext_docker_image` command with that new flag to build the docker image.
2. Modify `build_upload_colocated_python_image.sh` to only build docker image and use `docker_upload_runner.sh` to upload docker image to GCR.
3. Update https://github.com/AI-Hypercomputer/maxtext/blob/main/src/dependencies/scripts/upload_maxtext_docker_image.py to upload colocated python image to GCR.

SurbhiJainUSC (approved):

ycchenzheng:
> Synced with @ycchenzheng offline and here are the next steps that we discussed: This PR restricts users to run this script only by cloning Git repository. However, MaxText is only Pip installable, so we suggest:
> 
> 1. Add a new flag and call `src/dependencies/scripts/build_upload_colocated_python_image.sh` behind that flag here: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/dependencies/scripts/build_maxtext_docker_image.py#L27. This way, users can simply run `build_maxtext_docker_image` command with that new flag to build the docker image.
> 2. Modify `build_upload_colocated_python_image.sh` to only build docker image and use `docker_upload_runner.sh` to upload docker image to GCR.
> 3. Update https://github.com/AI-Hypercomputer/maxtext/blob/main/src/dependencies/scripts/upload_maxtext_docker_image.py to upload colocated python image to GCR.

Tracking future implementation in b/528026019

PR #4097, opened 2026-06-06
Add validation to block unsupported parameters in post-training SFT, …

shralex:
…RL, and DPO

- SFT, RL, and DPO trainers now error out on optimizer_memory_host_offload=True since the underlying Tunix trainers do not support host offloading.
- DPO and RL trainers now error out on num_vocab_tiling > 1 since they require full vocabulary projections to compute preference/policy log-probabilities, which is incompatible with vocabulary tiling.
- Created unit tests for SFT, RL, and DPO validation functions to verify these restrictions.

FIXES: b/520780393

# Tests

Added unit tests

# Checklist

### DIFF
--- a/src/maxtext/trainers/post_train/dpo/train_dpo.py
+++ b/src/maxtext/trainers/post_train/dpo/train_dpo.py
@@ -182,6 +182,22 @@ def train(mt_config, goodput_recorder=None):
   return trainer, mesh
 
 
+def validate_config(config):
+  """Validates the configuration parameters for DPO training."""
+  if config.optimizer_memory_host_offload:
+    raise ValueError(
+        "optimizer_memory_host_offload=True is not supported on the post-training "
+        "DPO path because the underlying Tunix DPOTrainer does not support "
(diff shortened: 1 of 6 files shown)


src/maxtext/trainers/post_train/dpo/train_dpo.py
```diff
@@ -182,6 +182,22 @@ def train(mt_config, goodput_recorder=None):
   return trainer, mesh
 
 
+def validate_config(config):
```
SurbhiJainUSC, line 185: This validation block can be moved to [pyconfig.py](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/configs/pyconfig.py).

shralex: Note that this is specific to DPO / SFT / RL, while pyconfig is generic. I'm not sure we should move it there.

SurbhiJainUSC (approved):

khatwanimohit (approved):

PR #4125, opened 2026-06-10
Add local_sa_* flags for local attention tuning

aireenmei:
# Description
* Gemma4 uses local-sliding attention and global attention with different head_count, head_dim, and requires different block size for the best perf. 
* Also fix a bug `block_kv_dkv_compute=min(global_block_kv_dkv_compute, query.shape[2])` should use `key.shape`

# Tests
Tested in tuning experiments

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -1098,11 +1098,26 @@ sa_use_fused_bwd_kernel: false
 sa_q_layout: "HEAD_DIM_MINOR"
 sa_k_layout: "HEAD_DIM_MINOR"
 sa_v_layout: "HEAD_DIM_MINOR"
+use_splash_scheduler: false # to use tokamax splash attention scheduler.
+# local_sa_* variants apply to local (sliding window) attention layers;
+# if None, each inherits from the corresponding global sa_* flag.
+local_sa_block_q: None  # inherits from sa_block_q if None
+local_sa_block_kv: None  # inherits from sa_block_kv if None
+local_sa_block_kv_compute: None  # inherits from sa_block_kv_compute if None
+local_sa_block_q_dkv: None  # inherits from sa_block_q_dkv if None
+local_sa_block_kv_dkv: None  # inherits from sa_block_kv_dkv if None
+local_sa_block_kv_dkv_compute: None  # inherits from sa_block_kv_dkv_compute if None
+local_sa_block_q_dq: None  # inherits from sa_block_q_dq if None
+local_sa_block_kv_dq: None  # inherits from sa_block_kv_dq if None
+local_sa_use_fused_bwd_kernel: None  # inherits from sa_use_fused_bwd_kernel if None
+local_sa_q_layout: None  # inherits from sa_q_layout if None
+local_sa_k_layout: None  # inherits from sa_k_layout if None
+local_sa_v_layout: None  # inherits from sa_v_layout if None
+local_use_splash_scheduler: None  # inherits from use_splash_scheduler if None
 use_max_logit_estimate: -1 # -1 means no estimate, any > 0 value will be used as max logit estimate
 cost_estimate_flops_fwd: -1 # -1 means using splash default cost estmiation, any >= 0 value will be used as cost estmiation for splash to overlap for communication (forward)
 cost_estimate_flops_bwd: -1 # -1 means using splash default cost estmiation, any >= 0 value will be used as cost estmiation for splash to overlap for communication (backward)
 dq_reduction_steps: 0 #the number of reduction steps. For now, only 3 or all the kv steps are supported.
-use_splash_scheduler: false # to use tokamax splash attention scheduler.
 ### Determine if we want to use load balance for context parallelism
 context_parallel_load_balance: true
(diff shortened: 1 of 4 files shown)


src/maxtext/configs/base.yml
```diff
@@ -1068,24 +1048,38 @@ ragged_block_size: 256
 
 ### Splash attention block sizes
 # These can be tuned for specific hardware generations, and can be set up to
-# the model's sequence length.
+# the model's sequence length. local_sa_* variants apply to local (sliding window)
+# attention layers; if null, each inherits from the corresponding global sa_* flag.
 sa_block_q: 512
+local_sa_block_q: None  # inherits from sa_block_q if null
```
NuojCheng: can we set them the same default numbers with sa_block_q instead of using None? 

Also, I might prefer move all local_* blocks sizes separate from other blocks, like
sa_block_q
...
sa_block_kv_dq
local_block_q
...
local_sa_block_kv_dq

src/maxtext/configs/types.py
```diff
@@ -2898,7 +2857,33 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+    if self.local_sa_block_kv_dq is None:
+      self.local_sa_block_kv_dq = self.sa_block_kv_dq
+    if self.local_sa_use_fused_bwd_kernel is None:
+      self.local_sa_use_fused_bwd_kernel = self.sa_use_fused_bwd_kernel
+    if self.local_sa_q_layout is None:
+      self.local_sa_q_layout = self.sa_q_layout
+    if self.local_sa_k_layout is None:
+      self.local_sa_k_layout = self.sa_k_layout
+    if self.local_sa_v_layout is None:
+      self.local_sa_v_layout = self.sa_v_layout
```
NuojCheng, line 2979: we can remove this logic if local_* are manually set in base.yml

src/maxtext/layers/attention_op.py
```diff
@@ -79,19 +79,6 @@
 # pylint: disable=line-too-long, g-doc-args, g-doc-return-or-yield, bad-continuation, g-inconsistent-quotes
```
NuojCheng, line 79: @dandragona Could you review the changes in this file making sure the logics are not impacted?

NuojCheng (approved):

src/maxtext/configs/base.yml
aireenmei: The reason I set the default to None is I want to fall back to the existing sa_block_* when user do not set local_sa_* explicitly, this will:
1. avoid any recipe regression. The current recipe uses the same sa_block_* for both local and global
2. unlike gemma4, gemma3 also uses local sliding attention, but the head_count head_dim are the same for local/global, people may prefer to tune them jointly, this "fallback" mechanism makes easy for both cases of tuning separately or jointly.

src/maxtext/layers/attention_op.py
dandragona: I think it LGTM me, all global splash attention parameters seem to be using the same fields they were using before.

src/maxtext/configs/base.yml
aireenmei: And I moved all local_* blocks sizes separate from other blocks as suggested.

src/maxtext/configs/types.py
aireenmei: Discussed in other comments above.

src/maxtext/layers/attention_op.py
```diff
@@ -492,6 +479,35 @@ def __init__(
...
+        self.block_q_dkv = self.config.local_sa_block_q_dkv
+        self.block_kv_dkv = self.config.local_sa_block_kv_dkv
+        self.block_kv_dkv_compute = self.config.local_sa_block_kv_dkv_compute
+        self.block_q_dq = self.config.local_sa_block_q_dq
+        self.block_kv_dq = self.config.local_sa_block_kv_dq
+        self.use_fused_bwd_kernel = self.config.local_sa_use_fused_bwd_kernel
+        self.q_layout = self.config.local_sa_q_layout
+        self.k_layout = self.config.local_sa_k_layout
+        self.v_layout = self.config.local_sa_v_layout
+        self.use_splash_scheduler = self.config.local_use_splash_scheduler
```
gagika, line 497: if  self.config.local_use_splash_scheduler is not set, should we just use self.config.use_splash_scheduler ?

src/maxtext/layers/attention_op.py
```diff
@@ -492,6 +479,35 @@ def __init__(
     self.quant = quant
     self.kv_quant = kv_quant
     self.attention_type = attention_type
+    if self.attention_kernel in ("flash", "autoselected"):
```
gagika: can some be tested in  tests/unit/attention_test.py ?

gagika (approved):
could you have:  a small attention_test.py case asserting unset -> inherits-global and explicit overrides?

it can be a follow up as well.

aireenmei:
> could you have: a small attention_test.py case asserting unset -> inherits-global and explicit overrides?
> 
> it can be a follow up as well.

Good point, added test

PR #4133, opened 2026-06-10
[NNX] Fix multi-token-prediction (MTP) eval under pure_nnx

ecnal-cienet:
## Description

Two NNX-only bugs prevent multi-token-prediction (MTP) eval (`mtp_num_layers>0` with `mtp_eval_target_module`) from running under `pure_nnx=True`. Both already live on `main` and surface only on the NNX path, since `main` still defaults to Linen (`pure_nnx=False`) — invisible by default, but they break the moment NNX is selected. Fixing them now keeps MTP working once the NNX defaults flip lands.

**1. Missing `logical_axis_rules` for `concat_embed` (and `num_activations`).**
The MTP projection kernel (`DenseGeneral`, `2*emb_dim → emb_dim`) is tagged `kernel_axes=("concat_embed", "embed")` in `multi_token_prediction.py`, and the fused gated-MLP `wi` uses `("embed", "num_activations", "mlp")` in `linears.py` — but neither `concat_embed` nor `num_activations` is in `logical_axis_rules`. On the NNX path, `nnx.get_abstract_model` (`train_utils.py`) resolves logical→physical via flax `core.spmd.from_sharding_rules`, which returns **unmapped names verbatim** (no replicate fallback), so the literal name reaches `NamedSharding` and is rejected at state init:

```
ValueError: Resource axis: concat_embed of P('concat_embed', ('fsdp','context','expert')) is not found in mesh
```

Linen never hits this because `linen.spmd.logical_to_mesh_axes` maps unmatched names to `None` (replicated). Fix: add both as replicated (`[]`) in `base.yml`. Note that mapping `concat_embed` to `embed`'s physical axes would be wrong — they are two axes of the *same* kernel, so reusing `fsdp`/`context`/`expert` raises `DuplicateSpecError`.

**2. MTP sown variables leak into the train state and silently zero the MTP loss.**
`mtp_losses` / `mtp_acceptance` are `nnx.Variable` subclasses, **not** `nnx.Intermediate`, so `loss_fn`'s `nnx.pop(model, nnx.Intermediate)` never extracts them. Two consequences:

- They remain on `model.mtp_block` and leak into the returned train state, so `out_shardings` (built from the abstract param state) no longer matches at `p_train_step.lower(...).compile()`:
  ```
  pytree structure error ... at pjit out_shardings[0]['model']['mtp_block']:
    prefix subtree has 1 child key (mtp_layer_1)
    but the full pytree has 5 (losses, mtp_layer_1, mtp_mask, mtp_preds, weights)
  ```
- They are absent from `intermediate_outputs`, so `calculate_mtp_loss` reads `None` and returns `0.0` — **the MTP loss is silently dropped from training** (a correctness bug independent of the crash).

Fix (`train.py` `loss_fn`): pop `mtp_losses` / `mtp_acceptance` into `intermediate_outputs` under their collection names so the loss/acceptance extractors find them, and exclude them from the returned state alongside `nnx.Intermediate`.

## Tests

Validated on TPU — both errors are structural (sharding resolution and `out_shardings` pytree shape), so they reproduce without a TPU:

```bash
PYTHONPATH=src python3 -m maxtext.trainers.pre_train.train src/maxtext/configs/base.yml \
  pure_nnx=True enable_nnx=True pure_nnx_decoder=True \
  model_name=gpt3-52k dataset_type=synthetic steps=6 \
  mtp_num_layers=1 mtp_eval_target_module=1 eval_interval=2 eval_steps=2 \
  attention=dot_product ici_fsdp_parallelism=4 skip_jax_distributed_system=True \
  tokenizer_path=src/maxtext/assets/tokenizers/tokenizer.llama2 \
  base_output_directory=/tmp/mtp run_name=r \
  --xla_force_host_platform_device_count=4
```
- Before fix (failed): https://paste.googleplex.com/6029517163462656
- After fix (passed): https://paste.googleplex.com/6601672909520896

End-to-end TPU run (`gpt3-6b`, the `27_mtp_eval` config) recommended before merge.

## Stats

- **Diff:** +21 / −4 across 2 files — `configs/base.yml` (+6, two `logical_axis_rules`) and `trainers/pre_train/train.py` (+19 / −4, the `loss_fn` pop/filter).

# Checklist

### DIFF
--- a/src/maxtext/trainers/pre_train/train.py
+++ b/src/maxtext/trainers/pre_train/train.py
@@ -45,7 +45,7 @@ from maxtext.utils import elastic_utils
 # Placeholder: internal
 
 # pylint: disable=too-many-positional-arguments
-from maxtext.layers.multi_token_prediction import calculate_mtp_acceptance_rate, calculate_mtp_loss
+from maxtext.layers.multi_token_prediction import calculate_mtp_acceptance_rate, calculate_mtp_loss, mtp_acceptance, mtp_losses
 from maxtext.common import checkpointing, profiler
 from maxtext.common.goodput import (
     GoodputEvent,
@@ -196,6 +196,16 @@ def loss_fn(model, config, data, dropout_rng, params, sparsity_state=None, is_tr
     )
     intermediates = nnx.pop(model, nnx.Intermediate)
     intermediate_outputs = intermediates.to_pure_dict()
+
+    # MTP sows mtp_losses/mtp_acceptance as custom Variable subclasses, not
+    # Intermediate, so the nnx.pop above misses them. Pop them here under their
+    # collection names so calculate_mtp_loss / calculate_mtp_acceptance_rate
+    # find them. Otherwise the MTP loss is silently zeroed. They are also
+    # excluded from the returned state below so they don't leak into
+    # out_shardings.
+    if config.mtp_num_layers > 0:
+      intermediate_outputs["mtp_losses"] = nnx.pop(model, mtp_losses).to_pure_dict()
+      intermediate_outputs["mtp_acceptance"] = nnx.pop(model, mtp_acceptance).to_pure_dict()
 
     if config.num_vocab_tiling > 1:
       hidden_state_key = ("decoder", "hidden_states")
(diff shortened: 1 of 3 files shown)


bvandermoon (approved):

src/maxtext/configs/base.yml
```diff
@@ -554,6 +554,12 @@ logical_axis_rules: [
                       ['embed', ['fsdp', 'tensor_transpose', 'context', 'expert']],
                       ['embed', ['fsdp', 'fsdp_transpose', 'context', 'expert']],
                       ['embed', ['fsdp', 'context', 'expert']],
+                      # Replicate: these are the contraction/fused dims of a kernel that already
+                      # shards 'embed', so they can't reuse those mesh axes. NNX requires an explicit
+                      # rule (Linen replicates unmapped names by default). concat_embed: MTP projection
+                      # input (2*emb); num_activations: fused gated-MLP wi.
+                      ['concat_embed', []],
+                      ['num_activations', []],
```
NuojCheng: It is unclear to me why adding these two logical names? If they are intended to be replicated, then we don't need to introduce them in the logical rule. They are defaulted to be replicated.

NuojCheng (approved):
LGTM except would recommend removing the added logical names

cgarciae (approved):

src/maxtext/configs/base.yml
ecnal-cienet: Fixed. Thank you.

ecnal-cienet:
Due to TPU UT check hanging issue remains unsolved, I will add "pull ready" manually and run TAP Presubmit on critique side first.

parambole (commented):
LGTM

parambole (approved):

PR #4149, opened 2026-06-11
Add reward_functions_path + reward_functions CLI knobs for custom rewards

py4:
## Summary

The post-train RL reward stack in `create_rl_components` is hardcoded to a 3-function list:

```python
reward_fns = [
    make_reward_fn(utils_rl.match_format_exactly),
    make_reward_fn(utils_rl.match_format_approximately),
    make_reward_fn(utils_rl.check_numbers),
]
```

Replacing it requires editing `train_rl.py`. This blocks any user-provided reward (VTC-style partial credit, `math_verify`-based scorer, a custom domain reward) without forking maxtext.

This PR adds two config knobs:
- `reward_functions_path` (str, default `""`): filesystem path to a user Python file
- `reward_functions` (str, default `""`): comma-separated list of function names to import from that file

When **both** are set, the built-in reward list is replaced entirely by the user-provided functions. Each function must accept `(prompts, completions, tmvp_config, **kwargs)` and return `list[float]`. Default behavior is unchanged when either knob is empty.

Reuses the `load_custom_callable` helper (added in #4031) for the file import, so this is a small wiring change on top of an already-reviewed loader.

## Verified on TPU (v6e 8x8)

Ran two live GRPO smoke tests on a Qwen3-1.7B / GSM8K recipe driving the reward through these knobs:

1. **VTC partial-credit reward** (`reward_functions=match_format_exactly` from a user file): full run completed (pre-eval, 3 training steps, checkpoint, post-checks) with exit 0. Confirmed `reward_fns: using 1 custom reward function(s) ['match_format_exactly']`.

2. **Random reward** (returns `uniform(0,1)`, an unforgeable signal): confirmed the custom function is actually called and averaged, not silently defaulted:
   - Called during training: `n=32` batches (4 prompts x 8 generations) scored each step.
   - Called during eval: the eval `mean_reward` exactly equalled the average of the printed random values (pre-RL: mean of 8 values = 0.6065 == reported 0.6065; post-RL: 0.3761 == 0.3761).

## Files

| File | Change |
|---|---|
| `src/maxtext/configs/types.py` | New `reward_functions_path` + `reward_functions` fields on the `RLDataset` mixin |
| `src/maxtext/configs/post_train/rl.yml` | Defaults (`''`) + comment |
| `src/maxtext/trainers/post_train/rl/train_rl.py` | Load + use custom reward fns when both knobs set; else built-in stack |

## Checklist
- [x] Pydantic schema updated in `types.py` (`RLDataset` mixin)
- [x] Backward compatible: default-off when either knob is empty
- [x] No effect on non-RL paths
- [x] Verified end-to-end on TPU with two distinct custom rewards (VTC + random)

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -2056,6 +2056,20 @@ class RLDataset(BaseModel):
           "When set, replaces the built-in dataset processor for custom datasets."
       ),
   )
+  reward_functions_path: str = Field(
+      "",
+      description=(
+          "Optional path to a user Python file containing custom reward functions. "
+          "Used with `reward_functions` to fully replace the built-in reward stack."
+      ),
+  )
+  reward_functions: str = Field(
+      "",
+      description=(
+          "Comma-separated names of reward functions to import from `reward_functions_path`. "
+          "Each function signature: (prompts, completions, tmvp_config, **kwargs) -> list[float]."
+      ),
+  )
 
 
 class RLEvaluation(BaseModel):
--- a/src/maxtext/configs/post_train/rl.yml
+++ b/src/maxtext/configs/post_train/rl.yml
@@ -238,6 +238,13 @@ eval_dataset_name: 'openai/gsm8k'
 # tmvp_config, x) -> dict with keys {prompts, question, answer}. When empty
 # (default), the built-in utils_rl.process_data is used.
 dataset_processor_path: ''
+# Optional: path to a user-provided Python file with custom reward functions,
+# AND a comma-separated list of function names to import from that file.
+# Each function signature: fn(prompts, completions, tmvp_config, **kwargs) -> list[float].
+# When both are set, the built-in [match_format_exactly, match_format_approximately,
+# check_numbers] reward list is REPLACED entirely.
+reward_functions_path: ''
+reward_functions: ''
 train_split: 'train'
 eval_split: 'test'
 hf_name: 'main' # subset of Hugging Face dataset
--- a/src/maxtext/trainers/post_train/rl/train_rl.py
+++ b/src/maxtext/trainers/post_train/rl/train_rl.py
@@ -46,7 +46,7 @@ python3 -m maxtext.trainers.post_train.rl.train_rl src/maxtext/configs/post_trai
 from __future__ import annotations
 import contextlib
 from functools import wraps
-from typing import Any, Optional, Sequence
+from typing import Any, Callable, Optional, Sequence
 
 import datasets
 import grain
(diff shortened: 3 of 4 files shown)


A9isha (approved):
Thanks Pooya!

src/maxtext/configs/types.py
```diff
@@ -2061,6 +2056,20 @@ class RLDataset(BaseModel):
...
       ),
   )
+  reward_functions_path: str = Field(
+      "",
+      description=(
+          "Optional path to a user Python file containing custom reward functions. "
+          "Used with `reward_functions` to fully replace the built-in reward stack."
+      ),
+  )
+  reward_functions: str = Field(
```
SurbhiJainUSC, line 2066: Can we have this default to 
utils_rl.match_format_exactly, utils_rl.match_format_approximately, utils_rl.check_numbers?

py4: @SurbhiJainUSC the default is actually that :

https://github.com/AI-Hypercomputer/maxtext/pull/4149/changes#diff-2ce30b45a1ac12cb32447489f599d3d23f20cabba546fbbefa6f34f27c311086R384

```
return [
      make_reward_fn(utils_rl.match_format_exactly),
      make_reward_fn(utils_rl.match_format_approximately),
      make_reward_fn(utils_rl.check_numbers),
  ]
```
do you think we need a change?

SurbhiJainUSC: I am suggesting:
1. Creating a new module `src/maxtext/trainers/post_train/rl/math_rewards.py` and move `match_format_exactly`, `match_format_approximately`, and `check_numbers` to this module.

2. Set the following in rl.yml:
```
reward_functions_path: 'src/maxtext/trainers/post_train/rl/math_rewards.py'
reward_functions: 'match_format_exactly, match_format_approximately, check_numbers'
```

3. Update `build_reward_fns()` in train_rl.py:
```
def build_reward_fns(trainer_config: Any, make_reward_fn: Callable) -> list:
  rewards_path = getattr(trainer_config, "reward_functions_path", "")
  rewards_names = getattr(trainer_config, "reward_functions", "")
  if custom_rewards_path and custom_rewards_names:
    reward_names = [n.strip() for n in custom_rewards_names.split(",") if n.strip()]
    reward_fns = [make_reward_fn(utils_rl.load_custom_callable(custom_rewards_path, n)) for n in names]
    max_logging.log(f"reward_fns: using {len(reward_fns)} custom reward function(s) {names} from {custom_rewards_path}")
    return reward_fns
max_logging.log("No reward functions provided")
raise Exception
```

py4: Thanks Surbhi, I like the direction. A dedicated rewards module reads better than burying them in utils_rl, and a single path for rewards is cleaner. Before I make the change, a few things to sort out, mostly around defaulting rl.yml to a file path:

1. Path resolution. `load_custom_callable` resolves `reward_functions_path` with `os.path.isfile()` on the raw string, with no anchoring to the package or repo root. So `reward_functions_path: 'src/maxtext/.../math_rewards.py'` is relative to the process CWD. It works when you launch from the repo root, but it breaks for a pip-installed / packaged maxtext (no `src/maxtext/...` under CWD, so it raises "file does not exist"). The knob was meant for user files outside the repo via absolute paths; pointing it at an in-tree file by relative path is the fragile case. If we want the default to load from a file, I'd anchor it, e.g. `os.path.join(os.path.dirname(maxtext.__file__), "trainers/post_train/rl/math_rewards.py")`.

2. The move isn't free. `check_numbers` depends on `extract_answer`, `preprocess_math_string`, `boxed`, `math_verify_func`, and `FALLBACK_ANSWER`, which live in utils_rl and are also used by evaluate_rl.py. So math_rewards.py would still import utils_rl, and it has to use absolute imports since `load_custom_callable` execs it as a package-less module. Moving the three functions also touches rl_utils_test / evaluate_rl_test / math_verify_pool_test.

3. Defaulting both knobs and raising when unset flips this PR's intended behavior (default-off, built-in stack when the knobs are empty), which is what's currently approved.

A middle ground that gets your structure without the path fragility: add `math_rewards.py` and have `build_reward_fns` import those three from it as the in-process default (keep the `''` knobs and the fallback). Same clean module, no CWD dependence, stays backward compatible, and the file-load path stays for real user rewards.

Happy to do it that way in this PR, or keep this one as the minimal wiring and do the module move as a follow-up. Which do you prefer?

SurbhiJainUSC: Agreed to this change in a follow-up PR.

SurbhiJainUSC (approved):

PR #4150, opened 2026-06-11
extract_answer: prefer boxed{N} extraction, fall back to legacy tags

py4:
## Summary

`extract_answer` (used by the built-in `check_numbers` reward) returned the raw `<answer>` content. Modern reasoning models (Qwen3, DeepSeek-R1, etc.) emit the final answer as `\boxed{N}` (the maxtext GSM8K chat template itself asks for `{solution_start_token}\boxed{}{solution_end_token}`). So the extractor returned the literal string `\boxed{42}`, which `math_verify` cannot match against a bare numeric gold like `42`. Result: ~0% accuracy on Qwen3/GSM8K even when the model's numeric answer is correct, a silent scoring failure rather than a model failure.

This makes the extractor consistent with maxtext's own chat template.

## Strategy (priority order)

1. If `<answer>...</answer>` is present, scope to the last block's content; otherwise use the full response.
2. Inside the scope, extract the last `\boxed{N}` via a brace-balanced scan (handles nested LaTeX like `\boxed{\frac{1}{2}}`), with a permissive regex fallback.
3. If no `\boxed` is found, fall back to the legacy `{solution_start_token}...{solution_end_token}` regex, so recipes that emit plain-text answers are unaffected.

Step 3 is what keeps this backward compatible.

## Tests

`tests/post_training/unit/extract_answer_test.py` (10 cases, `cpu_only` + `post_training`):

Boxed extraction:
- inside `<answer>` tags, without tags, nested LaTeX, multiple boxed (last wins), whitespace stripping, negatives, and `<answer>`-tag scoping over a `\boxed` that appears in `<reasoning>`.

Legacy fallback (no `\boxed`):
- plain-text answer inside `<answer>` tags still extracts; last-answer-wins preserved; no-answer returns `FALLBACK_ANSWER`.

All 10 executed and passed against the real function (`Ran 10 tests ... OK`).

## Files

| File | Change |
|---|---|
| `src/maxtext/trainers/post_train/rl/utils_rl.py` | `extract_answer`: boxed extraction + legacy fallback (+40/-4) |
| `tests/post_training/unit/extract_answer_test.py` | new, 10 cases |

## Checklist
- [x] Backward compatible: legacy plain-text answers still extract via the tier-3 fallback
- [x] No effect on non-RL paths
- [x] Unit test covering boxed extraction + legacy fallback, verified passing

### DIFF
--- a/src/maxtext/trainers/post_train/rl/utils_rl.py
+++ b/src/maxtext/trainers/post_train/rl/utils_rl.py
@@ -484,12 +484,53 @@ def check_numbers(
 
 
 def extract_answer(response: str, tmvp_config: Any) -> str:
-  """Function to extract the answer from the text based on the tmvp_config format."""
-  answer_fallback = get_answer_fallback_regex(tmvp_config)
-  # Find the *last* occurrence of the answer tag (most likely the final answer).
-  fallback_matches = answer_fallback.findall(response)
-  extracted_response = fallback_matches[-1].strip() if fallback_matches else FALLBACK_ANSWER
-  return extracted_response
+  """Extract the final numeric answer from the model's response.
+
+  Strategy (priority order):
+    1. Narrow the search scope to the LAST
+       `{solution_start_token}...{solution_end_token}` block (default
+       `<answer>...</answer>`) if present; otherwise use the full response.
+    2. Inside the search scope, find the last `\\boxed{N}` via a brace-
+       balanced scan (handles nested braces in LaTeX). Fall back to a
+       permissive `\\boxed{N}` regex if no balanced match is found.
+    3. If no boxed expression is found, fall back to the same configured
+       solution-tag regex over the full response, for recipes that emit the
+       answer as plain text rather than `\\boxed{N}`.
+
+  Step 1 + 2 are required for modern reasoning models (Qwen3, DeepSeek-R1,
+  etc.) that emit `<think>...</think>\\boxed{N}` or `<answer>\\boxed{N}</answer>`
+  framing. Without `\\boxed` extraction the legacy regex returns the raw
+  `\\boxed{N}` string and math_verify cannot match it against a bare numeric
+  gold. Step 3 keeps the function backward-compatible with recipes that
+  emit plain-text answers inside the configured solution tags.
+
+  The solution tags have a single source of truth: both the scoping (step 1)
+  and the plain-text fallback (step 3) reuse `get_answer_fallback_regex`,
+  built from `tmvp_config.solution_start_token` / `solution_end_token`.
+  """
+  answer_tag_regex = get_answer_fallback_regex(tmvp_config)
+  answer_matches = answer_tag_regex.findall(response)
+  content = answer_matches[-1] if answer_matches else response
+  boxed_matches: list[str] = []
+  stack: list[int] = []
+  for i, ch in enumerate(content):
+    if ch == "{":
+      stack.append(i)
+    elif ch == "}":
+      if not stack:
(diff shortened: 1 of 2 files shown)


src/maxtext/trainers/post_train/rl/utils_rl.py
```diff
@@ -484,12 +484,48 @@ def check_numbers(
...
+       that emit the answer as plain text rather than `\\boxed{N}`.
+
+  Step 1 + 2 are required for modern reasoning models (Qwen3, DeepSeek-R1,
+  etc.) that emit `<think>...</think>\\boxed{N}` or `<answer>\\boxed{N}</answer>`
+  framing. Without `\\boxed` extraction the legacy regex returns the raw
+  `\\boxed{N}` string and math_verify cannot match it against a bare numeric
+  gold. Step 3 keeps the function backward-compatible with recipes that
+  emit plain-text answers inside the configured solution tags.
+  """
+  answer_matches = re.findall(r"<answer>(.*?)</answer>", response, re.DOTALL)
```
SurbhiJainUSC: Isn't this logic same as extracting answer at line 524 by setting `solution_start_token` as `<answer>`: https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/configs/types.py#L2116?

py4: Thanks! Refactored. can you please take a look again?

src/maxtext/trainers/post_train/rl/utils_rl.py
```diff
@@ -484,12 +484,53 @@ def check_numbers(
 
 
 def extract_answer(response: str, tmvp_config: Any) -> str:
```
SurbhiJainUSC, line 486: I think this method can be simplified:
```
def extract_answer():
  answer_tag_regex = get_answer_regex(tmvp_config) # rename get_answer_fallback_regex
  # Extract content from tag or use full response
  answer_matches = answer_tag_regex.findall(response)
  content = answer_matches[-1] if answer_matches else response

  # Find the last \boxed{} match
  boxed_pattern = re.compile(r"\\boxed\s*\{((?:[^{}]|\{[^{}]*\})*)\}")
  boxed_matches = boxed_pattern.findall(content)
  if boxed_matches:
      return boxed_matches[-1].strip()

  # Fallback to simpler regex if \boxed{} structure was malformed
  matches = re.search(r"\\boxed\s*\{?\s*([a-zA-Z0-9\.,\-]+)\s*\}?", content)
  if matches:
      return m.group(1).strip()
  
  # Final fallback
  return answer_matches[-1].strip() if answer_matches else fallback_answer
```
wdyt?

A9isha: Even though the regex approach is cleaner than the stack-based approach, standard Python regular expressions (re) do not support recursive matching, which means they cannot parse arbitrarily nested braces.

The regex provided `(r"\\boxed\s*\{((?:[^{}]|\{[^{}]*\})*)\}")` hardcodes support for exactly one level of nesting. While this works for simple cases like `\boxed{\frac{1}{2}}`, it will fail to extract the full string on deeper LaTeX structures that have two or more levels of nested braces (e.g., `\boxed{a{b{c}d}e}`).

The stack-based for-loop is necessary because it robustly handles nested braces in LaTeX up to any depth by mathematically tracking the opening and closing brackets.

A9isha (approved):

SurbhiJainUSC (approved):

PR #4160, opened 2026-06-13
Add low-memory streaming conversion for unscanned DeepSeek-family checkpoints

### LINKED ISSUE #4071: Kimi-K2.6 unscanned checkpoint converter OOMs on hosts < ~2.5 TB RAM (buffers all 64 shards before write)
# Summary
`convert_deepseek_family_unscanned_ckpt --model_size kimi-k2.6-text` accumulates every dequantized bf16 tensor for all 64 HF safetensors shards in a single in-memory `chkpt_vars` dict before assembling and writing Orbax. For Kimi-K2.6 that's ~2.3 TB of CPU RAM at peak. There's no streaming/low-memory mode, and the documented runbook doesn't mention a host-size requirement.

# Environment
- `maxtext==0.2.2` installed via `pip install maxtext[tpu]` (PyPI), Python 3.12, CPU-only `torch==2.12.0+cpu` (Note: `maxtext[tpu]` pulls `torch==2.12.0+cu130` whose `torch._dynamo` segfaults at import on a CPU-only host — forced reinstall with `--index-url https://download.pytorch.org/whl/cpu` to fix.)
- Host: GCP `n2-standard-32` (32 vCPU, 128 GB RAM), 2 TB pd-balanced disk, Ubuntu 22.04.
- Zone: `us-east5-a` (chosen to be local to v5e TPU quota in this project; us-east5 has no `m1/m2/m3-ultramem` SKUs).
- HF source: `moonshotai/Kimi-K2.6` (96 files, 64 `*.safetensors` shards, 555 GB on disk).

# Repro
```sh
python -m maxtext.checkpoint_conversion.standalone_scripts.convert_deepseek_family_unscanned_ckpt \
    --model_size kimi-k2.6-text \
    --base_model_path /home/$USER/kimi-k2.6-hf \
    --maxtext_model_path gs://<bucket>/k2.6-unscanned
```

# Memory growth observed (RSS via `free -g`)
| After loading shard | RSS used |
| --- | --- |
| 2 | 16 GB |
| 3 | 53 GB |
| 4 | 90 GB |
| 5 | 124 GB |
| 6 | **SIGKILL (rc=137)** |

~35–37 GB per shard, linear. Extrapolating across all 64 shards → ~2.3 TB RSS at peak.

# Root cause (from reading the script)
In [`convert_deepseek_family_unscanned_ckpt.py`](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/checkpoint_conversion/standalone_scripts/convert_deepseek_family_unscanned_ckpt.py), the outer loop over `ckpt_paths` populates a single `chkpt_vars = {}` with every `dequantize_pack_quantized_int4(...)` result before the assembly phase begins. Nothing is freed; nothing is written until the entire model is resident.

# Asks
1. Is there a supported low-memory / streaming mode for the K2.6 (or any K2-Thinking/K2.5) converter? The K2.6 runbook doesn't mention host-memory requirements.
2. If not, what host SKU does the MaxText team use to convert K2.6? `m1-ultramem-160` (3.75 TB) appears necessary but is unavailable in zones where v5e Lite TPU quota lives (`us-east5`, `europe-west4`).
3. Would a per-layer streaming patch be welcome upstream? Happy to PR it if there's interest.

# Workarounds considered
- Cross-region: convert in `us-central1` (has `m1-ultramem-160`), pay ~\$80 GCS egress to land the Orbax in `us-east5`.
- Patch the script to dequant + emit per layer, freeing after each.
- Use the scanned converter then unscan at decode time — does this avoid the same accumulation?

discobot:
# Description

Fixes #4071.

The unscanned converter buffered every dequantized tensor for all shards before
assembly, OOM-ing on hosts with less than ~2.5 TB RAM for Kimi-K2.6.

Two things beyond what the issue establishes: assembly builds a second full fp16 copy
of the model while the buffered dict is still alive, and even with loading fixed, the
default save path (`simulated_cpu_devices_count=16`) re-materializes the whole pytree
as RAM-resident jax.Arrays inside `shard_jax_weights` — so streaming the loads alone
is not enough.

This PR makes tensor loading lazy unconditionally (a header-only index over the shards;
each tensor is read and int4-dequantized exactly once, when assembly consumes it) and
adds an opt-in `--low_memory` flag that stages converted leaves in read-only disk-backed
memmaps under `TMPDIR` and saves via the single-device path. Peak RSS drops from
O(2x model) to O(one tensor); on an 8-shard synthetic kimi-style checkpoint, RSS during
the shard scan goes from +400 MB monotonic growth to flat (+0.5 MB).

Checkpoints are bit-identical in all directions (verified via Orbax restore): old
converter vs new default, new default vs `--low_memory`, and 16-simulated-device
sharded save vs the low-memory single-device save — the saved checkpoint is
topology-independent, so skipping simulated-device sharding in low-memory mode does not
change the artifact.

Adds `tests/unit/convert_deepseek_unscanned_low_memory_test.py` (synthetic int4
multi-shard checkpoint; asserts no tensor reads before assembly, no re-reads, bit-exact
low-memory equivalence, and a save/restore round trip against independently computed
values), ignored in default CI like the other torch-dependent conversion tests, and
documents the flag in the Kimi runbook.

# Tests

- `python3 -m pytest tests/unit/convert_deepseek_unscanned_low_memory_test.py tests/unit/dequantize_pack_quantized_int4_test.py` (9 passed)
- CLI end-to-end on a synthetic checkpoint with and without `--low_memory true`; restored pytrees bit-identical
- `pylint` 10.00/10 and `pyink --pyink-indentation=2 --line-length=122` clean on touched files

# Checklist

This change was developed with assistance from Claude Code.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -10,6 +10,7 @@ addopts =
     --ignore=tests/integration/smoke/train_int8_smoke_test.py
     --ignore=tests/integration/smoke/train_smoke_test.py
     --ignore=tests/integration/smoke/train_using_ragged_dot_smoke_test.py
+    --ignore=tests/unit/convert_deepseek_unscanned_low_memory_test.py
     --ignore=tests/unit/dequantize_mxfp4_test.py
     --ignore=tests/unit/dequantize_pack_quantized_int4_test.py
     --ignore=tests/unit/gemma3_layers_test.py
--- a/src/maxtext/checkpoint_conversion/standalone_scripts/convert_deepseek_family_unscanned_ckpt.py
+++ b/src/maxtext/checkpoint_conversion/standalone_scripts/convert_deepseek_family_unscanned_ckpt.py
@@ -19,6 +19,12 @@ Example cmd:
 python3 -m maxtext.checkpoint_conversion.standalone_scripts.convert_deepseek_family_unscanned_ckpt \
     --base_model_path <path/to/meta/ckpt> \
     --maxtext_model_path <GCS/path/to/save/new/maxtext/ckpt> --model_size deepseek2-16b
+
+Pass --low_memory true to keep peak host RAM at O(one tensor) instead of
+O(2x model size): tensors are read from the safetensors shards one at a time
+and converted weights are staged in disk-backed memmaps (under TMPDIR) until
+the checkpoint is written. Recommended for trillion-parameter checkpoints such
+as kimi-k2.6-text, which otherwise needs ~2.5 TB of host RAM.
 """
 
 # pylint: disable=line-too-long
@@ -30,6 +36,8 @@ import pathlib
 import os
 import gc
 import logging
+import shutil
+import tempfile
 import absl
 
 import numpy as np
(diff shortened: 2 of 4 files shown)


shralex:
How does this approach compare to lazy_load_tensors (which is now true by default) mode in src/maxtext/checkpoint_conversion/to_maxtext.py ?

PR #4166, opened 2026-06-15
Add weight for ragged gather kernel and enable fan out in bwd ragged sort

NuojCheng:
# Description

This PR fixes the usage of `ragged_gather` kernel under `_ring_ragged_unsort_bwd`. As a result, it 

* Removes the `bitcast_add` in backward pass
* Reduce HBM usage

Previously, this kernel was applied in non-fan-out way, since the backward pass the unsort requires weight balance, while the ragged gather kernel did not support. This PR adds this support and enable a fan-out style of this kernel, successfully avoid materializing the full padded-shape tensor in fp32.

FIXES: b/522526813

# Tests

* Performance: 
  * Not use ragged buffer: [17.79 s before](http://xprof.corp.google.com/trace_viewer/chengnuojin-9288426616673894265) [17.80s after](http://xprof.corp.google.com/trace_viewer/chengnuojin-5925128231703123303). HBM usage gets 5GB lower.
  * Use ragged buffer: [17.53 s before](https://xprof.corp.google.com/trace_viewer/ranran-8695996070095900488) [17.53s after](http://xprof.corp.google.com/trace_viewer/chengnuojin-745726469142310301). HBM usage gets 14GB lower.
* Solve ragged buffer size memory issue: EP=8 and EP=16 successful compilation https://paste.googleplex.com/6743176042905600
* EP=8 and FSDP=64 use `ragged_buffer_factor=3.0`: [xprof](http://xprof.corp.google.com/trace_viewer/chengnuojin-1738938222007968418)

# Checklist

### DIFF
--- a/tests/unit/moe_test.py
+++ b/tests/unit/moe_test.py
@@ -618,7 +618,7 @@ class RoutedMoeTest(unittest.TestCase):
           enable_checkpointing=False,
           model_name="mixtral-8x7b",
           override_model_config=True,
-          base_emb_dim=256,
+          base_emb_dim=2048,  # we want emb dim being multiple of 1024 for fully using the kernel
           base_mlp_dim=256,
           base_moe_mlp_dim=256,
           dtype="bfloat16",
--- a/src/maxtext/kernels/ragged/ragged_gather.py
+++ b/src/maxtext/kernels/ragged/ragged_gather.py
@@ -41,6 +41,7 @@ def main_kernel(
     end_ref: jax.Ref,
     in_hbm_ref: jax.Ref,
     indices_hbm_ref: jax.Ref,
+    weights_hbm_ref: jax.Ref,
     # Outputs.
     out_hbm_ref: jax.Ref,
     # Scratch.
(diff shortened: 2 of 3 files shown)


tests/unit/moe_test.py
```diff
@@ -618,7 +618,7 @@ def _build_cfg(use_ragged_sort: bool):
           enable_checkpointing=False,
           model_name="mixtral-8x7b",
           override_model_config=True,
-          base_emb_dim=256,
+          base_emb_dim=2048,  # we want emb dim being multiple of 1024 for fully using the kernel
```
gobbleturk, line 621: this is a bit strange, why does emb dim matter?

NuojCheng: ragged gather reduce kernel hardcoded num of column partitions being 8. Since the partition size should be multiple of 128, it means emb size should be multiple of 1024 for using this kernel. I updated the emb dim in test case for better demonstrating this knowledge, see context in https://github.com/AI-Hypercomputer/maxtext/blob/61c225f6507581f361644630a503f0f4e20f96c3/src/maxtext/kernels/ragged/ragged_gather_reduce.py#L434-L458.

gobbleturk: hmm this is a hard constraint for a lot of models, would this fail fast with a clear error message on these sizes? Ideally we can somehow support

gobbleturk: I would create a bug to track

gobbleturk (approved):

Shuwen-Fang (approved):

tests/unit/moe_test.py
NuojCheng: Agree, added in b/524661949

PR #4167, opened 2026-06-15
Disable JAX compilation cache when `dump_hlo` is True

igorts-git:
# Description

When `dump_hlo=true` is set, MaxText relies on XLA compiler flags to dump HLO files during compilation. However, if `jax_cache_dir` is configured, subsequent runs reuse the JAX compilation cache on disk. Since compilation is bypassed, XLA does not run, and no HLO files are dumped. This silently fails to produce HLOs if the user clears the dump directory between runs.

**Solution:** Automatically bypass the JAX compilation cache initialization in pyconfig.py if `dump_hlo` is True. This forces recompilation and ensures that HLOs are always generated as requested. A warning is logged to inform the user:
```
JAX compilation cache is disabled because dump_hlo is True. HLO dumping requires recompilation.
```

**Alternative considered:** We explored programmatically extracting the final HLO text from the cached executable using compiled.as_text(). While this successfully retrieves the post-optimization HLO even on cache hits, it is unable to recover the full suite of XLA compiler intermediate output files (like before_optimizations.txt, tpu_comp_env.txt, memory usage reports, etc.) which are never generated when compilation is skipped. Forcing recompilation remains the only way to guarantee the complete dump.

FIXES: b/440500124

# Tests

Tested on my TPU VM using a tiny model before and after the fix to confirm that when `dump_hlo==True` XLA compilation gets triggered and HLO is dumped.

# Checklist

### DIFF
--- a/src/maxtext/configs/pyconfig.py
+++ b/src/maxtext/configs/pyconfig.py
@@ -541,9 +541,14 @@ def _initialize_pydantic(argv: list[str] | None = None, **kwargs) -> MaxTextConf
   if "pytest" not in sys.modules:
     max_utils.maybe_initialize_jax_distributed_system(pydantic_kwargs)
   if pydantic_kwargs.get("jax_cache_dir"):
-    from jax.experimental.compilation_cache import compilation_cache  # pylint: disable=import-outside-toplevel
+    if pydantic_kwargs.get("dump_hlo"):
+      max_logging.warning(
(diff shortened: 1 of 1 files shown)


khatwanimohit (approved):

hengtaoguo (approved):

gobbleturk (approved):
This PR implies any hero run should not dump HLO, which is fine I guess since our main HLO use cases are for one-off debugging, but something to note in PR description perhaps, maybe also in types.py documentation

PR #4007, opened 2026-05-28
Fix MoE throughput regression: restore expert sharding + stack remat …

yeandy:
# Description

Fixes https://github.com/AI-Hypercomputer/maxtext/issues/4005

**Changes (Option B''):**

`src/maxtext/configs/base.yml` — 1 line:
```yaml
# Before
['activation_batch_moe', ['data', 'fsdp', 'fsdp_transpose']],
# After
['activation_batch_moe', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
```

`src/maxtext/layers/moe.py` — 9 sites, pattern:
```python
# Before
layer_w0 = adc.checkpoint_name(layer_w0, "moe_mlpwi_0")
# After
layer_w0 = adc.checkpoint_name(adc.checkpoint_name(layer_w0, "mlpwi_0"), "moe_mlpwi_0")
```

# Notes: 
- The `base.yml` change restores expert-axis sharding for MoE activations in the
  einsum/dense_matmul path, fixing the 2.5x regression.
- The stacked `checkpoint_name` pattern ensures `minimal_flash` (which matches `mlpwi_*`)
  correctly checkpoints MoE intermediates via the inner name, while preserving the `moe_*`
  outer name for custom remat policies (PR #3414's original intent).
- No changes to `nnx_decoders.py` or `types.py` required.
- If the maintainers prefer a different option (e.g., Option A to fully revert, or Option C
  for semantic precision), those are also tested and ready.

# Tests

- Benchmarked on MI325X (8 GPU): 120 → 360 TFLOP/s (3x recovery), matching or exceeding
  the `release/v26.4` baseline of 318 TFLOP/s.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [ ] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [ ] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ ] I have run end-to-end tests tests and provided workload links above if applicable.
- [ ] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #4179, opened 2026-06-16
Make MoE dispatch/MLP expert-axis batch sharding configurable (fix Mixtral EP throughput)

gulsumgudukbay:
# Description
PR #4007 added `'expert'` to `activation_batch_moe` to fix a DeepSeek MoE throughput regression. That change is applied to the post-dispatch core MoE activations (`dispatch_axis`/`mlp_axis`) as well, where the tensor's expert dim `E` is already
sharded via `activation_exp`. 
For models with few large experts (e.g. Mixtral), mapping the batch dim onto `'expert'` too double-maps two tensor dims onto one mesh axis, so GSPMD abandons the expert-local layout and falls back to FSDP-style AllGather+ReduceScatter instead of expert-parallel AllToAll, creatiing a large throughput regression under single-node expert parallelism (`ici_expert_parallelism=-1`).
This PR adds a config flag `moe_dispatch_no_expert_sharding` (default `false`) that selects, for the training dispatch/MLP batch axis only, a new `activation_batch_no_exp` rule (`['data','fsdp','fsdp_transpose']`, i.e. without `'expert'`). Mixtral-8x7b uses it as **true**.
This is the per-model knob anticipated in the #4007 review discussion (sharding core MoE activations by the expert physical axes rather than the batch dimension), without changing the default for any other model.

## Behavior
- Default `false` is byte-identical to current `main` for every model.
- Only `mixtral-8x7b` opts in; no other config inherits it.
- For `mixtral-8x7b`, the flag only changes sharding when the `expert` mesh axis size > 1  (expert parallelism active). When `expert` is size 1 (TPU/FSDP-primary, or non-expert parallelism GPU) the two axis rules are identical, so those paths are unaffected.

## Scope (why only the training dispatch/MLP axes)
- `mask_axes` is intentionally left on `activation_batch_moe`: switching it too regresses  Mixtral bs=11 (~10,900 -> ~8,400 tok/s/device, train-step all-to-all 5 -> 3).
- The inference-mode `dispatch_axis`/`mlp_axis` are left unchanged: the inference path is already expert-parallel with `activation_batch_moe` (no FSDP fallback) at both prefill (bs=1: all-to-all=290, all-gather=0) and decode (bs=8: all-to-all=193, all-gather=0), so the regression does not occur there. The change is scoped to the training path.

# Tests

Measured on 1x MI355X node (8 GPUs), JAX 0.9.1, `ici_expert_parallelism=-1`, `capacity_factor>0`:
| Model | bs | tok/s/device (before) | tok/s/device (this PR) | train-step all-to-all |
|---|---|---|---|---|
| Mixtral-8x7b | 11 | ~7,400 | ~10,900 | 0 -> 5 |
| Mixtral-8x7b | 6  | ~7,500 | ~11,400 | 0 -> 5 |
| DeepSeek-v2-lite-16b | 8 | ~17,800 | ~17,800 (unchanged) | 0 (unchanged) |

Mixtral recovers ~47% throughput via restored expert-parallel AllToAll; DeepSeek (flag off by default) is unchanged.

Added a unit regression test `test_moe_dispatch_keeps_expert_on_expert_dim` (`tests/unit/moe_test.py`), parametrized over `mixtral-8x7b`, `mixtral-8x22b`, and `deepseek3-671b`: when `moe_dispatch_no_expert_sharding` is set it asserts the expert dim is sharded by the `expert` mesh axis and the batch dim is not (guarding the expert-parallel dispatch/MLP sharding). Passing locally.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -218,6 +218,10 @@ load_balance_loss_weight: 0.0 # weight for the load balance loss
 use_random_routing: false # whether to use random routing for debug/test purpose
 use_custom_sort_vjp: true # whether to use a custom VJP sort for efficient backward pass processing in sparse matmul
 use_ring_of_experts: false # whether to use ring of experts for sparse matmul expert parallelism
+# If true, peel the 'expert' mesh axis off the MoE dispatch/MLP batch dim so the expert GEMM
+# stays expert-parallel (AllToAll); false keeps 'expert' on the batch dim (activation_batch_moe).
+# Only affects the dense (dense_matmul) MoE path; the sparse (shard_map) path is unaffected.
+moe_dispatch_no_expert_sharding: false
 use_ragged_sort: false # whether to use the Pallas ragged-sort kernels in the MoE permute path; valid both with and
                        # without `use_ring_of_experts` (with EP > 1). When `use_ring_of_experts=True` the kernels run
                        # inside `permute`/`unpermute`; otherwise they run inside `local_permute`/local-unpermute.
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -43,7 +43,7 @@ from maxtext.utils import max_logging
 from maxtext.utils import max_utils
 from maxtext.utils import maxtext_utils
 from maxtext.utils.sharding import create_sharding, maybe_shard_with_logical, maybe_shard_with_pspec
-from maxtext.utils.sharding import logical_to_mesh_axes
+from maxtext.utils.sharding import logical_to_mesh_axes, remove_expert_from_partition_spec
 import numpy as np
 import qwix
 from qwix.contrib.sparsity import sparsity_module
@@ -623,6 +623,18 @@ class RoutedMoE(nnx.Module):
         debug_sharding=self.config.debug_sharding,
         extra_stack_level=1,
     )
+
+  def _maybe_shard_moe_dispatch(self, inputs, logical_axis, peel_expert):
+    """Shard a MoE dispatch/MLP activation. When `peel_expert` is set, drop the 'expert'
+    mesh axis from the batch dim (index 1) so the GEMM stays expert-parallel (AllToAll)
+    instead of double-mapping E and B onto 'expert'. Each logical dim is resolved
+    independently so the shared 'expert' axis is not deduped off the expert dim before
+    the peel."""
+    if not peel_expert:
+      return self._maybe_shard_with_logical(inputs, logical_axis)
+    spec = [None if name is None else self._logical_to_mesh_axes((name,))[0] for name in logical_axis]
+    pspec = remove_expert_from_partition_spec(jax.sharding.PartitionSpec(*spec), dims_to_peel=(1,))
+    return self._maybe_shard_with_pspec(inputs, pspec)
 
   def get_expert_parallelism_size(self):
     # When expert parallelism has more than one physical axes, take product of their shapes
@@ -2173,12 +2185,18 @@ class RoutedMoE(nnx.Module):
 
     if self.config.capacity_factor > 0:
       # token dropping if needed
+      moe_peel_expert = False  # only the training dispatch/MLP path peels 'expert' from the batch dim
       if self.config.model_call_mode != "inference":
         # TODO(b/425930949): remove this pylint by refactoring the logic here.
         dispatch_mask, combine_mask = self.generate_masks(
             top_k_indices, weights  # pylint: disable=undefined-variable,possibly-used-before-assignment
         )
         mask_axes = ("activation_batch_moe", "activation_norm_length_moe", None, None)
+        # Dispatch/MLP are already expert-sharded via "activation_exp". With
+        # moe_dispatch_no_expert_sharding we peel 'expert' off the batch dim of these specs
+        # (see _maybe_shard_moe_dispatch) so the GEMM stays expert-parallel (AllToAll) instead
+        # of double-mapping E and B onto 'expert' (FSDP-style fallback).
+        moe_peel_expert = self.config.moe_dispatch_no_expert_sharding
         dispatch_axis = (
             "activation_exp",
             "activation_batch_moe",
@@ -2283,10 +2301,7 @@ class RoutedMoE(nnx.Module):
                   "activation_embed_moe",
               ),
           )
-        dispatch = self._maybe_shard_with_logical(
-            dispatch,
-            dispatch_axis,
-        )
+        dispatch = self._maybe_shard_moe_dispatch(dispatch, dispatch_axis, moe_peel_expert)
       with jax.named_scope("wi_0"):
         w0_kernel_axes = ("exp", None, "mlp")
         w0_kernel = self.maybe_all_gather_kernel_weight_in_expert_parallelism(w0_kernel, w0_kernel_axes)
@@ -2299,10 +2314,7 @@ class RoutedMoE(nnx.Module):
 
         if self.config.activations_in_float32:
           layer_w0 = layer_w0.astype(jnp.float32)
-        layer_w0 = self._maybe_shard_with_logical(
-            layer_w0,
-            mlp_axis,
-        )
+        layer_w0 = self._maybe_shard_moe_dispatch(layer_w0, mlp_axis, moe_peel_expert)
         layer_w0 = adc.checkpoint_name(adc.checkpoint_name(layer_w0, "mlpwi_0"), "moe_mlpwi_0")
       with jax.named_scope("wi_1"):
         w1_kernel_axes = ("exp", None, "mlp")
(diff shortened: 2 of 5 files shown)


src/maxtext/configs/models/mixtral-8x7b.yml
```diff
@@ -31,3 +31,6 @@ num_experts: 8
 num_experts_per_tok: 2
 rope_max_timescale: 1_000_000
 decoder_block: "mixtral"
+# Few large experts: keep the MoE dispatch/MLP GEMM expert-parallel (AllToAll) rather than
+# letting the batch dim double-map onto 'expert' and fall back to FSDP-style collectives.
+moe_dispatch_no_expert_sharding: true
```
mfrancepillois: This line should probably be added to the `mixtral-8x22b.yml` as well

src/maxtext/layers/moe.py
```diff
@@ -2176,15 +2176,20 @@ def dense_matmul(
...
         mask_axes = ("activation_batch_moe", "activation_norm_length_moe", None, None)
+        # Dispatch/MLP are already expert-sharded via "activation_exp"; with no_expert_sharding the
+        # batch dim drops 'expert' so the GEMM stays expert-parallel (AllToAll) instead of FSDP.
+        moe_dispatch_batch_axis = (
+            "activation_batch_no_exp" if self.config.moe_dispatch_no_expert_sharding else "activation_batch_moe"
+        )
         dispatch_axis = (
             "activation_exp",
-            "activation_batch_moe",
+            moe_dispatch_batch_axis,
```
mfrancepillois: Should `activation_batch_moe` also be replaced with `moe_dispatch_batch_axis` in the `dispatch_axis` and the `mlp_axis` below?

i-chaochen (commented):
can we add test into [moe_test.py](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/moe_test.py) for `activation_batch_no_exp` and `moe_dispatch_batch_axis` ? a pure logical check to make sure batch doesn't steal `expert`

gulsumgudukbay:
> can we add test into [moe_test.py](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/moe_test.py) for `activation_batch_no_exp` and `moe_dispatch_batch_axis` ? a pure logical check to make sure batch doesn't steal `expert`

you beat me! I just pushed it seconds after this comment

src/maxtext/configs/models/mixtral-8x7b.yml
gulsumgudukbay: added, thanks for the reminder. 
mixtral-8x22b has the same geometry (8 experts, moe_mlp_dim=16384, decoder_block: mixtral), so it hits the identical batch-dim/expert double-mapping under EP. Added moe_dispatch_no_expert_sharding: true, and it's covered by the new regression test (...[mixtral-8x22b] passes).

src/maxtext/layers/moe.py
gulsumgudukbay: I benchmarked replacing activation_batch_moe  with moe_dispatch_batch_axis in other sites like dispatch_axis and mlp_axis and it was not beneficial. 

the inference path is already expert-parallel at both prefill bs=1 (jit_prefill_aot: a2a=290, all-gather=0) and decode bs=8 (jit_generate_aot: a2a=193, all-gather=0), no FSDP fallback, unlike unpatched training (a2a=0). So gating the inference dispatch_axis/mlp_axis isn't needed; it only lowers the a2a count without fixing a fallback. I've kept the change scoped to the training path.
I also confirmed earlier that gating training mask_axes regresses throughput ~10.9k to 8.4k, so that's left as-is too.

yeandy:
I tested across many models on MI355 single node. Mixtral regression is resolved without any impact to deepseek-v2-lite. 

Seeing slight regression to llama2-7b, not suspecting that this change is the cause.

tests/unit/moe_test.py
```diff
@@ -1534,5 +1534,38 @@ def test_prefused_vs_sparse_softmax(self):
     self.assertIsNone(bias_updates)
 
 
+@pytest.mark.parametrize("model_name", ["mixtral-8x7b", "mixtral-8x22b", "deepseek3-671b"])
```
yeandy: do we need `deepseek3-671` in the test if `moe_dispatch_no_expert_sharding` is not set to true for this model? we will never hit the `if` conditional, so the test won't fail or anything, but we're also not `assert`ing or testing anything else for models that don't set `moe_dispatch_no_expert_sharding` to true.

gulsumgudukbay: > do we need `deepseek3-671` in the test if `moe_dispatch_no_expert_sharding` is not set to true for this model? we will never hit the `if` conditional, so the test won't fail or anything, but we're also not `assert`ing or testing anything else for models that don't set `moe_dispatch_no_expert_sharding` to true.

you are right deepseek3-671b wasn't asserting anything. Dropped it and now test both states explicitly: flag on (mixtral-8x7b/8x22b) asserts expert is on the expert dim and not the batch dim; a forced flag-off case asserts the default keeps expert on the batch dim. i used Mixtral for the off-case rather than deepseek3-671b since its custom mesh doesn't map activation_exp to expert, so there was nothing stable to assert there.

NuojCheng (approved):

src/maxtext/configs/models/mixtral-8x22b.yml
```diff
@@ -31,3 +31,6 @@ num_experts: 8
 num_experts_per_tok: 2
 rope_max_timescale: 1_000_000
 decoder_block: "mixtral"
+# Few large experts: keep the MoE dispatch/MLP GEMM expert-parallel (AllToAll) rather than
+# letting the batch dim double-map onto 'expert' and fall back to FSDP-style collectives.
+moe_dispatch_no_expert_sharding: true
```
gobbleturk: models.yml files are meant to be core definitions of the model. Implementation details such as sharding are not meant to live in these files. We've made some exceptions for tokenizers since those are a pain to specify and sort of tied to the model

src/maxtext/configs/base.yml
```diff
@@ -218,6 +218,9 @@ load_balance_loss_weight: 0.0 # weight for the load balance loss
 use_random_routing: false # whether to use random routing for debug/test purpose
 use_custom_sort_vjp: true # whether to use a custom VJP sort for efficient backward pass processing in sparse matmul
 use_ring_of_experts: false # whether to use ring of experts for sparse matmul expert parallelism
+# If true, peel the 'expert' mesh axis off the MoE dispatch/MLP batch dim so the expert GEMM
```
gobbleturk, line 221: Can you add this is only relevant for what we call the "dense" case - not our "sparse" shard map case

src/maxtext/configs/models/mixtral-8x22b.yml
gobbleturk: Can you instead specify moe_dispatch_no_expert_sharding=True in the commands you care about?

src/maxtext/utils/sharding.py
```diff
@@ -732,6 +732,32 @@ def _remove_fsdp_from_partition_spec(named_sharding):
   return jax.tree.map(_remove_fsdp_from_partition_spec, sharding_tree)
 
 
+def remove_expert_from_partition_spec(pspec, dims_to_peel):
```
gobbleturk, line 735: this function makes me sad, but I trust the team with help from @NuojCheng can at some point remove the need for these

can keep this function in this PR

In general we should have separate rules for the moe layer vs not the moe layer, which we almost have succesfully refactored into. Any rule named "no_exp" is sad, it should be named logically (as opposed to physically) because it is a logical rule

tests/unit/moe_test.py
```diff
@@ -1534,5 +1535,53 @@ def test_prefused_vs_sparse_softmax(self):
     self.assertIsNone(bias_updates)
 
 
+@pytest.mark.parametrize(
+    "model_name,override_flag",
+    [
+        ("mixtral-8x7b", None),  # opts in via model config (flag on)
```
gobbleturk: can set your new flag in these tests instead of defining them in the model

Shuwen-Fang (approved):

tests/unit/moe_test.py
gulsumgudukbay: addressed in new commit https://github.com/AI-Hypercomputer/maxtext/pull/4179/commits/5709de1951960f63513dde17f21bb10435f80bcb

src/maxtext/configs/base.yml
gulsumgudukbay: Addressed in https://github.com/AI-Hypercomputer/maxtext/pull/4179/commits/5709de1951960f63513dde17f21bb10435f80bcb

src/maxtext/configs/models/mixtral-8x22b.yml
gulsumgudukbay: addressed in https://github.com/AI-Hypercomputer/maxtext/pull/4179/commits/5709de1951960f63513dde17f21bb10435f80bcb

gobbleturk (approved):

PR #4187, opened 2026-06-17
Add ragged sort kernel fallback mechanism and version guard

NuojCheng:
# Description

This PR

* Introduces two flags, `ragged_gather_fallback` and `ragged_gather_reduce_fallback`. When they are true, a fully JAX, non-ragged version  instead of the kernel version gonna be used
* Add version guard protecting ragged gather kernel.

# Tests

Based on xprofs, both flags work effectively.

## Fall back ragged gather reduce only
```
smoke_train model_name=deepseek2-16b ici_expert_parallelism=4 per_device_batch_size=1 max_target_length=4096 use_random_routing=true use_ring_of_experts=true use_ragged_sort=true ragged_gather_reduce_fallback=true debug_sharding=false profiler=xplane ragged_gather_fallback=false enable_tpu_profiling_options=true
```
[xprof](http://xprof.corp.google.com/trace_viewer/chengnuojin-13733567731680657379)

## Fall back ragged gather only
```
smoke_train model_name=deepseek2-16b ici_expert_parallelism=4 per_device_batch_size=1 max_target_length=4096 use_random_routing=true use_ring_of_experts=true use_ragged_sort=true ragged_gather_reduce_fallback=false debug_sharding=false profiler=xplane ragged_gather_fallback=true enable_tpu_profiling_options=true
```
[xprof](http://xprof.corp.google.com/trace_viewer/chengnuojin-11626661074307699646)

## Fall back both kernels
```
smoke_train model_name=deepseek2-16b ici_expert_parallelism=4 per_device_batch_size=1 max_target_length=4096 use_random_routing=true use_ring_of_experts=true use_ragged_sort=true ragged_gather_reduce_fallback=true debug_sharding=false profiler=xplane ragged_gather_fallback=true enable_tpu_profiling_options=true
```
[xprof](http://xprof.corp.google.com/trace_viewer/chengnuojin-561626777463547215)

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -222,6 +222,10 @@ use_ragged_sort: false # whether to use the Pallas ragged-sort kernels in the Mo
                        # without `use_ring_of_experts` (with EP > 1). When `use_ring_of_experts=True` the kernels run
                        # inside `permute`/`unpermute`; otherwise they run inside `local_permute`/local-unpermute.
 use_gather_mosaic_kernel: false # whether to use a custom mosaic kernel for token gather ops
+ragged_gather_fallback: false # when true, unconditionally use the JAX reference implementation instead of the
+                              # ragged gather SparseCore kernel. When false (default), use the SparseCore kernel.
+ragged_gather_reduce_fallback: false # when true, unconditionally use the JAX reference implementation instead of the
+                                     # ragged gather reduce SparseCore kernel. When false (default), use the SparseCore kernel.
 # tunable tiling dimensions used for mlp gmm
 # megablox/jax ragged dot - supports forward pass only (6 configs: `wi_tile_fwd...` and `wo_tile_fwd_...`)
 # tokamax ragged dot - supports all 18 configs
--- a/src/maxtext/kernels/ragged/ragged_gather.py
+++ b/src/maxtext/kernels/ragged/ragged_gather.py
@@ -16,7 +16,6 @@
 # Source from https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/sparse_core/ragged_gather.py
 
 import functools
-
 import jax
 import jax.numpy as jnp
 from jax.experimental import pallas as pl
@@ -24,15 +23,23 @@ from jax.experimental.pallas import tpu as pltpu
 from jax.experimental.pallas import tpu_sc as plsc
 from packaging.version import Version
 
-
 # JAX <= 0.10.0 used `out_shape`/`scratch_shapes` kwargs for `pl.kernel`; later
 # versions renamed them to `out_type`/`scratch_types`.
 if Version(jax.__version__) <= Version("0.10.0"):
   _OUT_KW = "out_shape"
   _SCRATCH_KW = "scratch_shapes"
+  _COMPILER_PARAMS = {
+      "use_tc_tiling_on_sc": True,
+      "disable_bounds_checks": True,
+  }
 else:
   _OUT_KW = "out_type"
   _SCRATCH_KW = "scratch_types"
+  _COMPILER_PARAMS = {
+      "use_tc_tiling_on_sc": True,
+      "disable_bounds_checks": True,
+      "needs_layout_passes": False,
+  }
 
 
 def main_kernel(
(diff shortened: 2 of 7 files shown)


src/maxtext/kernels/ragged/ragged_gather.py
```diff
@@ -331,12 +355,9 @@ def ragged_gather(
...
   sc_info = pltpu.get_tpu_info().sparse_core
-  if sc_info is None:
-    # Sparse core is not available. Fallback to regular gather.
-    out = x[indices]
-    if has_weights:
-      out = out * weights[:, None]
-    return out
+  if sc_info is None or enforce_fallback:
+    # Sparse core is not available or fallback is enforced. Use JAX reference.
+    return _fallback_implementation(x, indices, weights, has_weights)
```
gobbleturk, line 360: this is a bit surprising to see a non ragged fallback inside of a function called ragged_gather - should the decision to use ragged_gather vs a fallback happen at a higher level function (e.g. somewhere higher in the call stack?)

tests/unit/moe_test.py
```diff
@@ -723,6 +731,13 @@ def test_ragged_sort_loss_and_grad_ring_of_experts(self):
   def test_ragged_sort_loss_and_grad_ring_of_experts_ragged_buffer(self):
     self._run_ragged_sort_loss_and_grad(use_ring_of_experts=True, ragged_buffer_factor=1.5)
 
+  @pytest.mark.tpu_only
+  @pytest.mark.skip_on_tpu7x
```
gobbleturk, line 735: why skip on tpu7x?

src/maxtext/kernels/ragged/ragged_gather.py
```diff
@@ -264,6 +271,19 @@ def dma_write_loop(col_vmem_start):
   inner_kernel()
 
 
+def _fallback_implementation(
+    x: jax.Array,
+    indices: jax.Array,
+    weights: jax.Array | None = None,
+    has_weights: bool = False,
+) -> jax.Array:
+  """Fallback to JAX implementation for ragged gather."""
```
gobbleturk: is this a ragged implementation? Doesn't this grow with the full buffer size or no?

src/maxtext/configs/base.yml
```diff
@@ -222,6 +222,10 @@ use_ragged_sort: false # whether to use the Pallas ragged-sort kernels in the Mo
                        # without `use_ring_of_experts` (with EP > 1). When `use_ring_of_experts=True` the kernels run
                        # inside `permute`/`unpermute`; otherwise they run inside `local_permute`/local-unpermute.
 use_gather_mosaic_kernel: false # whether to use a custom mosaic kernel for token gather ops
+ragged_gather_fallback: false # when true, unconditionally use the JAX reference implementation instead of the
```
Shuwen-Fang, line 225: could this just be:
ragged_gather: true --> use ragged kernel
ragged_gather:false --> use non ragged fallback

src/maxtext/kernels/ragged/ragged_gather.py
NuojCheng: The new `ragged_gather_fallback` and `ragged_gather_reduce_fallback` flags control kernel-level fallback logic. We avoid using these flags directly in `moe.py` because we may need the native JAX implementation for the forward pass and the custom kernel for the backward pass. While we could theoretically use them in `ragged_sort.py`, a fallback mechanism already exists in the primary kernel wrapper. We leverage that existing structure instead of introducing redundant if-else conditions.

tests/unit/moe_test.py
NuojCheng: I think there were some issues using ragged sort on bloom. @darisoy do we have a buganizer tracking this?

src/maxtext/kernels/ragged/ragged_gather.py
NuojCheng: There are only two options:
1. Jax sort, no raggedness
2. sparse core kernels, ragged
Update the comments to better reflect JAX is non-ragged.

src/maxtext/configs/base.yml
NuojCheng: There are two operations:
1: ragged permute (fwd: ragged gather; bwd: ragged gather reduce)
2. ragged unpermute (fwd: ragged gather reduce, bwd: ragged gather)

Technically we can introduce 4 flags, controlling whether we want kernel version/JAX version respectively. However, it is probably not necessary. Instead, we have one flag `use_ragged_sort` indicating whether we want to use kernel version for all, and `ragged_gather_fallback` and `ragged_gather_reduce_fallback` controlling kernels respectively. 

I think you are suggesting re-naming these two flags, with opposite meanings? I don't have strong opinion on this, but I can update if you think they are much better...

Shuwen-Fang: It is a nit naming suggestion, but I think naming it to ragged_sort: false can imply that the fallback mechanism is not ragged.

Shuwen-Fang (approved):

src/maxtext/kernels/ragged/ragged_gather.py
gobbleturk: I see the fwd and bwd coupling makes this challenging to separate cleanly, thank you for explaining

gobbleturk (approved):

PR #4205, opened 2026-06-19
Fix logical sharding resolution in NNX

xibinliu:
# Description

In pure NNX training runs, model variables retrieve physical PartitionSpecs via `get_nnx_named_sharding_with_scan_axis` in `maxtext_utils.py`. Previously, this helper used Flax core SPMD's `from_sharding_rules` to map logical names to physical axes. However, `from_sharding_rules` resolves rules by converting the rules list into a dictionary (last-write-wins). This caused fallback rules sharing the same logical name (e.g. 'embed') to overwrite preceding specific rules, dropping essential axes like `fsdp_transpose` and leading to unsharded parameter percentage assertion errors.

Additionally, resolving specifications independently for each dimension without tracking assigned axes could bind a single physical axis (like `fsdp_transpose`) to multiple positional dimensions of a tensor, causing `DuplicateSpecError`.

To fix this:
1. Replaced `from_sharding_rules` with a Rules-first resolution loop that matches rules sequentially (first-match-wins), matching Flax Linen's mapping behavior.
2. Implemented an `assigned_axes` tracker within the loop to ensure physical mesh axes are bound to at most one dimension per tensor.
3. Added unit tests covering sequential matching (first-match-wins) and duplicate physical axis prevention during resolution.

# Tests

[Log with Gemma3-12B (2x v6e-256)](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22tpu-prod-env-one-vm%22%0Aresource.labels.location%3D%22southamerica-west1%22%0Aresource.labels.cluster_name%3D%22bodaborg-v6e-256-lcscld-c%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name:%22xib-gemma3-12b-2-062300-0z2-slice-job-0-0-%22%0Aseverity%3E%3DDEFAULT%0ASEARCH%2528%22completed%20step%22%2529;storageScope=project;cursorTimestamp=2026-06-23T01:00:59.634697213Z;startTime=2026-06-23T00:50:55.612Z;endTime=2026-06-23T01:10:55.612Z?project=tpu-prod-env-one-vm)

# Checklist

### DIFF
--- a/src/maxtext/utils/maxtext_utils.py
+++ b/src/maxtext/utils/maxtext_utils.py
@@ -20,7 +20,6 @@ import os
 from typing import Sequence
 
 from flax import nnx, linen as nn
-from flax.core.spmd import composite_rules, from_sharding_rules, get_logical_axis_rules
 from flax.linen import partitioning as nn_partitioning
 from flax.training.train_state import TrainState
 
@@ -28,7 +27,7 @@ import numpy as np
 
 import jax
 import jax.numpy as jnp
-from jax.sharding import AxisType, Mesh, NamedSharding, PartitionSpec
+from jax.sharding import AxisType, Mesh, NamedSharding
 from jax.experimental import mesh_utils
 from jax.experimental.serialize_executable import deserialize_and_load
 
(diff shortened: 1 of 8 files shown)


src/maxtext/utils/maxtext_utils.py
```diff
@@ -1612,6 +1612,43 @@ def move(path, x):
   )
 
 
+def _resolve_logical_sharding(out_sharding, context_rules, local_rules) -> list:
```
NuojCheng: could you move this function to `utils/sharding.py`? The goal is to move all sharding related util functions to this file.

tests/unit/maxtext_utils_test.py
```diff
@@ -1690,6 +1691,60 @@ def test_string_out_sharding_is_wrapped_into_tuple(self):
     # The single string 'fsdp' is turned into a list, and 'layers' is prepended.
     self.assertEqual(result_sharding.spec, PartitionSpec("layers", "fsdp"))
 
+  def test_sequential_matching_first_match_wins(self):
+    """Multiple rules for the same logical axis are matched sequentially, first-match-wins."""
+    # We define rules for 'embed' mapping to 'fsdp' (specific) then 'layers' (fallback)
+    rules = (
+        ("embed", "fsdp"),
+        ("embed", "layers"),
```
NuojCheng: layers is not a physical axis name? maybe for something else, expert?

tests/unit/maxtext_utils_test.py
```diff
@@ -1690,6 +1691,60 @@ def test_string_out_sharding_is_wrapped_into_tuple(self):
...
+    )
+    with nn_partitioning.axis_rules(rules):
+      with jax.set_mesh(self.mesh):
```
NuojCheng: with jax.set_mesh(self.mesh), nn_partitioning.axis_rules(rules):

tests/unit/maxtext_utils_test.py
```diff
@@ -1690,6 +1691,60 @@ def test_string_out_sharding_is_wrapped_into_tuple(self):
...
+      self.assertEqual(result_sharding.spec, PartitionSpec("fsdp"))
+
+  def test_deduplicates_assigned_physical_axes(self):
+    """Physical axes already bound to a dimension cannot be bound to another dimension."""
+    # Define rules where 'embed' maps to ('fsdp', 'layers') and 'mlp' maps to 'fsdp'.
+    # Because 'embed' is defined first, it binds 'fsdp'.
+    # When matching 'mlp', 'fsdp' is already bound, so it is skipped (unassigned/None).
+    rules = (
+        ("embed", ("fsdp", "layers")),
+        ("mlp", "fsdp"),
```
NuojCheng: IMO this is an error caused by the rule. We should throw an error if there is a conflict instead of silently solving it.

NuojCheng (commented):
Overall I like this feature. I was concerned that this function would silently hide some errors that should be explicit raised, e.g. logical rule conflicting part. I agree we should improve the `from_sharding_rules` if it simply treats logical rule as dictionary.

tests/unit/maxtext_utils_test.py
xibinliu: In [base.yaml](https://github.com/AI-Hypercomputer/maxtext/blob/0b9f60450397378a445417e165af7cc91a1abcdb/src/maxtext/configs/base.yml#L552-L557): both `mlp` and `embed` have the `"fsdp_transpose"` physical match:
```
                      ['mlp', ['fsdp_transpose', 'tensor', 'tensor_sequence', 'autoregressive']],
                      ...
                      ['embed', ['fsdp', 'fsdp_transpose', 'tensor_transpose', 'context', 'expert']],
```
And this is used in [MLP layer](https://github.com/AI-Hypercomputer/maxtext/blob/dd9c26764fafe6131e8ad035842ec19ffba05795/src/maxtext/layers/linears.py#L436):
```
            kernel_axes=("embed", "mlp"),
```

I guess this is not a config issue.

Linen resolved this by calling `remove_size_one_mesh_axis` on the physical specs. We will now apply the same to NNX.

src/maxtext/utils/maxtext_utils.py
xibinliu: Moved the `get_nnx_named_sharding_with_scan_axis()` to sharding.py, and moved the corresponding unit tests accordingly.

tests/unit/maxtext_utils_test.py
xibinliu: Changed to "stage", and use "layers" only as logical name.

tests/unit/maxtext_utils_test.py
xibinliu: Modified.

tests/unit/maxtext_utils_test.py
NuojCheng: I would expect 'embed' downgraded to the next item, https://github.com/AI-Hypercomputer/maxtext/blob/0b9f60450397378a445417e165af7cc91a1abcdb/src/maxtext/configs/base.yml#L558, so the conflict no longer exists. With your function that auto move to next list, I think the error you mentioned should be solved?

xibinliu: You are right. Checked the Linen implementation, and it does fall to the next item.
I modified the NNX code to re-use the linen nn.logical_to_mesh_axes(), instead of implementing the same func for NNX.
Now they should have the same behaviour.

xibinliu:
> Overall I like this feature. I was concerned that this function would silently hide some errors that should be explicit raised, e.g. logical rule conflicting part. I agree we should improve the `from_sharding_rules` if it simply treats logical rule as dictionary.

With pure NNX implementation of MaxText models, are we still allowed to import linen functions? If the answer is yes, then we can continue to re-use Linen implementation. I think we can decide during the Linen removal phase. (PR#12)

tests/unit/maxtext_utils_test.py
```diff
@@ -26,6 +26,7 @@
 
 from flax import linen as nn
 from flax import nnx
+# Removed unused partitioning import
```
NuojCheng: remove this line?

tests/unit/sharding_nnx_test.py
```diff
@@ -108,24 +99,19 @@ def test_extract_param_only_skips_non_param_variables(self):
...
     cfg = _Cfg(shard_optimizer_over_data=True)
-    state_mesh_shardings = _build_state_mesh_shardings(
-        self.model, optax.adam(1e-3)
-    )
+    state_mesh_shardings = _build_state_mesh_shardings(self.model, optax.adam(1e-3))
 
     # Mutate the optimizer mu leaves in place so the function picks up a distinct PartitionSpec.
     mesh = Mesh(
-        np.array(jax.local_devices()[:1]).reshape(1, 1), ("data", "model")
+        devices=np.array(jax.local_devices()[:1]).reshape(1, 1),
```
NuojCheng: For CI tests we should have more than 1 device, right? Can we make the mesh 2x2?

I am concerned that a 1x1 mesh cannot test the functionality when using remove_size_one_axis.

tests/unit/sharding_nnx_test.py
```diff
@@ -172,24 +148,254 @@ def test_chained_optimizer_recursion_finds_adam_mu(self):
...
+  """Unit tests for nnx_construct_named_sharding covering every branch.
+
+  The helper resolves a NamedSharding for each NNX Variable inside an nnx.State and
+  — unlike flax.nnx.spmd.get_var_pspec — also inserts the `nnx.PARTITION_NAME` axis at
+  `param_scan_axis` when scanned-layers metadata is present.
+  """
+
+  def setUp(self):
+    # Mesh needs to contain every axis name the tests reference in partition specs.
+    self.mesh = Mesh(np.array(jax.local_devices()[:1]).reshape(1, 1, 1), ("fsdp", "stage", "expert"))  # pylint: disable=too-many-function-args
```
NuojCheng: same concern as above, can we make mesh 2x2?

NuojCheng (approved):
Thank you Xibin! Could you try setting mesh as 2x2 for CI tests? If it does not work then 1x1 mesh should be fine.

tests/unit/maxtext_utils_test.py
xibinliu: Done!

tests/unit/sharding_nnx_test.py
xibinliu: Use this for the mesh in unit tests: 
```
def _create_2d_test_mesh(axis_names=("data", "model")):
  devices = jax.local_devices()
  num_devices = len(devices)
  if num_devices >= 4:
    mesh_devices = np.array(devices[:4]).reshape(2, 2)
  elif num_devices >= 2:
    mesh_devices = np.array(devices[:2]).reshape(2, 1)
  else:
    mesh_devices = np.array(devices[:1]).reshape(1, 1)
  return Mesh(devices=mesh_devices, axis_names=axis_names)
```

tests/unit/sharding_nnx_test.py
xibinliu: same as above.

SurbhiJainUSC (approved):

PR #4228, opened 2026-06-22
Add cost estimates for ragged sort kernels

NuojCheng:
# Description

Add cost estimates for ragged gather and ragged gather reduce kernels, for better XLA compiler scheduling.

This PR also adds four flags for ragged kernel cost estimations:
* ragged_gather_cost_estimate_flops
* ragged_gather_reduce_cost_estimate_flops
* ragged_gather_cost_estimate_bytes_accessed
* ragged_gather_reduce_cost_estimate_bytes_accessed

Their default values are -1 and any positive input value will replace the flop/bytes estimations.

FIXES: b/525538961

# Tests

* Before (xids/265679573): [step time 15.53s](http://xprof/?session_id=chengnuojin-12833556763352028825)
* After (xids/265674304): [step time 15.48s](http://xprof/?session_id=chengnuojin-7985950153437521179)

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -226,6 +226,10 @@ ragged_gather_fallback: false # when true, unconditionally use the JAX reference
                               # ragged gather SparseCore kernel. When false (default), use the SparseCore kernel.
 ragged_gather_reduce_fallback: false # when true, unconditionally use the JAX reference implementation instead of the
                                      # ragged gather reduce SparseCore kernel. When false (default), use the SparseCore kernel.
+ragged_gather_cost_estimate_flops: -1 # -1 means auto-compute, any > 0 value overrides the flop cost estimate for the ragged gather kernel
+ragged_gather_reduce_cost_estimate_flops: -1 # -1 means auto-compute, any > 0 value overrides the flop cost estimate for the ragged gather reduce kernel
+ragged_gather_cost_estimate_bytes_accessed: -1 # -1 means auto-compute, any > 0 value overrides the bytes_accessed cost estimate for the ragged gather kernel
+ragged_gather_reduce_cost_estimate_bytes_accessed: -1 # -1 means auto-compute, any > 0 value overrides the bytes_accessed cost estimate for the ragged gather reduce kernel
 # tunable tiling dimensions used for mlp gmm
 # megablox/jax ragged dot - supports forward pass only (6 configs: `wi_tile_fwd...` and `wo_tile_fwd_...`)
 # tokamax ragged dot - supports all 18 configs
--- a/src/maxtext/kernels/ragged/ragged_gather.py
+++ b/src/maxtext/kernels/ragged/ragged_gather.py
@@ -271,6 +271,59 @@ def main_kernel(
   inner_kernel()
 
 
+def get_cost_estimate(
+    out_size: int,
+    hidden_size: int,
+    dtype_bytes: int,
+    has_weights: bool,
+    flops_override: int = -1,
+    bytes_accessed_override: int = -1,
+) -> pl.CostEstimate:
+  """Returns a cost estimate for the ragged gather kernel.
+
+  The ragged gather is primarily a data-movement kernel: it gathers rows from
+  an input table according to an index array.  When ``has_weights`` is True an
+  additional element-wise multiply is performed.
+
+  Args:
+    out_size: Number of output rows (after padding).
+    hidden_size: Number of columns in the input / output.
+    dtype_bytes: Size of one element in bytes (e.g. 2 for bf16, 4 for f32).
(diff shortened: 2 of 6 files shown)


src/maxtext/kernels/ragged/ragged_gather.py
```diff
@@ -271,6 +271,47 @@ def dma_write_loop(col_vmem_start):
   inner_kernel()
 
 
+def get_cost_estimate(
```
gobbleturk, line 274: can we allow users to plumb in their own cost estimate and override this? The kernels are not very efficient so estimating cost from theory is not reflective of the time taken

gobbleturk: e.g. just like for splash there is fwd and bwd cost estimate in base.yml

Shuwen-Fang (approved):

src/maxtext/configs/base.yml
```diff
@@ -230,6 +226,8 @@ ragged_gather_fallback: false # when true, unconditionally use the JAX reference
                               # ragged gather SparseCore kernel. When false (default), use the SparseCore kernel.
 ragged_gather_reduce_fallback: false # when true, unconditionally use the JAX reference implementation instead of the
                                      # ragged gather reduce SparseCore kernel. When false (default), use the SparseCore kernel.
+ragged_gather_cost_estimate_flops: -1 # -1 means auto-compute, any > 0 value overrides the flop cost estimate for the ragged gather kernel
```
gobbleturk, line 229: Is there benefit to exposing four separate cost estimates (fwd dispatch, fwd combine, bwd dispatch, bwd combine?) I guess there are only two main kernels - gather and gather reduce, so only need these two?

src/maxtext/kernels/ragged/ragged_gather.py
```diff
@@ -271,6 +271,53 @@ def dma_write_loop(col_vmem_start):
...
+    out_size: int,
+    hidden_size: int,
+    dtype_bytes: int,
+    has_weights: bool,
+    flops_override: int = -1,
```
gobbleturk, line 279: nit: I would just name it cost_estimate instead of flops_override, with the same -1 gets set to a nice default behavior (identical code just variable name)

gobbleturk: oh nvm I see this just overrides the flops, maybe this makes sense. I forget to consider cost estimate has both bytes and flops... I'm not sure which is a more useful tuning knob for e2e performance (maybe even both...), I think this is fine for now. The end goal is to help XLA schedule collectives

src/maxtext/kernels/ragged/ragged_gather_reduce.py
```diff
@@ -49,6 +49,55 @@ def _align_to(a, b):
...
+  Returns:
+    A ``pl.CostEstimate`` suitable for XLA scheduling.
+  """
+  # Flops:
+  #   - one multiply per element for weighting: padded_input_size * aligned_hidden_size
+  #   - one add per element for reduction:       padded_input_size * aligned_hidden_size
+  if flops_override > 0:
+    flops = flops_override
+  else:
+    flops = 2 * padded_input_size * aligned_hidden_size
```
gobbleturk, line 84: I'm not sure there are any flops here actually - do flops of a cost_estimate refer to MXU flops? The additions performed here won't happen on the MXU. The flops here are tiny anyway unless the user is setting a large override to tune scheduling, so maybe this is fine, but we should default to 0 if flops of this cost_estimate refers to MXU flops

gobbleturk (approved):

src/maxtext/kernels/ragged/ragged_gather.py
gobbleturk: it makes more sense to me that we can override the bytes_accessed since that is the main cost here, but I don't understand exactly how these affect our goal of helping tune XLA schedules

src/maxtext/configs/base.yml
NuojCheng: I think two is enough, though there is slight diff between using weight/not using weight for ragged gather kernel

src/maxtext/kernels/ragged/ragged_gather.py
NuojCheng: Added both now

src/maxtext/kernels/ragged/ragged_gather_reduce.py
NuojCheng: The flops are for tokens/gradients accumulation... Agree they are not very useful..

src/maxtext/kernels/ragged/ragged_gather.py
NuojCheng: done

PR #4233, opened 2026-06-22
Enabled auxilillary loss free load balancing and sequence wise load b…

dipakg-lang:
## Description
    Implement DSV4 auxillary loss free and sequence-wise load balancing via pure NNX MoEBiasVar

This commit migrates the DeepSeek V4 auxiliary-loss-free routing bias to a pure nnx.Variable (MoEBiasVar), automatically isolating it from the optimizer and standard sequence-wise gradients without requiring the jax.lax.stop_gradient hacks or Optax global masking.
    
DeepSeek V3 backward compatibility remains completely untouched and functional via the legacy nnx.Param paths.

FIXES: b/509933890
FIXES: b/521990776

## Tests
Tested by running training loop with new tiny Deeepseek V4 model added as part of the commit,
here are the logs for testing and commands used for this :

```
export JAX_PLATFORM_NAME=cpu & export XLA_FLAGS=--xla_force_host_platform_device_count=8 & python3 -m maxtext.trainers.pre_train.train src/maxtext/configs/base.yml override_model_config=True model_name=deepseek4-tiny enable_checkpointing=False base_output_directory=/tmp/maxtext_output/ dataset_type=synthetic hardware=cpu skip_jax_distributed_system=True attention=dot_product per_device_batch_size=1 steps=100 max_target_length=256 async_checkpointing=false dtype=bfloat16 weight_dtype=bfloat16 megablox=False sparse_matmul=False ici_expert_parallelism=-1 sharding_tolerance=1.0 ici_fsdp_parallelism=1 indexer_topk=16 routed_bias=True load_balance_loss_weight=0.0001 routed_bias_update_rate=0.001
```

    
    To prove the variance reduction and model convergence, the DeepSeek-V4-Flash (284B) model was trained for 300 steps on a Ironwood cluster. The following analysis was performed:
    
    * Run A (Baseline): Variance at Step 300 with NO load balancing active (routed_bias_update_rate=0.0, load_balance_loss_weight=0.0). Shows natural degradation and expert collapse.
    * Run B (With Load Balancing): Variance at Step 300 with FULL load balancing active (routed_bias_update_rate=0.001, load_balance_loss_weight=0.0001). Shows healthy token distribution and stable loss curve comparable or better than the PR-4497 base architecture.
    
    === DeepSeek V4 Load Balancing Variance Analysis (Step 300) ===
    
    | Layer Index | Routing Type | Baseline (No LB) | With Load Balancing | Improvement |
    |-------------|--------------|------------------|---------------------|-------------|
    |           0 | Hash Routed  |        587520.00 |           587520.00 |       0.00% |
    |           1 | Hash Routed  |        587520.00 |           587520.00 |       0.00% |
    |           2 | Hash Routed  |        587520.00 |           587520.00 |       0.00% |
    |           3 | Top-K Routed |         14623.82 |             2447.45 |      83.26% |
    |           4 | Top-K Routed |         15871.75 |             3436.97 |      78.35% |
    |           5 | Top-K Routed |         15261.11 |             3110.88 |      79.62% |
    |           6 | Top-K Routed |         15717.74 |             1812.94 |      88.47% |
    |           7 | Top-K Routed |         15728.96 |             2615.92 |      83.37% |
    |           8 | Top-K Routed |         15999.51 |             4967.23 |      68.95% |
    |           9 | Top-K Routed |         15240.05 |             3919.24 |      74.28% |
    |          10 | Top-K Routed |         16086.67 |             2348.59 |      85.40% |
    |          11 | Top-K Routed |         16092.98 |             2607.32 |      83.80% |
    |          12 | Top-K Routed |         14255.88 |             2896.68 |      79.68% |
    |          13 | Top-K Routed |         14632.31 |             4420.16 |      69.79% |
    |          14 | Top-K Routed |         15010.89 |             4668.00 |      68.90% |
    |          15 | Top-K Routed |         16004.66 |             2190.59 |      86.31% |
    |          16 | Top-K Routed |         14625.14 |             2506.02 |      82.86% |
    |          17 | Top-K Routed |         13549.16 |             2880.44 |      78.74% |
    |          18 | Top-K Routed |         15119.30 |             3142.03 |      79.22% |
    |          19 | Top-K Routed |         14413.22 |             4134.32 |      71.32% |
    |          20 | Top-K Routed |         14239.44 |             2910.45 |      79.56% |
    |          21 | Top-K Routed |         13616.19 |             2687.55 |      80.26% |
    |          22 | Top-K Routed |         14360.10 |             1947.84 |      86.44% |
    |          23 | Top-K Routed |         14887.30 |             3880.00 |      73.94% |
    |          24 | Top-K Routed |         14978.84 |             4074.12 |      72.80% |
    |          25 | Top-K Routed |         14996.25 |             1401.96 |      90.65% |
    |          26 | Top-K Routed |         14232.96 |             3889.90 |      72.67% |
    |          27 | Top-K Routed |         14977.17 |             1858.18 |      87.59% |
    |          28 | Top-K Routed |         14378.70 |             2400.12 |      83.31% |
    |          29 | Top-K Routed |         13691.07 |             2181.08 |      84.07% |
    |          30 | Top-K Routed |         15055.45 |             3521.05 |      76.61% |
    |          31 | Top-K Routed |         14677.20 |             5095.23 |      65.28% |
    |          32 | Top-K Routed |         16188.89 |             3978.03 |      75.43% |
    |          33 | Top-K Routed |         14369.13 |             2921.98 |      79.66% |
    |          34 | Top-K Routed |         14483.14 |             5714.06 |      60.55% |
    |          35 | Top-K Routed |         15321.41 |             2911.53 |      81.00% |
    |          36 | Top-K Routed |         13667.09 |             2495.93 |      81.74% |
    |          37 | Top-K Routed |         14704.00 |             3494.09 |      76.24% |
    |          38 | Top-K Routed |         14339.20 |             2385.22 |      83.37% |
    |          39 | Top-K Routed |         14203.30 |             2454.55 |      82.72% |
    |          40 | Top-K Routed |         13582.94 |             2371.51 |      82.54% |
    |          41 | Top-K Routed |         16522.23 |             3767.62 |      77.20% |
    |          42 | Top-K Routed |         16277.68 |             1494.26 |      90.82% |
    |-------------|--------------|------------------|---------------------|-------------|
    | TOTAL/AVG   | Top-K Only   |        595982.83 |           123941.04 |      79.20% |
    
[Run : With Load Balancing Logs]: https://paste.googleplex.com/5739467624284160
[Run : PR-4497 Legacy Baseline Logs]: https://paste.googleplex.com/6184775940440064

## Checklist

- [x] I have performed a self-review of my code. For an optional AI review, add the gemini-review label.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in our documentation.

### DIFF
--- a/src/maxtext/configs/models/deepseek4-tiny.yml
+++ b/src/maxtext/configs/models/deepseek4-tiny.yml
@@ -11,16 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# Test Model config for DeepSeek-V4-Flash 284B (https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash)
 
-base_emb_dim: 4096
-base_num_query_heads: 64
+# Tiny model config for DeepSeek V4 for CPU execution and testing
+
+base_emb_dim: 64
+base_num_query_heads: 4
 base_num_kv_heads: 1
-base_num_decoder_layers: 7
-base_mlp_dim: 2048
-base_moe_mlp_dim: 2048
+base_num_decoder_layers: 43
+base_mlp_dim: 64
+base_moe_mlp_dim: 64
 vocab_size: 129280
-head_dim: 512
+head_dim: 32
+qk_rope_head_dim: 32
 
 # --- Standard Defaults ---
 enable_dropout: false
@@ -31,40 +33,38 @@ normalization_layer_epsilon: 1.0e-6
 decoder_block: "deepseek4"
 mhc_expansion_rate: 4
 first_num_hash_layers: 3
-indexer_head_dim: 128
-indexer_n_heads: 64
-indexer_topk: 512
+indexer_head_dim: 32
+indexer_n_heads: 4
+indexer_topk: 16
 
 # Note: Layers (0, 1, 2) are prefix layers as `first_num_hash_layers=3`.
-# The 6th layer (MTP module with compress_ratio=0) has been explicitly dropped for now.
-# This leaves exactly 7 layers: 3 prefix [0,0,4] + 4 scanned.
+# The 44th layer (MTP module with compress_ratio=0) has been explicitly dropped for now.
+# This leaves exactly 43 layers: 3 prefix [0,0,4] + 40 scanned.
 # `compress_ratio=0` uses sliding window attention. In this case, layer (0, 1).
-# This is a tiny version of deepseek4 with fewer layers and less experts for debugging.
-compress_ratios: [0, 0, 4, 128, 4, 128, 4]
+compress_ratios: [0, 0, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4, 128, 4]
 
 # --- MoE configuration ---
 mlp_activations: ["silu", "linear"]
-num_experts: 8
-num_experts_per_tok: 3
+num_experts: 16
+num_experts_per_tok: 4
 mlp_activations_limit: 10
 shared_experts: 1
 routed_score_func: "sqrtsoftplus"
 routed_bias: true
-routed_scaling_factor: 1.5
-
+routed_bias_update_rate: 0.001
+load_balance_loss_weight: 0.0001
+log_moe_bias_norms: false
 
 # --- Attention configuration ---
 attention_type: 'compressed'
-attention: 'dot_product'
-q_lora_rank: 1024
-o_groups: 8
-o_lora_rank: 1024
-sliding_window_size: 128
+q_lora_rank: 16
+o_groups: 4
+o_lora_rank: 16
+sliding_window_size: 32
 
 # --- RoPE ---
+
 rope_type: "default"
 rope_max_timescale: 10000                  # Main RoPE theta
 compressed_rope_max_timescale: 160000      # Compressed RoPE theta
-max_position_embeddings: 1048576
-original_max_position_embeddings: 65536
-
+max_position_embeddings: 4096
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -232,6 +232,10 @@ class Tid2EidVar(nnx.Variable):
   """Custom variable to hold tid2eid without trainable param overhead."""
 
 
+class MoEBiasVar(nnx.Variable):
+  """Custom NNX Variable for Auxiliary-Loss-Free MoE Routing Bias (DSV4)."""
+
+
 class GateLogit(nnx.Module):
   """A layer used to compute gate logits, allowing to return the pre bias values for DeepSeek routing."""
 
@@ -307,10 +311,17 @@ class GateLogit(nnx.Module):
     if self.use_bias:
       bias_axes = self.kernel_axes[-len(self.out_features_shape) :]
       bias_shape = kernel_shape[-len(self.out_features_shape) :]
+      # DSV3 was using nnx.Param and that code we are keeping the same
       self.bias = nnx.Param(
           default_bias_init(rngs.params(), bias_shape, self.weight_dtype),
           out_sharding=bias_axes,
       )
+      if self.model_name.startswith("deepseek4"):
+        # DSV4 uses MoEBiasVar to naturally isolate from sequence-wise updates
+        self.bias = MoEBiasVar(
+            default_bias_init(rngs.params(), bias_shape, self.weight_dtype),
+            out_sharding=bias_axes,
+        )
     else:
       self.bias = None
 
@@ -2582,13 +2593,29 @@ class RoutedMoE(nnx.Module):
 
   # See Switch Transformer (https://arxiv.org/abs/2101.03961) for more details.
   def load_balance_loss(self, top_k_indices, logits) -> jax.Array:
-    """Compute the load balance loss."""
+    """Compute the sequence-wise load balance loss.
+
+    For DeepSeek V4 like models, standard load balancing across an entire batch can
+    be inadequate due to heterogeneous prompt lengths and varying sequence
+    characteristics. This method implements sequence-wise load balancing by
+    computing the token density and routing probabilities on a per-sequence basis.
+
+    The resulting loss is scaled by `self.config.load_balance_loss_weight`.
+    When this configuration value is set > 0, the computed loss is aggregated
+    into the total training loss. By minimizing this scaled auxiliary loss,
+    the optimizer updates the routing parameters to actively enforce an even
+    distribution of tokens to experts within each individual sequence.
+    """
     expert_mask = jax.nn.one_hot(top_k_indices, num_classes=self.num_experts, dtype=jnp.int32)
     summed_expert_mask = jnp.sum(expert_mask, axis=2)
     # Get fraction of tokens dispatched to each expert
+    # jnp.mean over axis=1 (sequence length) isolates the token density per sequence.
     density = jnp.mean(summed_expert_mask, axis=1)
     # get fraction of probability allocated to each expert
+    # jnp.mean over axis=1 isolates the routing probability per sequence.
     density_prob = jnp.mean(logits, axis=1)
+    # The sequence-wise densities and probabilities are multiplied and then averaged
+    # over the batch dimension, scaled by the required constant.
     loss = jnp.mean(density * density_prob) * (self.num_experts**2) * self.config.load_balance_loss_weight
     return loss
 
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -916,6 +916,7 @@ class NNXDecoder(nnx.Module):
       kv_caches_stacked=None,
       skip_block_remat: bool = False,
       unroll: int = 1,
+      metadata_axis_name: str = "layers",
       **kwargs,
   ):
     """Runs the layer stack using nnx.scan.
@@ -938,6 +939,10 @@ class NNXDecoder(nnx.Module):
         e.g. per-layer) remat internally, to avoid double rematerialization.
       unroll: Number of scan iterations to unroll into straight-line code
         (forwarded to jax.lax.scan). unroll >= length fully unrolls the loop.
+      metadata_axis_name: The name of the scan axis used during layer initialization.
+        This must perfectly match the string passed to `_create_scanned_layers`
+        (e.g., "layers", "scanned_blocks") to prevent strict JAX `pjit` PyTree
+        metadata mismatch errors when using custom `nnx.Variable` types (like `MoEBiasVar`).
       **kwargs: Keyword args forwarded to the layer (filtered by the layer signature).
 
     Returns:
@@ -1065,7 +1070,7 @@ class NNXDecoder(nnx.Module):
 
       # Move the scan axis to each variable's param_scan_axis and restore its name
       # in the sharding metadata. jax.lax.scan emits it at position 0.
-      scanned_state = maxtext_utils_nnx.nnx_add_and_sync_scan_axis(scanned_state, "layers")
+      scanned_state = maxtext_utils_nnx.nnx_add_and_sync_scan_axis(scanned_state, metadata_axis_name)
 
       returned_kv_stacked = None
 
@@ -2027,6 +2032,7 @@ class NNXDecoder(nnx.Module):
           deterministic,
           model_mode,
           length=num_full_blocks,
+          metadata_axis_name="scanned_blocks",
           **layer_call_kwargs,
       )
 
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -432,9 +432,17 @@ class NNXDecoder(nnx.Module):
 
     self.scanned_layers = None
     self.is_deepseek = self.config.decoder_block == DecoderBlockType.DEEPSEEK
+    self.is_deepseek4 = self.config.decoder_block == DecoderBlockType.DEEPSEEK4
     self.is_gemma3 = self.config.decoder_block == DecoderBlockType.GEMMA3
     self.is_gemma4 = self.config.decoder_block == DecoderBlockType.GEMMA4
     self.is_gemma4_small = self.config.decoder_block == DecoderBlockType.GEMMA4_SMALL
+
+    if config.mhc_expansion_rate > 1 and config.decoder_block == DecoderBlockType.DEEPSEEK4:
+      self.hc_head = mhc.DeepSeek4HyperHead(
+          config=config,
+          mesh=self.mesh,
+          rngs=self.rngs,
+      )
 
     self._init_decoder_layers(decoder_block_classes, rngs, mesh)
 
@@ -532,12 +540,38 @@ class NNXDecoder(nnx.Module):
     """Initializes decoder layers with scanning (non-pipeline)."""
     if self.is_deepseek:
       self._init_scanned_deepseek(decoder_block_classes, rngs)
+    elif self.is_deepseek4:
+      self._init_scanned_deepseek4(rngs)
     elif self.is_gemma3:
       self._init_scanned_gemma3(decoder_block_classes, rngs, mesh)
     elif self.is_gemma4:
       self._init_scanned_gemma4(decoder_block_classes, rngs, mesh)
     else:
       self._init_scanned_generic(decoder_block_classes, rngs)
+
+  def _init_scanned_deepseek4(self, rngs):
+    """Initializes DeepSeek V4 scanned layers: unrolls first_num_hash_layers prefix layers and scans remaining full blocks."""
+    config = self.config
+    self.layers = nnx.List([])
+    num_hash_layers = config.first_num_hash_layers
+    for layer_idx in range(num_hash_layers):
+      self._create_and_register_layer(
+          deepseek4.DeepSeek4DecoderLayer,
+          rngs,
+          "layers",
+          layer_idx,
+          layer_idx=layer_idx,
+      )
+
+    num_remaining_layers = config.num_decoder_layers - num_hash_layers
+    num_full_blocks = num_remaining_layers // 2
+    if num_full_blocks > 0:
+      self.scanned_blocks = self._create_scanned_layers(
+          deepseek4.DeepSeek4ScannableBlock,
+          length=num_full_blocks,
+          metadata_axis_name="scanned_blocks",
+          rngs=rngs,
+      )
 
   def _init_scanned_deepseek(self, decoder_block_classes, rngs):
     """Initializes scanned DeepSeek layers with optional Engram support."""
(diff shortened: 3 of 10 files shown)


tests/unit/deepseek_routed_bias_test.py
```diff
@@ -0,0 +1,126 @@
+"""Tests for DeepSeek routed bias updates."""
```
parambole: Hey @Dipak Gaikwad,

Could we also add a unit test case for deepseek4-tiny in tests/unit/deepseek_routed_bias_test.py?

Since DeepSeek V4 introduces a hybrid routing structure—using Hash Routing for the first few layers (which lack bias/aux-loss) and standard Top-K routing for the later layers—having a test that uses model_name=deepseek4-tiny, decoder_block=deepseek4, and first_num_hash_layers=2 would help ensure that our dynamic traversal cleanly handles this mixed topology without crashing and correctly ignores the hash routing layers.

parambole (commented):
Hey @dipakg-lang, on a high level, the changes look good.
a) Can you add test cases for the dsv4-like structure?
b) Also, could you add the bug number for sequence load balancing loss?

tests/unit/deepseek_routed_bias_test.py
dipakg-lang: Done.

dipakg-lang:
> Hey @dipakg-lang, on a high level, the changes look good. a) Can you add test cases for the dsv4-like structure? b) Also, could you add the bug number for sequence load balancing loss?



> The sequence wise changes are already present in the code they are enabled with the multiplier :  load_balance_loss_weight

parambole (approved):
LGTM. Thank you for adding these changes.

RissyRan:
Note: This is a fork repo, so unfortunately Gemini review doesn't work due to the API key thing.

src/maxtext/configs/models/deepseek4-tiny.yml
```diff
@@ -0,0 +1,69 @@
...
+num_experts_per_tok: 4
+shared_experts: 1
+routed_score_func: "sqrtsoftplus"
+routed_bias: true
+routed_bias_update_rate: 0.001
+load_balance_loss_weight: 0.0001
+adamw_mask: [".*gate.*bias.*"]
+
+# --- Attention configuration ---
+attention: 'dot_product'
```
RissyRan: We usually just put model configs in this yml file. The experimental config (like dot_product or flash, etc) usually is passed from cmd.

src/maxtext/configs/models/deepseek4-tiny.yml
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tiny model config for DeepSeek V4 for CPU execution and testing
+
+base_emb_dim: 64
```
RissyRan, line 17: Do you know why those model configs (if from v4) are not included in the 284B model? if using default value from function, it's risky - https://screenshot.googleplex.com/4na5Jwtha49zfZM

src/maxtext/layers/moe.py
```diff
@@ -348,8 +348,11 @@ def __call__(self, inputs: jax.Array, _initializing: bool = False) -> Tuple[jax.
       pre_bias_logits = output
 
     if self.use_bias:
+      # Architectural Note:  Bias is an nnx.Param rather than nnx.Variable due to Linen/NNX state
```
RissyRan: I think we have turned on NNX by default. Shall we actually use nnx.variable for a cleaner solution? It seems nnx.param + stop_gradient is a shortcut/hack. Using nnx.variable, I would expect we don't need `adamw_mask: [".*gate.*bias.*"]`. Please let me know your thoughts!

cc @shuningjin involved discussion before

src/maxtext/optimizers/optimizers.py
```diff
@@ -238,6 +239,21 @@ def get_optimizer(config, learning_rate_schedule, model=None):
         lambda params: jax.tree_util.tree_map(lambda x: "frozen" if x else "trainable", freeze_mask_fn(params)),
     )
 
+  if getattr(config, "routed_bias", False) and getattr(config, "routed_bias_update_rate", 0.0) > 0.0:
```
RissyRan: This is needed if using nnx.variable?

src/maxtext/trainers/pre_train/train.py
```diff
@@ -501,12 +477,47 @@ def move(path, value):
     else:
       new_state = state.apply_gradients(grads=full_grads)
 
-    # Apply updates for Auxiliary-Loss-Free load balancing for DeepSeek family
-    if config.routed_bias and config.routed_bias_update_rate > 0.0 and moe_bias_updates is not None:
-      target_path = ("params", "decoder", "moe_layers", "DeepSeekMoeBlock_0", "MoeBlock_0", "gate", "bias")
-      # Updates the shape to be aligned with state.
-      moe_bias_updates = jnp.array(moe_bias_updates[0]).transpose()
-      new_state = maxtext_utils.update_state_param(new_state, target_path, moe_bias_updates)
+    # Apply updates for Auxiliary-Loss-Free load balancing for the DeepSeek family.
```
RissyRan: Could you help make some comments and explanation how the sequence-wise is applied in the code for future readers? Thank you!

RissyRan:
Thanks for the change! I have a few high level comments:

1) based on previous discussion, it seems nnx.variable is the right way to go. Did you meet some issues to leverage it?
2) baed on your tests, it seems the [loss](https://screenshot.googleplex.com/9pD2bH3gfZwzN7v) doesn't change, do you happen to know why? Even use synthetic dataset, we should see the decrease
3) we are very actively working on DS v3 optimization. would like to ensure there is no perf regression with current change. Could you help have a run using small/tiny v3 configs (should have in configs) to sanity check before/after your changes and see if tflops/s/device is the same? regular DS v3 using pure FSDP should be sufficient

src/maxtext/configs/models/deepseek4-tiny.yml
dipakg-lang: adding @parambole to this conversation as well who has added the 284B model, so that we can resolve this.

src/maxtext/configs/models/deepseek4-tiny.yml
dipakg-lang: okey, would you suggest we skip them and have the user pass it instead? as we only support the dot product right now, I felt it would be safer to have that value present here and later change when we remove it from there when we start supporting more options, thoughts?

src/maxtext/layers/moe.py
dipakg-lang: Yes, I had the same suggestion earlier, but then we will have to manage the overhead of managing this variable ourselves that is why we have gone through this route, please let me know if you strongly feel about going through the variable route then I will have to create a plan on this variable management.

src/maxtext/optimizers/optimizers.py
dipakg-lang: No, this part would be saved then. but we will have own variable management separately if not param

src/maxtext/trainers/pre_train/train.py
dipakg-lang: Sure, it is implemented through load_balance_loss_weight and it's implementation already present in the code!

dipakg-lang:
> Thanks for the change! I have a few high level comments:
> 
> 1. based on previous discussion, it seems nnx.variable is the right way to go. Did you meet some issues to leverage it?
> 2. baed on your tests, it seems the [loss](https://screenshot.googleplex.com/9pD2bH3gfZwzN7v) doesn't change, do you happen to know why? Even use synthetic dataset, we should see the decrease
> 3. we are very actively working on DS v3 optimization. would like to ensure there is no perf regression with current change. Could you help have a run using small/tiny v3 configs (should have in configs) to sanity check before/after your changes and see if tflops/s/device is the same? regular DS v3 using pure FSDP should be sufficient

1. Explained in the other comment happy to do if we feel strongly about it

2. I have extensively tested this change for longer time, I was able to see loss constantly going down and similar behavior to  code without this change.

3. can you share the recipe with me for V3 so that I can make sure there are no regressions there, but as this code change is only enabled through these two constants, I don't think there will any regression for v3 as I don't see them being used in v3's config., but good to test it.

	Auxiliary loss free load balancing  :  routed_bias_update_rate.
 
	Sequence wise load balancing : load_balance_loss_weight

src/maxtext/configs/models/deepseek4-tiny.yml
RissyRan: In this case, we commonly put assertion like this to indicate which case is not supported:

https://github.com/AI-Hypercomputer/maxtext/blob/7d8d169a094d33dc2d40849de7bc9e8eb7a47664/src/maxtext/configs/types.py#L2567-L2572

RissyRan:
> > Thanks for the change! I have a few high level comments:
> > 
> > 1. based on previous discussion, it seems nnx.variable is the right way to go. Did you meet some issues to leverage it?
> > 2. baed on your tests, it seems the [loss](https://screenshot.googleplex.com/9pD2bH3gfZwzN7v) doesn't change, do you happen to know why? Even use synthetic dataset, we should see the decrease
> > 3. we are very actively working on DS v3 optimization. would like to ensure there is no perf regression with current change. Could you help have a run using small/tiny v3 configs (should have in configs) to sanity check before/after your changes and see if tflops/s/device is the same? regular DS v3 using pure FSDP should be sufficient
> 
> 1. Explained in the other comment happy to do if we feel strongly about it
> 2. I have extensively tested this change for longer time, I was able to see loss constantly going down and similar behavior to  code without this change.
> 3. can you share the recipe with me for V3 so that I can make sure there are no regressions there, but as this code change is only enabled through these two constants, I don't think there will any regression for v3 as I don't see them being used in v3's config., but good to test it.
>    Auxiliary loss free load balancing  :  routed_bias_update_rate.
>    Sequence wise load balancing : load_balance_loss_weight

2. I see. It seems strange, as we usually see obvious loss decreasing within every 10 steps. One [example](https://paste.googleplex.com/6062473349300224). So I feel your test seems interesting.

3. Thanks for the sanity check. You could just this [script](https://github.com/AI-Hypercomputer/maxtext/blob/7d8d169a094d33dc2d40849de7bc9e8eb7a47664/tests/end_to_end/tpu/deepseek/v3-671b/2_test_deepseek_mtp.sh#L33-L53).  But without `load_parameters_path` (so pre-training) and without     `mtp_num_layers=1  mtp_loss_scaling_factor=0.1 ` for simplicity. If you'd like to run it in local like v5p-8, you could manually reduce # of layers in v3 yml config.

src/maxtext/layers/moe.py
RissyRan: > we will have to manage the overhead of managing this variable 

Oh I didn't realize that. I thought we were thinking of nnx.variable in this [thread](https://chat.google.com/room/AAAAHtQgNdM/o4ylPKvfh0c/o4ylPKvfh0c?cls=10). Please let me know if I missed something. 

Gemini says "Yes, the DeepSeek router bias is an absolute textbook use case for nnx.Variable"

src/maxtext/trainers/pre_train/train.py
RissyRan: Hmm... we are actually implemented sequence-wise already? Oh, I thought it's not based on previous discussion.

src/maxtext/layers/moe.py
```diff
@@ -2111,13 +2114,29 @@ def generate_masks(self, top_k_indices, softmax_probs):
 
   # See Switch Transformer (https://arxiv.org/abs/2101.03961) for more details.
   def load_balance_loss(self, top_k_indices, logits) -> jax.Array:
-    """Compute the load balance loss."""
+    """Compute the sequence-wise load balance loss.
```
shuningjin, line 2596: This seems sequence-wise already. I also added note in b/521990776. Could you double check? @RissyRan

src/maxtext/layers/moe.py
shuningjin: I would prefer `nnx.variable` to avoid manipulation of optimizer mask. Could you elaborate what is the "overhead of managing this variable"?

src/maxtext/configs/models/deepseek4-tiny.yml
shuningjin: I think [deepseek4-284b.yml](https://github.com/AI-Hypercomputer/maxtext/blob/8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a/src/maxtext/configs/models/deepseek4-284b.yml#L15) from main branch is the up-to-date version. In particular, it has been updated in [PR4153](https://github.com/AI-Hypercomputer/maxtext/pull/4153) based on review.

Should make this tiny version consistent with 284b when possible.

src/maxtext/configs/models/deepseek4-tiny.yml
shuningjin: This already addressed in [PR4153](https://github.com/AI-Hypercomputer/maxtext/pull/4153)
- This `attention: 'dot_product'` has been removed from [deepseek4-284b.yml](https://github.com/AI-Hypercomputer/maxtext/blob/8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a/src/maxtext/configs/models/deepseek4-284b.yml#L15) 
- types check is already added https://github.com/AI-Hypercomputer/maxtext/blob/8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a/src/maxtext/configs/types.py#L3135-L3136

Please remove `attention: 'dot_product'` here as well.

src/maxtext/trainers/pre_train/train.py
shuningjin: The logic of routed bias update (in loss-free load balancing) seems really complicated. Is it possible to simplify?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1265,6 +1265,7 @@ def get_norm_layer(self, num_features: int, rngs: nnx.Rngs):
         DecoderBlockType.MISTRAL,
         DecoderBlockType.MIXTRAL,
         DecoderBlockType.DEEPSEEK,
+        DecoderBlockType.DEEPSEEK4,
```
shuningjin, line 1268: Please merge main. This is already in head.
https://github.com/AI-Hypercomputer/maxtext/blob/8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a/src/maxtext/layers/nnx_decoders.py#L1124

shuningjin (commented):
Thanks for the feature! I agree with @RissyRan in many aspects.

1. More testing with Deepseek4 & Deepseek3
2. loss-free lb - Routed bias implementation:
  - it currently uses: `nnx.Param` + stop gradient + optimizer mask for adamw and muon
  - Can we simplify via `nnx.Variable` to avoid optimizer mask?
3. (additionally) loss-free lb - Routed bias update: 
  - in `train.py` uses complicated logic based on path. 
  - Can we simplify?

For lb loss, perhaps the existing [load_balance_loss](https://github.com/AI-Hypercomputer/maxtext/blob/8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a/src/maxtext/layers/moe.py#L2113-L2122) is already sequence-wise. I added some notes in b/521990776

src/maxtext/configs/models/deepseek4-tiny.yml
dipakg-lang: Made consistent with 284b perfectly only scaled down version of it.

src/maxtext/configs/models/deepseek4-tiny.yml
dipakg-lang: Removed dot produce here as well

src/maxtext/layers/moe.py
dipakg-lang: Declaring the bias as an nnx.Variable instead of an nnx.Param introduces framework overhead because JAX requires custom state to be manually threaded through the train_step pipeline
  (passed as inputs and returned as outputs) on every iteration. Additionally, because MaxText's sharding and checkpointing are built for the params tree, using a new state collection
  would require writing custom PartitionSpec logic to distribute the bias across the mesh and custom serialization code to save and restore it. This is what motivated me to do design it in this way, otherwise if you remember I was the one pointing out need for this to be a variable. Please let me know if you would like to have discussion on this.

src/maxtext/layers/moe.py
dipakg-lang: The sequence wise building blocks have been reused for DSV4 instead of rebuilding from sctratch to avoid the code replication.

src/maxtext/layers/nnx_decoders.py
dipakg-lang: Sure, rebased.

src/maxtext/trainers/pre_train/train.py
dipakg-lang: Yes, that was my understanding but diving deep into the code, looks like you are the one who developed it in the past :)

dipakg-lang: I think you are talking about this code which is about locating the bias in the tree, considering the nature of path of the bias, we have to have logic to locate it in such way that we don't miss it and the code continue to work for any other models as well is resulting in this code, if you have any ideas we can always incorporate them!

src/maxtext/configs/models/deepseek4-tiny.yml
RissyRan: It seems it still [here](https://screenshot.googleplex.com/Bv9WCRahbbaNQrd)?

src/maxtext/configs/models/deepseek4-tiny.yml
RissyRan: Thanks! Didn't see qk_rope_head_dim in 284-version configs?

dipakg-lang:
> Thanks for the change! I have a few high level comments:
> 
> 1. based on previous discussion, it seems nnx.variable is the right way to go. Did you meet some issues to leverage it?
> 2. baed on your tests, it seems the [loss](https://screenshot.googleplex.com/9pD2bH3gfZwzN7v) doesn't change, do you happen to know why? Even use synthetic dataset, we should see the decrease
> 3. we are very actively working on DS v3 optimization. would like to ensure there is no perf regression with current change. Could you help have a run using small/tiny v3 configs (should have in configs) to sanity check before/after your changes and see if tflops/s/device is the same? regular DS v3 using pure FSDP should be sufficient

RissyRan (approved):
LGTM at high level, please address following before merge:


1. remove unneeded configs `adamw_mask: [".gate.*bias."]`
2. ensure configs are aligned with tiny and flash, i.e. cross check qk_rope_head_dim set up
3. please consider to have a debug flag to enable required norm bias metrics.

dipakg-lang:
> LGTM at high level, please address following before merge:
> 
> 1. remove unneeded configs `adamw_mask: [".gate.*bias."]`
> 2. ensure configs are aligned with tiny and flash, i.e. cross check qk_rope_head_dim set up
> 3. please consider to have a debug flag to enable required norm bias metrics.

1. Done
2. Done, added in both the configs.
3. Done, moved behind a debug flag

PR #3526, opened 2026-03-31
[NNX] NNX migration (11/N): set pure_nnx / enable_nnx / pure_nnx_decoder defaults to True

ecnal-cienet:
## NNX Migration Route Map
1. ✅ Add NNX scaffolding: `pure_nnx` flag, `init_state_fn`, `TrainStateNNX`, NNX utils. Linen workflow unchanged. (PR #3427)
2. ✅ NNX sharding utilities: `get_abstract_state_nnx`, `get_named_sharding_nnx`, `set_named_sharding_nnx`, `get_partition_spec_nnx`, `get_mesh_from_config`. (PR #3470)
3. ✅ NNX fully supported end-to-end: `TrainStateNNX`, model creation, gradient accumulation, checkpointing, and training loop dispatch. (PR #3500)
4. ✅ Sharding diagnostics on NNX + post-training bugfixes that surfaced once the NNX path was exercised end-to-end. (PR #3652)
5. ✅ NNX correctness fixes, feature enablements, vocab tiling on NNX.
6. ✅ NNX-native DPO.
7. ✅ NNX-native MaxEngine inference. (PR #3821)
8. ✅ NNX-native LoRA + GRPO. (PR #3824)
9. ✅ NNX-aware QK-Clip + remaining checkpoint utilities. (PR #3836)
9.5. ✅ NNX + AQT in MaxEngine + serve-mode reload + gpt3 prefill fix. (PR #3844)
10. ✅ Vocab tiling `custom_vjp` for NNX. (PR #3849)
11. 🔄 **[This PR]** Flip `enable_nnx`, `pure_nnx`, `pure_nnx_decoder` from `False` → `True` in `base.yml`. Bundle the NNX-only fixes that surface once `pure_nnx=True`.
12. ❌ Delete Linen-specific code paths and NNX compatibility flags. (PR #4038)

## Description

PR6–PR10 promoted every routed-to-Linen feature to NNX-native; #2885 added NNX-native pipeline parallelism; #4040 added Qwix on NNX. This PR flips the three defaults in `base.yml` and bundles the NNX-only fixes that surface once `pure_nnx=True`.

## Changes

**`src/maxtext/configs/base.yml`** — flip defaults
- `enable_nnx: False → True`, `pure_nnx: False → True`, `pure_nnx_decoder: False → True`.

**`src/maxtext/utils/sharding.py`** — Zero-1 on flat `nnx.State`
[...]

PR #4255, opened 2026-06-24
NNX: fix Linen-parity gaps + unit tests

ecnal-cienet:
# Description
NNX (`pure_nnx=True`) had Linen-only / silently-divergent gaps across train / loss / decoder / metrics / GRPO. This closes them and adds correctness unit tests. The fixes apply on `main` independently; PR #3526 (flip defaults to NNX) makes them the live default path, and the UTs pin the behavior either way.

## Fixes

| # | Issue |
|---|-------|
| 1 | `skip_step_on_spikes`: silent no-op on NNX (`apply_gradients` didn't forward `loss`/`grad_norm`). Now forwarded + metric surfaced. |
| 2 | `loss_fn`: NNX checked vocab-tiling before indexer warm-up; reordered to match Linen. |
| 3 | NNX decoder logits guards used `self.model_mode` instead of the call-arg `model_mode`. |
| 4 | `routed_bias`: updates silently dropped on NNX (Linen `"intermediates"` prefix absent on NNX dict). Now matched by suffix. |
| 5 | `record_internal_nn_metrics`: KeyError on NNX. Now NNX-aware via suffix collection. |
| 6 | qwix: crashed under `pure_nnx` with the bridged decoder; bridge now skips qwix's non-`Variable` attrs + a config guard rejects bridged-decoder+qwix. |
| 7 | `maxengine.set_engine_vars_from_base_engine`: AttributeError on NNX; now uses `get_kv_cache_annotations_nnx`. |
| 8 | GRPO `gradient_accumulation_steps>1`: NotImplementedError on NNX. Implemented; also fixed the GA loss metric (`sum/GA`, not `sum/total_weights`). |
| 9 | GRPO `scan_layers=False`: NotImplementedError on NNX. Guard removed (NNX policy already matches the inference layout). |
| 10 | GRPO `optimizer_memory_host_offload`: ignored on NNX; now moves opt state to device before the update. |

Also re-declared the legacy GRPO config fields (`inference_replicas` / `inference_devices_per_replica` /
`inference_rollouts` / `use_pathways_reshard`) in `types.py` — they were dropped from the schema so
`grpo.yml` couldn't load (pre-existing on `main`).

## Tests

`tests/unit/{train_nnx_test,grpo_nnx_test,maxengine_nnx_test,nnx_quant_guard_test}.py` — 27 pass on CPU:

```bash
PYTHONPATH=src JAX_PLATFORMS=cpu python3 -m pytest \
  tests/unit/grpo_nnx_test.py tests/unit/train_nnx_test.py \
  tests/unit/maxengine_nnx_test.py tests/unit/nnx_quant_guard_test.py -q
```

# Checklist

### DIFF
--- a/src/maxtext/experimental/rl/grpo_trainer.py
+++ b/src/maxtext/experimental/rl/grpo_trainer.py
@@ -471,8 +471,8 @@ def _train_step_nnx(model_graphdef, config, state_mesh_shardings, state, data):
   Args:
     model_graphdef: NNX `GraphDef` of the `TrainStateNNX`.
     config: Training configuration object.
-    state_mesh_shardings: Sharding spec for the train state. Unused on this
-      path; kept for signature parity with `train_step`.
+    state_mesh_shardings: Sharding spec for the train state; used to move the
+      optimizer state to device when `optimizer_memory_host_offload` is set.
     state: Flat `nnx.State` matching `model_graphdef`.
     data: A batch dict produced by the GRPO input pipeline.
 
@@ -480,14 +480,6 @@ def _train_step_nnx(model_graphdef, config, state_mesh_shardings, state, data):
     A tuple `(new_state, metrics)`. `new_state` is filtered to exclude
     `nnx.Intermediate`. `metrics` is a dict shaped like the Linen path's.
   """
-  del state_mesh_shardings  # Host-offload paths are not yet wired up here.
-
-  if config.gradient_accumulation_steps > 1:
-    raise NotImplementedError(
-        "GRPO + pure_nnx + gradient_accumulation_steps>1 not supported yet. "
-        "Set gradient_accumulation_steps=1 or pure_nnx=False."
-    )
-
   state = nnx.merge(model_graphdef, state)  # Reconstruct the TrainStateNNX.
   policy_graphdef, curr_params, rest = nnx.split(state.model, nnx.Param, ...)
   # Split the reference model into (graphdef, state) so we pass `ref_state` as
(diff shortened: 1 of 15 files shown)


src/maxtext/experimental/rl/grpo_trainer.py
```diff
@@ -505,13 +497,61 @@ def diff_wrapper(param, rest, ref_state, config, data):
...
+    }
+    carry, aux = jax.lax.scan(accumulate_gradient, init_carry, ga_data, length=config.gradient_accumulation_steps)
+    # total_loss is already a per-batch mean (and includes moe_lb), so the full-batch
+    # loss is the mean across the equal-sized microbatches.
+    loss = carry["loss"] / config.gradient_accumulation_steps
+    raw_grads = jax.tree_util.tree_map(lambda arr: arr / carry["total_weights"], carry["grad"])
+    aux = jax.tree.map(lambda x: jnp.sum(x, axis=0), aux)
+    new_rest = carry["rest"]
+  else:
+    (loss, (aux, new_rest)), raw_grads = grad_func(curr_params, rest, ref_state, config, data)
```
igorts-git: Nit, I think the code would be more readable if it was written such that this short part comes first:
```
if config.gradient_accumulation_steps <= 1:
    (loss, (aux, new_rest)), raw_grads = grad_func(curr_params, rest, ref_state, config, data)
else:
     # The big code block here
```

src/maxtext/trainers/pre_train/train.py
```diff
@@ -208,15 +208,15 @@ def loss_fn(model, config, data, dropout_rng, params, sparsity_state=None, is_tr
       intermediate_outputs["mtp_losses"] = nnx.pop(model, mtp_losses).to_pure_dict()
       intermediate_outputs["mtp_acceptance"] = nnx.pop(model, mtp_acceptance).to_pure_dict()
 
-    if config.num_vocab_tiling > 1:
-      hidden_state_key = ("decoder", "hidden_states")
-      hidden_states = maxtext_utils.get_nested_value(intermediate_outputs, hidden_state_key)[0]
-      xent_sum, total_z_loss = vocab_tiling_nnx_loss(model, hidden_states, data, config, is_train)
-    elif (config.use_indexer and not config.indexer_sparse_training) and is_train:
+    if (config.use_indexer and not config.indexer_sparse_training) and is_train:
```
NuojCheng, line 211: Could you help me understand why change the order of `if config.num_vocab_tiling > 1` and `if (config.use_indexer and not config.indexer_sparse_training) and is_train`?

src/maxtext/trainers/pre_train/train.py
```diff
@@ -293,8 +293,25 @@ def loss_fn(model, config, data, dropout_rng, params, sparsity_state=None, is_tr
   # get MoE routed bias term updates
   moe_bias_updates = None
   if config.routed_bias and config.routed_bias_update_rate > 0.0:
-    nested_key = ("intermediates", "decoder", "moe_layers", "moe_bias_updates")
-    moe_bias_updates = maxtext_utils.get_nested_value(intermediate_outputs, nested_key, None)
+    if isinstance(model, nn.Module):
```
NuojCheng, line 296: does this branch get triggered only when `enable_nnx=False`?

NuojCheng (approved):

src/maxtext/trainers/pre_train/train.py
ecnal-cienet: `isinstance(model, nn.Module)`is the Linen path — the same predicate this function uses for the Linen/NNX split throughout (the forward at the top, etc.); the `else` is the NNX path. It keys on the model type, not a flag directly: a pure-NNX model (`pure_nnx=True`, built with `rngs` → `models.Transformer`) takes the `else`; everything else is a Linen `nn.Module` and takes this branch. So it's `pure_nnx=False`, which is broader than `enable_nnx=False` — it also covers `enable_nnx=True, pure_nnx=False`, where `transformer_as_linen` bridges the NNX core into a Linen `nn.Module`.

src/maxtext/trainers/pre_train/train.py
ecnal-cienet: This restores parity with the Linen branch just above (lines 152–160), which checks the indexer dense warm-up before `num_vocab_tiling > 1`; the NNX branch had them reversed. The two conditions are mutually exclusive, so the order only matters when both are enabled (an indexer model trained with vocab tiling): during dense warm-up the main LM loss must be skipped entirely (`xent_sum = 0` — main params frozen, only the indexer trains via KL). With tiling checked first, a warm-up step would run the vocab-tiling loss instead of skipping it. Putting the warm-up check first gives it precedence and matches Linen.

igorts-git (approved):

src/maxtext/experimental/rl/grpo_trainer.py
ecnal-cienet: Fixed. Thank you.

PR #4288, opened 2026-06-29
fix: decoders and pipeline parity gaps of linen to nnx migrations 

mesakhcienet:
# Description

Closes the remaining behavioral gaps between the pure-NNX decoder/pipeline path and the Linen reference. 
The NNX path now reproduces Linen for DeepSeek-V4, per-stage remat, and the pipeline's non-trainable / repeat-level-remat handlning.

### DeepSeek-V4 NNX decoder port

  - Registered `DEEPSEEK4` in `NNXDecoder.get_decoder_layer`(was missing → ValueError at construction) and added full decoder-level handling: norm dispatch (RMSNorm), scanned + non-scanned init, `_apply_deepseek4_scanned_blocks` (prefix first_num_hash_layers unroll + paired HCA/CSA scan), global layer_idx, and decoder_input_tokens threading — matching Linen `_apply_deepseek4_scanned_blocks`.

###  Per-stage pipeline remat parity (set_remat_policy_on_layers_per_stage)

  - The flag was a no-op in the NNX pipeline after the Linen→NNX migration. Restored per-stage remat (jax.checkpoint) + params-only host-offload in `NNXSequentialPipelineStage` / `NNXScannedPipelineStage`, wired from both stage builders, incl. `num_layers_per_pipeline_stage == 1`.
  - Fix: decoupled "apply remat" from the policy value. `remat_policy='full'` resolves to `None` (== full remat, as Linen nn.remat(policy=None)); the old `if policy is  not None` gate silently dropped remat for the default 'full' policy. Now gated on the flag via an explicit `apply_remat` argument.

### Pipeline Linen→NNX migration parity (pipeline.py)

  - `non_trainable` collection: the migration asserted the iteration-scan catch-all was RngState-only, crashing any pipelined model with a non-trainable variable (e.g. the `DeepSeek-V4` hash-routing table). Non-circular now broadcasts `non_trainable` as a loop-invariant constant (4-way state split); circular carries it via
  `carry_state`. 
  - circular repeat-level remat: made unconditional to match the Linen reference (whose flag-check was dead code, always rematting). Default flag path unchanged; only flag=False configs regain the dropped rematerialization.

###  Unit Tests

  - tests/unit/nnx_decoders_test.py (+ DeepSeek-V4 construct/forward/scan parity, pipeline-stage forward + remat transparency, per-stage-remat-applied guards, layer_map registration guard).
  - tests/unit/nnx_pipeline_test.py (new): coverage for NNXPipeline / NNXCircularPipeline — non-circular + circular forward, non_trainable partitioning, repeat-remat output transparency. 
  
# Tests
[Sheet](https://docs.google.com/spreadsheets/d/1rqMs8vRCdecI-xi4EQasjzaioxHNp2lWbbd8-SRnWGk/edit?gid=32811007#gid=32811007) combination of  `set_remat_policy_on_layers_per_stage` flag.

# Checklist

### DIFF
--- a/src/maxtext/layers/decoders.py
+++ b/src/maxtext/layers/decoders.py
@@ -33,7 +33,12 @@ from maxtext.layers import linears
 from maxtext.layers import mhc
 from maxtext.layers import normalizations
 from maxtext.layers import pipeline
-from maxtext.layers.nnx_decoders import NNXDecoderLayer, NNXSequentialPipelineStage, NNXScannedPipelineStage
+from maxtext.layers.nnx_decoders import (
+    NNXDecoderLayer,
+    NNXSequentialPipelineStage,
+    NNXScannedPipelineStage,
+    _make_single_layer_remat_stage_cls,
+)
 from maxtext.layers import quantizations
 from maxtext.layers.attentions import attention_as_linen
 from maxtext.layers.embeddings import attend_on_embedding, embed_as_linen, positional_embedding_as_linen
@@ -526,6 +531,7 @@ class Decoder(nn.Module):
         DecoderBlockType.SIMPLE: [simple_layer.SimpleDecoderLayer],
         DecoderBlockType.SIMPLE_MLP: [simple_layer.SimpleMlpDecoderLayer],
         DecoderBlockType.DEEPSEEK: [deepseek.DeepSeekDenseLayer, deepseek.DeepSeekMoELayer],
+        DecoderBlockType.DEEPSEEK4: get_scannable(deepseek4.DeepSeek4DecoderLayer, deepseek4.DeepSeek4ScannableBlock),
         DecoderBlockType.LLAMA4: get_scannable(llama4.Llama4DecoderLayer, llama4.Llama4ScannableBlock),
         DecoderBlockType.OLMO3: get_scannable(olmo3.Olmo3DecoderLayer, olmo3.Olmo3ScannableBlock),
     }
@@ -567,52 +573,49 @@ class Decoder(nn.Module):
     cfg = self.config
     base_stage_cls = decoder_blocks[1] if cfg.decoder_block == DecoderBlockType.DEEPSEEK else decoder_blocks[0]
 
+    # Per-stage-layer remat (+ params-only host-offload inside the stage) when the flag is set.
+    # apply_per_stage_remat is the boolean decision; per_stage_remat is the policy value
+    # (None == full remat for remat_policy='full', matching Linen nn.remat(policy=None)).
+    apply_per_stage_remat = cfg.set_remat_policy_on_layers_per_stage
+    per_stage_remat = self.get_remat_policy() if apply_per_stage_remat else None
+
     if cfg.num_layers_per_pipeline_stage == 1:
+      if apply_per_stage_remat:
+        # Linen nn.remat parity: keep params TOP-LEVEL (no 'layers_0' nesting).
+        stage_cls = _make_single_layer_remat_stage_cls(base_stage_cls)
+        return stage_cls(
+            config=cfg,
+            mesh=self.mesh,
+            quant=self.quant,
+            model_mode=self.model_mode,
+            rngs=rngs,
+            remat_policy=per_stage_remat,
+            apply_remat=True,
+        )
       return base_stage_cls(config=cfg, mesh=self.mesh, quant=self.quant, model_mode=self.model_mode, rngs=rngs)
     elif cfg.scan_layers_per_stage:
       return NNXScannedPipelineStage(
-          base_stage_cls, cfg.num_layers_per_pipeline_stage, cfg, self.mesh, self.quant, self.model_mode, rngs=rngs
+          base_stage_cls,
+          cfg.num_layers_per_pipeline_stage,
+          cfg,
+          self.mesh,
+          self.quant,
+          self.model_mode,
+          rngs=rngs,
+          remat_policy=per_stage_remat,
+          apply_remat=apply_per_stage_remat,
       )
     return NNXSequentialPipelineStage(
-        base_stage_cls, cfg.num_layers_per_pipeline_stage, cfg, self.mesh, self.quant, self.model_mode, rngs=rngs
+        base_stage_cls,
+        cfg.num_layers_per_pipeline_stage,
+        cfg,
+        self.mesh,
+        self.quant,
+        self.model_mode,
+        rngs=rngs,
+        remat_policy=per_stage_remat,
+        apply_remat=apply_per_stage_remat,
     )
-
-  def get_pipeline_stage_module(self, decoder_blocks):
-    """get pipeline stage module"""
-
-    def get_layer_to_pipeline(blocks, cfg):
-      if cfg.decoder_block == DecoderBlockType.DEEPSEEK:
-        return blocks[1]  # return the sparse block
-      else:
-        return blocks[0]
-
-    cfg = self.config
-    base_stage = get_layer_to_pipeline(decoder_blocks, cfg)
-    if cfg.set_remat_policy_on_layers_per_stage:
-      policy = self.get_remat_policy()
-      base_stage = self.set_remat_policy([base_stage], policy)[0]
-    if cfg.num_layers_per_pipeline_stage == 1:
-      stage_module = base_stage(config=cfg, mesh=self.mesh, quant=self.quant, model_mode=self.model_mode)
-    elif cfg.scan_layers_per_stage:
-      stage_module = self.scan_decoder_layers(
-          cfg,
-          base_stage,
-          cfg.num_layers_per_pipeline_stage,
-          "layers_per_stage",
-          self.mesh,
-          in_axes_tuple=(nn.broadcast,) * 4,
-          model_mode=self.model_mode,
-      )
-    else:
-      stage_module = SequentialBlockDecoderLayers(
-          decoder_layer=base_stage,
-          num_decoder_layers=cfg.num_layers_per_pipeline_stage,
-          config=cfg,
-          mesh=self.mesh,
-          quant=self.quant,
-          model_mode=self.model_mode,
-      )
-    return stage_module
 
   def get_norm_layer(self, num_features: int):
     """get normalization layer (return type inherits from nn.Module)"""
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -252,6 +252,96 @@ def deepstack_process(hidden_states, bidirectional_mask, visual_embeds):
   return hidden_states
 
 
+def _run_stage_layer_with_remat(
+    graphdef,
+    params,
+    state,
+    inputs,
+    run_fn,
+    *,
+    host_offload,
+    remat_policy,
+    prevent_cse,
+):
+  """Run one pipeline-stage layer under ``jax.checkpoint`` with params-only host-offload.
+
+  Shared by ``NNXSequentialPipelineStage`` and ``NNXScannedPipelineStage``, whose per-layer
+  remat bodies were byte-identical. The layer must already be split into ``(graphdef, params,
+  state)`` so the host-offload targets params only.
+  Inside a single ``jax.checkpoint(policy=remat_policy, prevent_cse=prevent_cse)`` this:
+
+    1. optionally offloads params to host memory via ``jax.device_put(device_space())``,
+    2. re-merges the layer with ``nnx.merge(graphdef, params, state)``,
+    3. runs it through ``run_fn(merged_layer, inputs)`` -- ``run_fn`` captures the call site's
+       ``decoder_segment_ids`` / ``decoder_positions`` / ``deterministic`` / ``model_mode`` / kwargs,
+    4. unwraps a tuple layer output to its first element.
+
+  Returns ``(out, new_state)``; the caller round-trips ``new_state`` back via ``nnx.update``.
+  """
+
+  def pure_fn(params_in, state_in, x_in):
+    if host_offload:
+      params_in = jax.tree.map(lambda x: jax.device_put(x, max_utils.device_space()), params_in)
+    merged = nnx.merge(graphdef, params_in, state_in)
+    out_inner = run_fn(merged, x_in)
+    out_inner = out_inner[0] if isinstance(out_inner, tuple) else out_inner
+    return out_inner, nnx.state(merged)
+
+  checkpointed_fn = jax.checkpoint(pure_fn, policy=remat_policy, prevent_cse=prevent_cse)
+  return checkpointed_fn(params, state, inputs)
+
+
+@functools.lru_cache(maxsize=None)
+def _make_single_layer_remat_stage_cls(base_stage_cls):
+  """Single-layer pipeline-stage subclass of base_stage_cls: applies per-stage jax.checkpoint
+  (+params-only host-offload) around the base layer's own __call__ WITHOUT a wrapper module, so
+  nnx.split(stage, nnx.Param, ...) yields param paths identical to a bare layer (top-level, no
+  'layers_0') . Cached per base class for stable nnx type identity."""
+
+  class NNXSingleLayerRematPipelineStage(base_stage_cls):
+    """Single decoder layer + per-stage remat/host-offload, params kept top-level (Linen parity)."""
+
+    def __init__(self, config, mesh, quant, model_mode, *, rngs, remat_policy=None, apply_remat=True):
+      super().__init__(config=config, mesh=mesh, quant=quant, model_mode=model_mode, rngs=rngs)
+      # Static (non-Variable) attrs -> graphdef, never the param tree. apply_remat decoupled from
+      # remat_policy (None == full rematerialization, matches Linen nn.remat(policy=None)).
+      self.apply_remat = apply_remat
+      self.remat_policy = remat_policy
+      self.prevent_cse = maxtext_utils.should_prevent_cse_in_remat(config)
+
+    def __call__(self, inputs, decoder_segment_ids, decoder_positions, deterministic, model_mode, **kwargs):
+      if not self.apply_remat:
+        return base_stage_cls.__call__(
+            self, inputs, decoder_segment_ids, decoder_positions, deterministic, model_mode, **kwargs
+        )
+
+      def run_fn(merged_layer, x_in):
+        # BASE __call__ explicitly: merged_layer is a THIS-subclass instance; merged_layer(...) recurses.
+        return base_stage_cls.__call__(
+            merged_layer, x_in, decoder_segment_ids, decoder_positions, deterministic, model_mode, **kwargs
+        )
+
+      graphdef, params, rest = nnx.split(self, nnx.Param, ...)
+      out, new_state = _run_stage_layer_with_remat(
+          graphdef,
+          params,
+          rest,
+          inputs,
+          run_fn,
+          host_offload=self.config.parameter_memory_host_offload,
+          remat_policy=self.remat_policy,
+          prevent_cse=self.prevent_cse,
+      )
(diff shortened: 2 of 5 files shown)


src/maxtext/layers/decoders.py
```diff
@@ -582,52 +583,49 @@ def _build_nnx_pipeline_stage(self, decoder_blocks, rngs):
...
-          mesh=self.mesh,
-          quant=self.quant,
-          model_mode=self.model_mode,
+          self.quant,
+          self.model_mode,
+          rngs=rngs,
+          remat_policy=per_stage_remat,
+          apply_remat=apply_per_stage_remat,
       )
-    return stage_module
```
mesakhcienet, line 615: remove `get_layer_to_pipeline` dead code (unused anymore)

src/maxtext/layers/nnx_decoders.py
```diff
@@ -669,6 +716,34 @@ def _init_scanned_gemma4(self, decoder_block_classes, rngs, mesh):
...
+
+    # 1. Prefix (hash-routing) layers: unrolled, heterogeneous attention per layer_idx.
+    self.num_prefix_layers = config.first_num_hash_layers
+    for layer_idx in range(self.num_prefix_layers):
+      self._create_and_register_named_layer(
+          deepseek4.DeepSeek4DecoderLayer, rngs, "layers", layer_idx, layer_idx=layer_idx
+      )
+
+    # 2. Scanned alternating HCA(128)/CSA(4) blocks.
+    num_full_blocks = (config.num_decoder_layers - self.num_prefix_layers) // 2
```
NuojCheng: add assertion making sure `config.num_decoder_layers - self.num_prefix_layers` is even

src/maxtext/layers/nnx_decoders.py
```diff
@@ -346,6 +377,19 @@ def __call__(
...
+
+        def pure_fn(params_in, state_in, x_in):
+          if self.config.parameter_memory_host_offload:
+            params_in = jax.tree.map(lambda x: jax.device_put(x, max_utils.device_space()), params_in)
+          inner_layer = nnx.merge(graphdef, params_in, state_in)
+          out = inner_layer(x_in, decoder_segment_ids, decoder_positions, deterministic, model_mode, **kwargs)
+          out = out[0] if isinstance(out, tuple) else out
+          return out, nnx.state(inner_layer)
+
+        checkpointed_fn = jax.checkpoint(pure_fn, policy=self.remat_policy, prevent_cse=self.prevent_cse)
```
NuojCheng: if this pure_fn+checkpointed_fn appears multiple times, e.g. line 301 - 309, maybe we can move them to a separate function util?

tests/unit/nnx_pipeline_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _run_pipeline(config, stage_factory):
+  devices_array = maxtext_utils.create_device_mesh(config)
+  mesh = Mesh(devices_array, config.mesh_axes)
+  inputs, seg, positions = _inputs(config)
+  my_pipeline = pipeline.create_pipeline(config=config, layers=stage_factory, mesh=mesh)
+  with jax.set_mesh(mesh), nn_partitioning.axis_rules(config.logical_axis_rules):
+    params = my_pipeline.init(jax.random.PRNGKey(0), inputs, seg, positions, True, MODEL_MODE_TRAIN)
+    out = my_pipeline.apply(params, inputs, seg, positions, True, MODEL_MODE_TRAIN)
```
NuojCheng, line 107: could you add backward pass check, e.g. jax.grad.

tests/unit/nnx_decoders_test.py
```diff
@@ -951,3 +964,338 @@ def mock_donor_idx(lyr, layer_types, num_kv_shared):
...
+    segment_ids = jnp.full((batch, seq_len), DECODING_ACTIVE_SEQUENCE_INDICATOR)
+    positions = jnp.broadcast_to(jnp.arange(seq_len)[None], (batch, seq_len))
+    logits, _, _ = decoder(
+        shared_embedding,
+        ids,
+        positions,
+        decoder_segment_ids=segment_ids,
+        deterministic=True,
+        model_mode=MODEL_MODE_TRAIN,
+    )
```
NuojCheng, line 1559: backward pass check is missing

tests/unit/nnx_decoders_test.py
```diff
@@ -951,3 +964,338 @@ def mock_donor_idx(lyr, layer_types, num_kv_shared):
...
+    self.assertIsNotNone(decoder.scanned_blocks)
+
+  def test_non_scan_init_builds_deepseek4_layers(self):
+    """non-scan init builds num_decoder_layers DeepSeek4DecoderLayer instances (with layer_idx)."""
+    from maxtext.models import deepseek4  # pylint: disable=import-outside-toplevel
+
+    cfg = self._make_ds4_config(scan_layers=False)
+    mesh = _make_mesh(cfg)
+    decoder = NNXDecoder(config=cfg, mesh=mesh, model_mode=MODEL_MODE_TRAIN, rngs=nnx.Rngs(params=0, dropout=1))
+    self.assertEqual(len(decoder.layers), cfg.num_decoder_layers)
```
NuojCheng: these are very weak checks. If we only care about shapes/length, instead of real values, we could use CPU AOT test. We don't need real TPU devices to run these tests.

tests/unit/nnx_decoders_test.py
```diff
@@ -951,3 +964,338 @@ def mock_donor_idx(lyr, layer_types, num_kv_shared):
...
+  def test_forward_scan(self):
+    """deepseek4 scan forward (unrolled hash-routing prefix + scanned HCA/CSA blocks).
+
+    End-to-end numeric forward exercising _apply_deepseek4_scanned_blocks: the prefix layers run
+    unscanned, then the alternating HCA(128)/CSA(4) blocks run via _apply_layers_sequentially.
+    """
+    cfg = self._make_ds4_config(scan_layers=True)
+    _, logits, expected = self._build_and_run(cfg)
+    self.assertEqual(logits.shape, expected)
+    self.assertTrue(jnp.all(jnp.isfinite(logits)))
```
NuojCheng: jnp.isfinite is also a weak test. We could use AOT test for similar functionalities

NuojCheng (commented):
Nice changes! Some comments in the test section, I feel the tests are weak, mostly checking shapes/isinfinite, which can be replaced by CPU AOT tests. Also backward pass tests seems missing. Can we compare gradients values say with/without using the selected code path, e.g. https://github.com/AI-Hypercomputer/maxtext/blob/d30c749e036aefbbebb92954bb7345ec967d6b06/tests/unit/moe_test.py#L604-L723.

mesakhcienet:
Hi @NuojCheng 
Thank you for the review! Really apprciate!

Quick summary of the gradient-parity test change.

**What we changed.** We now do the exact gradient check on CPU only (CI's `cpu-unit` runs it on every
PR). On GPU/TPU we just check the gradients are finite, non-zero, and the loss matches. The previous
version widened the tolerances instead, which either still failed on TPU or got loose enough to hide
real bugs. Only the test file changed — no production code.

**Why.** That's only exactly true on CPU. On TPU the recomputed backward runs its matmuls in
bfloat16 and rounds a little didferrently, so the gradients drift — a little for the dense layers, and
by a big jump for DeepSeek-V4 when a top-k routing tie flips to the other expert. It's normal
accelerator rounding, not a real bug: the gradients still match on CPU at a tight tolerance, so the
backwadr math is correct.

**What was failing.** The remat gradient tests on `tpu-unit`. They check that turning
rematerialization on vs off gives the same gradients.


**Things we tried that didn't work:**

- Loosening the tolerances — hides real regressions, and still failed on TPU.
- `matmul_precision="highest"` — on TPU that's still bfloat16 under the hood, so it doesn't help.
- Saving the matmuls (`dots_saveable`) — checked on TPU, gradients still drift ~38%.

If you have anny suggestions/idea — a better way to check the backward on TPU, or a tolerance you'd prefer, or better implementation—
happy to change it. Thanks!

tests/unit/nnx_pipeline_test.py
mesakhcienet: i have added them on each individual unit test cases. Thanks!

mesakhcienet:
I have moved the check so that it now runs on all platforms, including TPU. 

Before proceeding further, I would be happy to hear your thoughts and suggestions on the current approach. Below is the context on what was failing and how it is currently addressed.

# What was failing
The tests initially used an exact per-element check (`assert_allclose` with `rtol=atol=1e-2`). While this passes on CPU, it fails on TPU because the rematerialized backward pass recomputes matmuls in `bfloat16` and rounds them differently. In DeepSeek-V4, a top-k routing tie can flip during this recomputation, shifting individual gradient elements by about 4.

We tried several tolerance configurations, but none of them worked reliably:
  `1e-2 / 1e-2` (original): Fails on TPU.
  `5e-2 / 2.0`: Fails.
  `3e-1 / 3.0` (DeepSeek-V4) + `5e-1 / 20.0` (dense stages): Still fails, as the flip exceeds `atol=3`. Raising the atol high enough to pass (~20) would make the test too loose to catch actual gradient regressions.
  
We also tried using float32 activations and setting `matmul_precision="highest"`, but on TPU "highest" still uses `bfloat16`, so the routing flip remained.

# Current Fix (already pushed on current branch)
Instead of a per-element check, the current implementation bounds the aggregate relative L2 error over the entire gradient to under 5% on all platforms:
`|g_remat - g_no_remat| / |g_no_remat| < 0.05`

A functional backward regression typically moves this norm by around 100%, whereas the `bfloat16` noise and isolated routing flips have a minimal impact.

On TPU (v6e-8), we observed the following relative errors:
  DeepSeek-V4: 0.29% (with a max single-element diff of 4.2)
  Dense pipeline stages: Up to 1.2% on the scanned stage (with a max single-element diff of 16.7)
  Host-offload: 0%
  CPU: 0%

The individual outlier differences are relatively large, but they wash out in the overall norm check. 

Please let me know if you have any suggestions on this validation method or if you would prefer a different approach. I am happy to adjust based on your feedback. Thank you!

src/maxtext/layers/pipeline.py
```diff
@@ -891,32 +891,18 @@ def unbox_val(x):
...
-    # it is being carried through scan instead of broadcast.
-    # NOTE: is_leaf stops jax.tree.leaves from traversing *into* Variable nodes,
-    # so we see actual Variable instances (not raw arrays).
-    assert all(
-        isinstance(v, nnx.RngState)
-        for v in jax.tree.leaves(layers_mutables, is_leaf=lambda x: isinstance(x, nnx.Variable))
-        if isinstance(v, nnx.Variable)
-    ), (
-        "Non-RngState variable found in layers_mutables catch-all partition. "
-        "Only RngState variables (RngKey/RngCount) should be present."
```
NuojCheng, line 910: could you help me understand why the assertion is removed?

mesakhcienet: The assertion was originally added as a safety check. In the old 3-way split (line 894), `layers_mutables` was the catch-all bucket, and we carried it through the `jax.lax.scan` loop. The assert ensured we didn't accidentally carry `non_trainable` variables (like BatchStat) through the scan, which would cause issues.

  However, with models like `DeepSeek-V4`, the layers now legitimately contain `non_trainable` variables (specifically, the static hash routing tables used in MoE). These fell into the `layers_mutables` catch-all and triggered the assertion crash.

  To fix this and support `non_trainable` variables, I updated the code to use a 4-way split: 
   1. We explicitly extract `nnx.RngState` into its own bucket (`layers_rng`), which is the only thing we carry through the loop. 
   2. The remaining variables fall into `layers_non_trainable`, which we safely broadcast (loop-invariant) across the scan and discard the output copy.

  Because we now explicitly separate the RNG state from the non-trainable state and handle both correctly, the old assertion is no longer needed.

mesakhcienet:
Hi @NuojCheng — pushed an update and rebased onto latest main. Ready for re-review.

**Summary:** You were right on both points you raised. I was wrong on both. Two of my earlier answers
are retracted below. Acting on your "weak checks" comment found 15 tests that were not testing
anything.

---

## Correction 1 — the assertion removal was hiding a real bug

**What I said:** the non-trainable state is "safely broadcast, and the output copy discarded."

**What is actually true:** discarding that copy silently loses mutations.

I ran a stage that adds 1 to a non-trainable value on every call, starting from 0:

| Schedule | Result |
|---|---|
| non-circular | `[0.0, 0.0, 0.0, 0.0]` — **mutation lost, no error** |
| circular | `[8, 9, 10, 11, ...]` — mutation kept |

The output was still finite and the right shape. Nothing warned us.

**Linen does this correctly.** It carries the value when the collection is mutable
(`53dea32b7:898-900`). So this was a regression against Linen, introduced by this PR.

**I also had the cost backwards.** I avoided carrying it because I thought it would stack per
iteration. It does not. A `jax.lax.scan` carry is threaded, not stacked. Measured on TPU with a real
decoder layer: **0.00 MB extra** at both 4 KB and 4 MB of state.

**Fixed**, with a regression test that fails on the old code.

---

## Correction 2 — my gradient explanation named the wrong cause

**What I said:** bf16 recompute plus DeepSeek Top-K router flips.

**What the measurements show:**

| Stage | rel_L2 |
|---|---|
| Sequential | **0.00e+00** |
| Single-layer | **0.00e+00** |
| Scanned | 1.17e-02 |

Three things are wrong with my explanation:

1. **That test has no router.** The config is `num_experts=1`, `decoder_block=llama2`.
2. **Recompute is not the cause.** `everything_saveable` recomputes nothing, and shows the same
   deviation. In full float32 the deviation is exactly zero. It is a precision effect.
3. **Two of the three stages are bit-identical.** They never needed a loose tolerance at all.

---

## "L2 norm is too weak" — you were right, and it is now fixed

**The problem.** The old check was a single aggregate: `||g_a - g_b|| / ||g_b||` over all gradients
concatenated. That is dominated by whichever tensor carries most of the norm.

Measured on the scanned stage (8 gradient leaves; these are norm *fractions*, so they combine in
quadrature and do not sum to 1):

- the largest leaf holds 94.1% of the gradient norm → it must be ~5.3% wrong before the aggregate moves
- the smallest holds 5.6% → it can be **89.2% wrong and the aggregate never notices**

So a bug confined to a bias or a layer-norm gain was invisible.

**What I changed.** `_assert_grad_parity` now applies **two** bounds instead of one:

| Bound | Before | After |
|---|---|---|
| Aggregate | 5% | **2%** |
| Per-leaf | none | **added** |

Two different quantities are involved below, so to be unambiguous: **aggregate error** is over all
gradients concatenated; **per-leaf error** is `||d_leaf|| / ||leaf||` for one tensor. They are not
comparable, and the small numbers below are aggregate while the large ones are per-leaf.

The 2% aggregate is measured, not guessed: the worst real *aggregate* deviation anywhere is 1.17e-2,
on the scanned stage only. Every other stage is bit-identical.

The per-leaf bound is deliberately **not** a plain relative check. It is a combined absolute+relative
criterion, analogous to `np.allclose` but norm-based rather than elementwise:

```
||g_other - g_ref||  >  5% * ||g_ref_leaf||   +   1e-3 * ||g_total||
                        \_ relative term _/       \_ absolute term _/
```

**Why the absolute term is required — and this is the part I got wrong first.** A relative-only bound
is unusable for tiny tensors. Many leaves (scalars, per-head biases) sit at 1e-6..1e-5 of the total
gradient norm. bf16 rounding gives those a *per-leaf* error of ~1% on the TPU I develop on and 24.5%
on the CI TPU — while their contribution to the aggregate is only 1e-9..1e-7. I shipped a
relative-only version first and it failed CI on exactly such a leaf (`shape (1,)`, 0.00% of the norm).
That is a hardware-dependent false alarm, not a bug, which is why the absolute term exists.

`1e-3` is chosen from that measurement: ~100x above the largest error any noise leaf contributes
(1e-7..1e-5 of total) and ~50x below a genuine defect (the 5.6% leaf case contributes 5.0e-2).

**Where the bound does and does not fire.** Swept, not asserted — leaf size as a fraction of total
gradient norm, against per-leaf relative error:

| leaf frac | 5% wrong | 20% wrong | 50% wrong | 100% wrong |
|---|---|---|---|---|
| 1e-06 | tolerated | tolerated | tolerated | tolerated |
| 1e-05 | tolerated | tolerated | tolerated | tolerated |
| 1e-04 | tolerated | tolerated | tolerated | tolerated |
| 1e-03 | tolerated | tolerated | tolerated | tolerated |
| **1e-02** | tolerated | **caught** | **caught** | **caught** |
| **5e-02** | tolerated | **caught** | **caught** | **caught** |

**Stated limitation:** a leaf holding ≤0.1% of the total gradient norm can be arbitrarily wrong and
will not be flagged. That is deliberate — such a leaf can shift the parameter update by at most 0.1%,
below the bf16 noise floor — but it is a real limit and I would rather state it than have you find it.
If you want that tightened, lowering `per_leaf_atol_frac` is a one-line change; the cost is
reintroducing hardware-dependent failures.

Note that tightening the aggregate alone would **not** have fixed the original problem: in the 3%-leaf
case the aggregate error is 1.199%, which passes even the new 2% bound. The per-leaf term is what
catches it.

No false failures: decoder suite green on CPU and TPU, including the stages that show the real bf16
deviation.

---

## Your "weak checks" comment found the most

I mutation-tested the suite. The method: break the code a test covers, then check the test turns red.
A test that stays green is protecting nothing.

**15 tests stayed green while the thing they tested was broken.**

**8 were in this PR.** All fixed, each verified by mutation:

- pipeline forward/backward tests passed with **stage-to-stage chaining completely cut**
- circular remat tests passed with **remat removed entirely**
- a DeepSeek4 test only checked `isinstance`, while every layer's `layer_idx` silently became `-1`
- one test compared `remat_policy='full'` vs `'minimal'` on the **same** object — both had the same
  bug, so they always agreed

**7 were pre-existing.** I fixed the worst one: making the decoder ignore its input tokens entirely
left the whole suite green. The rest are outside this PR — I will do them in a follow-up.

**Two blind spots, both closed:**

- nothing detected rematerialisation being turned off
- the stage-output unwrap had zero coverage

Both now check the traced jaxpr by primitive identity, not by string match. (`remat2` is a legacy
name, and JAX has renamed a sibling primitive before.)

---

## Four source fixes

| | Problem | How it was verified |
|---|---|---|
| **F1** | non-trainable state silently discarded | regression test red → green; 0.00 MB cost |
| **F2** | tuple-unwrap applied to only one of the two pipelines | fixed two reproducible crashes |
| **F3** | `scan_pipeline_repeats` was never read — scanned and unrolled compiled identically | now mirrors Linen; HLO differs |
| **F4** | circular per-iteration remat was hardcoded on; Linen gated it (`53dea32b7:1365`) | flag now works; loss and gradients bit-identical |

---

## Verified clean

- **DeepSeek4 decoder on TPU**, with the megablox kernel confirmed running: 306/306 parameters match,
  `rel_L2 = 0.000e+00`.
- **all-False vs all-True** on llama2: bit-identical.
- **First Linen-vs-NNX pipeline comparison:** the non-circular pipeline is bit-identical to the
  pre-migration Linen pipeline.

---

## How to verify

```bash
XLA_FLAGS=--xla_force_host_platform_device_count=4 JAX_PLATFORMS=cpu \
  python3 -m pytest tests/unit/nnx_pipeline_test.py -q                    # 15 passed
JAX_PLATFORMS=cpu python3 -m pytest tests/unit/nnx_decoders_test.py -q    # 76 passed, 4 skipped
JAX_PLATFORMS=tpu python3 -m pytest tests/unit/nnx_{pipeline,decoders}_test.py -q      # 95 passed
JAX_PLATFORMS=tpu python3 -m pytest tests/integration/pipeline_parallelism_test.py -q  # 15 passed
```

---

Thanks for pushing back on both points. The non-trainable bug only surfaced because you asked.

PR #4289, opened 2026-06-29
Fix TPU Splash local sliding window size

huytransformer:
# Description

Dense local sliding keeps exactly `sliding_window_size` tokens. Splash `LocalMask` uses an inclusive left window, so passing `sliding_window_size` keeps 1 extra token. This changes the Splash window to `sliding_window_size - 1`.

# Tests
Ran successfully
```bash
python3 -m pytest tests/unit/attention_test.py -k SplashLocalMaskTest
```

# Checklist

### DIFF
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -1286,7 +1286,7 @@ class AttentionOp(nnx.Module):
         raise ValueError("Sliding_window_size must be set if Local Sliding attention type")
       mask &= mask_module.LocalMask(
           shape=(query.shape[2], key.shape[2]),
-          window_size=(self.sliding_window_size, self.sliding_window_size),
+          window_size=(self.sliding_window_size - 1, self.sliding_window_size),
           offset=0,
       )
     elif self.attention_type == AttentionType.CHUNK:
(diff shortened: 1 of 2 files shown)


src/maxtext/layers/attention_op.py
```diff
@@ -1286,7 +1286,7 @@ def create_sa_config(config, query, key, attn_logits_soft_cap):
         raise ValueError("Sliding_window_size must be set if Local Sliding attention type")
       mask &= mask_module.LocalMask(
           shape=(query.shape[2], key.shape[2]),
-          window_size=(self.sliding_window_size, self.sliding_window_size),
+          window_size=(self.sliding_window_size - 1, 0),
```
NuojCheng: I am not fully confident this change is valid. Why it is 0 for the 2nd index? cc @gagika for wisdom on local mask

huytransformer: because when we logical AND with causal mask the right side doesn't matter; we could leave the right side as it was if that would make things more clear

huytransformer: Done : )

igorts-git (approved):

src/maxtext/layers/attention_op.py
aireenmei: The Gemma4 implementation in Tunix (https://github.com/google/tunix/blob/main/tunix/models/gemma4/model.py#L925-L929) uses window_size=(self.sliding_window_size - 1, 0)

NuojCheng (approved):

PR #4290, opened 2026-06-29
ci: add test duration tracking workflow

hsuan-lun-chiang:
# Description

This PR implements a comprehensive automated mechanism to detect and prevent test suite run-time inflation before code is merged into `main`, addressing the need for a fast, reliable, and responsive CI pipeline.

### How the Workflow Works:
When a PR is opened or updated, the pipeline runs the standard test suites (CPU, GPU, TPU) and generates JUnit XML reports. Once those tests complete, the new `track_performance` job runs and does the following:

1. **Micro-Level Tracking**: It runs `enforce_test_limits.py` to parse the XML reports and ensures no *individual* test function regresses significantly compared to its stored baseline. The fine-grained baseline (`per_test_baseline.json`) is maintained directly on the `gh-pages` branch.
2. **Macro-Level Tracking**: It runs `parse_junit_to_benchmark.py` to aggregate the *total* runtime for each device. It then uses `github-action-benchmark` to compare these aggregated total runtimes against the baseline stored on the `gh-pages` branch to prevent overall suite inflation.

### Changes Made:
1. **Macro-Level Suite Tracking (`gh-pages`)**: 
   - Integrated `github-action-benchmark` to track macro-level execution times (suite/shard level).
   - Added `tests/utils/parse_junit_to_benchmark.py` to convert JUnit XML test results into the JSON format required by the benchmark action.
   - Results are automatically pushed to the `gh-pages` branch for historical trend visualization.

2. **Micro-Level Individual Test Tracking**:
   - Uses a custom, highly reliable Python-based baseline tracking mechanism.
   - Saves a fine-grained `per_test_baseline.json` dictionary containing the execution times of individual test functions to the `gh-pages` branch.

3. **Regression Enforcement**:
   - Added `tests/utils/enforce_test_limits.py` to automatically flag and block any test that significantly regresses compared to its stored per-test baseline.
   - **Relative Regression Limits**: Any test that is **50% slower** (for Train Compile tests) or **20% slower** (for Unit and Integration tests) than its baseline AND exceeds its absolute noise threshold (+15s for Unit, +120s for Train Compile, +30s for Integration) will be blocked. This protects against JAX JIT compilation jitter on shared CI runners while keeping non-compile tests tightly constrained.
  
# Tests

- Validated macro-level tracking via `github-action-benchmark` in CI.
- Validated micro-level test baseline generation, caching, and evaluation in CI.
- Tested `enforce_test_limits.py` in CI to ensure that test jitter thresholds are accurately tuned.

# Checklist

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -416,3 +416,15 @@ jobs:
     with:
       failed_run_id: '${{ github.run_id }}'
     secrets: inherit
+
+  track_performance:
+    name: Track Test Performance
+    needs: [tpu-tests, gpu-tests, cpu-tests]
+    if: ${{ always() && (needs.cpu-tests.result == 'success' || needs.gpu-tests.result == 'success' || needs.tpu-tests.result == 'success') }}
+    uses: ./.github/workflows/track_performance.yml
+    permissions:
+      contents: write
+      id-token: write
+      pull-requests: write
+    secrets: inherit
+
--- a/.github/workflows/run_tests_against_package.yml
+++ b/.github/workflows/run_tests_against_package.yml
@@ -19,6 +19,9 @@ name: Run Tests Against MaxText Package
 on:
   workflow_call:
     inputs:
+      flavor:
+        required: true
+        type: string
       device_type:
         required: true
         type: string
@@ -211,6 +214,7 @@ jobs:
               -v \
               -m "${FINAL_PYTEST_MARKER}" \
               --durations=0 \
+              --junitxml=test-results-${INPUTS_FLAVOR}-${INPUTS_WORKER_GROUP}.xml \
               $PYTEST_COV_ARGS \
               $SPLIT_ARGS \
               ${INPUTS_PYTEST_EXTRA_ARGS}
@@ -220,6 +224,7 @@ jobs:
           INPUTS_IS_SCHEDULED_RUN: ${{ inputs.is_scheduled_run }}
           INPUTS_PYTEST_MARKER: ${{ inputs.pytest_marker }}
           INPUTS_DEVICE_TYPE: ${{ inputs.device_type }}
+          INPUTS_FLAVOR: ${{ inputs.flavor }}
           INPUTS_PYTEST_ADDOPTS: ${{ inputs.pytest_addopts }}
           INPUTS_TOTAL_WORKERS: ${{ inputs.total_workers }}
           INPUTS_WORKER_GROUP: ${{ inputs.worker_group }}
@@ -243,3 +248,11 @@ jobs:
           # If scheduled, upload to scheduled flag only. If PR, upload to regular flag only.
           flags: ${{ inputs.is_scheduled_run == 'true' && 'scheduled' || 'regular' }}
           verbose: true
+      - name: Upload Test Results XML
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-results-${{ inputs.flavor }}-${{ inputs.worker_group }}-${{ github.run_id }}
+          path: test-results-*.xml
+          retention-days: 1
+          if-no-files-found: ignore
--- a/tests/utils/process_test_results.py
+++ b/tests/utils/process_test_results.py
@@ -1,0 +1,243 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Enforces test limits and generates macro-level benchmark metrics in a single pass."""
+
+import argparse
+import glob
+import json
+import os
+import sys
+import xml.etree.ElementTree as ET
+
+# Relative regression threshold ratio (1.20 = 20% slower)
+REL_REGRESSION_OTHER_RATIO = 1.20
+
+# Absolute increase thresholds to filter out noise
+ABS_INCREASE_UNIT_SEC = 15.0
+ABS_INCREASE_INTEGRATION_SEC = 30.0
+
+
+def process_testcase(testcase, xml_file, baseline_data, new_baseline_data):
+  """Processes a single testcase and checks for limit violations or regressions."""
+  failed = False
+
+  # 1. Skip processing for skipped tests to avoid corrupting the baseline with ~0s durations
+  if testcase.find("skipped") is not None:
+    return False
+
+  # 2. Skip processing for failed/errored tests to prevent capturing abnormally short durations
+  if testcase.find("failure") is not None or testcase.find("error") is not None:
+    return False
+
+  time_val = float(testcase.get("time", 0.0))
+  name = testcase.get("name", "unknown")
+  classname = testcase.get("classname", "unknown")
+  full_name = f"{classname}.{name}"
+
+  # Parse custom properties to extract markers
+  markers = set()
+  properties_elem = testcase.find("properties")
+  if properties_elem is not None:
+    for prop in properties_elem.findall("property"):
+      if prop.get("name") == "marker":
+        markers.add(prop.get("value"))
+
+  is_integration = "integration_test" in markers
+  is_cpu = "cpu" in os.path.basename(xml_file).lower()
+
+  if is_integration:
+    abs_noise_threshold = ABS_INCREASE_INTEGRATION_SEC
+    rel_regression_ratio = REL_REGRESSION_OTHER_RATIO
+    test_type = "Integration Test"
+  else:
+    abs_noise_threshold = ABS_INCREASE_UNIT_SEC
+    rel_regression_ratio = REL_REGRESSION_OTHER_RATIO
+    test_type = "Unit Test"
+
+  new_baseline_data[full_name] = time_val
+
+  # Skip regression checking for CPU tests due to shared CPU multi-tenancy noise
+  skip_regression = is_cpu
+
+  # Check relative regression if baseline exists
+  if not skip_regression and full_name in baseline_data:
+    base_time = baseline_data[full_name]
+    if base_time > 0:
+      ratio = time_val / base_time
+      increase = time_val - base_time
+      if ratio >= rel_regression_ratio and increase > abs_noise_threshold:
+        print(f"::error::[REGRESSION ALERT] {test_type} significantly degraded!")
+        print(f"  Test: {full_name}")
+        print(f"  File: {os.path.basename(xml_file)}")
+        print(f"  Previous Duration: {base_time:.2f}s")
+        print(f"  New Duration: {time_val:.2f}s")
+        print(f"  Increase: +{increase:.2f}s ({(ratio - 1) * 100:.1f}%)")
+        print(f"  Thresholds: >{(rel_regression_ratio - 1) * 100:.0f}% AND >{abs_noise_threshold}s")
+        print("-" * 50)
+        failed = True
+
+  return failed
+
+
+def main():
+  """Parses XML files and processes test results."""
+  parser = argparse.ArgumentParser(
+      description="Enforce test limits, check regressions, and parse JUnit XML to Benchmark format."
+  )
+  parser.add_argument("xml_dir", help="Directory containing JUnit XML files")
+  parser.add_argument("--baseline", type=str, help="Path to the baseline JSON file", default=None)
+  parser.add_argument(
+      "--save-baseline",
+      type=str,
+      help="Path to save the new baseline JSON file",
+      default=None,
+  )
+  parser.add_argument(
+      "--warn-only",
+      action="store_true",
+      help="Report regressions but do not return a non-zero exit code",
+  )
+  parser.add_argument(
+      "--output-benchmark",
+      type=str,
+      help="Path to save the benchmark JSON file",
+      default=None,
+  )
+  args = parser.parse_args()
+
+  xml_files = glob.glob(os.path.join(args.xml_dir, "*.xml"))
+  if not xml_files:
+    print(f"No XML files found in {args.xml_dir}")
+    sys.exit(0)
+
+  baseline_data = {}
+  if args.baseline and os.path.exists(args.baseline):
+    try:
+      with open(args.baseline, "r", encoding="utf-8") as f:
+        baseline_data = json.load(f)
+    except Exception as e:  # pylint: disable=broad-exception-caught
+      print(f"Error loading baseline {args.baseline}: {e}")
+
+  # Initialize with a copy of baseline_data to preserve history for tests not run or skipped
+  new_baseline_data = dict(baseline_data)
+  has_regression = False
+  has_errors = False
+
+  total_times_by_job = {}
+  total_tests_by_job = {}
+
+  for xml_file in xml_files:
+    basename = os.path.basename(xml_file)
+    parts = basename.replace(".xml", "").split("-")
+    if len(parts) >= 4 and parts[0] == "test" and parts[1] == "results":
+      job_name = "-".join(parts[2:-1])
+    else:
+      # Fallback to device type if naming is simple (e.g. test-results-cpu-1.xml)
+      if len(parts) >= 3:
+        job_name = parts[2]
+      else:
+        job_name = "unknown"
+
+    try:
+      tree = ET.parse(xml_file)
+      root = tree.getroot()
+
+      job_time = 0.0
+      job_count = 0
+
+      for testcase in root.iter("testcase"):
+        job_count += 1
+        time_val = float(testcase.get("time", 0.0))
+        job_time += time_val
+
+        # Micro-level regression check
+        if process_testcase(testcase, xml_file, baseline_data, new_baseline_data):
+          has_regression = True
+
+      if job_name != "unknown" or job_count > 0:
+        total_times_by_job[job_name] = total_times_by_job.get(job_name, 0.0) + job_time
+        total_tests_by_job[job_name] = total_tests_by_job.get(job_name, 0) + job_count
+
+    except Exception as e:  # pylint: disable=broad-exception-caught
+      print(f"Error parsing or processing {xml_file}: {e}")
+      has_errors = True
+
+  # Output macro-level benchmark JSON if requested
+  if args.output_benchmark:
+    benchmarks = []
+    for job, total_time in total_times_by_job.items():
+      benchmarks.append(
+          {
+              "name": f"Total {job.upper()} Tests Duration",
+              "unit": "sec",
+              "value": total_time,
+          }
+      )
+      benchmarks.append(
+          {
+              "name": f"Total {job.upper()} Tests Count",
+              "unit": "count",
+              "value": total_tests_by_job.get(job, 0),
+          }
+      )
+
+    try:
+      dirname = os.path.dirname(args.output_benchmark)
+      if dirname:
+        os.makedirs(dirname, exist_ok=True)
+      with open(args.output_benchmark, "w", encoding="utf-8") as f:
+        json.dump(benchmarks, f, indent=2)
+      print(f"Saved benchmark results to {args.output_benchmark}")
+    except Exception as e:  # pylint: disable=broad-exception-caught
+      print(f"Error saving benchmark results: {e}")
+      has_errors = True
+
+  # Save baseline only if no critical execution/parsing errors occurred
+  if args.save_baseline:
+    if has_errors:
+      print(
+          "Warning: Critical errors occurred during parsing. "
+          "Skipping saving of the new baseline to prevent corruption."
+      )
+    else:
+      try:
+        dirname = os.path.dirname(args.save_baseline)
+        if dirname:
+          os.makedirs(dirname, exist_ok=True)
+        with open(args.save_baseline, "w", encoding="utf-8") as f:
+          json.dump(new_baseline_data, f, indent=2)
+        print(f"Saved new baseline to {args.save_baseline}")
+      except Exception as e:  # pylint: disable=broad-exception-caught
+        print(f"Error saving baseline {args.save_baseline}: {e}")
+        has_errors = True
+
+  # Determine final exit status
+  if has_errors:
+    print("\nOne or more critical errors occurred during execution.")
+    sys.exit(1)
+  elif has_regression:
+    print("\nOne or more tests regressed significantly.")
+    if args.warn_only:
+      print("Non-blocking mode active: exiting with code 0.")
+      sys.exit(0)
+    else:
+      sys.exit(1)
+  else:
+    print("All tests passed execution time and regression checks.")
+    sys.exit(0)
+
+
+if __name__ == "__main__":
+  main()
--- a/.github/workflows/run_tests_coordinator.yml
+++ b/.github/workflows/run_tests_coordinator.yml
@@ -109,6 +109,7 @@ jobs:
 
     uses: ./.github/workflows/run_tests_against_package.yml
     with:
+      flavor: ${{ inputs.flavor }}
       # Infrastructure Mapping
       device_type: >-
         ${{ fromJSON('{
--- a/.github/workflows/track_performance.yml
+++ b/.github/workflows/track_performance.yml
@@ -1,0 +1,86 @@
+# Copyright 2023-2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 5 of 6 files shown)


.github/workflows/run_tests_against_package.yml
```diff
@@ -227,3 +228,10 @@ jobs:
           # If scheduled, upload to scheduled flag only. If PR, upload to regular flag only.
           flags: ${{ inputs.is_scheduled_run == 'true' && 'scheduled' || 'regular' }}
           verbose: true
+      - name: Upload Test Results XML
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-results-${{ inputs.device_type }}-${{ inputs.worker_group }}
```
xibinliu: I think we should append `${{ github.run_id }}` to the artifact name. This is to avoid mixing files from different runs.

For download, we should download files of the same run_id.

Can we also add the expiration time of these temporary files, such as:

    retention-days: 1 # Automatically delete the artifact from the cloud after 24 hours

.github/workflows/ci_pipeline.yml
```diff
@@ -351,3 +351,68 @@ jobs:
...
+          name: MaxText Test Execution Times
+          tool: 'customSmallerIsBetter'
+          output-file-path: benchmark-results.json
+          external-data-json-path: ./cache/benchmark-data.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          alert-threshold: '150%'
+          comment-on-alert: true
+          fail-on-alert: true
+
+      - name: Upload Updated Baseline to GCS
```
xibinliu: Can we use gh-pages to store the benchmark data? github-action-benchmark can generate a index.html so we can see the interactive chart of the benchmark history.

.github/workflows/run_tests_against_package.yml
hsuan-lun-chiang: Thanks for the suggestions, I have addresses all of these points:

  1. Added  `${{ github.run_id }}`  to the artifact name in the upload step and updated the download pattern in  ci_pipeline.yml  to match.
  2. Ensure temporary files are automatically cleaned up.

.github/workflows/ci_pipeline.yml
hsuan-lun-chiang: Yes, we used a GCS bucket as a temporary method to test the core logic. Now we have configured the `github-action-benchmark` step to automatically push the generated index.html to the `gh-pages` branch (under the `dev/bench` directory)."

hsuan-lun-chiang: For the live github.io URL, we need to manually flips the switch in repo settings to tell GitHub to host the site. Until then, we can download the artifacts to view the interactive chart instead.

tests/utils/parse_junit_to_benchmark.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for testsuite in root.iter("testsuite"):
+      time_val = float(testsuite.get("time", 0.0))
+
+      # Recommend NOT adding individual file-level times to the benchmark action list
+      # to prevent rebalancing/sharding from triggering false regression alerts.
+      # Let Codecov handle the micro-level tracking.
+
+      total_times_by_device[device] = total_times_by_device.get(device, 0.0) + time_val
+
```
xibinliu: can we also track number of tests, so we know if a regresssion is caused by adding new tests?

We should be able see the number of tests and total execution time in the benchmark html page.

tests/utils/enforce_test_limits.py
```diff
@@ -0,0 +1,75 @@
+"""Enforces absolute maximum duration limits for unit and integration tests."""
+
+import xml.etree.ElementTree as ET
+import glob
+import os
+import sys
+
+UNIT_TEST_LIMIT_SEC = 80.0
+TRAIN_COMPILE_TEST_LIMIT_SEC = 1500.0
+INTEGRATION_TEST_LIMIT_SEC = 300.0
```
xibinliu: how are these number obtained?

.github/workflows/ci_pipeline.yml
xibinliu: To see the benchmark https://AI-Hypercomputer.github.io/maxtext/dev/bench/, @shralex could you enable the Build and deployment? According the gemini:

Step-by-Step Instructions to Flip the Switch
1. Navigate to Repository Settings: Go to the home page of your repository on GitHub (e.g., github.com/AI-Hypercomputer/maxtext) and click on the Settings tab at the top of the page.
2. Access the Pages Menu: In the left-hand sidebar menu, scroll down to the Code and automation section and click on Pages.
3. Configure the Build and Deployment Source: Under the Build and deployment section:
- For the Source dropdown, select Deploy from a branch.
4. Select the Target Branch: Under the Branch configuration:
- Click the branch dropdown (which defaults to None) and select gh-pages.
- Leave the folder dropdown set to / (root).
5. Save the Configuration: Click the Save button.

.github/workflows/ci_pipeline.yml
```diff
@@ -367,3 +351,56 @@ jobs:
...
+
+      - name: Parse JUnit XML to Benchmark format
+        run: |
+          mkdir -p ./cache
+          python3 tests/utils/parse_junit_to_benchmark.py test-results benchmark-results.json
+          echo "Parsed Benchmark Results:"
+          cat benchmark-results.json
+
+      - name: Track Test Durations
+        uses: benchmark-action/github-action-benchmark@v1
```
xibinliu: If the PR is modified, or abandoned, will the uploaded benchmark file benchmark-results.json be kept in the benchmark history? (which is undesired)

.github/workflows/ci_pipeline.yml
```diff
@@ -367,3 +351,56 @@ jobs:
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          alert-threshold: '115%'
+          comment-on-alert: true
+          fail-on-alert: ${{ github.ref != 'refs/heads/main' }}
+          auto-push: ${{ github.ref == 'refs/heads/main' }}
+          gh-pages-branch: 'gh-pages'
+          benchmark-data-dir-path: 'dev/bench'
+
+      - name: Upload Test Results to Codecov
+        uses: codecov/test-results-action@v1
```
xibinliu: save as above, uploading for each PR run is not desired. We should upload once every day from a scheduled run on the main branch.

.github/workflows/ci_pipeline.yml
hsuan-lun-chiang: The benchmark file is uploaded only when the workflow runs on the main branch.

.github/workflows/ci_pipeline.yml
shralex: Done, thanks for the detailed instructions!

tests/utils/enforce_test_limits.py
hsuan-lun-chiang: These limits were obtained empirically by profiling the maximum execution times from our recent CI runs.

tests/utils/enforce_test_limits.py
```diff
@@ -0,0 +1,75 @@
...
+      for testcase in root.iter("testcase"):
+        time_val = float(testcase.get("time", 0.0))
+        name = testcase.get("name", "unknown")
+        classname = testcase.get("classname", "unknown")
+        full_name = f"{classname}.{name}"
+
+        is_integration = (
+            "integration" in xml_file.lower()
+            or "tests/integration/" in full_name.lower()
+            or ".integration" in full_name.lower()
```
shralex: Can we rely on markers here instead of parsing the file name:

Gemini says this:

Since enforce_test_limits.py is a standalone script that runs after tests have completed, it only has access to the generated JUnit XML files and does not directly interact with Pytest. However, you can easily instruct Pytest to automatically write its markers into the JUnit XML as custom properties, which your parsing script can then read.

Here is the exact, two-step implementation to achieve this:

Step 1: Save Markers to the XML in tests/conftest.py
Add the pytest_runtest_setup hook to your tests/conftest.py file. Pytest automatically writes any key-value pairs added to item.user_properties directly into the <properties> section of the JUnit XML.

Add the following block to your tests/conftest.py:


def pytest_runtest_setup(item):
  """Hook to inject markers as properties into the test item."""
  for marker in item.iter_markers():
    item.user_properties.append(("marker", marker.name))
This records all applied markers (like integration_test, tpu_only, etc.) as <property> nodes inside <testcase>. For example, the output XML will contain:


<testcase classname="tests.integration.train_tests.TrainTests" name="test_tpu_base" time="240.5">
  <properties>
    <property name="marker" value="integration_test" />
    <property name="marker" value="tpu_only" />
  </properties>
</testcase>
Step 2: Update tests/utils/enforce_test_limits.py to Parse the Properties
Modify the logic in your limit-checking script to read these XML property tags.

Search and Replace in tests/utils/enforce_test_limits.py:
Find the following block:


            is_integration = (
                "integration" in xml_file.lower()
                or "tests/integration/" in full_name.lower()
                or ".integration" in full_name.lower()
            )
And replace it with:


            # Parse custom properties to extract markers
            markers = set()
            properties_elem = testcase.find("properties")
            if properties_elem is not None:
                for prop in properties_elem.findall("property"):
                    if prop.get("name") == "marker":
                        markers.add(prop.get("value"))

            # Determine test type (with directory/filename as fallback checks)
            is_integration = (
                "integration_test" in markers
                or "integration" in xml_file.lower()
                or "tests.integration." in full_name.lower()
            )
Why This is Better
Decoupled from Directory Structure: If you ever rename the tests/integration/ folder or relocate a file, the duration checker won't break.
Single Source of Truth: A test is recognized as an integration test purely because it is annotated with @pytest.mark.integration_test.
Extensible: You can easily use this same pattern for other types of tests or markers in the future without writing brittle file path checks.

shralex (commented):
Instead of relying on hardcoded limits, can we store per-test time during scheduled run, and detect regressions directly ? for example test X became 20% slower. 

Aggregate metrics are still useful to detect total time regressions.

tests/utils/enforce_test_limits.py
shralex: I'm not sure we even need this if we track individual test times.

tests/utils/parse_junit_to_benchmark.py
hsuan-lun-chiang: Thanks for the suggestion. I've updated the  `parse_junit_to_benchmark.py`  script. It now parses the  tests="N"  attribute from the XML reports, aggregates the total count for each device, and outputs an additional metric (e.g.,  `Total CPU Tests Count` ).

tests/utils/enforce_test_limits.py
hsuan-lun-chiang: Thanks for the idea. I have updated the code to use markers instead of parsing the file name.

hsuan-lun-chiang:
> Instead of relying on hardcoded limits, can we store per-test time during scheduled run, and detect regressions directly ? for example test X became 20% slower.
> 
> Aggregate metrics are still useful to detect total time regressions.

Thanks for sharing the idea. That was actually the initial approach. We originally tried tracking per-test regressions directly, but ran into two roadblocks that forced us to pivot to the current macro/hard-limit design:
1. Runner Noise: Many of the unit tests are extremely fast (e.g.,  0.01s ). Normal machine jitter can easily make a test take  0.1s . A naive 'detect 20% slower' rule immediately flags this as a 900% regression.
2. Dashboard Crashing: If we feed thousands of individual test functions into the  github-action-benchmark  action, the  data.js database grows exponentially and eventually crashes the browser when trying to render the interactive  gh-pages  dashboard. 

However, we will build a custom script that saves a separate per_test_baseline.json  to  `gh-pages`. We then compare the current run against that custom baseline and apply a compound threshold (e.g., Fail only if the test is >20% slower AND the absolute time increased by >5 seconds). This filters out the noise for fast tests while catching real regressions.

hsuan-lun-chiang:
> Instead of relying on hardcoded limits, can we store per-test time during scheduled run, and detect regressions directly ? for example test X became 20% slower.
> 
> Aggregate metrics are still useful to detect total time regressions.

Hi @shralex, we now store another `per_test_baseline.json` in gh-pages (which will be updated on every schedule run), and detect regressions with it. I also wanted to share some context around how the regression thresholds are configured in `tests/utils/enforce_test_limits.py`:

  #### 1. Why setting a global threshold is difficult

  • **Shared Runner Jitter**: GitHub Actions virtual machines run on shared multi-tenant hosts, the performance vary run-to-run.
    • **Fast-Test Relative Noise**: For very fast tests (e.g., executing under 5 seconds), minor fluctuations (such as a 4-second delay in network/file lookup) represent a massive 80% relative regression. If we relied solely on a relative threshold, we would suffer constant false-positive CI failures.

**The Solution**: To counteract this, we enforce a dual-gate condition. _A test is only flagged if it both exceeds a relative percentage regression AND exceeds a significant absolute noise-reduction threshold_ (e.g., +15s for unit tests, +120s for compile tests).

  #### 2. Why we use different thresholds: Train Compile (50%) vs. Other Tests (20%)

We observed that JAX/XLA graph compile times naturally have a high variance (often fluctuating by up to 30-40% under ordinary runner load, for example in this [run](https://github.com/AI-Hypercomputer/maxtext/actions/runs/28913579679/job/85778287824?pr=4290)).

Let me know if you have any questions or feedback on these approches!

.github/workflows/ci_pipeline.yml
```diff
@@ -367,3 +367,76 @@ jobs:
...
+        continue-on-error: true
+        run: |
+          git fetch origin gh-pages || true
+          git show origin/gh-pages:dev/bench/per_test_baseline.json > per_test_baseline.json || echo "{}" > per_test_baseline.json
+
+      - name: Enforce Individual Test Limits
+        run: |
+          python3 tests/utils/enforce_test_limits.py test-results \
+            --baseline per_test_baseline.json \
+            --save-baseline new_per_test_baseline.json
```
shralex: If this fails, the subsequent step Push per-test baseline to gh-pages is skipped. On main branch (scheduled runs), if a test regresses (either due to an intentional change that was merged, or environment changes), the job will fail and the baseline will not be updated. This creates a situation where the check will continue to fail until the baseline is manually updated on gh-pages.

Suggestion: Introduce a flag to enforce_test_limits.py to allow it to run in "non-blocking" mode (e.g., --no-fail or --warn-only), where it reports regressions but exits with code 0. Use this flag on main scheduled runs, similar to how fail-on-alert: false is used for github-action-benchmark.

tests/utils/enforce_test_limits.py
```diff
@@ -0,0 +1,132 @@
...
+  for xml_file in xml_files:
+    try:
+      tree = ET.parse(xml_file)
+      root = tree.getroot()
+      for testcase in root.iter("testcase"):
+        if process_testcase(testcase, xml_file, baseline_data, new_baseline_data):
+          failed = True
+
+    except Exception as e:  # pylint: disable=broad-exception-caught
+      print(f"Error parsing {xml_file}: {e}")
```
shralex: Here the script sets failed = True, but continues processing other files and eventually saves the baseline if --save-baseline is provided.

If we implement the suggestion allowing scheduled runs to pass even with regressions, we should ensure we do not save/push a baseline if there were actual errors (like parsing failures), as this would result in a partial baseline missing many tests.

Suggestion: Separate regression failures from execution/parse errors. Only save the baseline if there were no critical errors during the run.

  has_regression = False
  has_errors = False
  # ...
  for xml_file in xml_files:
    try:
      # ...
    except Exception as e:
      # ...
      has_errors = True
      
  if args.save_baseline and not has_errors:
    # Save baseline
  
  if has_errors:
    sys.exit(1) # Always fail on errors
  if has_regression and fail_on_regression:
    sys.exit(1)

tests/utils/enforce_test_limits.py
```diff
@@ -0,0 +1,132 @@
...
+  # Parse custom properties to extract markers
+  markers = set()
+  properties_elem = testcase.find("properties")
+  if properties_elem is not None:
+    for prop in properties_elem.findall("property"):
+      if prop.get("name") == "marker":
+        markers.add(prop.get("value"))
+
+  is_integration = (
+      "integration_test" in markers or "integration" in xml_file.lower() or "tests.integration." in full_name.lower()
```
shralex: I think the marker will always include integration for integration tests, we don't need to check the file name

tests/utils/parse_junit_to_benchmark.py
```diff
@@ -0,0 +1,64 @@
...
+    for testsuite in root.iter("testsuite"):
+      time_val = float(testsuite.get("time", 0.0))
+      test_count = int(testsuite.get("tests", 0))
+
+      # Recommend NOT adding individual file-level times to the benchmark action list
+      # to prevent rebalancing/sharding from triggering false regression alerts.
+      # Let Codecov handle the micro-level tracking.
+
+      total_times_by_device[device] = total_times_by_device.get(device, 0.0) + time_val
```
shralex: Currently, macro-level tracking aggregates results by device_type (CPU, GPU, TPU). However, different jobs might run different subsets of tests on the same device type (e.g., tpu-unit vs tpu-integration). Aggregating them together makes it hard to detect regressions in specific test suites and can cause noise if the test distribution changes.

Can we aggregate by job (e.g., tpu-unit, pathways integration, gpu-integration, etc.) instead of device type

tests/utils/parse_junit_to_benchmark.py
```diff
@@ -0,0 +1,64 @@
...
+  for xml_file in xml_files:
+    basename = os.path.basename(xml_file)
+    # e.g., test-results-tpu-1.xml -> device = tpu
+    device = "unknown"
+    parts = basename.replace(".xml", "").split("-")
+    if len(parts) >= 3:
+      device = parts[2]
+
+    try:
+      tree = ET.parse(xml_file)
```
shralex: The workflow currently parses all JUnit XML files twice: once in enforce_test_limits.py for micro-level enforcement, and once in parse_junit_to_benchmark.py for macro-level dashboard data generation.

Recommendation: Combine these two scripts into a single script (e.g., process_test_results.py) that performs both tasks in a single pass over the XML files. This improves efficiency and simplifies the workflow.

.github/workflows/ci_pipeline.yml
```diff
@@ -367,3 +367,76 @@ jobs:
     with:
       failed_run_id: '${{ github.run_id }}'
     secrets: inherit
+
+  track_performance:
```
shralex, line 420: ci_pipeline.yml is already very large. Adding the performance tracking logic directly to it makes it harder to maintain.

Recommendation: Move the track_performance job to a separate reusable workflow file, e.g., .github/workflows/track_performance.yml, and call it from ci_pipeline.yml. Reusable workflows inherit the execution context (including run_id for downloading artifacts) making this clean and modular.

tests/utils/parse_junit_to_benchmark.py
hsuan-lun-chiang: Done. Merged both into a unified `process_test_results.py` script.

.github/workflows/ci_pipeline.yml
hsuan-lun-chiang: Done. Extracted the tracking blocks into `.github/workflows/track_performance.yml`.

tests/utils/enforce_test_limits.py
hsuan-lun-chiang: Done. Replaced filename checks with direct marker properties extraction from the JUnit XML blocks.

tests/utils/parse_junit_to_benchmark.py
hsuan-lun-chiang: Done. The changes also reflect in https://ai-hypercomputer.github.io/maxtext/dev/bench/.

.github/workflows/ci_pipeline.yml
hsuan-lun-chiang: Done. Added a  --warn-only command-line argument to our new results processor and passed this flag in the  `.github/workflows/track_performance.yml` schedule workflow.

tests/utils/enforce_test_limits.py
hsuan-lun-chiang: Done. Now the script never saves or updates a baseline if `has_errors` is true, ensuring a corrupt/partial run never overwrites the baseline.

hsuan-lun-chiang:
Hi @shralex,
  I wanted to share an update from our workflow testing. We encountered a [Track Test Performance failure](https://github.com/AI-Hypercomputer/maxtext/actions/runs/28989148119/job/86027569717?pr=4290) that perfectly highlights the volatile runner noise we discussed:

  - test_moe_deepseek_with_device_limit  spiked from 27.89s to 183.99s (+156s / +559.7%) on the shared CPU runner.
  -  test_moe_deepseek_scanned_bf16  spiked from 112.11s to 236.90s (+124s / +111.3%).

  To proceed, we see two main directions. How would you prefer we approach this?

  #### Option 1: Skip regression tracking for `train_compile_test`

  #### Option 2: Stabilize compile test runtimes on runners

  If we want to keep compile tests in our performance tracking, we need to mitigate XLA's compilation volatility. Here are the main ways we can do this (Suggested by Gemini):

  #### A. Enable Persistent XLA Compilation Caching

  • Pros: Skips compilation entirely on cache hits, reducing subsequent test runs to sub-seconds.
  • Cons: Cache keys are highly sensitive. Any dependency updates (JAX/jaxlib) or config/shape changes will trigger cache misses, re-exposing the test to compilation spikes. Also introduces large binary storage overhead in GitHub cache.

  #### B. Enforce Runner Thread Limits (Thread-Thrashing Mitigation)

  • Pros: Restricts OMP_NUM_THREADS and MKL_NUM_THREADS to 2 to prevent XLA from over-subscribing threads and thrashing CPU caches on multi-tenant hosts.
  • Cons: Hardcoding low thread limits will throttle execution speeds on larger self-hosted runners or local developer workstations unless configured dynamically.

  #### C. Measure Pure Execution (Warm Up Step 0)

  • Pros: Performs a warmup step first to compile the graph, then times subsequent steps (1 +). This isolates the LLVM/XLA compilation overhead completely from our performance metrics.
  • Cons: Requires minor refactoring of the pytest timing wrappers to separate Step 0 from the main loop.
  ──────
  Would love to hear your thoughts on which direction fits best.

tests/utils/process_test_results.py
```diff
@@ -0,0 +1,235 @@
+"""Enforces test limits and generates macro-level benchmark metrics in a single pass."""
+
+import argparse
+import glob
+import json
+import os
+import sys
+import xml.etree.ElementTree as ET
+
+REL_REGRESSION_TRAIN_COMPILE_PCT = 1.50  # 50% slower
```
shralex: since this is called pct can we make it a number like 150, or call the variable frac or ratio, also below

tests/utils/process_test_results.py
```diff
@@ -0,0 +1,235 @@
...
+    test_type = "Train Compile Test"
+  else:
+    abs_noise_threshold = ABS_INCREASE_UNIT_SEC
+    rel_regression_pct = REL_REGRESSION_OTHER_PCT
+    test_type = "Unit Test"
+
+  new_baseline_data[full_name] = time_val
+
+  # Skip regression checking for CPU and train-compile tests due to shared CPU multi-tenancy noise
+  skip_regression = is_cpu or is_train_compile
```
shralex: Are all train compile tests cpu tests ? if so, can we simplify this condition to just check for is_cpu. Since the train compile thresholds and checks are effectively skipped can we remove them ?

shralex (approved):

tests/utils/parse_junit_to_benchmark.py
shralex: Can we add this link to the PR description so its easily accessible. Can we please add a doc (potentially as a follow up PR) similar to this one https://maxtext.readthedocs.io/en/maxtext-v0.2.3/development/hlo_diff_testing.html 
describing performance tracking, how to view results, the error message that the user will see, that this blocks PRs but not scheduled tests, etc. Also what the developer needs to do when a new test job is added to CI.

Also, can we add the v7x tests to tracking here please.

tests/utils/process_test_results.py
hsuan-lun-chiang: Done. Renamed `REL_REGRESSION_TRAIN_COMPILE_PCT` and `REL_REGRESSION_OTHER_PCT` to `REL_REGRESSION_TRAIN_COMPILE_RATIO` and `REL_REGRESSION_OTHER_RATIO`, and updated `rel_regression_pct` to `rel_regression_ratio`.

.github/workflows/run_tests_against_package.yml
```diff
@@ -207,10 +199,11 @@ jobs:
           if [ "${INPUTS_IS_UPDATE_HLO}" == "true" ]; then
             python3 tests/utils/update_hlo_references.py
           else
-            $PYTHON_EXE -m pytest ${INPUTS_PYTEST_ADDOPTS} \
+             $PYTHON_EXE -m pytest ${INPUTS_PYTEST_ADDOPTS} \
```
xibinliu: Remove the extra leading space.

tests/utils/process_test_results.py
hsuan-lun-chiang: Thanks for the suggestion! Indeed, all train compile tests run as CPU jobs, so their XML output filenames all contain `cpu`.

I have simplified the logic in `tests/utils/process_test_results.py`:
1.  Removed `is_train_compile` and simplified `skip_regression = is_cpu`.
2. Removed the unused `REL_REGRESSION_TRAIN_COMPILE_RATIO` and `ABS_INCREASE_TRAIN_COMPILE_SEC` threshold constants.

xibinliu (approved):

.github/workflows/run_tests_against_package.yml
hsuan-lun-chiang: Done! Removed the extra leading space in `.github/workflows/run_tests_against_package.yml`.

PR #4298, opened 2026-06-29
Fix load-balanced TPU Splash masks

huytransformer:
# Description

Fixes TPU Splash local sliding masks when CP load balancing is enabled. The causal mask already uses the load-balanced query order, but the local sliding mask used the default query order. This PR adds a load-balanced local mask so both masks use the same query positions.

BUGS: b/416110599

# Tests

Ran:

```bash
python3 -m pytest tests/unit/attention_test.py -k LoadBalancedMaskTest
```

# Checklist

### DIFF
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -1275,20 +1275,28 @@ class AttentionOp(nnx.Module):
     else:
       mask = mask_module.CausalMask(shape=mask_shape)
 
-    # Create LoadBalancedCausalMask if cp and load_balancing
-    if cp_size > 1 and load_balanced_context_parallel:
+    use_load_balanced_cp = cp_size > 1 and load_balanced_context_parallel
+    if use_load_balanced_cp and self.attention_type != AttentionType.FULL:
       mask = LoadBalancedCausalMask(shape=mask_shape, cp_size=cp_size)
 
-    # TODO: figure out local_sliding attention + load_balancing, default is global
     # Apply local masking if local sliding attention is enabled.
     if self.attention_type == AttentionType.LOCAL_SLIDING:
       if self.sliding_window_size is None:
         raise ValueError("Sliding_window_size must be set if Local Sliding attention type")
-      mask &= mask_module.LocalMask(
-          shape=(query.shape[2], key.shape[2]),
-          window_size=(self.sliding_window_size - 1, self.sliding_window_size),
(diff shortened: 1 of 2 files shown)


Shuwen-Fang (approved):

src/maxtext/layers/attention_op.py
```diff
@@ -2163,6 +2171,13 @@ def __call__(
       return prefill_unnormalized_output / prefill_exponentials_sum
 
 
+def _load_balanced_q_sequence(shape: tuple[int, int], cp_size: int):
+  """Reorders query positions the same way as load-balanced input tokens."""
+  arr = np.arange(shape[0])
+  out = max_utils.reorder_mask_load_balancing(arr[None, :, None, None], cp_size, 1)
+  return out[0, :, 0, 0]
```
NuojCheng: i feel sad reading line 2177 and 2178. reorder_mask_load_balancing only works for 4D tensor?

NuojCheng (approved):

src/maxtext/layers/attention_op.py
```diff
@@ -2223,3 +2232,23 @@ def __hash__(self):
...
+
+class LoadBalancedLocalMask(splash_attention_mask.LocalMask):
+  """Lazy local mask using load-balanced query order."""
+
+  def __init__(
+      self,
+      shape: tuple[int, int],
+      window_size: tuple[int | None, int | None],
+      offset: int,
+      shard_count: int = 1,
```
gobbleturk, line 2244: what is the difference betwen shard_count and cp_size? Can you include comments for each property?

src/maxtext/layers/attention_op.py
```diff
@@ -2223,3 +2232,23 @@ def __hash__(self):
...
+      cp_size: int = 4,
```
gobbleturk: I would prefer defaulting cp_size to -1 so users are forced to set it

gobbleturk (approved):

src/maxtext/layers/attention_op.py
huytransformer: noo don't be sad it was copy&paste from 2196-2200. fixed :)

src/maxtext/layers/attention_op.py
huytransformer: mm actually that's a good idea. done!!

PR #4307, opened 2026-06-30
Add use_te_comm_gemm_overlap: integrates TransformerEngine collective GEMMs for tensor sequence parallelism

dinodeep:
# Description

This PR integrates the necessary changes for enabling TransformerEngine's collective GEMM API for MLP layers as well as adding a configurable flag (`use_te_comm_gemm_overlap`, defaults to false) to enable it.

Collective GEMMs are useful because they allow for fine-grained overlap of communication and computation of distributed GEMM computations, reducing the distributed GEMM latency and improving overall E2E step time when working with tensor sequence parallelism.

Enabling the flag requires the usage of tensor sequence parallelism, running on single-rank-per-GPU, and enabling a Transformer Engine quantization recipe.

It works by telling TransformerEngine which type of collective will be performed for specific parts of the MLP. Up-projections require all-gathers, and down projections require reduce-scatters under tensor sequence parallelism. Overall benefit is reduced exposed communication time for large distributed GEMMs.

The change is primarily in `quantizations.py` in which we also determine the type of collective the dense layer is performing if it is in the MLP.

# Tests

Validated that enabling the collective GEMMs introduces them via profiling and looking at HLOs. Loss curves stay close when disabling/enabling collective GEMMs.

# Checklist

### DIFF
--- a/tests/integration/train_tests.py
+++ b/tests/integration/train_tests.py
@@ -296,6 +296,22 @@ class TrainTests(unittest.TestCase):
   @pytest.mark.skip(reason="No runner with GPU arch >= 89 is available")
   @pytest.mark.integration_test
   @pytest.mark.gpu_only
+  def test_gpu_te_fp8_delayedscaling_tsp_cgemm(self):
+    if jax.process_count() <= 1:
+      pytest.skip("Requires rank-per-GPU launch (JAX_PROCESS_COUNT > 1)")
+    if jax.local_device_count() != 1:
+      pytest.skip(f"Requires rank-per-GPU launch (local_device_count==1), " f"got {jax.local_device_count()}")
+    if not jax.distributed.is_initialized():
+      pytest.skip("Requires jax.distributed.initialize() (hardware=gpu_multiprocess)")
+
+    train_main(
+        TrainTests.CONFIGS["te_fp8_delayedscaling"]
+        + ["attention=dot_product", "ici_tensor_sequence_parallelism=2", "use_te_comm_gemm_overlap=true"]
+    )
+
+  @pytest.mark.skip(reason="No runner with GPU arch >= 89 is available")
+  @pytest.mark.integration_test
+  @pytest.mark.gpu_only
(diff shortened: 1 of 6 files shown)


src/maxtext/configs/pyconfig.py
```diff
@@ -550,6 +550,9 @@ def _initialize_pydantic(argv: list[str] | None = None, **kwargs) -> MaxTextConf
 
       compilation_cache.set_cache_dir(os.path.expanduser(pydantic_kwargs["jax_cache_dir"]))
 
+  if pydantic_kwargs["use_te_comm_gemm_overlap"]:
```
abhinavgoel95: Could you add an assertion to check if TPSP>1

gobbleturk (approved):

src/maxtext/configs/pyconfig.py
dinodeep: I added a check for that in the changes for `max_utils.py` in `bootstrap_transformer_engine_cgemm`. Would that be sufficient?

gobbleturk (approved):
Can we include a test with the collective matmul enabled? In train_tests.py I think

src/maxtext/configs/pyconfig.py
gobbleturk: we prefer these kind of assertions in types.py

tests/integration/train_tests.py
```diff
@@ -279,6 +293,22 @@ def test_gpu_nanoo_fp8(self):
   def test_gpu_te_fp8_delayedscaling(self):
     train_main(TrainTests.CONFIGS["te_fp8_delayedscaling"] + ["attention=dot_product"])
 
+  @pytest.mark.skip(reason="No runner with GPU arch >= 89 is available")
```
gobbleturk, line 296: this test is getting skipped for two different reasons lol, unfortunately this feature is not protected by our CI

gobbleturk (approved):

PR #4308, opened 2026-06-30
Fix hlo_diff_test to compile the real NNX path

ecnal-cienet:
# Description

`hlo_diff_test` loaded each model yml (e.g. `models/llama3-8b.yml`) directly as the top-level config. Those model ymls declare no `base_config` parent, so `base.yml` was never merged in. Two things broke as a result:

- `pure_nnx` fell back to its pydantic schema default (`False`) instead of `base.yml`'s `True`, so the test silently kept compiling the **Linen** path even after the NNX defaults were flipped. Its reference HLOs were therefore stale Linen graphs.
- `logical_axis_rules` lives only in `base.yml`, so it was left empty.

Compile via `base.yml` + `model_name` (the normal training path) instead, so the config inherits `base.yml`'s `logical_axis_rules` and exercises the real NNX path. The three reference HLOs (`deepseek3`, `llama3_8b`, `qwen3_1.7b`) are regenerated for NNX via the **Update HLO References** workflow.

# Tests

- Ran the **Update HLO References** workflow on this branch to regenerate `tests/utils/reference_hlo_*.txt` in the GitHub Actions TPU (v6e-4) environment.
- `tests/integration/hlo_diff_test.py` passes against the regenerated references in CI.

# Checklist

### DIFF
--- a/tests/integration/hlo_diff_test.py
+++ b/tests/integration/hlo_diff_test.py
@@ -93,11 +93,11 @@ class TestHloDiff:
     return False
 
   @pytest.mark.parametrize(
-      "test_id, config_file, overrides",
+      "test_id, model_name, overrides",
       [
           (
               "deepseek3",
-              "src/maxtext/configs/models/deepseek3-test.yml",
+              "deepseek3-test",
               {
                   "compile_topology": "v6e-4",
                   "base_num_decoder_layers": 4,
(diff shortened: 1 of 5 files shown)


tests/integration/hlo_diff_test.py
xibinliu: Since we rely on pyconfig.py to read the config file based on the model name, can we we just pass in the model name "deepseek3-test"? Same for the other models.

xibinliu (commented):
Can we also change the NNX flag default values to "True", so even without including the base.yaml, the tests can use the True flags.

tests/integration/hlo_diff_test.py
ecnal-cienet: Fixed

xibinliu (approved):

NuojCheng (approved):

PR #4313, opened 2026-07-01
Fix HF token validation to support Hugging Face CLI cache

Shuwen-Fang:
This PR fixes a bug where `train_rl.py` (and other post-training scripts) would fail with `ValueError: hf_access_token must be provided when not providing a pre-existing checkpoint` even if the user had authenticated via Hugging Face CLI (`hf auth login`).

### Cause
`model_creation_utils.py:from_pretrained` strictly validated `config.hf_access_token` (which only comes from `HF_TOKEN` environment variable) and did not check the Hugging Face CLI cache.

### Fix
Updated `from_pretrained` to fallback to `huggingface_hub.get_token()` to retrieve the token from the CLI cache if `config.hf_access_token` is not set. The retrieved token is also passed to the `to_maxtext` subprocess via the `HF_TOKEN` environment variable to ensure it can authenticate.

This makes the on-the-fly conversion during training work seamlessly after running `hf auth login` without needing to manually export `HF_TOKEN`.

Buganizer: https://b.corp.google.com/issues/528366193

# Checklist

### DIFF
--- a/src/maxtext/utils/model_creation_utils.py
+++ b/src/maxtext/utils/model_creation_utils.py
@@ -38,6 +38,7 @@ import subprocess
 import sys
 from typing import Callable, overload
 from etils import epath
+from huggingface_hub import get_token
 from flax import nnx
 from flax.core.meta import Partitioned
 import flax.linen as nn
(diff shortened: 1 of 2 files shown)


src/maxtext/utils/model_creation_utils.py
```diff
@@ -811,8 +811,14 @@ def from_pretrained(
   if config.convert_checkpoint_if_possible and not config.load_parameters_path:
     if not (epath.Path(config.base_output_directory) / "0" / "items").exists():
       # Try to convert checkpoint on the fly
-      if not config.hf_access_token:
-        raise ValueError("hf_access_token must be provided when not providing a pre-existing checkpoint")
+      from huggingface_hub import get_token  # pylint: disable=import-outside-toplevel
```
igorts-git: Is there a reason to have this import here? Often AI agents are just lazy.

igorts-git (approved):

src/maxtext/utils/model_creation_utils.py
Shuwen-Fang: good catch, updated

PR #4315, opened 2026-07-01
feat(ci): add ci fail-fast implementation plan

mesakhcienet:
## Description

This PR refactors the MaxText CI pipeline to orchestrate quality checks (linting/pre-commit) and documentation builds as the first phase of a Directed Acyclic Graph (DAG) before triggering heavy functional tests. 

Previously, `ci_pipeline.yml`, `code_quality.yml`, and `check_docs_build.yml` ran as entirely independent, concurrent workflows on Pull Requests. If a simple linting or documentation build error occurred, the pipeline would continue running the expensive, long-running TPU/GPU/CPU test suites for upwards of > 20 minutes. This resulted in wasted cloud compute resources and forced developers to wait unnecessarily for runs that were already bound to fail.

1. **`code_quality.yml`**: Added support for `workflow_call` and an optional `maxtext_sha` input, fallback checkout logic (`${{ inputs.maxtext_sha || github.sha }}`), and removed the separate concurrency block to prevent interference with the caller.
2. **`check_docs_build.yml`**: Relaxed the `maxtext_sha` input to be optional (`required: false`) so the workflow can still run independently when needed while gracefully falling back to `github.sha` if not provided.
3. **`ci_pipeline.yml`**: Added the `code_quality_check` and `docs_build_check` jobs at the top of the file, forced the heavy test matrixes (`tpu-tests`, `gpu-tests`, `cpu-tests`, `maxtext_tpu_pathways_unit_tests`, and `maxtext_tpu_pathways_integration_tests`) to require their success, and updated the `all_tests_passed` reporter job.

FIXES: [b/527592378](https://b.corp.google.com/issues/527592378)

---

## Tests

The changes have been thoroughly validated:

### 1. Local Syntax and Schema Validation
Validated all modified GitHub Actions files to ensure correct YAML structure:
```bash
python -c "import yaml; yaml.safe_load(open('.github/workflows/ci_pipeline.yml'))"
python -c "import yaml; yaml.safe_load(open('.github/workflows/code_quality.yml'))"
python -c "import yaml; yaml.safe_load(open('.github/workflows/check_docs_build.yml'))"
```
*Result:* All files parsed successfully with zero syntax errors.

### 2. Live GitHub Actions Run Verification
We successfully triggered a dry run of the refactored pipeline here: 
https://github.com/AI-Hypercomputer/maxtext/actions/runs/28509679093

* **Observations:** 
  * The fast phase jobs started concurrently at `10:05:16Z`.
  * `docs_build_check / build-docs` successfully passed in **21 seconds**.
  * `code_quality_check` failed in **46 seconds** due to pre-commit issues.
  * Immediately after the quality failure, **all downstream heavy test jobs were instantly skipped** (`tpu-tests`, `gpu-tests`, `cpu-tests`, and pathways tests), and running notebook jobs were automatically cancelled.
  * This confirmed a total fail-fast execution time of under **1 minute**, preventing any TPU/GPU compute from being scheduled or billed.

# Checklist

### DIFF
--- a/.github/workflows/check_docs_build.yml
+++ b/.github/workflows/check_docs_build.yml
@@ -17,12 +17,11 @@
 name: Check the Documentation Build
 
 on:
-  pull_request:
-    types: [opened, synchronize]
   workflow_call:
     inputs:
       maxtext_sha:
-        required: true
+        description: 'The specific MaxText commit SHA.'
+        required: false
         type: string
   workflow_dispatch:
 
@@ -38,7 +37,7 @@ jobs:
         with:
           persist-credentials: false
           fetch-depth: 0
-          ref: ${{ inputs.maxtext_sha }}
+          ref: ${{ inputs.maxtext_sha || github.sha }}
 
       - name: Check if only documentation changed
         id: check
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -126,12 +126,23 @@ jobs:
             echo "run_notebooks=false" >> $GITHUB_OUTPUT
           fi
 
+  code_quality_check:
+    uses: ./.github/workflows/code_quality.yml
+    with:
+      maxtext_sha: ${{ inputs.maxtext_sha || github.sha }}
+
+  docs_build_check:
+    uses: ./.github/workflows/check_docs_build.yml
+    with:
+      maxtext_sha: ${{ inputs.maxtext_sha || github.sha }}
+
   build_and_upload_maxtext_package:
-    needs: analyze_code_changes
-    # Run if either tests or notebooks need to run
+    needs: [analyze_code_changes, code_quality_check, docs_build_check]
+    # Run if either tests or notebooks need to run; on PRs, gate on code quality + docs passing
     if: |
-      needs.analyze_code_changes.outputs.run_tests == 'true' ||
-      needs.analyze_code_changes.outputs.run_notebooks == 'true'
+      always() &&
+      (needs.analyze_code_changes.outputs.run_tests == 'true' || needs.analyze_code_changes.outputs.run_notebooks == 'true') &&
+      (github.event_name != 'pull_request' || (needs.code_quality_check.result == 'success' && needs.docs_build_check.result == 'success'))
     uses: ./.github/workflows/build_package.yml
     with:
       device_type: tpu
@@ -154,10 +165,29 @@ jobs:
     secrets:
       HF_TOKEN: ${{ secrets.HF_TOKEN }}
 
+  gate_test_run:
+    name: Gate and Formalize Parameters
+    needs: [analyze_code_changes, build_and_upload_maxtext_package]
+    if: |
+      always() &&
+      needs.analyze_code_changes.outputs.run_tests == 'true' &&
+      needs.build_and_upload_maxtext_package.result == 'success'
+    runs-on: ubuntu-latest
+    outputs:
+      total_workers: ${{ steps.set-params.outputs.total_workers }}
+      worker_groups: ${{ steps.set-params.outputs.worker_groups }}
+    steps:
+      - id: set-params
+        name: Formalize Test Suite Parameters
+        run: |
+          TPU_UNIT_TOTAL_WORKERS=2
+          TPU_UNIT_WORKER_GROUPS='[1, 2]'
+          echo "total_workers=${TPU_UNIT_TOTAL_WORKERS}" >> "$GITHUB_OUTPUT"
+          echo "worker_groups=${TPU_UNIT_WORKER_GROUPS}" >> "$GITHUB_OUTPUT"
+
   tpu-tests:
     name: ${{ matrix.flavor || 'TPU' }} tests
-    needs: [build_and_upload_maxtext_package]
-    if: needs.analyze_code_changes.outputs.run_tests == 'true'
+    needs: [build_and_upload_maxtext_package, gate_test_run]
     uses: ./.github/workflows/run_tests_coordinator.yml
     strategy:
       fail-fast: false
@@ -171,8 +201,8 @@ jobs:
 
   tpu7x-tests:
     name: TPU7X tests
-    needs: [build_and_upload_maxtext_package]
-    if: needs.analyze_code_changes.outputs.run_tests == 'true' && github.ref == 'refs/heads/main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
+    needs: [build_and_upload_maxtext_package, gate_test_run]
+    if: github.ref == 'refs/heads/main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
     uses: ./.github/workflows/run_tests_coordinator.yml
     strategy:
       fail-fast: false
@@ -186,8 +216,7 @@ jobs:
 
   gpu-tests:
     name: ${{ matrix.flavor || 'GPU' }} tests
-    needs: [build_and_upload_maxtext_package]
-    if: needs.analyze_code_changes.outputs.run_tests == 'true'
+    needs: [build_and_upload_maxtext_package, gate_test_run]
     strategy:
       fail-fast: false
       matrix:
@@ -201,8 +230,7 @@ jobs:
 
   cpu-tests:
     name: ${{ matrix.flavor || 'CPU' }} tests
-    needs: [build_and_upload_maxtext_package]
-    if: needs.analyze_code_changes.outputs.run_tests == 'true'
+    needs: [build_and_upload_maxtext_package, gate_test_run]
     uses: ./.github/workflows/run_tests_coordinator.yml
     strategy:
       fail-fast: false
@@ -214,31 +242,13 @@ jobs:
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
       maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
 
-  setup-pathways-parameters:
-    name: Setup Pathways Parameters
-    runs-on: ubuntu-latest
-    outputs:
-      total_workers: ${{ steps.set-params.outputs.total_workers }}
-      worker_groups: ${{ steps.set-params.outputs.worker_groups }}
-    steps:
-      - id: set-params
-        name: Set Pathways Worker Parameters
-        run: |
-          # TPU worker constants
-          TPU_UNIT_TOTAL_WORKERS=2
-          TPU_UNIT_WORKER_GROUPS='[1, 2]'
-
-          echo "total_workers=${TPU_UNIT_TOTAL_WORKERS}" >> "$GITHUB_OUTPUT"
-          echo "worker_groups=${TPU_UNIT_WORKER_GROUPS}" >> "$GITHUB_OUTPUT"
-
   maxtext_tpu_pathways_unit_tests:
-    needs: [build_and_upload_maxtext_package, setup-pathways-parameters]
-    if: needs.analyze_code_changes.outputs.run_tests == 'true'
+    needs: [build_and_upload_maxtext_package, gate_test_run]
     uses: ./.github/workflows/run_pathways_tests.yml
     strategy:
         fail-fast: false
         matrix:
-          group: ${{ fromJSON(needs.setup-pathways-parameters.outputs.worker_groups) }}
+          group: ${{ fromJSON(needs.gate_test_run.outputs.worker_groups || '[1, 2]') }}
     with:
       device_type: tpu
       device_name: v6e-4
@@ -251,12 +261,11 @@ jobs:
       container_resource_option: "--privileged"
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
       maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
-      total_workers: ${{ needs.setup-pathways-parameters.outputs.total_workers }}
+      total_workers: ${{ needs.gate_test_run.outputs.total_workers || '2' }}
       worker_group: ${{ matrix.group }}
 
   maxtext_tpu_pathways_integration_tests:
-    needs: build_and_upload_maxtext_package
-    if: needs.analyze_code_changes.outputs.run_tests == 'true'
+    needs: [build_and_upload_maxtext_package, gate_test_run]
     uses: ./.github/workflows/run_pathways_tests.yml
     strategy:
         fail-fast: false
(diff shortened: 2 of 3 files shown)


.github/workflows/ci_pipeline.yml
```diff
@@ -126,6 +126,18 @@ jobs:
             echo "run_notebooks=false" >> $GITHUB_OUTPUT
           fi
 
+  code_quality_check:
+    needs: analyze_code_changes
```
xibinliu: It doesn't depend on analyze_code_changes - dependency can be removed. Same for docs_build_check.

.github/workflows/ci_pipeline.yml
shralex: Not modified in the PR, but perhaps you can still fix it. There is a typo in the variable here, it should be  NEEDS_ANALYZE_CODE_CHANGES_OUTPUTS_RUN_TESTS  lets remove this if since the code below already checks that tests passed or skipped but not failed.

.github/workflows/code_quality.yml
shralex: Previously, code_quality.yml was configured to run only on pull request events. In this PR, code_quality_check is being integrated as a job directly inside the main ci_pipeline.yml workflow. But ci_pipeline.yml runs on both pull requests and scheduled runs. The pull request-specific environment variables $GITHUB_BASE_REF and $GITHUB_HEAD_REF are empty on scheduled runs. The commands to fetch and branch them in code_quality.yml will fail and cause the scheduled build to fail.

So there are 2 issues here:
1. On scheduled (or manual) runs on main, we want to collect as much test coverage and signal as possible. Having a minor code quality or documentation warning skip the entire TPU and GPU test suite is counterproductive for nightly or scheduled builds.

Suggestion: Add the following if: condition to all test suites that depend on code_quality_check and docs_build_check (such as tpu-tests, gpu-tests, cpu-tests, and pathways tests):


    if: |
      always() &&
      needs.analyze_code_changes.outputs.run_tests == 'true' &&
      needs.build_and_upload_maxtext_package.result == 'success' &&
      (github.event_name != 'pull_request' || (needs.code_quality_check.result == 'success' && needs.docs_build_check.result == 'success'))


2. Update .github/workflows/code_quality.yml to check the event name:

    - name: Run pre-commit checks on just the files that have changed
      run: |
        . "$GITHUB_WORKSPACE"/venv/bin/activate
        if [ "${{ github.event_name }}" == "pull_request" ]; then
          git fetch origin "$GITHUB_BASE_REF":"$GITHUB_BASE_REF"
          git branch "$GITHUB_HEAD_REF"
          pre-commit run --from-ref "$GITHUB_BASE_REF" --to-ref "$GITHUB_HEAD_REF" --show-diff-on-failure
        else
          echo "Not a pull request event (${{ github.event_name }}), checking all files."
          pre-commit run --all-files --show-diff-on-failure
        fi

.github/workflows/ci_pipeline.yml
shralex: we need to add code_quality_check and docs_build_check here

shralex: we need to add code_quality_check and docs_build_check here.

shralex (changes requested):
Thanks for the changes!

Please rebase your PR as tpu7x tests were added, and its "needs" should also be updated in your PR

.github/workflows/ci_pipeline.yml
mesakhcienet: Done — removed the early-exit block entirely rather than just fixing the variable name. The contains(needs.*.result, 'failure') check below already handles passed / skipped / failed correctly.

Worth noting the block was actually dead code: it read NEEDS_ANALYZE_CHANGES_OUTPUTS_RUN_TESTS while the env declared NEEDS_ANALYZE_CODE_CHANGES_OUTPUTS_RUN_TESTS (the typo you spotted), so it never matched — and had it matched, it would have exit 0'd on docs-only runs and masked a code_quality_check / docs_build_check failure. Removing it fixes that too.

.github/workflows/code_quality.yml
mesakhcienet: Done, both points addressed:

1. **Gating scoped to PRs.** The test suites (`tpu-tests`, `tpu7x-tests`, `gpu-tests`, `cpu-tests`, and both pathways jobs) now use the condition you suggested:
   ```yaml
   if: |
     always() &&
     needs.analyze_code_changes.outputs.run_tests == 'true' &&
     needs.build_and_upload_maxtext_package.result == 'success' &&
     (github.event_name != 'pull_request' || (needs.code_quality_check.result == 'success' && needs.docs_build_check.result == 'success'))
   ```
   So on `schedule` / `workflow_dispatch` the suites run regardless of the code-quality/docs outcome (full nightly signal), and on PRs they still gate.

2. **Event guard in `code_quality.yml`.** The pre-commit step now branches on `github.event_name`: on `pull_request` it runs the diff-scoped `--from-ref/--to-ref` check; otherwise it runs `pre-commit run --all-files`. (`check_docs_build.yml` already had the equivalent non-PR guard, so no change needed there.)

.github/workflows/ci_pipeline.yml
mesakhcienet: Done

mesakhcienet: Done

mesakhcienet:
Thanks @shralex and @xibinliu for the careful review. Rebased onto latest `main` (picking up the new `tpu7x` tests) and pushed fixes for every thread. Summary:

**Addressed**
- **Unused dep (@xibinliu):** dropped `needs: analyze_code_changes` from `code_quality_check` + `docs_build_check`.
- **`all_tests_passed` early-exit (@shralex):** removed it — it was dead code (variable-name mismatch) and would have masked quality/docs failures on docs-only runs. `contains(needs.*.result, 'failure')` already covers every case.
- **Scheduled-run breakage (@shralex):**
  - Test suites now use `always() && … && (github.event_name != 'pull_request' || (code_quality_check + docs_build_check success))` — nightly/dispatch runs get full test signal even when code-quality/docs warn; PRs still gate.
  - `code_quality.yml` guards the PR-only `git fetch`/`branch` steps by `github.event_name` and runs `pre-commit --all-files` on non-PR events.
- **Failure notifications (@shralex):** added `code_quality_check` + `docs_build_check` to `notify_failure` and `investigate_failure` `needs`, so scheduled quality/docs failures create the tracking issue and trigger investigation.
- **Rebase:** onto latest `main`; `tpu7x-tests` `needs` updated to the fail-fast pattern, and it's wired into `all_tests_passed` / `notify_failure` / `investigate_failure`.
- **Build-failure masking (follow-up spotted during review):** added `build_and_upload_maxtext_package` to `all_tests_passed`'s `needs`. A real build failure skips every test job, and skipped ≠ failure, so the aggregate previously reported **green** on a broken build. It now fails correctly — and as a bonus the release gate blocks properly on an unbuildable package.

.github/workflows/check_docs_build.yml
shralex: should we remove the pull_request trigger since it now runs from ci_pipeline ?

.github/workflows/code_quality.yml
```diff
@@ -18,23 +18,22 @@ name: Code Quality
 
 on:
   pull_request:
```
shralex: should we remove the pull_request trigger since it now runs from ci_pipeline ?

shralex (approved):

.github/workflows/ci_pipeline.yml
xibinliu: Running this bare runner should be gated, too.

I would suggest to add a `gate_test_run` job, and also consolidate this `setup-pathways-parameters` task into the same job.

With this approach, we don't need add the `if` block to all tests, and just need them to depend on the `gate_test_run`.

```
  gate_test_run:
    name: Gate and Formalize Parameters
    needs: [analyze_code_changes, build_and_upload_maxtext_package, code_quality_check, docs_build_check]
    if: |
      always() &&
      needs.analyze_code_changes.outputs.run_tests == 'true' &&
      needs.build_and_upload_maxtext_package.result == 'success' &&
      (github.event_name != 'pull_request' || (needs.code_quality_check.result == 'success' && needs.docs_build_check.result == 'success'))
    runs-on: ubuntu-latest
    outputs:
      # Formalize pathways parameters directly from the gatekeeper job
      pathways_total_workers: ${{ steps.set-params.outputs.total_workers }}
      pathways_worker_groups: ${{ steps.set-params.outputs.worker_groups }}
    steps:
      - id: set-params
        name: Formalize Test Suite Parameters
        run: |
          echo "Pre-conditions met. Formalizing downstream parameters."
          
          # Consolidate Pathways test parameters
          TPU_UNIT_TOTAL_WORKERS=2
          TPU_UNIT_WORKER_GROUPS='[1, 2]'
          
          echo "total_workers=${TPU_UNIT_TOTAL_WORKERS}" >> "$GITHUB_OUTPUT"
          echo "worker_groups=${TPU_UNIT_WORKER_GROUPS}" >> "$GITHUB_OUTPUT"

```

.github/workflows/check_docs_build.yml
mesakhcienet: Done. Removed `pull_request:` from `check_docs_build.yml` (kept `workflow_call` and `workflow_dispatch`). 

It still runs on PRs, because `ci_pipeline.yml` runs on PRs and calls it as the `docs_build_check` job. So nothing is lost — this just stops it running twice.

Heads-up: the PR check name changes from `Check the Documentation Build / build-docs` to `MaxText Package Tests / docs_build_check / build-docs`. If branch protection requires the old name, someone with admin will need to update it — I can't see those settings from the fork.

.github/workflows/code_quality.yml
mesakhcienet: Done, same as above. Removed `pull_request:` from `code_quality.yml` (kept `workflow_call`). It still runs on PRs through `ci_pipeline`'s `code_quality_check` job.

Same heads-up: the check name changes from `Code Quality / qa` to `MaxText Package Tests / code_quality_check / qa`, so it may need updating in the required checks list.

.github/workflows/ci_pipeline.yml
mesakhcienet: Done.

Added a `gate_test_run` job that holds the `always() && …` gate, moved `setup-pathways-parameters` into it (so that ungated runner is gated now), and the test suites just use `needs: [build_and_upload_maxtext_package, gate_test_run]` — their `if:` blocks are gone. That's −28 lines in `ci_pipeline.yml`.

A few notes:

- `tpu7x-tests` still keeps a small `if: github.ref == 'refs/heads/main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')`. That rule is specific to tpu7x, so it stays on the job instead of going into the shared gate.
- The gate can now skip (docs-only PRs, build failures, etc.), and a skipped gate has empty outputs. So pathways-unit uses `fromJSON(needs.gate_test_run.outputs.worker_groups || '[1, 2]')` and `... || '2'`, so the matrix never gets an empty value. The fallback only matters if the matrix is built on a skip path — the job still runs or skips based on `needs: [gate_test_run]`.
- Kept the output names `total_workers` / `worker_groups` (same as the old job), so `run_pathways_tests.yml` needs no changes.
- Small bonus: `gate_test_run` lists `analyze_code_changes` as a direct `need`, so `run_tests` reads cleanly instead of relying on it being pulled in indirectly.
- `gate_test_run` is also in `all_tests_passed`'s `needs`, so if the gate ever fails, the overall check still goes red.

.github/workflows/ci_pipeline.yml
```diff
@@ -214,31 +221,34 @@ jobs:
       is_scheduled_run: ${{ github.event_name == 'schedule' }}
       maxtext_sha: ${{ needs.build_and_upload_maxtext_package.outputs.maxtext_sha }}
 
-  setup-pathways-parameters:
-    name: Setup Pathways Parameters
+  gate_test_run:
```
xibinliu: Can you move this job before all the test jobs?

.github/workflows/ci_pipeline.yml
xibinliu, line 154: this test should also be gated by gate_test_run.

.github/workflows/ci_pipeline.yml
```diff
@@ -126,6 +126,16 @@ jobs:
...
+    uses: ./.github/workflows/code_quality.yml
+    with:
+      maxtext_sha: ${{ inputs.maxtext_sha || github.sha }}
+
+  docs_build_check:
+    uses: ./.github/workflows/check_docs_build.yml
+    with:
+      maxtext_sha: ${{ inputs.maxtext_sha || github.sha }}
+
   build_and_upload_maxtext_package:
```
xibinliu, line 139: We should make this job depends on code_quality_check and docs_build_check. There is no point building images if quality check fails.

xibinliu (approved):

.github/workflows/ci_pipeline.yml
mesakhcienet: Done — moved `gate_test_run` above the test jobs. Job order doesn't change how they run (GitHub goes by `needs`, not file order), but it reads better with the gatekeeper defined first.

.github/workflows/ci_pipeline.yml
mesakhcienet: Quick check on this one before I wire it up. `gate_test_run` is gated on `run_tests`, but the notebooks job runs on `run_notebooks`. On a notebook-only PR (`run_tests=false`), `gate_test_run` skips — so if the notebooks job depended on it, notebooks would never run.

I think what you want is for notebooks to get the same quality gating. If we make `build_and_upload_maxtext_package` depend on the quality checks (your other comment), the notebooks job already depends on `build`, so it'd be quality-gated automatically — no need to wire it to `gate_test_run` directly.

Does that cover it, or did you want notebooks gated separately from build?

.github/workflows/ci_pipeline.yml
mesakhcienet: Agreed — no point building if quality fails. One thing to keep: on scheduled/nightly runs we still want everything to run even if there's a lint failure (per @shralex's earlier note), so I'd gate this on PRs only:

```yaml
build_and_upload_maxtext_package:
  needs: [analyze_code_changes, code_quality_check, docs_build_check]
  if: |
    always() &&
    (needs.analyze_code_changes.outputs.run_tests == 'true' || needs.analyze_code_changes.outputs.run_notebooks == 'true') &&
    (github.event_name != 'pull_request' || (needs.code_quality_check.result == 'success' && needs.docs_build_check.result == 'success'))
```

That skips the build on a PR when quality fails (so tests and notebooks skip too — fail-fast), but keeps nightly running for full signal. And once the build is quality-gated, `gate_test_run`'s own quality check becomes redundant, so I can simplify it at the same time. Is it sounds good ?

.github/workflows/ci_pipeline.yml
xibinliu: Checking analyze_code_changes, looks like `run_tests` is always `true` if `run_notebooks` is true. So this shouldn't be a problem for notebook-only PR.

Edit - just found we do have run_notebooks only case, where run_tests is false

.github/workflows/ci_pipeline.yml
xibinliu: This looks good!

.github/workflows/ci_pipeline.yml
```diff
@@ -154,10 +164,30 @@ jobs:
     secrets:
       HF_TOKEN: ${{ secrets.HF_TOKEN }}
 
+  gate_test_run:
```
xibinliu, line 168: If we've gated build_and_upload_maxtext_package, which depends on analyze_code_changes, code_quality_check, and docs_build_check, we don't need this gating step anymore.

mesakhcienet: Good point — you're right that the quality gate becomes redundant in `gate_test_run` once `build_and_upload_maxtext_package` depends on the quality checks. I'd simplify it rather than remove it, though, because it still does two things the build doesn't:

1. **The `run_tests` gate.** The build runs on `run_tests || run_notebooks`, so on a notebook-only PR (`run_tests=false`, `run_notebooks=true`) the build runs. If the test suites depended only on `build`, they'd run too — but they shouldn't, since there are no code changes. `gate_test_run`'s `run_tests == 'true'` check is what keeps them skipped. If we drop the whole job, every suite needs its own `if: run_tests == 'true'` back — the per-suite duplication we removed earlier.
2. **The pathways params** (`total_workers` / `worker_groups`) that `maxtext_tpu_pathways_unit_tests` reads for its matrix.

So I'd keep `gate_test_run` but strip the now-redundant quality clause:

```yaml
gate_test_run:
  needs: [analyze_code_changes, build_and_upload_maxtext_package]
  if: |
    always() &&
    needs.analyze_code_changes.outputs.run_tests == 'true' &&
    needs.build_and_upload_maxtext_package.result == 'success'
  outputs:
    total_workers: ...
    worker_groups: ...
```

That keeps the suites gated on `run_tests` and depending on one job (no per-suite `if:` blocks), while the quality gate lives at the build. Is it sounds good?

PR #4339, opened 2026-07-02
fix(ci): Execute newly added or modified `scheduled_only` tests in pre-submits

mesakhcienet:
# Description

This change makes the pre-submit execute the `scheduled_only` tests that a PR **introduces or modifies**, so each is verified at least once before merge, while unrelated `scheduled_only` tests remain nightly-only (unchanged).

### How it works

- A pytest collection hook in `tests/conftest.py` computes the set of tests a PR adds or modifies — via `git diff --unified=0 origin/<base>...HEAD` — and dynamically applies a new `newly_added` marker to the matching collected tests.
- The two test-runner workflows change the pre-submit marker from `<marker> and not scheduled_only` to `<marker> and (not scheduled_only or newly_added)`. Scheduled (nightly) runs are untouched and continue to run everything.

### Detection details / robustness

- Diff parsing lives in a small, **stdlib-only, unit-tested** module `tests/utils/newly_added_detection.py` (`parse_changed_tests` is pure; `get_changed_tests` wraps git). Keeping it out of `conftest.py` lets it be tested without importing absl/pytest/jax.
- Matching is **file-qualified** — `(file_path, test_name)` — so identically named tests in different files (e.g. `test_basic_call` appears in 6 files) do not cross-contaminate and pull unrelated `scheduled_only` tests into pre-submit.
- `.gitattributes` sets `*.py diff=python` so `git diff` hunk headers expose the enclosing `def test_*` for edits **inside class methods** (the suite is overwhelmingly class-based). This enables detection of *modified* tests, not just brand-new ones.
- The package-test checkout uses `fetch-depth: 0` so the three-dot `origin/main...HEAD` diff has a reachable merge-base (otherwise the shallow clone forces a less precise fallback).
- The base branch falls back to `GITHUB_BASE_REF` (then `main`), so PRs targeting a non-`main` branch are handled.

### Files changed

| File | Change |
|------|--------|
| `tests/conftest.py` | Collection hook applies `newly_added` (file-qualified match) |
| `tests/utils/newly_added_detection.py` *(new)* | Pure diff parser + git wrapper |
| `tests/unit/newly_added_detection_test.py` *(new)* | Unit tests for the parser |
| `.github/workflows/run_tests_against_package.yml` | Marker expression + `fetch-depth: 0` |
| `.github/workflows/run_pathways_tests.yml` | Marker expression |
| `.gitattributes` *(new)* | `*.py diff=python` |
| `pytest.ini` | Register `newly_added` marker |

### Shortcomings / future work

- Detection is a no-op when `maxtext_installed=true` (the runner does `cd /deps`, which is not a git work tree). Not on the current PR pre-submit path, but worth noting.
- `run_pathways_tests.yml` checks out via a hardcoded `git clone https://github.com/google/maxtext.git`, which does not resolve fork build SHAs (pre-existing, out of scope here).
- A change that only *deletes* lines from a test (no added lines) is not flagged as modified.

# Tests

New unit tests in `tests/unit/newly_added_detection_test.py` (marked `cpu_only`) cover the diff parser across: new top-level function, new class method, modified method detected via the hunk header, cross-file name-collision isolation, non-test / `src/` changes ignored, and multi-file diffs.

All commands run locally against the repo `.venv` (Python 3.12).

**1. Unit tests**

```console
$ python3 -m pytest tests/unit/newly_added_detection_test.py -q
6 passed, 1 warning in 0.14s
```

**2. Lint gate** (same hooks as `code_quality.yml`)

```console
$ pre-commit run --files tests/conftest.py tests/utils/newly_added_detection.py \
    tests/unit/newly_added_detection_test.py .gitattributes \
    .github/workflows/run_tests_against_package.yml pytest.ini
codespell....................Passed
pylint.......................Passed
pyink........................Passed
yamllint.....................Passed
```

**3. End-to-end through the real `conftest.py`** — a temporary `scheduled_only` probe test committed on the branch, collected through the actual `tests/conftest.py`:

```console
# The new scheduled_only test IS rescued by the newly_added marker:
$ python3 -m pytest -m "cpu_only and (not scheduled_only or newly_added)" tests/unit/_probe_test.py -q
1 passed

# Baseline: under the old marker it would have been skipped:
$ python3 -m pytest -m "cpu_only and not scheduled_only" tests/unit/_probe_test.py -q
1 deselected
```

This confirms the full chain: `git diff` → file-qualified parse → `newly_added` marker → survives `-m` selection → the `scheduled_only` test runs. The `.gitattributes` change was separately verified to make `git diff --unified=0` emit `@@ ... @@ def test_column(...)` headers for edits inside class methods (instead of the enclosing `class ...` line).

**4. CI:** the two modified workflows exercise this change on the actual TPU / GPU / CPU and Pathways pre-submit runners in this PR.

# Checklist

- [x] I have run end-to-end tests tests and provided workload links above if applicable. — N/A: CI-infrastructure change with no ML workload; verified via unit tests, the lint gate, an end-to-end conftest probe (above), and the modified workflows running in this PR's pre-submit.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files). — No user-facing docs affected.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -115,6 +115,7 @@ if os.getenv("JAX_PLATFORMS") == "proxy":
   import maxtext  # pylint: disable=unused-import
 
 from maxtext.common.gcloud_stub import is_decoupled
+from tests.utils.newly_added_detection import get_changed_tests
 
 # Configure JAX to use unsafe_rbg PRNG implementation to match main scripts.
 if is_decoupled():
@@ -149,6 +150,15 @@ def pytest_collection_modifyitems(config, items):
   - Deselect tests marked as external_serving/training in decoupled mode.
   - Mark remaining tests with the `decoupled` marker when running decoupled.
   """
+
+  changed_tests = get_changed_tests()  # set[(file_path, test_name)]
+  if changed_tests:
+    for item in items:
+      item_file = item.nodeid.split("::", 1)[0]
+      base_name = getattr(item, "originalname", item.name)
+      if (item_file, base_name) in changed_tests or (item_file, item.name) in changed_tests:
+        item.add_marker(pytest.mark.newly_added)
+
   decoupled = is_decoupled()
   remaining = []
   deselected = []
@@ -199,6 +209,7 @@ def pytest_configure(config):
       "external_training: goodput integrations",
       "decoupled: marked on tests that are not skipped due to GCP deps, when DECOUPLE_GCLOUD=TRUE",
       "skip_on_tpu7x: skip test if running on TPU7x platform",
+      "newly_added: newly introduced or modified tests in PRs, executed even if scheduled_only",
   ]:
     config.addinivalue_line("markers", m)
 
--- a/tests/unit/newly_added_detection_test.py
+++ b/tests/unit/newly_added_detection_test.py
@@ -1,0 +1,194 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for the line-range test-change detector in ``newly_added_detection``."""
+
+import textwrap
+
+import pytest
+
+from tests.utils.newly_added_detection import _build_diff_commands
+from tests.utils.newly_added_detection import _is_test_file
+from tests.utils.newly_added_detection import parse_changed_line_map
+from tests.utils.newly_added_detection import touched_test_names
+
+pytestmark = pytest.mark.cpu_only
+
+
+# --- _is_test_file -----------------------------------------------------------
+
+
+def test_is_test_file_accepts_test_suffixes_under_tests():
+  assert _is_test_file("tests/unit/a_test.py")
+  assert _is_test_file("tests/integration/b_tests.py")
+
+
+def test_is_test_file_rejects_non_tests_and_non_suffix():
+  assert not _is_test_file("tests/utils/test_helpers.py")  # helper, not *_test.py
+  assert not _is_test_file("src/maxtext/foo_test.py")  # outside tests/
+  assert not _is_test_file("tests/unit/helper.py")  # not a test suffix
+
+
+# --- _build_diff_commands ----------------------------------------------------
+
+
+def test_build_diff_commands_uses_only_merge_base_threedot_ranges():
+  # Only three-dot (merge-base) ranges: remote first (CI / local-with-origin),
+  # then local (developer without an origin remote). Two-dot tip-vs-tip ranges are
+  # excluded because they over-report every commit the base gained past the fork
+  # point, dragging unrelated scheduled_only tests into pre-submit.
+  cmds = _build_diff_commands("main")
+  assert cmds == [
+      ["git", "diff", "--unified=0", "origin/main...HEAD"],
+      ["git", "diff", "--unified=0", "main...HEAD"],
+  ]
+  # Guard against a two-dot form sneaking back in: every range arg is three-dot.
+  range_args = [cmd[-1] for cmd in cmds]
+  assert all("..." in arg for arg in range_args)
+
+
+def test_build_diff_commands_honours_non_main_base():
+  cmds = _build_diff_commands("release/v2")
+  assert cmds == [
+      ["git", "diff", "--unified=0", "origin/release/v2...HEAD"],
+      ["git", "diff", "--unified=0", "release/v2...HEAD"],
+  ]
+
+
+# --- parse_changed_line_map --------------------------------------------------
+
+
+def test_line_map_added_region_from_header_range():
+  diff = (
+      "diff --git a/tests/unit/a_test.py b/tests/unit/a_test.py\n"
+      "--- a/tests/unit/a_test.py\n"
+      "+++ b/tests/unit/a_test.py\n"
+      "@@ -0,0 +1,3 @@\n"
+      "+line1\n+line2\n+line3\n"
+  )
+  assert parse_changed_line_map(diff) == {"tests/unit/a_test.py": {1, 2, 3}}
+
+
+def test_line_map_single_line_default_length():
+  diff = "+++ b/tests/unit/a_test.py\n" "@@ -10 +12 @@\n" "+changed\n"
+  assert parse_changed_line_map(diff) == {"tests/unit/a_test.py": {12}}
+
+
+def test_line_map_pure_deletion_marks_boundary():
+  # `+4,0` = pure deletion; the join sits between new-file lines 4 and 5.
+  diff = "+++ b/tests/unit/a_test.py\n" "@@ -5,2 +4,0 @@\n" "-gone1\n-gone2\n"
+  assert parse_changed_line_map(diff) == {"tests/unit/a_test.py": {4, 5}}
+
+
+def test_line_map_accumulates_multiple_hunks_and_files():
+  diff = (
+      "+++ b/tests/unit/a_test.py\n"
+      "@@ -0,0 +1,1 @@\n"
+      "+x\n"
+      "@@ -8,0 +10,2 @@\n"
+      "+y\n+z\n"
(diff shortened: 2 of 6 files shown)


tests/utils/newly_added_detection.py
```diff
@@ -0,0 +1,192 @@
+# Copyright 2025 Google LLC
```
xibinliu: 2026

tests/unit/newly_added_detection_test.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2025 Google LLC
```
xibinliu: 2026

tests/utils/newly_added_detection.py
```diff
@@ -0,0 +1,192 @@
...
+  except SyntaxError:
+    return set()
+  found = set()
+  for name, start, end in _iter_test_defs(tree):
+    if any(start <= line <= end for line in touched_lines):
+      found.add(name)
+  return found
+
+
+def get_changed_tests(base_ref=None):
```
xibinliu, line 156: Do we want to detect both newly added tests and the modified tests, or just new tests? @shralex

shralex: detecting modified tests sounds great

tests/conftest.py
```diff
@@ -149,6 +150,15 @@ def pytest_collection_modifyitems(config, items):
   - Deselect tests marked as external_serving/training in decoupled mode.
   - Mark remaining tests with the `decoupled` marker when running decoupled.
   """
+
+  changed_tests = get_changed_tests()  # set[(file_path, test_name)]
+  if changed_tests:
+    for item in items:
+      item_file = item.nodeid.split("::", 1)[0]
```
shralex, line 157: item_file = os.path.abspath(item.nodeid.split("::", 1)[0])

tests/conftest.py
```diff
@@ -199,6 +209,7 @@ def pytest_configure(config):
       "external_training: goodput integrations",
       "decoupled: marked on tests that are not skipped due to GCP deps, when DECOUPLE_GCLOUD=TRUE",
       "skip_on_tpu7x: skip test if running on TPU7x platform",
+      "newly_added: newly introduced tests in PRs, executed even if scheduled_only",
```
shralex: add "or modified"

tests/utils/newly_added_detection.py
```diff
@@ -0,0 +1,192 @@
...
+    if inside.returncode != 0:
+      return set()
+    if os.environ.get("GITHUB_ACTIONS") == "true":
+      subprocess.run(
+          ["git", "fetch", "origin", f"{base}:refs/remotes/origin/{base}"],
+          stdout=subprocess.DEVNULL,
+          stderr=subprocess.DEVNULL,
+          check=False,
+      )
+    try:
```
shralex: gemini proposes the following (more fallback strategies):

    # Diff strategies to try in order of preference.
    # Remote specs are tried first (for CI), falling back to local specs (for local development).
    diff_strategies = [
        ["origin/{base}...HEAD"],
        ["origin/{base}", "HEAD"],
        ["{base}...HEAD"],
        ["{base}", "HEAD"],
    ]

    diff_text = None
    for strategy in diff_strategies:
        try:
            diff_text = subprocess.check_output(
                ["git", "diff", "--unified=0"] + [arg.format(base=base) for arg in strategy],
                text=True,
                stderr=subprocess.DEVNULL,
            )
            break
        except Exception: # pylint: disable=broad-exception-caught
            continue

    if diff_text is None:
        return set()

tests/utils/newly_added_detection.py
```diff
@@ -0,0 +1,192 @@
...
+          stderr=subprocess.DEVNULL,
+      )
+  except Exception:  # pylint: disable=broad-exception-caught
+    return set()
+
+  changed = set()
+  for path, touched_lines in parse_changed_line_map(diff_text).items():
+    if not _is_test_file(path):
+      continue
+    try:
```
shralex, line 199: Lets convert to absolute path here and return that instead of path
            # Resolve the absolute path to open the file and return in the set
            full_path = os.path.abspath(os.path.join(repo_root, path))

tests/utils/newly_added_detection.py
```diff
@@ -0,0 +1,192 @@
...
+      with open(path, "r", encoding="utf-8") as handle:
+        source = handle.read()
+    except OSError:
+      continue
+    for name in touched_test_names(source, touched_lines):
+      changed.add((path, name))
```
shralex, line 205: full_path

tests/utils/newly_added_detection.py
mesakhcienet: Thank you for the suggestion!

Currently I kept the three-dot ranges from this list and dropped the two-dot ones. The reasons is: 

- **Three-dot** (`main...HEAD`) = only what my branch added since it forked from main.
- **Two-dot** (`main HEAD`) = every difference between the tips, so it also blames commits that landed on main *after* I branched.

Every test we detect gets force-run in pre-submit (even `scheduled_only`), so two-dot's extra hits drag in unrelated tests: on this branch three-dot found 3 changed test files, two-dot found 11, and with a stale local `main` two-dot ballooned to ~64. 

So I kept `origin/{base}...HEAD` (CI + local with a remote) and `{base}...HEAD` (local dev with no `origin`), both merge-base based. CI never falls through to the second one because `origin/{base}` is always fetched first. Extracted into `_build_diff_commands()` with tests. 

Glad to add a two-dot fallback back if there's a case that needs it!

tests/conftest.py
mesakhcienet: Kept it relative (reasoning in the `newly_added_detection.py:185` / `:191`). `nodeid` is rootdir-relative but `abspath` is CWD-relative, so from a subdir it doubles the path and the match breaks. `pytest.ini` pins rootdir = repo root, so relative-vs-relative is already exact.

tests/utils/newly_added_detection.py
mesakhcienet: Thanks for the suggestion 

I tried it but ended up keeping the repo-relative `path`, because `abspath` turned out to break the match. It resolves against the current working directory, while the conftest side (`item.nodeid`) is relative to the pytest rootdir. When those differ they diverge: collecting from `tests/unit/` turns `os.path.abspath(nodeid)` into `.../tests/unit/tests/unit/..._test.py`, so the tuple no longer matches. 

Since `pytest.ini` fixes rootdir = repo root = git toplevel, the `git diff` paths and the nodeid paths are already the same string, so the current relative compare is exact and CWD-independent. abspath would also pull in a `repo_root` lookup that isn't available on the `maxtext_installed:true` `/deps` runner.

I verified this empirically, but happy to revisit if there's a case where the paths don't line up, or you want us to strictly implement , thanks!

mesakhcienet:
Thanks for the detailed review, @shralex @xibinliu — pushed an update addressing the feedback. summary of what changed:

**Applied as suggested**
- **Detect modified tests, not just new ones** — the detector now maps each changed line (from `git diff`) onto the AST span of every test, so an edited test is flagged as well as a brand-new one (covers funtcions, class methods, `async def`, and decorator lines).
- **Wording** — the `newly_added` marker help now reads "newly introduced **or modified** tests".
- **Copyright headers** bumped to 2026 on both new files.

**Adjusted with a small deviation (details in the threads)**
- **Diff fallback strategies** — adopted the three-dot (merge-base) ranges but dropped the two-dot ones. Two-dot compares tip-vs-tip and over-reports commits that landed on the base after the branch point (3 → 11 test files here, ~64 with a stale local `main`), which would force unrelated `scheduled_only` tests to run. Kept `origin/{base}...HEAD` and a local `{base}...HEAD` fallback, exctracted into `_build_diff_commands()` with unit tests.
- **Absolute paths for the match key** — kept the match relative instead. `item.nodeid` is rootdir-relative while `os.path.abspath()` is CWD-relative, so it breaks when pytest runs from a subdirectory. `pytest.ini` fixes rootdir = repo root = git toplevel, so the current relative compare is already exact. Verified empirically — happy to revisit if there's a case I've missed.

**Verification**
- 17 unit tests (line-range parsing + line→test mapping, incl. the modified-test cases) — all green.
- Lint gate (codespell / pylint / pyink) — passing.
- Real-git behavioural check + an end-to-end conftest probe confirming a `scheduled_only` test the PR touches now runs in pre-submit while an untouched one stays deselected.

Feel free to suggest changes. Thank you!

tests/utils/newly_added_detection.py
mesakhcienet: comment below

shralex (approved):

xibinliu (approved):

PR #4340, opened 2026-07-02
refactor: remove obsolete dynamic checkpoint mismatch handling (cleanup PR #4099)

RexBearIU:
# Description

This PR cleans up the dynamic checkpoint mismatch error-handling logic originally introduced in PR #4099. Since the proactive validation of `scan_layers` from checkpoint metadata (PR #4304) catches mismatches early, the after-the-fact recovery logic is no longer needed.

### Key Changes:
- Removed obsolete helpers (`handle_checkpoint_mismatch`, `is_structural_or_shape_mismatch`) from `checkpointing.py`.
- Simplified the `from_pretrained` loading block in `model_creation_utils.py`.
- Deleted obsolete unit tests (`TestCheckpointMismatchHandling` and related test cases) in `checkpointing_nnx_load_test.py`.
- Updated `test_scan_layers_mismatch_tpu` in `checkpointing_test.py` to assert on proactive verification error messages.
- Fixed an f-string quoting compatibility issue in `inspect_checkpoint.py` to allow pylint to pass on Python < 3.12.

### Dependency Note:
⚠️ **This PR must be merged AFTER PR #4304** (which implements the proactive metadata-based validation).

# Tests

Tested this change by running the following unit and integration tests under the CPU platform:
- `tests/unit/checkpointing_nnx_load_test.py`
- `tests/unit/model_creation_utils_test.py`
- `tests/post_training/unit/lora_utils_test.py`
- `tests/integration/checkpointing_test.py`

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #4341, opened 2026-07-02
[NNX] Checkpoint restore: error on weights missing from the checkpoint

ecnal-cienet:
## Description

On the pure_nnx resume path, a weight present in the model but absent from the checkpoint is handled unsafely. Now that proactive `scan_layers` validation has landed (#4304) and the reactive mismatch guards were removed (#4340), `_populate_pure_dict_from_partial` fills every absent leaf with a default:
- a weight Orbax omits entirely → silently zero-filled (silent accuracy loss);
- a weight Orbax leaves under `partial_restore` as an unmaterialized `jax.ShapeDtypeStruct` → propagated as-is, then failing with a cryptic, low-level compile/allocation error deep in the first `train_step`.

The `scan_layers` check only validates the `scan_layers` flag, so neither case is caught for an arbitrary missing weight (a changed `emb_dim`/`mlp_dim`, a different number of layers, a dropped or renamed module, etc.).

This folds a descriptive guard into `_populate_pure_dict_from_partial`, tracking the tree path and whether the leaf sits under an NNX-only `rngs`/`dropout` subtree:

- rngs/dropout absent → filled with a deterministic default, as before (a Linen-layout checkpoint never carried it).
- any other absent-or-unmaterialized weight → governed by the new `checkpoint_missing_param_policy`:
  - `"error"` (default): raise naming the parameter path and expected shape.
  - `"warn"`: log a WARNING listing the zero-filled path and continue.

`"error"` is the default because a zero-filled or unmaterialized weight is a silentcorrectness failure. A surviving `Shape DtypeStruct` is treated the same as an absent leaf, so the previously cryptic first-`train_step` compile error becomes an explicit, path-named error at restore time.

## Tests

- `tests/unit/checkpointing_nnx_load_test.py::TestMissingParamPolicy` — present weights pass through; rng/dropout absence defaults even under `error`; a missing weight raises naming the path; a surviving `ShapeDtypeStruct` is treated as missing; `warn` zero-fills and logs; default policy is `error`.
- `tests/unit/checkpointing_nnx_missing_param_test.py` — end-to-end through the real `create_orbax_checkpoint_manager → maybe_save_checkpoint → load_state_if_possible` stack: a weight missing from the checkpoint raises naming its path, `warn` zero-fills and continues, and a matching-arch restore returns concrete arrays.

# Checklist

### DIFF
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -194,6 +194,16 @@ def _default_for_sds(sds):
   return jax.jit(_make, out_shardings=sharding)()
 
 
+def _deep_merge(base, overlay):
+  """Recursively merges `overlay` into `base`, returning a new dict. Overlay wins on leaves."""
+  if not (isinstance(base, dict) and isinstance(overlay, dict)):
+    return overlay if overlay is not None else base
+  out = dict(base)
+  for k, v in overlay.items():
+    out[k] = _deep_merge(base.get(k), v) if k in base else v
+  return out
+
+
 def _populate_pure_dict_from_partial(abstract_pure, partial_concrete):
   """Fills `abstract_pure` with values from `partial_concrete` (by path), defaulting the rest.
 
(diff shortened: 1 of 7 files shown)


xibinliu:
`_assert_no_shaped_dtype_struct` has been removed with Jacky's change. Please rebase and probably implement the check in `_populate_pure_dict_from_partial`.

NuojCheng (approved):

src/maxtext/common/checkpointing.py
```diff
@@ -194,23 +194,58 @@ def _make():
...
         for k, v in abstract_pure.items()
     }
-  if partial_concrete is not None and not isinstance(partial_concrete, dict):
+  # A present, materialized leaf takes the restored value. A surviving ShapeDtypeStruct
+  # means Orbax had no matching array, so treat it as absent (handled below).
+  if partial_concrete is not None and not isinstance(partial_concrete, (dict, jax.ShapeDtypeStruct)):
     return partial_concrete
+
+  # Absent (or unmaterialized) leaf. rngs/dropout are expected to be absent from a
+  # Linen-layout checkpoint; anything else is a real weight the checkpoint should have had.
```
xibinliu: If we handles only weights, we don't need handle rngs. Also, using hard-coded NNX_RNG_STATE_KEYS is not a safe solution. Please refer to the comments in PR [#4343](https://github.com/AI-Hypercomputer/maxtext/pull/4343) to manipulate weights/rngs with nnx.state_split

ecnal-cienet: Done. Rebased onto #4343 and reworked to filter by Variable type instead of key names:

- Dropped `NNX_RNG_STATE_KEYS` entirely. The check now does `nnx.split_state(abstract_state, nnx.Param, ...)` and only inspects the `nnx.Param` partition — so as you said, we don't touch rngs at all. nnx.RngState/nnx.BatchStat live in #4343's nnx_aux and are never flagged.
- Kept it a minimal add-on: `_linen_items_to_nnx` and its tests are unchanged; the new `_enforce_missing_weight_policy` just runs at the restore entry points before reshaping.

A missing weight (absent, or left by Orbax as an unmaterialized `ShapeDtypeStruct`) now errors naming the parameter path, or zero-fills under `checkpoint_missing_param_policy=warn`.

PR #4343, opened 2026-07-02
[NNX] Checkpoint: persist rngs/dropout across resumes; policy for missing weights

ecnal-cienet:
## Description

A pure_nnx run saves in the Linen on-disk layout, which has no place for NNX-only `rngs`/`dropout` and batch stats, so that state was dropped on save and refilled on resume from a base `jax.random.key(0)`. Every preemption reset the dropout RNG stream even though the data iterator was correctly restored. The optimizer state was silently reset too (see the `_opt_state_from_linen` fix below).

`to_checkpoint_dict` now partitions the state by Variable type via `split_for_checkpoint` into `(persistent, aux, ephemeral)`. `persistent` is the weights, the optimizer, and any other variable that has to survive a resume; it maps to the Linen `params`/`opt_state`/`step` layout. `aux` is `nnx.RngState` and `nnx.BatchStat`, which Linen never carried, and goes under an `nnx_aux` subtree of `items`. `ephemeral` is `nnx.Cache` and `nnx.Intermediate`, which are recomputed and so are not checkpointed. Selecting the ephemeral types rather than whitelisting `nnx.Param` matters: a custom `nnx.Variable` — such as the frozen `tid2eid` routing table DeepSeek V4 hash routing reads (`deepseek4-284b.yml` sets `first_num_hash_layers: 3`) — otherwise falls out of the checkpoint silently and reverts to zeros on resume, sending every token to expert 0. The missing-weight check only inspects `nnx.Param`, so nothing would warn.

- `nnx_aux` lives inside the single `items` tree, not as a sibling item, so every manager persists it — including the emergency manager, which writes one state tree (an emergency preemption previously lost the RNG stream). A Linen restore targets only `params`/`opt_state`/`step` and ignores the extra `items/nnx_aux`, so the two formats stay interchangeable.
- Restore is the inverse of save. `_linen_items_to_nnx` loads the checkpoint's two disjoint pieces — weights + optimizer, and rngs/dropout/batch stats — back into the matching pieces of `split_for_checkpoint(abstract)` and recombines them with `nnx.merge_state`. The split copies, so the result is the abstract with the checkpoint's values in place and unmaterialized `ShapeDtypeStruct`s wherever the checkpoint carried nothing. `setup_initial_state` builds a fresh concrete init and fills only those placeholders: a present leaf comes from the checkpoint, an absent one keeps its real init value — no zero or `key(0)` defaults. Restore never depends on knowing exactly what was saved, so an old checkpoint with no `nnx_aux`, or a weight the model added, simply keeps its init value. `setup_initial_state` no longer overlays onto the abstract itself and no longer mutates it.
- Fixes a latent bug in `_opt_state_from_linen`: it re-wrapped MaxText's flat opt_state (`{count, mu, nu}`) under a `{0: ...}` chain index, so on resume the optimizer silently reset to zero under the old default-fill. It now round-trips flat, and the step and adam moments are genuinely restored.
- Makes `_opt_state_to_linen` symmetric: it flattened a chain whose only non-empty entry sits at index 0, which `from_linen` cannot rebuild. It now maps a chain to the list-with-None layout and back. This one is latent rather than a live crash — `get_optimizer` always passes a learning-rate schedule, and `scale_by_schedule` carries state at a higher chain index, so MaxText's own optimizers (`adamw` → `{0, 2}`, `sgd` → `{1}`, `adam_pax` → flat) never reduce to the single-entry shape. A scalar learning rate (`optax.adamw(1e-3)`) does, and reproduces the failure.
- `checkpoint_missing_param_policy` gives visibility into a weight the checkpoint doesn't carry. Weights are filtered by Variable type (`nnx.Param`) so only real weights are checked — rngs/dropout/batch stats are never flagged. `warn` (default) logs a warning naming the weight and keeps its init value; `error` raises.
- The params-only load folds back into `load_params_from_path`. On disk the weights live at `params/params/...`: an outer key naming the item, and Flax's `params` collection inside it. A Linen `TrainState.params` is that collection; an NNX params state sits one level below it, so it is wrapped going in and unwrapped coming out. The single-use `_load_linen_params_into_nnx` and `_rebuild_nnx_with_values` are gone, and native `nnx.replace_by_pure_dict` writes the restored values back into the state.

Also addressed in review: the NNX full-state branch restores the grain data iterator; two dead `isinstance(..., nnx.State)` checks were removed; the orbax-v1 "orbax" layout routes `nnx.State` through the Linen→NNX conversion; and a TODO marks that `checkpoint_missing_param_policy` is only enforced on the full-state restore, not on the params-only route.

## Tests

- `tests/unit/checkpointing_nnx_load_test.py` (24) — `TestToCheckpointDict` (typed split: params/opt_state/step vs `nnx_aux`; batch stats split from weights; caches/intermediates excluded; a custom `nnx.Variable` persists rather than being silently dropped; a flat opt_state round-trips without a `{0}` chain wrapper; a chained opt_state round-trips back onto the state), `TestLinenItemsToNnx` (present aux kept; a leaf not on disk kept as a placeholder; no `nnx_aux` key leaves rng as a placeholder; a weights-only checkpoint leaves the optimizer as a placeholder), `TestMissingWeightPolicy` (the `nnx.Param`-type check: present weights pass; a missing weight raises naming its path under `error`; a surviving `ShapeDtypeStruct` counts as missing; `warn` logs and keeps init; `_expected_and_restored_params` splits weights from rng/dropout), and `TestLoadParamsIntoNNX` (a Linen-layout params checkpoint restores into the NNX params state, and the same function returns a Linen `TrainState.params` collection unchanged), plus the emergency-recovers-`nnx_aux` and param-view-split cases.
- `tests/unit/checkpointing_nnx_roundtrip_test.py` (12) — end-to-end through the real `create_orbax_checkpoint_manager → maybe_save_checkpoint → load_state_if_possible` stack, on a chained optimizer so the opt_state reshape is exercised for real: exact full-state fidelity (weights + optimizer + dropout rng + batch stats), per-component restore, old-checkpoint-missing-fields-keep-init, cache exclusion, the restored state loading into a live model and running, resume continuity of the rng stream, and the `load_full_state_from_path` route under both orbax v0 and v1.
- `tests/unit/setup_initial_state_nnx_test.py` (5) — drives the real `setup_initial_state` on CPU (patching only `get_abstract_state` for a tiny model), on an un-chained optimizer so the flat opt_state path is covered too: full restore via overlay, fresh init when no checkpoint exists, missing fields keeping their init value, the emergency manager's un-wrapped overlay, and a params-only load that writes the weights and leaves the optimizer and rng stream at init.
- `tests/unit/checkpointing_nnx_missing_param_test.py` (3) — end-to-end through the real save/restore stack: `error` raises naming the weight, `warn` keeps init (the missing weight stays a placeholder the trainer fills), and a matching-arch restore carries the weight.
- `tests/unit/train_state_nnx_checkpoint_test.py` (19) — the save path now reshapes with `to_checkpoint_dict(state)` (which splits by Variable type) instead of `to_linen_checkpoint_dict(state.to_pure_dict())`, so the `maybe_save_checkpoint` tests probe that call to assert save work is dispatched on a checkpoint step and skipped otherwise.
- Verified end-to-end on a v6e-8, for both opt_state shapes: a pure_nnx `train → save → resume` on `adam_pax` (flat) and on `adamw` (chained) each continues past the checkpoint step. Inspecting the on-disk checkpoint for the adamw run confirms `opt_state` is the chain layout with `count=10` and all 16 mu/nu leaves non-zero (a fresh init would be all zeros), and `nnx_aux` carries the advanced RNG stream (`dropout/rngs/dropout/count=20`) rather than a base `key(0)`. The checkpoint the `(persistent, aux, ephemeral)` split writes is path-for-path identical to the one the earlier `nnx.Param` split wrote (82 leaves, same tree), so the fix changes nothing for a model without custom `nnx.Variable`s.
- 44 passed across the four files above, plus `train_state_nnx_checkpoint_test.py` (19, covering nested-chain optimizers).

# Checklist

### DIFF
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -24,7 +24,6 @@ from etils import epath
 from flax import nnx
 from flax.training import train_state
 import jax
-import jax.numpy as jnp
 from maxtext.utils.globals import DEFAULT_OCDBT_TARGET_DATA_FILE_SIZE
 from maxtext.input_pipeline.multihost_dataloading import MultiHostDataLoadIterator
 from maxtext.input_pipeline.multihost_dataloading import RemoteIteratorWrapper
@@ -170,48 +169,79 @@ class GrainCheckpointRestore(ocp.args.CheckpointArgs):
   process_count: Optional[int] = None
 
 
-def _default_for_sds(sds):
-  """Returns a deterministic value matching `sds` shape/dtype/sharding.
+def _weight_mismatches(want, have, path=()):
+  """Returns `(path, problem)` for each weight in `want` that `have` didn't restore faithfully.
 
-  Used to fill NNX-only state (rngs/dropout) that the Linen on-disk layout never
-  carried. Materializes under jit with the target out_shardings so it works on
-  multi-host meshes (device_put can't place a global sharding whose devices
-  aren't
-  all addressable from this process).
+  A weight is wrong if the checkpoint didn't carry it -- absent, or left by Orbax as an
+  unmaterialized ShapeDtypeStruct -- or carried it at a different shape. Only the shape can
+  disagree: Orbax casts a restored array to the target's dtype.
   """
-  if not (hasattr(sds, "dtype") and hasattr(sds, "shape")):
-    return sds
+  if isinstance(want, dict):
+    out = []
+    for k, v in want.items():
+      out.extend(_weight_mismatches(v, have.get(k) if isinstance(have, dict) else None, path + (k,)))
+    return out
+  name = "/".join(str(p) for p in path)
+  if have is None or isinstance(have, jax.ShapeDtypeStruct):
+    return [(name, f"missing (model expects {getattr(want, 'shape', '?')} {getattr(want, 'dtype', '?')})")]
+  want_shape, got_shape = getattr(want, "shape", None), getattr(have, "shape", None)
+  if want_shape is not None and got_shape is not None and tuple(want_shape) != tuple(got_shape):
+    return [(name, f"shape {tuple(got_shape)} but the model expects {tuple(want_shape)}")]
+  return []
 
-  def _make():
-    if "key" in str(sds.dtype):
-      base = jax.random.key(0)
-      return base if sds.shape == () else jax.random.split(base, int(np.prod(sds.shape))).reshape(sds.shape)
-    return jnp.zeros(sds.shape, dtype=sds.dtype)
 
-  sharding = getattr(sds, "sharding", None)
-  if sharding is None:
-    return _make()
-  return jax.jit(_make, out_shardings=sharding)()
+def _expected_and_restored_params(abstract_nnx_state, restored_linen):
+  """Returns the model's expected weights and the checkpoint's restored weights, as pure dicts.
 
-
-def _populate_pure_dict_from_partial(abstract_pure, partial_concrete):
-  """Fills `abstract_pure` with values from `partial_concrete` (by path), defaulting the rest.
-
-  Paths present in `partial_concrete` take the restored value; paths absent from
-  it
-  (NNX-only state the Linen checkpoint never had) get `_default_for_sds`.
+  Splits the abstract by Variable type (nnx.Param) so only real weights are compared --
+  rngs/dropout/batch stats live in `nnx_aux` and are restored separately.
   """
-  if isinstance(abstract_pure, dict):
-    return {
-        k: _populate_pure_dict_from_partial(
-            v,
-            partial_concrete.get(k) if isinstance(partial_concrete, dict) else None,
-        )
-        for k, v in abstract_pure.items()
-    }
-  if partial_concrete is not None and not isinstance(partial_concrete, dict):
-    return partial_concrete
-  return _default_for_sds(abstract_pure)
+  want = nnx.split_state(abstract_nnx_state, nnx.Param, ...)[0].to_pure_dict().get("model", {})
+  have = restored_linen.get("params", {}).get("params", {})
+  return want, have
+
+
+def _raise_on_weight_mismatch(want, have):
+  """Raises if the restored weights (`have`) don't match what the model expects (`want`).
+
+  Both are pure dicts, so this works for any structure. `partial_restore` returns a weight the
+  checkpoint doesn't carry as an unmaterialized ShapeDtypeStruct, and Orbax restores a stored
+  array at its own shape rather than the target's. Either way it reaches the model as an
+  untrained init value (a silent accuracy loss) or fails much later, deep in the first step,
+  without naming the weight.
+  """
+  problems = _weight_mismatches(want, have)
+  if not problems:
+    return
+  lines = "\n".join(f"  - '{p}': {why}" for p, why in problems)
+  raise ValueError(
+      "Checkpoint does not match the model:\n"
+      f"{lines}\n"
+      "Verify the checkpoint matches the model architecture (emb_dim, mlp_dim, num layers, scan_layers)."
+  )
+
+
+def _linen_items_to_nnx(restored_linen, abstract_nnx_state):
+  """Reshapes a restored Linen-layout `items` dict into an NNX state.
+
+  The inverse of `to_checkpoint_dict`, over the same `split_for_checkpoint` partition. The Linen
+  weights + optimizer fill `linen_state`; the `nnx_aux` state (rngs/dropout, batch stats, custom
+  variables) fills `aux`; the two are recombined with `nnx.merge_state`. The split copies, so the
+  caller's abstract is untouched. Leaves the checkpoint didn't carry -- including the caches it
+  never stores -- stay unmaterialized `ShapeDtypeStruct`s; the caller fills them from a fresh init.
+  """
+  linen_state, aux_state, ephemeral = train_state_nnx.split_for_checkpoint(abstract_nnx_state)
+  weights = train_state_nnx.from_linen_checkpoint_dict(restored_linen)
+  if "model" in weights:
+    nnx.replace_by_pure_dict(linen_state, {"model": weights["model"]})
+  if "optimizer" in weights:
+    nnx.replace_by_pure_dict(linen_state, {"optimizer": weights["optimizer"]})
+
+  nnx_aux = restored_linen.get("nnx_aux")
+  if nnx_aux:
+    nnx.replace_by_pure_dict(aux_state, nnx_aux)
+
+  return nnx.merge_state(linen_state, aux_state, ephemeral)
 
 
 def _load_linen_checkpoint_into_nnx(
@@ -223,12 +253,12 @@ def _load_linen_checkpoint_into_nnx(
 ):
   """Restores a Linen-layout checkpoint into an NNX state (pure_nnx resume).
 
-  Restores against a Linen-shape abstract, reshapes back via
-  `from_linen_checkpoint_dict`, then fills NNX-only rngs/dropout with defaults.
+  Restores a Linen-shape target that includes `nnx_aux`, then reshapes back via
+  `_linen_items_to_nnx`. rngs/dropout/batch stats come from `items/nnx_aux` when
+  present, else keep their fresh init value. A genuinely-missing weight raises.
   """
   max_logging.log(f"Restoring Linen-layout checkpoint into NNX state at {path}")
-  nnx_abstract_pure = abstract_nnx_state.to_pure_dict()
-  linen_abstract = train_state_nnx.to_linen_checkpoint_dict(nnx_abstract_pure)
+  linen_abstract = train_state_nnx.to_checkpoint_dict(abstract_nnx_state)
   ckptr = ocp.Checkpointer(
       ocp.PyTreeCheckpointHandler(
           restore_concurrent_gb=checkpoint_storage_concurrent_gb,
@@ -240,8 +270,8 @@ def _load_linen_checkpoint_into_nnx(
   restore_args = ocp.checkpoint_utils.construct_restore_args(linen_abstract)
   restored = ocp.args.PyTreeRestore(item=linen_abstract, restore_args=restore_args, partial_restore=True)
   restored = ckptr.restore(epath.Path(path), args=restored)
-  partial_nnx = train_state_nnx.from_linen_checkpoint_dict(restored)
-  return _populate_pure_dict_from_partial(nnx_abstract_pure, partial_nnx)
+  _raise_on_weight_mismatch(*_expected_and_restored_params(abstract_nnx_state, restored))
+  return _linen_items_to_nnx(restored, abstract_nnx_state)
 
 
 def _restore_emergency_linen_checkpoint_into_nnx(
@@ -250,10 +280,14 @@ def _restore_emergency_linen_checkpoint_into_nnx(
     abstract_nnx_state,
     map_to_pspec,
 ):
-  """Restores an emergency Linen-layout checkpoint into an NNX state."""
+  """Restores an emergency Linen-layout checkpoint into an NNX state.
+
+  The `nnx_aux` subtree is stored inside `items`, so an emergency checkpoint
+  carries it too; it's restored when present and otherwise kept at its fresh
+  init value. A genuinely-missing weight raises.
+  """
   max_logging.log(f"Restoring emergency Linen-layout checkpoint into NNX state at step {step}")
-  nnx_abstract_pure = abstract_nnx_state.to_pure_dict()
-  linen_abstract = train_state_nnx.to_linen_checkpoint_dict(nnx_abstract_pure)
+  linen_abstract = train_state_nnx.to_checkpoint_dict(abstract_nnx_state)
   restore_args = jax.tree_util.tree_map(map_to_pspec, linen_abstract)
   checkpoint_args = ocp.args.PyTreeRestore(
       item=linen_abstract,
@@ -261,52 +295,8 @@ def _restore_emergency_linen_checkpoint_into_nnx(
       partial_restore=True,
   )
   restored = checkpoint_manager.restore(step, args=Composite(state=checkpoint_args)).state
-  partial_nnx = train_state_nnx.from_linen_checkpoint_dict(restored)
-  return _populate_pure_dict_from_partial(nnx_abstract_pure, partial_nnx)
-
-
-def _rebuild_nnx_with_values(abstract_nnx_state, concrete_weights):
-  """Fills each Variable in `abstract_nnx_state` with the matching restored array."""
-  leaves, treedef = jax.tree_util.tree_flatten(abstract_nnx_state, is_leaf=lambda x: isinstance(x, nnx.Variable))
-  concrete = jax.tree_util.tree_leaves(concrete_weights)
-  if len(leaves) != len(concrete):
-    raise ValueError(
-        f"Params load leaf-count mismatch: {len(leaves)} abstract Variables vs" f" {len(concrete)} restored."
-    )
-  new_leaves = [v.replace(value=a) if isinstance(v, nnx.Variable) else a for v, a in zip(leaves, concrete)]
-  return jax.tree_util.tree_unflatten(treedef, new_leaves)
-
-
-def _load_linen_params_into_nnx(
-    path,
-    nnx_params_abstract,
-    checkpoint_storage_concurrent_gb,
-    use_ocdbt,
-    use_zarr3,
-):
-  """Weight-only load of a Linen-layout checkpoint into an NNX params state.
-
-  Reuses `to_linen_checkpoint_dict` (wrapping the params under `model`) to build
-  the
-  `params/params/...` restore target, then rebinds the restored weights into the
-  NNX params Variables.
-  """
-  max_logging.log(f"Restoring Linen-layout params into NNX state at {path}")
-  linen_abstract = train_state_nnx.to_linen_checkpoint_dict({"model": nnx_params_abstract.to_pure_dict()})
-  ckptr = ocp.Checkpointer(
-      ocp.PyTreeCheckpointHandler(
-          restore_concurrent_gb=checkpoint_storage_concurrent_gb,
-          save_concurrent_gb=checkpoint_storage_concurrent_gb,
-          use_ocdbt=use_ocdbt,
-          use_zarr3=use_zarr3,
-      )
-  )
-  restore_args = ocp.checkpoint_utils.construct_restore_args(linen_abstract)
-  restored = ckptr.restore(
-      epath.Path(path),
-      args=ocp.args.PyTreeRestore(item=linen_abstract, restore_args=restore_args, partial_restore=True),
-  )
-  return _rebuild_nnx_with_values(nnx_params_abstract, restored["params"]["params"])
+  _raise_on_weight_mismatch(*_expected_and_restored_params(abstract_nnx_state, restored))
+  return _linen_items_to_nnx(restored, abstract_nnx_state)
 
 
 def _load_full_state_from_path(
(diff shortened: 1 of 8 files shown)


src/maxtext/common/checkpointing.py
```diff
@@ -194,6 +194,16 @@ def _make():
   return jax.jit(_make, out_shardings=sharding)()
 
 
+def _deep_merge(base, overlay):
```
NuojCheng: add a test protecting this funciton?

NuojCheng (approved):

src/maxtext/common/checkpointing.py
ecnal-cienet: Added, thank you.

src/maxtext/common/checkpointing.py
xibinliu, line 860: This case is not covered for `pure_nnx=True`, ie, abstract_unboxed_pre_state is nnx.State.

We need handle `dataset_type="grain"` for NNX.

xibinliu: This is the dead code because the previous checks have returned for nnx.State cases. It is guaranteed that at this point the abstract_unboxed_pre_state is linen.

xibinliu, line 1175: Since emergency checkpointers only write a single state tree, we should move "nnx_aux" into the "items" structure for all cases.

Can we have this for NNX checkpoints?

```
step_1000/
├── items/                  <── The unified items checkpointable
│   ├── params/             <── Learnable weight matrices
│   ├── opt_state/          <── Optimizer states (mapped from dict to list)
│   ├── step                <── Step counter (uint32 cast to int32)
│   └── nnx_aux/            <── RNGs and dropout state saved directly inside items/
└── iter/                   <── Optional dataset iterator progress (e.g., PyGrain)
```

src/maxtext/common/checkpointing.py
```diff
@@ -250,7 +289,11 @@ def _restore_emergency_linen_checkpoint_into_nnx(
     abstract_nnx_state,
     map_to_pspec,
 ):
-  """Restores an emergency Linen-layout checkpoint into an NNX state."""
+  """Restores an emergency Linen-layout checkpoint into an NNX state.
+
+  Emergency checkpoints don't carry the separate rngs/dropout `nnx_aux` item, so
```
xibinliu: Left a comment in save path: can we pack nnx_aux under "items"

src/maxtext/common/checkpointing.py
xibinliu: At this point, `abstract_unboxed_pre_state` is surely instance of `nnx.State`

xibinliu, line 340: Linen convertion is missing here?

src/maxtext/common/train_state_nnx.py
```diff
@@ -102,6 +102,26 @@ def _strip_rng_state(tree):
   return out
 
 
+def extract_rng_state(tree):
```
xibinliu: There could be other non-weights variables in NNX state (see [nnx Variables](https://flax.readthedocs.io/en/stable/api_reference/flax.nnx/variables.html)), I think we should only leave weights.

Please see the other comment about manipulating on nnx.State, instead of pure dict.

src/maxtext/common/checkpointing.py
```diff
@@ -224,7 +255,8 @@ def _load_linen_checkpoint_into_nnx(
   """Restores a Linen-layout checkpoint into an NNX state (pure_nnx resume).
 
   Restores against a Linen-shape abstract, reshapes back via
-  `from_linen_checkpoint_dict`, then fills NNX-only rngs/dropout with defaults.
+  `from_linen_checkpoint_dict`, then restores the separately-saved rngs/dropout
+  state when present (falling back to deterministic defaults otherwise).
   """
   max_logging.log(f"Restoring Linen-layout checkpoint into NNX state at {path}")
   nnx_abstract_pure = abstract_nnx_state.to_pure_dict()
```
xibinliu: Before converting to pure dict, I think we should use nnx.state_split to filter out nnx.Param (weights), nnx.RngState (RngKey / RngCount), and nnx.BatchStat. Store nnx.Param in Linen compatible format, with nnx.RngState and nnx.BatchStatand goes to nnx_aux.

This should be more accurate than working on pure dict.
```
params, rngs, batchstat, rest = nnx.state_split(abstract_nnx_state, nnx.Param, nnx.RngState, nnx.BatchStat, ...)

# params goes to items.params.params
# items.nnx_aux contains rngs, batchstat
...
```
In the restore path, we should do:
```
restored_state = nnx.state.merge(rest, restored_params, restored_rngs, restored_batchstat)

```

src/maxtext/common/train_state_nnx.py
ecnal-cienet: The string-keyed `extract_rng_state` is removed. The type-based split now correctly separates non-weight variables, so BatchStat (batch-norm running stats) goes to `nnx_aux` instead of being saved as a weight, which the old key-name approach missed.

src/maxtext/common/checkpointing.py
ecnal-cienet: Done. `to_checkpoint_dict` now splits by Variable type via `nnx.split_state(state, nnx.Param, nnx.RngState, nnx.BatchStat, ...)` and recombines on restore with `nnx.merge_state`.

src/maxtext/common/checkpointing.py
ecnal-cienet: Done. `nnx_aux` is now a subtree inside `items` (`items/{params, opt_state, step, nnx_aux}`) rather than a sibling composite item. Dropped the separate item handler and the `nnx_aux_state` plumbing.

src/maxtext/common/checkpointing.py
ecnal-cienet: Removed.

ecnal-cienet: The NNX full-state branch now restores the grain iterator, routing through the existing `_restore_grain_iterator` (reusing the multi-process/elastic/remote handling), then reshapes `items` back to NNX.

ecnal-cienet: Done. Because `nnx_aux` now lives inside the single `items` tree, the emergency path persists and restores it too — previously emergency preemption lost the RNG stream. Added a unit test (`test_emergency_restore_recovers_nnx_aux`).

ecnal-cienet: I read this the same as the line-880 thread: `nnx.State` already returns earlier in this branch, so this `isinstance(..., nnx.State)` check is dead and I removed it.

ecnal-cienet: Fixed. The orbax-v1 "orbax" layout path now routes `nnx.State` through `_load_linen_checkpoint_into_nnx` (the Linen -> NNX conversion), matching the v0 path.

src/maxtext/common/train_state_nnx.py
```diff
@@ -193,3 +195,20 @@ def from_linen_checkpoint_dict(linen_pure_dict):
...
+  """Reshapes an nnx.State into the on-disk checkpoint layout, splitting by Variable type.
+
+  Weights (nnx.Param) and the optimizer state map to the Linen
+  params/opt_state/step layout, so pure_nnx and Linen checkpoints stay
+  interchangeable. rngs/dropout (nnx.RngState) and batch stats (nnx.BatchStat) go
+  under an `nnx_aux` subtree so they survive a resume. Works on a concrete state
+  (save) or an abstract state (restore target).
+  """
+  params, rng_state, batch_stats, rest = nnx.split_state(state, nnx.Param, nnx.RngState, nnx.BatchStat, ...)
+  linen = to_linen_checkpoint_dict(nnx.merge_state(params, rest).to_pure_dict())
```
xibinliu: We should not merge `rest` with the `params` for linen. I think `rest` might contain cache, intermediates, and other variables of the model, and they should be ignored from the checkpointing. We should pass only `params` to `to_linen_checkpoint_dict`.

src/maxtext/common/train_state_nnx.py
```diff
@@ -66,7 +66,9 @@ def apply_gradients(self, grads: Any, **kwargs):
 #   2. weights: model/... -> params/params/... (Linen `params` collection)
 #   3. opt_state: int-keyed dict (empty entries skipped) -> list with None for EmptyState,
 #      and mu/nu wrapped under the `params` collection
-# NNX-only rngs/dropout state is dropped (Linen never had it).
+# NNX-only state Linen never had (rngs/dropout and batch stats) is split out by
```
xibinliu: Rephrase:
```
# NNX-only state (rngs/dropout and batch stats), which Linen never has,  is split...
```

src/maxtext/common/checkpointing.py
```diff
@@ -214,6 +224,25 @@ def _populate_pure_dict_from_partial(abstract_pure, partial_concrete):
   return _default_for_sds(abstract_pure)
 
 
+def _all_materialized(tree):
+  """True if Orbax restored every leaf, i.e. none is left as a jax.ShapeDtypeStruct."""
+  return not any(isinstance(leaf, jax.ShapeDtypeStruct) for leaf in jax.tree_util.tree_leaves(tree))
+
+
+def _linen_items_to_nnx(restored_linen, nnx_abstract_pure):
+  """Reshapes a restored Linen-layout `items` dict back into the NNX pure-dict state.
```
xibinliu: 1. `nnx_abstract_pure` should be passed in with type of nnx.State, not a pure dict
2. Here we should split the state to params, rng_states, norm_stats, rest
3. Get restored params from restored linen
4. get restored rng_states and norm_stats from restored linen
5. merge (restored params, restored rng_states and norm_stats, rest) to NNX state

ecnal-cienet: Restructured per the suggestion — `_linen_items_to_nnx` now takes the abstract as an `nnx.State`. It restores `nnx.Param` weights and the optimizer from the Linen layout and `nnx.RngState`/`nnx.BatchStat` from `nnx_aux`, filling whatever the checkpoint didn't carry (rng when nnx_aux is absent, caches/intermediates) from the abstract's defaults. The optimizer is sourced from the checkpoint's `opt_state`/`step`, not the abstract — `test_optimizer_state_restored_from_checkpoint` locks that in. I kept the `opt_state` Linen <-> NNX reshape in `from_linen_checkpoint_dict` since it's a structural remap, not a plain type split.

src/maxtext/common/train_state_nnx.py
ecnal-cienet: Reworded

src/maxtext/common/train_state_nnx.py
ecnal-cienet: Actually, `rest` also holds the optimizer state (mu/nu/step aren't `nnx.Param`), which must be checkpointed — so I pass params + the optimizer explicitly and drop the rest: `to_linen_checkpoint_dict({"model": params, "optimizer": rest["optimizer"]})`. Added `test_cache_and_intermediates_excluded`.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1494,44 +1494,53 @@ def setup_initial_state(
...
+        is_emergency = isinstance(
+            checkpoint_manager,
+            (
+                emergency_checkpoint_manager.CheckpointManager,
+                emergency_replicator_checkpoint_manager.ReplicatorCheckpointManager,
+            ),
+        )
+        # data_iterator state is updated in place during restore.
+        overlay = restored if is_emergency else restored["items"]
+        nnx.replace_by_pure_dict(state, overlay)
```
xibinliu: `ValueError` is key exists in overlay but is absent from state.

src/maxtext/common/checkpointing.py
```diff
@@ -170,48 +169,88 @@ class GrainCheckpointRestore(ocp.args.CheckpointArgs):
...
-        for k, v in abstract_pure.items()
-    }
-  if partial_concrete is not None and not isinstance(partial_concrete, dict):
-    return partial_concrete
-  return _default_for_sds(abstract_pure)
+  overlay = train_state_nnx.from_linen_checkpoint_dict(restored_linen)
+  aux = restored_linen.get("nnx_aux")
+  if aux:
+    overlay = _deep_merge(overlay, aux)
+  return _drop_unmaterialized(overlay)
```
xibinliu: Can we do something like this:

```
nnx_partial_state = nnx.replace_from_pure_dict(state, overlay)
aux_state = nnx.replace_from_pure_dict(state, aux)

nnx_state = nnx.merge_state(nnx_partial_state, aux_state)
```

src/maxtext/common/checkpointing.py
```diff
@@ -170,48 +169,88 @@ class GrainCheckpointRestore(ocp.args.CheckpointArgs):
...
+    out = []
+    for k, v in want.items():
+      out.extend(_missing_weight_paths(v, have.get(k) if isinstance(have, dict) else None, path + (k,)))
+    return out
+  if have is None or isinstance(have, jax.ShapeDtypeStruct):
+    return [("/".join(str(p) for p in path), getattr(want, "shape", "?"), getattr(want, "dtype", "?"))]
+  return []
+
+
+def _enforce_missing_weight_policy(abstract_nnx_state, restored_linen, missing_param_policy="warn"):
```
xibinliu: Pass in abstract params and restored params so this function become a general one and can be used for any structures.

ecnal-cienet:
Hi @xibinliu, I managed to remove `_drop_unmaterialized` helper from the code. For `_deep_merge`, it unions the two disjoint on-disk pieces of a restored checkpoint into one overlay: the Linen-format weights+optimizer (`from_linen_checkpoint_dict`) and the nnx_aux rngs/dropout/batch-stats. Both nest under model/ but at different leaves (`model/<layer>/kernel` vs `model/<layer>/rngs/...`), so a shallow dict.update at model would clobber the whole weights subtree with the rngs-only one — the merge has to recurse to the leaf level. There's no cleaner built-in for a leaf-level union of two partial pure dicts, so a small recursive merge is the simplest correct tool here. Added a comment in the code.

src/maxtext/common/train_state_nnx.py
```diff
@@ -193,3 +198,40 @@ def from_linen_checkpoint_dict(linen_pure_dict):
...
+    aux:        rngs/dropout and batch stats, which Linen never carried. Goes under `nnx_aux`.
+    ephemeral:  caches and intermediates, which are recomputed and so are not checkpointed.
+
+  Selecting the ephemeral variables by type rather than whitelisting nnx.Param means a new
+  Variable subclass persists by default instead of being silently dropped. Save and restore must
+  partition identically, so both go through here.
+  """
+  rng_state, batch_stats, caches, intermediates, persistent = nnx.split_state(
+      state, nnx.RngState, nnx.BatchStat, nnx.Cache, nnx.Intermediate, ...
+  )
```
xibinliu, line 222: let's get params explicitly, in case the `persistent` leaves some nnx.Variable in the "model" part. We want the "model" to be pure nnx.Param so they can covert to linen params.
```
  params, rng_state, batch_stats, reset = nnx.split_state(
      state, nnx.Param, nnx.RngState, nnx.BatchStat, ...
  )


return params, nnx.merge_state(rng_state, batch_stats), rest
```

src/maxtext/common/train_state_nnx.py
```diff
@@ -193,3 +198,40 @@ def from_linen_checkpoint_dict(linen_pure_dict):
...
+
+def to_checkpoint_dict(state: nnx.State):
+  """Reshapes an nnx.State into the on-disk checkpoint layout.
+
+  Weights and the optimizer map to the Linen params/opt_state/step layout, so pure_nnx
+  and Linen checkpoints stay interchangeable; rngs/dropout and batch stats go under an
+  `nnx_aux` subtree so they survive a resume. Works on a concrete state (save) or an
+  abstract state (restore target).
+  """
+  persistent, aux_state, _ = split_for_checkpoint(state)
```
xibinliu:
```
params_state, aux_state, rest = split_for_checkpoint(state)

linen = to_linen_checkpoint_dict({"model": params_state.model.to_pure_dict(), "optimizer": rest.optimizer.to_pure_dict() if rest.optimizer is not None else {})
```

src/maxtext/common/checkpointing.py
```diff
@@ -170,48 +169,80 @@ class GrainCheckpointRestore(ocp.args.CheckpointArgs):
...
-        k: _populate_pure_dict_from_partial(
-            v,
-            partial_concrete.get(k) if isinstance(partial_concrete, dict) else None,
-        )
-        for k, v in abstract_pure.items()
-    }
-  if partial_concrete is not None and not isinstance(partial_concrete, dict):
-    return partial_concrete
-  return _default_for_sds(abstract_pure)
+  persistent, aux_state, ephemeral = train_state_nnx.split_for_checkpoint(abstract_nnx_state)
```
xibinliu:
```
params_state, aux_state, rest = split_for_checkpoint(abstract_nnx_state)
restored_state = train_state_nnx.from_linen_checkpoint_dict(restored_linen)

if "model" in restored_state:
    nnx.replace_by_pure_dict(params_state.model, restored_state["model"]) # Make sure only replace the "model" part
if "optimizer" in restored_state:
    nnx.replace_by_pure_dict(rest, {"optimizer": restored_state["optimizer"]})

...

return nnx.merge_state(params_state, aux_state, rest)
```

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1501,44 +1501,62 @@ def setup_initial_state(
...
+        # The restore already overlaid the checkpoint onto a copy of the abstract, so a leaf it
+        # didn't carry is still an unmaterialized placeholder. Fill those from the fresh init: a
+        # present leaf comes from the checkpoint, an absent one keeps its init value.
+        overlay = restored if is_emergency else restored["items"]
+        merged = jax.tree.map(
+            lambda ckpt, init: init if isinstance(ckpt, jax.ShapeDtypeStruct) else ckpt,
+            overlay.to_pure_dict(),
+            state.to_pure_dict(),
+            is_leaf=lambda x: isinstance(x, jax.ShapeDtypeStruct),
+        )
```
xibinliu, line 1534:
```
# restored_state: a mix of restored params, optimizers, nnx_aux (rngs, batch_stat), and rest un-initialized abstract variables.
# state: initialized variables, including params, optimizers, nnx_aux, and the rest.
# The goal: replace state with params, optimizers, nnx_aux from the restored_state, and leave the rest initialized variables

restored_state = restored if is_emergency else restored["items"] 

params_restored, aux_restored, rest_restored = train_state_nnx.split_for_checkpoint(restored_state)

nnx.update(state, params_restored, aux_restored)
nnx.update(state.optimizer, rest_restored.optimizer)
```

ecnal-cienet: I tested but it doesn't work here, because it writes the placeholder for anything the checkpoint didn't carry.

When a checkpoint has no `nnx_aux` (a Linen-era checkpoint, or any leaf that wasn't stored), `_linen_items_to_nnx` returns that leaf as an unmaterialized `jax.ShapeDtypeStruct`. `nnx.update` / `replace_by_pure_dict` write whatever value is in the source, so they'd copy the `ShapeDtypeStruct` straight into the init state — the rng ends up as a placeholder instead of a real key. I checked it:

| checkpoint | `nnx.update(state, restored)` | current `jax.tree.map` |
|---|---|---|
| has `nnx_aux` | dropout count concrete (Yes) | (Yes) |
| no `nnx_aux` (Linen-era) | dropout count → `ShapeDtypeStruct` (No) | concrete (Yes) |

The `is_leaf=lambda x: isinstance(x, jax.ShapeDtypeStruct)` in the `jax.tree.map` is exactly the "skip the placeholder, keep init" that `nnx.update` can't express, and that no-`nnx_aux` path is covered by `test_old_checkpoint_without_nnx_aux_keeps_init_value`.

src/maxtext/common/train_state_nnx.py
ecnal-cienet: Done

src/maxtext/common/train_state_nnx.py
ecnal-cienet: Done

src/maxtext/common/checkpointing.py
ecnal-cienet: Done

xibinliu (approved):

PR #4344, opened 2026-07-03
feat(scan_layers): dynamically set scan_layers from checkpoint metadata

RexBearIU:
# Description

This PR is a follow-up to PR #4304 and PR #4340. It introduces dynamic synchronization of `scan_layers` configuration from checkpoint metadata when a checkpoint path is provided and `scan_layers` is not explicitly specified by the user on the command-line, in environment variables, or as programmatic keyword arguments.

### Key Changes:
- **Robust Tracking of Explicit User Configurations (`src/maxtext/configs/pyconfig.py`):**
  - Keeps track of overridden configuration parameters from the environment (`env_keys`).
  - Sets Pydantic's internal `__pydantic_fields_set__` to only the union of explicit CLI, environment, and keyword argument keys. This avoids the default Pydantic behavior where unpacking a dictionary of default configurations marks every single field as explicitly set.
- **Helper for Sync and Verification (`src/maxtext/utils/model_creation_utils.py`):**
  - Extracted dynamic configuration resolution and verification into `verify_and_sync_scan_layers(config)`.
  - Inside `from_pretrained`, resolved the active config via `config = verify_and_sync_scan_layers(config)` without verbose comments, keeping the code highly readable.
  - If `scan_layers` is explicitly specified by the user and mismatches checkpoint metadata, a `ValueError` configuration mismatch error is raised.

# Tests

- Added a comprehensive suite of unit tests in `TestVerifyAndSyncScanLayers` inside `tests/unit/model_creation_utils_test.py`:
  - `test_sync_to_false_when_implicit`: Asserts that `scan_layers` dynamically updates to `False` from checkpoint metadata when not explicitly specified by the user.
  - `test_sync_to_true_when_implicit`: Asserts that `scan_layers` stays/updates to `True` when checkpoint metadata is `True`.
  - `test_explicit_match_raises_no_error`: Asserts that providing matching explicit values does not trigger validation errors.
  - `test_explicit_mismatch_raises_value_error`: Asserts that an explicit conflict raises the expected `ValueError` configuration mismatch error.
- Verified all tests in `model_creation_utils_test.py` pass successfully:
  ```bash
  PYTHONPATH=src python3 -m pytest tests/unit/model_creation_utils_test.py
  ```

# Checklist

### DIFF
--- a/docs/guides/checkpointing_solutions/convert_checkpoint.md
+++ b/docs/guides/checkpointing_solutions/convert_checkpoint.md
@@ -78,7 +78,7 @@ You can find your converted checkpoint files under `${BASE_OUTPUT_DIRECTORY}/0/i
 ### Key Parameters
 
 - `model_name`: The specific model identifier. It must match a supported entry in the MaxText [globals.py](https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/src/maxtext/utils/globals.py#L46C1-L46C7).
-- `scan_layers`: Controls whether the output uses a scanned (`scan_layers=true`) or unscanned (`scan_layers=false`) checkpoint format. Refer [to the Checkpoints guide](checkpoints) for more information. **IMPORTANT:** This setting *must* match the `scan_layers` value used during model training or loading. A mismatch will cause PyTree loading errors (though MaxText will intercept these and raise a descriptive `ValueError` explaining the mismatch).
+- `scan_layers`: Controls whether the output uses a scanned (`scan_layers=true`) or unscanned (`scan_layers=false`) checkpoint format. Refer [to the Checkpoints guide](checkpoints) for more information. **Note:** When resuming or loading a checkpoint in MaxText, this setting will automatically be loaded from the checkpoint metadata, meaning you do not have to manually specify it during model execution. If you do explicitly specify a value for `scan_layers`, it must match the checkpoint's saved configuration, or a `ValueError` mismatch error will be raised.
 - `use_multimodal`: Indicates if multimodality is used, important for Gemma3.
 - `base_output_directory`: The path where the converted Orbax checkpoint will be stored; it can be Google Cloud Storage (GCS) or local.
 - `hardware=cpu`: The conversion script runs on a CPU machine.
@@ -298,9 +298,9 @@ Here is an example [PR to add support for gemma3 multi-modal model](https://gith
 
 - **Error:** `Type ShapeDtypeStruct is not a valid JAX type` or generic **PyTree structure/shape mismatches** (e.g., Orbax reporting `"X/Y paths matched"`, such as `143/145 paths`).
 
-  - **Cause: Configuration mismatch** (e.g., `scan_layers`) between the checkpoint conversion script (e.g., `to_maxtext.py` or `to_huggingface.py`) and the trainer/inference runner (e.g., `train.py`).
+  - **Cause: Configuration mismatch** (e.g., `scan_layers`) between the checkpoint conversion script (e.g., `to_maxtext.py` or `to_huggingface.py`) and the trainer/inference runner (e.g., `train.py`). (Note: Since MaxText automatically loads `scan_layers` from the checkpoint's saved metadata, you should only encounter this error if you explicitly set a mismatching value on the command line, which raises a `ValueError` mismatch error).
 
-  - **Solution:** Ensure the `scan_layers` flag is set to the exact same value (`True` or `False`) in both the conversion command and your training/execution command.
+  - **Solution:** Omit the `scan_layers` parameter from your training or execution command to allow MaxText to automatically resolve it from the checkpoint metadata, or ensure any explicitly specified `scan_layers` parameter matches the format of the loaded checkpoint.
 
 - **Error:** The converted checkpoint loads without errors but produces nonsensical output.
 
--- a/docs/reference/core_concepts/checkpoints.md
+++ b/docs/reference/core_concepts/checkpoints.md
@@ -74,7 +74,7 @@ In MaxText, the **`scan_layers`** configuration parameter is used to control thi
 - `scan_layers=false` tells MaxText to keep layer parameters unstacked (often required for inference and certain model architectures).
 
 > [!IMPORTANT]
-> **PyTree Structure Compatibility:** Because JAX expects the loaded PyTree structure to exactly match the model's instantiated structure, the value of the `scan_layers` flag during execution (training, SFT, RL, DPO, or decoding) **must** match the format of the checkpoint being loaded. A mismatch will cause PyTree loading or shape/path mismatch errors (which MaxText will intercept to raise a descriptive `ValueError` pointing to the scan_layers setting).
+> **Automatic scan_layers Resolution:** MaxText automatically loads `scan_layers` from the checkpoint's saved metadata when resuming (via `load_parameters_path`) if you do not explicitly specify `scan_layers` on the command-line. If you explicitly specify a value for `scan_layers` that conflicts with the checkpoint format, MaxText will raise a descriptive `ValueError` mismatch error to prevent JAX PyTree structure or shape mismatch errors during loading.
 
 ### Takeaways
 
--- a/docs/tutorials/posttraining/dpo.md
+++ b/docs/tutorials/posttraining/dpo.md
@@ -104,11 +104,11 @@ export MAXTEXT_CKPT_PATH=<CKPT_PATH> # e.g., gs://my-bucket/my-model-checkpoint/
 ```
 
 > [!IMPORTANT]
-> **Matching the `scan_layers` Parameter:**
-> The `scan_layers` setting during your fine-tuning run **must match** the setting used when creating the checkpoint at `MAXTEXT_CKPT_PATH`.
+> **Automatic `scan_layers` Resolution:**
+> MaxText automatically loads `scan_layers` from the checkpoint's saved metadata when resuming (via `load_parameters_path`) if you do not explicitly specify it on the command-line.
 >
-> - If the checkpoint was converted or saved with `scan_layers=False` (which is common for Hugging Face conversions and inference-ready models), you **must also provide `scan_layers=False` in the MaxText command.**
-> - If `scan_layers` does not match, MaxText will raise a `ValueError`.
+> - You do not need to manually supply `scan_layers=False` (or `scan_layers=True`) when loading checkpoints; MaxText will configure this automatically.
+> - If you do explicitly provide a `scan_layers` argument, it must match the checkpoint's saved setting or a `ValueError` mismatch error will be raised.
 >   See the [Checkpoints concept guide](../../reference/core_concepts/checkpoints.md) for more details.
 
 ## Running DPO Training
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -365,7 +365,8 @@ engram: 'remat'
 
 optimizer_memory_host_offload: false
 parameter_memory_host_offload: false
-scan_layers: true # We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations.
+# Whether to use jax.lax.scan over layers (stacked/unstacked checkpoint). We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations. When resuming from a checkpoint, this flag is auto-determined from metadata.
+scan_layers: true
 param_scan_axis: 1
 
 # The attention parameter dictates the specific algorithm/methodology used to compute the attention scores
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -972,7 +972,13 @@ class HardwareAndMesh(BaseModel):
   )
   shard_mode: ShardMode = Field("auto", description="can be either auto or explicit")
   inhomogeneous_layer_cycle_interval: int = Field(1, description="The interval of repeated inhomogeneous layer patterns.")
-  scan_layers: bool = Field(True, description="Whether to use jax.lax.scan over layers.")
+  scan_layers: bool = Field(
+      True,
+      description=(
+          "Whether to use jax.lax.scan over layers (stacked/unstacked checkpoint). "
+          "When resuming from a checkpoint, this flag is auto-determined from metadata."
+      ),
+  )
   param_scan_axis: int = Field(1, description="Axis to scan over for parameters.")
   context_parallel_load_balance: bool = Field(True, description="Whether to use load balancing for context parallelism.")
   context_parallel_strategy: str = Field(
--- a/src/maxtext/utils/model_creation_utils.py
+++ b/src/maxtext/utils/model_creation_utils.py
@@ -790,6 +790,44 @@ def create_models_and_meshes(trainer_config, sampler_config, trainer_devices, sa
   return reference_model, reference_mesh, actor_model, actor_mesh, rollout_mesh
 
 
+def verify_and_sync_scan_layers(config):
+  """Verify and sync scan_layers based on checkpoint metadata."""
+  if not config.load_parameters_path:
+    return config
+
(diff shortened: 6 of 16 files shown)


src/maxtext/utils/model_creation_utils.py
```diff
@@ -790,6 +790,49 @@ def create_models_and_meshes(trainer_config, sampler_config, trainer_devices, sa
   return reference_model, reference_mesh, actor_model, actor_mesh, rollout_mesh
 
 
+def verify_and_sync_scan_layers(config):
+  """Verify and sync scan_layers based on checkpoint metadata."""
```
shralex, line 794: There seem to be some repeated code and checks here, would something like this work where we extract the metadata first


def verify_and_sync_scan_layers(config):
  """Verify and sync scan_layers based on checkpoint metadata."""
  if not config.load_parameters_path:
    return config

  custom_metadata = checkpointing.load_checkpoint_metadata(config.load_parameters_path)
  saved_scan_layers = custom_metadata.get("scan_layers")
  if not isinstance(saved_scan_layers, bool):
    return config

  # Extract the Pydantic config (or use direct config if already a Pydantic model)
  pydantic_config = getattr(config, "_pydantic_config", config)
  model_fields_set = getattr(pydantic_config, "model_fields_set", None)

  # If model metadata tracking isn't supported, fall back to matching check (True)
  is_explicit = "scan_layers" in model_fields_set if model_fields_set is not None else True

  if is_explicit:
    if saved_scan_layers != config.scan_layers:
      raise ValueError(
          f"Configuration mismatch: Your run specifies scan_layers={config.scan_layers}, "
          f"but the checkpoint was saved with scan_layers={saved_scan_layers}."
      )
  else:
    new_pydantic_config = pydantic_config.model_copy(update={"scan_layers": saved_scan_layers})
    # Wrap back in HyperParameters if the original config was wrapped
    if getattr(config, "_pydantic_config", None) is not None:
      config = pyconfig.HyperParameters(new_pydantic_config)
    else:
      config = new_pydantic_config

  return config

shralex (commented):
Thanks Jacky!

Can you add some comments to wherever scan_layers is described (types.py, base.yml) describing this behavior when resuming from a checkpoint. 

Also, can we change docs in appropriate places to explain this new behavior. I remember I added a bunch of warnings related to scan_layers previously to the docs, these might need to be updated since you no longer need to specify scan_layers when resuming from a checkpoint.

src/maxtext/utils/model_creation_utils.py
RexBearIU: Thank you so much for the feedback! I've refactored the function using your elegant fallback structure—it is indeed much cleaner and completely avoids the repetitive code.

I've also added comments explaining the automatic synchronization behavior to both `types.py` and `base.yml`, and updated the outdated manual `scan_layers` warnings/guides throughout our documentation (including `checkpoints.md`, `convert_checkpoint.md`, `run_maxtext_localhost.md`, and all post-training tutorials).

docs/guides/checkpointing_solutions/convert_checkpoint.md
```diff
@@ -78,7 +78,7 @@ You can find your converted checkpoint files under `${BASE_OUTPUT_DIRECTORY}/0/i
 ### Key Parameters
 
 - `model_name`: The specific model identifier. It must match a supported entry in the MaxText [globals.py](https://github.com/AI-Hypercomputer/maxtext/blob/16b684840db9b96b19e24e84ac49f06af7204ae3/src/maxtext/utils/globals.py#L46C1-L46C7).
-- `scan_layers`: Controls whether the output uses a scanned (`scan_layers=true`) or unscanned (`scan_layers=false`) checkpoint format. Refer [to the Checkpoints guide](checkpoints) for more information. **IMPORTANT:** This setting *must* match the `scan_layers` value used during model training or loading. A mismatch will cause PyTree loading errors (though MaxText will intercept these and raise a descriptive `ValueError` explaining the mismatch).
+- `scan_layers`: Controls whether the output uses a scanned (`scan_layers=true`) or unscanned (`scan_layers=false`) checkpoint format. Refer [to the Checkpoints guide](checkpoints) for more information. **Note:** When resuming or loading a checkpoint in MaxText, this setting will automatically synchronize from checkpoint metadata unless explicitly overridden, meaning you do not have to manually specify it during model execution.
```
shralex: The text sounds like a user can override what's in the checkpoint metadata. 

Can we say instead: 

**Note**: When resuming or loading a checkpoint in MaxText, this setting will automatically be loaded from the checkpoint metadata, meaning you do not have to manually specify it during model execution. If you do explicitly specify a value for scan_layers, it must match the checkpoint's saved configuration, or a ValueError mismatch error will be raised.

docs/guides/checkpointing_solutions/convert_checkpoint.md
```diff
@@ -298,9 +298,9 @@ Here is an example [PR to add support for gemma3 multi-modal model](https://gith
 
 - **Error:** `Type ShapeDtypeStruct is not a valid JAX type` or generic **PyTree structure/shape mismatches** (e.g., Orbax reporting `"X/Y paths matched"`, such as `143/145 paths`).
 
-  - **Cause: Configuration mismatch** (e.g., `scan_layers`) between the checkpoint conversion script (e.g., `to_maxtext.py` or `to_huggingface.py`) and the trainer/inference runner (e.g., `train.py`).
```
shralex, line 301: Lets not remove the examples. 

**Cause: Configuration mismatch** (e.g., `scan_layers`) between the checkpoint conversion script (e.g., `to_maxtext.py` or `to_huggingface.py`) and the trainer/inference runner (e.g., `train.py`). Since MaxText automatically loads `scan_layers` from the checkpoint's saved metadata, you should only encounter this error if you explicitly set a mismatching value on the command line).

Actually, should we also auto-detect the setting in to_huggingface ?

docs/guides/checkpointing_solutions/convert_checkpoint.md
```diff
@@ -298,9 +298,9 @@ Here is an example [PR to add support for gemma3 multi-modal model](https://gith
...
+  - **Cause: Configuration mismatch** (e.g., `scan_layers`) between the checkpoint conversion script and explicit user configurations. (Note: MaxText automatically synchronizes `scan_layers` from the checkpoint's saved metadata when resuming, so you only encounter this error if you explicitly set a mismatching value on the command line).
 
-  - **Solution:** Ensure the `scan_layers` flag is set to the exact same value (`True` or `False`) in both the conversion command and your training/execution command.
+  - **Solution:** Omit the `scan_layers` parameter from your training or execution command to allow MaxText to automatically resolve it from the checkpoint metadata, or ensure any explicitly specified `scan_layers` parameter matches the format of the loaded checkpoint.
```
shralex, line 303: **Solution:** Omit the `scan_layers` parameter from your checkpoint conversion command to allow MaxText to automatically resolve it from the checkpoint metadata, or ensure any explicitly specified `scan_layers` parameter matches the format of the loaded checkpoint.

docs/reference/core_concepts/checkpoints.md
```diff
@@ -74,7 +74,7 @@ In MaxText, the **`scan_layers`** configuration parameter is used to control thi
 - `scan_layers=false` tells MaxText to keep layer parameters unstacked (often required for inference and certain model architectures).
 
 > [!IMPORTANT]
-> **PyTree Structure Compatibility:** Because JAX expects the loaded PyTree structure to exactly match the model's instantiated structure, the value of the `scan_layers` flag during execution (training, SFT, RL, DPO, or decoding) **must** match the format of the checkpoint being loaded. A mismatch will cause PyTree loading or shape/path mismatch errors (which MaxText will intercept to raise a descriptive `ValueError` pointing to the scan_layers setting).
+> **Automatic scan_layers Resolution:** MaxText automatically detects and synchronizes `scan_layers` from the checkpoint's saved metadata when resuming (via `load_parameters_path`) if you do not explicitly specify `scan_layers` on the command-line. If you explicitly specify a value for `scan_layers` that conflicts with the checkpoint format, MaxText will raise a descriptive `ValueError` mismatch error to prevent JAX PyTree structure or shape mismatch errors during loading.
```
shralex: Suggested:

**Automatic scan_layers Resolution:** MaxText automatically loads `scan_layers` from the checkpoint's metadata when resuming (via `load_parameters_path`) so it isn't necessary to specify `scan_layers` on resume. If provided, `scan_layers` must match the checkpoint metadata, otherwise a `ValueError` error is raised.

docs/tutorials/posttraining/dpo.md
```diff
@@ -104,11 +104,11 @@ export MAXTEXT_CKPT_PATH=<CKPT_PATH> # e.g., gs://my-bucket/my-model-checkpoint/
 ```
 
 > [!IMPORTANT]
-> **Matching the `scan_layers` Parameter:**
-> The `scan_layers` setting during your fine-tuning run **must match** the setting used when creating the checkpoint at `MAXTEXT_CKPT_PATH`.
+> **Automatic `scan_layers` Resolution:**
+> MaxText automatically detects and synchronizes `scan_layers` from the checkpoint's saved metadata when resuming (via `load_parameters_path`) if you do not explicitly specify it on the command-line.
```
shralex: I would say "loads" instead of "detects and synchronizes" here and everywhere

src/maxtext/configs/base.yml
```diff
@@ -365,7 +365,7 @@ engram: 'remat'
 
 optimizer_memory_host_offload: false
 parameter_memory_host_offload: false
-scan_layers: true # We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations.
+scan_layers: true # We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations. Automatically syncs with checkpoint when resuming unless overridden.
```
shralex: I propose:   

Whether to use jax.lax.scan over layers (stacked/unstacked checkpoint). We recommend setting this to false when using pipeline parallelism, instead scanning the PP iterations. When resuming from a checkpoint, this flag is auto-determined from metadata.

src/maxtext/configs/types.py
```diff
@@ -972,7 +972,13 @@ class HardwareAndMesh(BaseModel):
   )
   shard_mode: ShardMode = Field("auto", description="can be either auto or explicit")
   inhomogeneous_layer_cycle_interval: int = Field(1, description="The interval of repeated inhomogeneous layer patterns.")
-  scan_layers: bool = Field(True, description="Whether to use jax.lax.scan over layers.")
+  scan_layers: bool = Field(
+      True,
+      description=(
+          "Whether to use jax.lax.scan over layers. Automatically syncs"
+          " with the checkpoint when resuming unless overridden."
```
shralex: Whether to use jax.lax.scan over layers (stacked/unstacked checkpoint). When resuming from a checkpoint, this flag is auto-determined from metadata.

shralex (changes requested):
Thanks for the changes. I made a few suggestions to update comments. The main missing thing is updates to conversion scripts: to_huggingface should auto-detect the setting as well from the provided checkpoint (and error if its setting conflicts with the checkpoint:

Consider calling the new helper function in to_huggingface.py's main function right after initialising the config:

# Initialize maxtext config
config = pyconfig.initialize_pydantic(argv)

# Auto-resolve scan_layers from checkpoint metadata if not explicitly provided
config = verify_and_sync_scan_layers(config)

to_maxtext should write the scan_layers metadata into the checkpoint it creates, but in the current implementation, it does not actually do so. I believe we should provide a config when calling save_weights_to_checkpoint,

src/maxtext/utils/model_creation_utils.py
```diff
@@ -790,6 +790,40 @@ def create_models_and_meshes(trainer_config, sampler_config, trainer_devices, sa
...
+  if not config.load_parameters_path:
+    return config
+
+  custom_metadata = checkpointing.load_checkpoint_metadata(config.load_parameters_path)
+  saved_scan_layers = custom_metadata.get("scan_layers")
+  if not isinstance(saved_scan_layers, bool):
+    return config
+
```
xibinliu, line 802: We should return early if the flags are the same:
```
if saved_scan_layers == config.scan_layers:
    return config
```

src/maxtext/utils/model_creation_utils.py
```diff
@@ -790,6 +790,40 @@ def create_models_and_meshes(trainer_config, sampler_config, trainer_devices, sa
...
+  # Extract the Pydantic config (or use direct config if already a Pydantic model)
+  pydantic_config = getattr(config, "_pydantic_config", config)
+  model_fields_set = getattr(pydantic_config, "model_fields_set", None)
+
+  # If model metadata tracking isn't supported, fall back to matching check (True)
+  is_explicit = "scan_layers" in model_fields_set if model_fields_set is not None else True
+
+  if is_explicit:
+    if saved_scan_layers != config.scan_layers:
```
xibinliu, line 814: No need to check if the check in the above comment is implemented.

src/maxtext/utils/model_creation_utils.py
```diff
@@ -790,6 +790,40 @@ def create_models_and_meshes(trainer_config, sampler_config, trainer_devices, sa
...
+      raise ValueError(
+          f"Configuration mismatch: Your run specifies scan_layers={config.scan_layers}, "
+          f"but the checkpoint was saved with scan_layers={saved_scan_layers}."
+      )
+  else:
+    new_pydantic_config = pydantic_config.model_copy(update={"scan_layers": saved_scan_layers})
```
xibinliu, line 821: We should print log message and let user know the scan_layers has been changed.

xibinliu (commented):
I think we should also cover the following metadata handling when doing checkpoints conversion:

1. to_maxtext.py
Generate metadata data with scanned_layers and lora parameters and attach the metadata to the MaxText checkpoints.

2. to_hugging_face.py:
Besides the lora parameters, we should also read scan_layers from the MaxText metadata and use it for conversion.

docs/reference/core_concepts/checkpoints.md
RexBearIU: I have gone through all 8 of the documentation manuals, tutorials, and configurations, and updated the wording to use 'loads' or 'auto-determined' everywhere instead of 'detects and synchronizes' or 'automatically syncs' to match your feedback. Thank you!

docs/tutorials/posttraining/dpo.md
RexBearIU: This section has also been updated to use 'loads' and match the rest of the documentation. Thanks again!

src/maxtext/configs/base.yml
RexBearIU: Excellent phrasing. I have adopted your proposed description exactly for the `scan_layers` comment in `base.yml`. Thank you!

src/maxtext/configs/types.py
RexBearIU: Perfect. This is now fully updated in `types.py` description as well. Thank you!

src/maxtext/utils/model_creation_utils.py
RexBearIU: Good suggestion! I've added the early return check at the very beginning of the verification logic, returning the config immediately when `saved_scan_layers == config.scan_layers`.

src/maxtext/utils/model_creation_utils.py
RexBearIU: Correct, with the early return in place, that duplicate check was removed. Thanks!

src/maxtext/utils/model_creation_utils.py
RexBearIU: That makes total sense. I have added logging using `max_logging.log` so the user is informed explicitly whenever the `scan_layers` value is automatically loaded/changed from the checkpoint metadata.

shralex (approved):

src/maxtext/utils/model_creation_utils.py
xibinliu: Remove this line: 

`if saved_scan_layers != config.scan_layers:`

xibinliu (approved):

PR #4348, opened 2026-07-03
Fuse the Gated Delta Net inter-chunk scan into a Pallas TPU kernel

### LINKED ISSUE #4347: Qwen3-Next: sequential GDN paths (lax.scan recurrence, solve_triangular) bound TPU step time
### Problem

In the Qwen3-Next Gated Delta Net (`jax_chunk_gated_delta_rule`), the two sequential computations dominate TPU step time:

- the inter-chunk recurrence runs as `lax.scan`, so the recurrent state round-trips HBM on every chunk step, and
- the UT-transform inverse `A = (I+S)^-1` uses `solve_triangular`, which substitutes row by row and barely uses the MXU.

Neither overlaps with other work, so they bound the step directly.

### Measurement

v5e-256, Qwen3-Next dense-8B config, seq 4096, bs 1, remat full (details in #4348):

| build | step time | tokens/s/device |
|---|---|---|
| main | 3.373s | 1,214 |
| #4348 (Pallas kernels) | **2.571s** | **1,593 (+31%)** |

Loss trajectories match to bf16 rounding level.

bzantium:
# Description

Fuses the sequential parts of the Qwen3-Next Gated Delta Net into Pallas TPU kernels: the inter-chunk recurrence, the UT-transform triangular inversion, and head/tile batching to keep the MXU pipelined through both. Together: **3.373s → 2.571s/step (+31% throughput)** at the reference config.

FIXES: #4347

## Changes

- **Inter-chunk scan kernel** — the recurrence runs as one kernel per (batch, heads) with the state in VMEM, instead of a `lax.scan` whose state round-trips HBM every chunk. Backward walks the chunks in reverse (FLA-style checkpointing); initial/final state are first-class, so the `custom_vjp` covers the full state chain. The chunk-parallel WY math stays in XLA on purpose — TPU grid cells run sequentially per core, and fusing batched matmuls into the walk measured slower.
- **Triangular inversion kernel** — replaces `solve_triangular` (row-sequential, no MXU use) for `A = (I+S)^-1`. Blockwise ladder: with `X = (I+S_b)^-1` at block size `b`, the size-`2b` inverse is exactly `X - X @ S_l @ X`, so doubling from `b=1` needs `2·log2(C)` full-width matmuls. It never forms powers of `S` (those overflow f32 once `|S| > 1`), so intermediates stay at the true inverse's scale — stable wherever the answer is representable. Products use manual bf16x3 operand splitting (Mosaic's default f32 dot truncates to bf16, which breaks `(I+S)A = I` enough to destabilize the recurrence; `Precision.HIGH` is unsupported). Backward is analytic: `dS = -(A^T dA A^T)`, strict-lower masked.
- **Head/tile batching** — the walks are sequential, but heads (scan) and chunk tiles (inversion) are independent. Batching them per grid cell (scan fwd 8 / bwd 4 heads, inversion 8 tiles — bounded by the 16MB scoped-VMEM limit) turns each dot into a batched `dot_general` the MXU can pipeline instead of paying fill/drain per short matmul.

Peak memory is unchanged — the eliminated HBM traffic was latency-hidden, which is also why the gains come specifically from the sequential paths that cannot overlap.

## Performance

v5e-256, Qwen3-Next dense-8B (emb 4096, 36 layers, GDN 16/32 heads, head dim 128), seq 4096, bs 1, bf16, FSDP, `gdn_chunk_size=128`, remat full, synthetic data:

| build | step time | tokens/s/device |
|---|---|---|
| main | 3.373s | 1,214 |
| + scan kernel | 3.119s | 1,313 |
| + head batching | 3.076s | 1,331 |
| + inversion ladder | **2.571s** | **1,593** |

Loss trajectories match to bf16 rounding level at every step.

# Tests

`python3 -m pytest tests/unit/gated_delta_network_kernel_test.py` — 9 tests on CPU (interpret mode): scan kernel vs a `lax.scan` reference (forward, final state, all gradients, non-zero initial state and final-state cotangent, bf16 dtypes); inversion vs `solve_triangular` (random, all-ones, unnormalized gaussian; forward + gradients). End-to-end training verified on v5e-256 (table above). pyink / pylint clean.

# Checklist

### DIFF
--- a/src/maxtext/kernels/attention/gated_delta_network.py
+++ b/src/maxtext/kernels/attention/gated_delta_network.py
@@ -1,0 +1,518 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Pallas TPU kernels for the Gated Delta Net chunked delta rule.
+
+The chunked gated delta rule splits into a chunk-parallel precomputation
+(the WY factors, handled well by XLA) and a strictly sequential inter-chunk
+recurrence. The recurrence is the throughput bottleneck when expressed as
+`lax.scan`: every step is a separate XLA loop iteration whose recurrent
+state round-trips through HBM between many small fusions.
+
+This module fuses the recurrence into a single Pallas kernel. The grid is
+(batch, head-tiles, num_chunks) with the chunk dimension marked "arbitrary"
+so each program walks its chunks sequentially while the recurrent state
+lives in a VMEM scratch buffer for the whole walk. The walk is sequential
+per head, but heads are independent, so each grid cell batches a tile of
+heads and every matmul is a leading-axis dot_general: Mosaic pipelines the
+MXU across heads instead of stalling on the chain of chunk dependencies.
+
+Forward additionally materializes the per-chunk input states (FLA-style
+chunkwise checkpointing) so the backward kernel can walk the chunks in
+reverse with the same fusion structure, carrying the state cotangent in
+VMEM. The initial recurrent state is a first-class input and the final
+state a first-class output, so the kernel is differentiable end-to-end
+including through the recurrent-state chain.
+
+Matmul operands are cast to `compute_dtype` (bf16 by default) with float32
+accumulation via `preferred_element_type` — the MXU fast path. The
+recurrent state and the log-decay math stay float32 throughout.
+
+Shapes (per chunk c, head-batched by the grid):
+  w, u, q, k: [C, D]   g (cumulative log-decay within chunk): [C]
+  recurrent state h: [D_k, D_v]
+
+The module also provides `invert_unit_lower`, a Pallas blockwise inversion
+of the unit-lower-triangular UT-transform system that replaces the
+row-sequential TPU triangular solve in the chunk-parallel stage.
+"""
+
+# Pallas kernel bodies receive their operand refs positionally in the order
+# declared by pallas_call, so the many-positional-arguments rule cannot apply.
+# pylint: disable=too-many-positional-arguments
+
+import functools
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+
+
+def _bdot(a, b, contract, compute_dtype):
+  """Batched MXU dot over the leading (heads) axis: contract a-dim vs b-dim."""
+  return jax.lax.dot_general(
+      a.astype(compute_dtype),
+      b.astype(compute_dtype),
+      dimension_numbers=(((contract[0],), (contract[1],)), ((0,), (0,))),
+      preferred_element_type=jnp.float32,
+  )
+
+
+def _tril_mask(chunk_size: int, include_diag: bool = True) -> jax.Array:
+  rows = jax.lax.broadcasted_iota(jnp.int32, (chunk_size, chunk_size), 0)
+  cols = jax.lax.broadcasted_iota(jnp.int32, (chunk_size, chunk_size), 1)
+  return rows >= cols if include_diag else rows > cols
+
+
+def _gdn_scan_fwd_kernel(
+    w_ref,
+    u_ref,
+    q_ref,
+    k_ref,
+    g_ref,
+    h0_ref,
+    o_ref,
+    h_saved_ref,
+    h_final_ref,
+    h_scratch,
+    *,
+    chunk_size: int,
+    num_chunks: int,
+    compute_dtype: jnp.dtype,
+):
+  """Forward inter-chunk recurrence for one (batch, head-tile) program."""
+  n = pl.program_id(2)
+
+  @pl.when(n == 0)
+  def _init():
+    h_scratch[...] = h0_ref[0].astype(jnp.float32)
+
+  # [TH, d_k, d_v] — the walk is sequential per head, but the TH heads in
+  # this cell are independent, so every dot below is batched over the
+  # leading axis and Mosaic can pipeline the MXU across heads.
+  h = h_scratch[...]
+  # Save the chunk-input state for the backward pass.
+  h_saved_ref[0, 0] = h
+
+  # MXU dots: operands in the compute dtype (bf16), float32 accumulation
+  # via preferred_element_type. State and decay math stay f32.
+  def mxu_dot(a, b, contract=(2, 1)):
+    return _bdot(a, b, contract, compute_dtype)
+
+  w = w_ref[0, 0]
+  u = u_ref[0, 0].astype(jnp.float32)
+  q = q_ref[0, 0]
+  k = k_ref[0, 0]
+  g = g_ref[0, 0, :, :, 0].astype(jnp.float32)  # [TH, C]
+
+  exp_g = jnp.exp(g)
+  q_g = q.astype(jnp.float32) * exp_g[..., None]
+  # Merged matmul: [q_g ; w] @ h doubles the LHS rows for the MXU.
+  both = mxu_dot(jnp.concatenate([q_g, w.astype(jnp.float32)], axis=1), h)
+  attn_inter, v_prime = both[:, :chunk_size], both[:, chunk_size:]
+  v_new = u - v_prime
+
+  p = mxu_dot(q, k, contract=(2, 2))
+  g_diff = g[:, :, None] - g[:, None, :]
+  mask = _tril_mask(chunk_size)
+  decay = jnp.where(mask, jnp.exp(jnp.where(mask, g_diff, 0.0)), 0.0)
+  s = p * decay
+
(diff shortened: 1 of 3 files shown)


WandLZhang:
Independent reproduction, on a dense GatedDeltaNet config: d_model 4096, 32 layers, one full-attention layer in four, `vocab_size=512`, bf16, FSDP, `gdn_chunk_size=128`, sequence 8192, batch 1 per device, synthetic data. Both columns are the same checkout, with the control built by removing these two commits.

| | main | this PR | gain |
|---|---|---|---|
| v6e-8 | 3,216 tok/s/chip, 137.1 TFLOP/s | 3,838, 163.6 | **+19.4%** |
| v5p, 4 chips, batch 1 | 3,121 | 3,883 | +24.4% |
| v5p, 4 chips, batch 2 | 3,711 | 4,470 | +20.4% |

Loss matches the control to three decimals in every pair. It rebases onto current main with one conflict, an import block in `models/qwen3.py` where both sides are wanted.

Two things I hit that aren't in the description.

**Peak memory falls by about a third.** The description says peak memory is unchanged. On this config sequence 16384 runs on a v6e-8 with the kernel and OOMs without it. Where both OOM, the reported requirement falls 31% at 32768 (86.39G to 59.59G) and 35% at 65536 (259.68G to 169.83G). Worth claiming, since it buys a sequence length main can't reach.

**The `if use_pallas:` branch returns from `jax_chunk_gated_delta_rule` above everything below it,** so any work later in that function is skipped. We carry a patch that makes the inter-chunk recurrence context-parallel by composing each device's affine map at that point, and with the kernel on it never runs: every device does an isolated recurrence from a zero state. Training loss doesn't show it, since 60 steps at `ctx=2` match to three decimals. Comparing the output tensors under a two-shard context mesh does. The shard that should inherit state is off by 7.9e-1 max, against 5.7e-2 for the shard that has nothing upstream to inherit, and the final recurrent state by 3.78. This only reaches callers who run context parallelism on this path, so it may be out of scope, but it's worth a line in the kernel docstring.

Would you consider reopening? The gain is large and it's a rebase rather than new work. Happy to carry that.

WandLZhang:
cc @mmcsa

bzantium:
@WandLZhang Reopened. Thanks for taking the time to reproduce this, and for offering to carry the rebase.

On the numbers, our original run was v5e-256 at around +8%, so it's good to see +19-24% on v6e/v5p. Nice that it holds up on the newer chips.

On peak memory, you're right that "unchanged" was too conservative. The recurrent state stays in VMEM across the inter-chunk walk instead of getting written back to HBM every scan step, so the drop makes sense. A 31-35% reduction that lets you fit a sequence length main OOMs on is worth calling out, I'll update the description.

On the context-parallel issue, good catch. The use_pallas branch returns before the code below it, so the fused path only threads a single initial_state/final_h and never composes state across sequence-sharded shards. Each shard just starts from its own h0. It doesn't surface in this PR's configs because main keeps the sequence replicated (batch/head sharding only), so there's no upstream state for a shard to inherit, which is why ctx=1 loss matches to three decimals. It only shows up once you shard the sequence, like your ctx=2 comparison. I added a note on the branch (dd5f9b038) so anyone running context parallelism knows to compose per-shard state around the call. If you want to upstream your affine-composition patch on top of this, happy to coordinate.

Branch is rebased and pushed. I'll take the import conflict in models/qwen3.py.

WandLZhang:
Thanks for reopening, and for the NOTE.

Re-verified against your rebase. Applying our patch to the new base `b18ceac97` reproduces the branch exactly in both files, and the kernel blob is unchanged, so the +19% to +24% above still stands as measured. We carry the two commits plus your note locally until this lands.

On the context-parallel side, the note covers it for callers. The composition itself is in #4932. The inter-chunk step is

```
h_new = exp(g_last) * h + k_g^T (u - w h) = (exp(g_last) I - k_g^T w) h + k_g^T u = A h + B
```

`A` and `B` don't depend on `h`, so affine maps compose associatively and each device can fold its local chunks into one `(A, B)` pair, all-gather the D pairs, take the exclusive prefix for its incoming state, and replay locally. `A` and `B` are 128x128 for every published Qwen 3.5 size, so the gathered volume stays small. It needs `jax.checkpoint` on the local scan body; without it the autodiff residuals reach 103 GB at sequence 262,144.

I can send that as a PR on top once yours merges. Let me know.

bzantium:
@WandLZhang thanks for re-verifying against the rebase, glad the numbers held.

The affine-composition approach makes sense to me. Folding each device's local chunks into a single `(A, B)` pair and taking the exclusive prefix over the gathered pairs is a clean way to do it, and with `A, B` at 128x128 the all-gather stays cheap. Good catch on needing `jax.checkpoint` around the local scan body too, that residual blowup at long sequence would be easy to miss.

Please do send the context-parallel PR on top once this lands, that's a natural follow-up and I'd be happy to review it. I'll ping the maintainers to take another look here so it can merge.

WandLZhang:
Thanks. I'll send the context-parallel PR once this lands.

One design question:

`gdn_inter_chunk_scan` takes `h0` and returns `(o, h_final)`, so it evaluates `A·h0 + B` for a given `h0`. The exclusive prefix needs `A` on its own: each device has to publish its affine map before it knows its inbound state, otherwise the devices serialise.

Two options:

Return it. `_fwd_pallas` already forms the pieces, `gamma = exp(g_last)` and the `k_gᵀw` term. An optional third output of `[B, H, D_k, D_k]`, off by default so nothing changes for existing callers.

Or form `A_loc` in XLA next to the kernel. It's a product of `(exp(g_last)·I − k_gᵀw)` over the device's chunks, 128x128 each. At 1M over 64 devices with `gdn_chunk_size=64` that's about 244 small matmuls per device, negligible against the attention in the same layer. The sequential walk stays in your kernel.

I lean to the second: your signature stays untouched and the composition stays where the collective already is. If returning `A` is close to free, the first is tidier.

WandLZhang:
Rethinking the ordering.

At 1M the two can't both be live. The composition needs each device's `A`, and the fused path returns before it, so our patch sets `use_pallas` false whenever a context axis is set. The 1M run I mentioned used the XLA scan, not your kernel.

So they're disjoint rather than stacked. I can send the context-parallel PR against `main` and you rebase #4348 on top, which is the `models/qwen3.py` conflict you already absorbed once. Let me know.

bzantium:
@WandLZhang makes sense, disjoint is the right call. Gating `use_pallas` off whenever a context axis is set keeps the fused path for the common case and hands the sequence-sharded case to the XLA scan where `A` is available for the prefix. That also lines up with the note on the branch, the caller is expected to handle composition, and falling back to the scan is exactly that.

I checked the fallback is loss-preserving before agreeing: the kernel path and the XLA scan path match to about 6e-6 relative on the output with no initial state, and 7e-6 output / 2e-5 final-state with an `h0` passed in (float32, interpret mode). So switching to the scan under a context axis changes speed, not math, and state threading through `h0` behaves the same on both paths, which is what the replay relies on.

On ordering, since they're disjoint either way works and it's just the one `models/qwen3.py` import block. #4348 is already rebased, green, and mergeable, so landing it first and rebasing your PR on top is probably the least work overall, but if the maintainers would rather take the context-parallel PR first I'm happy to rebase #4348 onto it instead. Either way, ping me when yours is up.

WandLZhang:
@bzantium it's up: #4968, against `main` as we discussed. Small — `models/gdn_cp.py` plus the two pspec fixes, 121 lines. No dependency on #4348, so rebase whenever suits.

WandLZhang:
Heads up that #4968 adds context parallelism to the GatedDeltaNet scan which conflicts with this PR

gdn_inter_chunk_scan replaces the same lax.scan(scan_body, h_init, xs) that #4968's context-parallel branch replays. The fused path returns before that code. So whichever lands second, the other stops working and there would be no error thrown.

bzantium:
@WandLZhang thanks for the heads up, and you're right about the silent breakage. Whichever lands second, the fused path returns above the `cp_axis` branch, so context parallelism would just be ignored with no error.

I reproduced #4968's composition on a 4-device CPU mesh to be sure of it before touching anything: folding the local chunks into `(A_loc, B_loc)`, the Hillis-Steele prefix over the context axis, and replaying each shard reproduces the monolithic scan's final state to about 1.5e-6 relative. The `A_i = exp(g_last) I - k_g^T w_i`, `B_i = k_g^T u_i` decomposition and the `jax.checkpoint` on the local scan both look right.

To make the two disjoint at the code level rather than by landing order, I pushed a gate to this PR (6ec471b): `jax_chunk_gated_delta_rule` takes a `cp_axis` argument, and the fused kernel now runs only when `cp_axis is None`. When a context axis is set it falls through to the scan path, which is exactly where #4968 adds the affine composition. `cp_axis` defaults to `None` so nothing changes for existing callers, and the kernel stays on for the common case. I left the Pallas triangular inversion ungated since it is independent of the inter-chunk recurrence, so the scan path keeps that speedup too.

With this in, the only real overlap left is the one `models/qwen3.py` region. Whoever rebases second gets a clean three-way merge on the `if use_pallas and cp_axis is None:` block plus your `else` branch, instead of a silent runtime break. Happy to rebase onto #4968 or have you rebase onto this, whichever the maintainers prefer.

WandLZhang:
Thanks, gating it is the right fix, cleaner than depending on landing order

your 1.5e-6 lines up with what I see. also worth knowing the backward composes: there's no custom_vjp in gdn_cp, and the CP gradient matches the sequential recurrence to 2.3e-7 on an 8-device CPU mesh

on rebase order, #4968 is further along, two approvals and NuojCheng's threads all resolved, so probably less churn if you rebase onto it. only overlap left is that qwen3.py block. fine either way if the maintainers prefer otherwise

PR #4372, opened 2026-07-07
[NNX] Emergency checkpoint: build the manager with a checkpoint-layout abstract

ecnal-cienet:
## Description

This PR aims to fix issue [b/530068041](https://buganizer.corp.google.com/issues/530068041)

The `maxtext_emc_resume_from_gcs` e2e job fails when a `pure_nnx` run resumes from its GCS checkpoint:

```
ValueError: User-provided restore item and on-disk value metadata tree structures do not match
  model:     - Source: {'decoder': ...}   - Target: MISSING
  opt_state: - Source: MISSING            - Target: [{'count': ...}, ...]
```

A `pure_nnx` run saves in the on-disk checkpoint layout — Linen `params`/`opt_state`/`step` plus an `nnx_aux` subtree, via `to_checkpoint_dict` — so pure_nnx and Linen checkpoints stay interchangeable. Both the regular and the emergency manager save that way.

The difference is on the restore side. The regular manager restores against the item it is given at restore time. The emergency manager instead **bakes in the abstract it is constructed with** and restores against that, ignoring the restore-time item. `create_orbax_emergency_checkpoint_manager` was handing it `get_abstract_state`'s NNX-native tree (`model`/`optimizer`), so Orbax compared a checkpoint-layout tree on disk against an NNX target and raised. This is also why converting the restore-time item alone did not help: the emergency manager never looks at that item.

Fix: convert the abstract with `to_checkpoint_dict` in `create_orbax_emergency_checkpoint_manager` when it is an `nnx.State`, so the blueprint the manager holds matches what `maybe_save_checkpoint` writes. `split_for_checkpoint` already requires that save and restore partition the state identically; the manager's baked-in abstract is a third copy of that partition, and it now goes through the same function as the other two. The restore path reshapes back to NNX, so the state we train on is still NNX, and the on-disk format is unchanged. Six lines, in the constructor. A Linen state is not an `nnx.State`, so it is passed through untouched.

## Tests

`tests/unit/checkpointing_emergency_nnx_test.py` (5) — the emergency-manager counterpart of `checkpointing_nnx_roundtrip_test.py`:

- the manager is handed the checkpoint layout (`params`/`opt_state`/`step`/`nnx_aux`), not the NNX `model`/`optimizer` roots. Orbax is mocked, so this one runs in normal CI and catches a regression here without an emergency-checkpoint setup.
- a Linen state is passed through unchanged.
- a full round trip through a **real** `EmergencyCheckpointManager` is leaf-exact: weights, optimizer moments, dropout rng, batch stats.
- `nnx_aux` is genuinely restored rather than reset to a fresh init — checked by value (rng count, batch-norm mean, optimizer step), since a reset keeps the tree structure intact and would slip past the structural check.
- caches stay ephemeral: excluded from the baked-in abstract, so they keep their init value on restore while the rng beside them still comes back from `nnx_aux`.

Reverting the six lines fails four of the five (the fifth is the Linen pass-through case, which correctly still passes).

### End-to-end on a v6e-8 (jax 0.10.0, orbax-checkpoint 0.12.1)
1. Save checkpoint for later tests: (passed, [log](https://paste.googleplex.com/6410789710856192))
2. EMC resume WITHOUT fix (failed expected, [log](https://paste.googleplex.com/5343369239855104))
3. EMC resume WITH fix (passed, [log](https://paste.googleplex.com/6057983078432768))

# Checklist

### DIFF
--- a/tests/unit/checkpointing_emergency_nnx_test.py
+++ b/tests/unit/checkpointing_emergency_nnx_test.py
@@ -1,0 +1,293 @@
+# Copyright 2025-2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for restoring pure_nnx emergency checkpoints.
+
+pure_nnx saves in the on-disk checkpoint layout: Linen params/opt_state/step, plus an nnx_aux
+subtree for the rngs and batch stats Linen has no place for. The emergency manager, unlike the
+regular one, bakes in the abstract it is built with and restores against that, ignoring the
+restore-time item. So that abstract has to be in the on-disk layout already, or Orbax compares
+it against the checkpoint and raises.
+
+checkpointing_nnx_roundtrip_test covers the same ground for the regular manager. The emergency
+manager writes one state tree, and nnx_aux is part of it, so the same partition has to survive
+there too. These tests check that it does.
+"""
+
+import os
+import shutil
+import tempfile
+import unittest
+from types import SimpleNamespace
+from unittest import mock
+
+from etils import epath
+from flax import nnx
+import jax
+import jax.numpy as jnp
+from maxtext.common import checkpointing
+from maxtext.common import train_state_nnx
+import optax
+
+
+class _Model(nnx.Module):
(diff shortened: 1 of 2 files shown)


tests/unit/checkpointing_emergency_nnx_test.py
```diff
@@ -0,0 +1,275 @@
...
+  def test_full_state_round_trip_is_exact(self):
+    """Every leaf -- weights, optimizer, dropout rng, batch stats -- round-trips unchanged."""
+    manager = self._manager(_Model, "exact")
+    state = self._trained(_Model(nnx.Rngs(0)))
+    saved = nnx.state(state).to_pure_dict()
+    self._save(manager, state)
+    restored = self._restore(manager, _Model)
+
+    self.assertEqual(jax.tree_util.tree_structure(saved), jax.tree_util.tree_structure(restored))
+    matches = jax.tree_util.tree_map(_leaf_equal, saved, restored)
```
NuojCheng: might be better adding `is_leaf=lambda...`

xibinliu (approved):

NuojCheng (approved):

PR #4392, opened 2026-07-08
Fix RL LR schedule default

AntonyMei:
RL computes its real training length from num_batches, rl.num_iterations, train_fraction, and num_epoch, but rl.yml inherited base.yml steps=150001. Config normalization then converted learning_rate_schedule_steps=-1 into that inherited step count before the RL optimizer saw max_train_steps, causing warmup/decay to use a much longer horizon than the actual RL run.

Set both steps and learning_rate_schedule_steps to -1 in the RL base config so the optimizer falls back to the computed max_train_steps unless a recipe explicitly pins the schedule horizon. Also warn when a positive learning_rate_schedule_steps differs from the computed RL training length, since that may be intentional but is easy to inherit accidentally.

# Checklist

### DIFF
--- a/src/maxtext/configs/post_train/rl.yml
+++ b/src/maxtext/configs/post_train/rl.yml
@@ -111,6 +111,13 @@ debug:
 enable_tunix_perf_metrics: false
 
 # ====== Training ======
+# RL computes actual optimizer updates as:
+# num_batches * rl.num_iterations * train_fraction * num_epoch.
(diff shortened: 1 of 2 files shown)


A9isha:
@py4 had made [this change](https://github.com/AI-Hypercomputer/maxtext/pull/4225), do you still need further changes?

AntonyMei:
I think after that change, this change is not necessary. I will close this PR.

PR #4404, opened 2026-07-09
 fix: resolve vLLM and RL training gibberish outputs for Gemma 3/4 scanned layers

mesakhcienet:
# Description

  This PR fixes multiple issues causing gibberish outputs and initialization crashes when using scan_layers=True with Gemma 3 and 4 in vLLM workflows.

  Context and Details:
   - Why is this change being made: To enable successful RL training and generation with Gemma 3 and 4 by ensuring KV caches and scanned layer weights are correctly handled in the
     vLLM rollout engine.
   - The problems solved:
     1. vLLM Decoder Gibberish: Fixed missing `kv_caches` argument in `_apply_gemma3_scanned_blocks`, which caused loss of attention context.
     2. KV Cache Shape Mismatches: `MaxTextVllmRollout` ignored `rollout_vllm_additional_config`, falling back to base configs that conflicted with HuggingFace config head dimensions.
     4. Native NNX Keys & Scalar Bugs: The weight sync preprocessor failed to unstack nested scanned weights because it mismatched on native NNX root keys, and a bug in the `scan_length` calculation for scalar RNG states caused it to drop all decoder parameters, resulting in random noise generation.
   - Implementation details:
     - Explicitly passed `kv_caches` and attention_metadata in decoder blocks.
     - Updated the custom preprocessor in MaxTextVllmSampler to use robust string matching for NNX keys, safely ignore scalars when calculating `scan_length`, and correctly broadcast scalar values during unstacking.

# Tests
## inference vllm_decode
Run script : [here](https://paste.googleplex.com/5013364639137792) 
Run output : [log](https://paste.googleplex.com/5962439937097728)

# Checklist

### DIFF
--- a/src/maxtext/layers/decoders.py
+++ b/src/maxtext/layers/decoders.py
@@ -1374,18 +1374,19 @@ class Decoder(nn.Module):
         start_idx = scan_length * attention_pattern_length
         remainder_kv = tuple(kv_caches[start_idx : start_idx + num_remaining_layers])
 
-      y_and_kv = layer(
-          y,
+      remainder_args = (
           decoder_segment_ids,
           decoder_positions,
           deterministic,
           model_mode,
-          previous_chunk=previous_chunk,
-          slot=slot,
-          bidirectional_mask=bidirectional_mask,
-          kv_cache=remainder_kv,
-          attention_metadata=attention_metadata,
+          slot,
+          None,  # page_state
+          previous_chunk,
+          bidirectional_mask,
+          remainder_kv,
+          attention_metadata,
       )
+      y_and_kv = layer(y, *remainder_args)
 
       if isinstance(y_and_kv, tuple):
         y = y_and_kv[0]
--- a/src/maxtext/trainers/post_train/rl/train_rl.py
+++ b/src/maxtext/trainers/post_train/rl/train_rl.py
@@ -64,6 +64,9 @@ from flax import nnx
 from orbax import checkpoint as ocp
 from pprint import pprint
 from transformers import AutoTokenizer
+import maxtext.integration.vllm.maxtext_vllm_adapter as adapter
+
+adapter.register()
 import functools
 from tunix.rl import rl_cluster as rl_cluster_lib
 from tunix.rl.rollout import base_rollout
@@ -505,7 +508,7 @@ def create_rl_components(
               "enable_expert_parallel": sampler_config.enable_expert_parallel,
               "enable_prefix_caching": True,  # Enable prefix caching to speed up generation for long prompts
               # Ensures vLLM model initializes with correct dtype (not float32 default)
-              "dtype": trainer_config.weight_dtype,
+              "dtype": trainer_config.weight_dtype.value,
           },
           rollout_vllm_sampling_kwargs={
               "stop": trainer_config.stop_strings,
--- a/tests/unit/nnx_decoders_test.py
+++ b/tests/unit/nnx_decoders_test.py
@@ -23,6 +23,8 @@ Tests cover:
 
 import sys
 import unittest
+from unittest.mock import MagicMock, patch
+
 import pytest
 
 import jax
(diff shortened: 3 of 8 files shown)


src/maxtext/layers/attentions.py
```diff
@@ -1060,7 +1054,6 @@ def forward_serve_vllm(
         q_scale,
         k_scale,
         v_scale,
-        update_kv_cache=update_kv_cache,
```
hengtaoguo, line 1067: Could you remind me why removing this argument?

src/maxtext/layers/decoders.py
```diff
@@ -1374,18 +1374,19 @@ def _apply_gemma3_scanned_blocks(
...
-          kv_cache=remainder_kv,
-          attention_metadata=attention_metadata,
+          slot,
+          None,  # page_state
+          previous_chunk,
+          bidirectional_mask,
+          remainder_kv,
+          attention_metadata,
       )
+      y_and_kv = layer(y, *remainder_args)
```
hengtaoguo, line 1389: This means previously some args were not used, causing the output gibberish? Can we identify which args are the most critical cause?

src/maxtext/trainers/post_train/rl/train_rl.py
```diff
@@ -445,16 +448,23 @@ def create_rl_components(
...
   vllm_config_path = os.path.join(MAXTEXT_CONFIGS_DIR, "inference", "vllm.yml")
   argv_list = ["", str(vllm_config_path), "log_config=False"]
   vllm_config = pyconfig.initialize(argv_list)
 
-  rl_rollout_engine = (
-      functools.partial(MaxTextVllmRollout, maxtext_config=trainer_config)
-      if trainer_config.use_standalone_converter
-      else "vllm"
-  )
+  rl_rollout_engine = functools.partial(MaxTextVllmRollout, maxtext_config=trainer_config)
```
hengtaoguo: I recall when `rl_rollout_engine="vllm"`, there's another set of [standalone hook functions](src/maxtext/integration/tunix/weight_mapping) which perform the weight sync for dense models like qwen3-4b:

- Creates `TunixMaxTextAdapter`
- Expose `to_hf_mappings, to_hf_transpose_keys, to_hf_hook_fns` [here](https://github.com/AI-Hypercomputer/maxtext/blob/409f43b0f375f652c739e3e14e38c047f125e5f3/src/maxtext/integration/tunix/weight_mapping/__init__.py#L29).


This means we probably should not remove `else "vllm"` here, unless you've validated the run for other models such as qwen3-4b?

tests/unit/nnx_decoders_test.py
```diff
@@ -743,6 +745,111 @@ def _make_shared_embedding(self, cfg):
         rngs=self.rngs,
     )
 
+  def test_gemma_scan_layers_equivalence(self):
```
hengtaoguo, line 748: Thanks for adding these tests!

src/maxtext/layers/attentions.py
mesakhcienet: my bad, i used older version of vllm. I have reverted this changes

src/maxtext/layers/decoders.py
mesakhcienet: Yes, exactly. In the legacy Flax-Linen pipeline (`pure_nnx_decoder=False`), passing `**kwargs` to the layer caused critical arguments like `slot` and `bidirectional_mask` to be misrouted or silently dropped due to positional signature mismatches. Without the correct `slot` assignment and mask, the attention layers misidentified their roles (global vs local) and failed to mask tokens, corrupting the KV caches and producing gibberish. Passing them strictly as positional `remainder_args` ensures they are safely routed and fixes the legacy path.

However, please note that for our actual native NNX pipeline runs (`pure_nnx_decoder=True`), the `decoders.py` file is entirely bypassed in favor of `nnx_decoders.py` (which did not have this positional bug). The gibberish we experienced during our XPK RL evaluation was actually caused by a separate issue: vLLM expects flat layers, but Gemma 3/4 scanned blocks are nested arrays. We had to restore the `unroll_gemma_scanned_weights` workaround in `maxtext_vllm_rollout.py` to correctly unstack the weights before they reach the vLLM converter.

src/maxtext/trainers/post_train/rl/train_rl.py
mesakhcienet: apologize, i used older version of vllm. I have reverted this change too

hengtaoguo (approved):
Thanks for the fix! Please resolve the code-style failures before merging.

SurbhiJainUSC (approved):

SurbhiJainUSC:
@mesakhcienet - Please fix code quality issue in this PR

PR #4410, opened 2026-07-09
[ DeepSeek-V4 ] training FLOPs and MFU calculation

parambole:
Description
===========

This PR adds training FLOPs and MFU calculation support for DeepSeek-V4 architectures.

### Details & Implementation:

-   Alternating Multi-Track Attention: DeepSeek-V4 alternates layers with different compression rates (sliding-window prefix C=0C=0, HCA C=128C=128, and CSA C=4C=4). This change dynamically checks the `compress_ratios` list inside the config to scale attention calculations based on the specific topology track of each layer.
-   Manifold-Constrained Hyper-Connections (mHC): Calculates the exact projection and contraction overheads across the expanded manifold space (k=4k=4), which run twice per layer (wrapping the attention and FFN blocks).
-   Golden Target Validation: Connects the DeepSeek-V4 block type to the main MFU calculations and checks estimations against dry-run targets in the test suite.
-   MTP Exclusion Note: Multi-Token Prediction (MTP) is not currently supported in the MaxText pretraining configuration. The MTP layer calculations are explicitly bypassed and set to 0, which aligns with standard training configurations where `mtp_num_layers = 0`.

Tests
=====

The changes were verified using the MaxText test suite in a local virtual environment.

Run the test suite:

```
python3 -m unittest tests/unit/flop_calculation_test.py
```

Specifically, `test_deepseek4_284b_flops` validates:

-   Computed total training FLOPs are approx `1402.72` TFLOPs.
-   Computed weight FLOPs are approx `1282.89` TFLOPs.
-   Computed attention FLOPs are approx `119.83` TFLOPs.

# Checklist

### DIFF
--- a/src/maxtext/utils/maxtext_utils.py
+++ b/src/maxtext/utils/maxtext_utils.py
@@ -702,7 +702,7 @@ def get_dense_moe_layers(config):
     num_moe_layers = config.num_decoder_layers // config.interleave_moe_layer_step
     num_dense_layers = config.num_decoder_layers - num_moe_layers
     return num_dense_layers, num_moe_layers
-  elif config.decoder_block in (DecoderBlockType.QWEN3_NEXT, DecoderBlockType.QWEN3_5):
+  elif config.decoder_block in (DecoderBlockType.QWEN3_NEXT, DecoderBlockType.QWEN3_5, DecoderBlockType.DEEPSEEK4):
     return 0, config.num_decoder_layers
   elif config.decoder_block == DecoderBlockType.DEFAULT:
     raise ValueError("Unsupported decoder block for dense/MoE layer calculation")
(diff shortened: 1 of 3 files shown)


tests/unit/flop_calculation_test.py
```diff
@@ -377,6 +377,19 @@ def test_deepseek32_671b_flops(self):
...
+  def test_deepseek4_284b_flops(self):
+    """Test DeepSeek V4 284B MFU FLOP calculation"""
+    cfg = self._initialize_model_config(
+        "deepseek4-284b",
+        max_target_length=4096,
+        per_device_batch_size=4,
+        attention="dot_product",
+    )
+    total_flops, weight_flops, attn_flops = calculate_tflops_training_per_device(cfg, log=False)
+    self.assertFlopsAlmostEqual(total_flops, 1402.557)
```
RissyRan: Maybe you have explain the source of those hardcoded numbers in bug? But they could be confusing for future readers. Wondering if you could find a way to calculated it from 6BP rule, like other tests?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,149 @@ def calculate_vision_encoder_tflops(config):
...
+      * (seq_len * sliding_window_size - 0.5 * sliding_window_size**2)
+      * config.num_query_heads
+      * config.head_dim
+  )
+
+  # - Track 2: HCA Compressed Branch (C=128)
+  #   Includes HCA compressor dense projection: D -> D_h
+  hca_pooler = 4 * batch_size * seq_len * config.emb_dim * config.head_dim
+  #   Calculates prefix sliding window attention + global cross-attention to the S/128 elements
+  hca_attn = prefix_attn + 2 * batch_size * seq_len * (seq_len / 128) * config.num_query_heads * config.head_dim
```
entrpn: I see 128 and 4 used in various places, can you create a constant, ex: hca_ratio, csa_ratio so its more clear.

entrpn (changes requested):

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,155 @@ def calculate_vision_encoder_tflops(config):
...
+  wkv_flops = 2 * batch_size * seq_len * config.emb_dim * (config.num_kv_heads * config.head_dim)
+  # - Grouped Out projection: projects out groups of heads down to o_lora_rank, then up to emb_dim.
+  #   FLOPs = 2 * B * S * (H * D_h) * o_lora_rank + 2 * B * S * (o_groups * o_lora_rank) * D
+  grp_feat = (config.num_query_heads * config.head_dim) // config.o_groups
+  o_flops = (2 * batch_size * seq_len * config.o_groups * grp_feat * config.o_lora_rank) + (
+      2 * batch_size * seq_len * (config.o_groups * config.o_lora_rank) * config.emb_dim
+  )
+  base_attention_weights = wq_flops + wkv_flops + o_flops
+
+  # 3. Manifold-Constrained Hyper-Connections (mHC) Overheads
```
RissyRan, line 1022: Thanks for adding this! Could you help extract this into a separate function? So that it can be triggered when `mhc_expansion_rate > 1`? Then you could directly call that function here.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,155 @@ def calculate_vision_encoder_tflops(config):
...
+  Returns:
+    A tuple of (total_attn_flops_in_tflops, total_weight_flops_in_tflops).
+  """
+  batch_size = config.per_device_batch_size
+  seq_len = config.max_target_length
+  sliding_window_size = config.sliding_window_size
+  num_layers = config.num_decoder_layers
+
+  # DeepSeek-V4 Attention topology compression rates (HCA = 128, CSA = 4)
+  HCA_RATIO = 128
```
RissyRan, line 987: Wondering if we could set them as config? So in the attn, you could leverage directly?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,155 @@ def calculate_vision_encoder_tflops(config):
...
+  CSA_RATIO = 4
+
+  # Note: Multi-Token Prediction (MTP) is not supported in the current MaxText configuration.
```
RissyRan: Shall we add assertion into types.py for DSv4 + MTP if not?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,155 @@ def calculate_vision_encoder_tflops(config):
   return mm_total_tflops, mm_learnable_weight_tflops, mm_attention_tflops
 
 
+def calculate_deepseek4_tflops_training_per_device(config, total_ffn_flops_all_layers, embedding_flops):
```
RissyRan, line 961: Trying to see if those are aligned with Megatron calculation, but it seems they haven't onboarded it yet: https://github.com/NVIDIA/Megatron-LM/blob/48a887fecd01674346f724dcf44f417f455989f0/megatron/training/training.py#L546

RissyRan (commented):
Thanks for the change!

src/maxtext/utils/maxtext_utils.py
shuningjin: Megatron does have FLOP calculation for ds v4 attention: 
- https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/e65a34658895a9e72314a23456c1b0d50d0d81b3/src/megatron/bridge/training/utils/flop_utils.py#L766-L856
- Here is AI-generated [comparison](https://docs.google.com/document/d/1qXQdv3A5xJXm6GeU4-H4AkZ5ad-rbgHnEBS8KRbPiPk/edit?usp=sharing). It seems calculation is more precise in this PR.
- Do we need to be identical to Megatron? @RissyRan

tests/unit/flop_calculation_test.py
parambole: Done. Removed the hardcoded FLOP values from the assertions in test_deepseek4_284b_flops. The test now dynamically derives the active parameter size directly from the configuration hyperparameters, and validates our computed FLOPs against the analytical 6BP rule baseline within the unit test's tolerance margins, aligning with the pattern used in the other decoder tests.

src/maxtext/utils/maxtext_utils.py
parambole: The sequence compression ratios are already dynamically represented by the config parameter compress_ratios (e.g. [0, 0, 4, 128...] in deepseek4-284b.yml). The constants HCA_RATIO = 128 and CSA_RATIO = 4 in maxtext_utils.py are just local helper aliases used for parsing and counting those occurrences. Adding duplicate top-level config keys for them is redundant.

entrpn (approved):

src/maxtext/utils/maxtext_utils.py
RissyRan: Nice finding! We don't need to be exactly the same, but at least high level alignment will be good. Also taking a look at comparison.

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,166 @@ def calculate_vision_encoder_tflops(config):
   return mm_total_tflops, mm_learnable_weight_tflops, mm_attention_tflops
 
 
+def calculate_mhc_flops_per_layer(config):
+  """Calculates the training FLOPs per layer for Manifold-Constrained Hyper-Connections (mHC)."""
+  batch_size = config.per_device_batch_size
+  seq_len = config.max_target_length
+  k_mhc = getattr(config, "mhc_expansion_rate", 4)
```
RissyRan: Nit: for the default value, shall we align https://github.com/AI-Hypercomputer/maxtext/blob/48c7da3e28e8f57fe36ddf09d00eec53072a2ea0/src/maxtext/configs/base.yml#L1275?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,166 @@ def calculate_vision_encoder_tflops(config):
...
+  Returns:
+    A tuple of (total_attn_flops_in_tflops, total_weight_flops_in_tflops).
+  """
+  batch_size = config.per_device_batch_size
+  seq_len = config.max_target_length
+  sliding_window_size = config.sliding_window_size
+  num_layers = config.num_decoder_layers
+
+  # Note: Multi-Token Prediction (MTP) is not supported in the current MaxText configuration.
+  # Therefore, MTP FLOPs are explicitly bypassed and set to 0.
```
RissyRan: Nit: shall we remove this comment? as MTP is calculated here for all models at the end? https://github.com/AI-Hypercomputer/maxtext/blob/48c7da3e28e8f57fe36ddf09d00eec53072a2ea0/src/maxtext/utils/maxtext_utils.py#L1100

src/maxtext/utils/maxtext_utils.py
```diff
@@ -938,6 +938,166 @@ def calculate_vision_encoder_tflops(config):
...
+  grp_feat = (config.num_query_heads * config.head_dim) // config.o_groups
+  o_flops = (2 * batch_size * seq_len * config.o_groups * grp_feat * config.o_lora_rank) + (
+      2 * batch_size * seq_len * (config.o_groups * config.o_lora_rank) * config.emb_dim
+  )
+  base_attention_weights = wq_flops + wkv_flops + o_flops
+
+  # 3. Manifold-Constrained Hyper-Connections (mHC) Overheads
+  # Inside mhc.py, inputs are projected to an expanded manifold space.
+  # These operations are executed twice per layer (wrapping the attention block and FFN block).
+  if getattr(config, "mhc_expansion_rate", 0) > 1:
```
RissyRan: Nit: could you help align with default? the above default for this config is 4.

RissyRan (approved):
Thanks for the change! Please ensure all comments are addressed from reviewers.

src/maxtext/utils/maxtext_utils.py
parambole: Done. Aligned the fallback default value for mhc_expansion_rate to 1 in both functions to match the default configuration value defined in base.yml

src/maxtext/utils/maxtext_utils.py
parambole: Done. Removed the local comment noting the bypass since MTP parameters are calculated globally at the end anyway.

src/maxtext/utils/maxtext_utils.py
parambole: Done. Aligned the fallback default value for mhc_expansion_rate to 1 in both functions to match the default configuration value defined in base.yml.

src/maxtext/utils/maxtext_utils.py
parambole: Resolving this for now. @shuningjin Thank you for the analysis.

src/maxtext/utils/maxtext_utils.py
parambole: @RissyRan Resolving this for now. Please reopen if you have additional concerns.

PR #4414, opened 2026-07-10
Add dense_fsdp_use_two_stage_all_gather

aireenmei:
# Description

Add flag `dense_fsdp_use_two_stage_all_gather` similar to `moe_fsdp_use_two_stage_all_gather` added in cl/805434569, achieving similar effect on dense layer then using both fadp and fadp_transpose. This will remove some reshape ops during all-gather.

Refactored based on Nuojin's suggestion:
* Instead of adding new logical axis, use helper functions to construct the two stages
* The existing helper functions such as `remove_fsdp_sharding` and `remove_expert_from_partition_spec`, shares similar logic on different axis. They are now consolidated into generic helper functions. Here is a summary table on helper functions:

| Old function (pre-PR) | Now | How |
|---|---|---|
| *(none — new)* | **`remove_mesh_axes_from_partition_spec(pspec, axes_to_remove, dims=None)`** | The single core peel loop; absorbs the bodies of all three old loops |
| *(none — new)* | **`remove_mesh_axes_from_sharding(tree, axes_to_remove)`** | Generic tree-map + NamedSharding wrapper (the reusable part of old `remove_fsdp_sharding`) |
| *(none — new)* | **`get_physical_spec_without_axes(full_logical, mesh, axes_to_remove, rules=None)`** | logical→physical + peel; the parameterized entry point the layer calls |
| `remove_fsdp_sharding(tree)` — inline `_remove_fsdp_from_partition_spec` loop, peels `fsdp`+`fsdp_transpose` from a NamedSharding tree | **Deleted** | Its only caller (`get_physical_spec_no_fsdp`) was rewired; the tree-peel behavior lives on in `remove_mesh_axes_from_sharding(tree, FSDP_MESH_AXES)` |
| `remove_expert_from_partition_spec(pspec, dims_to_peel)` — inline loop peeling `"expert"` from given dims | **Kept, now a 1-line wrapper** | Delegates to `remove_mesh_axes_from_partition_spec(pspec, ("expert",), dims=dims_to_peel)`; same signature, callers/test untouched |
| `get_physical_spec_no_fsdp(full_logical, mesh, rules)` — `logical_to_mesh_sharding` + `remove_fsdp_sharding` | **Kept, now a 1-line wrapper** | Delegates to `get_physical_spec_without_axes(full_logical, mesh, FSDP_MESH_AXES, rules)` |

# Tests
* Added unit test
* Profile shows the fwd all-gather-copy-reshape block reduces from 17.5 ms ([screenshot](https://screenshot.googleplex.com/3sBioQCz4t8f3e3)) to 6 ms ([screenshot](https://screenshot.googleplex.com/66GcS56UGmkFeZn))

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -656,6 +656,9 @@ check_vma: False
 # Enable ZeRO-1 optimizer sharding over data axis
 shard_optimizer_over_data: false
 
+# when dense MLP weight matrices are sharded on both fsdp and fsdp-transpose axes, use two separate all-gather calls
+dense_fsdp_use_two_stage_all_gather: false
+
 # Unless explicitly specified, the number of TPU slices is automatically determined. It should only be set for
 # disaggregated reinforcement learning workloads using multiple slices. For ahead of time compilation,
 # you should set compile_toplogy_num_slices, which will in turn set this value. For non-TPU environments this is set to 1.
--- a/src/maxtext/layers/linears.py
+++ b/src/maxtext/layers/linears.py
@@ -23,7 +23,7 @@ import jax
 import jax.numpy as jnp
 
 from jax import lax
-from jax.sharding import NamedSharding, Mesh
+from jax.sharding import NamedSharding, Mesh, PartitionSpec
 from jax.ad_checkpoint import checkpoint_name
 
 from flax import nnx
@@ -38,6 +38,9 @@ from maxtext.layers.quantizations import AqtQuantization as Quant
 from maxtext.utils import max_logging
 from maxtext.utils import max_utils
 from maxtext.utils.sharding import maybe_shard_with_logical
+from maxtext.utils.sharding import maybe_shard_with_name
+from maxtext.utils.sharding import get_physical_spec_without_axes
+from maxtext.utils.sharding import FSDP_MESH_AXES
 
 
 def _convert_to_activation_function(fn_or_string: str | Callable[..., Any]) -> Callable[..., Any]:
@@ -121,6 +124,9 @@ class DenseGeneral(nnx.Module):
       shard_mode: ShardMode = ShardMode.AUTO,
       matmul_precision: str = "default",
       parameter_memory_host_offload: bool = False,
+      mesh: Mesh | None = None,
+      use_two_stage_all_gather: bool = False,
+      debug_sharding: bool = False,
       *,  # Following arguments are keyword-only
       rngs: nnx.Rngs = None,
   ):
@@ -140,6 +146,13 @@ class DenseGeneral(nnx.Module):
       shard_mode: auto or explicit shard mode.
       matmul_precision: Precision for matrix multiplication.
       parameter_memory_host_offload: Determines whether to offload params to host
+      mesh: Mesh of devices and physical axes, needed for two-stage all-gather.
+      use_two_stage_all_gather: when the kernel is sharded on both the fsdp and
+        fsdp_transpose axes, gather the two axes with two separate all-gather
+        calls (separated by an optimization barrier) to avoid the relayout
+        transpose XLA emits for a single combined 2-axis all-gather.
+      debug_sharding: when True, log the logical/physical sharding of the
+        two-stage all-gather constraints to the sharding dump files.
       rngs: RNG state for initialization in nnx.
     """
     self.in_features_shape = canonicalize_tuple(in_features_shape)
@@ -154,6 +167,9 @@ class DenseGeneral(nnx.Module):
     self.shard_mode = shard_mode
     self.matmul_precision = matmul_precision
     self.parameter_memory_host_offload = parameter_memory_host_offload
+    self.mesh = mesh
+    self.use_two_stage_all_gather = use_two_stage_all_gather
+    self.debug_sharding = debug_sharding
 
     # Parameter initialization
     kernel_shape = self.in_features_shape + self.out_features_shape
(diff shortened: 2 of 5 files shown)


RissyRan (commented):
Also discuss offline. 2 stage all-gather is a workaround before 3D all-gather support.

src/maxtext/configs/base.yml
```diff
@@ -607,6 +607,9 @@ logical_axis_rules: [
                       # ==========================================
                       ['mlp_no_fsdp', ['tensor', 'tensor_sequence', 'autoregressive']],
                       ['exp_with_fsdp', 'fsdp'],
+                      # Helper rules for dense_fsdp_use_two_stage_all_gather (two-stage MLP-weight all-gather).
+                      ['embed_fsdp', ['fsdp']],
+                      ['embed_no_fsdp', []],
```
RissyRan: You could use None if duplicated. Like 

https://github.com/AI-Hypercomputer/maxtext/blob/83527b8c75530a0f6b9f9d76c4b764147abe21ad/src/maxtext/layers/moe.py#L2134

src/maxtext/configs/base.yml
```diff
@@ -607,6 +607,9 @@ logical_axis_rules: [
                       # ==========================================
                       ['mlp_no_fsdp', ['tensor', 'tensor_sequence', 'autoregressive']],
                       ['exp_with_fsdp', 'fsdp'],
+                      # Helper rules for dense_fsdp_use_two_stage_all_gather (two-stage MLP-weight all-gather).
+                      ['embed_fsdp', ['fsdp']],
```
RissyRan: I see this section is going to be removal, potentially you could leverage ['embed', ['fsdp', 'context', 'expert']] above? expert will not applied in your case.

RissyRan (commented):
Thanks! Minor comments about annotation.

src/maxtext/configs/base.yml
aireenmei: Sounds good, removed 'embed_no_fsdp' rule and use None instead.

src/maxtext/configs/base.yml
aireenmei: There are two embed rules and in this case the first one will win, we need to use another name here.
```
['embed', ['fsdp', 'fsdp_transpose', 'context', 'expert']],
['embed', ['fsdp', 'context', 'expert']],
```
Thanks for reminding this is the "deprecation" section of the config. I now moved 'embed_fsdp' up following the 'emb' rules and add comments for the use case. Also since this feature uses 'mlp_no_fsdp', I moved it out of the deprecation section to follow the 'mlp' rule, with comments on use case.

NuojCheng (commented):
I suggest using helper function, e.g. https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/utils/sharding.py#L812-L834, to directly edit sharding/partitions instead of adding new rule in base.yml. It complicates a general-purpose rule for a limited use case. 

We plan to remove all rules with physical axis names in them, e.g. `mlp_no_fsdp` and `exp_with_fsdp`, since we want to separate the logical rule/physical rule systems.

src/maxtext/layers/linears.py
```diff
@@ -200,6 +209,37 @@ def quant_dot_general(self) -> nnx_wrappers.ToNNX | None:
...
+    transpose to fix the layout. Splitting into two single-axis gathers separated
+    by an `optimization_barrier` makes each stage produce a contiguous layout, so
+    no transpose is emitted. Mirrors `moe_fsdp_use_two_stage_all_gather`.
+    """
+    # Stage axes per supported MLP kernel layout; fused_mlp
+    # ("embed", "num_activations", "mlp") is absent and thus skipped below.
+    stage_axes = {
+        ("embed", "mlp"): (("embed_fsdp", "mlp_no_fsdp"), (None, "mlp_no_fsdp")),
+        ("mlp", "embed"): (("mlp", None), ("mlp_no_fsdp", None)),
+    }.get(self.kernel_axes)
```
NuojCheng: I would use physical partitions to do the 2-step sharding:
1. collect physical partitions using logical to mesh axes, e.g. https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/utils/sharding.py#L251;
2. peel the fsdp from physical partition using https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/utils/sharding.py#L812 or https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/utils/sharding.py#L865;
3. apply maybe_shard_with_pspec (or similar) to apply the sharding logic. 

Also, I think AG over FSDP transpose first (especially if they are over core) then AG over FSDP could be faster than the other way (in theory, no overlap)? Have you tried AG FSDP_T first?

aireenmei: Good suggestion and thanks for the pointer! I revised the PR, also refactor the helper functions, see details in the PR description following "Refactored based on Nuojin's suggestion:" And the revised version use AG over FSDP_T first

src/maxtext/layers/linears.py
```diff
@@ -200,6 +212,39 @@ def quant_dot_general(self) -> nnx_wrappers.ToNNX | None:
...
+    # lookup treats it as a single spec rather than a pytree of strings.
+    full_logical = PartitionSpec(*self.kernel_axes)
+    # Stage 1 gathers fsdp_transpose, stage 2 gathers the remaining fsdp.
+    stage1 = get_physical_spec_without_axes(full_logical, self.mesh, ("fsdp_transpose",))
+    stage2 = get_physical_spec_without_axes(full_logical, self.mesh, FSDP_MESH_AXES)
+    if stage1.spec == stage2.spec:
+      # Not sharded on both FSDP axes, so a single all-gather is already optimal.
+      return kernel
+
+    shard = functools.partial(maybe_shard_with_name, shard_mode=self.shard_mode)
```
NuojCheng: could you also add debug_sharding=self.config.debug_sharding?

src/maxtext/utils/sharding.py
```diff
@@ -806,34 +806,50 @@ def get_formatted_sharding_annotations(params, mesh=None):
...
 
 
-def remove_fsdp_sharding(sharding_tree):
-  """Recursively traverses the sharding tree to remove fsdp axes."""
+FSDP_MESH_AXES = ("fsdp", "fsdp_transpose")
 
-  def _remove_fsdp_from_partition_spec(named_sharding):
-    """Removes 'fsdp' and 'fsdp_transpose' from a PartitionSpec."""
+
+def remove_mesh_axes_from_partition_spec(pspec, axes_to_remove, dims=None):
```
NuojCheng, line 828: love this function thank you!

NuojCheng (approved):
Thank you for the fix and love your added functions in sharding.py!

src/maxtext/configs/base.yml
aireenmei: The function has been refactored with no logical rules added. Please take another look.

src/maxtext/configs/base.yml
aireenmei: The function has been refactored with no logical rules added. Please take another look.

src/maxtext/layers/linears.py
aireenmei: The class doesn't have self.config, I added self.debug_sharding to pass this config

aireenmei:
I add pull ready to generate internal cl for testing. Won't submit until PR approved

RissyRan (approved):

PR #4419, opened 2026-07-10
SFT: tools rendering + developer role in chat-template preprocessing

pekach-d:
# Description

Add first-class support for **tool-calling SFT data** and the **`developer` role** to the
grain/HF chat-template preprocessing, so tool-augmented conversations render and mask correctly —
including across multi-turn conversations. Today the SFT pipeline only understands `system`/`user`/
`assistant` turns and a single primary data column; there is no way to feed tool/function
definitions to the chat template, the `developer` role is silently dropped, and a multi-turn
conversation loses its leading turn after the first round.

This change is fully opt-in and backward compatible: datasets without a `tools` column and without
`developer` turns behave exactly as before (`tools_column_name` defaults to `None`).

## What this adds

**Tools column.** A dataset may now provide an auxiliary `tools` column alongside `messages`
(supported column set `["messages", "tools"]`). The tools (a JSON string or a list of tool
definitions) are split off the primary data column, threaded through `apply_chat_template`, rendered
into every `apply_chat_template` call so tool-calling turns tokenize correctly, and dropped from the
dataset afterward. The chat-template wrapper also learns the `tool` role (tool results accumulate
into the current round without emitting an extra prompt/completion segment).

- `data_processing_utils.py`: `TOOLS_COLUMN = "tools"`; accept `["messages", "tools"]`.
- `hf_data_processing.py` / `grain_data_processing.py`: separate the `tools` column, thread it into
  `apply_chat_template`, drop it afterward (HF path also deserializes JSON string columns).
- `input_pipeline_utils.py`: `apply_chat_template` / `_get_completion_in_chat_template` accept the
  tools and pass them to every chat-template call; `enable_thinking=True` keeps the
  completion-boundary detection correct for thinking-channel templates (e.g. Gemma).

**Developer role.** `developer` is treated like `system` (a leading, index-0 role). Any other/
unknown role now raises instead of being silently ignored.

**Optional tools column (mixture-friendly).** `ParseFeatures` / `NormalizeFeatures` / `KeepFeatures`
treat `tools` as optional (`OPTIONAL_COLUMNS = {"tools"}`): a record without it is skipped rather
than raising, so a single grain mixture can blend tools and non-tools datasets. Required columns
still raise when missing.

**Multi-turn tools persistence.** Each new round is re-seeded with the leading system/developer turn
(`leading_msgs`) instead of clearing `round_msgs`. Clearing dropped the leading turn, so rounds 2+
of a multi-turn conversation rendered as a bare `<bos>user` prompt with no tools; under
completion-only loss that trained those rounds' tool-calls as (no-tools → call), which makes the
model emit spurious tool calls at inference. Re-seeding keeps the tools/persona rendered in every
round.

## Backward compatibility

- `tools_column_name` defaults to `None` → existing `messages`-only / `prompt,completion` /
  `question,answer` datasets are unchanged.
- The `developer` role and the optional-`tools` handling are additive; no existing behavior changes
  unless the data uses them.

## Shortcomings / future work

- Tools are rendered whenever present; there is no per-turn tool visibility control.
- `enable_thinking=True` is currently unconditional (kept for correct completion-boundary detection
  on thinking-channel templates); making it configurable is a possible follow-up.

# Tests

The change preserves the existing `apply_chat_template` signature for non-tools callers, so the
existing SFT data-processing tests (`tests/post_training/unit/sft_data_processing_test.py`) continue
to exercise the system/user/assistant paths for the qwen3 / llama2 / gemma4 tokenizers:

```bash
python -m pytest tests/post_training/unit/sft_data_processing_test.py -v
```

The new control-flow (developer role, unknown-role error, `tool`-role accumulation, multi-turn
leading-turn re-seeding, and the optional-`tools` skipping) was validated against the migrated
functions with a mock tokenizer:
- `developer` accepted as a leading (index-0) turn with tools passed through; `developer` at a
  non-zero index raises.
- an unsupported role raises (the new `else` branch).
- a `tool` message accumulates into the round and emits no extra segment.
- round 2 of a multi-turn conversation still renders the leading `developer` turn + tools (not a
  bare user prompt).
- `NormalizeFeatures`/`KeepFeatures` skip an absent `tools` column but still raise on an absent
  required column.

# Checklist

### DIFF
--- a/src/maxtext/input_pipeline/data_processing_utils.py
+++ b/src/maxtext/input_pipeline/data_processing_utils.py
@@ -24,6 +24,8 @@ from maxtext.input_pipeline import input_pipeline_utils
 from maxtext.input_pipeline import tokenizer
 from maxtext.utils import elastic_utils
 
+TOOLS_COLUMN = "tools"
+
 
 def parse_and_keep_features(dataset, config, data_columns, tokenize):
   """Parse arrayrecord features or keep specified columns for other formats."""
@@ -58,7 +60,7 @@ def validate_and_configure_sft_columns(data_columns, tokenizer_model, chat_templ
   if chat_template and hasattr(tokenizer_model, "chat_template"):
     tokenizer_model.chat_template = chat_template
 
-  supported_columns = [["prompt", "completion"], ["messages"], ["question", "answer"]]
+  supported_columns = [["prompt", "completion"], ["messages"], ["messages", TOOLS_COLUMN], ["question", "answer"]]
   assert any(
       set(data_columns) == set(supported) for supported in supported_columns
   ), f"Dataset column names mismatch. Expected columns to match one of {supported_columns}, but got {data_columns}"
--- a/src/maxtext/input_pipeline/grain_data_processing.py
+++ b/src/maxtext/input_pipeline/grain_data_processing.py
@@ -307,26 +307,34 @@ def dpo_preprocessing_pipeline(
 
 def _format_chat_template_grain(element, data_columns, tokenizer_model):
   """Grain-compatible mapping function to format raw columns into conversational messages."""
+  tools_column_name = data_processing_utils.TOOLS_COLUMN if data_processing_utils.TOOLS_COLUMN in data_columns else None
+  primary_columns = [c for c in data_columns if c != data_processing_utils.TOOLS_COLUMN]
+
   # Convert raw columns to conversational messages
-  if "messages" in data_columns:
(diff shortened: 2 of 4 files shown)


aireenmei (commented):
Sorry we missed this PR. Could you:
1. Sign CLA according to the link in cla bot message. So that CI tests can run
2. Rebase to head
3. Add unit tests. Can add to tests/post_training/unit/
  sft_data_processing_test.py for apply_chat_template changes and prompt masking.

PR #4444, opened 2026-07-13
[Deepseek V4] Decoding Support & Integration with MaxEngine

Rohan-Bierneni:
# Description

# Enable End-to-End DeepSeek-V4 Decoding, Caching, and API Server Support

## Overview

This pr enables end-to-end decoding and inference server support for **DeepSeek-V4** architectures, which rely on Compressed Attention (Hybrid Compressed Attention and Compressed Sliding Attention) and interleaved Rotary Positional Embeddings.

This PR enables full **autoregressive decoding, KV cache management, and OpenAI-compatible API server serving** for DeepSeek-V4 models in MaxText and MaxEngine.

---

### 1. Multi-Tier Autoregressive KV Caching (`src/maxtext/inference/`)

- **DeepSeek-V4 Cache State (`KVCache`)**: Extends `KVCache` to manage DeepSeek-V4's multi-component cache state alongside standard key/value buffers:
  - **Accumulator Buffers** (`leftover_buffer_kv`, `leftover_buffer_gate`): Buffers incoming tokens until a full compression window is reached.
  - **Overlap Registers** (`overlap_kv`, `overlap_gate`): Carries trailing window state across sequence chunks for Compressed Sliding Attention (CSA).
  - **Block Trackers**: Tracks total entry counts and valid autoregressive block lengths per sequence.
- **MaxEngine Integration (`maxengine.py`)**: Adds DeepSeek-V4 cache leaves (`_DEEPSEEK_V4_CACHE_KEYS`) to engine slot operations (`bulk_insert`, `insert`, `replace`), using XLA `DynamicUpdateSlice` primitives to update concurrent decode slots in-place without HBM copies.

### 2. Compressed Attention Decoding Pipeline (`src/maxtext/layers/`)

- **Autoregressive Token-by-Token Compression (`attention_compressed.py`)**:
  - Implements decoding execution paths for both **HCA** (non-overlapping pooling) and **CSA / Indexer** (overlapping window pooling).
  - Integrates clean helper abstractions for prefill chunk pooling, cache state priming, and safe RoPE keyword argument dispatching.
- **Dynamic Validity Masking**:
  - Constructs 3D/4D binary validity masks combining statically padded prefill blocks and autoregressive blocks so queries attend only to completed, valid compressed representations.
- **Batched Sliding-Window Attention (`attention_op.py`)**:
  - Ensures each sequence in a batched decoding request independently evaluates its sliding-window token positions (`abs_k`), supporting multi-sequence workloads where active autoregressive sequences and newly started prompts coexist in the same batch.
  

### 3. API Server & Tokenization (`benchmarks/api_server/`)

- **Harmony Tokenizer (`encoding_dsv4.py`)**: Adds native DeepSeek-V4 / Harmony tokenization and formatting support.
- **Server Integration**: Updates `maxtext_server.py`, `server_models.py`, and `server_utils.py` to route DeepSeek-V4 requests, configure special tokens, and support generation parameters.
- **Concurrent Benchmark Tests**: Adds `test_batching_chat_completions.py` and `test_batching_text_completions.py` for verifying multi-request serving and throughput.

FIXES: b/123456

# Tests

## Forward pass logit checker

 - Script: https://paste.googleplex.com/4927710567727104
 - Result: https://paste.googleplex.com/4776313389973504. Successful run based on comment in [b/509930698#comment18](https://b.corp.google.com/issues/509930698#comment18)

## Api_server Run

XPK Cmd: https://paste.googleplex.com/5792652892176384

#### Sample Curl Request for 500 tokens decoding
 - Cmd: https://paste.googleplex.com/4719573701099520
 - Output: https://paste.googleplex.com/5832623954984960

#### test_batching_chat_completions.py
 - Result: https://paste.googleplex.com/4510990040432640

## Training run on deepseek4-tiny.yml (on v5p-8 TPU vm)

 - Script: https://paste.googleplex.com/4891722382442496
 - Result: https://paste.googleplex.com/5805793009074176

# Checklist

### DIFF
--- a/benchmarks/api_server/maxtext_server.py
+++ b/benchmarks/api_server/maxtext_server.py
@@ -59,7 +59,7 @@ from benchmarks.api_server.server_models import (
     ChatMessage,
 )
 from benchmarks.api_server import server_utils
-from benchmarks.api_server.encoding import encoding_dsv32
+from benchmarks.api_server.encoding import encoding_dsv32, encoding_dsv4
 
 # ----------------------------
 # Init
@@ -187,14 +187,25 @@ def _build_chat_completion_response(request, completion_result, llm):
     except (ValueError, IndexError) as e:
       logger.error("Harmony parsing failed for gpt-oss: %s. Falling back to raw text.", e, exc_info=True)
 
+  t_mode = getattr(request, "thinking_mode", "thinking")
+
   if server_utils.is_dsv32_encoding_enabled(request.model):
     try:
       # DeepSeek-V3.2 models often generate thinking block.
-      parsed = encoding_dsv32.parse_message_from_completion_text(text_out, thinking_mode="thinking")
+      parsed = encoding_dsv32.parse_message_from_completion_text(text_out, thinking_mode=t_mode)
       text_out = parsed.get("content", text_out)
       reasoning_out = parsed.get("reasoning_content")
     except (AssertionError, ValueError, IndexError) as e:
       logger.error("DeepSeek-V3.2 parsing failed: %s. Falling back to raw text.", e, exc_info=True)
+
+  if server_utils.is_dsv4_encoding_enabled(request.model):
+    try:
+      # DeepSeek-V4 models often generate thinking block.
+      parsed = encoding_dsv4.parse_message_from_completion_text(text_out, thinking_mode=t_mode)
+      text_out = parsed.get("content", text_out)
+      reasoning_out = parsed.get("reasoning_content")
+    except (AssertionError, ValueError, IndexError) as e:
+      logger.error("DeepSeek-V4 parsing failed: %s. Falling back to raw text.", e, exc_info=True)
 
   want_top_logprobs = (
       (request.top_logprobs or 0) > 0 if isinstance(request, ChatCompletionRequest) else (request.logprobs or 0) > 0
@@ -322,9 +333,17 @@ def _prepare_batch_for_broadcast(batched_items):
   request_info_map = []
   for req_id, req in batched_items:
     is_chat = isinstance(req, ChatCompletionRequest)
-    prompts_for_req = server_utils.get_prompts_for_request(req, LLM)
-    all_prompts.extend(prompts_for_req)
-    request_info_map.append((req_id, req, is_chat, len(prompts_for_req)))
+    try:
+      prompts_for_req = server_utils.get_prompts_for_request(req, LLM)
+      all_prompts.extend(prompts_for_req)
+      request_info_map.append((req_id, req, is_chat, len(prompts_for_req)))
+    except ValueError as e:
+      logger.error("Failed to parse request %s: %s", req_id, e)
+      with response_lock:
+        response_dict[req_id] = {"error": f"Invalid request format or missing chat template: {e}"}
+
+  if not all_prompts:
+    return 0, b"", []  # Signal other ranks to skip if all requests failed
 
   broadcast_payload = {"prompts": all_prompts, "params": params}
   payload_bytes = json.dumps(broadcast_payload).encode("utf-8")
--- a/benchmarks/api_server/test_batching_chat_completions.py
+++ b/benchmarks/api_server/test_batching_chat_completions.py
@@ -1,0 +1,87 @@
+# Copyright 2023–2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Client test script for verifying concurrent chat completion batching against the MaxText API server."""
+
+import concurrent.futures
+import time
+import requests
+
+URL = "http://localhost:8000/v1/chat/completions"
+HEADERS = {"Content-Type": "application/json"}
+
+PROMPTS = [
+    "What is attention in transformers?",
+    "Explain the concept of backpropagation.",
+    "How does a Convolutional Neural Network work?",
+    "What are the benefits of using a learning rate scheduler?",
+]
+
+
+def send_request(req_id, prompt):
+  """Send a single chat completion request to the API server."""
+  print(f"[{req_id}] Sending request: '{prompt}'")
+  start_time = time.time()
+
+  payload = {
+      "model": "deepseekv4",
+      "messages": [{"role": "user", "content": prompt}],
+      "thinking_mode": "chat",
+      "max_tokens": 20,
+      "temperature": 0.7,
+      "top_p": 0.95,
+  }
+
+  try:
+    response = requests.post(URL, json=payload, headers=HEADERS)
+    duration = time.time() - start_time
+    print(f"[{req_id}] Received response in {duration:.2f}s. Status: {response.status_code}")
+    return req_id, prompt, response.json()
+  except (requests.RequestException, ValueError) as e:
+    print(f"[{req_id}] Request failed: {e}")
+    return req_id, prompt, None
+
+
+def test_batching():
+  """Run concurrent requests to test batching."""
+  num_requests = len(PROMPTS)
+  print(f"Testing batching by firing {num_requests} concurrent requests with different prompts...")
+
+  # Use a ThreadPoolExecutor to fire requests simultaneously
+  with concurrent.futures.ThreadPoolExecutor(max_workers=num_requests) as executor:
+    # Submit all tasks at once
+    futures = {executor.submit(send_request, i, prompt): i for i, prompt in enumerate(PROMPTS)}
+
+    # Wait for all of them to finish
+    for future in concurrent.futures.as_completed(futures):
+      req_id = futures[future]
+      try:
+        ret_req_id, prompt, data = future.result()
+        if data and "choices" in data and len(data["choices"]) > 0:
+          choice = data["choices"][0]
+          message = choice.get("message", {})
+          content = message.get("content", "")
+          reasoning = message.get("reasoning_content", "")
+
+          print(f"\n--- Result for Request [{ret_req_id}] ---")
+          print(f"Prompt: {prompt}")
+          if reasoning:
+            print(f"Thinking: {reasoning.strip()}")
+          print(f"Output: {content.strip()}")
+      except (RuntimeError, ValueError) as exc:
+        print(f"[{req_id}] Generated an exception: {exc}")
+
+
+if __name__ == "__main__":
+  test_batching()
--- a/src/maxtext/inference/kvcache.py
+++ b/src/maxtext/inference/kvcache.py
@@ -14,7 +14,7 @@
 
 """Implementation of the kvcache."""
 
-from typing import Any
+from typing import Any, Optional, Callable
 
 import jax
 import jax.numpy as jnp
@@ -175,6 +175,9 @@ def kv_cache_as_linen(
     use_chunked_prefill: bool = False,
     model_mode: str = MODEL_MODE_PREFILL,
     is_gdn: bool = False,
+    is_deepseek_v4: bool = False,
+    compress_rate: int = 1,
+    is_indexer: bool = False,
     conv_kernel_size: int = 0,
     conv_dim: int = 0,
     name: str | None = None,
@@ -228,6 +231,9 @@ def kv_cache_as_linen(
       use_chunked_prefill=use_chunked_prefill,
       model_mode=model_mode,
       is_gdn=is_gdn,
+      is_deepseek_v4=is_deepseek_v4,
+      compress_rate=compress_rate,
+      is_indexer=is_indexer,
       conv_kernel_size=conv_kernel_size,
       conv_dim=conv_dim,
       metadata_fn=variable_to_logically_partitioned,
@@ -272,6 +278,9 @@ class KVCache(BaseCache):
       use_chunked_prefill: bool = False,
       model_mode: str = MODEL_MODE_PREFILL,
       is_gdn: bool = False,
+      is_deepseek_v4: bool = False,
+      compress_rate: int = 1,
+      is_indexer: bool = False,
       conv_kernel_size: int = 0,
       conv_dim: int = 0,
       *,
@@ -324,12 +333,41 @@ class KVCache(BaseCache):
     self.model_mode = model_mode
     self.use_chunked_prefill = use_chunked_prefill
     self.is_gdn = is_gdn
+    self.is_deepseek_v4 = is_deepseek_v4
+    self.compress_rate = compress_rate
+    self.is_indexer = is_indexer
     self.conv_kernel_size = conv_kernel_size
     self.conv_dim = conv_dim
 
     if model_mode in (MODEL_MODE_PREFILL, MODEL_MODE_AUTOREGRESSIVE):
       self._initialize_prefill_caches(model_mode)
       self._initialize_ar_cache_vars(model_mode)
+
+      if self.is_deepseek_v4 and self.compress_rate > 1:
+        cache_batch_axis_name = CACHE_BATCH_PREFILL if model_mode == MODEL_MODE_PREFILL else CACHE_BATCH
+
+        self.entry_count = nnx.Cache(
+            jnp.zeros((self.batch, 1), dtype=jnp.int32), out_sharding=(cache_batch_axis_name, None)
+        )
+        self.accumulator_index = nnx.Cache(
+            jnp.zeros((self.batch, 1), dtype=jnp.int32), out_sharding=(cache_batch_axis_name, None)
+        )
+        self.leftover_buffer_kv = nnx.Cache(
+            jnp.zeros((self.batch, self.compress_rate, self.key_heads, self.key_head_size), dtype=dtype),
+            out_sharding=(cache_batch_axis_name, None, None, None),
+        )
+        self.leftover_buffer_gate = nnx.Cache(
+            jnp.zeros((self.batch, self.compress_rate, self.key_heads, self.key_head_size), dtype=dtype),
+            out_sharding=(cache_batch_axis_name, None, None, None),
+        )
+        self.overlap_kv = nnx.Cache(
+            jnp.zeros((self.batch, self.key_heads, self.compress_rate, self.key_head_size), dtype=dtype),
+            out_sharding=(cache_batch_axis_name, None, None, None),
+        )
+        self.overlap_gate = nnx.Cache(
+            jnp.zeros((self.batch, self.key_heads, self.compress_rate, self.key_head_size), dtype=dtype),
+            out_sharding=(cache_batch_axis_name, None, None, None),
+        )
 
   @property
   def prefill_key_vars(self):
@@ -718,6 +756,7 @@ class KVCache(BaseCache):
     if decoder_segment_ids is not None:
       assert cached_prefill_segment_id_var is not None
       cached_prefill_segment_id_var.set_value(decoder_segment_ids)
+
     return key, value, decoder_segment_ids
 
   def update_ar_key_value(
@@ -928,6 +967,132 @@ class KVCache(BaseCache):
         cached_ar_segment_id_var.get_value(),
         cache_ar_lengths_var.get_value(),
     )
+
+    return cached_prefill, cached_ar
+
+  def kv_cache_autoregressive_v4(
+      self,
+      key: Array,
+      value: Array,
+      gate: Optional[Array] = None,
+      use_ragged_attention: bool = False,
+      compressor_fn: Optional[Callable] = None,
+  ):
+    """DeepSeek-V4 aware token-by-token caching matrix (Functionally Pure for JAX tracing)."""
+    if self.compress_rate == 1:
+      return self.kv_cache_autoregressive(key, value, use_ragged_attention)
+
+    assert gate is not None, "gate cannot be None when compress_rate > 1 in DeepSeek-V4 autoregressive cache."
+    assert self.cached_ar_key is not None, "cached_ar_key cannot be None when updating autoregressive cache."
+
+    incoming_batch = key.shape[0]
+
+    # 1. Update the Accumulator Buffers
+    current_index_array = self.accumulator_index.get_value()
+    current_index = jnp.reshape(current_index_array, (-1,))[0]
+
+    buffer_kv = jax.lax.dynamic_update_slice_in_dim(self.leftover_buffer_kv.get_value(), key, current_index, 1)
+    buffer_gate = jax.lax.dynamic_update_slice_in_dim(self.leftover_buffer_gate.get_value(), gate, current_index, 1)
+
+    self.leftover_buffer_kv.set_value(buffer_kv)
+    self.leftover_buffer_gate.set_value(buffer_gate)
+
+    next_index = current_index + 1
+    window_complete = next_index == self.compress_rate
+
+    # 2. Compute the compressed block using the layer's math callback
+    if compressor_fn is not None:
+      compressed_block, next_overlap_kv, next_overlap_gate = compressor_fn(buffer_kv, buffer_gate)
+
+      # Conditionally update the overlap registers only if the window finished
+      if next_overlap_kv is not None:
+        new_overlap_kv = jnp.where(window_complete, next_overlap_kv, self.overlap_kv.get_value())
+        self.overlap_kv.set_value(new_overlap_kv)
+
+      if next_overlap_gate is not None:
+        new_overlap_gate = jnp.where(window_complete, next_overlap_gate, self.overlap_gate.get_value())
+        self.overlap_gate.set_value(new_overlap_gate)
+    else:
+      gate_weights = jax.nn.softmax(buffer_gate, axis=1).astype(buffer_kv.dtype)
+      compressed_block = jnp.sum(buffer_kv * gate_weights, axis=1, keepdims=True)
+
+    potential_update_key = compressed_block
+
+    operand_shape = self.cached_ar_key.get_value().shape
+    if potential_update_key.shape[3] < operand_shape[3]:
+      # DeepSeek-V4 compressed latent dimension d_c can be smaller than cached key_head_size
+      # (which is often padded to power-of-two / systolic array boundaries for TPU efficiency).
+      # Pad trailing zeros along head_dim (axis 3) to match cache operand shape.
+      pad_amt = operand_shape[3] - potential_update_key.shape[3]
+      potential_update_key = jnp.pad(potential_update_key, ((0, 0), (0, 0), (0, 0), (0, pad_amt)))
+
+    # If window incomplete, we will just write dummy zeros
+    dummy_key = jnp.zeros_like(potential_update_key)
+    update_key = jnp.where(window_complete, potential_update_key, dummy_key)
+
+    # 3. Fetch current AR cache state variables
+    (
+        cached_ar_key_vars,
+        cached_ar_value_vars,
+        cached_ar_segment_id_var,
+        _,
+        cache_ar_lengths_var,
+    ) = self._get_ar_cache_vars()
+    assert cached_ar_segment_id_var is not None, "cached_ar_segment_id_var cannot be None"
+    ar_index = self.cache_ar_index.get_value()
+
+    # 4. Write to the cache array
+    self.update_ar_key_value(
+        update_key,
+        update_key,  # Value is identical to key for DeepSeek V4 compressed states
+        cached_ar_key_vars,
+        cached_ar_value_vars,
+        ar_index,
+        cache_ar_lengths_var.get_value(),
+        use_ragged_attention,
+    )
+
+    # 5. Conditionally write the segment ID mask for the incoming batch
+    active_indicator = jnp.zeros((incoming_batch, 1), dtype=jnp.int32) + DECODING_ACTIVE_SEQUENCE_INDICATOR
+    zero_indicator = jnp.zeros_like(active_indicator)
+    indicator_to_write = jnp.where(window_complete, active_indicator, zero_indicator)
+
+    temp_ar_seg = cached_ar_segment_id_var.get_value()
+
+    new_ar_seg = jax.lax.dynamic_update_index_in_dim(temp_ar_seg, indicator_to_write, jnp.squeeze(ar_index), 1)
+    cached_ar_segment_id_var.set_value(new_ar_seg)
+
+    # 6. Conditionally advance all the pointers and counters
+    next_ar_index = jnp.mod(ar_index + 1, self.max_target_length - self.max_prefill_length)
+    self.cache_ar_index.set_value(jnp.where(window_complete, next_ar_index, ar_index))
+
+    self.entry_count.set_value(jnp.where(window_complete, self.entry_count.get_value() + 1, self.entry_count.get_value()))
+
+    cache_ar_lengths_var.set_value(
+        jnp.where(window_complete, cache_ar_lengths_var.get_value() + 1, cache_ar_lengths_var.get_value())
+    )
+
+    self.accumulator_index.set_value(
+        jnp.full_like(current_index_array, jnp.where(window_complete, jnp.int32(0), next_index))
+    )
+
+    # --- 7. UNPACK JAX ARRAYS TO MATCH STANDARD ATTENTION PIPELINE ---
+    cached_prefill_key_vars, cached_prefill_value_vars, cached_prefill_segment_id_var = self._get_prefill_cache_vars()
+    assert cached_prefill_segment_id_var is not None, "cached_prefill_segment_id_var cannot be None"
+
+    cached_prefill = (
+        self.get_cached_values(cached_prefill_key_vars, key.dtype, self.prefill_cache_axis_order),
+        self.get_cached_values(cached_prefill_value_vars, value.dtype, self.prefill_cache_axis_order),
+        cached_prefill_segment_id_var.get_value(),
+    )
+
+    cached_ar = (
+        self.get_cached_values(cached_ar_key_vars, key.dtype, self.ar_cache_axis_order),
+        self.get_cached_values(cached_ar_value_vars, value.dtype, self.ar_cache_axis_order),
+        cached_ar_segment_id_var.get_value(),
+        cache_ar_lengths_var.get_value(),
+    )
+
     return cached_prefill, cached_ar
 
   def __call__(
@@ -938,6 +1103,8 @@ class KVCache(BaseCache):
       model_mode: str,
       use_ragged_attention: bool = False,
       previous_chunk: Any = None,
+      gate: Optional[Array] = None,
+      compressor_fn: Optional[Callable] = None,
   ) -> tuple:
     """KV cache takes the current state and updates the state accordingly.
 
@@ -962,6 +1129,8 @@ class KVCache(BaseCache):
       else:
         return self.kv_cache_prefill(key, value, decoder_segment_ids), None
     elif model_mode == MODEL_MODE_AUTOREGRESSIVE:
+      if self.is_deepseek_v4 and self.compress_rate > 1:
+        return self.kv_cache_autoregressive_v4(key, value, gate, use_ragged_attention, compressor_fn)
       return self.kv_cache_autoregressive(key, value, use_ragged_attention)
     else:
       raise ValueError(f"Model Mode isn't supported! {model_mode=}")
@@ -1134,6 +1303,7 @@ class MlaKVCache(KVCache):
       model_mode: str,
       use_ragged_attention: bool = False,
       previous_chunk: Any = None,
+      gate: Optional[Array] = None,
   ) -> tuple[
       None | tuple[Array, Array, Array],
       None | tuple[Array, Array, Array, Array],
--- a/src/maxtext/inference/maxengine/maxengine.py
+++ b/src/maxtext/inference/maxengine/maxengine.py
@@ -57,6 +57,16 @@ config_lib, engine_api, token_utils, tokenizer_api, _token_params_ns = jetstream
 TokenizerParameters = getattr(_token_params_ns, "TokenizerParameters", object)  # type: ignore[assignment]
 TokenizerType = getattr(_token_params_ns, "TokenizerType", object)  # type: ignore[assignment]
 
+_DEEPSEEK_V4_CACHE_KEYS = frozenset(
+    (
+        "entry_count",
+        "accumulator_index",
+        "leftover_buffer_kv",
+        "leftover_buffer_gate",
+        "overlap_kv",
+        "overlap_gate",
+    )
+)
 
 warnings.simplefilter("ignore", category=FutureWarning)
 DecodeState = Any
@@ -1495,6 +1505,12 @@ class MaxEngine(_BaseEngine):  # pyrefly: ignore[invalid-inheritance]
       if batch_idx < 0:
         raise ValueError(f"Batch index {batch_idx=} shouldn't be less than zero for {path_key}, got {annotations=}")
 
+      if path_key in _DEEPSEEK_V4_CACHE_KEYS:
+        # Copy these states by explicitly overwriting the target slots matching current request id
+        for slot in slots:
+          full_cache = jax.lax.dynamic_update_index_in_dim(full_cache, partial_cache, slot, batch_idx)
+        return full_cache
+
       for slot in slots:
         if path_key == "cache_ar_segment_id":
           ### goal: zero this out in case there is existing data
@@ -1609,6 +1625,10 @@ class MaxEngine(_BaseEngine):  # pyrefly: ignore[invalid-inheritance]
 
       if batch_idx < 0:
         raise ValueError(f"Batch index {batch_idx=} shouldn't be less than zero for {path_key}, got {annotations=}")
+
+      if path_key in _DEEPSEEK_V4_CACHE_KEYS:
+        # Copy these states by explicitly overwriting the target slot matching current request id
+        return jax.lax.dynamic_update_index_in_dim(full_cache, partial_cache, slot, batch_idx)
 
       if path_key == "cache_ar_segment_id":
         s = list(full_cache.shape)
@@ -1743,6 +1763,10 @@ class MaxEngine(_BaseEngine):  # pyrefly: ignore[invalid-inheritance]
 
       if batch_idx < 0:
         raise ValueError(f"Batch index {batch_idx=} shouldn't be less than zero for {path_key}, got {annotations=}")
+
+      if path_key in _DEEPSEEK_V4_CACHE_KEYS:
+        # Direct batch slot index overwrite for fixed-size metadata trackers
+        return jax.lax.dynamic_update_index_in_dim(full_cache, partial_cache, slot, batch_idx)
 
       if path_key == "cache_ar_segment_id":
         ### goal: zero this out in case there is existing data
--- a/src/maxtext/layers/attention_compressed.py
+++ b/src/maxtext/layers/attention_compressed.py
@@ -15,12 +15,14 @@
 """Compressed Attention Layer (DeepSeek-V4)."""
 
 
+import enum
 from typing import Any, Optional, Tuple
 
 import jax
 import jax.numpy as jnp
 from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
+from maxtext.utils import max_utils
 
 from flax import nnx
 
@@ -29,6 +31,7 @@ from maxtext.common.common_types import (
     Config,
     DType,
     MODEL_MODE_TRAIN,
+    MODEL_MODE_AUTOREGRESSIVE,
     AttentionType,
     DEFAULT_MASK_VALUE,
 )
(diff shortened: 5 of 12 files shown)


entrpn (approved):

src/maxtext/configs/models/deepseek4-tiny.yml
```diff
@@ -11,16 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# Test Model config for DeepSeek-V4-Flash 284B (https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash)
 
-base_emb_dim: 4096
-base_num_query_heads: 64
+# Model config for DeepSeek-V4-Flash 284B (https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash)
```
parambole: @Rohan-Bierneni - remove this line and add a comment around how the tiny config was chosen.

src/maxtext/configs/models/deepseek4-tiny.yml
```diff
@@ -11,16 +11,17 @@
...
-base_mlp_dim: 2048
-base_moe_mlp_dim: 2048
-vocab_size: 129280
-head_dim: 512
+base_mlp_dim: 512
+base_moe_mlp_dim: 512
+vocab_size: 32000
+head_dim: 64
 
 # --- Standard Defaults ---
```
parambole, line 25: Rephrase this as it's not the standard defaults.

src/maxtext/configs/base.yml
```diff
@@ -524,7 +524,7 @@ logical_axis_rules: [
                       # Attention Activations
                       ['activation_batch_attn', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_heads', ['tensor', 'tensor_sequence', 'autoregressive']],
-                      ['activation_kv_heads', ['tensor', 'tensor_sequence']],
```
parambole, line 533: @Rohan-Bierneni - Why have we removed these configs ?

src/maxtext/configs/base.yml
```diff
@@ -535,7 +535,7 @@ logical_axis_rules: [
                       # Attention Weights
                       ['heads', ['tensor', 'tensor_sequence', 'autoregressive']],
                       ['q_heads', ['tensor', 'tensor_sequence', 'autoregressive']],
-                      ['kv_heads', ['tensor', 'tensor_sequence', 'autoregressive']],
```
parambole, line 544: @Rohan-Bierneni - Why have we removed these configs ?

src/maxtext/configs/base.yml
```diff
@@ -594,8 +594,8 @@ logical_axis_rules: [
                       ['activation_prefill_kv_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['decode_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['decode_length', []],
-                      ['cache_heads', ['autoregressive', 'tensor', 'tensor_sequence']],
```
parambole, line 603: @Rohan-Bierneni - Why have we removed these configs ?

src/maxtext/inference/maxengine/maxengine.py
```diff
@@ -1492,6 +1502,12 @@ def copy(path, partial_cache, full_cache, annotations):
       if batch_idx < 0:
         raise ValueError(f"Batch index {batch_idx=} shouldn't be less than zero for {path_key}, got {annotations=}")
 
+      if path_key in _DEEPSEEK_V4_CACHE_KEYS:
+        # Copy these states by explicitly overwriting the target slots matching current request id
+        for slot in slots:
+          full_cache = jax.lax.dynamic_update_index_in_dim(full_cache, partial_cache, slot, batch_idx)
+        return full_cache
+
```
parambole, line 1513: @Rohan-Bierneni - Since the batch id is known can we avoid the for loop and write something like this ?

```
if path_key in _DEEPSEEK_V4_CACHE_KEYS:
    # If batch_idx is always axis 0:
    return full_cache.at[slots, ...].set(partial_cache)
```

If not lets use more of a jax primitive to do this

```
if path_key in _DEEPSEEK_V4_CACHE_KEYS:
     # 2. Perform a vectorized, JAX-idiomatic update
    return full_cache.at[tuple(idx)].set(partial_cache)
```

src/maxtext/inference/kvcache.py
```diff
@@ -924,6 +963,124 @@ def kv_cache_autoregressive(
...
+    else:
+      gate_weights = jax.nn.softmax(buffer_gate, axis=1).astype(buffer_kv.dtype)
+      compressed_block = jnp.sum(buffer_kv * gate_weights, axis=1, keepdims=True)
+
+    potential_update_key = compressed_block
+
+    operand_shape = self.cached_ar_key.get_value().shape
+    if potential_update_key.shape[3] < operand_shape[3]:
+      pad_amt = operand_shape[3] - potential_update_key.shape[3]
+      potential_update_key = jnp.pad(potential_update_key, ((0, 0), (0, 0), (0, 0), (0, pad_amt)))
```
parambole, line 1027: @Rohan-Bierneni - Can you elaborate on under which scenarios do we need padding ?

src/maxtext/inference/kvcache.py
```diff
@@ -924,6 +963,124 @@ def kv_cache_autoregressive(
...
+    (
+        cached_ar_key_vars,
+        cached_ar_value_vars,
+        cached_ar_segment_id_var,
+        _,
+        cache_ar_lengths_var,
+    ) = self._get_ar_cache_vars()
+    ar_index = self.cache_ar_index.get_value()
+
+    # 4. Unconditionally write to the cache array
```
parambole: nit: Remove Unconditionally

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
-) -> Array:
+def csa_overlap_pooling(chunk_kv, chunk_gate, kv_norm, head_dim, prior_kv, prior_gate):
   """Shared utility for Compressed Sparse Attention (CSA) overlap pooling.
 
-  Implements the overlapping Ca/Cb pooling logic shared by both the CSA Compressor
-  and the CSA Indexer. It splits the projected states into two halves (Ca and Cb),
-  shifts the first half forward by one window, and concatenates them to form
-  overlapping windows over which softmax gating is applied.
+  Implements overlapping Ca/Cb pooling logic shared by both the CSA Compressor
+  and the CSA Indexer.
```
parambole: @Rohan-Bierneni - Can you add back the comments and fix the formatting  ?

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
 
   Args:
-    hidden_states: Input token embeddings. Shape: `[batch, seq_len, emb_dim]`.
-    kv_proj: Dense layer projecting to `2 * head_dim`.
-    gate_proj: Dense layer projecting to `2 * head_dim`.
-    position_bias: Bias tensor. Shape: `[compress_rate, 2 * head_dim]`.
-    kv_norm: RMSNorm instance.
-    compress_rate: Compression rate for CSA.
-    head_dim: Standard head dimension.
```
parambole, line 69: @Rohan-Bierneni - Please add back the removed description.

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
   Returns:
-    compressed: The pooled overlapping states. Shape: `[batch, n_windows, head_dim]`.
-
-  Shape Transformations:
-    1. Projections: `[batch, seq_len, emb_dim]` -> `[batch, seq_len, 2 * head_dim]`
-    2. Reshape: -> `[batch, n_windows, compress_rate, 2 * head_dim]`
-    3. Split: -> 2x `[batch, n_windows, compress_rate, head_dim]`
-    4. Shift: Ca shifted forward by one window.
-    5. Concat (Ca + Cb): -> `[batch, n_windows, 2 * compress_rate, head_dim]`
-    6. Gating & Sum: -> `[batch, n_windows, head_dim]`
```
parambole, line 80: @Rohan-Bierneni  - Please put back the comments.

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
-  )
-  a_gate_shifted = jnp.concatenate(
-      [
-          jnp.full((batch_size, 1, compress_rate, head_dim), -jnp.inf, dtype=a_gate.dtype),
-          a_gate[:, :-1],
-      ],
-      axis=1,
-  )
-
-  # Concatenate shifted Ca and unshifted Cb to form the final overlapping window
```
parambole, line 125: @Rohan-Bierneni - Can you elaborate on why these changes are required ?

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
     """
     batch_size, seq_len, _ = hidden_states.shape
-
-    # Project hidden states to KV and Gate components
-    # [batch, seq_len, emb_dim] -> [batch, seq_len, head_dim]
     kv = self.kv_proj(hidden_states)
-    # [batch, seq_len, emb_dim] -> [batch, seq_len, head_dim]
     gate = self.gate_proj(hidden_states)
 
-    # Truncate sequence to the nearest multiple of the compression rate
```
parambole, line 288: Please add back the comments.

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
+    # --- AUTOREGRESSIVE DELEGATION ---
```
parambole: Rephrase the comment.

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
+      # Subtract ar_valid to get the constant, unchanging number of true prefill blocks!
+      prefill_blocks_count = cache.entry_count.get_value() - ar_valid  # [B, 1]
+
+      entry_indices = jnp.arange(compressed_kv.shape[1])[None, None, None, :]  # [1, 1, 1, L]
+
+      is_prefill = entry_indices < max_prefill_comp
+      is_valid_prefill = is_prefill & (entry_indices < jnp.expand_dims(prefill_blocks_count, axis=(1, 3)))
+
+      is_ar = entry_indices >= max_prefill_comp
+      is_valid_ar = is_ar & ((entry_indices - max_prefill_comp) < jnp.expand_dims(ar_valid, axis=(1, 3)))
```
parambole: @Rohan-Bierneni - We should pull this out in a separate helper method.

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
     else:
-      # Provide an empty tensor when the sequence is shorter than the compression rate
       compressed = jnp.zeros((batch_size, 0, self.head_dim), dtype=self.dtype)
 
-    # Expand the feature dimension to match the standard KV projection shape
-    # -> [batch, n_windows, 1, head_dim]
     compressed_kv = jnp.expand_dims(compressed, axis=2)
     compressed_len = compressed_kv.shape[1]
 
-    # Skip causal mask generation during decoding (seq_len == 1) or if no blocks were pooled
```
parambole, line 323: Please add back the comments.

src/maxtext/layers/attention_op.py
```diff
@@ -816,32 +816,61 @@ def generate_attention_mask(
...
+          output_mask = uncompressed_mask_5d * output_mask
+        else:
+          output_mask = uncompressed_mask_5d
+        return jnp.where(output_mask, 0.0, DEFAULT_MASK_VALUE)
 
-      # Broadcast uncompressed_mask to match compressed_mask's layout
       target_shape = compressed_mask.shape[:-1] + (s_len,)
-      padded_shape = (1,) * (len(target_shape) - 2) + uncompressed_mask.shape
-      uncompressed_mask = jnp.broadcast_to(uncompressed_mask.reshape(padded_shape), target_shape)
+      uncompressed_mask_5d = jnp.broadcast_to(uncompressed_mask_5d, target_shape)
```
parambole: @Rohan-Bierneni - Can you rename the variable and avoid _5d type naming?

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
-      ],
-      axis=1,
-  )
-
-  # Concatenate shifted Ca and unshifted Cb to form the final overlapping window
+  # 2. Safely handle cache priors
+  if prior_kv is None or prior_kv.size != expected_size:
+    prior_a_kv = jnp.zeros((B, 1, C, head_dim), dtype=a_kv.dtype)
+    # Note: Empty gate prior must be -inf so softmax(gate) = 0
+    prior_a_gate = jnp.full((B, 1, C, head_dim), -jnp.inf, dtype=a_gate.dtype)
```
parambole: @Rohan-Bierneni 

>(B, 1, C, head_dim)

a) For W=1 add an enum and use that everywhere rather than hardcoding it. 
b) Also add comments around it as well

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
+  else:
+    prior_kv = prior_kv.reshape((B, 1, C, D2))
+    prior_gate = prior_gate.reshape((B, 1, C, D2))
+    prior_a_kv, _ = jnp.split(prior_kv, 2, axis=-1)
+    prior_a_gate, _ = jnp.split(prior_gate, 2, axis=-1)
+
+    # KVCache initializes with zeros. If it's the very first step,
+    # gate must be -inf for softmax to equal 0.
+    is_empty = jnp.all(prior_a_kv == 0)
+    prior_a_gate = jnp.where(is_empty, -jnp.inf, prior_a_gate)
```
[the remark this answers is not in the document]
Rohan-Bierneni: This bug only triggers in a staggered / continuous batching (in-flight batching) regime where individual slots are reset independently while the rest of the batch continues decoding. 

However, maxengine/jetstream does synchronous batching meaning that we will not hit this bug.

src/maxtext/layers/attention_compressed.py
```diff
@@ -41,98 +43,75 @@
...
+    prior_a_gate = jnp.full((B, 1, C, head_dim), -jnp.inf, dtype=a_gate.dtype)
+  else:
+    prior_kv = prior_kv.reshape((B, 1, C, D2))
+    prior_gate = prior_gate.reshape((B, 1, C, D2))
+    prior_a_kv, _ = jnp.split(prior_kv, 2, axis=-1)
+    prior_a_gate, _ = jnp.split(prior_gate, 2, axis=-1)
+
+    # KVCache initializes with zeros. If it's the very first step,
+    # gate must be -inf for softmax to equal 0.
+    is_empty = jnp.all(prior_a_kv == 0)
```
parambole, line 119: Pass an integer step/token counter or check emptiness per-sequence instead of using a global jnp.all(...) == 0 ?

benchmarks/api_server/maxtext_server.py
```diff
@@ -322,9 +333,17 @@ def _prepare_batch_for_broadcast(batched_items):
...
-    all_prompts.extend(prompts_for_req)
-    request_info_map.append((req_id, req, is_chat, len(prompts_for_req)))
+    try:
+      prompts_for_req = server_utils.get_prompts_for_request(req, LLM)
+      all_prompts.extend(prompts_for_req)
+      request_info_map.append((req_id, req, is_chat, len(prompts_for_req)))
+    except ValueError as e:
+      logger.error("Failed to parse request %s: %s", req_id, e)
+      with response_lock:
+        response_dict[req_id] = {"error": f"Invalid request format or missing chat template: {e}"}
```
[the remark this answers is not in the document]
Rohan-Bierneni: Will not make changes here as it is copied from the HF repo directly.

src/maxtext/layers/attention_compressed.py
```diff
@@ -453,90 +516,226 @@ def __call__(
...
+        chunk_gate_reshaped = jnp.swapaxes(buf_gate, 1, 2) + self.position_bias.value
 
-      compressed = self.rotary_emb(compressed, positions, unsqueeze_dim=None)
-    else:
-      # Return empty top-k selections when sequence is too short to form any windows
-      return jnp.zeros(
-          (batch_size, seq_len, min(self.index_topk, compressed_len)),
-          dtype=jnp.int32,
+        prior_kv = jnp.transpose(cache.overlap_kv.get_value(), (0, 2, 1, 3))
+        prior_gate = jnp.transpose(cache.overlap_gate.get_value(), (0, 2, 1, 3))
```
parambole: @Rohan-Bierneni - All of these transpose seem unnecessary can we instead change the caching

```
In src/maxtext/inference/kvcache.py (KVCache.__init__), swap key_heads and compress_rate in the zero shape:

self.overlap_kv = nnx.Cache(
    jnp.zeros((self.batch, self.key_heads, self.compress_rate, self.key_head_size), dtype=dtype),
    out_sharding=(cache_batch_axis_name, None, None, None),
)
self.overlap_gate = nnx.Cache(
    jnp.zeros((self.batch, self.key_heads, self.compress_rate, self.key_head_size), dtype=dtype),
    out_sharding=(cache_batch_axis_name, None, None, None),
)
```

Here and else where.

src/maxtext/configs/models/deepseek4-tiny.yml
Rohan-Bierneni: I am reverting the changes made to this file. Previously I made changes so that I could iterate faster on cpu vm but the previous tiny.yml configs are better for testing on tpu vm since it is closer to actual model configs.

src/maxtext/layers/attention_compressed.py
```diff
@@ -601,58 +800,171 @@ def __call__(
...
+
+        positions = jnp.arange(compressed_len) * self.compress_rate + position_ids[:, 0:1]
+        compressed = self.rotary_emb(compressed, positions, unsqueeze_dim=None)
+      else:
+        compressed = jnp.zeros((batch_size, 0, self.head_dim), dtype=self.dtype)
+        compressed_len = 0
+        next_prior_kv = prior_kv
+        next_prior_gate = prior_gate
+
+      compressed_kv = jnp.expand_dims(compressed, 2)
```
parambole: @Rohan-Bierneni these blocks are used multiple times across the code base. 

We can mvoe them to `src/maxtext/inference/kvcache.py and paste this helper function inside class KVCache(BaseCache):` ?

src/maxtext/configs/models/deepseek4-tiny.yml
Rohan-Bierneni: As mentioned above reverting the changes made to this file.

src/maxtext/configs/base.yml
Rohan-Bierneni: For testing, replicating the kv heads across all devices allows me to increase the TP value upto 64, instead of being limited to TP=2. This allowed me to test a global batch of 1 on TPU cluster for debugging the batching logic. 

However, this is not needed anymore and I will put a comment in the bug for anyone who wants to repro and debug with a similar setup as this.

Reverting these changes.

src/maxtext/configs/base.yml
Rohan-Bierneni: Same reponse as from comment above:

"
For testing, replicating the kv heads across all devices allows me to increase the TP value upto 64, instead of being limited to TP=2. This allowed me to test a global batch of 1 on TPU cluster for debugging the batching logic. 

However, this is not needed anymore and I will put a comment in the bug for anyone who wants to repro and debug with a similar setup as this.

Reverting these changes.
"

src/maxtext/configs/base.yml
Rohan-Bierneni: Same reponse as from comment above:

"
For testing, replicating the kv heads across all devices allows me to increase the TP value upto 64, instead of being limited to TP=2. This allowed me to test a global batch of 1 on TPU cluster for debugging the batching logic. 

However, this is not needed anymore and I will put a comment in the bug for anyone who wants to repro and debug with a similar setup as this.

Reverting these changes.
"

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
+          update_blocks = jnp.pad(update_blocks, ((0, 0), (0, 0), (0, 0), (0, pad_amt)))
+
+        operand = cache_key_var.get_value()
+        batch_axis = cache.prefill_cache_axis_order.index(0)
+
+        # Broadcast batch=1 prefill requests to fit the max_concurrent_decodes cache
+        if operand.shape[batch_axis] != update_blocks.shape[batch_axis]:
+          repeats = operand.shape[batch_axis] // update_blocks.shape[batch_axis]
+          update_blocks = jnp.repeat(update_blocks, repeats, axis=batch_axis)
+
```
parambole: We should add a comment over here and end where that we are copying the same batch multiple times.

src/maxtext/layers/attention_op.py
```diff
@@ -816,32 +816,61 @@ def generate_attention_mask(
...
+
+            # Safely trace AR cache vs Prefill cache without triggering a ConcretizationTypeError
+            is_ar_cache = jnp.max(max_valid, axis=(0, 1), keepdims=True) >= 0
+
+            i = jnp.arange(s_len)[None, None, :]
+            abs_k_ar = abs_q - max_valid[:, :, None] + i
+            abs_k_prefill = jnp.broadcast_to(i, abs_k_ar.shape)
+
+            abs_k = jnp.where(is_ar_cache, abs_k_ar, abs_k_prefill)
+            distance = abs_q - abs_k
```
parambole, line 849: @Rohan-Bierneni Do not reduce across axis=(0, 1). Let each sequence in the batch determine its own position using its own max_valid slice per row ?

src/maxtext/layers/attention_op.py
```diff
@@ -2257,19 +2287,25 @@ def __call__(
     if model_mode != MODEL_MODE_TRAIN:
       assert prefill_kv_cache
       key, value, decoder_segment_ids = prefill_kv_cache
+    pass_comp_to_prefill = (compressed_kv is not None) and (ar_kv_cache is None)
+    k_prefill = key
+    v_prefill = value
+    if pass_comp_to_prefill:
+      k_prefill = jnp.concatenate([k_prefill, compressed_kv], axis=1)
+      v_prefill = jnp.concatenate([v_prefill, compressed_kv], axis=1)
```
parambole, line 2316: Instead of copying and concatenating the entire max_target_length cache tensor with compressed_kv in HBM on every decode step, either:

Concatenate only along the active valid sequence slice rather than the entire static cache buffer, or keep compressed_kv inside KVCache so it is already in place without runtime concatenation ?

benchmarks/api_server/test_batching_chat_completions.py
```diff
@@ -0,0 +1,73 @@
+"""Client test script for verifying concurrent chat completion batching against the MaxText API server."""
```
shuningjin, line 15: Need license in the head. Same for the other test file.

src/maxtext/inference/maxengine/maxengine.py
Rohan-Bierneni: We should keep `jax.lax.dynamic_update_index_in_dim` here rather than using `.at[slots, ...].set(partial_cache)`. 
1. `dynamic_update_index_in_dim` maps directly to XLA's `DynamicUpdateSlice` instruction, which guarantees in-place buffer donation without HBM copies. `.at[slots, ...].set(...)` compiles to an XLA `Scatter`, which can break buffer donation and increase peak memory footprint.
2. This also maintains consistency with the rest of `bulk_insert`, where all standard KV cache tensors (`cached_prefill_key`, `cached_ar_key`, etc.) use this exact same `for slot in slots: dynamic_update_index_in_dim` pattern.

tests/unit/train_compile_test.py
```diff
@@ -1229,3 +1229,29 @@ def test_vocab_tiling_bf16_nnx(self):
             "pure_nnx_decoder=true",
         )
     )
+
+  @pytest.mark.cpu_only
+  def test_dsv4(self):
+    """AOT test for deepseek4"""
+    compiled_trainstep_file = "/tmp/test_dsv4"
```
shuningjin: We already have DS-V4 AOT test. Is it any different?
https://github.com/AI-Hypercomputer/maxtext/blob/63fffd4458f689ab5c5bf05fd45b8bdd59ddfc1b/tests/unit/train_compile_test.py#L862-L888

parambole (commented):
Hey @Rohan-Bierneni thank you for making the changes. I have left a few comments around
1. Putting back the docustrings
2. Refactoring the caching code in attention_compressed and ops. 

PTAL

src/maxtext/inference/kvcache.py
Rohan-Bierneni: We need padding for DeepSeek-V4 CSA (compress_ratio == 4) and the Indexer due to a 2x size difference between un-pooled tokens and pooled latent blocks:

#### KVCache size (2 * head_dim):
kv_proj outputs 2 * head_dim (split into C_a and C_b for overlapping pooling). KVCache is initialized to this larger size so buffers like leftover_buffer_kv can store the full un-pooled projections.

#### **Compressed block size (head_dim):**

When a 4-token window completes, compressor_fn pools them down into a single latent block of size head_dim.

**Why we pad:** cached_ar_key shares the KVCache's 2 * head_dim allocation. Writing the smaller head_dim compressed block into it requires zero-padding the trailing coordinates.

**How it is unpacked:** When read back during attention (attention_compressed.py:L293), we cleanly strip the padded zeros by slicing: compressed_full[:batch_size, :, 0, :comp_dim].

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: These changes support Autoregressive (AR) Mode by preserving overlapping window math across step and chunk boundaries:

#### The Bug

The original csa_overlap_pooling hardcoded the prior window (C_a) to zeros/-inf. This broke AR decoding and chunked prefill because every new block treated its preceding window as empty, corrupting the overlap history.

#### The Solution

We moved projections (kv_proj, gate_proj) and chunking logic to the callers. In AR mode, chunks now correctly come from accumulated window buffers rather than raw, unchunked states.

We introduced prior inputs and outputs to track overlap history.

  - Inputs (prior_kv, prior_gate): We pass None for Step 0 (defaults to zeros/-inf), and pass the cached overlaps for t > 0 to seamlessly continue the pooling from preceding tokens.
  - Outputs (next_prior_kv, next_prior_gate): Returns the trailing window so callers can update cache.overlap_kv and cache.overlap_gate for subsequent AR steps.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_op.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Added a CSAPoolingConfig(enum.IntEnum) with OVERLAP_WINDOWS = 1 and replaced all hardcoded 1s across csa_overlap_pooling along with explanatory comments. Thanks!

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,70 +243,154 @@ def __call__(
...
       compressed = jnp.zeros((batch_size, 0, self.head_dim), dtype=self.dtype)
 
-    # Expand the feature dimension to match the standard KV projection shape
-    # -> [batch, n_windows, 1, head_dim]
     compressed_kv = jnp.expand_dims(compressed, axis=2)
     compressed_len = compressed_kv.shape[1]
 
-    # Skip causal mask generation during decoding (seq_len == 1) or if no blocks were pooled
+    # --- PREFILL CACHE PRIMING ---
+    if cache is not None:
```
shuningjin, line 623: This block appear twice (CSA, HCA). can you wrap into function and reuse?
https://screenshot-v2.corp.google.com/AZTFZ7pRLUHwk8R

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Since maxengine uses synchronous batching, we can use jnp.all() to check for initial step instead of doing it be per sequence which should be a feature of asynchronous batching. 

For now, I will skip implementing this as jetstream is deprecated and don't want to introduce other points of failure as the purpose of this pr is to enable functionality for decoding.

We can revisit this or improve the perf in a follow up pr.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Swapped key_heads and compress_rate in KVCache.__init__ for overlap_kv and overlap_gate so their shape is (batch, key_heads, compress_rate, key_head_size). Removed all jnp.transpose(..., (0, 2, 1, 3)) calls across the CSA compressor and indexer. 

This should improve the perf of decoding by a good amount. Thanks!

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Resolved.

src/maxtext/layers/attention_op.py
Rohan-Bierneni: Removed the reduction across axis=(0, 1) and replaced it with ```is_ar_cache = max_valid[:, :, None] >= 0 ([batch, 1, 1])```.

src/maxtext/layers/attention_op.py
Rohan-Bierneni: This concatenation actually does not run during autoregressive steps—it only runs during prefill.
Notice that `pass_comp_to_prefill = (compressed_kv is not None) and (ar_kv_cache is None)`. 

During autoregressive mode, `ar_kv_cache` is not `None` (as `kv_cache_autoregressive_v4` returns the AR cache tuple), so `pass_comp_to_prefill` evaluates to `False`.

Furthermore, for autoregressive decoding we already follow your suggestion: `kv_cache_autoregressive_v4` writes newly compressed blocks directly into `KVCache` (`ar_key_vars` / `ar_value_vars`), so `compressed_kv` is already in place in `ar_kv_cache` without any runtime HBM concatenation during decode steps.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Pulled out the AR validity mask computation into a reusable top-level helper function update_ar_cache_and_get_validity_mask(...) and replaced the duplicate masking blocks across the HCA compressor and indexer.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Pulled out the prefill chunking, prior fetching, overlap pooling, and RoPE embedding logic into a shared helper function compute_csa_prefill_chunk_pooling.

However, since this function is specific to CSA/Indexer classes, we shouldn't add it to KVCache.py as then it would apply to all models instantiated with a KVCache object.

Future models that only use comressed attention will still be supported with this change though

tests/unit/train_compile_test.py
Rohan-Bierneni: Removing this as I think I had it added before testing on v5p-512.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Actually this logic happens across all three classes (HCACompressor, CSACompressor, and CSAIndexer).

Replaced it now with a single call to prime_prefill_cache_state() that should account for the minor differences in logic between CSA & HCA

benchmarks/api_server/test_batching_chat_completions.py
Rohan-Bierneni: Resolved.

shuningjin (approved):
Thank you!

src/maxtext/layers/attention_compressed.py
```diff
@@ -453,90 +561,164 @@ def __call__(
...
-    # Broadcast the compressed KV representations across all indexer heads
-    # -> [batch, 1, n_windows, index_head_dim]
+      comp_dim = self.index_head_dim
+      compressed_full = jnp.concatenate([cached_prefill[0], cached_ar[0]], axis=1)
+
+      # Restore back to original batch_size and strip RoPE/dummy head
+      compressed = compressed_full[:batch_size, :, 0, :comp_dim]
+      compressed_len = compressed.shape[1]
+
+      # --- NEW VALIDITY MASK FOR AR MODE ---
```
parambole: @Rohan-Bierneni nit: can you please elaborate on the AR mode here ?

src/maxtext/layers/attention_compressed.py
```diff
@@ -1021,85 +1329,94 @@ def __call__(
...
+          compressed_segment_mask,
+          model_mode,
+          self.compressor_cache,
+          self.indexer_cache,
+      )
 
     # Apply segment masking to the compressed blocks
     if compressed_segment_mask is not None and compressed_mask is not None:
-      # compressed_segment_mask is [batch, q_len, num_compressed_blocks]
-      # compressed_mask is [batch, 1, q_len, num_compressed_blocks]
```
parambole, line 1050: @Rohan-Bierneni - Please put back comments

src/maxtext/layers/attention_compressed.py
```diff
@@ -1021,85 +1329,94 @@ def __call__(
...
       compressed_mask = compressed_mask + jnp.expand_dims(
           compressed_segment_mask[:, :, : compressed_mask.shape[-1]], axis=1
       )
 
-    # Extend local KV tensors with the compressed blocks
-    if compressed_kv is not None:
-      kv = jnp.concatenate([kv, compressed_kv], axis=1)
-
```
parambole, line 1058: @Rohan-Bierneni - Why was this blocked removed ?

src/maxtext/layers/attention_compressed.py
```diff
@@ -1021,85 +1329,94 @@ def __call__(
...
     kv = checkpoint_name(kv, "kv_proj")
 
-    # Prepare the mask shape for the underlying AttentionOp
```
parambole, line 1061: Please put back comment

src/maxtext/layers/attention_compressed.py
```diff
@@ -1021,85 +1329,94 @@ def __call__(
...
     if compressed_mask is not None:
       compressed_mask = jnp.expand_dims(compressed_mask, axis=2)
 
-    # Scale queries if a pre-attention scalar is defined
```
parambole, line 1065: Please put back comment

src/maxtext/layers/attention_compressed.py
```diff
@@ -1021,85 +1329,94 @@ def __call__(
...
     grouped_out = attn_out.reshape(b, s, self.config.o_groups, (h * d) // self.config.o_groups)
-
-    # -> [batch, q_length, o_groups, out_features_per_group]
     grouped_out = self.o_a_proj(grouped_out)
-
-    # -> [batch, q_length, o_groups * out_features_per_group]
     grouped_flat = grouped_out.reshape(b, s, -1)
-
-    # -> [batch, q_length, emb_dim]
     final_out = self.o_b_proj(grouped_flat)
```
parambole, line 1577: @Rohan-Bierneni - Please put back comments

src/maxtext/layers/attention_op.py
```diff
@@ -816,32 +816,74 @@ def generate_attention_mask(
         sliding_mask = sliding_mask[:, None, None, :, :]
       output_mask = sliding_mask * output_mask
     elif self.attention_type == AttentionType.COMPRESSED:
-      if compressed_mask is None:
-        raise ValueError("compressed_mask must be provided for COMPRESSED attention type")
```
parambole, line 825: @Rohan-Bierneni - Isn't this check required for future consumers who are not DSV4 ?

src/maxtext/layers/attention_compressed.py
```diff
@@ -264,34 +319,78 @@ def __call__(
...
+      model_mode: The execution mode (e.g. train, prefill, or autoregressive).
+      cache: Optional KV cache instance used during inference.
 
     Returns:
       compressed_kv: The pooled KV tensors. Shape: `[batch, n_windows, 1, head_dim]`.
       compressed_causal_mask: Causal mask preventing queries from seeing future blocks.
                               Shape: `[batch, 1, seq_len, n_windows]`.
     """
     batch_size, seq_len, _ = hidden_states.shape
-
```
parambole, line 281: nit: Add new line

src/maxtext/layers/attention_compressed.py
```diff
@@ -954,7 +1255,12 @@ def compressed_query_projection(self, inputs_q: Array, inputs_positions: Array,
     q_up_normed = self.q_up_norm(q_up)
 
     # -> [batch, seq_len, num_query_heads, head_dim]
-    q_out = self.rotary_embedding(q_up_normed, inputs_positions, unsqueeze_dim=-2)
+    try:
+      q_out = self.rotary_embedding(q_up_normed, inputs_positions, unsqueeze_dim=-2)
+    except TypeError:
+      # If the embedding rejects the kwarg, reshape manually before passing it in,
+      # or rely on the embedding's internal broadcasting.
+      q_out = self.rotary_embedding(q_up_normed, inputs_positions)
```
parambole: @Rohan-Bierneni - Can we avoid using `try ... except TypeError:` for optional kwargs here and in `kv_out`/`attn_out`?
**Problem:**  
Catching a broad `TypeError` to test method signatures is fragile—it can silently swallow legitimate runtime bugs (such as shape/dtype mismatches or internal tensor errors) and adds unnecessary exception-handling clutter during JAX tracing.
**Potential Solution:**  
Instead of relying on runtime exception trapping for polymorphic dispatch, we should either check the embedding type/capability explicitly before calling (e.g., via an `isinstance` or attribute check), or unify the base `RotaryEmbedding` signature so it safely accepts/ignores optional kwargs like `unsqueeze_dim` and `reverse`.

parambole (commented):
@Rohan-Bierneni - I have left a few comments PTAL

src/maxtext/layers/attention_op.py
Rohan-Bierneni: Removing the ValueError is intentional because in AttentionType.COMPRESSED architectures (both DSV4 and future hybrid compressed/sliding-window models), compressed_mask can be None.

Specifically:

Short Sequences & Early Prefill: When a sequence is shorter than compress_rate (or during early prefill steps before any full compression windows can be pooled), there are 0 compressed blocks.

Sliding-Window-Only Fallback: When compressed_mask is None, the layer should gracefully fall back to attending only over the local uncompressed sliding window (s_len = kv_seq_len), which is handled by:


```
if compressed_mask is None:
  if output_mask is not None:
    output_mask = expanded_uncompressed_mask * output_mask
  else:
    output_mask = expanded_uncompressed_mask
  return jnp.where(output_mask, 0.0, DEFAULT_MASK_VALUE)
```

Requiring compressed_mask to never be None would break valid execution for any model using compressed attention whenever a request has fewer tokens than the compression window size. Falling back to the local sliding window mask is the correct general behavior for any consumer of AttentionType.COMPRESSED.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: The concatenation was moved inside AttentionOp.__call__ ():

For prefill:
```
    if pass_comp_to_prefill:
      k_prefill = jnp.concatenate([k_prefill, compressed_kv], axis=1)
      v_prefill = jnp.concatenate([v_prefill, compressed_kv], axis=1)
```

For autoregressive decoding:
```
    if compressed_kv is not None:
      key_ar = jnp.concatenate([key_ar, compressed_kv], axis=1)
      value_ar = jnp.concatenate([value_ar, compressed_kv], axis=1)
```


Why we moved it:

In autoregressive decoding with caching, prefill_kv_cache and ar_kv_cache are managed separately. 

Concatenating compressed_kv onto kv at the outer layer (DeepseekV4Attention) would mix uncompressed local sliding-window tokens and compressed KV blocks into a single tensor before caching/kernel execution.

By passing compressed_kv explicitly as a keyword argument (attention_op(..., compressed_kv=compressed_kv)), AttentionOp can correctly append compressed_kv to k_prefill/v_prefill during prefill, or append it to key_ar/value_ar during autoregressive decoding, while keeping the physical cache structures clean.

src/maxtext/layers/attention_compressed.py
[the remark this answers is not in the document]
Rohan-Bierneni: This comment is incorrect regarding the input shapes. `cached_ar[3]` returned by the cache is a 1D array of shape `[batch]`, not 2D `[batch, 1]`.
Therefore, `ar_valid = jnp.expand_dims(cached_ar[3], axis=1)` produces `[batch, 1]`, matching `cache.entry_count.get_value()`. Subtraction produces `[batch, 1]` without any `[batch, batch, 1]` broadcasting, and `jnp.expand_dims(..., axis=(1, 3))` cleanly produces `[batch, 1, 1, 1]` for 4D mask broadcasting against `[1, 1, 1, L]`.

src/maxtext/layers/attention_compressed.py
```diff
@@ -453,90 +516,226 @@ def __call__(
...
+      # Calculate constant true prefill blocks
+      prefill_blocks_count = cache.entry_count.get_value() - ar_valid  # [B, 1]
+
+      entry_indices = jnp.arange(compressed_len)[None, None, :]  # [1, 1, L]
+
+      is_prefill = entry_indices < max_prefill_comp
+      is_valid_prefill = is_prefill & (entry_indices < jnp.expand_dims(prefill_blocks_count, axis=1))
+
+      is_ar = entry_indices >= max_prefill_comp
+      is_valid_ar = is_ar & ((entry_indices - max_prefill_comp) < jnp.expand_dims(ar_valid, axis=1))
```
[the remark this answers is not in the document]
Rohan-Bierneni: Incorrect assumption as same response from above comment applies here.

src/maxtext/layers/attention_compressed.py
Rohan-Bierneni: Removed all three `try...except TypeError` blocks and replaced them with an explicit `apply_rotary_embedding` helper method on `DeepseekV4Attention`, following the existing pattern in `attentions.py`. It explicitly checks `isinstance(self.rotary_embedding, DeepSeekV4RotaryEmbedding)` before passing `unsqueeze_dim` and `reverse` kwargs, avoiding exception trapping during JAX tracing.

src/maxtext/inference/kvcache.py
parambole, line 1027: @Rohan-Bierneni 

nit: Can you please add comments around padding 

```
   operand_shape = self.cached_ar_key.get_value().shape
    if potential_update_key.shape[3] < operand_shape[3]:
      # DeepSeek-V4 compressed latent dimension d_c can be smaller than cached key_head_size
      # (which is often padded to power-of-two / systolic array boundaries for TPU efficiency).
      # Pad trailing zeros along head_dim (axis 3) to match cache operand shape.
      pad_amt = operand_shape[3] - potential_update_key.shape[3]
      potential_update_key = jnp.pad(potential_update_key, ((0, 0), (0, 0), (0, 0), (0, pad_amt)))```

**Review by parambole — APPROVED** [review]
LGTM 🚀 great work !

File: src/maxtext/inference/kvcache.py (hunk 7 above)
Comment (Rohan-Bierneni) [reply src/maxtext/inference/kvcache.py:1027]: Resolved.

File: src/maxtext/layers/decoders.py [hunk 40]
```diff
@@ -854,6 +854,8 @@ def __call__(
     mesh = self.mesh
     assert decoder_input_tokens.ndim == 2  # [batch, len]
 
+    decoder_input_tokens = decoder_input_tokens.astype(jnp.int32)
```
shuningjin: Why is this type cast necessary?

shuningjin:
It would be good if we can have unit test for DSV4 decoding. Can be follow up. 

Similar to:

https://github.com/AI-Hypercomputer/maxtext/blob/eb9153df80ef3dba4054e6f3d897bb41f1d2aa19/tests/unit/attention_test.py#L692

https://github.com/AI-Hypercomputer/maxtext/blob/eb9153df80ef3dba4054e6f3d897bb41f1d2aa19/tests/unit/attention_test.py#L1985

https://github.com/AI-Hypercomputer/maxtext/blob/eb9153df80ef3dba4054e6f3d897bb41f1d2aa19/tests/unit/attention_test.py#L2820

src/maxtext/layers/decoders.py
Rohan-Bierneni: Issue was caused by hash routing for the first 3 v4 layers. Moved the typecast to moe.py instead of here.

aireenmei (approved):

RissyRan (approved):
The PR has been reviewed by the team, approval for unblocking.

jrplatin (approved):

PR #4448, opened 2026-07-13
Keep fused MoE weights intact through FSDP gather (1.5% perf improvement on DeepSeek-v3)

abhinavgoel95:
# Description

Preserve prefused MoE input weights through `shard_map` so FSDP performs one all-gather of the combined weight tensor before the grouped GEMM.

Previously, MaxText split the prefused `[experts, embed, 2 * mlp]` weight into separate `w0` and `w1` tensors before `shard_map`. This could cause XLA to all-gather each half independently and concatenate them again immediately before the grouped GEMM.

The updated flow is:

```text
prefused weight
    -> one FSDP all-gather
    -> one grouped GEMM
    -> split output activations for SwiGLU
```

Only the GMM output is split into gate and up-projection activations. This split is required for SwiGLU and is considerably cheaper than splitting, gathering, and concatenating the large weight tensors.

The implementation:

- Passes the prefused weight directly into the sparse MoE `shard_map`.
- Uses one partition specification for the combined weight.
- Preserves the combined weight through one-stage and two-stage FSDP all-gather paths.
- Runs one grouped GEMM using the combined weight.
- Splits the resulting activations before applying the configured FFN activation.
- Preserves existing behavior for quantized serving, weight sparsity, dense MoE computation, and vLLM RPA inference.

The optimization currently applies to sparse MoE training when prefused weights can be consumed directly. Quantized serving and sparsified-weight paths continue using separate `w0` and `w1` tensors. Supporting fused quantized-weight retrieval could be considered as a future improvement.

# Tests

The original experimental change was profiled with a DeepSeek MoE workload and removed the separate weight-slice operations before the grouped GEMM.

The version rebased onto current `main` has not been run end-to-end. The patch was checked with:

```bash
git diff --check
```

To reproduce the performance comparison, run the same sparse MoE workload with:

```text
prefuse_moe_weights=true
sparse_matmul=true
```

Compare the generated HLO or GPU profile before and after this change. The updated version should contain one FSDP all-gather for the combined MoE input weight and should not split and concatenate the weight before the grouped GEMM.

# Checklist


src/MaxText/assets/tokenizers/deepseek-v3-tokenizer/tokenizer.json
gobbleturk, line 1: sync issue? I assume this file should be untouched in this PR, please note if this understanding is not correct

src/maxtext/configs/base.yml
```diff
@@ -432,19 +397,14 @@ qk_nope_head_dim: 128
...
 # QK-Clip (Muon Clip) Configuration
-use_qk_clip: false  # Enable QK-Clip (supported in MLA with DotProduct or Tokamax Splash)
+use_qk_clip: False  # Enable QK-Clip (supported in MLA with DotProduct or Tokamax Splash)
 qk_clip_threshold: 100.0  # Threshold for clipping (tau in the paper)
 
 # Combine matmuls for QKV and MLP
-fused_qkv: false
-fused_mlp: false
+fused_qkv: False
+fused_mla_lora_proj: False  # Fuse MLA Q+KV LoRA up-projections (wq_a+wkv_a) into a single matmul. Requires q_lora_rank > 0.
```
gobbleturk, line 406: can this be put into a separate PR? Keep PRs small for easier readability and reviewing

src/maxtext/configs/types.py
```diff
@@ -87,20 +86,26 @@ class QuantizationType(str, Enum):
...
   INT8 = "int8"
   INTMP = "intmp"
-  FP8_E5M2 = "fp8_e5m2"
-  FP8_E4M3 = "fp8_e4m3"
   FP8 = "fp8"
   NANOO_FP8 = "nanoo_fp8"
   FP8_NANO_V2 = "fp8_nanoo"
   FP8_GPU = "fp8_gpu"
   FP8_FULL = "fp8_full"
+  TE_NO_QUANT = "te_no_quant"
```
gobbleturk, line 94: can this be part of a separate PR? Is this quantization necessary?

gobbleturk: I see maybe this quantization technique has some dependency with fused vs non fused (which is surprising to me), maybe since fused just doesn't support quantization for now?

src/maxtext/trainers/pre_train/train.py
```diff
@@ -630,124 +512,23 @@ def eval_step(model, config, state, data, dropout_rng=None):
...
-  # Pack mutated state back to dicts
-  jax_device_state["state"] = state
-  python_vars["last_step_completion"] = last_step_completion
-
-
 def train_loop(config, recorder, state=None):
   """Main Training loop."""
+  # Initialize HybridEP buffer manager BEFORE setup_train_loop, because model creation
+  # during setup traces shard_map which calls hybrid_ep_dispatch_ffi (needs buffer manager).
+  if config.use_hybrid_ep:
```
gobbleturk, line 525: would prefer to put this in a method (to keep train_loop as lean and readable as possible). Either a method defined somewhere outside of train_loop, or possibly in a new te_utils file

PR #4459, opened 2026-07-14
Keep fused MoE weights intact through FSDP gather

abhinavgoel95:
# Description

Keep prefused MoE input weights intact through the FSDP all-gather, avoiding an unnecessary weight slice/concatenate sequence and improving DeepSeek V3 performance by approximately 1.5% on GB300.

Previously, `prefuse_moe_weights` stored the two MoE input projections as one `[experts, embed, 2 * mlp]` tensor, but the sparse MoE path split this tensor into `w0` and `w1` before entering `shard_map`. This caused FSDP to all-gather the two halves separately, after which they were concatenated again immediately before the grouped GEMM.

This change passes the prefused weight through `shard_map` intact. The sparse MoE path now performs:

1. One FSDP all-gather of the `[experts, embed, 2 * mlp]` weight.
2. One grouped GEMM producing `[tokens, 2 * mlp]`.
3. A split of the activation output immediately before the SwiGLU activation.

This removes the pre-GEMM weight slicing and avoids the second independent FSDP all-gather. The implementation also preserves the fused weight through two-stage FSDP all-gather and applies the appropriate partition specifications to the combined tensor.

The optimization is restricted to sparse matmul with prefused weights, outside quantized serving and weight-sparsity paths. Existing split-weight paths remain unchanged as fallbacks. There is no change to the checkpoint format or default behavior because the optimization is controlled by the existing `prefuse_moe_weights` option.

# Tests

Tested with DeepSeek V3 training on NVIDIA GB300 using:

```yaml
prefuse_moe_weights: true
sparse_matmul: true
```

Compared against the same workload, configuration, and container without this change:

- HLO/profile inspection confirmed that the prefused weight remains intact through FSDP all-gather.
- The unnecessary pre-GEMM weight slice operations were removed.
- The two independent input-weight all-gathers were replaced by one combined all-gather.
- End-to-end training performance improved by approximately 1.5%.
- `git diff --check` passes.

# Checklist

- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files). No documentation changes are required for this internal optimization.

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1370,6 +1370,7 @@ class RoutedMoE(nnx.Module):
       inputs,
       gate_logits,
       pre_bias_logits,
+      wi_kernel,
       w0_kernel,
       w1_kernel,
       wo_kernel,
@@ -1379,6 +1380,11 @@ class RoutedMoE(nnx.Module):
       input_ids=None,
   ):
     """Perform sparse matrix multiplication of inputs and Experts."""
+
+    # Keep prefused [experts, embed, 2 * mlp] weights intact through shard_map
+    # so FSDP gathers the combined tensor once before the grouped GEMM.
+    if self.config.prefuse_moe_weights and wi_kernel is None:
+      raise ValueError("prefuse_moe_weights=True requires a fused wi kernel in sparse_matmul.")
 
     def jax_ragged_dot_gmm(inputs, kernel, tiling, group_sizes, expert_assignments, padding_amount):
       """Execute jax.lax.ragged_dot, with potential quantization"""
@@ -1561,14 +1567,25 @@ class RoutedMoE(nnx.Module):
           return True
       return False
 
-    def maybe_aqt_partition(w0_kernel, w0_pspec, w1_kernel, w1_pspec, wo_kernel, wo_pspec):
+    def maybe_aqt_partition(
+        wi_kernel,
+        wi_pspec,
+        w0_kernel,
+        w0_pspec,
+        w1_kernel,
+        w1_pspec,
+        wo_kernel,
+        wo_pspec,
+    ):
+      if isinstance(wi_kernel, aqt.QTensor):
+        wi_pspec = aqt.partition_spec(wi_pspec, (1,), wi_kernel.dtype, use_bias=False)
       if isinstance(w0_kernel, aqt.QTensor):
         w0_pspec = aqt.partition_spec(w0_pspec, (1,), w0_kernel.dtype, use_bias=False)
       if isinstance(w1_kernel, aqt.QTensor):
         w1_pspec = aqt.partition_spec(w1_pspec, (1,), w1_kernel.dtype, use_bias=False)
       if isinstance(wo_kernel, aqt.QTensor):
         wo_pspec = aqt.partition_spec(wo_pspec, (1,), wo_kernel.dtype, use_bias=False)
-      return w0_pspec, w1_pspec, wo_pspec
+      return wi_pspec, w0_pspec, w1_pspec, wo_pspec
 
     def get_routed_moe_shardings(is_batch_sharded_by_expert, has_input_ids):
       if is_batch_sharded_by_expert:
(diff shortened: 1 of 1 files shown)


src/maxtext/layers/moe.py
```diff
@@ -1429,28 +1445,28 @@ def get_routed_moe_shardings(is_batch_sharded_by_expert, has_input_ids):
         quantization_rule = qpl.get_current_rule("gmm")
         if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
           # special sharding when using static scaling for weights in quantization with shard_exp_on_fsdp
-          w0_pspec = self._logical_to_mesh_axes(self.wi_kernel_axes)
-          w1_pspec = self._logical_to_mesh_axes(self.wi_kernel_axes)
+          wi_base_pspec = self._logical_to_mesh_axes(self.wi_kernel_axes)
```
gobbleturk: does this break the prefuse_moe_weights=False case? Don't we still need to specify w0 and w1 shardings in this case?

gobbleturk: I guess I see w0 and w1 are set to wi down below, can you include a comment here on first setting wi_base_pspec (line 1448) - that w0 and w1 will be set to wi

src/maxtext/layers/moe.py
```diff
@@ -1700,7 +1721,16 @@ def gmm_up(
...
+      if use_fused_wi:
+        out = gmm_fn(x, wi, tiling=wi_tile_size, weight_gather_axes=wi_gather_axes)
+        n = out.shape[-1] // 2
+        layer_w0, layer_w1 = out[:, :n], out[:, n:]
+        if self.config.mlp_bias:
+          layer_w0 = layer_w0 + w0_bias
+          layer_w1 = layer_w1 + w1_bias
+        layer_w0 = adc.checkpoint_name(adc.checkpoint_name(layer_w0, "mlpwi_0"), "moe_mlpwi_0")
+        layer_w1 = adc.checkpoint_name(layer_w1, "moe_mlpwi_1")
+      elif self.config.prefuse_moe_weights:
```
gobbleturk: this is confusing to me to have both a use_fused_wi=True branch and a self.config.prefuse_moe_weights branch - I would expect only one use_fused_wi branch - when would the self.config.prefuse_moe_weights be hit (and why wouldn't we prefer the use_fused_wi branch?)

abhinavgoel95: I removed `use_fused_wi` and made it the only supported path. There shouldn't be a need to not use it.

src/maxtext/layers/moe.py
abhinavgoel95: Done

src/maxtext/layers/moe.py
```diff
@@ -1380,6 +1381,11 @@ def sparse_matmul(
   ):
     """Perform sparse matrix multiplication of inputs and Experts."""
 
+    # Keep prefused [experts, embed, 2 * mlp] weights intact through shard_map
+    # so FSDP gathers the combined tensor once before the grouped GEMM.
+    if self.config.prefuse_moe_weights and wi_kernel is None:
```
RissyRan, line 1386: This is not supported with bias right? If so, shall we also add an assertion in types.py file indicating 
`prefuse_moe_weights` with `mlp_bias` are not supported? 

https://github.com/AI-Hypercomputer/maxtext/blob/58741fa8ece1d01ecc8c6e8f5829cca22712a104/src/maxtext/configs/base.yml#L287

src/maxtext/layers/moe.py
```diff
@@ -3064,15 +3112,26 @@ def __call__(
...
     fused_kernel = None
+    wi_kernel = None
     w0_kernel = None
     w1_kernel = None
     if cfg.prefuse_moe_weights and cfg.attention == "vllm_rpa" and not self.is_hash_routing:
       fused_kernel = jnp.asarray(self.wi[...], self.dtype)
     elif cfg.prefuse_moe_weights:
+      if cfg.sparse_matmul and quantizations.in_serve_mode(self.quant):
+        raise ValueError("prefuse_moe_weights with sparse_matmul does not support quantized serving.")
+      if cfg.sparse_matmul and self.wi_0_sparsity_module is not None:
```
RissyRan, line 3123: Nit, shall we put like bellow

`(wi_1_sparsity_module or wi_0_sparsity_module or)`?

RissyRan (approved):
LGTM at high level

PR #4470, opened 2026-07-15
Support TE Collective GEMM for Attention Layer Projections

dinodeep:
# Description

This PR adds support for performing collective GEMMS on projections within the attention layer: Q, K, V, and output projections when the user was previously only allowed to perform collective GEMMs on MLP GEMMs. Collective GEMMs reduce E2E execution time by better overlapping computation with exposed communication, and GEMMs in the attention layer can be large enough to benefit from such overlap (especially with QKV are fused).

This PR changes the way the user enables collectives from a boolean to using `te_comm_gemm_overlap=(disabled|mlp|full)`. "disabled" means that there are no collective GEMMs enabled; "mlp" means that only the MLP GEMMs are turned into CGEMMs, and "full" means that the MLP GEMMs and the QKV/output GEMMs from the attention layer are overlapped. The default remains as disabled/off.

Also, some small pre-commit changes were included for formatting.

# Tests

I tested this change by varying the flags across the three different configurations and validating via profiling that each of the three configurations produces collective GEMMs in the appropriate places.

# Checklist


src/maxtext/configs/base.yml
```diff
@@ -521,6 +519,7 @@ logical_axis_rules: [
                       # ==========================================
                       # Attention Activations
                       ['activation_batch_attn', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
+                      ['activation_input_length_attn', ['tensor_sequence', 'context']],
```
gobbleturk, line 539: @NuojCheng this is what I drew as "Layer Inputs" in my component breakdown... Generally layer inputs have the same rules as other components, but tensor_sequence is an exception, so layer inputs need their own component to support tensor_sequence

NuojCheng (approved):

PR #4477, opened 2026-07-15
fix: restore two-layer defense to load checkpoint 

mesakhcienet:
# Description

Restores the two-layer defense against loading a checkpoint that does not fit
the model in the training flow, and fixes a metadata-lookup bug that silently
disabled the config-level check everywhere it was already wired in.

## Problem

MaxText has two protections that were designed to work together:

1. **Config-level pre-flight** — `verify_and_sync_scan_layers` reads the
   `scan_layers` value recorded in the checkpoint's custom metadata and either
   rejects the run early with `Configuration mismatch` (user explicitly set a
   conflicting value) or silently syncs the config (user left it unset).
2. **Weight-level backstop** — `_raise_on_weight_mismatch` compares the model's
   expected weights against what the checkpoint carries and raises
   `Checkpoint does not match the model`, naming every offending weight.

A June refactor (`f5f3af7e7`) deleted the training-side mismatch handling while
the replacement metadata check was only wired into `from_pretrained` and
`to_huggingface` — the training entrypoint never ran it. A `scan_layers`
mismatch on a training resume was therefore only caught late, by the generic
weight error, which is why `test_scan_layers_mismatch_tpu` stopped seeing the
`Configuration mismatch` message it asserts on.

A second, independent bug made the pre-flight useless even where it *was*
wired in: `CheckpointManager.save(custom_metadata=…)` writes metadata at the
**step directory** (`…/0/`), but the documented `load_parameters_path`
convention points at `…/0/items`, and `load_checkpoint_metadata` read exactly
the given path — so it always returned `{}` and the check silently skipped.

# Tests

**Integration (TPU, `scheduled_only`)** — `tests/integration/checkpointing_test.py`:

- `test_scan_layers_mismatch_tpu` — saves with default `scan_layers=True`,
  reruns with explicit `scan_layers=False`; asserts the config-level
  `Configuration mismatch` + `scan_layers`. Exercises the pre-flight layer.
- `test_weight_mismatch_tpu` (new) — saves an unscanned single-layer
  checkpoint, reruns with `base_num_decoder_layers=2` and the same
  `scan_layers=False` (metadata matches, pre-flight passes); asserts
  `Checkpoint does not match the model` + `decoder/layers/1/`. Exercises the
  weight-level layer in isolation.

**Unit** — run with `python -m pytest` from the repo root:

- `tests/unit/checkpointing_metadata_test.py` (new) — against real Orbax:
  metadata readable at the step dir and at `…/items` via the parent-first
  lookup, `{}` for missing checkpoints; `latest_checkpoint_step_dir` newest
  step / `None` cases; `PreRestoreWeightCheckTest` drives
  `_raise_on_weight_mismatch` from real stored `item_metadata` on both the
  manager form (pinning the indexed `["items"]` access — `Composite.items` is
  a method, attribute access would silently no-op the check) and the
  params-only form.
- `tests/unit/model_creation_utils_test.py` — pre-flight honors an explicit
  `checkpoint_path`, falls back to `load_full_state_path`, and is a no-op with
  no resume source.
- `tests/unit/checkpointing_test.py` — metadata mocks updated to patch
  `ocp.Checkpointer`, matching the sync-checkpointer refactor.

Local runs: metadata + model-creation suites 52 passed / 25 skipped
(TPU-only/scheduled markers); checkpointing unit file 10 passed. CI pathways
integration jobs (both TPU mismatch tests) and all unit jobs green on the
pushed head.

# Checklist

### DIFF
--- a/tests/unit/checkpointing_test.py
+++ b/tests/unit/checkpointing_test.py
@@ -308,7 +308,7 @@ class SourceCheckpointLoadingTest(parameterized.TestCase):
 class CheckpointMetadataTest(parameterized.TestCase):
   """Tests for loading checkpoint custom metadata."""
 
-  @mock.patch.object(checkpointing.ocp, "StandardCheckpointer")
+  @mock.patch.object(checkpointing.ocp, "Checkpointer")
   def test_load_checkpoint_metadata(self, mock_checkpointer_cls):
     mock_ckptr = mock_checkpointer_cls.return_value
     mock_metadata = mock.MagicMock()
@@ -319,7 +319,7 @@ class CheckpointMetadataTest(parameterized.TestCase):
     self.assertEqual(loaded_metadata.get("lora"), {"lora_rank": 8, "lora_alpha": 16.0})
     mock_ckptr.metadata.assert_called_once()
 
-  @mock.patch.object(checkpointing.ocp, "StandardCheckpointer")
+  @mock.patch.object(checkpointing.ocp, "Checkpointer")
   def test_load_checkpoint_metadata_handles_exceptions(self, mock_checkpointer_cls):
     mock_ckptr = mock_checkpointer_cls.return_value
     mock_ckptr.metadata.side_effect = Exception("Checkpoint read error")
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -237,6 +237,19 @@ def _stored_metadata_tree(read_tree):
   return tree if isinstance(tree, dict) else None
 
 
+def _pre_restore_weight_check(want, have):
+  """Runs the weight check against stored metadata, but only when `have` was located.
+
+  An empty `have` means the params collection wasn't found where this layout expects it
+  (an unfamiliar on-disk nesting), not that the checkpoint is missing every weight -- so
+  skip rather than hard-fail a valid restore with a "missing" for every weight. A real
+  partial mismatch leaves the other weights in `have`, so it still raises here; a
+  genuinely empty checkpoint is caught after restore.
+  """
+  if isinstance(have, dict) and have:
+    _raise_on_weight_mismatch(want, have)
+
+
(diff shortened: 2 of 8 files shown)


.github/workflows/ci_pipeline.yml
```diff
@@ -110,21 +111,20 @@ jobs:
...
-              echo "run_notebooks=false" >> $GITHUB_OUTPUT
-            fi
+          # 5. Check for notebook (.ipynb) changes
+          if echo "$CHANGED_FILES" | grep '\.ipynb$' > /dev/null; then
+            echo "Notebook files changed, enabling notebook run."
+            echo "run_notebooks=true" >> $GITHUB_OUTPUT
+          else
+            echo "No notebook changes, skipping notebook run."
+            echo "run_notebooks=false" >> $GITHUB_OUTPUT
           fi
```
mesakhcienet: from https://github.com/AI-Hypercomputer/maxtext/pull/4474. We will remove this before merge

tests/unit/checkpointing_test.py
```diff
@@ -319,7 +319,7 @@ def test_load_checkpoint_metadata(self, mock_checkpointer_cls):
     self.assertEqual(loaded_metadata.get("lora"), {"lora_rank": 8, "lora_alpha": 16.0})
     mock_ckptr.metadata.assert_called_once()
 
-  @mock.patch.object(checkpointing.ocp, "StandardCheckpointer")
```
mesakhcienet, line 322: `StandardCheckpointer` is just shorthand for `AsyncCheckpointer(StandardCheckpointHandler())`. The async wrapper exists to make saves non-blocking: each instance spins up background save infrastructure on construction and joins it on close().

 `load_checkpoint_metadata` never saves — it only calls .metadata(), a synchronous, read-only lookup of `_CHECKPOINT_METADATA`. Since this helper constructs and closes a checkpointer on every call and runs on the initialize() pre-flight path (executed by every process in multi-host runs), the async machinery is pure per-call overhead.

PR #4478, opened 2026-07-15
Fix packed-sequence masking in cudnn_flash_jax and add cutlass_flash GPU backend

### LINKED ISSUE #4476: cudnn_flash_jax silently attends across packed sequence boundaries (decoder_segment_ids ignored in training)
## Bug report

**Summary**

With `attention=cudnn_flash_jax`, the training path ignores `decoder_segment_ids` entirely. `AttentionOp.cudnn_jax_flash_attention` receives the segment ids but the train-mode branch calls `jax._src.cudnn.fused_attention_stablehlo.dot_product_attention` with `mask_type=MaskType.CAUSAL` only (see `src/maxtext/layers/attention_op.py`, the `else` branch of `cudnn_jax_flash_attention`). As a result, when sequence packing is enabled (`packing=true`, the default, with any real dataset), tokens of a packed segment attend to all earlier segments in the same row. There is no guard and no error, so training proceeds with silently corrupted attention outputs, loss, and gradients.

For comparison, the `cudnn_flash_te` path handles packing correctly by lowering segment ids to a THD `SequenceDescriptor`.

**Steps to reproduce**

Any GPU training run with packing and `cudnn_flash_jax` is affected, e.g.:

```
  run_name=repro hardware=gpu attention=cudnn_flash_jax \
  dataset_type=tfds packing=true ...
```

A minimal numerical repro (comparing `AttentionOp` outputs against the `dot_product` reference on the same packed batch, two segments per row) gives, on A100 with bf16:

```
cudnn_flash_jax   unpacked control   max|diff| segment1=0.009  segment2=0.008  -> OK (bf16 noise)
cudnn_flash_jax   packed 2 segments  max|diff| segment1=0.009  segment2=5.406  -> cross-segment leak
```

The first segment matches the reference exactly (its causal history is identical), while every token after the packing boundary diverges, which is the signature of attending across the boundary. The repro script can be shared as part of a PR (it parametrizes over kernels and packed/unpacked/unaligned boundaries and checks both the forward output and the gradient wrt query).

**What did you expect to happen?**

Either correct per-segment causal masking (the jax cuDNN SDPA API supports packed inputs through `q_seqlen`/`q_offsets` with `MaskType.PADDING_CAUSAL`), or a hard error stating that packing is unsupported with this kernel.

**Additional observations**

1. The jax cuDNN packed (offsets) layout is only accepted on Hopper and newer; on A100 it raises `NotImplementedError: Packed layout requires a GPU with at least Hopper architecture`. So a correct fix needs a pre-Hopper fallback.
2. Related but separate: on GPUs/shapes where cuDNN declines the fused kernel, `cudnn_flash_te`'s unfused fallback crashes with `AssertionError: sequence_descriptor must be None or ndarray, but got <class 'transformer_engine.jax.attention.SequenceDescriptor'>` instead of computing unfused attention. On A100 with TE 2.17 and cuDNN 9.24 this makes `cudnn_flash_te` + packing fail for every shape we tried, since cuDNN provides no THD `PADDING_CAUSAL` kernel on sm80. Happy to file this separately if useful.

## Logs/Output

```
NotImplementedError: Packed layout requires a GPU with at least Hopper architecture.
```
(from jax when passing q_offsets on A100; the silent-corruption case produces no log at all, which is the core problem)

## Environment Information

- MaxText commit: 8e7ff2d7b5197eccec252ae5c3efe0b2ebc7bd1a
- jax 0.10.2, Transformer Engine 2.17.0, cuDNN 9.24.0 (pip), CUDA 12
- GPU: NVIDIA A100-SXM4-80GB (sm80), single node x 8
- dtype bf16, `per_device_batch_size=8`, `max_target_length=2048` for the throughput numbers

I have a fix prepared (packed seqlens/offsets lowering on Hopper+, fused fallback on older GPUs, config validation, and a kernel-vs-reference packing test) and will open a PR referencing this issue.

bzantium:
# Description

This PR fixes a silent correctness bug in the `cudnn_flash_jax` attention kernel and adds `cutlass_flash`, an optional FlashAttention-2 (CUTLASS, via `flash-attn-jax`) backend for GPUs, together with a pallas kernel used for the shapes FA2 cannot run. The two changes are coupled: the fused fallback that makes the `cudnn_flash_jax` fix practical on pre-Hopper GPUs is provided by the new backend.

### Problem 1: cudnn_flash_jax ignores decoder_segment_ids

`AttentionOp.cudnn_jax_flash_attention` received `decoder_segment_ids` but its train branch called the jax cuDNN SDPA with `mask_type=MaskType.CAUSAL` only. With sequence packing enabled (the default `packing: true` with any real dataset), tokens attended across packed segment boundaries: no guard, no error, silently wrong loss and gradients. In a kernel-vs-reference test on packed batches, every token after the packing boundary diverged from the `dot_product` reference (max abs diff 5.4 vs 0.009 bf16 noise on the first segment).

Fix, in order of preference at runtime:
- On Hopper and newer, lower segment ids to the cuDNN packed layout: per-segment `q_seqlen`/`kv_seqlen` `[B, M]` plus `q_offsets`/`kv_offsets` `[B, M+1]` (absent segments filled with -1) with `MaskType.PADDING_CAUSAL`. This is the layout `jax._src.cudnn.fused_attention_stablehlo.dot_product_attention` documents for packed inputs.
- On older GPUs the cuDNN packed layout is rejected by jax (`Packed layout requires a GPU with at least Hopper architecture`), so the op falls back to a fused flash kernel that masks segment boundaries (see below) and logs the downgrade, instead of silently attending across boundaries.
- `max_segments_per_seq` is now validated at config time for `cudnn_flash_jax` (and `cutlass_flash`) with packing, matching the existing `cudnn_flash_te` validation. The check skips `dataset_type=synthetic`, which has no real segments and never takes the packed path. Not every packing implementation caps the run count at that bound (grain `best_fit` and `concat_then_split`, and the tfds pipeline, do not), so ids above it fold into the last segment rather than being left out of every segment, which is what previously handed cuDNN uninitialized memory.

### Problem 2: no fast, correct packed attention on pre-Hopper GPUs

On A100 we found no working fused path for packed sequences: the jax cuDNN packed layout is Hopper-only, and `cudnn_flash_te` (TE 2.17, cuDNN 9.24) finds no sm80 THD kernel and then crashes in its unfused fallback (`sequence_descriptor must be None or ndarray`). The remaining correct option, unfused `dot_product`, costs about 39 percent of step throughput at seq 2048 on a qwen3-4b benchmark.

### Key changes

- `attention=cutlass_flash`: FlashAttention-2 CUTLASS kernels through the optional `flash-attn-jax` package (import-guarded; the option errors clearly when the package is absent). It stays opt-in rather than a declared dependency: `flash-attn-jax` pins `jax<0.9` while MaxText needs `jax>=0.10`, and installing it downgrades jax until the package stops importing. That also means CI never exercises this backend, and its tests skip. Native GQA/MQA, sliding window (LOCAL_SLIDING), and sequence packing lowered to FA2's varlen kernel with `cu_seqlens` built inside jit. Requires bf16/fp16 and head_dim <= 256; decode falls back to unfused attention.
- `src/maxtext/kernels/attention/gpu_pallas_flash.py`: a fork of `jax.experimental.pallas.ops.gpu.attention` extended with sliding-window masking, gemma2-style logit soft cap (with the matching backward chain rule), and chunked head_dim accumulation for head_dim > 256, so every layer FA2 cannot run still has a fused kernel.

  Since `attention=flash` moved to Tokamax, the obvious question is whether this fork can just be `tokamax_pallas_triton.PallasTritonFlashAttention`. It cannot yet, for the two cases that are exactly this backend's job. Measured on A100 (sm80) against tokamax 0.0.12, the version MaxText resolves today:

  | case | tokamax 0.0.12 |
  |---|---|
  | packed masking via `Mask(k_start=..., is_causal=True)` | works, fwd rel err 0.0028 / grad 0.0058 vs unfused reference |
  | sliding window via `k_start` | works |
  | `logits_soft_cap` forward | works |
  | `logits_soft_cap` backward | `NotImplementedError: logits_soft_cap unsupported.` (`pallas_triton_vjp.py:385`) |
  | head_dim > 256 forward | works |
  | head_dim > 256 backward | `RESOURCE_EXHAUSTED: Shared memory size limit exceeded: requested 203136, available: 166912` |

  So soft-capped and large-head_dim layers have no backward pass there and cannot train. Both gaps now have fixes in review upstream (openxla/tokamax#1121 and #1122); once they land this file should be deleted, and the kernel carries a comment saying so.

  On the motivation behind #4465 — that `jax.experimental.pallas.ops.gpu` is heading for removal from JAX: this file is a vendored copy, not an import of those ops, so it does not hold MaxText to a module JAX plans to drop. It does mean MaxText owns the kernel, which is the cost being paid to keep soft-capped and large-head_dim layers trainable on GPU until Tokamax covers them.
- A single backend dispatch (`AttentionOp._gpu_flash_backend_dispatch`) serves `attention=cutlass_flash` and the pre-Hopper packed fallback of `cudnn_flash_jax`: FA2 where eligible, pallas otherwise. Eligibility covers dtype (bf16/fp16), head_dim, soft cap, and packed-metadata availability, so no configuration silently loses segment masking. `attention=flash` on GPU is left on the Tokamax path that landed in #4465 and is not touched by this PR.
- Context parallelism is explicitly rejected for these two backends (the kernels compute masks from shard-local positions); `cudnn_flash_te` remains the kernel to use with context parallelism.

### Measurements (A100-SXM4-80GB x 8, qwen3-4b, bf16, seq 2048, per-device batch 8, synthetic data)

| attention | TFLOP/s/device | notes |
|---|---|---|
| cudnn_flash_te | 177.8 | crashes with packing on sm80 (TE fallback bug) |
| cutlass_flash | 176.6 | fast and correct with packing; measured before the current `jax>=0.10` pin, and not reproducible on it (see above) |
| dot_product | 108.0 | previous only correct packed option on sm80 |

### Shortcomings / future work

- The Hopper-native packed path of `cudnn_flash_jax` (seqlens/offsets) is implemented per the jax API contract and covered by the new test, but our verification hardware is A100; a run of `tests/unit/gpu_attention_packing_test.py` on H100 or newer would exercise it directly.
- The TE unfused-fallback crash (`SequenceDescriptor` assertion) is a separate upstream issue and is not addressed here.
- `attention=cutlass_flash` is unverified end to end. Running it needs `flash-attn-jax`, which cannot coexist with the pinned jax, so the CUTLASS kernel and its shard_map specs are reviewed but not executed. Everything else here is exercised by the new test.
- `backward_pass_impl='auto'` in the pallas kernel now always takes the fused backward. The XLA fallback is the more accurate of the two (bf16 gradients agree with the reference to ~4e-3 against ~2e-2) but materializes `[b, h, q, kv]`, needing 12.5GiB against 3.3GiB at 8k tokens on an A100, so it does not scale. It remains available explicitly.
- K/V now use `activation_kv_heads` rather than the query head axis. On a training mesh the two resolve identically, since they differ only by `autoregressive`, which is size one there and dropped; this is a semantic correction, not a fix for MQA divisibility under tensor parallelism, which is unaffected.

FIXES: #4476

# Tests

New test: `tests/unit/gpu_attention_packing_test.py` (marked `gpu_only`). For each kernel (`cudnn_flash_jax`, `cutlass_flash`, and the pallas backend, reached by hiding the cutlass library and pretending the GPU predates the cuDNN packed layout), it compares the forward output and the gradient wrt query against the `dot_product` reference on an unpacked control, a packed batch, and a packed batch with a tile-unaligned boundary, reporting relative error separately before and after the packing boundary so cross-segment leaks are directly visible. A host-side test covers the segment-ids-to-seqlens/offsets encoding.

Results on A100 (seq 2048, GQA 32/8 heads, bf16): all kernels match the reference within 2 percent relative error on both forward and gradients, for all three boundary cases. Reproduce with:

```bash
PYTHONPATH=src:. python -m pytest tests/unit/gpu_attention_packing_test.py -v
```

`pre-commit` (repo `.pre-commit-config.yaml`: codespell, `pylint` 3.3.8, `pyink` 24.10.1, yamllint) passes on the changed files.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -396,7 +396,7 @@ param_scan_axis: 1
 
 # The attention parameter dictates the specific algorithm/methodology used to compute the attention scores
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
-attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
+attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te, cutlass_flash (GPU FlashAttention-2, head_dim<=256; needs flash-attn-jax, which currently pins jax<0.9)
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla, full, compressed, block_diffusion
 share_kv_projections: false # Note: Not compatible with attention_type='mla'
 attention_bias: false # If true, adds a learnable bias to the query, key, and value projections
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -380,7 +380,7 @@ param_scan_axis: 1
 
 # The attention parameter dictates the specific algorithm/methodology used to compute the attention scores
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
-attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
+attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te, cutlass_flash (GPU FlashAttention-2, head_dim<=256; needs flash-attn-jax, which currently pins jax<0.9)
 attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
 share_kv_projections: false # Note: Not compatible with attention_type='mla'
 attention_bias: false # If true, adds a learnable bias to the query, key, and value projections
--- a/src/maxtext/configs/pyconfig_deprecated.py
+++ b/src/maxtext/configs/pyconfig_deprecated.py
@@ -105,6 +105,7 @@ def validate_attention_kernel(s: str) -> None:
       "flash",
       "cudnn_flash_te",
       "cudnn_flash_jax",
+      "cutlass_flash",
       "vllm_rpa",
       "vllm_batched_rpa",
   )
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -4198,8 +4198,20 @@ class MaxTextConfig(
           "STRIPED reorder strategy requires Transformer Engine and is only supported on GPUs. "
           f"Got hardware={self.hardware!r}."
       )
-    if self.hardware == "gpu" and self.packing and self.attention == "cudnn_flash_te" and self.max_segments_per_seq <= 0:
-      raise ValueError("max_segments_per_seq must be set when using TransformerEngine attention and packing")
+    # Mirrors AttentionOp._needs_packed_masking: synthetic data carries no real
+    # segments, so the packed path is never taken and the bound is not needed.
+    if (
+        self.hardware == "gpu"
+        and self.packing
+        and self.dataset_type != DatasetType.SYNTHETIC
+        and self.attention in ("cudnn_flash_te", "cudnn_flash_jax", "cutlass_flash")
+        and self.max_segments_per_seq <= 0
+    ):
+      raise ValueError(
+          f"max_segments_per_seq must be set to the maximum number of packed segments per "
+          f"sequence when using attention={self.attention!r} with packing; these kernels need "
+          f"that bound at trace time to lay out per-segment metadata."
+      )
     dcn_product = (
         self.dcn_data_parallelism
         * self.dcn_pipeline_parallelism
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -3487,8 +3487,20 @@ class MaxTextConfig(
           "STRIPED reorder strategy requires Transformer Engine and is only supported on GPUs. "
           f"Got hardware={self.hardware!r}."
       )
-    if self.hardware == "gpu" and self.packing and self.attention == "cudnn_flash_te" and self.max_segments_per_seq <= 0:
-      raise ValueError("max_segments_per_seq must be set when using TransformerEngine attention and packing")
+    # Mirrors AttentionOp._needs_packed_masking: synthetic data carries no real
+    # segments, so the packed path is never taken and the bound is not needed.
+    if (
+        self.hardware == "gpu"
+        and self.packing
+        and self.dataset_type != DatasetType.SYNTHETIC
+        and self.attention in ("cudnn_flash_te", "cudnn_flash_jax", "cutlass_flash")
+        and self.max_segments_per_seq <= 0
+    ):
+      raise ValueError(
+          f"max_segments_per_seq must be set to the maximum number of packed segments per "
+          f"sequence when using attention={self.attention!r} with packing; these kernels need "
+          f"that bound at trace time to lay out per-segment metadata."
+      )
     dcn_product = (
         self.dcn_data_parallelism
         * self.dcn_pipeline_parallelism
--- a/src/maxtext/kernels/attention/cutlass_flash.py
+++ b/src/maxtext/kernels/attention/cutlass_flash.py
@@ -1,0 +1,110 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""CUTLASS FlashAttention-2 wrapper (via flash-attn-jax) for MaxText.
+
+Feature parity with the GPU ``flash`` (pallas) path:
+  - causal self-attention with optional sliding window (LOCAL_SLIDING),
+  - native GQA/MQA (num_kv_heads < num_query_heads, no KV repeat),
+  - sequence packing via segment ids, lowered to FA2's varlen kernel with
+    cu_seqlens built inside jit (fixed-size, empty tail sequences are skipped
+    by the kernel).
+
+Constraints (callers must fall back for these): head_dim <= 256, sm80+,
+training/prefill only (no decode), no attention sinks / logit soft-cap.
+
+flash-attn-jax is not a declared dependency: it pins ``jax<0.9`` while MaxText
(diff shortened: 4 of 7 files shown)


Cjkkkk:
Thanks, the cudnn flash jax part of fix looks good to me. Let me run locally to test it out on B200.

Cjkkkk:
Verified on my end, this is working. But needs one additional fix in jax side `jax/jax/_src/cudnn/fused_attention_stablehlo.py` for `_fix_seqlen_offsets` to update
```
q_offsets = q_offsets * N * H
kv_offsets = kv_offsets * N * H
```
to
```
q_offsets = q_offsets * N * H
kv_offsets = kv_offsets * N_KV * H_KV
```
so it handles MQA correctly. Right now it only assumes MHA.

bzantium:
Thanks for running it, and for spotting the MQA case. I've filed that fix as jax-ml/jax#39521. It takes the kv stride from `key.shape` (`N_kv * H_kv`) rather than the query's, which also covers MLA, where the head dim differs too.

Until that lands, the Hopper+ packed path here is only correct for MHA: under GQA/MQA `kv_offsets` ends up scaled by `N / N_kv` and attention reads the wrong keys, silently. The tests here run on A100, which takes the pre-Hopper fallback, so they never reach that path; it needed your B200 run to surface. The jax-side test is verified on Blackwell.

That PR also notes a separate `BNTH` stride issue in the same function, left out to keep the fix narrow.

bzantium:
@Cjkkkk jax-ml/jax#39521 is merged (`2243550f3`) along with a regression test, so the kv offset stride you spotted is fixed upstream.

It doesn't help here yet, though. It landed after v0.11.0, so it isn't in a release, and CI here still resolves `jax>=0.10.0` to 0.10.0. Until a release picks it up and the pin moves, `attention=cudnn_flash_jax` with packing is still wrong under GQA/MQA on Hopper and newer, where the packed cuDNN layout is used. Pre-Hopper is unaffected, since it takes the fused fallback and never builds those offsets.

bzantium:
Rebased onto main, the conflicts were in `base.yml` and the `attention_op.py` imports and are resolved.

@Cjkkkk the kv offset stride fix you spotted is in upstream jax now (jax-ml/jax#39521), so the MQA case is handled once a release picks it up and the pin here moves. The rest is unchanged from what you verified on B200.

One note on CI: the `compile_cache_test.py::test_train_step_cache_hit` failure on the previous run is unrelated to this change. It is the flake fixed by #4692, where JAX skips caching a computation that compiles in under a second, so the cache is empty and there is nothing for the runtime to hit. Happy to have that land first if it makes the signal here cleaner.

PR #4488, opened 2026-07-15
Add Flash Attention Support to Deepseek v4

octatrifan:
# Description

This PR introduces Flash (Splash) Attention Support for DeepSeek-V4 Heavily Compressed Attention (HCA) and Compressed Sparse Attention (CSA) layers. Details in Bug: b/534788783

DeepSeek-V4's HCA (ratio 128) and CSA (ratio 4) require specialized block-aligned attention masks. This PR integrates these layers with TPU-optimized Pallas/Splash attention kernels (`tokamax_splash_kernel`), allowing dynamic block-masked execution directly in the forward loop. Furthermore, we adapt the mask to support document packing.

### Problem Solved
* Enabled compilation and execution of compressed attention kernels on TPUs.
* This change allows running DSV4 with increased sequence lengths that would OOM on the dot-product path. Performance testing coming soon.
* Fix sequence boundaries for Document Packing for HCA and CSA layers.

* Added `indexer_mask` support for CSA and HCA that can be passed into the Pallas Splash Kernel (Structured correct causal and sliding window masking sequences for CSA and HCA)
* Added block-size padding alignment (`sa_block_kv`) inside `CompressedAttention` that prepends padding to compressed KV blocks (to make sure the mask is a multiple of the block size, as required by the splash Attention kernel)
* Corrected sequence tracking across packed inputs by generating compressed causal masks directly from true `position_ids`.
* Isolated overlapping pooling blocks in CSA to prevent cache leakage between independent sequences during document packing.

# Tests

The changes were verified using compilation, execution, and numerical parity checks on Cloud TPU VMs:

1. **Compressed Attention Compilation Validation:**
   * Appended `CompressedAttentionTest` to `tests/unit/attention_test.py` to compile and run HCA (ratio 128) and CSA (ratio 4) configurations under both CPU (`dot_product`) and TPU (`flash`) attention paths.
   * Command to reproduce: `python3 tests/unit/attention_test.py` 

2. **Correctness check against PyTorch Reference:**
   * Executed `forward_logit` test with full model checkpoints, KL divergence matches `dot_product` path.
   Across the 489-token prompt sequence, the average KL divergence relative to the HuggingFace reference distribution is 0.027071 nats
   * Executed `tests/unit/deepseek_v4_vs_reference_test.py` to confirm that output logits match the PyTorch reference implementation within a strict tolerance.
   * Command to reproduce: `python3 tests/unit/deepseek_v4_vs_reference_test.py` 
   * Due to precision change on TPU, we were forced to increase the tolerance for some of the tests to ensure they pass on TPU.
  
 
# Performance
We conduct microbenchmarks on 1 TPU v5p for Compressed Attention. We unlock context length up to 64k from 8k at batch size 1. For the DeepSeek-V4-Flash model, performance speedups (just for the Compressed Attention Layers) with the flash attention kernel begin at a context length of 4k for a batch size of 1, and as early as 2k for batch sizes of 2 or greater. For CSA the flash kernel delivers speedups of 1.09x at a context length of 2k with a batch size of 2, and 1.53x at a context length of 4k with a batch size of 8. For the HCA speedups are observed at a context length of 4,096, reaching 1.23x for a batch size of 2 and 1.32x for a batch size of 8, and a max of 1.61x for 8k seq lengths at batch size 2.
 

# Checklist

- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents.


src/maxtext/configs/types.py
```diff
@@ -2986,6 +2986,9 @@ def set_derived_and_validate_values(self) -> "MaxTextConfig":
     if self.steps == -1:
       self.steps = self.learning_rate_schedule_steps
 
+    if self.decoder_block == DecoderBlockType.DEEPSEEK4 and self.attention == "flash":
+      self.use_tokamax_splash = True
```
shuningjin: nit: use explicit check rather than hardcode. similar to
https://github.com/AI-Hypercomputer/maxtext/blob/9f9ac051647f0f03eb93b64329802e5ecc1d3e37/src/maxtext/configs/types.py#L3349-L3357

src/maxtext/layers/attention_compressed.py
```diff
@@ -584,10 +584,15 @@ def hca_compressor_fn(buf_kv, buf_gate):
       return compressed_kv, compressed_mask
 
     # --- PREFILL CHUNKING & PRIMING ---
-    # Truncate sequence to the nearest multiple of the compression rate
-    usable = (seq_len // self.compress_rate) * self.compress_rate
-    chunk_kv = kv[:, :usable]
-    chunk_gate = gate[:, :usable]
+    # Ceil-pad sequence to nearest multiple of compression rate so all tokens are included
```
shuningjin: Could you explain why padding is needed? 

Seems the torch reference is doing truncation for kv and gate:
https://github.com/huggingface/transformers/blob/fd12552d770f745fdbe41031ff4daa688f5ed57e/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L405-L406

src/maxtext/layers/attention_compressed.py
```diff
@@ -1534,10 +1539,54 @@ def __call__(
           compressed_segment_mask[:, :, : compressed_mask.shape[-1]], axis=1
       )
 
-    # Note: Unlike standard pre-attention concatenation (extending local KV tensors with compressed blocks),
-    # compressed_kv is passed separately to attention_op to support custom kernels and decoding caching.
+    # Note: compressed_kv is passed separately to attention_op to support custom kernels and decoding caching.
+    decoder_segment_ids_kv = decoder_segment_ids
```
shuningjin: Thanks for adding support for sequence packing (`is_packed=True`).

1. Add some note to explain, preferably add small inline examples.

2. Currently, it handles HCA. As CSA performs overlapped pooling, it may need a different path. [eg](https://paste.googleplex.com/5998806960308224)


3. The document packing test in `deepseek_v4_vs_reference_test.py` only covers HCA, may need a test for CSA.

src/maxtext/layers/attention_op.py
shuningjin: update indexer mask
```
if indexer_mask is not None:
  if pass_comp_to_prefill:
    # Pass the compressed KV blocks into the prefill/training attention
    # Forward full (L + C) mask for Compressed Attention
    indexer_mask_prefill = indexer_mask
  else:
    # Use prefill and autoregressive split
    prefill_len = key.shape[1]  
    indexer_mask_prefill = indexer_mask[:, :, :prefill_len]
    if ar_kv_cache is not None:
      indexer_mask_ar = indexer_mask[:, :, prefill_len:]
```

tests/unit/deepseek_v4_vs_reference_test.py
shuningjin, line 769: The document packing test here seems rather loose. We could have a unit test in `attention_test.py` and check equivalence between `is_packed=True` and `is_packed=False`. [eg](https://paste.googleplex.com/4593499314978816)

shuningjin (commented):
Thanks for the feature! I left some comments on "pad ceil" and "sequence packing".

src/maxtext/layers/attention_op.py
```diff
@@ -1772,7 +1796,19 @@ def wrap_flash_attention(
...
+          pad_q = mask_shape[0] - indexer_mask.shape[-2]
+          pad_kv = mask_shape[1] - indexer_mask.shape[-1]
+          if pad_q > 0 or pad_kv > 0:
+            pad_width = [(0, 0)] * (indexer_mask.ndim - 2) + [(0, pad_q), (0, pad_kv)]
+            indexer_mask = jnp.pad(
+                indexer_mask,
+                pad_width,
+                constant_values=False,
+            )
+
```
parambole, line 2123: There is a conflicting double-conversion on lines 1797–1798:
1. `indexer_mask = indexer_mask == 0.0` converts the additive float mask into a boolean (`0.0` becomes `True (1)`, and `-1e10` becomes `False (0)`).
2. `indexer_mask = jnp.isclose(indexer_mask, 0.0)` immediately checks if that boolean equals `0.0`. This turns `True (1)` into `False`, and `False (0)` into `True`!
This inverts the entire attention mask on TPU (the model attends to masked-out tokens and ignores valid tokens).
```diff
       if self.config.use_tokamax_splash:
         if indexer_mask is not None:
-          indexer_mask = indexer_mask == 0.0
-          indexer_mask = jnp.isclose(indexer_mask, 0.0)
+          # 0.0 (allowed) -> True, -1e10 (masked) -> False
+          indexer_mask = indexer_mask >= -1.0
           pad_q = mask_shape[0] - indexer_mask.shape[-2]

File: src/maxtext/configs/types.py [hunk 5]
```diff
@@ -3465,8 +3465,13 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
       raise ValueError("MoBA is only supported with dot_product attention.")
-    if self.decoder_block == DecoderBlockType.DEEPSEEK4 and self.attention != "dot_product":
-      raise ValueError("DeepSeek4 decoder block currently only supports dot_product attention.")
+    if self.decoder_block == DecoderBlockType.DEEPSEEK4:
+      supports_dot_product = self.attention == "dot_product"
+      supports_flash_splash = self.attention == "flash" and self.use_tokamax_splash
+      if not (supports_dot_product or supports_flash_splash):
+        raise NotImplementedError(
+            "DeepSeek4 is only supported with dot_product attention or flash attention with use_tokamax_splash set to True."
+        )
```
parambole: **Suggestion (Exception Type Consistency):**
MaxText's Pydantic config validators consistently raise `ValueError` for unsupported configurations across `types.py`. Let's use `ValueError` instead of `NotImplementedError` so it aligns with the rest of the file and CI exception parsers:
```diff
     if self.decoder_block == DecoderBlockType.DEEPSEEK4:
       supports_dot_product = self.attention == "dot_product"
       supports_flash_splash = self.attention == "flash" and self.use_tokamax_splash
       if not (supports_dot_product or supports_flash_splash):
-        raise NotImplementedError(
-            "DeepSeek4 is only supported with dot_product attention or flash attention with use_tokamax_splash set to True."
-        )
+        raise ValueError(
+            "DeepSeek4 is only supported with `dot_product` attention or `flash` attention with `use_tokamax_splash=True`."
+        )

**Review by parambole — COMMENTED** [review]
Hey @octatrifan thank you for making the changes. I have added a few comments PTAL
 
* **Mask Inversion:**
  In `src/maxtext/layers/attention_op.py` (lines 1797–1798), delete `indexer_mask = jnp.isclose(indexer_mask, 0.0)` so only `indexer_mask = indexer_mask >= -1.0` remains.
* **Resolve Nit (Exception Type):**
  In `src/maxtext/configs/types.py` (line 3471), change `NotImplementedError` to `ValueError`.

File: src/maxtext/configs/types.py [hunk 6]
```diff
@@ -3481,8 +3481,13 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
       raise ValueError("MoBA is only supported with dot_product attention.")
-    if self.decoder_block == DecoderBlockType.DEEPSEEK4 and self.attention != "dot_product":
-      raise ValueError("DeepSeek4 decoder block currently only supports dot_product attention.")
+    if self.decoder_block == DecoderBlockType.DEEPSEEK4:
+      supports_dot_product = self.attention in ("dot_product", "autoselected")
+      supports_flash_splash = self.attention in ("flash", "autoselected") and self.use_tokamax_splash
+      if not (supports_dot_product or supports_flash_splash):
+        raise ValueError(
+            "DeepSeek4 is only supported with `dot_product` attention or `flash` attention with `use_tokamax_splash=True`."
+        )
```
dandragona: nit: Can we do a match statement? http://google3/third_party/py/maxtext/src/maxtext/utils/maxtext_utils.py;l=1502-1517;rcl=963109149

I think it might simplify.

dandragona:
> DeepSeek-V4's HCA (ratio 128) and CSA (ratio 4) require specialized block-aligned attention masks. This PR integrates these layers with TPU-optimized Pallas/Splash attention kernels (`tokamax_splash_kernel`), allowing dynamic block-masked execution directly in the forward loop. Furthermore, we adapt the mask to support document packing.

Can you explain this a bit more? Specifically a simple example of the previous behavior w.r.t. the mask contrasted with after this change? Feel free to simplify as much as possible.

src/maxtext/layers/attention_op.py
```diff
@@ -869,6 +871,8 @@ def generate_attention_mask(
       segment_positions: Optional `Array` of shape `[batch_size,
         q_sequence_length]`. Identifies original positions for load-balanced
         context-parallel inputs.
```
dandragona, line 932: needs a comment explaining pad_kv_total

src/maxtext/layers/attention_op.py
```diff
@@ -1697,7 +1720,17 @@ def wrap_ulysses_splash_kernel(single_head_mask):
...
+      block_kv = sa_config.block_kv
+      if self.attention_type == AttentionType.COMPRESSED and (
+          (query.shape[2] % block_q != 0) or (key.shape[2] % block_kv != 0)
+      ):
+        padded_q_len = ((query.shape[2] + block_q - 1) // block_q) * block_q
+        padded_kv_len = ((key.shape[2] + block_kv - 1) // block_kv) * block_kv
+        mask_shape = (padded_q_len, padded_kv_len)
+      else:
+        mask_shape = (query.shape[2], key.shape[2])  # (q_seq_len, kv_seq_len)
+
```
dandragona, line 1851: This block is hard to follow, i.e., what happens if we are doing compressed attn and the second condition is not true?

Do we always expect the second condition to be true when doing compressed attn?

src/maxtext/configs/types.py
octatrifan: Addressed

src/maxtext/layers/attention_compressed.py
octatrifan: I have reverted this padding back to the huggingface reference

src/maxtext/layers/attention_compressed.py
octatrifan: Added note to explain changes, CSA overlapped pooling support and tests to cover CSA.

src/maxtext/layers/attention_op.py
octatrifan: Thank you for pointing it out, the issue comes from a duplicated instruction upon accepting Github Gemini's comment. I have reverted back to the correct implementation.

src/maxtext/layers/attention_op.py
octatrifan: The possible paths are:
1. NOT Compressed attention -> no padding required.
2. Compressed attention but the query and key lengths are multiples of the block sizes -> no padding required.
3. Compressed attention but the query and key are NOT multiples of block sizes -> we have to pad.

I have added a comment regarding this.

src/maxtext/layers/attention_op.py
octatrifan: Comment added.

src/maxtext/configs/types.py
octatrifan: Added match statement as proposed.

src/maxtext/configs/models/deepseek4-284b.yml
```diff
@@ -60,6 +60,7 @@ routed_scaling_factor: 1.5
 
 # --- Attention configuration ---
 attention_type: 'compressed'
+use_tokamax_splash: true
```
shuningjin: model config should not include low-level implementation (like which attention / moe backend), can remove, same for other file

src/maxtext/configs/types.py
```diff
@@ -3644,8 +3644,17 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
         raise ValueError("`local_checkpoint_period` must be > 0 for emergency checkpointing.")
     if self.moba and self.attention not in ("dot_product"):
       raise ValueError("MoBA is only supported with dot_product attention.")
-    if self.decoder_block == DecoderBlockType.DEEPSEEK4 and self.attention != "dot_product":
-      raise ValueError("DeepSeek4 decoder block currently only supports dot_product attention.")
+    if self.decoder_block == DecoderBlockType.DEEPSEEK4:
+      match (self.attention, self.use_tokamax_splash):
+        case ("dot_product" | "autoselected", _):
+          pass
+        case ("flash" | "autoselected", True):
```
shuningjin: maybe remove `autoselected` logic for simplicity. 
- I prefer users explicit setting flags for complicated models like ds-v4, so they know what attention backend they are using.
- also not sure whether it works out of the box with `attention_compressed.py`

src/maxtext/layers/attention_compressed.py
```diff
@@ -79,18 +86,10 @@ def csa_overlap_pooling(
...
-      - compressed: The pooled overlapping states. Shape: `[batch, n_windows, head_dim]`.
-      - next_prior_kv: Updated KV prior for the next window.
-      - next_prior_gate: Updated gate prior for the next window.
-
-  Shape Transformations:
-    1. Split: `[batch, n_windows, compress_rate, 2 * head_dim]` -> 2x `[batch, n_windows, compress_rate, head_dim]`
-    2. Shift: Ca shifted forward by one window (prepending cache prior if available).
-    3. Concat (Ca + Cb): -> `[batch, n_windows, 2 * compress_rate, head_dim]`
-    4. Gating & Sum: -> `[batch, n_windows, head_dim]`
+    Tuple of (compressed, next_prior_kv, next_prior_gate).
```
shuningjin: nit: retain comments `next_prior_kv, next_prior_gate`

dandragona (approved):
I think attention_compressed.py needs a large refactor but im okay with leaving it outside the scope of your PR.

src/maxtext/configs/models/deepseek4-284b.yml
octatrifan: Removed flag

src/maxtext/configs/types.py
octatrifan: Removed autoselected logic

src/maxtext/layers/attention_compressed.py
octatrifan: Restored comments

src/maxtext/layers/attention_compressed.py
```diff
@@ -79,6 +86,7 @@ def csa_overlap_pooling(
     head_dim: Target head dimension.
     prior_kv: Previous window KV prior from cache (optional).
     prior_gate: Previous window gate prior from cache (optional).
+    is_same_doc: Boolean tensor indicating if a window belongs to the same document as its predecessor.
```
shuningjin: `is_same_doc: Boolean tensor of shape [batch, n_windows] indicating if window i
  belongs to the same document as previous window i - 1. Without sequence packing,
  this is always True. With packing, this is False at document boundaries, causing
  the shifted prior window (Ca) to be masked out to prevent cross-document leakage.`

shuningjin (commented):
Can you add `flash` to train compile test? currently ds4 only test dot product
https://github.com/AI-Hypercomputer/maxtext/blob/f9d53d4840ab7addd42c4a446b2393252d7d0426/tests/unit/train_compile_test.py#L829-L834

nit: Add the bug in the description `Bug: b/534788783`.

shuningjin (approved):
Thank you for adding Flash Attention support to DeepSeek-V4, along with sequence packing! 
- Appreciate the thorough validation, including the unit tests and forward logit checks. 
- Also great work on the comprehensive micro-benchmarking comparing memory usage and speedup against dot product. Looks promising!
- As a next step, we can explore more end-to-end results.

parambole (approved):
LGTM! 

Thank you @octatrifan for the great work on adding 1/N Flash (Splash) Attention support to DeepSeek-V4 for compressed attention.

🚀

PR #4493, opened 2026-07-15
FSDP Prefetch: Forward pass only

darisoy:
# Description

### Overview
This PR implements **single-buffer lookahead parameter prefetching** for Fully Sharded Data Parallelism (FSDP / ZeRO-3) across `NNXDecoderLayer` and `nnx.scan` layer boundaries (`NNXScannedPipelineStage` / `NNXDecoder`). 
In standard FSDP training, parameter All-Gather collectives lie exposed on the critical execution path before every layer's forward computation, stalling TensorCores while waiting for interconnect transfers. By restructuring our scanned layers and `custom_vjp` stack to prefetch weights for layer `l+1` asynchronously while layer `l` executes, this PR decouples parameter reconstruction from immediate consumption and enables **complete communication-compute overlap**.

### Key Architectural Changes
1. **Double-Stacked `custom_vjp` Architecture for Scanned Weight Prefetching**
   - **Outer Custom VJP (`_custom_vjp_prefetch_scan`)**:
     - Decorates the full scanned decoder stack (`@functools.partial(jax.custom_vjp, nondiff_argnums=(0, 6, 7))`).
     - **Forward Pass (`_custom_vjp_prefetch_scan_fwd`)**: Executes the `jax.lax.scan` loop over layers while lookahead prefetching weights for upcoming layers (`scheduling_group(group_id=43)`). Evaluates per-layer forward autodiff passes (`_prefetch_single_layer_custom_vjp_fwd`) and collects the sequence of per-layer `vjp_fn` closures into `vjp_seq`.
     - **Primal Weight Clearing (`vjp_fn.args_res[1] = None`)**: Drops un-sharded primal weight arrays from residual memory immediately during the forward pass to minimize peak HBM footprint.
     - **Backward Pass (`_custom_vjp_prefetch_scan_bwd`)**: Scans backward (`jax.lax.scan(..., reverse=True)`), performing lookahead prefetching for reverse layer passes, restoring `vjp_fn_i.args_res[1] = w_curr` JIT right before evaluating each layer's VJP, and applying Reduce-Scatter ops (`_reduce_scatter_param_grads`) to parameter gradients.
   - **Inner Custom VJP (`_prefetch_single_layer_custom_vjp`)**:
     - Decorates single-layer execution (`@functools.partial(jax.custom_vjp, nondiff_argnums=(0, 1, 2, 3, 4, 5))`).
     - Encapsulates per-layer rematerialization policies (`jax.checkpoint(_single_layer_fn, policy=...)`) and manages single-step VJP evaluation.
2. **Scan Carry & Rematerialization Policy Separation**
   - `scan_step_fn` handles the scanned carry tuple `(y_out, w_prefetched)` and XLA scheduling groups, while `_prefetch_single_layer_custom_vjp` cleanly manages per-layer rematerialization and checkpointing.
3. **Scan Boundary All-Gather Wrap-Around Prevention**
   - Adds explicit boundary guards (`_maybe_skip` with `idx < length - 1` in forward and `idx > 0` in backward) to prevent `_apply_sharding_hint` from initiating multi-gigabyte All-Gathers on wrapped-around boundary parameters from `jnp.roll`.
   - **XLA Scheduling Group Safety**: Places conditional control flow (`jax.lax.cond`) *outside* of `scheduling_group(group_id=43)` so that the scheduling annotation wraps only pure data-flow operations (`_apply_sharding_hint` and `checkpoint_name`), completely avoiding XLA C++ `RemoveTrivialGroups()` iterator crashes on nested control flow.
4. **Array vs. Tracer Parameter Splitting (`diff_args_leaves` / `diff_kwargs_leaves`)**
   - Keeps integer/array arguments (e.g., `decoder_positions`, `decoder_segment_ids`, and attention masks) cleanly split as differentiable/non-differentiable leaves when passed into `_custom_vjp_prefetch_scan`.
   - Ensures JAX `Tracer` objects never enter `ctx` at `nondiff_argnums=(0, 6, 7)`, preventing `UnexpectedTracerError` while keeping `mesh`, `shard_mode`, and treedefs static.
5. **Unified `NNXDecoder.__call__` Layer Execution**
   - Redirects to `_apply_layers_sequentially_prefetch` when `prefetch_fsdp_weights` is enabled, unifying prefetching across scanned decoder stacks.

### Trade-offs & Best Practices

* **Ideal Use Case**: Maximum step-time savings occur when All-Gather communication duration (`T_AG`) closely matches GMM/dense compute duration (`T_GMM`).
* **Pros**: Significantly faster forward pass execution due to 100% overlap of All-Gather communication behind GMM compute blocks.
* **Cons**: Slightly higher High Bandwidth Memory (HBM) footprint since each device must hold two reconstructed layer weight matrices in memory simultaneously during the lookahead window.

#### Recommended XLA Compiler Flags
For optimal SparseCore offloading, copy elision, and latency hiding without systolic array interruption, run with:
```bash
--xla_tpu_rerun_latency_hiding_scheduler_post_sc_assignment=true \
--xla_tpu_num_sparse_cores_for_gather_offloading=1 \
--xla_tpu_enable_copy_fusion=true \
--xla_tpu_copy_elision_analysis_allowance=150000 \
--xla_tpu_copy_insertion_use_region_analysis_limit=10000 \
--xla_tpu_use_single_sparse_core_for_all_gather_offload=true
```

### Next Steps & Future Work
* **Backward Pass Reduce-Scatter Overlap**: While All-Gathers are now completely hidden in the forward pass, parameter gradient `reduce-scatter` collectives in the backward pass (`_custom_vjp_prefetch_scan_bwd`) remain exposed at the tail of each step (waiting for `A^T x dY` GMMs to finish). Future iterations will introduce lag-behind reverse scan carry pipelining to overlap backward `reduce-scatters` with TGMMs.

BUGS: b/515348405

# Tests & Verification

### XManager & XProf Benchmarks (TPU7x-64)
* **Baseline (Without Prefetching)**: `xid/271422723`
* **With Lookahead Prefetching**: `xid/273154030 `

### Results
* **Forward Pass Layer Execution**: In a balanced compute/communication regime, single sparse decoder layer duration in the forward pass improved from ~95.6 ms to ~88.0 ms—a ~8% reduction in forward layer execution time achieved by hiding nearly 100% of the FSDP All-Gather communication overhead behind active systolic array compute.
* **Compilation Verification**: Passed Ahead-of-Time (AoT) lowering checks (`train_compile.py`) across 15 layers of DeepSeek-V3 MoE (`num_experts=128`, `enable_nnx=True`, `pure_nnx_decoder=True`).

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -1049,6 +1049,9 @@ class HardwareAndMesh(BaseModel):
           "When resuming from a checkpoint, this flag is auto-determined from metadata."
       ),
   )
+  prefetch_fsdp_weights: bool = Field(
+      False, description="Prefetch all gathering the FSDP weights for layer N+1 during the compute of layer N."
+  )
   param_scan_axis: int = Field(1, description="Axis to scan over for parameters.")
   context_parallel_load_balance: bool = Field(True, description="Whether to use load balancing for context parallelism.")
   context_parallel_strategy: str = Field(
@@ -3092,6 +3095,10 @@ class MaxTextConfig(
         ), "Pipeline weight prefetching does not support pipeline delay."
         assert not self.quantization, "Quantization is currently not supported for pipeline prefetching."
         assert not self.scan_layers_per_stage, "Pipeline weight prefetching currently does not support scan."
+
+      if self.prefetch_fsdp_weights:
+        assert self.scan_layers, "FSDP weight prefetching requires scan_layers=True."
+        assert not self.quantization, "Quantization is currently not supported for FSDP prefetching."
 
       assert (num_stages * self.num_pipeline_repeats * self.num_layers_per_pipeline_stage) == (
           self.pipeline_parallel_layers
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -16,6 +16,7 @@
 # pylint: disable=arguments-differ
 # pylint: disable=no-name-in-module
 
+import contextlib
 import functools
 import inspect
 from typing import Any
@@ -26,7 +27,7 @@ from flax import nnx
 import jax
 from jax.ad_checkpoint import checkpoint_name
 import jax.numpy as jnp
-from jax.sharding import Mesh
+from jax.sharding import Mesh, NamedSharding
 from maxtext.common.common_types import (
     Config,
     DecoderBlockType,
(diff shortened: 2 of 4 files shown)


src/maxtext/configs/types.py
```diff
@@ -1010,6 +996,9 @@ class HardwareAndMesh(BaseModel):
           "When resuming from a checkpoint, this flag is auto-determined from metadata."
       ),
   )
+  prefetch_fsdp_weights: bool = Field(
+      False, description="Enable BSW dual-buffer Windowed Scan prefetching for pure FSDP."
```
gobbleturk: lol too much love for BSW acronym. I would simply description - prefetch all gathering the FSDP weights for layer N+1 during the compute of layer N

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,240 @@ def layer_fn(carry, scanned_vars):
...
+  state = jax.tree_util.tree_unflatten(state_treedef, state_leaves)
+
+  reconstructed_args_leaves = [s if s is not None else d for d, s in zip(diff_args_leaves, static_args_leaves)]
+  args_tuple = jax.tree_util.tree_unflatten(args_treedef, reconstructed_args_leaves)
+
+  reconstructed_kwargs_leaves = [s if s is not None else d for d, s in zip(diff_kwargs_leaves, static_kwargs_leaves)]
+  kwargs_tuple = jax.tree_util.tree_unflatten(kwargs_treedef, reconstructed_kwargs_leaves)
+  valid_kwargs = dict(kwargs_tuple)
+
+  def _strip_fsdp_from_spec(spec):
```
gobbleturk: strip_fsdp_from_spec is a super useful function, can we expose this somewhere more broady, e.g. in max_utils / maxtext_utils?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,240 @@ def layer_fn(carry, scanned_vars):
...
+
+  with scheduling_group(group_id=40):
+    w_0 = jax.tree.map(lambda x: x[0], params)
+    w_init = jax.tree.map(_apply_sharding_hint, w_0)
+
+  def scan_step_fn(carry, xs):
+    y_curr, w_curr = carry
+    st_i, nxt_params_i, idx = xs
+
+    with scheduling_group(group_id=43):
```
gobbleturk: I would include a comment here why we are using scheduling group - in this case its only the FSDP all gathers all in the same schedling group, to encourage XLA to schedule them together which should make XLA's scheduling job easier IIUC

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,240 @@ def layer_fn(carry, scanned_vars):
     return final_carry
 
 
+def _run_prefetch_pipeline_fwd(
```
gobbleturk: I'm having a hard time figuring out what is_fwd means - could you include a comment  here?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,240 @@ def layer_fn(carry, scanned_vars):
...
+    w_prefetched = jax.tree.map(_maybe_skip, w_prefetched_ag, w_curr)
+
+    layer = nnx.merge(graphdef, w_curr, st_i)
+    layer_out = layer(y_curr, *args_tuple, **valid_kwargs)
+
+    y_out = layer_out[0] if isinstance(layer_out, tuple) else layer_out
+    _, _, new_state_i = nnx.split(layer, nnx.Param, ...)
+
+    if is_fwd:
+      return (y_out, jax.tree.map(jax.lax.stop_gradient, w_prefetched)), (new_state_i, y_curr)
```
gobbleturk: can you include a comment why we are calling stop_gradient here? I assume its because the real compute associated with w_prefetched is used in the next iteration

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,240 @@ def layer_fn(carry, scanned_vars):
...
+    y_i, orig_w_i, prev_params_i, st_i, idx = xs
+
+    with scheduling_group(group_id=43):
+      w_prev_ag = jax.tree.map(_apply_sharding_hint, prev_params_i)
+      w_prefetched_ag = checkpoint_name(w_prev_ag, "fsdp_bsw")
+
+    def _maybe_skip(ag_w, curr_w):
+      return jax.lax.cond(idx > 0, lambda: ag_w, lambda: curr_w)
+
+    w_prefetched = jax.tree.map(_maybe_skip, w_prefetched_ag, w_curr)
```
gobbleturk: can you include a comment with _maybe_skip is achieving? IIUC idx=0 is the last layer in the bwd (the first layer in fwd), which has no more prefetching to do

gobbleturk (approved):

src/maxtext/configs/types.py
darisoy: lol good catch. i forgot to update this description earlier. updated now

src/maxtext/layers/nnx_decoders.py
darisoy: moved to src/maxtext/utils/sharding.py

src/maxtext/layers/nnx_decoders.py
darisoy: added the comment

src/maxtext/layers/nnx_decoders.py
darisoy: yes it's to prevent JAX from calculating the gradient for the next layer's weights . added a comment

src/maxtext/layers/nnx_decoders.py
darisoy: yes, this is to skip all-gathering when computing the last layer. added a comment

src/maxtext/layers/nnx_decoders.py
darisoy: added a comment. 
essentially i am using `_run_prefetch_pipeline_fwd` as a refactored function for both the primal function during inference `_custom_vjp_prefetch_pipeline` and forward pass `_custom_vjp_prefetch_pipeline_fwd`. 
`is_fwd` is used during `_custom_vjp_prefetch_pipeline_fwd` and the only difference is that it saves the activations for gradient calculations during backward pass.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+
+  reconstructed_kwargs_leaves = [s if s is not None else d for d, s in zip(diff_kwargs_leaves, static_kwargs_leaves)]
+  kwargs_tuple = jax.tree_util.tree_unflatten(kwargs_treedef, reconstructed_kwargs_leaves)
+  valid_kwargs = dict(kwargs_tuple)
+
+  def _apply_sharding_hint(w):
+    w_sharding = getattr(jax.typeof(w), "sharding", None) or getattr(w, "sharding", None)
+    if isinstance(w_sharding, NamedSharding) and w_sharding.spec is not None:
+      w_sharding_clean = NamedSharding(mesh, w_sharding.spec)
+      w = jax.lax.with_sharding_constraint(w, w_sharding_clean)
```
NuojCheng: use `sharding.maybe_shard_with_name`?

NuojCheng: when using `maybe_shard_with_name`, set `debug_sharding=config.debug_sharding`. This feature is for sharding debug purpose.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+      stripped_spec = sharding.strip_fsdp_from_spec(w_sharding.spec)
+      w_sharding_stripped = NamedSharding(mesh, stripped_spec)
+      return sharding.maybe_shard_with_name(w, w_sharding_stripped, shard_mode=shard_mode)
+    return sharding.maybe_shard_with_name(w, w_sharding, shard_mode=shard_mode)
+
+  def _roll_and_shard(x):
+    rolled = jnp.roll(x, -1, axis=0)
+    x_sharding = getattr(jax.typeof(x), "sharding", None) or getattr(x, "sharding", None)
+    if isinstance(x_sharding, NamedSharding):
+      return jax.lax.with_sharding_constraint(rolled, x_sharding)
```
NuojCheng: use maybe shard with name

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+    layer = nnx.merge(graphdef, w_curr, st_i)
+    layer_out = layer(y_curr, *args_tuple, **valid_kwargs)
+
+    y_out = layer_out[0] if isinstance(layer_out, tuple) else layer_out
+    _, _, new_state_i = nnx.split(layer, nnx.Param, ...)
+
+    # Gradients are stopped on the prefetched weights here because they are just
+    # being carried over to the next iteration. The actual gradient computation
+    # for these weights happens in the next iteration where they are used as w_curr.
+    if is_fwd:
```
NuojCheng: I would rename `is_fwd` to `return_residual` or something more informative.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+
+
+@functools.partial(jax.custom_vjp, nondiff_argnums=(0, 6, 7))
+def _custom_vjp_prefetch_pipeline(
+    graphdef, x_in, params_leaves, state_leaves, diff_args_leaves, diff_kwargs_leaves, length, ctx
+):
+  """Forward pipeline wrapper for custom_vjp prefetching."""
+  return _run_prefetch_pipeline_fwd(
+      graphdef, x_in, params_leaves, state_leaves, diff_args_leaves, diff_kwargs_leaves, length, ctx, is_fwd=False
+  )
```
NuojCheng, line 415: what I usually do is 
```
reutrn _custom_vjp_prefetch_pipeline_fwd(...)[0]
```
so the `is_fwd` flag can be removed.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+        step_wrapped, (x_in, w_init), (state, nxt_params_seq, jnp.arange(length))
+    )
+    return (final_y, scanned_state), (
+        y_seq,
+        params_leaves,
+        state_leaves,
+        scanned_state,
+        diff_args_leaves,
+        diff_kwargs_leaves,
+    )
```
NuojCheng: it is unclear to me what `scanned_state` represents here and why we need them

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+  """Forward autodiff pass wrapper for custom_vjp prefetching."""
+  return _run_prefetch_pipeline_fwd(
+      graphdef, x_in, params_leaves, state_leaves, diff_args_leaves, diff_kwargs_leaves, length, ctx, is_fwd=True
+  )
+
+
+def _custom_vjp_prefetch_pipeline_bwd(graphdef, length, ctx, res, g_out):
+  """Backward autodiff pass wrapper for custom_vjp prefetching."""
+  y_seq, params_leaves, state_leaves, _, diff_args_leaves, diff_kwargs_leaves = res
+  y_bar_final, _ = g_out
```
NuojCheng: gradient of scanned state is dropped here

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+):
+  """Forward autodiff pass wrapper for custom_vjp prefetching."""
+  return _run_prefetch_pipeline_fwd(
+      graphdef, x_in, params_leaves, state_leaves, diff_args_leaves, diff_kwargs_leaves, length, ctx, is_fwd=True
+  )
+
+
+def _custom_vjp_prefetch_pipeline_bwd(graphdef, length, ctx, res, g_out):
+  """Backward autodiff pass wrapper for custom_vjp prefetching."""
+  y_seq, params_leaves, state_leaves, _, diff_args_leaves, diff_kwargs_leaves = res
```
NuojCheng: scanned state in res is dropped here

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+
+  def _apply_sharding_hint(w):
+    w_sharding = getattr(jax.typeof(w), "sharding", None) or getattr(w, "sharding", None)
+    if isinstance(w_sharding, NamedSharding) and w_sharding.spec is not None:
+      w_sharding_clean = NamedSharding(mesh, w_sharding.spec)
+      w = jax.lax.with_sharding_constraint(w, w_sharding_clean)
+      stripped_spec = sharding.strip_fsdp_from_spec(w_sharding.spec)
+      w_sharding_stripped = NamedSharding(mesh, stripped_spec)
+      return sharding.maybe_shard_with_name(w, w_sharding_stripped, shard_mode=shard_mode)
+    return sharding.maybe_shard_with_name(w, w_sharding, shard_mode=shard_mode)
```
NuojCheng: repeated function with fwd pass, maybe combine them?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+        grad_w,
+        orig_w,
+    )
+
+  def _roll_fwd_and_shard(x):
+    rolled = jnp.roll(x, 1, axis=0)
+    x_sharding = getattr(jax.typeof(x), "sharding", None) or getattr(x, "sharding", None)
+    if isinstance(x_sharding, NamedSharding):
+      return jax.lax.with_sharding_constraint(rolled, x_sharding)
+    return rolled
```
NuojCheng: repeatedly defined function

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+    w_prefetched = jax.tree.map(_maybe_skip, w_prefetched_ag, w_curr)
+
+    def _single_layer_fn(y_inp, w_inp):
+      layer = nnx.merge(graphdef, w_inp, st_i)
+      layer_out = layer(y_inp, *args_tuple, **valid_kwargs)
+      y_out = layer_out[0] if isinstance(layer_out, tuple) else layer_out
+      return y_out
+
+    _, vjp_fn = jax.vjp(_single_layer_fn, y_i, w_curr)
+    y_bar_prev, w_bar_unreduced = vjp_fn(y_bar_curr)
```
NuojCheng: I think vjp functions should be defined in fwd pass, not in backward. The current code should lead to checkpoint failure. To verify, you could run experiments but using `remat_policy=custom, context=device` and check if the forward pass splash exists in backward pass. If my theory is correct, then the forward pass splash should still exist and xprof should point to this vjp function, which means the remat policy definition fails when we use fsdp prefetch.

NuojCheng: We should use `jax.vjp` in forward pass, and save the VJP object in residual, so that backward pass can use it directly, e.g. 
```
def f(inputs):
  return _fwd(inputs)[0]

def _fwd(inputs):
  y, vjp = g(inputs)
  return y, vjp

def _bwd(res, g_outputs):
  g_inputs = vjp(g_outputs)
  return g_inputs

f.defvjp(_fwd, _bwd)
```

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+
+    w_bar_sharded_i = _reduce_scatter_param_grads(w_bar_unreduced, orig_w_i)
+
+    # Gradients are stopped on the prefetched weights here because they are just
+    # being carried over to the next iteration. The actual gradient computation
+    # for these weights happens in the next iteration where they are used as w_curr.
+    return (y_bar_prev, jax.tree.map(jax.lax.stop_gradient, w_prefetched)), w_bar_sharded_i
```
NuojCheng: it is unclear to me if the `stop_gradient` usage is correct

src/maxtext/layers/nnx_decoders.py
```diff
@@ -372,6 +373,260 @@ def layer_fn(carry, scanned_vars):
...
+
+  def _zero_tangent(x):
+    if isinstance(x, jax.Array) or (hasattr(jax.core, "Tracer") and isinstance(x, jax.core.Tracer)):
+      return jnp.zeros_like(x)
+    return None
+
+  state_bar_leaves = tuple(_zero_tangent(x) for x in state_leaves)
+  diff_args_bar_leaves = tuple(_zero_tangent(x) for x in diff_args_leaves)
+  diff_kwargs_bar_leaves = tuple(_zero_tangent(x) for x in diff_kwargs_leaves)
+  return y_bar_in, params_bar_leaves, state_bar_leaves, diff_args_bar_leaves, diff_kwargs_bar_leaves
```
NuojCheng: I think the backward pass logic need to be revised. The current implementation, though works, not very performant.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1019,6 +1350,71 @@ def layer_fn(carry, scanned_vars):
...
+    prevent_cse = maxtext_utils.should_prevent_cse_in_remat(self.config)
+    graphdef, params, state = nnx.split(layers, nnx.Param, ...)
+
+    scan_axis = self.config.param_scan_axis
+    if scan_axis != 0:
+      params = jax.tree.map(lambda x: jnp.moveaxis(x, scan_axis, 0), params)
+
+    layer_cls = layers.__class__
+    sig = inspect.signature(layer_cls.__call__)
+    valid_kwargs = {k: v for k, v in kwargs.items() if k in sig.parameters or "kwargs" in sig.parameters}
```
NuojCheng, line 1204: flax magic

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1019,6 +1350,71 @@ def layer_fn(carry, scanned_vars):
...
+    diff_args_leaves = tuple(x if _is_diff_leaf(x) else dummy_arr for x in args_leaves)
+    static_args_leaves = tuple(None if _is_diff_leaf(x) else x for x in args_leaves)
+
+    diff_kwargs_leaves = tuple(x if _is_diff_leaf(x) else dummy_arr for x in kwargs_leaves)
+    static_kwargs_leaves = tuple(None if _is_diff_leaf(x) else x for x in kwargs_leaves)
+
+    params_leaves, params_treedef = jax.tree_util.tree_flatten(params)
+    params_leaves = tuple(params_leaves)
+    state_leaves, state_treedef = jax.tree_util.tree_flatten(state)
+    state_leaves = tuple(state_leaves)
```
NuojCheng, line 1230: I don't understand what line 1369 - 1389 are doing. We need some comments to explain

src/maxtext/layers/nnx_decoders.py
```diff
@@ -1019,6 +1350,71 @@ def layer_fn(carry, scanned_vars):
...
+
+    ctx = (
+        self.mesh,
+        self.config.shard_mode,
```
NuojCheng, line 1234: add self.config.debug_sharding for maybe_shard_with_name functions

src/maxtext/utils/sharding.py
```diff
@@ -919,3 +919,10 @@ def all_gather_over_fsdp(variables, sharding_info, mesh, logical_axis_rules, sha
   # Apply the constraint to the model's current variables. This tells JAX to
   # gather the weights into this layout.
   return maybe_shard_with_name(variables, physical_constraint_no_fsdp, shard_mode=shard_mode)
+
+
+def strip_fsdp_from_spec(spec):
+  """Strip FSDP mesh axes from partition spec."""
+  if spec is None:
+    return None
+  return P(*(None if axis in ("fsdp", "fsdp_transpose") else axis for axis in spec))
```
NuojCheng: to me this function does the same thing with https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/utils/sharding.py#L812-L834, can we re-use old code?

tests/unit/nnx_decoders_test.py
```diff
@@ -30,7 +30,8 @@
 import numpy as np
 from flax import linen as nn
 from flax import nnx
-from jax.sharding import Mesh
+from jax._src.sharding_impls import GSPMDSharding  # pylint: disable=protected-access
```
NuojCheng: import jax._src may fail in g3 build

src/maxtext/layers/nnx_decoders.py
darisoy: removed the is_fwd  flag

src/maxtext/layers/nnx_decoders.py
darisoy: removed this flag all together

src/maxtext/utils/sharding.py
darisoy: reusing the existing code and removed the function i added previously

tests/unit/nnx_decoders_test.py
darisoy: removed

src/maxtext/layers/nnx_decoders.py
darisoy: added debug_sharding's to maybe_shard_with_name's

src/maxtext/layers/nnx_decoders.py
darisoy: good point it doesnt seem like its needed. removed

src/maxtext/layers/nnx_decoders.py
darisoy: updated to use sharding.maybe_shard_with_name with debug_sharding=

src/maxtext/layers/nnx_decoders.py
darisoy: done

src/maxtext/layers/nnx_decoders.py
darisoy: moved _apply_sharding_hint to a higher level helper function both fwd and bwd pass can use

src/maxtext/layers/nnx_decoders.py
darisoy: refactored

src/maxtext/layers/nnx_decoders.py
darisoy: will work on bwd pass optimization in the next pr. exposed RS's seem to be biggest culprit

src/maxtext/layers/nnx_decoders.py
darisoy: okay custom_vjp is working. got the same result as before: https://xprof.corp.google.com/trace_viewer/darisoy-6614830898965459162

src/maxtext/layers/nnx_decoders.py
darisoy: `jax.lax.stop_gradient` is applied to `w_prefetched` in `scan_step_fn` because those weights are just being carried over in the scan carry to act as `w_curr` in iteration `i + 1`. `stop_gradient` cuts redundant gradient computation across the scan carry boundary during the next layer's AG. The actual weight gradients are computed in iteration `i + 1` when `w_curr` is passed into `_single_layer_fn_remat`.
I've added an inline comment explaining this in the code.

src/maxtext/layers/nnx_decoders.py
darisoy: apparently this is needed for next layer's compatibility

src/maxtext/layers/nnx_decoders.py
darisoy: reformatting is needed to the only args pass to custom_vjp are jax arrays. added a comment explaining this.

PR #4506, opened 2026-07-16
add util function get_save_and_offload_names

aireenmei:
# Description
Extract the logics from decoders.py into a util function get_save_and_offload_names in maxtext_utils.py, so it allows other modules to call. Use case: gemma4 requires customized handling of offload, specifically for the global layer scan implementation (up-coming PR).

# Tests

Unit test

# Checklist

### DIFF
--- a/src/maxtext/utils/maxtext_utils.py
+++ b/src/maxtext/utils/maxtext_utils.py
@@ -195,6 +195,39 @@ def should_prevent_cse_in_remat(config):
   return True
 
 
+def get_offload_remat_names(config):
+  """Returns ``(save_names, offload_names)`` for offloading remat policies, else ``None``.
+
+  Single source of truth for which checkpointed tensors a remat policy saves on
+  device vs. offloads to pinned host. Shared by ``Decoder.get_remat_policy`` (which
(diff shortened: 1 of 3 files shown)


src/maxtext/utils/maxtext_utils.py
```diff
@@ -195,6 +195,45 @@ def should_prevent_cse_in_remat(config):
...
+        "qkv_proj",
+        "out_proj",
+        "mlpwi_0",
+        "mlpwi_1",
+        "mlpwi",
+        "mlpwo",
+    ]
+  if config.remat_policy == "custom":
+    return list(config.tensors_on_device or []), list(config.tensors_to_offload or [])
+  return None
```
NuojCheng: should we return None, None?

NuojCheng (approved):
Thank you for the refactoring. It looks much better now

hengtaoguo (approved):

src/maxtext/utils/maxtext_utils.py
aireenmei: Good call. I update it to return [],[] to be more consistent

PR #4525, opened 2026-07-17
Fix NNX MTP zero-loss, qwix MTP crash, and DeepSeek scan-axis sharding corruption

ecnal-cienet:
Fixes three bugs in the NNX path surfaced by running DeepSeek-V3 with multi-token prediction. They are independent; the third is not an MTP bug at all and was hiding behind the second.

## 1. `mtp_loss` is silently 0 under NNX (b/535659954)

`mtp_losses` / `mtp_acceptance` subclass `nnx.Intermediate`, and NNX type filters match by `isinstance`. So the generic `nnx.pop(model, nnx.Intermediate)` in `loss_fn` consumed them before the dedicated pop ran, which then returned `{}`; `calculate_mtp_loss` fell through to its `return 0.0` default. No error, just a wrong number.

The comment claiming these variables are "not Intermediate, so the nnx.pop above misses them" was the false premise behind the bug.

Fix: pop the specific subclasses before the generic `nnx.Intermediate` pop. Linen is unaffected — it keys mutable collections by literal name.

## 2. `TypeError: roll requires ndarray ... got NoneType` with qwix (b/535662227)

NNX modules are stateful, so `qwix.quantize_model` must run a real forward pass. MaxText passed only tokens/positions/segment_ids, so the MTP block received `target_ids=None` and `jnp.roll` raised. Linen never hits this because `quantize_linen_model` rejects model inputs and never traces a forward.

Fix: supply dummy targets at that call site, gated on `mtp_num_layers > 0`, restoring the semantics of the `is_initializing()` guard that the NNX port left commented out in `models.py`.

## 3. DeepSeek params silently replicated under NNX (pre-existing)

Surfaced once (2) was fixed: `AssertionError: Unsharded parameter percentage (2.60%) exceeds tolerance (2.00%)`.

`_create_scanned_layers` records each stack's axis name in `nnx.PARTITION_NAME` — `dense_layers` / `moe_layers` for DeepSeek — but `_apply_layers_sequentially` hardcoded the literal `"layers"` when round-tripping the scan axis. The name never matched, so the fallback `while len(l) > ndim: l.pop(0)` deleted the real leading logical axis and `nnx_add_scan_axis` inserted a bogus `'layers'`:

```
decoder/dense_layers/mlp/wi_0/kernel
  before: ('embed', 'dense_layers', 'mlp')
  after : ('dense_layers', 'layers', 'mlp')   <- after ONE forward pass
```

`'embed'` is the only carrier of `fsdp`, so the params became `P(None, ...)`.

This is neither an MTP bug nor a qwix bug: it reproduces at `mtp_num_layers=0` and with qwix entirely absent (29 of 32 params corrupted by a single plain forward). It normally stays invisible because `get_abstract_state_nnx` snapshots shardings before the first train step, so the damage lands on a discarded traced copy. qwix makes it fatal only because it quantizes from inside model construction, i.e. inside the traced snapshot. Standard models pass `metadata_axis_name="layers"`, matching the literal — only DeepSeek-style multi-stack models are exposed.

Fix: resolve the axis name per-variable from `nnx.PARTITION_NAME` rather than a caller literal, and raise instead of blindly stripping an axis when the metadata is inconsistent. Flax's own `nnx.spmd.remove_axis` raises on this same mismatch; this reimplementation corrupted silently.

`sharding_tolerance` is untouched — the params are genuinely sharded again.

FIXES: b/535662227
FIXES: b/535659954

## Test
```
  base_output_directory=/home/wanglance_google_com/tmp/out run_name=nnx_mtp \
  model_name=deepseek3-tiny per_device_batch_size=1 max_target_length=128 steps=20 \
  dataset_type=synthetic enable_checkpointing=false sparse_matmul=false \
  mtp_num_layers=1 mtp_loss_scaling_factor=0.1
```
+ `enable_nnx=false pure_nnx_decoder=false pure_nnx=false` for Linen, + `use_qwix_quantization=true quantization=fp8_full` for qwix.

- [nnx_mtp](https://paste.googleplex.com/4724701682335744)
- [nnx_mtp_qwix](https://paste.googleplex.com/6535000483823616)
- [linen_mtp](https://paste.googleplex.com/6555822703378432)
- [linen_mtp_qwix](https://paste.googleplex.com/4700876357894144)

# Checklist

### DIFF
--- a/src/maxtext/layers/nnx_decoders.py
+++ b/src/maxtext/layers/nnx_decoders.py
@@ -998,13 +998,9 @@ class NNXDecoder(nnx.Module):
       final_carry, scanned_state = jax.lax.scan(layer_fn_wrapped, x_in, (params, state))
       returned_kv_stacked = None
 
-      # Ensure metadata rank matches the stacked values
-      scanned_state = maxtext_utils_nnx.nnx_add_scan_axis(scanned_state, "layers", 0)
-
-      if scan_axis != 0:
-        new_params, new_rest = scanned_state.split(nnx.Param, ...)
-        new_params = maxtext_utils_nnx.nnx_sync_moveaxis(new_params, 0, scan_axis)
-        scanned_state = nnx.merge_state(new_params, new_rest)
+      # Move the scan axis to each variable's param_scan_axis and restore its name
+      # in the sharding metadata. jax.lax.scan emits it at position 0.
+      scanned_state = maxtext_utils_nnx.nnx_add_and_sync_scan_axis(scanned_state, "layers")
 
       returned_kv_stacked = None
 
--- a/src/maxtext/utils/maxtext_utils_nnx.py
+++ b/src/maxtext/utils/maxtext_utils_nnx.py
@@ -218,33 +218,6 @@ def nnx_update_sharding_meta(variable, transform_fn):
   return variable
 
 
-def nnx_sync_moveaxis(tree, from_axis, to_axis):
-  """Moves an axis in both values and sharding metadata of nnx.Variables."""
-  if from_axis == to_axis:
-    return tree
-
-  def _op(x):
-    is_var = isinstance(x, nnx.Variable)
-    val = x.get_value() if is_var else x
-    if not hasattr(val, "shape"):
-      return x
-
-    new_val = jnp.moveaxis(val, from_axis, to_axis)
-    if not is_var:
-      return new_val
-
-    def move_fn(l):
-      while len(l) < val.ndim:
-        l.append(None)
-      if len(l) > max(from_axis, to_axis):
-        l.insert(to_axis, l.pop(from_axis))
-      return l
-
-    return nnx_update_sharding_meta(x.replace(value=new_val), move_fn)
-
-  return jax.tree.map(_op, tree, is_leaf=lambda x: isinstance(x, nnx.Variable) or hasattr(x, "shape"))
-
-
 def nnx_remove_scan_axis(tree, name="layers"):
   """Removes the given scan axis from the PartitionSpec."""
 
@@ -252,11 +225,25 @@ def nnx_remove_scan_axis(tree, name="layers"):
     if not isinstance(x, nnx.Variable):
       return x
 
+    # Scanned stacks record their own axis name, such as "dense_layers" or "moe_layers",
+    # so prefer it over the caller's default. Otherwise the name never matches and the
+    # check below strips a real logical axis instead of the scan axis.
+    axis_name = x.get_metadata().get(nnx.PARTITION_NAME, name)
+
     def remove_fn(l):
-      if name in l:
-        l.remove(name)
-      while len(l) > x.get_value().ndim:
-        l.pop(0)
+      removed = axis_name in l
+      if removed:
+        l.remove(axis_name)
+      if len(l) > x.get_value().ndim:
+        if removed:
+          raise ValueError(
+              f"Sharding names {l} still exceed value rank {x.get_value().ndim} after removing scan axis "
+              f"{axis_name!r}; the partition metadata is inconsistent."
+          )
+        raise ValueError(
+            f"Scan axis {axis_name!r} not found in sharding names {l} for a rank-{x.get_value().ndim} value; "
+            "the partition metadata is inconsistent."
+        )
       return l
 
     return nnx_update_sharding_meta(x, remove_fn)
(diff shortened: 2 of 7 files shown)


src/maxtext/utils/maxtext_utils_nnx.py
```diff
@@ -252,11 +252,19 @@ def _op(x):
...
+    axis_name = x.get_metadata().get(nnx.PARTITION_NAME, name)
+
     def remove_fn(l):
-      if name in l:
-        l.remove(name)
-      while len(l) > x.get_value().ndim:
-        l.pop(0)
+      if axis_name in l:
+        l.remove(axis_name)
+      if len(l) > x.get_value().ndim:
```
xibinliu, line 237: there's a small edge case with the error message: if the axis_name is successfully found and removed, but the sharding list l is still longer than the tensor's rank due to some other metadata corruption .

We can change like this:
```
   was_removed = False
    if axis_name in l:
      l.remove(axis_name)
      was_removed = True
      
    if len(l) > x.get_value().ndim:
      if not was_removed:
        raise ValueError(
            f"Scan axis {axis_name!r} not found in sharding names {l} for a rank-{x.get_value().ndim} "
            "value; the partition metadata is inconsistent."
        )
      else:
        raise ValueError(
            f"Sharding names {l} exceed value rank {x.get_value().ndim} even after removing "
            f"scan axis {axis_name!r}; the partition metadata is inconsistent."
        )
```

src/maxtext/utils/maxtext_utils_nnx.py
```diff
@@ -271,11 +279,13 @@ def _op(x):
     if not isinstance(x, nnx.Variable):
       return x
 
+    axis_name = x.get_metadata().get(nnx.PARTITION_NAME, name)
+
```
xibinliu, line 273: same for pos, can we fetch `param_scan_axis` from the metadata and use pos as a fallback (for cases where nnx.vmap didn't inject the axis metadata):
```
      actual_pos = x.get_metadata().get("param_scan_axis", pos)
```

src/maxtext/utils/maxtext_utils_nnx.py
ecnal-cienet: Applied.

src/maxtext/utils/maxtext_utils_nnx.py
ecnal-cienet: Done. `nnx_add_scan_axis` now resolves the position from each variable's `param_scan_axis` (fallback `pos`) and moves the value to match, so the separate `nnx_sync_moveaxis` step is gone. Verified metadata is preserved (0/32 corrupted), the qwix sharding assert passes, and e2e losses are unchanged across deepseek/gpt3/gemma.

src/maxtext/utils/maxtext_utils_nnx.py
```diff
@@ -265,17 +279,30 @@ def remove_fn(l):
 
 
 def nnx_add_scan_axis(tree, name="layers", pos=0):
```
xibinliu: Rename to nnx_add_and_sync_scan_axis?

src/maxtext/utils/maxtext_utils_nnx.py
xibinliu: Remove this func. It is not used any more.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -998,13 +1072,9 @@ def layer_fn(carry, scanned_vars):
...
-      # Ensure metadata rank matches the stacked values
-      scanned_state = maxtext_utils_nnx.nnx_add_scan_axis(scanned_state, "layers", 0)
-
-      if scan_axis != 0:
-        new_params, new_rest = scanned_state.split(nnx.Param, ...)
-        new_params = maxtext_utils_nnx.nnx_sync_moveaxis(new_params, 0, scan_axis)
-        scanned_state = nnx.merge_state(new_params, new_rest)
+      # Move the scan axis to each variable's param_scan_axis and restore its name
+      # in the sharding metadata. jax.lax.scan emits it at position 0.
+      scanned_state = maxtext_utils_nnx.nnx_add_scan_axis(scanned_state, "layers")
```
xibinliu: How about:

      scanned_state = maxtext_utils_nnx.nnx_add_scan_axis(scanned_state, "layers", scan_axis)

ecnal-cienet: I instrumented this before applying: on this path every variable already carries `param_scan_axis` — 65/65 (`Param`→1, `RngKey`/`RngCount`→0) — so `pos` is never consulted and the change is inert today.

I kept the default `0` because if the fallback ever were hit, `scan_axis` would be wrong for the RNG state: those variables explicitly carry `param_scan_axis=0`, and the pre-refactor code only moved `nnx.Param` (`nnx_sync_moveaxis(new_params, ...)`), leaving non-Param state at 0. So `0` matches the old behaviour for anything that falls through.

src/maxtext/utils/maxtext_utils_nnx.py
ecnal-cienet: Removed

src/maxtext/utils/maxtext_utils_nnx.py
ecnal-cienet: Done

xibinliu (approved):

tests/unit/quantizations_nnx_mtp_test.py
```diff
@@ -0,0 +1,151 @@
+# Copyright 2023-2026 Google LLC
```
NuojCheng: I suggest moving tests to `quantization_test` or `multi_token_prediction_test.py`

ecnal-cienet: Done. Thanks.

NuojCheng (approved):

PR #4066, opened 2026-06-04
Fix: Gemma 3 & 4 Base Model and Flax Linen Decoding  

RexBearIU:
# Description
  
This PR introduces architectural configurations for Gemma3 and Gemma4, and implements critical decoding layer KV-cache propagation fixes for JAX/NNX
  (`nnx_decoders.py`) and Flax Linen (`decoders.py`) pipelines.
  
Before this change, Flax Linen scanned blocks (`_apply_gemma3_scanned_blocks` and `_apply_gemma4_scanned_blocks`) lacked mechanisms to propagate    
  intermediate KV-caches across scanned layers under `nn.scan`, causing end-to-end decodes to fail.
  
Key Changes:
    - **Flax Linen Decoder (`src/maxtext/layers/decoders.py`)**: Added `kv_caches` and `attention_metadata` propagation across Gemma3/4 scanned blocks  
  via a stack/unstack mapping over `nn.scan`.
    - **JAX/NNX Decoder (`src/maxtext/layers/nnx_decoders.py`)**: Unified dynamic KV-cache carry updates within scanned block execution.
    - **Model Scaffolding (`gemma3.py`, `gemma4.py`)**: Added `kv_cache` routing through scannable blocks and layer inputs.
  
# Tests
  
- Validated via end-to-end decoding with Gemma3-4B on an 8-device TPU mesh.
- Run unit tests:
```bash
python3 -m pytest tests/post_training/unit/lora_utils_test.py
```
- Decode test:
- [command and output](https://paste.googleplex.com/4521857989607424)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #4530, opened 2026-07-18
Rewrite Gemma4 scannable block

aireenmei:
# Description

Currently the implementation of Gemma4ScannableBlock makes XLA treat the whole 6-layer block as a "big layer". The all-gathers for all 6 layers are run at the beginning of each block. Similarly, in backward pass, remat runs the whole block. This causes elevated HBM usage and more exposed collectives. This PR builds on top of my previous PRs #4506 and #4507, which allows a more modular and clean implementation.

After this rewrite:
* 5 local layers via jax.lax.scan, each per-layer jax.checkpoint
* length-1 scan for global layer, also wrapped in jax.checkpoint
* remove block-level remat and unroll outer block scan

### Added after revision:
* Refactor in gemma4.py
* Fix ckpt conversion: update checkpoint conversion code for both to_maxtext and to_hf directions.
* Fix gemma4-26b: support sowing of `moe_lb_loss` from layer intermediates to fix a bug breaking 26b model
* Fix decode: add "local_layer" to logical rules to fix a bug breaking vllm decode introduced in #4066

# Tests
* added unit test
* Before rewrite: [xprof](http://xprof.corp.google.com/trace_viewer/aireenmei-16525735383985599335) for pdbs=4. [xprof](http://xprof.corp.google.com/trace_viewer/aireenmei-3752206657021917776) for pdbs=8
* After rewrite: [xprof](https://xprof.corp.google.com/trace_viewer/aireenmei-2068683819829996003?view_start=-6192.735&view_end=47760.970) for pdbs=4. [xprof](http://xprof/?session_id=aireenmei-3228254048924162039) for pdbs=8 + vocab_tiling=8 (OOM without vocab_tiling). 
* Before rewrite, the HBM peak is not on vocab peak([screenshot](https://screenshot.googleplex.com/BNLXydTTJqgsFGp)). After rewrite, vocab peak becomes the highest ([screenshot](https://screenshot.googleplex.com/oPgh9cAzZs7E6oW)) due to other peaks become smaller, also by having remat boundary at layer level adds HBM usage and overlap with the vocab peak. Therefore, without vocab tiling, the peak HBM usage doesn't change much compared with original. But when combined with vocab_tiling, we can effectively lower the HBM peak a lot (70 GB -> 40 GB for pdbs=4).

### Added after revision
* checkpoint conversion to_maxtext & logit check:
  * [31B log](https://paste.googleplex.com/6723341921681408)
  * [26B log](https://paste.googleplex.com/4505618714066944)
  * added unit test in parem_mapping_test.py
* checkpoint conversion to_hf:
  * [31B to_hf log](https://paste.googleplex.com/5038062890516480) 
* Decode:
  * [26B decode script](https://paste.googleplex.com/6634462606655488) similar to the test script in #4066
  * [26B decode log](https://paste.googleplex.com/6541973380005888)
  * added uinit test to guard the kv_cache support

# Checklist

### DIFF
--- a/src/maxtext/layers/nnx_scan.py
+++ b/src/maxtext/layers/nnx_scan.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Utilities for constructing stacks of scanned NNX layers."""
+"""Utilities for constructing and applying stacks of scanned NNX layers."""
 
 from collections.abc import Callable
+from typing import Any
 
 from flax import nnx
 import jax
@@ -87,3 +88,57 @@ def create_scanned_layers(
   stacked_params = add_scan_metadata(stacked_params, param_scan_axis)
   stacked_rest = add_scan_metadata(stacked_rest, 0)
   return nnx.merge(layer_graphdef, stacked_params, stacked_rest)
+
+
+def apply_scanned_layers(
+    layers: nnx.Module,
+    carry: Any,
+    *,
+    length: int,
+    param_scan_axis: int,
+    apply_fn: Callable[[nnx.Module, Any], Any],
+    remat: bool = False,
+    remat_policy: Callable[..., Any] | None = None,
+    prevent_cse: bool = True,
+    unroll: int = 1,
+) -> Any:
+  """Applies stacked NNX layers using ``jax.lax.scan``.
+
+  This helper owns the generic NNX state and scan-axis mechanics. ``apply_fn``
+  defines the model-specific module invocation and must return the next carry.
+  ``remat`` is separate from ``remat_policy`` because ``None`` is JAX's full
+  rematerialization policy, not an indication that rematerialization is off.
+
+  Externally managed per-layer state, such as KV caches, is not supported by
+  this scan path.
+
+  Note:
+    This is a minimal, model-agnostic scan primitive. The other NNX decoder
+    paths instead go through ``NNXDecoder._apply_layers_sequentially``, a heavier
+    applier that also threads external (vLLM) KV caches via a static unroll and
+    re-applies scan-axis metadata. Gemma4 is currently the only caller of this
+    function; unifying the two appliers is a follow-up cleanup.
+  """
+  if length <= 0:
+    return carry
+
+  layer_graphdef, params, state = nnx.split(layers, nnx.Param, ...)
+  if param_scan_axis != 0:
+    params = jax.tree.map(lambda x: jnp.moveaxis(x, param_scan_axis, 0), params)
+
+  def scan_body(current_carry, scanned_state):
+    current_params, current_state = scanned_state
+    current_layer = nnx.merge(layer_graphdef, current_params, current_state)
+    next_carry = apply_fn(current_layer, current_carry)
+    return next_carry, nnx.state(current_layer)
+
+  scan_fn = jax.checkpoint(scan_body, policy=remat_policy, prevent_cse=prevent_cse) if remat else scan_body
+  final_carry, scanned_state = jax.lax.scan(scan_fn, carry, (params, state), unroll=unroll)
+
+  if param_scan_axis != 0:
+    scanned_params, scanned_other = scanned_state.split(nnx.Param, ...)
+    scanned_params = jax.tree.map(lambda x: jnp.moveaxis(x, 0, param_scan_axis), scanned_params)
+    scanned_state = nnx.State.merge(scanned_params, scanned_other)
+
+  nnx.update(layers, scanned_state)
+  return final_carry
--- a/src/maxtext/models/gemma4.py
+++ b/src/maxtext/models/gemma4.py
@@ -15,18 +15,19 @@
 """Specialized layers for Gemma 4."""
 
 import jax
+from jax.experimental import xla_metadata
 from jax.ad_checkpoint import checkpoint_name
 from jax.sharding import Mesh
 import jax.numpy as jnp
 
 from flax import linen as nn
 from flax import nnx
-from typing import Optional
+from typing import Optional, Any
 
 from maxtext.common.common_types import Config, AttentionType, MODEL_MODE_PREFILL
 from maxtext.layers import initializers
 from maxtext.layers import moe
-from maxtext.layers import nnx_wrappers
+from maxtext.layers import nnx_scan, nnx_wrappers
 from maxtext.layers import quantizations
 from maxtext.layers.attentions import Attention
 from maxtext.layers.linears import MlpBlock
@@ -35,6 +36,7 @@ import jax.sharding
 from maxtext.layers.normalizations import RMSNorm
 from maxtext.layers.quantizations import AqtQuantization as Quant
 from maxtext.utils import max_utils
+from maxtext.utils import maxtext_utils
 
 
 GEMMA4_ATTENTION_PATTERN = (
@@ -409,7 +411,7 @@ Gemma4DecoderLayerToLinen = nnx_wrappers.to_linen_class(
 
 
 class Gemma4ScannableBlock(nnx.Module):
-  """A repeatable block of Gemma4 decoder layers."""
+  """A repeatable block of Gemma4 decoder layers, scanning local layers."""
 
   def __init__(
       self,
@@ -418,7 +420,9 @@ class Gemma4ScannableBlock(nnx.Module):
       model_mode: str,
       rngs: nnx.Rngs,
       quant: None | Quant = None,
-      num_of_layers: int = 1,
+      num_of_layers: int = 6,
+      remat_policy_fn: Any = None,
+      apply_internal_remat: bool = False,
   ):
     """Initializes the instance.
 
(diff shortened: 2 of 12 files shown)


aireenmei:
I add pull ready to generate internal cl for testing. Won't submit until PR approved

src/maxtext/models/gemma4.py
```diff
@@ -429,27 +433,110 @@ def __init__(
...
     self.rngs = rngs
     self.num_of_layers = num_of_layers
+    self.remat_policy_fn = remat_policy_fn
+    self.apply_internal_remat = apply_internal_remat
+
+    pattern_length = len(GEMMA4_ATTENTION_PATTERN)
+    if not 0 <= num_of_layers <= pattern_length:
+      raise ValueError(f"Gemma4ScannableBlock must contain between 0 and {pattern_length} layers; got {num_of_layers}.")
+
+    # Pattern is 5 local, 1 global.
```
RissyRan: It seems the pattern has been defined. Will something like this work?

```
    active_pattern = GEMMA4_ATTENTION_PATTERN[:num_of_layers]
    self.num_local = sum(1 for t in active_pattern if t == AttentionType.LOCAL_SLIDING)
    self.num_global = sum(1 for t in active_pattern if t == AttentionType.GLOBAL)
```

src/maxtext/models/gemma4.py
```diff
@@ -429,27 +433,110 @@ def __init__(
...
+              rngs=layer_rngs,
+              attention_type=AttentionType.LOCAL_SLIDING,
+              layer_idx=0,  # layer_idx is not used in the class
+          ),
+          length=self.num_local,
+          param_scan_axis=self.config.param_scan_axis,
+          metadata_axis_name="local_layers",
+          rngs=self.rngs,
+      )
+    else:
```
RissyRan, line 486: Will this be triggered for Gemma4?

src/maxtext/models/gemma4.py
```diff
@@ -471,22 +558,120 @@ def __call__(
     y = inputs
 
     updated_kvs = []
-    for layer_id in range(self.num_of_layers):
-      current_kv = kv_cache[layer_id] if kv_cache is not None else None
-      y, new_kv = getattr(self, f"layers_{layer_id}")(
+
+    if kv_cache is not None:
```
RissyRan, line 681: I guess you meet issue when calling `kv_cache[layer_id]`? 

Wondering if you tested for decoding with the change end-to-end and see reasonable outputs?

src/maxtext/models/gemma4.py
```diff
@@ -471,22 +558,120 @@ def __call__(
...
+          updated_kvs.append(new_kv)
+          per_layer_states.append(nnx.state(layer))
+
+        stacked_state = jax.tree.map(lambda *xs: jnp.stack(xs), *per_layer_states)
+        if scan_axis != 0:
+          stacked_params, stacked_other = stacked_state.split(nnx.Param, ...)
+          stacked_params = jax.tree.map(lambda x: jnp.moveaxis(x, 0, scan_axis), stacked_params)
+          stacked_state = nnx.State.merge(stacked_params, stacked_other)
+        nnx.update(self.local_layers, stacked_state)
+    elif self.local_layers is not None:
```
RissyRan: Is this the same as line 564?

RissyRan (commented):
Thanks for the change! Overall it looks good. I am a little bit lost in gemma4.py if/else branches, wondering if we could simplify a little bit?

src/maxtext/layers/nnx_scan.py
```diff
@@ -87,3 +88,50 @@ def update_leaf(leaf):
...
+    carry: Any,
+    *,
+    length: int,
+    param_scan_axis: int,
+    apply_fn: Callable[[nnx.Module, Any], Any],
+    remat: bool = False,
+    remat_policy: Callable[..., Any] | None = None,
+    prevent_cse: bool = True,
+    unroll: int = 1,
+) -> Any:
```
NuojCheng, line 104: is this funtion only used in gemma4? I imagine every models gonna use this function, for my own understanding, why this code block is created in a gemma4-related PR?

src/maxtext/models/gemma4.py
```diff
@@ -429,27 +433,110 @@ def __init__(
...
+      )
+      return layer_out[0] if isinstance(layer_out, tuple) else layer_out
+
+    apply_remat = self.apply_internal_remat and self.config.remat_policy != "none"
+    if apply_remat:
+      prevent_cse = maxtext_utils.should_prevent_cse_in_remat(self.config)
+      remat_policy = self.remat_policy_fn
+    else:
+      prevent_cse = True
+      remat_policy = None
```
NuojCheng: for my own understanding, why fix prevent cse as true when apply remat is false?

src/maxtext/models/gemma4.py
```diff
@@ -471,22 +558,120 @@ def __call__(
...
+        # working set and OOMs. Offloaded (pinned-host) residuals can't cross that
+        # boundary, so the global layer saves would-be-offloaded tensors on device
+        # instead; the local-layer scan (a real multi-iteration scan) still offloads.
+        global_remat_policy = self.remat_policy_fn
+        offload_names = maxtext_utils.get_save_and_offload_names(cfg)
+        if offload_names[0] or offload_names[1]:
+          save_names, offload_to_device = offload_names
+          global_remat_policy = jax.checkpoint_policies.save_only_these_names(*(save_names + offload_to_device))
+
+        if self.apply_internal_remat and self.config.remat_policy != "none":
```
NuojCheng: config.remat_policy is never "none"? When it can be "none"?

src/maxtext/models/gemma4.py
```diff
@@ -471,22 +558,120 @@ def __call__(
...
+          scan_global_layer = jax.checkpoint(
+              scan_global_layer,
+              policy=global_remat_policy,
+              prevent_cse=prevent_cse,
+          )
+
+        # Carry state through the loop instead of returning a stacked [1, ...]
+        # scan result: slicing that result previously introduced a bitcast
+        # between device and pinned-host memory under offload remat.
+        with xla_metadata.set_xla_metadata(**{"skip-simplify-while-loops_trip-count-one": "true"}):
```
NuojCheng: I am thinking we could move this context manager to train.py with a maxtext flag controlling it, wdyt? @gobbleturk @RissyRan

NuojCheng (approved):

src/maxtext/models/gemma4.py
NuojCheng: in a separate PR of course, let's keep it in this PR

src/maxtext/models/gemma4.py
aireenmei: code updated as suggested

src/maxtext/models/gemma4.py
aireenmei: Added comments line 466-469
```
    # num_of_layers can be 0: the decoders always construct a "remainder" block
    # for num_decoder_layers % pattern_length layers, which is 0 whenever the
    # layer count divides evenly (e.g. 31b = 60 layers). That block is built but
    # never applied, so num_local or num_global may legitimately be 0 here.
```

src/maxtext/models/gemma4.py
aireenmei: The kv_cache related changes are to preserve the feature introduced in #4066. I added unit tests "test_gemma4_block_external_kv_cache_matches_scanned_path" to guard this feature for future. Also attached end to end decode results in PR description test section

src/maxtext/models/gemma4.py
aireenmei: The original code checks "local_layers is not None" in both the kv_cache branch and the non-kv cache branch, causing confusion. This section has now been refactored with helper functions for each branch ([screenshot](https://screenshot-v2.corp.google.com/A3rwUymc3WrxEeF)). Hopefully it's easier to follow

src/maxtext/layers/nnx_scan.py
aireenmei: Other models have been using [NNXDecoder._apply_layers_sequentially](https://github.com/AI-Hypercomputer/maxtext/blob/6d89c4176b8785d60501b829911ad345f6eccb81/src/maxtext/layers/nnx_decoders.py#L865). I added cross reference in the docstring of both functions.
```
    This is a minimal, model-agnostic scan primitive. The other NNX decoder
    paths instead go through ``NNXDecoder._apply_layers_sequentially``, a heavier
    applier that also threads external (vLLM) KV caches via a static unroll and
    re-applies scan-axis metadata. Gemma4 is currently the only caller of this
    function; unifying the two appliers is a follow-up cleanup.
```

src/maxtext/models/gemma4.py
aireenmei: Thanks for catching. prevent_cse is actually a no-op when apply_remat is False, so the value doesn't matter. Now removed "prevent_cse = True", will leave it the default value.

src/maxtext/models/gemma4.py
aireenmei: decoders.get_remat_policy checks `if cfg.remat_policy != "none"` (https://github.com/AI-Hypercomputer/maxtext/blob/df95a1fcdccc78166bc453faf016c680351eeefb/src/maxtext/layers/decoders.py#L349). Looks like "none" is allowed, at least historically. We can deprecate it later if no longer needed. But for now, I will keep it to align with the current code.

src/maxtext/models/gemma4.py
aireenmei: sounds good

RissyRan (approved):

PR #4536, opened 2026-07-20
fix: resolve Gemma 3/4 RL rollout gibberish by unrolling scanned weights for vLLM adapter

mesakhcienet:
# Description

### Problem
When running `train_rl.py` RL rollouts for Gemma 3 and Gemma 4 with `scan_layers=true`, the vLLM integration generates pure gibberish (e.g., `"enenenen..."`). 
The root cause is a structural mismatch: Gemma 3/4's cyclic attention forces MaxText to split scanned layers into nested blocks (`layers.layers_0` and `layers_remainder`). Because Gemma relies on the `MaxTextForCausalLM` adapter (which expects flat, integer-keyed layers like standard Hugging Face models), the weight mapping silently failed. As a result, vLLM initialized with random weights, leading to the gibberish output.

### Solution
1. **Unroll Scanned Weights**: fix previously introduced `unroll_gemma_scanned_weights` in `maxtext_vllm_rollout.py`. This function detects the `layers.layers_0` structure and dynamically unrolls it into a flat, integer-keyed dictionary, ensuring exact compatibility with the `MaxTextForCausalLM` vLLM adapter.
2. **Type Safety Enforcement**: The unrolling logic strictly enforces `int` keys instead of `str` (e.g., `0` instead of `"0"`) to comply with the adapter's `nnx.List` target structure. 
3. **End-to-End Test Update**: Updated `tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh` to explicitly include the required `use_standalone_converter=true` flag.

### Why `use_standalone_converter=true` is required for Gemma 3/4

Unlike Llama 3 or Qwen 3, Gemma 3 and 4 are bleeding-edge architectures that are not yet natively supported by upstream vLLM. To run them, MaxText overrides the vLLM model registry using a custom Python adapter (`MaxTextForCausalLM`). 

Because of this custom adapter and Gemma's complex cyclic attention (`layers_0` and `layers_remainder`), the default `tunix` fast-mapping path (`transfer_state_with_mappings`) completely fails to match the tensor structures. When `tunix` fails, it silently falls back to random weights (resulting in gibberish). 

Setting `use_standalone_converter=true` explicitly bypasses `tunix` and routes the weights through our custom conversion script—triggering the new `unroll_gemma_scanned_weights` logic and ensuring the adapter receives a properly formatted dictionary.

 If you look at the exact configuration passed to the vLLM rollout adapter inside `src/maxtext/trainers/post_train/rl/train_rl.py (around line 499)`, you will see this explicitly defined
  flag:

```python
 rollout_vllm_additional_config=rollout_additional_config,
 rollout_vllm_init_with_random_weights=True, # <--- THIS LINE
 rollout_vllm_enable_dp_attention=trainer_config.enable_dp_attention,
```

   1. The Optimization Hack: To save memory and time during the RL loop, MaxText does not ask vLLM to load real weights from disk. Because MaxText is about to stream the Actor's trained weights directly from its own memory over to vLLM, it tells vLLM to just boot up instantly using `jax.random` (pure garbage initialization). This is what `rollout_vllm_init_with_random_weights=True` does.
   
   2. The tunix Failure Mode: tunix maps weights by matching dictionary keys. If a key from MaxText (e.g. `layers.layers_0`) doesn't perfectly match a key in the vLLM model registry (e.g. `layers.0`), tunix simply skips that tensor. It does not crash or throw a fatal error.
   
   3. The Resulting State: Because tunix skipped those tensors, it never overwrites the garbage memory. vLLM proceeds to execute the decoding rollout using those skipped, randomly initialized weights.

  When you feed random weights through a Transformer's attention layers, the math collapses into a repeating feedback loop of garbage tokens, which is exactly why it spit out `"enenenenenenenen"`.

  This single line of configuration (`rollout_vllm_init_with_random_weights=True`) combined with tunix's silent failure mode is the definitive proof of why it generates gibberish!

# Tests

## gemma3-4b unscan
- Script: [here](https://paste.googleplex.com/5331807120785408)
- Log: [here](https://paste.googleplex.com/6515263280381952)

## gemma3-4b scan
`use_standalone_converter` and `scan_layers` should be set to `true`
- Script: [here](https://paste.googleplex.com/6396478175182848), 
- Log: [here](https://paste.googleplex.com/5925278923751424)

# Checklist

### DIFF
--- a/tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh
+++ b/tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh
@@ -43,6 +43,7 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
     model_name=${MODEL_NAME} enable_single_controller=${use_pathways} \
     checkpoint_storage_use_zarr3=False checkpoint_storage_use_ocdbt=False \
     rollout_tensor_parallelism=1 \
+    use_standalone_converter=true \
     vllm_hf_overrides='{architectures: ["MaxTextForCausalLM"]}' \
     vllm_additional_config='{"maxtext_config": {"model_name": "gemma3-4b", "log_config": "false"}}'
 
--- a/src/maxtext/integration/vllm/maxtext_vllm_rollout.py
+++ b/src/maxtext/integration/vllm/maxtext_vllm_rollout.py
@@ -51,21 +51,38 @@ def _create_model_converter(model_name: str, config: Any, mesh: jax.sharding.Mes
   return None
 
 
+def _find_scanned_layer_idx(key_tuple, container_names=("layers", "scanned_blocks", "layers_remainder")):
+  """Returns (container_idx, container_name) if a scanned layer structure is found, else (-1, None)."""
+  for name in container_names:
+    for i in range(len(key_tuple) - 1):
+      if key_tuple[i] == name and isinstance(key_tuple[i + 1], str) and key_tuple[i + 1].startswith("layers_"):
+        return i, name
+  return -1, None
+
+
 def unroll_gemma_scanned_weights(weights):
   """Workaround for tunix unstacking bug with Gemma 3/4 scanned blocks.
 
-  tunix fails to map nested layers like `layers.layers_0` to `layers_X`.
-  We manually unroll them here if we detect the structure.
+  tunix fails to map nested layers like `layers.layers_0` to `layers_X`
+  if the target expects integer keys (as in nnx.List).
+  We manually unroll them here, keeping the keys as tuples with integers.
   """
-  if not hasattr(weights, "to_pure_dict"):
+  if hasattr(weights, "to_pure_dict"):
+    pure_dict = weights.to_pure_dict()
+  elif hasattr(weights, "to_dict"):
+    pure_dict = weights.to_dict()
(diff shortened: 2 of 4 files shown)


tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh
```diff
@@ -43,6 +43,7 @@ python3 -m maxtext.trainers.post_train.rl.train_rl \
     model_name=${MODEL_NAME} enable_single_controller=${use_pathways} \
     checkpoint_storage_use_zarr3=False checkpoint_storage_use_ocdbt=False \
     rollout_tensor_parallelism=1 \
+    use_standalone_converter=true \
```
mesakh123, line 46: it's required when the `scan_layers=true`

hengtaoguo (approved):

hengtaoguo:
Could you also take a look at the failing CPU tests and see if relevant? Thanks!

SurbhiJainUSC (approved):

mesakhcienet:
> Could you also take a look at the failing CPU tests and see if relevant? Thanks!

Looks like it was from main branch, i have rebased our branch and the old failing CPU tests are passed. Thank you!

PR #4566, opened 2026-07-22
Unify the NNX and Linen restore paths in load_state_if_possible

ecnal-cienet:
## Description

The NNX path had two early-return blocks that each duplicated the Linen restore flow, one for the standard manager and one for the emergency managers. Any change to restore logic had to be made in three places, and they had already drifted: the NNX grain branch dropped the iterator element that the Linen branch returns.

pure_nnx saves in the Linen on-disk layout, so both paths restore the same tree. This converts the NNX abstract to that layout going in, runs the existing `match` statement unchanged, and reshapes what comes back on the way out. Linen behavior is unchanged.

## Bugs fixed behind that duplication

- **Dynamic safetensors got the whole abstract state instead of the weights.** For NNX that is fatal before any weight is read: `opt_state` is keyed by integer chain indices, and the loader flattens its target with a `"."` separator, so it died in `flatten_dict` with `TypeError: sequence item 2: expected str instance, int found`.
- **A weight no mapping covered reached the model as an untrained init value.** Orbax rejects a shape conflict itself, but leaves a weight the checkpoint does not carry as an unmaterialized placeholder. The weights-only Orbax load already checks for this; the dynamic load now does too.
- **`checkpoint_conversion_fn` was unusable.** It arrives from config as a dotted string but was called directly, raising `TypeError` for any value that was set. It is now imported, and resolved before the checkpoint is read.
- **Full-state safetensors now handles NNX**, previously wired only for the `orbax` layout.

## Testing

92 unit tests pass across the six suites that touch this code:
```
python -m pytest tests/unit/checkpointing_nnx_load_test.py \
                 tests/unit/checkpointing_test.py \
                 tests/unit/checkpointing_nnx_roundtrip_test.py \
                 tests/unit/checkpointing_emergency_nnx_test.py \
                 tests/unit/setup_initial_state_nnx_test.py \
                 tests/unit/grain_utility_test.py
```

End-to-end on a v6e-8: 74 assertions across gpt3-52k / gemma2-2b / llama2-7b, 0 failures.

| Case | Path | Model | Evidence | Log |
| 1a/1b | NNX save → resume (standard manager) | llama2-7b | `restoring from this run's directory step 9`, resumed 10 → 19 | [log](https://paste.googleplex.com/4817163755061248) [log](https://paste.googleplex.com/5137482466328576) |
| 2a/2b | Linen save → resume (regression) | gemma2-2b | Resumed 10 → 19, behavior unchanged | [log](https://paste.googleplex.com/6452070436700160) [log](https://paste.googleplex.com/5952328023801856) |
| 3 | Linen checkpoint → NNX resume | llama2-7b | `Restoring Linen-layout checkpoint into NNX state`, continued from the Linen step 19 → 22 | [log](https://paste.googleplex.com/5725509517443072) |
| 4 | Weights-only load (`load_parameters_path`) | gemma2-2b | Params restored into the NNX state, training continues | [log](https://paste.googleplex.com/5798646502260736) |
| 5 | Full-state load (`load_full_state_path`) | llama2-7b | Reshaped into NNX, continued from the restored step 9 → 12 | [log](https://paste.googleplex.com/4788029062512640) |
| 9a–9d | Emergency manager save → resume | gemma2-2b | Local dir wiped between save and resume, so the restore came from the persistent GCS copy; NNX and Linen both 6 → 11 | [log](https://paste.googleplex.com/4763719077855232) [log](https://paste.googleplex.com/6244904744386560) [log](https://paste.googleplex.com/6441677999964160) [log](https://paste.googleplex.com/6211234927083520) |
| 6a | Shape mismatch is rejected | gpt3-52k | `Requested shape (2048, 16) is not compatible with the stored shape (1024, 16)`, zero training steps | [log](https://paste.googleplex.com/6406491086127104) |
| 6b | Missing weight is named | gpt3-52k | `'decoder/layers/2/self_attention/qkv_proj/kernel': missing (model expects (16, 3, 2, 8) float32)`, zero training steps | [log](https://paste.googleplex.com/4957342897668096) |
| 7 | `safetensors` without a conversion fn | gpt3-52k | Names the missing config flag; no raw `'NoneType' object is not callable` | [log](https://paste.googleplex.com/4843344718069760) |
| 8/8b | HuggingFace safetensors → NNX | qwen3-0.6b | `Successfully mapped 13 parameters`, trained 3 steps. Pre-fix, same command: `TypeError`, zero parameters mapped, zero steps | [log](https://paste.googleplex.com/4913930693640192) [log](https://paste.googleplex.com/5631954350374912) |

The resume also matches Linen numerically. Two independently trained gemma2-2b runs, across the save and the restore:

| step | NNX | Linen |
| --- | --- | --- |
| 9 (checkpoint written) | 5.488 | 5.498 |
| 17 (after restore) | 2.280 | 2.288 |
| 19 | 2.036 | 2.043 |

# Checklist

### DIFF
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -16,6 +16,7 @@
 """Create an Orbax CheckpointManager with specified (Async or not) Checkpointer."""
 
 import datetime
+import importlib
 import time
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


src/maxtext/common/checkpointing.py
```diff
@@ -547,49 +579,12 @@ def map_to_pspec(data):
...
-        return (
-            restored,
-            None,
-        )
-
-      # Only Linen TrainState reaches here; the NNX cases returned above.
-      restore_target = abstract_unboxed_pre_state
+      # pure_nnx saves in the Linen on-disk layout, so both paths restore the same tree through
+      # the same manager calls below: convert the NNX abstract going in, reshape it coming out.
+      is_nnx = isinstance(abstract_unboxed_pre_state, nnx.State)
```
xibinliu: move to the very beginning of the func so it can be re-used by all branches.

src/maxtext/common/checkpointing.py
```diff
@@ -158,33 +159,58 @@ def _load_linen_checkpoint_into_nnx(
...
+  return abstract_unboxed_pre_state.params
+
+
+def _bare_weights(tree):
+  """Strips the Flax `params` collection wrapper so weights compare at the same depth.
+
+  A Linen params tree is the collection, an NNX one the bare weights; the dynamic
+  safetensors loader always returns the collection.
+  """
+  return tree["params"] if isinstance(tree, dict) and set(tree) == {"params"} else tree
```
xibinliu: set(tree) == {"params"}

Nit: make it efficient and understandable:

len(tree) == 1 and "params" in tree

src/maxtext/common/checkpointing.py
ecnal-cienet: Done.

src/maxtext/common/checkpointing.py
ecnal-cienet: Done

xibinliu (approved):

NuojCheng (approved):

shuningjin (approved):

PR #4571, opened 2026-07-22
Remove all-to-all operation from TSP/CP in Attention Segmentation Mask

dinodeep:
# Description

When performing TSP/CP with packed sequences, the segmentation mask produced for attention from ids could introduce a large all-to-all operation (transporting bytes on the order of `O(S^2)`). This PR avoids that by introducing an all-gather on the small non-broadcasted to reduce communication.

More specifically, this line
```
mask = decoder_segment_ids[:, :, None] == decoder_key_segment_ids[:, None, :]
```
would produce an all-to-all on the broadcasted `decoder_key_segment_ids[:, None, :]` when running with TSP/CP since `decoder_segment_ids` is `[B, S]` following the input data sharding. The sequence dimension is sharded under TSP/CP constraints; however, this introduces a large all-to-all collective when constructing the `mask`. 

To fix this, we add a sharding constraint on the right hand side of the equation. This all-gathers `decoder_key_segment_ids` on the sequence dimension for the right-hand side prior to broadcasting for a much smaller collective with the same sharded output. This reduces communication latency and can improve performance when training with large sequence lengths.

# Tests

I observed all-to-all operations (300-400us) in the HLO for a Llama3-70b training run executing with tensor_sequence_parallelism. Then, after running the same configuration but with the new changes, I did not observe an all-to-all instruction in the HLO, and instead, the operation was replaced with a small all-gather (30us latency).

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -565,6 +565,7 @@ logical_axis_rules: [
                       # Standard MLP / Dense Layers / Model Structure
                       # ==========================================
                       # Dense Activations
+                      ['segment_ids_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_mlp', ['tensor', 'tensor_sequence']],
                       # Note activation batch and length also get used in vocab
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -590,6 +590,7 @@ logical_axis_rules: [
                       # Standard MLP / Dense Layers / Model Structure
                       # ==========================================
                       # Dense Activations
+                      ['segment_ids_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
                       ['activation_mlp', ['tensor', 'tensor_sequence']],
                       # Note activation batch and length also get used in vocab
                       ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -49,6 +49,7 @@ from maxtext.common.common_types import (
     CACHE_SCALE_SEQUENCE,
     CACHE_SEQUENCE,
     Config,
+    SEGMENT_ID_BATCH,
     DECODE_BATCH,
     DECODE_LENGTH,
     DECODING_ACTIVE_SEQUENCE_INDICATOR,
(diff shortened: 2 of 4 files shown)


src/maxtext/layers/attention_op.py
```diff
@@ -752,7 +748,15 @@ def generate_attention_mask(
     if model_mode == MODEL_MODE_AUTOREGRESSIVE and decoder_segment_ids is not None:
       mask = decoder_segment_ids[:, None, None, None, :] == DECODING_ACTIVE_SEQUENCE_INDICATOR
     elif decoder_segment_ids is not None:
-      mask = decoder_segment_ids[:, :, None] == decoder_segment_ids[:, None, :]
+
+      # With TSP/CP, all-gather prior to broadcast to avoid large all-to-all on broadcasted ids.
+      if not self.config.using_pipeline_parallelism:
+        key_sharding = self._logical_to_mesh_axes((DATA_EMB_BATCH, None))
```
gobbleturk: activation_embed_and logits is meant to only be used for the vocab activations (batch, sequence, vocab). You may need to create a different logical axis rule, or write a function explicitly all gather over TSP / CP

dinodeep: Thanks for the review, I created a new logical axis rule specifically for the segment ids `"segment_ids_batch"`

src/maxtext/configs/base.yml
```diff
@@ -507,8 +508,10 @@ mesh_axes: ['diloco', 'data', 'stage', 'fsdp', 'fsdp_transpose', 'context', 'con
 logical_axis_rules: [
                       ['circular_repeats', []],
                       # ==========================================
-                      # Vocabulary Embedding
+                      # Data / Vocabulary Embedding
                       # ==========================================
+                      # Data Segment IDs
+                      ['segment_ids_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
```
NuojCheng: could you move it under ` Standard MLP / Dense Layers / Model Structure` -> `Dense Activations`

dinodeep: Moved under dense activations

NuojCheng (approved):

src/maxtext/configs/base.yml
```diff
@@ -564,6 +565,7 @@ logical_axis_rules: [
                       # Standard MLP / Dense Layers / Model Structure
                       # ==========================================
                       # Dense Activations
+                      ['segment_ids_batch', ['data', 'stage', 'fsdp', 'fsdp_transpose', 'expert']],
```
gobbleturk: can you remove stage from here? PP should not be sharding this

src/maxtext/layers/attention_op.py
```diff
@@ -750,7 +748,15 @@ def generate_attention_mask(
     if model_mode == MODEL_MODE_AUTOREGRESSIVE and decoder_segment_ids is not None:
       mask = decoder_segment_ids[:, None, None, None, :] == DECODING_ACTIVE_SEQUENCE_INDICATOR
     elif decoder_segment_ids is not None:
-      mask = decoder_segment_ids[:, :, None] == decoder_segment_ids[:, None, :]
+
+      # With TSP/CP, all-gather prior to broadcast to avoid large all-to-all on broadcasted ids.
+      if not self.config.using_pipeline_parallelism:
```
gobbleturk: I don't think PP is special here, should do the same all gather within microbatches of PP - should be more performant also for PP and simpler code

gobbleturk (approved):

huytransformer (approved):
I think might need a rebase since main has some new tests that could cause trouble

PR #4577, opened 2026-07-23
Optimize chunked delta rule inversion.

h1t35h:
- Replaced `jax.scipy.linalg.solve_triangular` with `invert_unit_lower_triangular_log_depth`, a highly optimized log-depth Newton-Schulz iteration implementation with a custom VJP for computing `(I+S)^-1`.
- Updated `jax.lax.scan` unroll parameters in the naive implementation.

# Description

1. Improves performance of triangular identity solver (adds a jax implementation). 
2. Unrolls jax.lax.scan given the small size of the loop XLA is able to optimize on the fly.
3. Tested on TPU v6e.

# Tests

Tested manually attached xprof results.
**Before Optimization**
For chunk_size: 128 
fwd : 96.96 ms
bwd: 127 ms
xprof:
image

**After Optimization**
fwd: 23 ms
bwd: 52 ms
xprof:
image

# Checklist

### DIFF
--- a/src/maxtext/models/qwen3.py
+++ b/src/maxtext/models/qwen3.py
@@ -170,7 +170,7 @@ def naive_jax_chunk_gated_delta_rule(
 
     return new_last_recurrent_state, core_attn_out_i
 
-  final_state, core_attn_out_stacked = jax.lax.scan(scan_body, last_recurrent_state, xs)
+  final_state, core_attn_out_stacked = jax.lax.scan(scan_body, last_recurrent_state, xs, unroll=0)
 
   core_attn_out = jnp.transpose(core_attn_out_stacked, (1, 2, 0, 3, 4))
   core_attn_out = core_attn_out.reshape(batch_size, num_heads, -1, v_head_dim)
@@ -180,6 +180,57 @@ def naive_jax_chunk_gated_delta_rule(
   return core_attn_out, final_state if output_final_state else None
 
 
+@jax.custom_vjp
+def invert_unit_lower_triangular_log_depth(S):
+  """
+  Computes (I + S)^-1 for a strictly lower triangular matrix S
+  using log-depth Newton-Schulz iterations.
+
+  This is highly optimized for TPUs/GPUs and replaces
+  jax.scipy.linalg.solve_triangular for chunkwise linear attention.
+  """
+  chunk_size = S.shape[-1]
+
+  # Ensure S is strictly lower triangular (zero out diagonal and upper half)
+  # This guarantees mathematical correctness and stability
+  S_strict = jnp.tril(S, k=-1)
+
+  # Base identity matrix
+  identity = jnp.eye(chunk_size, dtype=S.dtype)
+
+  # Initial approximation and error term
+  A = identity - S_strict
+  E = jnp.tril(S_strict @ S_strict, k=-1)
+
+  # Log-depth Taylor series exact computation
+  steps = int(math.ceil(math.log2(chunk_size)))
+  for _ in range(steps - 1):
+    # Update inverse and error using batched matmuls
+    A = jnp.tril(A + A @ E)
+    E = jnp.tril(E @ E, k=-1)
+
+  return A
+
+
+@functools.partial(jax.named_call, name="invert_triangular_fwd")
+def _invert_unit_lower_triangular_log_depth_fwd(S):
+  A = invert_unit_lower_triangular_log_depth(S)
+  return A, A
+
+
+@functools.partial(jax.named_call, name="invert_triangular_bwd")
+def _invert_unit_lower_triangular_log_depth_bwd(res, g):
+  A = res
+  grad_S = jnp.tril(-(A.mT @ g @ A.mT), k=-1)
+  return (grad_S,)
+
+
+invert_unit_lower_triangular_log_depth.defvjp(
+    _invert_unit_lower_triangular_log_depth_fwd, _invert_unit_lower_triangular_log_depth_bwd
+)
+
+
+@functools.partial(jax.named_call, name="jax_chunked_delta_rule")
 def jax_chunk_gated_delta_rule(
     query: Array,
     key: Array,
@@ -264,11 +315,11 @@ def jax_chunk_gated_delta_rule(
   S = S * jnp.exp(g_diff)
   S = jnp.where(mask, S, 0.0)
 
-  # Inversion (A) - Strictly float32
-  identity = jnp.eye(chunk_size, dtype=jnp.float32)
-  identity_broadcasted = jnp.broadcast_to(identity, S.shape)
+  # Cast to float32 explicitly as you were doing before
+  S = S.astype(jnp.float32)
 
-  A = jax.scipy.linalg.solve_triangular(identity + S, identity_broadcasted, lower=True, unit_diagonal=True)
+  # Inversion (A) - Replaces solve_triangular entirely
+  A = invert_unit_lower_triangular_log_depth(S)
 
   # 5. WY Factors
   v_beta = v_c * beta_c[..., None]
--- a/tests/unit/qwen3_next_vs_reference_test.py
+++ b/tests/unit/qwen3_next_vs_reference_test.py
@@ -22,6 +22,7 @@ from flax import nnx
 import jax
 import jax.numpy as jnp
 from jax.sharding import Mesh
+from jax.test_util import check_grads
 from maxtext.configs import pyconfig
 from maxtext.layers import normalizations
 from maxtext.layers.normalizations import Qwen3NextRMSNorm, Qwen3NextRMSNormGated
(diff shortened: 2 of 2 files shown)


tests/unit/qwen3_triangular_solve_test.py
```diff
@@ -0,0 +1,49 @@
...
+import numpy as np
+import jax
+import jax.numpy as jnp
+from maxtext.models.qwen3 import invert_unit_lower_triangular_log_depth
+
+class TestQwen3TriangularSovle(unittest.TestCase):
+    def test_invert_unit_lower_triangular_log_depth(self):
+        """Test for loss across chunk_size in the following range [16, 32, 64, 128, 256]."""
+        jax.config.update('jax_enable_x64', True)  # Use float64 for precise testing
+        chunk_sizes = [16, 32, 64, 128, 256]
```
gobbleturk: I don't think its valuable to protect multiple chunk sizes, I would just pick one for this test

h1t35h: Thanks! Updated.

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
   return core_attn_out, final_state if output_final_state else None
 
 
+@jax.custom_vjp
```
gobbleturk, line 183: Since this following the current pattern of GDN being defined in this file I think its okay to define this here, but in the future its likely we want to refactor GDN into its own file so other non-qwen models can share such optimal implementations

tests/unit/qwen3_triangular_solve_test.py
gobbleturk, line 1: I would rename to just triangular_solve_test or gdn_triangular_solve_test, I don't think there is any qwen3 specific logic other than qwen3 uses GDN

gobbleturk (commented):
This generally LGTM, can you include an e2e manual correctness tests? Maybe reach out to parambole@ for how to run

tests/unit/qwen3_triangular_solve_test.py
NuojCheng: I would put it in [attention_test](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/attention_test.py) or [qwen3 next test](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/qwen3_next_vs_reference_test.py)

h1t35h: Updated moved it to [qwen3_next_test](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/qwen3_next_vs_reference_test.py) thanks for the suggestion.

tests/unit/qwen3_next_vs_reference_test.py
```diff
@@ -1037,6 +1038,40 @@ def run_jax(x):
     )
     print("test_qwen3_next_sparse_moe_block passed!")
 
+  def test_invert_unit_lower_triangular_log_depth(self):
+    """Test for loss at chunk_size 256."""
+    jax.config.update("jax_enable_x64", True)  # Use float64 for precise testing
```
suhithr, line 1043: This checks for mathematical correctness, but I think we should prefer testing the precision that we are using in the model (as `S` is cast to `float32` in `qwen3.py`).

suhithr (commented):
Thank you!
+1 could you include the e2e correctness verification?

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
...
+def invert_unit_lower_triangular_log_depth(S):
```
parambole, line 184: nit:

Small doc/naming suggestions, both optional:

The docstring says Newton–Schulz, but I don't think that's quite what this is. Newton–Schulz is an iterative *approximation* (`X ← X(2I − AX)`) with a convergence condition attached. This is an exact finite Neumann series evaluated by repeated squaring — exact because strictly lower triangular ⇒ nilpotent ⇒ `S^C = 0`. That distinction seems worth stating, since it's the reason the method is admissible at all, and a reader who reads "Newton–Schulz" will go looking for a tolerance parameter that doesn't exist.

Also, the argument isn't a unit lower triangular matrix — it's the strictly lower part of one. Maybe `invert_unit_lower_triangular(strictly_lower)` with the convention spelled out? And type annotations + Google-style `Args:`/`Returns:` would match `jax_chunk_gated_delta_rule` right below.

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
...
+  # Ensure S is strictly lower triangular (zero out diagonal and upper half)
+  # This guarantees mathematical correctness and stability
+  S_strict = jnp.tril(S, k=-1)
+
+  # Base identity matrix
+  identity = jnp.eye(chunk_size, dtype=S.dtype)
+
+  # Initial approximation and error term
+  A = identity - S_strict
+  E = jnp.tril(S_strict @ S_strict, k=-1)
```
parambole, line 203: Should these matmuls pass `precision=jax.lax.Precision.HIGHEST`? Every other matmul in this file sets it explicitly — `S`, `u_chunks`, `w_chunks`, `attn_inter`, `v_prime`, `update_term`, plus the AR and naive paths -- but let me know if there's a reason.

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
...
+
+  # Log-depth Taylor series exact computation
+  steps = int(math.ceil(math.log2(chunk_size)))
+  for _ in range(steps - 1):
+    # Update inverse and error using batched matmuls
+    A = jnp.tril(A + A @ E)
```
parambole, line 209: I think these `tril` calls may be no-ops ?

src/maxtext/models/qwen3.py
```diff
@@ -170,7 +170,7 @@ def scan_body(prev_state, x):
 
     return new_last_recurrent_state, core_attn_out_i
 
-  final_state, core_attn_out_stacked = jax.lax.scan(scan_body, last_recurrent_state, xs)
+  final_state, core_attn_out_stacked = jax.lax.scan(scan_body, last_recurrent_state, xs, unroll=0)
```
parambole, line 173: Could you please elaborate on unroll=0 ?

nit `unroll=True` over `unroll=0` for readability

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
...
+
+  # Base identity matrix
+  identity = jnp.eye(chunk_size, dtype=S.dtype)
+
+  # Initial approximation and error term
+  A = identity - S_strict
+  E = jnp.tril(S_strict @ S_strict, k=-1)
+
+  # Log-depth Taylor series exact computation
+  steps = int(math.ceil(math.log2(chunk_size)))
```
parambole, line 206: The loop computes S^2, S^4, S^8... Those go to zero eventually since S is nilpotent, but they can get huge first — for identical keys with beta=1 and no decay, S^64 entries hit ~1e37 and f32 caps at 3.4e38. Then `A + A @ E` adds that to an order-1 matrix.

Have you checked max|E| per iteration on real activations? If it stays small this is a non-issue, but the unit test uses ||S|| ~ 0.1 which wouldn't catch it.

src/maxtext/models/qwen3.py
```diff
@@ -180,6 +180,57 @@ def scan_body(prev_state, x):
...
+@functools.partial(jax.named_call, name="invert_triangular_fwd")
+def _invert_unit_lower_triangular_log_depth_fwd(S):
+  A = invert_unit_lower_triangular_log_depth(S)
+  return A, A
+
+
+@functools.partial(jax.named_call, name="invert_triangular_bwd")
+def _invert_unit_lower_triangular_log_depth_bwd(res, g):
+  A = res
+  grad_S = jnp.tril(-(A.mT @ g @ A.mT), k=-1)
```
parambole, line 224: Can you add a comment that dA = -A dS A and the tril here is the VJP of the forward's tril, not a projection tacked on? Also needs precision=HIGHEST.

src/maxtext/models/qwen3.py
```diff
@@ -264,11 +315,11 @@ def to_chunk_scalar(x):
   S = S * jnp.exp(g_diff)
   S = jnp.where(mask, S, 0.0)
 
-  # Inversion (A) - Strictly float32
-  identity = jnp.eye(chunk_size, dtype=jnp.float32)
-  identity_broadcasted = jnp.broadcast_to(identity, S.shape)
+  # Cast to float32 explicitly as you were doing before
+  S = S.astype(jnp.float32)
```
parambole, line 319: No-op — already f32 from the cast above?

parambole (commented):
Main thing I wanted to ask about is numerical stability. The loop computes S^2, S^4, S^8 — those go to zero since S is nilpotent, but I wasn't sure how large they get before they do. Left a comment inline.

Couple of smaller ones inline too, mostly precision= on the new matmuls and the test setup.

PR #4581, opened 2026-07-23
Fix pyconfig strict validation for multimodal keys in text-only runs

SurbhiJainUSC:
# Description

Fix pyconfig strict validation for multimodal keys in text-only runs

The Pydantic configuration validation in `pyconfig.py` was overly strict for text-only workflows (such as RL training). When processing model YAMLs that included multimodal parameters (like `vision_encoder_block` or `image_size_for_vit`), it would crash with a ValueError because those keys are not always in the active schema.

This commit updates `_prepare_for_pydantic` to check the `use_multimodal` flag. If `use_multimodal` is False, the parser now gracefully ignores these specific vision keys and drops them instead of crashing.

FIXES: b/538001451

# Tests

Local run

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -2171,6 +2171,10 @@ class AudioEncoder(BaseModel):
   max_sample_len_for_audio: int = Field(10000, description="Maximum sample length for audio input.")
 
 
+class Multimodal(MultimodalGeneral, VisionTower, VisionProjector, AudioEncoder):
+  """Configurations for multimodal"""
+
+
 class RLCluster(BaseModel):
   """Cluster configurations specific to RL training."""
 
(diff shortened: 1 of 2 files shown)


src/maxtext/configs/types.py
```diff
@@ -2171,6 +2171,10 @@ class AudioEncoder(BaseModel):
   max_sample_len_for_audio: int = Field(10000, description="Maximum sample length for audio input.")
 
 
+class Multimodal(MultimodalGeneral, VisionTower, VisionProjector, AudioEncoder):
+  """Configurations for multimodal"""
```
NuojCheng, line 2175: empty class?

NuojCheng (approved):

igorts-git (approved):

src/maxtext/configs/types.py
SurbhiJainUSC: This class inherits from other classes.

PR #1745, opened 2025-05-15
Update the random key

RissyRan:
# Description

Thanks @gagika for pointing it out with `make_rng` ! Remove the hardcode key. If you are interested in how it [works](https://screenshot.googleplex.com/9QXUk5eRjDpVK9K).

# Tests

Expect test is green.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed.

PR #4587, opened 2026-07-23
[maxtext] Add generalized block-diffusion CFT and SFT

ethannnnnn:
## Motivation

Block-diffusion CFT/SFT needs one explicit definition of target alignment, completion scope, corruption, and loss weights. Reusing segmentation fields for these different meanings can leak clean targets or bias training.

## Scope

- Add a default-off `block_diffusion` training objective while retaining `causal_lm` as the default.
- Emit distinct validity, completion, corruption, and loss masks from the data pipeline.
- Support the public `same_position/all_masked` and `shifted/seed_and_mask` model contracts.
- Consume explicit weighted targets in Linen, NNX, native gradient accumulation, and the SFT adapter.

## Design

Targets remain physically unshifted. A shared scoring utility maps model logits to physical targets according to the configured alignment. Corruption samples independently per eligible block, protects prompts and padding, guarantees progress, and handles partial blocks.

`completion_mask`, `corruption_mask`, and `targets_loss_mask` remain separate through shaped batches and context-parallel reordering. Loss normalization uses the sum of active weights, including under gradient accumulation; an all-zero batch returns finite zero loss and zero gradients.

The Tunix-backed diffusion SFT entry point constructs the stock Tunix `PeftTrainer` and configures it through `configure_diffusion_sft`, a raw-batch adapter, and a target-aligned logits function. MaxText's existing causal-SFT compatibility subclass remains on the causal path only; diffusion does not depend on that overridden train step.

## Compatibility

The causal pipeline keeps its existing trainer selection, next-token shift, segmentation-derived mask, configuration defaults, and checkpoint shape. Diffusion refuses missing explicit masks instead of guessing. Through the paired Tunix weighted-reduction dependency, the Tunix-backed path supports denominator-aware gradient accumulation and evaluation instead of averaging independently normalized microbatches.

## Extensibility

The objective is selected by stable configuration rather than a model name. Additional models can implement one of the declared alignment/canvas contracts or add a new validated contract without changing trainer ownership.

## Tests

- Focused MaxText suite: 196 passed, 14 skipped, 3 deselected, and 66 subtests passed.
- Tunix-backed post-training SFT suite: 11 passed.
- Tests cover partial blocks, prompt protection, deterministic corruption, reordered alignment, strict mask requirements, zero-weight gradients, and causal no-regression.
- Pylint for new scoring/loss tests, Pyink, YAML lint, codespell, Python compilation, and `git diff --check` passed.

## Known limitations

The Tunix-backed SFT path does not support optimizer-state host offload because the underlying `PeftTrainer` does not provide that execution mode. Native MaxText SFT remains available for the same diffusion objective.

## Stack

Depends on the preceding upstream PR: https://github.com/AI-Hypercomputer/maxtext/pull/4586

Cross-repository dependency: https://github.com/google/tunix/pull/1745 (`block-diffusion-tunix-pr3-sft` at `8dc0dd39951a`).

[MaxText block-diffusion design document](https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit)

## Checklist

- [x] Added or updated tests for the changed behavior.
- [x] Ran the relevant local tests and code-quality checks.


src/maxtext/input_pipeline/hf_data_processing.py
```diff
@@ -42,6 +42,40 @@ def _get_pad_id(tokenizer):
...
+    return input_pipeline_utils.BlockDiffusionCorruption(
+        block_size=config.block_diffusion_block_size,
+        mask_id=config.block_diffusion_mask_id,
+        min_noise=config.block_diffusion_min_noise,
+        axis=1,
+        completion_only=bool(use_sft and completion_only),
+        seed_first_token=config.block_diffusion_canvas_policy == "seed_and_mask",
+        include_seed_in_loss=config.block_diffusion_logit_alignment == "shifted",
+    )
+  if objective != "causal_lm":
```
entrpn: instead, use elif for clarity.

ethannnnnn:
Addressed the current review feedback and rebased the complete seven-PR stack onto the latest `main`.

Key changes in this update:
- make causal-vs-diffusion objective dispatch explicit with `elif`;
- metadata fields now zero-pad independently of a tokenizer's token `pad_id`;
- full-target CFT uses target validity, while completion-only SFT uses the explicit completion mask;
- block-diffusion SFT evaluation no longer divides an already averaged loss twice;
- focused regression tests cover nonzero pad IDs, CFT/SFT mask ownership, and eval aggregation.

Validation: 206 focused tests passed, 44 skipped, with 71 parameterized subtests; Pyink, compile, whitespace, and public-text checks passed. This PR retains its documented dependency on Tunix PR #1745.

Design: https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit

Vanilla public reproduction guide: https://docs.google.com/document/d/1rjRSopy888ZqLnGh_KUseW8h1_ZqPGEuSee5MM2Q9ac/edit

@entrpn, could you please take another look?

src/maxtext/input_pipeline/hf_data_processing.py
ethannnnnn: thanks resolved with latest version

PR #4592, opened 2026-07-23
[maxtext] Wire opt-in block-diffusion GRPO

ethannnnnn:
## Motivation

The exact rollout and replay contracts need to enter the existing MaxText/Tunix RL workflow without changing autoregressive GRPO or comparing incompatible probability semantics.

## Scope

- Add explicit, default-off diffusion-RL configuration and fail-closed validation.
- Construct the custom diffusion rollout and prepared scorer only for block-diffusion objectives.
- Route training and evaluation generation through the existing `RLCluster` mode contract.
- Fingerprint rollout/replay semantics in checkpoint metadata.

## Design

MaxText supplies a custom rollout model plus `diffusion_logits_fn` to Tunix. Rollout, live actor, detached reference actor, and the detached start-of-step actor snapshot used as the old policy score one full target-aligned trace. The snapshot remains on the actor role; it is not a separate model role. Training and evaluation have independent sampling configurations and generation contexts tied to global step, so resume does not silently change RNG progression.

Checkpoint metadata covers model, tokenizer, mask token, block size, alignment, canvas policy, temperature, filters, threshold, maximum denoising steps, stop IDs, sequence lengths, and full-trace semantics.

`RLConfig` derives from the complete `MaxTextConfig` model schema and adds only the RL-specific configuration groups. This keeps RL parsing aligned as model fields evolve, without maintaining a partial duplicate list or admitting arbitrary unknown fields.

Block-diffusion CFT/SFT and OPD still require the Hugging Face MaxText input pipeline. The dedicated diffusion-RL rollout uses the RL dataset loader, so it alone is exempt from `dataset_type=hf`; `packing=False` remains mandatory for every block-diffusion objective.

Tunix reshards model variables onto the sampler role mesh, while MaxText modules also retain graph-static sharding objects. The rollout constructor recursively rebinds direct `Mesh` values, `Mesh` or `NamedSharding` values cached in `functools.partial` arguments and keywords, and precomputed `NamedSharding` fields before compilation. It logs the number of rebound module fields. This changes neither parameter values nor partition specs and does not hardcode a model or device ID.

## Compatibility

Autoregressive configuration, vLLM registration, and existing rollout behavior remain unchanged when `rl.diffusion_rollout` is false. Unsupported agentic, packing, MTP, sampling-filter, or execution-mode combinations fail before training.

Full Qwen3 model configuration is validated through the same typed schema used by MaxText model construction; no model-name exception or permissive extra-field mode is introduced. Autoregressive RL still uses the existing configuration and rollout defaults.

Graph-mesh rebinding is confined to the custom in-process diffusion rollout. Existing vLLM and autoregressive rollout construction is unchanged.

## Extensibility

The initial MaxText routing supports GRPO and token-level GSPO. Tunix owns algorithm variants, advantage calculation, and optimizer orchestration; MaxText owns model-aware trace preparation and checkpoint identity.

## Tests

- Full `diffusion_rl_test.py` suite with the fix: 14 passed on two CPU devices.
- A negative control running the incomplete direct-`Mesh` binder deterministically reproduced JAX's incompatible-device error: the direct mesh and model state were on CPU device ID 1 while a cached partial's `NamedSharding` still referenced CPU device ID 0.
- Focused `RLConfig` validation: 14 passed, including the HF-only exemption for the dedicated diffusion-RL loader while preserving the packing rejection.
- A full Qwen3 model YAML and pyconfig instantiated through `RLConfig` with `emb_dim=1024`, 28 layers, `use_mrope=false`, and `use_qk_norm=true`.
- An AST audit covered 98 Qwen/model-layer configuration attributes with zero missing from `RLConfig`.
- Coverage also includes default-off AR construction, train/eval mode routing, checkpoint fingerprints, invalid combinations, and lazy vLLM isolation.
- The exact final Tunix CPU gate completed 119 tests in 107.08 seconds, with
  six optional vLLM/SGL-JAX integration tests deselected. A 4x4x4 TPU
  validation then completed on all 16 workers with
  zero restarts and all three configured actor updates. Losses and rewards
  were finite, importance-sampling ratio was 1, clip fraction was 0, and
  gradient norms were nonzero. The actor step-3 checkpoint finalized and the
  downstream model export was verified. This is correctness smoke evidence,
  not a throughput or model-quality benchmark.
- Pyink 24.10.1, Ruff, and `git diff --check` passed for the sampler-mesh change; the earlier changed-file Pylint, compilation, and YAML checks remain green.

## Known limitations

MaxText does not yet expose DAPO or DrGRPO routing for this path. The current implementation requires Pathways single-controller execution and does not support packing or agentic rollout. Recursive graph-static sharding rebinding is covered locally and the three-update TPU smoke completed with checkpoint/export verification, but broader throughput, memory scaling, convergence, and model-quality validation remain open.

## Stack

Depends on the preceding upstream PR: https://github.com/AI-Hypercomputer/maxtext/pull/4591

Cross-repository dependency: https://github.com/google/tunix/pull/1749 (`block-diffusion-tunix-pr7-policy-optimization` at `264e9f88b374`).

[MaxText block-diffusion design document](https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit)

## Checklist

- [x] Added or updated tests for the changed behavior.
- [x] Ran the relevant local tests and code-quality checks.


src/maxtext/configs/types.py
```diff
@@ -3782,48 +3927,13 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
-    DataTypes,
-    Tokenizer,
-    AdamW,
-    Optimizer,
-    Quantization,
-    # Debugging and Profiling
-    DevelopmentAndDebugging,
-    Profiling,
-    # For compatibility with trainer in post_train/rl
+    MaxTextConfig,
```
entrpn, line 3927: why all these fields removed? @hengtaoguo PTAL.

src/maxtext/configs/types.py
```diff
@@ -1611,7 +1644,34 @@ class Distillation(BaseModel):
   )
 
 
-class TrainingLoop(BaseModel):
+class BlockDiffusionObjective(BaseModel):
```
entrpn: @hengtaoguo can you help me review these changes.

src/maxtext/input_pipeline/hf_data_processing.py
```diff
@@ -321,12 +356,15 @@ def preprocessing_pipeline(
...
   operations = []
   if use_sft:
+    is_block_diffusion = getattr(config, "training_objective", "causal_lm") == "block_diffusion"
     operations.append(
         input_pipeline_utils.SFTPromptMasking(
             text_column_name=data_column_names[0],
             completion_only=sft_train_on_completion_only,
             max_target_length=max_target_length,
             unk_id=pad_id,
+            target_aligned=is_block_diffusion,
```
entrpn: SFTPromptMasking should accept the training_objective string directly. The transform itself should internally determine if it needs to align targets or emit completion masks based on the objective, keeping the pipeline construction clean.

src/maxtext/diffusion/denoise.py
```diff
@@ -0,0 +1,348 @@
...
+  for name, value in (
+      ("positions", positions),
+      ("validity_mask", validity_mask),
+      ("completion_mask", completion_mask),
+  ):
+    if tuple(value.shape) != expected_shape:
+      raise ValueError(f"{name} must match initial_tokens shape; received {tuple(value.shape)} and {expected_shape}")
+
+
+def _concrete_numpy(value):
```
entrpn: create a utility file and add _concrete_numpy as it is duplicated multiple times in the code.

src/maxtext/diffusion/block_diffusion/denoise.py
```diff
@@ -0,0 +1,348 @@
+# Copyright 2026 Google LLC
```
entrpn, line 1: the class name is to ambiguous. Since diffusion is broad (continuous/discrete), add a subfolder for block diffusion and add the block diffusion specific classes in there. Rename it to something like `BlockDiffusionRolloutTrace` or more broad `DiscreteUnmaskTrace`.

src/maxtext/diffusion/block_diffusion/target_alignment.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2026 Google LLC
```
entrpn, line 1: like denoise.py, the class name is very broad, add it under a block_diffusion folder and rename it to something like `block_alignment` or `target_alignment`.

src/maxtext/trainers/post_train/sft/hooks.py
```diff
@@ -29,7 +29,12 @@ class SFTTrainingHooks(BaseTrainingHooks):
   @override
   def get_total_weights(self, batch) -> jax.Array:
     """Calculate the number of non-padded tokens in the batch."""
-    return jnp.sum(batch["targets_segmentation"] != 0)
+    return jnp.sum(batch.get("targets_loss_mask", batch["targets_segmentation"]) != 0)
+
+  @override
+  def eval_loss_is_preaveraged(self) -> bool:
```
entrpn, line 35: loss preaveraging should be decoupled from block_diffusion, instead add a new config flag  loss_is_preaveraged so that it can be used more broadly.

src/maxtext/layers/attention_op.py
```diff
@@ -1409,13 +1498,24 @@ def create_sa_config(config, query, key, attn_logits_soft_cap):
...
       mask_module = tokamax_splash_mask if self.config.use_tokamax_splash else splash_attention_mask
+      use_load_balanced_cp = cp_size > 1 and load_balanced_context_parallel
       if self.attention_type == AttentionType.FULL:
         mask = mask_module.FullMask(mask_shape)
+      elif self.attention_type == AttentionType.BLOCK_DIFFUSION:
+        mask_type = LoadBalancedBlockCausalMask if use_load_balanced_cp else BlockCausalMask
+        mask_kwargs = {"cp_size": cp_size} if use_load_balanced_cp else {}
+        mask = mask_type(
+            shape=mask_shape,
+            block_size=self.block_diffusion_block_size,
```
entrpn: rename this to causal_block_size to make it more generic.

ethannnnnn:
Addressed the current review feedback and rebased the complete seven-PR stack onto the latest `main`.

Key changes in this update:
- move discrete denoising and target alignment under the explicit `maxtext.diffusion.block_diffusion` package;
- rename the trajectory type to `BlockDiffusionRolloutTrace` and `scoring.py` to `target_alignment.py`;
- replace duplicated eager-array conversion helpers with one shared utility;
- make `SFTPromptMasking` derive target alignment and completion-mask behavior from `training_objective`;
- expose the generic `causal_block_size` contract throughout attention, training, OPD, and RL;
- decouple weighted eval semantics from diffusion through the explicit `loss_is_preaveraged` config flag;
- keep `TrainingLoop` directly based on `BaseModel` and group the diffusion fields without adding a nested hierarchy;
- document and test that `RLConfig` inherits the complete `MaxTextConfig` schema rather than removing fields;
- resolve cumulative Pylint findings and add the repository-required PR checklists.

Validation: 209 focused tests passed, 44 skipped, with 71 parameterized subtests; cumulative Pylint scored 10.00/10; Pyink, compilation, whitespace, and public-text checks passed. This PR retains its documented dependency on Tunix PR #1745.

Design: https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit

Vanilla public reproduction guide: https://docs.google.com/document/d/1rjRSopy888ZqLnGh_KUseW8h1_ZqPGEuSee5MM2Q9ac/edit

@entrpn @hengtaoguo, could you please take another look?

src/maxtext/configs/types.py
hengtaoguo: Great catch, I would prefer not to change the `TrainingLoop` from inheriting `BaseModel` to `BlockDiffusionObjective`.

I understand this consolidated design choice, but introducing a nested hierarchy only for block-diffusion sets an unusual example. I would recommend author to either flatten these fields directly into TrainingLoop, or use sibling mixin. It keeps the configuration aligned with the current style. You could add a comment line to indicate such fields are for block diffusion purpose.

src/maxtext/configs/types.py
hengtaoguo: Based on the updated doc string below, it says inheriting `MaxTextConfig` and RL-specific mixins will cover all the above deleted args, so the pre/post-training won't be affected. Looks reasonable based on the design, but I would recommend author to double check whether the config was impacted or not.

PR #4593, opened 2026-07-23
Add custom MLP vision projector and tests

subawocit:
# Description

This PR adds support for configurable custom vision projectors in MaxText's multimodal models. It introduces new vision projector config options, handles checkpoint weight mismatches when attaching custom projector architectures to pre-trained base models, and adds a test script to verify projector initialization and embedding extraction.

Step 2 of a 5-step Proof-of-Concept for any-to-any multimodal alignment in MaxText. Overall goal is to align a pretrained vision encoders with another text-only LLM, and connecting the two with a dynamically configured adaptation layer (customized MLP connector), special tokenizer mapping for visual placeholder tokens, and train only the MLP using supervised fine tuning.

# Changes

### `src/maxtext/layers/encoders.py`
- Implemented `MultimodalMLPProjector` (`nnx.Module`), supporting configurable layer, activations, hidden size, and layer biases.
- Updated `VisionEncoder._setup_vision_encoder_layers()` to conditionally override the default vision projector with `MultimodalMLPProjector` whenever `vision_projector_type == "customized_vision_projector"`.

### `src/maxtext/configs/types.py`
- Extended the `VisionProjector` configuration class for customizing projector architecture.

### `src/maxtext/common/checkpointing.py`
- Added `_is_custom_projector_problem()` to check if missing/mismatched weight keys belong to custom vision projector layers.
- Updated `_raise_on_weight_mismatch()` to suppress weight mismatch errors for custom projector layers only, and raise a warning instead. This allows restoring base model parameters from standard checkpoints while leaving newly introduced custom projector weights randomly initialized.

### `src/maxtext/utils/model_creation_utils.py`
- Modified `_free_device_memory` inside `from_pretrained()`.
- Added path checks (`is_custom`) to skip deleting JAX arrays corresponding to custom layers (i.e., layers with "custom_linear" in their name).
  - Reason: Pretrained base model checkpoints do not contain weights for new custom components; if their device memory were freed prior to checkpoint restoration, their initializations would be deleted with no incoming checkpoint weights to replace them. Skipping deletion of these arrays ensures their randomly initialized state remains alive in device memory when base model parameters are loaded.

### `src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py`
- Initializes a multimodal model (`qwen3-vl-2b`) with a customized MLP vision projector.
- Checks output embeddings against standard configurations.
### `tests/unit/model_creation_utils_test.py`
- Tests the model creation utility with newly added customized projector parts.

# Example Usage
```python
config = pyconfig.initialize(
    model_name="qwen3-vl-2b",
    tokenizer_path="Qwen/Qwen3-VL-2B-Instruct",
    tokenizer_type="huggingface",
    load_parameters_path=<PATH_TO_QWEN3_CHECKPOINT>,
    use_multimodal=True,
    prompt="Describe this image",
    image_path="tests/assets/test_image.jpg",
    override_model_config=True,
    hf_access_token=<HF_ACCESS_TOKEN>,
    vision_projector_type="customized_vision_projector",
    vision_connector_num_layers=2,
    vision_connector_activation="gelu",
    vision_connector_use_bias=True,
    ... # other standard multimodal arguments
    )
```

# Tests

## 1. Unit Tests

**Command:**
```bash
pytest tests/unit/model_creation_utils_test.py
```

**Output:**
```
============================= 63 passed in 45.95s ==============================
```

**Command:**
```bash
pytest tests/unit/checkpointing_nnx_load_test.py
```

**Output:**
```
=============================== 24 passed in 6.82s ===============================
```

## 2. Custom Vision Projector Verification
Validates that vision embeddings are:
- same between the standard model and custom-projector model before the projector layer, 
- differ after the projector layer due to randomly initialized projector.

**Command:**
```bash
python3 src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py \
    --load_parameters_path=<your-path> \
    --hf_access_token=<your-token>
```

**Output:**
```
Image pixel_values shape: (1, 3, 2, 544, 640)
vision embeddings: (1, 1360, 1024)
First 10 values:
[-5.091886    1.7075098  -1.0072641  -0.16636705 -1.4430853  -5.3355365
  1.4629428   0.21186987 -2.9909272  -4.376259  ]
projector_embeddings: (1, 340, 2048)
First 10 values:
[  1.3062916   8.224888   16.922318   -7.4536076  -8.535779   33.199806
   3.0147345   2.200746  -18.904125    9.107975 ]

Original vision embeddings: (1, 1360, 1024)
First 10 values:
[-5.091886    1.7075098  -1.0072641  -0.16636705 -1.4430853  -5.3355365
  1.4629428   0.21186987 -2.9909272  -4.376259  ]
Original projector_embeddings: (1, 340, 2048)
First 10 values:
[-0.31920713  0.54748684  0.49574944 -0.88344455 -0.23951492 -0.14456059
  0.1383374   0.05032197  0.381779    0.00628588]

===================================
Comparing customized vs original embeddings
Vision embeddings match: True (Max diff: 0.0)
Projector embeddings match: False (Max diff: 313.9195556640625)
```

### A. Baseline Multimodal Run (Default Vision Projector)
Verifies standard multimodal inference functions normally.

**Command:**
```bash
python3 -m maxtext.inference.decode \
    src/maxtext/configs/base.yml \
    model_name=qwen3-vl-2b \
    tokenizer_path=Qwen/Qwen3-VL-2B-Instruct \
    tokenizer_type=huggingface \
    load_parameters_path=<your-path> \
    per_device_batch_size=1 \
    scan_layers=false \
    use_multimodal=true \
    prompt='Describe this image' \
    image_path='tests/assets/test_image.jpg' \
    max_prefill_predict_length=512 \
    max_target_length=768 \
    ici_tensor_parallelism=4 \
    override_model_config=true \
    attention='dot_product' \
    hf_access_token=<your-token>
```

**Output:**
```
Input `<|im_start|>user
<|vision_start|><|image_pad|><|vision_end|>Describe this image<|im_end|>
<|im_start|>assistant
` -> `This is a panoramic view of the Seattle skyline on a bright, sunny day. The image is taken from a high vantage point, looking down on the city.

The most prominent feature is the **Space Needle**, a distinctive observation tower located on the left side of the frame. It stands tall among the city's modern skyscrapers, which are a mix of glass and steel structures. The buildings are densely packed, creating a dense urban landscape.

In the background, a range of mountains is visible, with their peaks covered in snow, suggesting a high elevation. The sky is a clear, vibrant blue with a few scattered white clouds.

In the foreground, there are lush green trees and foliage, which add a natural element to the urban scene. The overall impression is one of a vibrant, modern city with a beautiful natural backdrop.
```

### B. Custom Vision Projector Pipeline Run
Verifies end-to-end data flow with custom projector flags (`customized_vision_projector`). Output text is bad due to random weight initialization, as expected.

**Command:**
```bash
python3 -m maxtext.inference.decode \
    src/maxtext/configs/base.yml \
    model_name=qwen3-vl-2b \
    tokenizer_path=Qwen/Qwen3-VL-2B-Instruct \
    tokenizer_type=huggingface \
    load_parameters_path=<your-path> \
    per_device_batch_size=1 \
    run_name=runner_image_2026-06-26-20-40 \
    scan_layers=false \
    use_multimodal=true \
    prompt='Describe this image' \
    image_path='tests/assets/test_image.jpg' \
    max_prefill_predict_length=512 \
    max_target_length=768 \
    ici_tensor_parallelism=4 \
    override_model_config=true \
    attention='dot_product' \
    hf_access_token=<your-token> \
    vision_projector_type=customized_vision_projector \
    vision_connector_num_layers=2 \
    vision_connector_activation=gelu \
    vision_connector_use_bias=true
```

**Output:**
```
Input `<|im_start|>user
<|vision_start|><|image_pad|><|vision_end|>Describe this image<|im_end|>
<|im_start|>assistant
` -> `The image is a detailed and intricate depiction of a complex, multi-layered, and highly detailed scene that appears to be a digital or conceptual representation rather than a physical photograph...
```

## 4. Text-Only Decoding Regression Check
Verifies custom vision projector configurations do not affect text-only inference mode (`use_multimodal=false`).

**Command:**
```bash
python3 -m maxtext.inference.decode \
    src/maxtext/configs/base.yml \
    model_name=qwen3-vl-2b \
    tokenizer_path=Qwen/Qwen3-VL-2B-Instruct \
    tokenizer_type=huggingface \
    load_parameters_path=<your-path> \
    per_device_batch_size=1 \
    max_target_length=256 \
    scan_layers=false \
    use_multimodal=false \
    prompt='How many times does the letter r appear in the word Strawberry?' \
    override_model_config=true \
    ici_fsdp_parallelism=1 \
    ici_tensor_parallelism=-1 \
    hf_access_token=<your-token> \
    vision_projector_type=customized_vision_projector \
    vision_connector_num_layers=2 \
    vision_connector_activation=gelu \
    vision_connector_use_bias=true
```

**Output:**
```
Input `How many times does the letter r appear in the word Strawberry?` -> ` To determine how many times the letter **r** appears in the word **Strawberry**, let's go step by step.

1. Write down the word: **Strawberry**
2. Break it down into individual letters:
   - S - t - r - a - w - b - e - r - r - y
3. Now, count how many times the letter **r** appears:
   - The first **r** is in the 3rd position.
   - The second **r** is in the 8`
```

# Step-by-Step Objectives
- `√` Multi-Directory Checkpoint Restoration: Implemented selective sub-tree parameter loading using Orbax such that we can initiaze a multimodal model with parameters from multiple checkpoints.
- `this` Dynamic MLP Connector: Add omni modal adapter layer (MLP) to connect vision tower output to the LLM decoder
- `next` Special Tokenizer & Placeholder Masking: Add special token <|image|> to tokenizer and make sure the masking is correct in the decoder.
- COCO-Narratives Data Pipeline Processing
- SFT Execution & Evaluation

# Checklist

### DIFF
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -107,6 +107,22 @@ def _expected_and_restored_params(abstract_nnx_state, restored_linen):
   return want, have
 
 
+def _is_custom_projector_problem(path: str, want: dict) -> bool:
+  """Returns True if a weight mismatch belongs to a newly attached custom vision projector."""
+  parts = [p for p in path.replace(".", "/").split("/") if p and p != "params"]
+  if len(parts) >= 2 and parts[0] == "vision_encoder":
+    proj_name = parts[1]
+    proj_dict = want.get("vision_encoder", {}).get(proj_name, {})
+    if isinstance(proj_dict, dict) and any("custom_linear" in k for k in proj_dict.keys()):
+      max_logging.warning(
+          f"===Warning: weight mismatch found in custom vision projector: {proj_name}.\n"
+          f"Path: {path}\n"
+          "This custom vision projector will be initialized with random weights.==="
+      )
+      return True
+  return False
+
+
 def _raise_on_weight_mismatch(want, have, config=None):
   """Raises if the restored weights (`have`) don't match what the model expects (`want`).
 
@@ -120,6 +136,8 @@ def _raise_on_weight_mismatch(want, have, config=None):
     want = _filter_lora_trainable_state(want)
 
   problems = _weight_mismatches(want, have)
+  # Ignore the weight mismatches in the custom projector so it can stay randomly initialized
+  problems = [(p, why) for p, why in problems if not _is_custom_projector_problem(p, want)]
   if not problems:
     return
   lines = "\n".join(f"  - '{p}': {why}" for p, why in problems)
--- a/src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py
+++ b/src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py
@@ -1,0 +1,166 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for custom vision projector implementation in MaxText.
+
+"""
+
+import argparse
+import gc
+import os
+
+from flax import nnx
+import jax
+import jax.numpy as jnp
+
+from maxtext.configs import pyconfig
+from maxtext.inference.maxengine import maxengine
+from maxtext.multimodal import processor as mm_processor
+from maxtext.utils.globals import MAXTEXT_PKG_DIR
+
+
+def test_load_custom_vision_projector(hf_access_token, param_path):
+  base_yml_path = os.path.join(MAXTEXT_PKG_DIR, "configs", "base.yml")
+
+  config = pyconfig.initialize(
+      base_config=base_yml_path,
+      model_name="qwen3-vl-2b",
+      tokenizer_path="Qwen/Qwen3-VL-2B-Instruct",
+      tokenizer_type="huggingface",
+      load_parameters_path=param_path,
+      per_device_batch_size=1,
+      scan_layers=False,
+      use_multimodal=True,
+      prompt="Describe this image",
+      image_path="tests/assets/test_image.jpg",
+      max_prefill_predict_length=512,
+      max_target_length=768,
+      ici_tensor_parallelism=4,
+      override_model_config=True,
+      attention="dot_product",
+      hf_access_token=hf_access_token,
+      vision_projector_type="customized_vision_projector",
+      vision_connector_num_layers=2,
+      vision_connector_activation="gelu",
+      vision_connector_use_bias=True,
+  )
+
+  print("Customized model")
+  print(f"Active Vision Projector Type: {config.vision_projector_type}")
+  print(f"Num Layers: {config.vision_connector_num_layers}")
+
+  engine = maxengine.MaxEngine(config)
+
+  # Load the parameters from the checkpoint
+  rng = jax.random.PRNGKey(1234)
+  params = engine.load_params(rng=rng)
+  nnx.update(engine.model, params)
+
+  # Preprocess the example image using the provided image_path in the config
+  processor_outputs = mm_processor.preprocess_mm_data(config)
+
+  images = processor_outputs.pixel_values
+  print(f"Image pixel_values shape: {images.shape}")
+
+  # Extract the vision wrapper and its underlying parts
+  vision_wrapper = engine.model.vision_encoder
+  pure_encoder = getattr(vision_wrapper, vision_wrapper.encoder_name)
+  projector = getattr(vision_wrapper, vision_wrapper.projector_name)
+
+  # Pass the image through the pure vision tower (before projector)
+  encoder_output = pure_encoder(images)
+
+  vision_embeddings_customized = encoder_output[0] if isinstance(encoder_output, tuple) else encoder_output
+
+  print("vision embeddings:", vision_embeddings_customized.shape)
+  print(f"First 10 values:\n{vision_embeddings_customized.flatten()[:10]}")
+
+  # Pass the image through the vision tower (including projector)
+  projected_embeddings_customized = projector(vision_embeddings_customized)
+
+  print(f"projector_embeddings: {projected_embeddings_customized.shape}")
+  print(f"First 10 values:\n{projected_embeddings_customized.flatten()[:10]}")
+
+  del engine, params, vision_wrapper, pure_encoder, projector
+  gc.collect()
+
+  print("\nOriginal model")
+  config_original = pyconfig.initialize(
+      base_config=base_yml_path,
+      model_name="qwen3-vl-2b",
+      tokenizer_path="Qwen/Qwen3-VL-2B-Instruct",
+      tokenizer_type="huggingface",
+      load_parameters_path=param_path,
+      per_device_batch_size=1,
+      scan_layers=False,
+      use_multimodal=True,
+      prompt="Describe this image",
+      image_path="tests/assets/test_image.jpg",
+      max_prefill_predict_length=512,
+      max_target_length=768,
+      ici_tensor_parallelism=4,
+      override_model_config=True,
+      attention="dot_product",
+      hf_access_token=hf_access_token,
+  )
+
+  print(f"Original Vision Projector Type: {getattr(config_original, 'vision_projector_type', None)}")
+
+  engine_original = maxengine.MaxEngine(config_original)
+
+  params_original = engine_original.load_params(rng=rng)
+  nnx.update(engine_original.model, params_original)
+
+  processor_outputs_original = mm_processor.preprocess_mm_data(config_original)
+  images_original = processor_outputs_original.pixel_values
+
+  vision_wrapper_original = engine_original.model.vision_encoder
+  pure_encoder_original = getattr(vision_wrapper_original, vision_wrapper_original.encoder_name)
+  projector_original = getattr(vision_wrapper_original, vision_wrapper_original.projector_name)
+
+  encoder_output_original = pure_encoder_original(images_original)
+
+  vision_embeddings_original = (
+      encoder_output_original[0] if isinstance(encoder_output_original, tuple) else encoder_output_original
+  )
+
+  print("Original vision embeddings:", vision_embeddings_original.shape)
+  print(f"First 10 values:\n{vision_embeddings_original.flatten()[:10]}")
+
+  projected_embeddings_original = projector_original(vision_embeddings_original)
+
+  print(f"Original projector_embeddings: {projected_embeddings_original.shape}")
+  print(f"First 10 values:\n{projected_embeddings_original.flatten()[:10]}")
+
+  print("\n===================================")
+  print("Comparing customized vs original embeddings")
+  vision_embeddings_match = jnp.allclose(vision_embeddings_customized, vision_embeddings_original, atol=1e-5)
+  vision_max_diff = jnp.max(jnp.abs(vision_embeddings_customized - vision_embeddings_original))
+  print(f"Vision embeddings match: {vision_embeddings_match} (Max diff: {vision_max_diff})")
+
+  projector_embeddings_match = jnp.allclose(projected_embeddings_customized, projected_embeddings_original, atol=1e-5)
+  projector_max_diff = jnp.max(jnp.abs(projected_embeddings_customized - projected_embeddings_original))
+  print(f"Projector embeddings match: {projector_embeddings_match} (Max diff: {projector_max_diff})")
+
+
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser()
+  parser.add_argument("--load_parameters_path", required=True)
+  parser.add_argument("--hf_access_token", default=os.getenv("HF_TOKEN", ""))
+
+  args = parser.parse_args()
+  test_load_custom_vision_projector(
+      hf_access_token=args.hf_access_token,
+      param_path=args.load_parameters_path,
+  )
--- a/src/maxtext/layers/encoders.py
+++ b/src/maxtext/layers/encoders.py
@@ -21,6 +21,7 @@ from jax.sharding import Mesh
 from maxtext.common.common_types import Config, VisionEncoderBlockType
 from maxtext.layers import nnx_wrappers
 from maxtext.layers import initializers
+from maxtext.layers import linears
 
 
 class VisionEncoder(nnx.Module):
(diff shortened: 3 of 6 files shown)


src/maxtext/layers/encoders.py
```diff
@@ -91,14 +87,25 @@ def _setup_vision_encoder_layers(self):
...
       supported_blocks = [block.value for block in VisionEncoderBlockType if block != VisionEncoderBlockType.NONE]
       raise ValueError(
           f"Unsupported vision_encoder_block={self.vision_encoder_block.value!r} "
           f"for model_name={self.config.model_name!r}. Supported values are: {supported_blocks}."
       )
 
+    # If vision_projector_type is explicitly set to 'customized_mlp',
+    # override the model's default projector with a custom MLP-based projector.
+    vision_projector_type_config = getattr(self.config, "vision_projector_type", "default")
+    if vision_projector_type_config == "customized_mlp":
```
hengtaoguo: Maybe change this name to "customized_vision_projector"?

Because later we will have audio components and prefer explicit naming pattern.

src/maxtext/common/checkpointing.py
```diff
@@ -98,6 +114,7 @@ def _raise_on_weight_mismatch(want, have):
   without naming the weight.
   """
   problems = _weight_mismatches(want, have)
+  problems = [(p, why) for p, why in problems if not _is_custom_vision_projector_problem(p, want)]
```
hengtaoguo: Add a one-line comment explaining this line, such as "if the problem is caused by customized projector, ignore/remove this problem" if I understand correctly.

src/maxtext/common/checkpointing.py
```diff
@@ -88,6 +88,22 @@ def _expected_and_restored_params(abstract_nnx_state, restored_linen):
   return want, have
 
 
+def _is_custom_vision_projector_problem(path: str, want: dict) -> bool:
```
hengtaoguo: Rename to `_is_custom_projector_problem` to be more generic? (we could reuse for audio projector too in the future)

tests/unit/model_creation_utils_test.py
```diff
@@ -711,6 +711,32 @@ def test_load_linen_checkpoint(self, mock_ocp):
...
+    mock_ckptr.metadata.return_value = self._make_nnx_metadata_mock()
+    mock_ckptr.restore.side_effect = lambda path, item=None, **kw: item
+    mock_ocp.Checkpointer.return_value = mock_ckptr
+    mock_ocp.PyTreeCheckpointHandler.return_value = MagicMock()
+    mock_ocp.checkpoint_utils.construct_restore_args.return_value = {}
+    mock_ocp.ArrayRestoreArgs = ocp.ArrayRestoreArgs
+
+    cfg = _make_config(
+        enable_checkpointing=True,
+        load_parameters_path="gs://fake/custom_vision_ckpt",
```
hengtaoguo, line 727: What is this "gs://fake/custom_vision_ckpt"? If it's actually unused, can we possibly remove it from the config?

src/maxtext/layers/encoders.py
```diff
@@ -127,6 +134,83 @@ def __call__(self, input_images, input_masks=None, video_grid_thw=None, determin
     return embeddings, deep_feats
 
 
+class MultimodalMLPProjector(nnx.Module):
```
hengtaoguo, line 137: I like this customizable block! Is it possible to reuse for audio tower in the future too? Asking because you named it `Multimodal` instead of `Vision`, which is great.

src/maxtext/utils/model_creation_utils.py
```diff
@@ -1105,12 +1111,12 @@ def _free_device_memory(node):
...
             if isinstance(leaf, jax.Array) and not leaf.is_deleted():
               leaf.delete()
-        elif isinstance(node, jax.Array) and not node.is_deleted():
+        elif isinstance(node, jax.Array) and not node.is_deleted() and not is_custom:
           node.delete()
 
         return node
 
-      jax.tree_util.tree_map(_free_device_memory, sharded_state, is_leaf=lambda n: isinstance(n, nnx.Variable))
+      jax.tree_util.tree_map_with_path(_free_device_memory, sharded_state, is_leaf=lambda n: isinstance(n, nnx.Variable))
```
hengtaoguo, line 1119: Why is this change needed?

hengtaoguo (changes requested):

src/maxtext/utils/model_creation_utils.py
subawocit: Standard `tree_map()` passes only the raw tensor (node). A tensor has no name so it's very difficult to distinguish a base model weight from a custom projector weight. We want to identify the custom projector layer so it doesnt get deleted during free memory. So I switched to `tree_map_with_path()` which passes pytree keys (like "custom_linear_0"), and I can filter by the name, skip freeing memory in custom_linear, and keep it randomly initialized

reference: https://docs.jax.dev/en/latest/_autosummary/jax.tree.map_with_path.html

src/maxtext/layers/encoders.py
subawocit: Yes, one just need to 1. define the input/output dimensions based on the audio config (L146), 2. define the connector hyperparameters (similar to L150-155), and 3. wire it up in `AudioEncoder._setup_audio_encoder_layers` 

This should be very straightforward

tests/unit/model_creation_utils_test.py
subawocit: It is a dummy path required to test checkpoint loading related functions. The same flag was passed in many other unit tests (e.g., [L691](https://github.com/AI-Hypercomputer/maxtext/blob/40aaa08f9a611f3781a785840da9c1ad22477c81/tests/unit/model_creation_utils_test.py#L691)). I was wondering why this is added throughout the unit tests so I added some print statements in the `_free_device_memory` during debugging, and without this load_parameters_path="gs://fake/custom_vision_ckpt", `_free_device_memory` never gets executed in unit tests.

src/maxtext/common/checkpointing.py
subawocit: Agree - updated the code

src/maxtext/common/checkpointing.py
subawocit: Updated

src/maxtext/layers/encoders.py
subawocit: Updated - I also changed the names in PR description accordingly

src/maxtext/common/checkpointing.py
```diff
@@ -88,6 +88,22 @@ def _expected_and_restored_params(abstract_nnx_state, restored_linen):
...
 
 
+def _is_custom_projector_problem(path: str, want: dict) -> bool:
+  """Returns True if a weight mismatch belongs to a newly attached custom vision projector."""
+  parts = [p for p in path.replace(".", "/").split("/") if p and p != "params"]
+  if len(parts) >= 2 and parts[0] == "vision_encoder":
+    proj_name = parts[1]
+    proj_dict = want.get("vision_encoder", {}).get(proj_name, {})
+    if isinstance(proj_dict, dict) and any("custom_linear" in k for k in proj_dict.keys()):
+      print(
```
NuojCheng: use [maxtext logging](https://source.corp.google.com/piper///depot/google3/third_party/py/maxtext/src/maxtext/utils/max_logging.py;rcl=862910344;l=36)

src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py
```diff
@@ -0,0 +1,156 @@
...
+import gc
+import os
+import sys
+
+from flax import nnx
+import jax
+import jax.numpy as jnp
+
+# Ensure the parent directory (src/) is in sys.path so we can import maxtext
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../..")))
```
NuojCheng: I don't think we need this. Is the maxtext installed from source properly? Follow [here](https://maxtext.readthedocs.io/en/maxtext-v0.2.3/install_maxtext.html)

src/maxtext/layers/encoders.py
```diff
@@ -127,6 +134,83 @@ def __call__(self, input_images, input_masks=None, video_grid_thw=None, determin
...
+      current_in = current_out
+
+  def __call__(self, x: jax.Array) -> jax.Array:
+    # for qwen3 models, concatenate tokens per block (e.g. 2x2 spatial patches) along feature dimension
+    if self.tokens_per_block > 1 and x.ndim == 3 and x.shape[1] % self.tokens_per_block == 0:
+      batch_size, seq_len, in_dim = x.shape
+      num_blocks = seq_len // self.tokens_per_block
+      x = x.reshape((batch_size, num_blocks, self.tokens_per_block * in_dim))
+
+    for i in range(self.num_layers):
```
NuojCheng, line 206: in JAX we usually use `jax.lax.scan` instead of for-loop for faster compilation and potentially better performance

tests/unit/model_creation_utils_test.py
NuojCheng: maybe use `enable_checkpointing=False`?

tests/unit/model_creation_utils_test.py
```diff
@@ -711,6 +711,32 @@ def test_load_linen_checkpoint(self, mock_ocp):
...
+        use_multimodal=True,
+        model_name="qwen3-vl-2b",
+        override_model_config=True,
+        scan_layers=False,
```
NuojCheng, line 731: why the scan is not used?

subawocit: `qwen3-vl-2b` uses Deepstack visual features, which strictly requires `scan_layers=False`. If set True, it will raise a qwen3-vl specific error: 
```
E     pydantic_core._pydantic_core.ValidationError: 1 validation error for MaxTextConfig
E       Value error, Deepstack visual embedding injection requires scan_layers=False. 
Set scan_layers=False in your config to use deepstack features. 
[type=value_error, input_value={'run_name': 'test', 'mod...xp_with_fsdp', 'fsdp']]}, input_type=dict]
```

I also tested `model_name=="gemma3-4b"` and `scan_layers=True`, the unit test passed.

src/maxtext/common/checkpointing.py
subawocit: Updated

tests/unit/model_creation_utils_test.py
subawocit: Just tried - Same result: `_free_device_memory` was never executed

src/maxtext/experimental/omni_poc/tests/custom_vision_projector_test.py
subawocit: Good catch - it was a leftover from my pilot file, I've removed the line and tested this file again

src/maxtext/layers/encoders.py
subawocit: `jax.lax.scan` requires all scanned iterations to have identical weight shapes, whereas the custom projector has input/output shape varying across layers (e.g., first layer is `visual_dim -> hidden_dim`, second is `hidden_dim -> text_dim`). I tried to replace the for loop with `jax.lax.scan` but encountered many errors. (But please correct me if I was wrong)

NuojCheng: you are right and I did not notice that. Would jax.lax.fori_loop work here?

tests/unit/model_creation_utils_test.py
NuojCheng: I think it is worth investigating why it happenes (not necessarily in this PR).

hengtaoguo (approved):

src/maxtext/layers/encoders.py
subawocit: Good point, I just tried it and similar errors occurred. Reading through the [jax doc](https://docs.jax.dev/en/latest/_autosummary/jax.lax.fori_loop.html), my understanding is that `fori_loop` has a very similar requirement that all weight shapes need to be identical, but our input/output shapes are different by design

NuojCheng (approved):

PR #4604, opened 2026-07-24
Add axis inference logic and disable EP flags when EP is not active

NuojCheng:
# Description

This PR adds a mechanism to infer the context and expert sharding physical axis names, we use 
* contents in logical name "activation_length" to infer context sharding physical axis name
* contents in "exp" to infer expert sharding physical axis name

Through this mechanism, we are able to generate physical axes names in the config initialization period and use them to get their ranks. Their ranks are used to determine certain flags should be disabled.

When EP is enabled, the following flags should be default values, otherwise ValueError throw

* use_random_routing
* use_ragged_sort
* ragged_buffer_factor
* use_ring_of_experts

FIXES: b/538596542

# Tests

- Regular EP check: https://paste.googleplex.com/5099839552618496
- Use CP-as-EP: https://paste.googleplex.com/5558558300176384
- Use EP-as-CP: https://paste.googleplex.com/4597690464862208

# Checklist


tests/unit/pyconfig_test.py
```diff
@@ -341,6 +342,72 @@ def test_eval_start_step_negative_raises_error(self):
...
+
+  def test_ep_rank_1_no_disabling(self):
+    """EP rank=1 -> no flags should be affected."""
+    ep_axes = infer_ep_axes([["exp", "expert"]])
+    ep_rank = 1
+    for _ in ep_axes:
+      ep_rank *= 1  # parallelism = 1
+    self.assertEqual(ep_rank, 1)
+
+  def test_normalize_axes_basics(self):
```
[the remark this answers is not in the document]
NuojCheng: Updated

src/maxtext/configs/types.py
```diff
@@ -2748,10 +2809,36 @@ def set_derived_and_validate_values(self) -> "MaxTextConfig":
...
+      cp_axes = infer_cp_axes(self.logical_axis_rules)
+      if cp_axes:
+        self.context_sharding = cp_axes[0]
+
+    # Infer EP rank from logical_axis_rules and disable incompatible flags when EP rank > 1.
+    ep_axes = infer_ep_axes(self.logical_axis_rules)
+    ep_rank = 1
+    for axis_name in ep_axes:
+      ici_val = getattr(self, f"ici_{axis_name}_parallelism", 1)
+      dcn_val = getattr(self, f"dcn_{axis_name}_parallelism", 1)
```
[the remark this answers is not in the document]
NuojCheng: Updated

src/maxtext/configs/types.py
```diff
@@ -2748,10 +2809,39 @@ def set_derived_and_validate_values(self) -> "MaxTextConfig":
       mesh_config = self._load_mesh_config_from_yaml(self.custom_mesh_and_rule.value)
 
       # Use setattr to dynamically apply attributes, keeping code compact
-      for field in ("mesh_axes", "logical_axis_rules", "data_sharding", "context_sharding"):
+      for field in ("mesh_axes", "logical_axis_rules", "data_sharding"):
         if field in mesh_config:
           setattr(self, field, mesh_config[field])
 
+      # Infer context_sharding from logical_axis_rules when using custom mesh rules.
```
Shuwen-Fang, line 2849: are you just trying to see if EP is enabled? if so, might be more robust to use this data structure: https://github.com/AI-Hypercomputer/maxtext/blob/8e4160b2452902818b8bdcc04af55caae157d263/src/maxtext/configs/types.py#L1179 (example use.: https://github.com/AI-Hypercomputer/maxtext/blob/8e4160b2452902818b8bdcc04af55caae157d263/src/maxtext/configs/types.py#L2719)

tests/unit/moe_test.py
```diff
@@ -573,6 +573,7 @@ def test_moe_emb_chunking_random_routing(self):
         weight_dtype="bfloat16",
         megablox=False,
         sparse_matmul=True,
+        ici_expert_parallelism=4,
```
Shuwen-Fang: ty for adding these!

Shuwen-Fang (changes requested):

src/maxtext/configs/types.py
NuojCheng: I intentionally avoid this since we may use other physical axes playing the EP role. This is one of the core ideas of component sharding and I would like to guard it. 

For example, we may use ici_context_parallelism=4 and the CP=4 axis plays expert parallleism in the MoE component. This can be enabled by `custom_mesh_and_rule=cp-as-ep`. In this case, the only way (I think?) to tell if the real EP is enabled is to check the logical rule. If a (or more than one) physical axes is used to shard the expert dimension, we can call the multiplication of them being the EP rank.

Shuwen-Fang: I see, makes sense to me. Maybe we could make this into a helper function so it's more clear what this is trying to do

Shuwen-Fang (approved):

src/maxtext/configs/types.py
NuojCheng: to be clear, do you suggest moving line 2816 - 2843 to helper instead of keeping them here? I am so sure to do that since their major logic already in helper functions, e.g. `infer_cp_axes` and `infer_ep_axes`. Any concerns put them here?

Shuwen-Fang: ya, something like:

ep_enabled = is_ep_enabled() # helper function
if not ep_enabled:
  --> code to remove unused flags

Shuwen-Fang (approved):

src/maxtext/configs/types.py
NuojCheng: Done

khatwanimohit (approved):

PR #4609, opened 2026-07-24
Fix roundtrip checkpoint conversion test

hengtaoguo:
# Description

The qwen3-30b-a3b roundtrip checkpoint conversion test was failing in CI ([logs](https://screenshot.googleplex.com/9vVHYB46LHkhju4)). There were **two underlying causes**:

1. **Unknown flag save_dtype**: This latent argument-isolation bug was recently exposed by the JAX 0.10.2 dependency upgrade, which changed when the Abseil-backed configuration was initialized. `--save_dtype` was parsed by the converter's local `argparse` parser but remained in the global `sys.argv`. When JAX initialized Abseil, Abseil attempted to parse the same argument and raised `UnrecognizedFlagError` because `save_dtype` is not an Abseil flag.  This change replaces `sys.argv` with the remaining MaxText arguments before JAX initialization.

2. **NotImplementedError on cpu**: For `flash` or `autoselected` attention kernels, `AttentionOp` assumed GPU hardware when the platform was not TPU. It did not check if the active platform was actually CPU, leading to it attempting to call the GPU-only `PallasTritonFlashAttention` and throwing a `NotImplementedError` on CPU. We fixed this by adding an explicit flag for `attention=dot_product` to fallback to standard dot-product attention for checkpoint conversion.
    - Recent updates in `attention_op` have actually changed the checkpoint conversion utils, as from now on we have to manually specify `attention=dot_product` to avoid crashes.

# Tests

Covered in CI ([logs](https://screenshot.googleplex.com/3HKEDzvgqTJUrim)), [logs](https://screenshot.googleplex.com/BiERWwVHGvrXSzk).

# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/to_maxtext.py
+++ b/src/maxtext/checkpoint_conversion/to_maxtext.py
@@ -1155,6 +1155,7 @@ if __name__ == "__main__":
   local_args, remaining_args = parser.parse_known_args()
   # Reconstruct model_args (script name + the args MaxText needs)
   model_args = [sys.argv[0]] + remaining_args
+  sys.argv = model_args
 
   # Set jax environment
   jax.config.update("jax_platforms", "cpu")
--- a/tests/integration/checkpoint_conversion_test.py
+++ b/tests/integration/checkpoint_conversion_test.py
@@ -50,6 +50,7 @@ class Qwen3CheckpointConversionTest(unittest.TestCase):
           "override_model_config=True",
           "scan_layers=False",
           "hardware=cpu",
+          "attention=dot_product",
           "skip_jax_distributed_system=True",
           "checkpoint_storage_use_ocdbt=False",
           "checkpoint_storage_use_zarr3=False",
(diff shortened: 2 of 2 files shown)


NuojCheng (approved):

huytransformer (approved):

src/maxtext/layers/attention_op.py
```diff
@@ -1124,8 +1124,8 @@ def apply_attention(
         return out, None, None
 
       else:
-        if model_mode == MODEL_MODE_AUTOREGRESSIVE:
-          # fallback to dot_product as pallas gpu flash attention doesn't support decode stage
+        if model_mode == MODEL_MODE_AUTOREGRESSIVE or target_hardware == "cpu":
```
khatwanimohit: I think changing this in attention_op will lead to increase in CI time as all tests running with cpu backend will end up using dot product attention, the test in CI which keeps tracks of CI time somehow failed for a different reason. 

Can you just change the hardware=cpu only in the checkpoint conversion test which is failing

abhinavclemson (approved):

khatwanimohit (changes requested):

src/maxtext/layers/attention_op.py
hengtaoguo: Great call. The checkpoint conversion tests were already having `hardware=cpu`. To bypass this, I'm now explicitly using `attention=dot_product`. Let me know how it looks!

khatwanimohit (approved):

PR #4613, opened 2026-07-26
Fix data iterator checkpointing and eval under CP load balancing

huytransformer:
# Description

When load balancing is on for context parallelism, the data iterator gets wrapped with map() to reorder batches which breaks the iterator. Eval calls reset() and grain checkpointing reads local_iterator and `map` has neither, so training crashes at the first eval/checkpoint save. This PR keeps the original iterator for checkpointing and only applies the reorder where batches are read. Also fixed standalone_dataloader to use the reordered loader.

# Tests

`python3 -m pytest tests/unit/train_utils_test.py`

22 passed

End-to end: grain + load-balanced CP + checkpointing + eval, 4 steps with checkpoint saves, then a second run restoring from step 3 and resuming through step 5.

# Checklist

### DIFF
--- a/src/maxtext/utils/standalone_dataloader.py
+++ b/src/maxtext/utils/standalone_dataloader.py
@@ -29,7 +29,6 @@ import jax
 
 from maxtext.configs import pyconfig
 from maxtext.trainers.pre_train.train import get_first_step
-from maxtext.common.data_loader import DataLoader
 from maxtext.utils import max_logging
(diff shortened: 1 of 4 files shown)


NuojCheng (approved):

gobbleturk (approved):
LGTM, lets have @aireenmei review as well

aireenmei (approved):
Thanks for the fix!

PR #4615, opened 2026-07-26
Honour partial_rotary_factor on the MRoPE rotary path

lettt325:
### Problem

`Attention.init_rotary_embedding` routes to `Qwen3OmniMoeThinkerTextRotaryEmbedding` whenever `use_mrope` is true. That class rotates the **full** head dimension and, before this change, had no `partial_rotary_factor` parameter at all — so the config value was silently dropped for every model on the MRoPE path.

Two shipped configs are affected, both of which set `use_mrope: true` together with `partial_rotary_factor: 0.25`:

- `src/maxtext/configs/models/qwen3.5-35b-a3b.yml`
- `src/maxtext/configs/models/qwen3.5-397b-a17b.yml`

For these, RoPE was applied to every channel of the head instead of the leading quarter. Nothing raises and every shape stays valid, so the disagreement with the reference implementation is silent.

`PartialRotaryEmbedding` already implements the intended behaviour, but only the non-MRoPE path can reach it.

### Reproduction on `main`

```python
layer = Qwen3OmniMoeThinkerTextRotaryEmbedding(
    min_timescale=1, max_timescale=10000, embedding_dims=16,
    cast_as_fprop_dtype=False, mrope_section=(2, 1, 1), rngs=nnx.Rngs(params=0),
)
out = layer(inputs, positions)          # inputs: [2, 8, 4, 16]
np.allclose(out[..., 4:], inputs[..., 4:])
# -> False   (with partial_rotary_factor 0.25 only the first 4 channels may rotate)
```

Passing the factor explicitly fails outright on `main`:

```
TypeError: __init__() got an unexpected keyword argument 'partial_rotary_factor'
```

### Fix

The layer now takes `partial_rotary_factor` (default `1.0`), splits off the leading `rotary_dim = head_dim * factor` channels, builds `inv_freq` over `rotary_dim` rather than `head_dim`, and concatenates the untouched remainder back. This is the same construction `PartialRotaryEmbedding` uses, so the two paths stay consistent by design.

`Attention.init_rotary_embedding` passes `config.partial_rotary_factor` through — a one-line change.

The default of `1.0` keeps the existing `qwen3-vl-*` and `qwen3-omni-*` behaviour bit-identical; those configs do not set the factor.

### Tests

`MropePartialRotaryEmbeddingTest` in `tests/unit/partial_rotary_embedding_test.py`:

- trailing channels are passed through untouched, leading ones are rotated
- a partial factor no longer produces the fully rotated result (the regression itself)
- omitting the factor reproduces the previous full-rotation output
- for text-only 1D positions, where MRoPE degenerates to ordinary RoPE, the MRoPE path agrees channel for channel with `PartialRotaryEmbedding` (`rtol=1e-5`)

All four pass against this branch; the third and fourth fail on `main`.

### DIFF
--- a/src/maxtext/layers/attentions.py
+++ b/src/maxtext/layers/attentions.py
@@ -819,6 +819,7 @@ class Attention(nnx.Module):
           cast_as_fprop_dtype=True,
           fprop_dtype=self.dtype,
           mrope_section=self.mrope_section,
+          partial_rotary_factor=self.config.partial_rotary_factor,
           rngs=self.rngs,
       )
 
--- a/src/maxtext/layers/embeddings.py
+++ b/src/maxtext/layers/embeddings.py
@@ -1668,6 +1668,7 @@ class Qwen3OmniMoeThinkerTextRotaryEmbedding(RotaryEmbedding):
       fprop_dtype: DType = jnp.bfloat16,
       mrope_section: tuple[int, int, int] | None = None,
       attention_scaling: float = 1.0,
+      partial_rotary_factor: float = 1.0,
       rngs: nnx.Rngs = None,
   ):
     """Initializes the Qwen3OmniMoeThinkerTextRotaryEmbedding module.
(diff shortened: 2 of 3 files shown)


lettt325:
Closing this in favour of #4616, which describes the same defect and the fix without requiring a CLA. Happy for a maintainer to pick it up from there.

PR #4623, opened 2026-07-27
fix: MTP CP-aware roll, segment-aware roll, and synthetic packing support

chiaotung97:
## Summary

Fixes 3 issues when combining Multi-Token Prediction (MTP) with
All-Gather Context Parallelism (AG-CP) and Packing.

## Checklist

- [x] CP-aware left shift (`_shift_left_one_cp_aware`) with ppermute backward ring
- [x] Segment-aware roll (`roll_and_mask_by_segment`) to prevent cross-document leakage
- [x] Synthetic data packed segment IDs (`_make_packed_segment_ids`)
- [x] Unit tests: 30/30 passing on TPU v6e-4 (4 devices)
- [x] Backward compatibility: all paths degrade gracefully
- [x] Merge conflicts with main resolved

## Problem

| # | Issue | Root Cause | Impact |
|---|-------|------------|--------|
| 1 | `jnp.roll` does not cross CP rank boundaries | MTP shifts left each iteration to predict the "next token", but `jnp.roll` only operates within the local shard | Cross-rank right-neighbor tokens are lost under CP |
| 2 | `roll_and_mask` is segment-unaware | Under packing, a left shift can cross document boundaries (e.g., DocA's last token shifts into DocB's first) | Cross-document target leakage contaminates gradient signals |
| 3 | Synthetic data has no real segment boundaries | `segment_ids` is always all-ones even though `base.yml` defaults to `packing: true` | Segment boundary logic is untestable on synthetic data; `train_utils.py` also blocks the combination outright |

### 1. CP-Aware Left Shift (`_shift_left_one_cp_aware`)

Uses `jax.lax.ppermute` in a backward ring: rank r sends its first token to
rank r−1, receives rank r+1's first token, and places it in its last
position. Degrades to `jnp.roll` when no CP axis is in scope — zero overhead.

### 2. Segment-Aware Roll (`roll_and_mask_by_segment`)

Shifts both `x` and `segment_ids` via `_shift_left_one_cp_aware`, then masks
positions where `seg_current != seg_next` (document boundary) or
`seg_current == 0` (padding). Falls back to `roll_and_mask` when
`segment_ids=None`. All rolling variables in
`MultiTokenPredictionBlock.__call__` now use this function.

### 3. Synthetic Data with Packed Segment IDs

`_make_packed_segment_ids` splits each row into 2..N randomly-sized segments
with sequential integer IDs starting from 1. Removed the `train_utils.py`
guard that rejected `synthetic + packing + CP`.

## Files Changed

| File | Change | +/− |
|------|--------|:---:|
| `layers/multi_token_prediction.py` | `_shift_left_one_cp_aware`, `roll_and_mask_by_segment`, wiring | +103/−4 |
| `input_pipeline/synthetic_data_processing.py` | `_make_packed_segment_ids` | +43/−2 |
| `utils/train_utils.py` | Remove synthetic+packing+CP guard | −5 |
| `tests/unit/multi_token_prediction_test.py` | 17 new tests (segment + CP + packed_ids) | +296 |
| `docs/design/ag_cp_mtp_packing_fix.md` | Design document | +141 |

## Unit Tests

**30 tests, all passing** (TPU v6e-4, 4 devices):

| Test class | Tests | Description |
|------------|:-----:|-------------|
| `TestRollAndMask` | 3 | Existing — local roll and mask |
| `TestRollAndMaskBySegment` | 8 | Segment boundary masking (no CP) |
| `TestMakePackedSegmentIds` | 6 | Segment ID generation: shape, contiguity, determinism |
| `TestShiftLeftOneCpAware` | 5 | CP-aware shift: 1D/2D/3D, last-rank zero, rank-boundary non-zero |
| `TestRollAndMaskBySegmentWithCp` | 4 | CP + segment boundaries, padding mask, single-seg, 3D |
| Upstream tests (layer, block, MTP loss, quantize) | 4 | Existing — unchanged |

## Backward Compatibility

- `_shift_left_one_cp_aware`: degrades to `jnp.roll` when CP=1 or no `"context"` axis
- `roll_and_mask_by_segment`: degrades to `roll_and_mask` when `segment_ids=None`
- `roll_and_mask(shift=-1)`: equivalent to original path when CP is off
- `synthetic_data_processing`: segment IDs remain `jnp.ones` when `packing=False`

## Verification

| # | Configuration | TPU v6e-4 |
|---|---------------|:---------:|
| 1 | MTP + packing | ✅ |
| 2 | MTP + CP=2 | ✅ |
| 3 | MTP + CP=2 + packing | ✅ |
| 4 | Unit tests | ✅ 30/30 |


RissyRan:
Hi reviewers, just an FYI that this PR is from a forked repo. Due to secret key limitations, the Gemini review couldn't be triggered.

src/maxtext/layers/multi_token_prediction.py
```diff
@@ -45,9 +45,53 @@ class mtp_acceptance(nnx.Intermediate):  # pylint: disable=invalid-name
...
+    axis_name: Mesh axis along which the sequence is sharded.
+
+  Returns:
+    Array shaped like x, left-shifted by 1 across CP boundaries.
+  """
+  local_rolled = jnp.roll(x, -1, axis=1)
+  local_rolled = local_rolled.at[:, -1:, ...].set(0)
+
+  try:
+    cp_size = jax.lax.psum(1, axis_name=axis_name)
```
NuojCheng: This is a very interesting way to collect cp_size. I suggest using

`jax.lax.axis_size(axis_name)`
OR
`mesh.shape.get(axis_name, 1)`

NuojCheng (approved):
LGTM and thank you for the contribution! Some minor comments:
* the cpu test failure can be solved by rebasing to main
* I don't think we need design doc included in this PR, but it is great for PR review. Can we move the design doc to the PR description?

src/maxtext/layers/multi_token_prediction.py
```diff
@@ -45,9 +45,53 @@ class mtp_acceptance(nnx.Intermediate):  # pylint: disable=invalid-name
...
+  shard_map / no CP) or when CP size is 1.
+
+  Args:
+    x: Input array with sequence dim at axis=1.
+    axis_name: Mesh axis along which the sequence is sharded.
+
+  Returns:
+    Array shaped like x, left-shifted by 1 across CP boundaries.
+  """
+  local_rolled = jnp.roll(x, -1, axis=1)
```
huytransformer, line 68: Would this still work with load balancing on? If not perhaps we should handle it differently or reject LB?

src/maxtext/layers/multi_token_prediction.py
huytransformer, line 480: Is rolled_target_mask a boolean mask?

src/maxtext/layers/multi_token_prediction.py
```diff
@@ -272,17 +360,26 @@ def __call__(
...
     mtp_preds_list = []
     mtp_masks_list = []
 
     for k in range(1, cfg.mtp_num_layers + 1):
-      rolled_input_ids = roll_and_mask(rolled_input_ids)
-      rolled_target_ids = roll_and_mask(rolled_target_ids)
-      rolled_target_mask = roll_and_mask(rolled_target_mask)
-      rolled_position_id = roll_and_mask(rolled_position_id)
+      rolled_input_ids = roll_and_mask_by_segment(rolled_input_ids, rolled_segment_ids)
+      rolled_target_ids = roll_and_mask_by_segment(rolled_target_ids, rolled_segment_ids)
```
huytransformer, line 432: Is rolled_segment_ids aligned with rolled_target_ids here? target_ids is already shifted by 1 so after this roll the actual target is 2 tokens ahead? segment IDs passed to the helper seem to track only 1 token

huytransformer (changes requested):

src/maxtext/layers/multi_token_prediction.py
chiaotung97: target_mask enters `MultiTokenPredictionBlock.__call__` as `jnp.bool_` (standard training pipeline behavior: `True` = valid position, `False` =  padded/ignore).

We will cast it to `jnp.int32` before the rolling loop because `roll_and_mask_by_segment` internally zeros boundary positions via `jnp.where`(mask, 0, rolled) — when rolled is `bool`and `0` is `int32`, JAX auto-promotes to `int32` with a scatter-input FutureWarning (will become a hard error in a future JAX release).

The effective value is 0/1 either way, and downstream mtp_xent * mask works identically for `int32` and bool. Verified on TPU v6e. A follow-up commit will add an explicit `bool` → `int32` cast at the entry of `__call__`.

chiaotung97:
> LGTM and thank you for the contribution! Some minor comments:
> 
> * the cpu test failure can be solved by rebasing to main
> * I don't think we need design doc included in this PR, but it is great for PR review. Can we move the design doc to the PR description?

OK, I will remove the design doc later, thanks for these comments.

src/maxtext/layers/multi_token_prediction.py
huytransformer: Thank you for your response.

`target_mask enters MultiTokenPredictionBlock.__call__ as jnp.bool_ (standard training pipeline behavior: True = valid position, False = padded/ignore).`

I think this is not true if you're doing packing right? e.g. targets_segmentation is an integer array

chiaotung97: Thank you for your response.

This is a real problem. `dataloader` in our private forked repo override some vars like `targets_segmentation`, and finally pass to `rolled_target_mask`, ensuring it as a boolean array. But this will cause bugs in this repo. I am working for fix this problem.

docs/design/ag_cp_mtp_packing_fix.md
```diff
@@ -0,0 +1,141 @@
+# Design Doc: MTP + CP + Packing Combined Fix
```
RissyRan: Thanks for the great summary! 

Instead of putting into `design`, I am thinking of putting into `docs/reference` for better alignment.

Could we rename the title to something like `MTP + CP + Packing`?

For main content, could we just include the first 3 highlights from design section? But feel free to add more details in this PR description. HDYT?

src/maxtext/input_pipeline/synthetic_data_processing.py
```diff
@@ -28,6 +28,42 @@
...
+
+  Returns:
+    Integer array [batch_size, seq_len] where 0 = padding.
+  """
+  if max_segments_per_seq < 2:
+    return jnp.ones((batch_size, seq_len), dtype=jnp.int32)
+
+  key = jax.random.PRNGKey(42)
+  segments = jnp.zeros((batch_size, seq_len), dtype=jnp.int32)
+  for b in range(batch_size):
```
RissyRan: Have you considered vectorized implementation for efficiency vs. this loop?

Gemini suggested following, but just for your reference (I didn't test the output alignment):

```
import jax
import jax.numpy as jnp

def _make_packed_segment_ids(
    key: jax.Array,
    batch_size: int,
    seq_len: int,
    max_segments_per_seq: int,
) -> jnp.ndarray:
  """Creates synthetic segment IDs simulating packed sequences in vectorized JAX.

  Args:
    key: JAX PRNG key.
    batch_size: Number of sequences.
    seq_len: Total sequence length.
    max_segments_per_seq: Maximum number of segments per sequence (>= 2).

  Returns:
    Integer array [batch_size, seq_len] with sequential segment IDs starting at 1.
  """
  if max_segments_per_seq < 2 or seq_len < 2:
    return jnp.ones((batch_size, seq_len), dtype=jnp.int32)

  max_segs = min(max_segments_per_seq, seq_len)
  k1, k2 = jax.random.split(key, 2)

  # 1. Sample random number of segments per row in [2, max_segs] -> shape (batch_size, 1)
  n_segs = jax.random.randint(k1, shape=(batch_size, 1), minval=2, maxval=max_segs + 1)
  num_splits = n_segs - 1  # Number of split boundaries needed in [1, seq_len - 1]

  # 2. Assign random uniform scores to positions 1..seq_len-1 and rank them
  scores = jax.random.uniform(k2, shape=(batch_size, seq_len - 1))
  # Double argsort computes the 0-based rank of each position per row
  ranks = jnp.argsort(jnp.argsort(scores, axis=-1), axis=-1)

  # 3. Position j is a split boundary if its rank < num_splits for that row
  split_mask = ranks < num_splits

  # 4. Position 0 is always the start of segment 1 (True); prepend to split_mask
  first_pos = jnp.ones((batch_size, 1), dtype=jnp.bool_)
  is_boundary = jnp.concatenate([first_pos, split_mask], axis=-1)

  # 5. Cumulative sum across sequence length gives sequential segment IDs: 1, 2, ...
  return jnp.cumsum(is_boundary, axis=-1, dtype=jnp.int32)
```

src/maxtext/layers/multi_token_prediction.py
```diff
@@ -57,9 +101,53 @@ def roll_and_mask(x: jnp.ndarray, shift: int = -1) -> jnp.ndarray:
...
+  seg_next = _shift_left_one_cp_aware(segment_ids)
+
+  # A position is a boundary if:
+  #   1. current segment != next segment (document boundary), OR
+  #   2. current segment == 0 (padding/EOD position)
+  is_boundary = (seg_current != seg_next) | (seg_current == 0)
+
+  mask = is_boundary
+  for _ in range(x.ndim - 2):
+    mask = jnp.expand_dims(mask, axis=-1)
```
RissyRan: Maybe something like bellow?

```
  mask = is_boundary[(...,) + (None,) * (x.ndim - 2)]
```

src/maxtext/layers/multi_token_prediction.py
```diff
@@ -45,9 +45,53 @@ class mtp_acceptance(nnx.Intermediate):  # pylint: disable=invalid-name
...
+  if cp_size == 1:
+    return local_rolled
+
+  cp_rank = jax.lax.axis_index(axis_name)
+  first_token = jax.lax.dynamic_slice_in_dim(x, 0, 1, axis=1)
+  # Backward ring: rank r sends first_token to rank r-1.
+  perm = [(r, (r - 1) % cp_size) for r in range(cp_size)]
+  next_first = jax.lax.ppermute(first_token, axis_name=axis_name, perm=perm)
+  next_first = jnp.where(cp_rank == cp_size - 1, jnp.zeros_like(next_first), next_first)
+  return local_rolled.at[:, -1:, ...].set(next_first)
```
RissyRan: We usually met performance bottleneck when using `.at[...].set()`. May pair with agent to get a solution with jnp.where with proper mask?

RissyRan (commented):
Thanks for putting this together! The overall approach looks great. I've left a few comments on performance, happy to chat more!

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Thank you for the review. Changed to jax.lax.axis_size(axis_name) in 1b33f67c. axis_size is more direct and avoids the unnecessary reduction.

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Thank you for the observation. mtp_num_layers > 1 with CP + packing has not been fully tested in our scenario. Added a guard in 092c3e87 that limits this combination to mtp_num_layers=1 for now. Multi-layer support will be added in a follow-up.

docs/design/ag_cp_mtp_packing_fix.md
chiaotung97: Thank you for the suggestion. Moved the design doc to docs/reference/mtp_cp_packing.md with the title "MTP + CP + Packing" in 8d74ad2b. Kept the key design highlights covering all 5 fixes (CP-aware shift, segment-aware roll, target_mask normalization, synthetic packing, and CP guards).

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Thank you for catching this. Under context_parallel_load_balance=True, the DUAL_CHUNK_SWAP reorder shuffles token positions, which breaks the ppermute-based neighbor fetch in _shift_left_one_cp_aware (the backward ring assumes sequential token ordering). Added an explicit guard in 092c3e87 that raises ValueError when context_parallel_load_balance=True is combined with CP and MTP.

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Fixed in 092c3e87 by adding target_mask = (target_mask != 0).astype(jnp.int32) at the top of __call__ to normalize to a proper binary mask before the rolling loop. Also added a guard for None in the mtp_num_layers=0 quantization trace path.

src/maxtext/input_pipeline/synthetic_data_processing.py
chiaotung97: Thank you for the suggestion. Applied the vectorized implementation in 3025f72b. Tested on TPU v6e: (137ms → 0.8ms at batch=32, seq=8192) and now supports JIT compilation. All existing unit tests pass unchanged.

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Thank you, applied in 3025f72b. Replaced the expand_dims loop with is_boundary[(...,) + (None,) * (x.ndim - 2)].

src/maxtext/layers/multi_token_prediction.py
chiaotung97: Thank you for the suggestion. Replaced both .at[:, -1:, ...].set() calls in _shift_left_one_cp_aware with jnp.where using a broadcast mask in 3025f72b.

docs/reference/mtp_cp_packing.md
```diff
@@ -0,0 +1,70 @@
+# MTP + CP + Packing
+
+## Summary
```
RissyRan: Recommend to remove this Summary section. I think `Highlights` should be good for reference purpose.

RissyRan (approved):

src/maxtext/layers/multi_token_prediction.py
huytransformer: Thank you for addressing this. Perhaps just quick follow up that I think the guard might not trigger. Is context_parallel_size actually present on cfg?

src/maxtext/layers/multi_token_prediction.py
huytransformer: Thank you. I think this can still happen with mtp_num_layers=1 though...

docs/reference/mtp_cp_packing.md
qiaotonggg: Thanks, already solved.

huytransformer (approved):

PR #4634, opened 2026-07-27
Fix DeepSeek decode running the prefill attention path under NNX

ecnal-cienet:
## Description

Decoding `deepseek2-16b` through MaxEngine on the NNX path emits a single repeated token. On the nightly checkpoint (v6e-8, `maxtext.inference.decode`, greedy, `mla_naive_kvcache=false`):

```
prompt: An attention function can be described as mapping a query and a set of key-value pairs to an output,
        where the query, keys, values, and outputs are all vectors. The output is

before:  a a a a a a a a a a a a a a a a a a a a a ...   (all 1024 tokens)
after:   a weighted sum of the values, where the weights are computed by a function of the query and the keys.
```

## Root cause

`DeepSeekGenericLayer.attention_op` passed `model_mode=self.model_mode` to attention — the value captured at `__init__` — instead of the `model_mode` its `__call__` receives. Same in `self_attention_with_norm_op` for the MHC branch.

MaxEngine's NNX path builds the model **once in PREFILL mode** and passes the mode per call: `_nnx_run_model` always merged `self.graphdef`, which is the prefill graphdef. So on every autoregressive step each DeepSeek layer still read `self.model_mode == "prefill"` and took the prefill KV-cache branch. Prefill logits stay correct, which is why the first token is right and everything after it is wrong.

Only DeepSeek is affected — llama2 and the other families already forward the call-time `model_mode`. A sweep of every non-`__init__` read of `self.model_mode` in `src/maxtext` found DeepSeek to be the only one feeding a stale mode into attention.

The bug was latent under Linen, which re-binds layers on each `apply` so `self.model_mode` is always current. It only surfaces once a model object is built once and reused across modes, which is exactly what pure NNX does.

## Fix

| File | Change |
|---|---|
| `models/deepseek.py` | `attention_op` and `self_attention_with_norm_op` take `model_mode` and forward the call-time value. |
| `models/deepseek4.py` | Updated call site (`DeepSeek4DecoderLayer` inherits both helpers). |
| `inference/maxengine/maxengine.py` | Cache a `graphdef_ar` and merge with the graphdef built for the requested mode. |

Either change alone restores correct output; both are included deliberately, so a future layer repeating the mistake cannot silently corrupt decode and the two protections cannot regress together unnoticed.

## Tests

Two new CPU-only files, neither needing a checkpoint. They close a real gap: golden-logit tests only exercise the forward pass, so nothing covered the prefill → autoregressive handoff.

**`tests/unit/decoder_layer_model_mode_test.py`** — 16 cases. Asserts at the attention boundary that a layer attends in the mode it was *called* with. Covers deepseek dense/moe, llama2, mistral, gemma, gemma2, qwen2 and qwen3, in **both** directions (built-prefill/called-AR and built-AR/called-prefill) so no layer can pass by hardcoding a mode.

**`tests/unit/deepseek_decode_consistency_test.py`** — 4 cases. Greedy decode through prefill/insert/generate must equal the argmax rollout of a plain forward pass over the same prefix. Parameterized over deepseek-moe, deepseek-dense and a generic model, so adding a family is one line. Includes a separate guard that `generate` merges an autoregressive graphdef, covering the engine side on its own.

## Verification

All on `wanglance-v6e-8` (TPU v6 lite, 8 chips). Rows marked *unmodified `main`* were run in a throwaway `git worktree`, so the branch was never reverted.

| Item | Result | Log |
| --- | --- | --- |
| Real-checkpoint decode, **with fix** | `-> a weighted sum of the values, where the weights are computed by a function of the query and the keys.` | [log](https://paste.googleplex.com/5269766309806080) |
| Real-checkpoint decode, **unmodified `main`** | `-> a a a a a a a a a ...` for all 1024 tokens — the bug reproduced | [log](https://paste.googleplex.com/5074171318108160) |

# Checklist

### DIFF
--- a/src/maxtext/inference/maxengine/maxengine.py
+++ b/src/maxtext/inference/maxengine/maxengine.py
@@ -145,11 +145,16 @@ class MaxEngine(_BaseEngine):  # pyrefly: ignore[invalid-inheritance]
       # nnx.merge. `rest` (RNG state etc.) is materialized in load_params.
       graphdef, _, _, _ = nnx.split(abstract_model, nnx.Param, nnx.Cache, ...)
       self.graphdef = graphdef
+      # Layers may bake their construction-time model_mode into static attributes,
+      # so a call must be merged with the graphdef built for that same mode.
+      graphdef_ar, _, _, _ = nnx.split(abstract_model_ar, nnx.Param, nnx.Cache, ...)
+      self.graphdef_ar = graphdef_ar
       self._create_model_fn = _create_model
       self._nnx_rest_state = None
     else:
       self.model = models.transformer_as_linen(config, mesh=self._mesh, quant=quant, model_mode=MODEL_MODE_PREFILL)
       self.graphdef = None
+      self.graphdef_ar = None
       self._create_model_fn = None
     self.replicated_sharding = jax.sharding.NamedSharding(self._mesh, P(None))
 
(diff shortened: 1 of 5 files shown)


xibinliu (approved):
Good catch of the model_mode difference for deepseek attention_op

shuningjin (approved):
Thank you for the fix!

ecnal-cienet:
Confirmed merged into main branch. Manually close this PR.

PR #4642, opened 2026-07-28
Run selected tests in Decoupled mode on CPU in CI

shralex:
This PR adds a targeted decoupled test suite executed with `DECOUPLE_GCLOUD=TRUE` on CPU in CI to validate fragile offline components (checkpointing, quantization, GRPO NNX loss dispatch).

Changes:
1. **`pytest.ini`**: Registered `decoupled_target` marker ("tests to be additionally validated in decoupled mode (DECOUPLE_GCLOUD=TRUE) on CPU in CI.").
2. **Test Modules**: Tagged 9 target test files across `tests/unit/` and `tests/integration/` with `pytestmark = [pytest.mark.decoupled_target]`.
3. **CI Pipeline ([.github/workflows/run_tests_against_package.yml](file:///.github/workflows/run_tests_against_package.yml))**: Added a step under `inputs.flavor == 'cpu-unit'` executing `$PYTHON_EXE -m pytest -v -n auto -m "decoupled_target"`.

# Checklist

### DIFF
--- a/.github/workflows/run_tests_against_package.yml
+++ b/.github/workflows/run_tests_against_package.yml
@@ -223,8 +223,9 @@ jobs:
       - name: Run Targeted Decoupled Tests
         if: inputs.flavor == 'cpu-unit'
         run: |
-          echo "Running targeted fragile tests in decoupled mode (DECOUPLE_GCLOUD=TRUE)..."
+          echo "Running targeted tests in decoupled mode (DECOUPLE_GCLOUD=TRUE)..."
           export DECOUPLE_GCLOUD=TRUE
+
           $PYTHON_EXE -m pytest -v -n auto -m "decoupled_target" \
             --junitxml=test-results-decoupled-targeted.xml
 
--- a/.github/workflows/run_tests_against_package.yml
+++ b/.github/workflows/run_tests_against_package.yml
@@ -220,6 +220,15 @@ jobs:
               ${INPUTS_PYTEST_EXTRA_ARGS}
           fi
 
+      - name: Run Targeted Decoupled Tests
+        if: inputs.flavor == 'cpu-unit'
+        run: |
+          echo "Running targeted fragile tests in decoupled mode (DECOUPLE_GCLOUD=TRUE)..."
+          export DECOUPLE_GCLOUD=TRUE
+          $PYTHON_EXE -m pytest -v -n auto -m "decoupled_target" \
+            --junitxml=test-results-decoupled-targeted.xml
+
+
         env:
           INPUTS_IS_SCHEDULED_RUN: ${{ inputs.is_scheduled_run }}
           INPUTS_PYTEST_MARKER: ${{ inputs.pytest_marker }}
--- a/pytest.ini
+++ b/pytest.ini
@@ -36,7 +36,8 @@ markers =
     cpu_only: marks tests to be run on CPUs only
     decoupled: tests that validate offline / DECOUPLE_GCLOUD=TRUE mode.
 	       NOTE: this marker is not to be used manually, it is auto-
-	       applied to tests with external_* or tpu_only marker. 
+	       applied to tests without external_* or tpu_only marker. 
+
     scheduled_only: marks tests to run only on scheduled CI runs
     newly_added: newly introduced or modified tests in PRs, executed even if scheduled_only
     integration_test: tests exercising larger portions of the system,
--- a/pytest.ini
+++ b/pytest.ini
@@ -45,4 +45,7 @@ markers =
     external_serving: JetStream / serving / decode server components
     external_training: goodput integrations
     post_training: marks tests for post-training code paths.
+    decoupled_target: tests to be additionally validated in decoupled mode
+                      (DECOUPLE_GCLOUD=TRUE) on CPU in CI.
+
 
--- a/pytest.ini
+++ b/pytest.ini
@@ -46,6 +46,9 @@ markers =
     external_training: goodput integrations
     post_training: marks tests for post-training code paths.
     decoupled_target: tests to be additionally validated in decoupled mode
-                      (DECOUPLE_GCLOUD=TRUE) on CPU in CI.
+                      (DECOUPLE_GCLOUD=TRUE) on CPU in CI. NOTE: conftest.py
+                      automatically applies the `decoupled` marker to these tests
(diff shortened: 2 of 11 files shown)


pytest.ini
SurbhiJainUSC, line 37: How about we reuse this marker instead of creating a new one `decoupled_target`?

shralex: The goal is different -- decoupled says which tests are able to execute without GCP dependencies. I believe its also applied automatically for tests not explicitly excluded. Decoupled_target says which tests should we run in decoupled mode as part of CI testing (currently none of them run). If you see a way to clarify this please let me know

shralex: I updated the comment for decoupled_target

SurbhiJainUSC (approved):

.github/workflows/run_tests_against_package.yml
```diff
@@ -220,6 +220,16 @@ jobs:
...
+        if: inputs.flavor == 'cpu-unit'
+        run: |
+          echo "Running targeted tests in decoupled mode (DECOUPLE_GCLOUD=TRUE)..."
+          export DECOUPLE_GCLOUD=TRUE
+
+          $PYTHON_EXE -m pytest -v -n auto -m "decoupled_target" \
+            --junitxml=test-results-decoupled-targeted.xml
+
+
         env:
```
gulsumgudukbay, line 233: This step landed between the previous step's run: and its env:, so that env: now belongs to this step. Run Tests loses all ten INPUTS_* vars for every flavor, it would run -m "", unsharded, with no --ignore=tests/post_training. I think this step should be moved below the env: block.

.github/workflows/run_tests_against_package.yml
```diff
@@ -220,6 +220,16 @@ jobs:
               ${INPUTS_PYTEST_EXTRA_ARGS}
           fi
 
+      - name: Run Targeted Decoupled Tests
+        if: inputs.flavor == 'cpu-unit'
```
gulsumgudukbay, line 224: cpu-unit is a 4-way matrix (CPU_UNIT_TOTAL_WORKERS=4), so this runs in all four shards with no --splits/--group, the same 48 tests four times per PR.

.github/workflows/run_tests_against_package.yml
```diff
@@ -220,6 +220,16 @@ jobs:
...
+        run: |
+          echo "Running targeted tests in decoupled mode (DECOUPLE_GCLOUD=TRUE)..."
+          export DECOUPLE_GCLOUD=TRUE
```
gulsumgudukbay, line 227: This doesn't reproduce decoupled mode here: cpu-unit installs the tpu extra, which pins google-cloud-storage>=3.13.0, and DECOUPLE_GCLOUD only changes what gcloud_stub returns, no meta_path hook exists. 

a module-scope from google.cloud import storage still resolves, so #4003, #4180 and #4637 would all have passed this gate, as would today's main. 

I strongly believe here decoupled requirements should be installed.

.github/workflows/run_tests_against_package.yml
```diff
@@ -220,6 +220,16 @@ jobs:
...
+
+          $PYTHON_EXE -m pytest -v -n auto -m "decoupled_target" \
```
gulsumgudukbay, line 229: Is $PYTHON_EXE empty here? since it's a shell variable from the previous step's run: body and it doesn't cross steps. So this becomes -m pytest ... 
and can cause the bash: -m: command not found error. we may also see this behavior for source .venv/bin/activate, PYTHONPATH, and the cd /deps branch.

.github/workflows/run_tests_against_package.yml
```diff
@@ -220,6 +220,16 @@ jobs:
...
+            --junitxml=test-results-decoupled-targeted.xml
```
gulsumgudukbay, line 230: I see arg issues here: no paths, so addopts --ignores apply and 4 of the 9 tagged files never collect (48 tests from 5 files); no --ignore=tests/post_training, and -m filters after collection, so there are 14 errors during collection; -n auto only works since when there are more than one worker xdist is installed

pytest.ini
```diff
@@ -36,7 +36,8 @@ markers =
     cpu_only: marks tests to be run on CPUs only
     decoupled: tests that validate offline / DECOUPLE_GCLOUD=TRUE mode.
 	       NOTE: this marker is not to be used manually, it is auto-
-	       applied to tests with external_* or tpu_only marker. 
+	       applied to tests without external_* or tpu_only marker.
```
gulsumgudukbay, line 39: tpu_only tests do get the decoupled marker

```suggestion
	       all tests except external_serving / external_training. 
```

pytest.ini
```diff
@@ -45,4 +46,12 @@ markers =
     external_serving: JetStream / serving / decode server components
     external_training: goodput integrations
     post_training: marks tests for post-training code paths.
+    decoupled_target: tests to be additionally validated in decoupled mode
+                      (DECOUPLE_GCLOUD=TRUE) on CPU in CI. NOTE: when running
+                      with DECOUPLE_GCLOUD=TRUE, conftest.py automatically
+                      selects at runtime only those tests that can run in
```
gulsumgudukbay, line 52: conftest.py only deselects external_serving / external_training; it doesn't judge whether a test can run decoupled. this PR selects `-m "decoupled_target"`, not `decoupled and decoupled_target`, so that marker isn't involved at all.

tests/unit/checkpointing_test.py
```diff
@@ -35,8 +35,11 @@
 from maxtext.common import grain_utility
 import numpy as np
 import optax
+import pytest
 import safetensors.numpy
 
+pytestmark = [pytest.mark.decoupled_target]
```
gulsumgudukbay, line 41: #4637 is a prerequisite for this to run without failures.

tests/integration/checkpoint_conversion_test.py
```diff
@@ -23,7 +23,7 @@
 
 from maxtext.checkpoint_conversion import to_maxtext
 
-pytestmark = [pytest.mark.integration_test]
+pytestmark = [pytest.mark.integration_test, pytest.mark.decoupled_target]
```
gulsumgudukbay, line 26: cpu-unit's marker deliberately excludes integration tests so we don't need this.

gulsumgudukbay (changes requested):

PR #4646, opened 2026-07-28
fp8 KV cache: pre-dispatch quantize_kv in forward_serve_vllm

lokic233:
## Problem

`Attention.forward_serve_vllm` passes bf16 `k`/`v` straight into the ragged-paged-attention (RPA) v3 kernel with `q_scale = k_scale = v_scale = None`. When the vLLM rollout allocates the KV cache in fp8 (`kv_cache_dtype=fp8`), the kernel raises:

```
ValueError: Expected kv_cache.dtype=float8_e4m3fn to be equal to k.dtype=bfloat16 and v.dtype=bfloat16.
```

The RPA v3 kernel's `static_validate_inputs` requires `kv_cache.dtype == k.dtype == v.dtype`, and its own comment states it *"expects the kv quantization happens outside of the RPA kernel."* The MaxText serving path never quantizes, so it trips the assert during `capture_model`/KV-init and again on any real decode step.

## Why the native path doesn't hit this

The native vLLM backend (`layers/vllm/backends/flash_attn.py`) already quantizes before dispatch — it calls `quantize_kv(...)` on `k`/`v` ahead of the attention call. `forward_serve_vllm` is simply missing the equivalent step.

## Fix

Mirror `flash_attn.py` on the MaxText serving path: when the KV cache is a 1-byte float (fp8 e4m3/e5m2), quantize `k`/`v` to the cache dtype via the shared `quantize_kv` helper and thread the scales through to the kernel. Non-fp8 caches are a no-op (scales stay `None`), so bf16 behavior is unchanged. Single code path, reusing the existing helper — no duplicated quantization logic and no in-kernel cast.

## Validation

Validated on TPU v7x (gemma-4-2B DAPO/GRPO RL, 64 chips, 4x4x4) with `kv_cache_dtype=fp8`:

- **Clears `capture_model`/KV-init** — no dtype assert; real decode runs on the same RPA kernel that previously asserted (peak ~500 tok/s).
- **Reward unchanged** — `rewards/mean = 0.1405`, matching the bf16-cache baseline (~0.13); advantages centered at ~0, no NaN/collapse. Properly-scaled fp8 preserves the reward signal.
- **Memory** — fp8 halves per-token KV bytes (2× token capacity) at the same HBM footprint.

## Usage

The fix requires no new flags — it activates automatically whenever the KV cache dtype is a 1-byte float on the MaxText serving path. To enable fp8 KV, request an fp8 cache the standard vLLM-TPU way:

- **`--kv-cache-dtype=fp8`** (or `fp8_e4m3`) → cache allocated in `float8_e4m3fn`
- **`--kv-cache-dtype=fp8_e5m2`** → e5m2 variant

On the MaxText path (`hf_overrides={'architectures': ['MaxTextForCausalLM']}`, i.e. `model_impl=maxtext`), `forward_serve_vllm` now quantizes `k`/`v` to the cache dtype before the RPA dispatch, so the run clears `capture_model` and decodes instead of hitting the `kv_cache.dtype == k.dtype` assert. Non-fp8 caches (bf16/fp16) are a no-op — scales stay `None` and behavior is unchanged.

**Scope:** this only touches the MaxText serving path. The native vLLM-TPU backend (`flash_attn.py`) already supported `--kv-cache-dtype=fp8`; #4646 brings the MaxText path to parity so both backends behave identically under the same flag.

### DIFF
--- a/src/maxtext/layers/attentions.py
+++ b/src/maxtext/layers/attentions.py
@@ -1052,6 +1052,7 @@ class Attention(nnx.Module):
       # pylint: disable=import-outside-toplevel
       # pytype: disable=import-error
       from tpu_inference.layers.common.attention_interface import sharded_ragged_paged_attention as rpa_ops
+      from tpu_inference.layers.common.quantization import quantize_kv
     except ImportError as e:
       raise ImportError(
           "vLLM RPA attention ops require the vllm-tpu package. Please install it with `pip install vllm-tpu`."
@@ -1072,7 +1073,15 @@ class Attention(nnx.Module):
     else:
       attention_chunk_size = None
 
+    # fp8 KV cache: the RPA v3 kernel requires kv_cache.dtype == k.dtype == v.dtype
+    # and expects KV quantization to happen outside the kernel. Mirror the native
+    # vLLM backend (flash_attn.py): when the cache is a 1-byte float (fp8), quantize
+    # k/v to the cache dtype here via quantize_kv and thread the scales through.
+    # Non-fp8 caches are a no-op (scales stay None).
     q_scale, k_scale, v_scale = None, None, None
+    if jnp.issubdtype(rpa_kv_cache.dtype, jnp.floating) and rpa_kv_cache.dtype.itemsize == 1:
+      k_scale = v_scale = 1.0
+      key, value = quantize_kv(rpa_kv_cache.dtype, key, value, k_scale, v_scale)
 
     md = rpa_metadata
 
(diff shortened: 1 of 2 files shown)


lokic233:
Added a minimal regression test (`test_forward_serve_vllm_fp8_kv_quantizes`) that mirrors the existing `test_forward_serve_vllm` — with an fp8 KV cache it asserts the k/v passed to the RPA kernel are cast to the fp8 cache dtype. Skipped like its sibling (needs `vllm-tpu`, not yet a MaxText dep); the behavior itself was validated on TPU (reward 0.1405, clears capture_model).

PR #4664, opened 2026-07-29
Mhc optimization: Rewrites einsums for better xla fusion

h1t35h:
# Description

XLA optimizations does not seem to gel well with einsum for cases where dimensions are being expanded. Leading to a ~10- 15% performance drop in the current implementation. This change implements expanded einsums with jnp.expand_dims leading to better XLA fusion.

Before (110us): 
image
After (97us):
image

# Tests

All existing tests passing.

# Checklist

### DIFF
--- a/src/maxtext/layers/mhc.py
+++ b/src/maxtext/layers/mhc.py
@@ -67,13 +67,10 @@ def sinkhorn(t, iters=20):
   t = jax.nn.softmax(t, axis=-1) + eps
   t = t / (jnp.sum(t, axis=-2, keepdims=True) + eps)
 
-  def body_fun(i, val):
-    val = val / (jnp.sum(val, axis=-1, keepdims=True) + eps)
-    val = val / (jnp.sum(val, axis=-2, keepdims=True) + eps)
-    return val
+  for _ in range(iters - 1):
+    t = t / (jnp.sum(t, axis=-1, keepdims=True) + eps)
+    t = t / (jnp.sum(t, axis=-2, keepdims=True) + eps)
 
-  # Use lax.fori_loop for an efficient, JIT-friendly loop
-  t = jax.lax.fori_loop(0, iters - 1, body_fun, t)
   return t.astype(initial_dtype)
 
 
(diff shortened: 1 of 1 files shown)


src/maxtext/layers/mhc.py
```diff
@@ -269,7 +267,8 @@ def __call__(
         1.0,
         eps=1e-6,
     )
-    layer_input = jnp.einsum("bskd,bsk -> bsd", x, pre_mapping, precision=self.matmul_precision)
+    # bskd, bsk -> bsd
+    layer_input = jnp.sum(x * jnp.expand_dims(pre_mapping, axis=3), axis=2)
```
NuojCheng, line 278: for my understanding, do you see perf gain by this change? I thought `einsum` would be more performant.

h1t35h: Yes, even I found it a bit counter-intuitive einsum being more performant is what my understanding was as well. I think while being generally true for some cases like this one it seems better optimization is coming by moving away from it. Specifically, I can see the all-reduce overlap. I tested the changes individually by just changing the `einsums` and `for_iloop` but that does not seem to work I'm only seeing the All-Reduce overlap when both the changes are present. 

Just `for_iloop` replaced : 
image
Just `for_iloop` and `einsum` replaced: 
image

I don't have a very solid reasoning behind why of it.

dandragona: This is also somewhat surprising to me. I am curious if you can share the xprof? Also is this an isolation test or was this taken from your training job?

I am curious how this change generalizes to other shapes, i.e., is the speedup an artifact of your workload?

dandragona: Another option is you can try to use mhclite. It removes the need for the sinkhorn algorithm so theres no while loop. It also might be easier to fuse because it replace the sinkhorn step with an einsum.

dandragona: Upon further digging, I believe the rewrite to expand_dims being faster may be because dot_general is potentially harder to fuse than element wise operations. The contracting dimensions for H_pre, H_post are quite small (H_post is literally 1) so potentially doing elementwise operations is the move here.

I still think its worth a shot to see if mhclite helps.

dandragona: Here and elsewhere let's leave a comment about this theory and possible link to this discussion.

dandragona (approved):

src/maxtext/layers/mhc.py
h1t35h: Done.

NuojCheng (approved):

PR #4687, opened 2026-07-31
Add TPU Ulysses context parallelism

huytransformer:
## Description

This PR adds `context_parallel_strategy="ulysses"`, DeepSpeed-Ulysses style context parallelism for the TPU splash attention (https://arxiv.org/abs/2309.14509).

Motivations: (1) the existing approaches degrade when the context parallelism degree is large relative to the sequence length. (2) It is also the head-parallel half of a Ulysses x ring hybrid (USP), which can be built on top of this PR.

## Performance

What we did: v5p-128 (64 chips), bf16, synthetic data, 2M tokens per step in every cell, `remat_policy=custom context=device`, `dq_reduction_steps=3`. Each uses its best known block sizes (2048 for Ulysses, 1024 for Ring and All-gather).

### llama2-7b (MHA, 32 KV heads), 32K context, cp=16, 2K tokens per device.

| Method | Step time | TFLOP/s/device | MFU |
|---|---|---|---|
| Ulysses | 8.90 s | 239.4 | 52.2% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-12469201863404864261) |
| All-gather | 8.83 s | 242.7 | 52.9% |
| Ring | 16.38 s | 130.8 | 28.5% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-11372060697866617231) |

Ulysses matches all-gather and runs 1.8x faster than ring, while holding 1/16 of all-gather's attention KV per device (2 of 32 heads).

### llama3.1-8b (GQA, 8 KV heads), cp=4.

| Context (chunk) | Ulysses | All-gather | Ring |
|---|---|---|---|
| 64K (16K) | 243.3 / 53.0% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-15912674528103332053) | 282.5 / 61.6% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-3264586155108480195) | 277.4 / 60.4% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-16826968503513023661) |
| 8K (2K) | 259.0 / 56.4% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-15551305366018762088) | 281.1 / 61.2% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-15050118006705884737) | 274.7 / 59.8% [Xprof](http://xprof.corp.google.com/trace_viewer/htn-7345102037726269701) |

**Correctness:** 20 training steps against `all_gather` on llama3.1-8b at 64K, cp=4, identical data, max loss difference is 0.001.

## Example repro

```
export LIBTPU_INIT_ARGS="--xla_tpu_scoped_vmem_limit_kib=65472 \
  --xla_tpu_use_enhanced_launch_barrier=true \
  --xla_tpu_enable_async_collective_fusion=true \
  --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true \
  --xla_tpu_enable_async_collective_fusion_multiple_steps=true \
  --xla_tpu_overlap_compute_collective_tc=true \
  --xla_enable_async_all_gather=true \
  --xla_enable_async_collective_permute=true \
  --xla_tpu_enable_sparse_core_collective_offload_all_reduce=true \
  --xla_tpu_enable_sparse_core_collective_offload_reduce_scatter=true \
  --xla_tpu_enable_sparse_core_collective_offload_3d_all_gather=true"
```

```
  base_output_directory=${OUTPUT_DIR} run_name=ulysses_cp16 \
  dataset_type=synthetic enable_checkpointing=false steps=10 \
  model_name=llama2-7b max_target_length=32768 per_device_batch_size=1 \
  ici_context_parallelism=16 ici_fsdp_parallelism=4 \
  attention=flash use_tokamax_splash=true use_jax_splash=false \
  context_parallel_strategy=ulysses context_parallel_load_balance=false \
  packing=false dq_reduction_steps=3 remat_policy=custom context=device \
  num_vocab_tiling=16 \
  sa_block_q=2048 sa_block_kv=2048 sa_block_q_dkv=2048 sa_block_kv_dkv=2048
```

## Tests

- `tests/unit/ulysses_attention_test.py`:  Passed.
- `tests/unit/ulysses_collective_test.py`:  Passed.
- `tests/unit/configs_value_test.py`: Passed.
- `tests/integration/train_tests.py`: train smoke test with ulysses cp=4. Passed.

# Checklist


tests/unit/ulysses_collective_test.py
```diff
@@ -0,0 +1,157 @@
...
+import jax
+import jax.numpy as jnp
+from jax.sharding import Mesh, PartitionSpec as P
+import numpy as np
+import pytest
+
+from maxtext.kernels.attention import ulysses_attention
+
+
+@pytest.mark.cpu_only
```
NuojCheng, line 39: I think the default now is cpu_only

src/maxtext/kernels/attention/context_parallel_utils.py
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Shared helpers for attention context-parallel sharding metadata."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import jax
+
```
NuojCheng: nit: I think all these three function are useful in general, we should consider moving them to `sharding.py` in the future

src/maxtext/kernels/attention/ulysses_attention.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def ulysses_all_to_all(tensor: Any, ulysses_axis: str):
+  """Moves `[B, H, S/U, D]` to `[B, H/U, S, D]`."""
+  return jax.lax.all_to_all(tensor, ulysses_axis, split_axis=1, concat_axis=2, tiled=True)
+
+
+def inverse_ulysses_all_to_all(tensor: Any, ulysses_axis: str):
+  """Moves `[B, H/U, S, D]` back to `[B, H, S/U, D]`."""
+  return jax.lax.all_to_all(tensor, ulysses_axis, split_axis=2, concat_axis=1, tiled=True)
```
NuojCheng, line 210: I love you using explicit all2all/shard_map instead of relying on XLA/auto_shard

src/maxtext/layers/attention_op.py
```diff
@@ -1405,6 +1512,30 @@ def create_sa_config(config, query, key, attn_logits_soft_cap):
...
+        splash_kernel = tokamax_splash_kernel.make_splash_mha(
+            mask=single_head_mask,
+            config=sa_config,
+            q_seq_shards=1,
+        )
+        return splash_kernel
+
+      splash_kernel = wrap_ulysses_splash_kernel(mask)
+      segment_axis_names_splash_kernel = jax.sharding.PartitionSpec(None)
+      splash_kernel = self._maybe_shard_with_pspec(splash_kernel, segment_axis_names_splash_kernel)
```
NuojCheng, line 1539: why we want splash kernel been fully replicated? or is it an explicit all gather?

src/maxtext/layers/attention_op.py
```diff
@@ -1579,6 +1712,32 @@ def wrap_flash_attention(
...
+              full_segment_ids,
+          )
+        else:
+          decoder_segment_ids_tuple = None
+        kernel = partial(splash_kernel, max_logit_value=max_logit_value)
+        attention_output = jax.vmap(lambda q, k, v, d, s: kernel(q, k, v, d, sinks=s), in_axes=(0, 0, 0, 0, None))(
+            query, key, value, decoder_segment_ids_tuple, sinks
+        )
+        attention_output = ulysses_attention.inverse_ulysses_all_to_all(attention_output, context_axis)
+        return attention_output, None
```
NuojCheng, line 1740: very clean logic, nice!

tests/unit/attention_test.py
```diff
@@ -1398,6 +1399,203 @@ def ring_loss(lnx):
...
+    generic_grad = jax.device_get(generic_grad)
+    ulysses_grad = jax.device_get(ulysses_grad)
+
+    self.assertTrue(
+        jax.numpy.allclose(generic_grad, ulysses_grad, rtol=1e-02, atol=1e-07, equal_nan=False),
+        msg="Input gradients from generic dot product and flash attention + Ulysses context parallelism are not "
+        f"close. ici_context_parallelism={ici_context_parallelism}.",
+    )
+
+  @pytest.mark.tpu_only
```
NuojCheng, line 1542: if we just wanna check HLO, no need for TPU test. Use CPU test instead

tests/unit/ulysses_attention_test.py
```diff
@@ -0,0 +1,265 @@
...
+      ulysses_attention.validate_ulysses_runtime(model_mode=MODEL_MODE_TRAIN, sinks=object())
+    with self.assertRaisesRegex(ValueError, "indexer"):
+      ulysses_attention.validate_ulysses_runtime(model_mode=MODEL_MODE_TRAIN, indexer_mask=object())
+    with self.assertRaisesRegex(ValueError, "bidirectional"):
+      ulysses_attention.validate_ulysses_runtime(model_mode=MODEL_MODE_TRAIN, bidirectional_mask=object())
+    with self.assertRaisesRegex(NotImplementedError, "record_max_logits"):
+      ulysses_attention.validate_ulysses_runtime(model_mode=MODEL_MODE_TRAIN, record_max_logits=True)
+
+  def test_with_sequence_axis_preserves_partition_spec_type(self):
+    spec = jax.sharding.PartitionSpec("data", None, None, "model")
```
NuojCheng: we don't have `'model'` axis name in maxtext?

tests/unit/hlo_test_utils_test.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2026 Google LLC
```
NuojCheng, line 1: this is a very interesting test

tests/unit/ulysses_attention_test.py
```diff
@@ -0,0 +1,265 @@
+# Copyright 2026 Google LLC
```
NuojCheng, line 1: move this test to attention_test?

tests/unit/ulysses_collective_test.py
```diff
@@ -0,0 +1,157 @@
+# Copyright 2026 Google LLC
```
NuojCheng, line 1: move this test to attention test?

tests/unit/ulysses_collective_test.py
```diff
@@ -0,0 +1,157 @@
...
+
+  np.testing.assert_array_equal(jax.device_get(round_trip(query)), jax.device_get(query))
+
+  # The forward exchange produces each rank's head subset over the full sequence.
+  @partial(
+      jax.shard_map,
+      mesh=mesh,
+      in_specs=P(None, None, "context", None),
+      out_specs=P(None, "context", None, None),
+      check_vma=False,
```
NuojCheng, line 111: what gonna happen if we use check_vma=True?

NuojCheng (approved):

src/maxtext/configs/types.py
```diff
@@ -3628,6 +3631,67 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
           f"ring_scan_unroll={self.ring_scan_unroll} was specified, but is only supported when "
           "context_parallel_strategy='ring'."
       )
+    if context_parallel_strategy == "ulysses":
+      if self.hardware != "tpu":
```
RissyRan, line 3635: Do you know why? If this feature hasn't been tested on GPUs yet, we should label it as 'experimental' and add a disclaimer that GPU execution is unverified and should be used with caution. Similar comments for bellow.

src/maxtext/configs/types.py
```diff
@@ -3628,6 +3631,67 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+        raise ValueError("Ulysses context parallelism (context_parallel_strategy='ulysses') is only supported on TPU.")
+      if self.context_sharding != "context":
+        raise ValueError("TPU Ulysses attention requires context_sharding='context'.")
+      ici_context_parallel_size = self.ici_context_parallelism
+      dcn_context_parallel_size = self.dcn_context_parallelism
+      if ici_context_parallel_size <= 0 or dcn_context_parallel_size <= 0:
```
RissyRan, line 3641: MaxText won't support `ici_context_parallel_size <= 0 or dcn_context_parallel_size <= 0` for other strategies right? This is invalid, not specific to this Ulysses. If so, we could just align with other sharding constratins.

src/maxtext/configs/types.py
```diff
@@ -3628,6 +3631,67 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+        raise ValueError("TPU Ulysses attention does not support context_parallel_load_balance=True.")
+      if self.use_ragged_attention:
+        raise ValueError("TPU Ulysses attention does not support ragged attention.")
+      if self.attention_sink:
+        raise ValueError("TPU Ulysses attention does not support attention sinks.")
+      if self.use_indexer:
+        raise ValueError("TPU Ulysses attention does not support sparse indexer masks.")
+      if self.use_chunked_prefill:
+        raise ValueError("TPU Ulysses attention does not support chunked prefill yet.")
+      if self.use_multimodal:
```
RissyRan, line 3674: Wondering if those features are not tested or needs feature to supported?

```
      if self.use_multimodal:
        raise ValueError("TPU Ulysses attention does not support multimodal attention.")
      if self.enable_dropout and self.dropout_rate > 0.0:
        raise ValueError("TPU Ulysses attention does not support dropout yet.")
...
```

src/maxtext/kernels/attention/context_parallel_utils.py
RissyRan: Yeah, agree. It seems not context parallelism specific.

src/maxtext/layers/attention_op.py
```diff
@@ -568,6 +569,57 @@ def __init__(
             dkv_dim_q=3,
             dkv_dim_kv=3,
         )
+    if self.attention_kernel == "flash" and ulysses_attention.is_context_parallel_ulysses_requested(self.config):
```
RissyRan, line 572: It seems in this PR, we have quite some assertions, inside of types.py, those functions, etc. Do you think we could just keep one version in types.py for early check if no overwritten is allowed/happens during runtime.

RissyRan (commented):
Thanks! LGTM at high level. Some minor comments.

tests/unit/ulysses_collective_test.py
```diff
@@ -0,0 +1,157 @@
...
+  def exchange(tensor):
+    return ulysses_attention.ulysses_all_to_all(tensor, ulysses_axis)
+
+  np.testing.assert_array_equal(jax.device_get(exchange(query)), jax.device_get(query))
+
+  @partial(
+      jax.shard_map,
+      mesh=mesh,
+      in_specs=(
+          P(None, None, "context", None),
```
RissyRan: It seems this test verified context sharding only. We usually will include fsdp + cp with long context. Could you help add a test and ensure those work (to avoid future change breakage)?

src/maxtext/kernels/attention/context_parallel_utils.py
huytransformer: done!

src/maxtext/layers/attention_op.py
huytransformer: no gather it's just mask meta data, after a2a every device need full mask

src/maxtext/configs/types.py
```diff
@@ -3698,6 +3631,67 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+      if self.dq_reduction_steps not in (0, 3):
+        raise ValueError("TPU Ulysses attention requires dq_reduction_steps to be 0 or 3.")
+      if self.use_qk_clip:
+        raise ValueError("TPU Ulysses attention does not support QK-Clip statistics yet.")
+      if self.max_target_length % context_parallel_size != 0:
+        raise ValueError(
+            "TPU Ulysses attention requires max_target_length "
+            f"({self.max_target_length}) to be divisible by context_parallel_size ({context_parallel_size})."
+        )
+      if self.num_query_heads % context_parallel_size != 0:
```
susanbao, line 3687: Is it possible to use ulysses with TP? If so, may check num_query_heads % (CP x TP) !=0.

src/maxtext/configs/types.py
```diff
@@ -3698,6 +3631,67 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+            "TPU Ulysses attention requires num_query_heads "
+            f"({self.num_query_heads}) to be divisible by context_parallel_size ({context_parallel_size})."
+        )
+      if self.num_kv_heads == 1:
+        raise ValueError("TPU Ulysses attention does not support MQA with context_parallel_size > 1.")
+      if self.num_kv_heads % context_parallel_size != 0:
+        raise ValueError(
+            "TPU Ulysses attention requires num_kv_heads "
+            f"({self.num_kv_heads}) to be divisible by context_parallel_size ({context_parallel_size})."
+        )
```
susanbao, line 3698: As "context_parallel_load_balance=true" is default. It is better to have the validation for context_parallel_load_balance when context_parallel_strategy == "ulysses" and explain why it must be disabled.

src/maxtext/utils/sharding.py
```diff
@@ -192,6 +190,36 @@ def remove_size_one_mesh_axis(spec, mesh):
...
+  """Returns the product of mesh sizes for a set of axes."""
+  size = 1
+  for axis in axes:
+    if axis not in mesh.shape:
+      raise ValueError(f"{label} requires mesh axis {axis!r} to exist.")
+    size *= mesh.shape[axis]
+  return size
+
+
+def with_axis_on_dim(axis_names, axis, dim):
```
susanbao, line 212: Add a check if axis_names is None: return None and validate dim in  axis_names with a descriptive error.

susanbao (commented):
Have added the comments. Please check. It is great to have Ulysses in maxtext.

src/maxtext/utils/sharding.py
huytransformer: Done, added both :)

src/maxtext/configs/types.py
huytransformer: ya can you check line 3660

src/maxtext/configs/types.py
huytransformer: it is, but i think TP doesnt always shards heads? validate_head_sharding already checks the local heads against the real mesh

src/maxtext/configs/types.py
huytransformer: it can't run on GPU at all (GPU flash goes through the TE/cudnn path) that's why hard error instead of experimental label :)

src/maxtext/configs/types.py
NuojCheng: TP always shards heads unless using customized logical rule

src/maxtext/configs/types.py
huytransformer: -1 is actually valid for all_gather/ring so can't ban it globally :) ulysses needs the concrete size here for the head/seq divisibility checks below

src/maxtext/configs/types.py
huytransformer: i think most require implementation (some may just need testing like packing which I will unlock next)

tests/unit/ulysses_collective_test.py
huytransformer: added, it now also runs a 2d fsdp=2 x context=2 mesh with batch sharded over fsdp

src/maxtext/layers/attention_op.py
huytransformer: TY. I removed the load_balance (already covered in types,py).
attention_type/dropout are per-layer args (e.g. gemma3 passes local_sliding per layer) and the head checks need the real mesh

tests/unit/ulysses_attention_test.py
huytransformer: this test launches its own file as a subprocess to fake 4 cpu devices before jax starts. if it live in attention_test it would launch all of attention_test instead

tests/unit/ulysses_attention_test.py
huytransformer: good catch, renamed to tensor

src/maxtext/kernels/attention/ulysses_attention.py
huytransformer: idk what this says. i see i love you i stopped reading ;)

tests/unit/attention_test.py
huytransformer: we need the final compiled HLO and this program only compiles on TPU

tests/unit/ulysses_collective_test.py
huytransformer: passed

src/maxtext/configs/types.py
huytransformer: mmm ok. also i think tensor_sequence also shard heads

src/maxtext/configs/types.py
RissyRan: I see. -1 means all/rest to that parallelism.

RissyRan (approved):

susanbao (approved):

PR #4691, opened 2026-07-31
Validate rampup batch size settings in the Pydantic config

### LINKED ISSUE #4690: Rampup batch size settings are not validated in the Pydantic config
The legacy config validates the rampup batch size settings, the Pydantic config does not. Invalid combinations are accepted and either switch rampup off with no diagnostic, or ramp to a batch size that was never requested.

`pyconfig_deprecated.validate_rampup_batch_size` (src/maxtext/configs/pyconfig_deprecated.py:203) asserts five things: `per_device_batch_size_start > 0`, `per_device_batch_size_increment > 0`, `global_rampup_samples > 0`, `per_device_batch_size - per_device_batch_size_start > 0`, and that the difference divides by the increment.

`DatasetGeneral` in src/maxtext/configs/types.py declares the same four fields around line 1365 with no validator. The schedule derivation in `MaxTextConfig` around line 3067 is written defensively, `if self.global_batch_size_to_load_increment > 0:` then `if num_increments > 0:`, so when either guard fails the block falls through and leaves `rampup_end_step = 0`.

Replicating that arithmetic for 8 devices, `expansion_factor_real_data=1`, `gradient_accumulation_steps=1`:

| settings | rampup_end_step | result |
| --- | --- | --- |
| `per_device_batch_size=8, start=4, increment=2, samples=500` | 14 | correct |
| `increment=0` | 0 | rampup silently off |
| `global_rampup_samples=0` | 0 | rampup silently off |
| `start=8, per_device_batch_size=4` | 0 | rampup silently off |
| `per_device_batch_size=9, start=4, increment=2` | 14 | ramps to global batch size 64, 72 was requested |

The first three mean a run configured with `enable_rampup_batch_size=True` trains at a constant batch size with nothing in the logs saying rampup never engaged. The last one is worse, `num_increments = diff // increment` truncates, so ramp-up finishes below the configured `per_device_batch_size` and stays there.

The Pydantic config should reject these at parse time the way the legacy path does. I can send a PR adding a `model_validator(mode="after")` on `DatasetGeneral` with unit tests in tests/unit/configs_value_test.py.

Atishyy27:
Closes #4690.

The legacy config validates the rampup batch size settings in `pyconfig_deprecated.validate_rampup_batch_size` (src/maxtext/configs/pyconfig_deprecated.py:203). The Pydantic config has no equivalent, so an unusable combination is accepted and then silently degrades.

The derivation in `MaxTextConfig` is guarded by `if self.global_batch_size_to_load_increment > 0:` and `if num_increments > 0:`, so a non-positive increment, non-positive `global_rampup_samples`, or a start at or above `per_device_batch_size` falls through and leaves `rampup_end_step = 0`. Rampup is off and nothing says so. When the change is not a multiple of the increment, `num_increments = diff // increment` truncates and ramp-up finishes below the configured `per_device_batch_size`.

Replicating the derivation arithmetic for 8 devices, `expansion_factor_real_data=1`, `gradient_accumulation_steps=1`:

| settings | rampup_end_step | result |
| --- | --- | --- |
| `per_device_batch_size=8, start=4, increment=2, samples=500` | 14 | correct |
| `increment=0` | 0 | rampup silently off |
| `global_rampup_samples=0` | 0 | rampup silently off |
| `start=8, per_device_batch_size=4` | 0 | rampup silently off |
| `per_device_batch_size=9, start=4, increment=2` | 14 | ramps to global batch size 64, 72 requested |

This adds a `model_validator(mode="after")` on `DatasetGeneral` with the same five checks as the legacy path, skipped entirely when `enable_rampup_batch_size` is False.

Tests in tests/unit/configs_value_test.py cover each rejection, a positive control asserting a well-formed config still produces `rampup_end_step > 0`, and a case confirming the checks do not fire when rampup is disabled.

# Checklist


NuojCheng (approved):

tests/unit/configs_value_test.py
```diff
@@ -268,6 +268,65 @@ def test_load_balanced_chunk_context_parallel_config(self):
...
+        argv = base_args + bad_args
+        with unittest.mock.patch("jax.devices", return_value=mock_devices):
+          with self.assertRaisesRegex((ValueError, pydantic.ValidationError), expected_regex):
+            pyconfig.initialize(argv)
+
+    # A well-formed rampup config is still accepted and still produces a schedule.
+    with unittest.mock.patch("jax.devices", return_value=mock_devices):
+      config = pyconfig.initialize(base_args)
+    self.assertTrue(config.enable_rampup_batch_size)
+    self.assertGreater(config.rampup_end_step, 0)
```
aireenmei: Should we make this match the exact end step, not just > 0 ?

aireenmei (approved):

Atishyy27:
Thanks for the review. Anything still needed from me here, or is this just waiting on a merge?

tests/unit/configs_value_test.py
Atishyy27: Done in 0a98cb9, it now asserts `rampup_end_step == 11`. With 8 devices the load batch goes 32 to 128 in three steps of 32, so 500 samples give 6 + 3 + 2 steps. Sorry for missing this thread.

PR #4704, opened 2026-08-03
Add Hy3 (Tencent Hunyuan V3) model support

weikuo0506:
Adds Tencent Hunyuan V3 (Hy3-295B) model support.

**Hy3** is a 295B-parameter Mixture-of-Experts (MoE) foundation model (`tencent/Hy3` on Hugging Face) characterized by:
- **Parameter Scale:** 295.034B total parameters, ~21B active parameters per token.
- **Layer Architecture:** 80 Transformer decoder layers, where **Layer 0 is dense** (`first_num_dense_layers = 1`) and **Layers 1–79 are MoE**.
- **Attention & Normalization:** Grouped Query Attention (GQA) with 64 Query heads, 8 KV heads (`head_dim = 128`), combined with head-wise RMS Query-Key Normalization (**QK-Norm**) and RoPE.
- **MoE Routing:** DeepSeek-style auxiliary-loss-free routing with **1 shared expert** and **192 routed experts** (Top-8 routed selected per token), utilizing Sigmoid scoring with learnable routing biases and a fixed routed scaling factor of `2.826`.

This is a new-model bring-up. No prior release knew about `decoder_block: "hy3"`, so every `== DEEPSEEK` that becomes `in (DEEPSEEK, HY3)` below is registration of a new model, not a fix for a defect that was reachable before.

This PR includes:
- MoE routing adjustments (`layers/moe.py`): support for sigmoid router
  routing gates (pre-bias logits capture + `routed_scaling_factor` application),
  and the relevant `types.py` validation guards (including the `ModelName` literal
  allowlist and the loss-free-load-balancing decoder_block check).
- New `src/maxtext/models/hy3.py`: `Hy3DenseLayer`/`Hy3MoELayer`.
- New `hy3-tiny.yml` / `hy3-295b.yml` model configs.
- Registered Hy3 in the checkpoint conversion framework (`hf_model_configs.py`,
  `param_mapping.py`, `hf_shape.py`, `globals.py`'s `HF_IDS`). `transformers`
  already ships a native `HYV3Config`; `hf_model_configs.py` uses it via the
  same try/except native-class-with-`PTConfig`-fallback pattern already used
  for Gemma 4, so no `trust_remote_code`-only workaround is needed. HF tensor
  names/shapes were cross-checked against the real `tencent/Hy3` checkpoint's
  `model.safetensors.index.json` (46,545 non-MTP tensors, zero missing/extra).
  MTP is not mapped — see the MTP section below.
- `tests/unit/hy3_vs_reference_test.py`: parity tests for Hy3's MLP, MoE block,
  dense decoder layer and MoE decoder layer, against the PyTorch reference
  imported from `transformers.models.hy_v3` (the same approach
  `deepseek_v4_vs_reference_test` and `qwen3_5_layers_test` take). Toy
  dimensions on random weights initialized in PyTorch and copied into the JAX
  modules, so no checkpoint download is needed. Guarded by a skipif: torch only
  ships in `tpu-post-train-requirements.txt`, and this file is not in
  pytest.ini's `--ignore` list, unlike every other `*_vs_reference_test.py`.
- `tests/unit/nnx_decoders_test.py`: `NNXDecoder` forward-pass test with the
  `hy3` decoder block (1 dense + 3 MoE layers, unscanned).
- `tests/end_to_end/tpu/hy3/Run_Hy3.md`: end-to-end user guide (checkpoint
  conversion, pre-training, fine-tuning, decoding, logit verification).
- FLOPs/MFU accounting (`maxtext_utils.py`): registered HY3 in
  `get_dense_moe_layers` and in 3 branches of
  `calculate_tflops_training_per_device`. This is new-model registration, not a
  fix — no prior release knew about `decoder_block: "hy3"` — but skipping it
  fails silently rather than loudly: an unregistered block falls through to the
  generic MoE path, which has no shared-expert term and applies one MoE-shaped
  formula to every layer, including the dense first layer.

  Only two of the four sites move numbers today. `get_dense_moe_layers` goes
  from (0 dense, 80 MoE) to (1, 79), and the FFN dispatch picks up
  `calculate_routed_and_shared_ffn_tflops_per_device`. The weight-sum branch is
  deliberately identical to the generic fallback, added so DeepSeek/Hy3 don't
  silently inherit a future change to it; the MTP branch is inert until
  `mtp_num_layers > 0`, which neither hy3 yml sets.

  Measured on `hy3-295b` (`per_device_batch_size=1`, `max_target_length=4096`)
  by deleting the registrations from a scratch copy: `learnable_weight_tflops`
  459.01 -> 495.95 (+8.0%), of which +36.65 is the shared expert across 79 MoE
  layers and +0.29 the net effect of treating layer 0 as dense.
  `attention_tflops` is unchanged, as expected. Reported TFLOP/s and MFU only;
  training itself is unaffected.
- `grpo_utils.py` / `generate_param_only_checkpoint.py`: extended the
  dense+MoE two-stack layer-group handling from DEEPSEEK to HY3 in 4 places
  (GRPO's `pathways_reshard`, plus the Linen, NNX and LoRA-NNX unroll paths).
  Like DeepSeek, Hy3 has a dense prefix (`first_num_dense_layers`), so its
  scanned parameter tree is split into `decoder/dense_layers` and
  `decoder/moe_layers` rather than a single `decoder/layers`. Without the
  registration the unroll paths look up a key that isn't there and fail
  outright — `ValueError: Missing layers in training_state or
  training_state_annotations.` in generate_param_only_checkpoint, `KeyError:
  "layers"` in `max_utils.unscan_train_state_params`. Only reachable when
  layers are being unscanned: `force_unroll` for the checkpoint path,
  `scan_layers=False` for GRPO.
- `train.py`: registered HY3 in `_MOE_BLOCK_ATTR_BY_DECODER_BLOCK`, the table
  the legacy Linen decoder bridge uses to find the MoE submodule by attribute
  name (`"DeepSeekMoeBlock_0"`). The default NNX path no longer reads this
  table — an upstream commit replaced it with a generic `_find_gate_bias`
  lookup — so this entry only affects the Linen bridge.
- Renamed `is_deepseek` to `has_dense_prefix` in `nnx_decoders.py`: adding Hy3
  to that flag turned an accurate name into a false one, and eight call sites
  read `if self.is_deepseek:`. Also updates the one consumer in `maxengine.py`,
  which reads the attribute by name via `getattr(..., False)` — missing it
  would silently select the wrong KV-cache layer keys rather than raise. No
  behavior change.
- `use_batch_split_schedule=true` now requires `attention_type=mla`. The batch-split path lives inside the `has_dense_prefix` branch, so putting Hy3 into that flag let Hy3 reach it for the first time; it reads MLA parameters by name and Hy3 uses plain GQA. Reproduced on a forward pass as `KeyError: 'wq_a'`. Guarded on the attention type rather than the decoder block, since the real precondition is that those parameters exist -- every current batch-split user (deepseek3-671b, deepseek3-671b-batchsplit, deepseek3-tiny, kimi-k2-1t) is already `mla`.

## Load balancing

`load_balance_loss_weight` (the gradient-based auxiliary loss) works in both
scan modes.

Load balancing: `routed_bias_update_rate` works with `scan_layers=true`. With `scan_layers=false` the shared `train.py` path applies the last MoE layer's update to layer 0 and leaves the rest untouched -- a pre-existing problem it shares with DeepSeek V3, documented at that line. Rather than change DeepSeek's behaviour from a model-onboarding PR, Hy3 now rejects the combination at config validation time, so it fails fast instead of training on a silently wrong bias. Both hy3 configs default `routed_bias_update_rate` to 0.0.

Hy3's router bias stays an `nnx.Param`. Upstream keeps it in a separate
`MoEBiasVar` collection for `("deepseek3", "deepseek4", "kimi-k2")`; Hy3 is
deliberately not in that list. `MoEBiasVar` moves the bias out of the `params`
collection, and checkpoint conversion reads only that collection
(`to_maxtext.py`: `get_abstract_param(...)["params"]`), so joining the list
would take the bias out from under Hy3's conversion mapping — verified
complete at 30 entries for 30 parameters. The isolation only matters when
`routed_bias_update_rate > 0`, which both hy3 configs leave at 0.0, so nothing
is lost by waiting. `test_moe_block_matches_deepseek` asserts the difference is
exactly `MoeBlock_0/gate/bias`, so any other divergence still fails.

## Multi-Token Prediction (MTP)

Hy3's checkpoint carries a multi-token-prediction layer at index 80. Not
implemented in this PR. Both hy3 configs leave `mtp_num_layers` at 0, so no MTP
module is built and nothing is left uninitialized. This matches upstream
transformers, which also drops the layer
(`_keys_to_ignore_on_load_unexpected = [r"model\.layers\.80.*"]`).

## Muon optimizer / layerwise quantization

Muon: added HY3 to the optimizer allowlist in `configs/types.py`. That
allowlist means "dimension numbers have been reviewed", so this is backed by
`TestHy3MatchesWhitelistedModels` in `tests/unit/muon_utils_test.py`, which
asserts Hy3's attention dimension numbers are identical to already-listed
qwen3's and its MoE ones identical to deepseek3's — Hy3 recombines two
reviewed layouts rather than introducing a new one. Establishes the dimension
numbers, not convergence. Only affects `opt_type=muon`; the default is adamw.

Not added: the Linen layerwise-quantization path still asserts DEEPSEEK. The
default pure-NNX path is model-agnostic and already handles Hy3, and the Linen
per-layer path is legacy, so adapting the `DeepSeek*ToLinen` wrappers and the
hardcoded `DeepSeekMoeBlock_0` AQT scope has no current benefit.

# Tests

- Random-init CPU forward pass on `hy3-tiny.yml`, `scan_layers=True/False`,
  with `mtp_num_layers=1` (MTP reuses the last decoder layer class generically
  per `models.py`'s existing wiring — verified this works for Hy3 too), and a
  real 3-step training run confirming loss decreases normally.
- Real-weight verification, `scan_layers=False`: converted the real
  `tencent/Hy3` checkpoint truncated to 2 layers via `to_maxtext.py`,
  generated golden logits from the real HF reference model (also truncated
  to 2 layers, same real weights), compared via `forward_pass_logit_checker.py`:
  - prompt "The capital of France is": max KL divergence 1.34e-07, top-10
    token overlap 10/10
  - prompt "I love to": max KL divergence ~0 (-4.37e-09, floating-point
    noise), top-10 token overlap 10/10
  (Both far under the `--max_kl_div=0.5` threshold used.)
- `pytest tests/unit/configs_test.py` — full suite (78 tests) passes.
- `pytest tests/unit/hy3_vs_reference_test.py` — 4/4 pass on CPU (4 skipped
  with torch uninstalled): Hy3's MLP, MoE block, dense decoder layer and MoE
  decoder layer match the `transformers.models.hy_v3` reference. 
- `pytest tests/unit/muon_utils_test.py` — passes, including
  `TestHy3MatchesWhitelistedModels`.
- `pytest tests/unit/nnx_decoders_test.py -k hy3` — 1/1 pass.
- `tests/unit/train_compile_test.py::test_hy3` — passes on a TPU-enabled JAX
  build at full 80-layer/295B scale, for both `linen_scanned` and
  `nnx_scanned`.
- Static conversion coverage: `tests/unit/param_mapping_test.py` gains seven Hy3 cases, running in CI without a checkpoint or a TPU. They cover both scan modes: that the dense and MoE layer groups are both present and that scanning collapses each into one key holding a list of HF names; that unscanned layers keep the HF index for dense layers and restart at 0 for MoE layers; that routed experts expand per layer over the expert axis; that the three HF names which differ from DeepSeek/Qwen convention (`mlp.router.gate.weight`, `mlp.expert_bias`, `mlp.shared_mlp.*`) are pinned literally; that layer `num_hidden_layers` (MTP) appears in no mapping value; that the mapping and `hf_shape.HY3_HF_WEIGHTS_TO_SHAPE` agree exactly -- 62 mapping keys expand to 89 HF names and the shape map holds those same 89, nothing extra on either side; and that the exported HF config declares no `num_nextn_predict_layers` and disables fp32 MoE combine. The second-to-last is the "zero missing, zero extra" property this PR claims against the real checkpoint, now enforced in CI at toy scale.
- Merged `upstream/main` (280 commits) into the branch; conflicts in `moe.py`,
  `nnx_decoders.py` and `muon_utils_test.py` resolved (see `9056a35`).
- Confirmed no regressions: `test_deepseek_configs` (6/6 configs) passes, and
  re-ran a live `deepseek3-tiny` training reproduction
  (`routed_bias_update_rate=0.05`, `load_balance_loss_weight=0.01`,
  `scan_layers=true`) — all 3 steps complete cleanly, `moe_lb_loss` nonzero
  throughout (`0.0814 -> 0.0805 -> 0.0803`). Our only change in `train.py` is
  additive (added `HY3` alongside the existing `DEEPSEEK` case).
- **Distributed Scaling Verification (Google Cloud TPU v5p-64)**:
  Successfully executed a 30-step pre-training benchmark of the **full 80-layer/295B model** on a Google Cloud TPU v5p-64 slice (`4x4x4` 3D Torus) using Pure FSDP (`TP=1, EP=1, FSDP=64`), `max_target_length=8192`, and `per_device_batch_size=4.0` (2.1M tokens per step).
  - Memory footprint: ~65 GB / 95.74 GB HBM (Completed without OOM for this v5p-64 workload).
  - Compute efficiency: **122.43 TFLOP/s / device (26.67% MFU)**.
  - Global throughput: **51,115 tokens / second**.
  - *(Note: Expert parallelism `ici_expert_parallelism > 1` collective behavior at runtime remains untested; FSDP-64 was utilized here).*
  <details><summary>Click to view verbatim execution log snippet (Steps 26-29)</summary>

  ```text
  I0812 04:04:13.536221 metric_logger.py:226] completed step: 26, seconds: 41.180, TFLOP/s/device: 121.979, Tokens/s/device: 795.727, total_weights: 2097152, loss: 11.839
  I0812 04:04:54.699895 metric_logger.py:226] completed step: 27, seconds: 41.168, TFLOP/s/device: 122.015, Tokens/s/device: 795.960, total_weights: 2097152, loss: 11.838
  I0812 04:05:35.857697 metric_logger.py:226] completed step: 28, seconds: 41.168, TFLOP/s/device: 122.015, Tokens/s/device: 795.960, total_weights: 2097152, loss: 11.838
  I0812 04:06:17.021406 metric_logger.py:226] completed step: 29, seconds: 41.164, TFLOP/s/device: 122.028, Tokens/s/device: 796.045, total_weights: 2097152, loss: 11.838
  Per train step:
   Total TFLOPs: 5023.10 
   split as 78.99% learnable weight flops and 21.01% attention flops
  ```
  </details>

# Checklist

- [x] I have made or will make corresponding changes to the doc if needed (added a
      `### Hy3` entry to `docs/reference/models/supported_models_and_architectures.md`
      and a `Run_Hy3.md` end-to-end guide).


src/maxtext/utils/maxtext_utils.py
```diff
@@ -1154,7 +1154,12 @@ def calculate_tflops_training_per_device(config, log=True):
         DecoderBlockType.QWEN3_5,
         DecoderBlockType.GEMMA4,
         DecoderBlockType.DEEPSEEK4,
+        DecoderBlockType.HY3,
     ):
+      # Hy3 has DeepSeek's routed + shared + leading-dense structure. The
```
entrpn: I don't fully understand this message, does that mean the MFU reporting will be wrong?

entrpn (commented):
Can you also add a compile test in this file: https://github.com/weikuo0506/maxtext/blob/main/tests/unit/train_compile_test.py

RissyRan:
I just sent you the permission to push branch directly to the MaxText repo. In the future, you could do it, so Gemini PR could be triggered (you currently forked it, and unfortunately, due to key issue, Gemini review is disabled)

docs/reference/models/supported_models_and_architectures.md
```diff
@@ -66,6 +66,11 @@ MaxText is an open-source, high-performance LLM framework written in Python/JAX.
 - **Variants**: K2 (1T), K2-Thinking (1T), K2.5 (text), K2.6 (text)
 - **Notes**: DeepSeek V3 architecture; MuonClip optimizer
 
+### Hy3
+
+- **Variants**: Hy3 (295B, MoE 21B-A active)
```
RissyRan, line 71: There is 3.8B MTP, are you planning to onboard/verify as well? If not, could you make a comment saying MTP is not supported for this model?

src/maxtext/configs/models/hy3-tiny.yml
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test config for Hy3 (Tencent Hunyuan V3, https://huggingface.co/tencent/Hy3).
+# Small dims/layer count for fast local iteration with random init; architecturally
+# significant fields (routing, norm epsilon, QK-Norm) keep their real hy3-295b values.
+
+decoder_block: "hy3"
```
RissyRan, line 20: Could you make a comment this is a test config?

src/maxtext/utils/maxtext_utils.py
RissyRan: +1, could you help explain a little bit more?

src/maxtext/utils/maxtext_utils.py
```diff
@@ -1248,7 +1253,9 @@ def calculate_tflops_training_per_device(config, log=True):
     attention_tflops, learnable_weight_tflops = calculate_deepseek4_tflops_training_per_device(
         config, total_ffn_flops_all_layers, embedding_flops
     )
-  elif config.decoder_block == DecoderBlockType.DEEPSEEK:
+  elif config.decoder_block in (DecoderBlockType.DEEPSEEK, DecoderBlockType.HY3):
+    # total_ffn_flops_all_layers is already summed over layers by the helper
+    # above; the generic branch would multiply by num_decoder_layers again.
```
RissyRan: `the generic branch would multiply by num_decoder_layers again` this seems wrong to me, could you help explain?

RissyRan (commented):
LGTM! Thanks for the very detailed explanation in PR description! 

At high level, could you help:

1) NNX of LB PR has merged, could you help check if works in your case?
2) Attach the forward logit test & test against reference implementation logs? As those are not tested in CI/CD
3) Check if [section 3 & 4](https://screenshot-v2.corp.google.com/5r23reu8lpbe0) are done, before the claim?

weikuo0506:
Thanks both for the thorough review! Addressed everything below.

**@entrpn**

1. Added `test_hy3` to `tests/unit/train_compile_test.py`, mirroring `test_deepseek4` (parameterized over `linen_scanned`/`nnx_scanned`, `compile_topology=v5p-256`). Verified it actually passes on a real TPU-enabled JAX build at the full declared 80-layer/295B scale (`2 passed` for both variants).

2. Re: the `maxtext_utils.py` comment: without this fix, Hy3's reported MFU/TFLOPs would be undercounted (missing shared-expert FLOPs, and the dense first layer counted as if it were MoE). Comment reworded to say that directly -- full detail in the reply to RissyRan's line comment below, since it's the same code.

**@RissyRan**

1. **NNX load-balancing PR**: confirmed, and mostly resolved -- rebased onto latest `main`, which picked up a second, more recent upstream fix (`263b8c18e`) beyond the one I originally found. Re-verified with a CPU smoke test (`hy3-tiny`, `routed_bias_update_rate=0.05`, `load_balance_loss_weight=0.01`), and this time checked each MoE layer's bias individually, not just the aggregate norm -- which turned up a real remaining gap, so here's the accurate picture:
   - `load_balance_loss_weight` (gradient-based aux loss): works correctly in both scan modes -- `moe_lb_loss` nonzero every step regardless of `scan_layers`.
   - `routed_bias_update_rate` (EMA-style bias update) with `scan_layers=true`: works correctly. The scanned decoder stores all layers' biases in one stacked array, so each layer updates independently and correctly -- confirmed per-layer over 4 steps (3 MoE layers): `[8.5e-05]*3 -> [0.1416, 0.1416, 0.1328] -> [0.2061, 0.2295, 0.1875] -> [0.2891, 0.1328, 0.2354] -> [0.2734, 0.2061, 0.1738]`.
   - `routed_bias_update_rate` with `scan_layers=false`: **still broken**, differently than before. `263b8c18e` fixed the previous total failure (crash / missing intermediates), but the code that applies the update still only finds the *first* MoE layer's bias (unscanned layers are separate attributes, not one stacked array) and applies whichever layer's delta was collected *last* -- so one layer gets updated with the wrong data, and the rest never get updated at all. Confirmed per-layer over 3 steps: layer 0 `8.5e-05 -> 0.1416 -> 0.2832 -> 0.4258` (using layer 2's delta), layers 1 and 2 essentially flat (`~1e-4`, only optimizer-gradient drift). This is upstream `train.py` code shared with DeepSeek, not something in scope to fix here -- documented in the PR description and `Run_Hy3.md`, with a recommendation not to enable `routed_bias_update_rate` under `scan_layers=false` until it's fixed upstream.

   This also let us drop our own attribute-name lookup from the NNX path entirely (still needed for the older Linen bridge, which the upstream fix didn't touch).

2. **Forward-logit reference logs**: re-ran the truncated (2-layer) real-weight verification and attached the log below.

   - "The capital of France is": max KL divergence `1.34e-07`, top-10 token overlap `10/10`.
   - "I love to": max KL divergence `~0` (`-4.37e-09`, floating-point noise), top-10 token overlap `10/10`.

   (Previously reported as 2.2e-4 / 4.5e-4 in the PR description from an earlier run -- both tighter this time, and the PR description now reflects these new numbers too.)

   <details>
   <summary>Forward-pass comparison log (both prompts)</summary>

   ```text
   --- Comparing forward pass for golden data index: 0 ---
   prompt="The capital of France is" raw ids=[628, 6801, 279, 9391, 316], logits.shape=(5, 120832)

   [numerical difference]
   Max absolute difference: 5.25e-06 at index (1, 3097) (Train: -4.76e+00, Golden: -4.76e+00)
   Max relative difference: 9.55e-02 at index (4, 60412) (Train: 5.36e-06, Golden: 5.93e-06)

   --- MaxText model top 10 tokens ---          --- HF model top 10 tokens ---
   | ID    | Token      | Score  |              | ID    | Token      | Score  |
   |-------|------------|--------|              |-------|------------|--------|
   | 66258 | spelled    | 3.3296 |              | 66258 | spelled    | 3.3296 |
   | 26587 | lice       | 3.2572 |              | 26587 | lice       | 3.2572 |
   | 76955 | abella     | 3.1835 |              | 76955 | abella     | 3.1835 |
   | 61561 | depended   | 3.1655 |              | 61561 | depended   | 3.1655 |
   | 74554 | contend    | 2.9235 |              | 74554 | contend    | 2.9235 |
   | 87508 | Berm       | 2.8601 |              | 87508 | Berm       | 2.8601 |
   | 23285 | fatal      | 2.8285 |              | 23285 | fatal      | 2.8285 |
   | 74587 | 等处        | 2.8185 |              | 74587 | 等处        | 2.8185 |
   | 86165 | Spell      | 2.8182 |              | 86165 | Spell      | 2.8182 |
   | 6266  | generally  | 2.8140 |              | 6266  | generally  | 2.8140 |

   --- Similarity Metrics of Top Tokens ---
   | overlap_count | 10/10 |  jaccard_similarity | 1.0 |  rank_agreement_percentage | 100.0 |

   [KL divergence]
   KL divergence = [1.3447776e-07, -7.0248136e-09, -5.7446812e-09, 4.6311371e-08, 6.3416479e-08]
   max KL divergence = 1.3447775870645273e-07 at index 0, token id 628

   --- Comparing forward pass for golden data index: 1 ---
   prompt="I love to" raw ids=[40, 6347, 287], logits.shape=(3, 120832)

   [numerical difference]
   Max absolute difference: 5.01e-06 at index (1, 78062) (Train: 2.59e+00, Golden: 2.59e+00)
   Max relative difference: 1.04e-01 at index (0, 67241) (Train: 9.08e-06, Golden: 8.23e-06)

   --- MaxText model top 10 tokens ---          --- HF model top 10 tokens ---
   | ID    | Token      | Score  |              | ID    | Token      | Score  |
   |-------|------------|--------|              |-------|------------|--------|
   | 13417 | recognize  | 3.6079 |              | 13417 | recognize  | 3.6079 |
   | 3695  | address    | 3.3049 |              | 3695  | address    | 3.3049 |
   | 5012  | maintain   | 3.2655 |              | 5012  | maintain   | 3.2655 |
   | 19447 | combat     | 3.0653 |              | 19447 | combat     | 3.0653 |
   | 69818 | forgive    | 2.9983 |              | 69818 | forgive    | 2.9983 |
   | 25102 | Arg        | 2.9601 |              | 25102 | Arg        | 2.9601 |
   | 3304  | consider   | 2.8604 |              | 3304  | consider   | 2.8604 |
   | 51604 | ,...       | 2.8603 |              | 51604 | ,...       | 2.8603 |
   | 61108 | TIM        | 2.8326 |              | 61108 | TIM        | 2.8326 |
   | 5718  | talk       | 2.8205 |              | 5718  | talk       | 2.8205 |

   --- Similarity Metrics of Top Tokens ---
   | overlap_count | 10/10 |  jaccard_similarity | 1.0 |  rank_agreement_percentage | 100.0 |

   [KL divergence]
   KL divergence = [-4.3702464e-09, -1.5038137e-07, -5.9303801e-08]
   max KL divergence = -4.370246386997678e-09 at index 0, token id 40
   ```

   Note: top tokens look incoherent because this is a 2-layer truncated model, not the full 80-layer Hy3 -- the point is that MaxText and HF produce the same (equally incoherent) output from the same truncated weights, not that the output is meaningful text.
   </details>

3. **The `maxtext_utils.py` comment on the `learnable_weight_tflops` branch**: you're right, that claim was wrong -- there's no double-counting bug there. I compared the formula against the actual generic fallback and they're byte-identical; extending the tuple at that spot doesn't change any computed value. Reworded to say the branch is kept explicit for future-proofing, not as a functional fix.

   The other comment a few lines up (on `total_ffn_flops`) also had a wrong detail -- fixed to state the real cause: no shared-expert term, and the fallback can't express a dense/MoE layer split.

4. **MTP**: that line was wrong, fixed. `supported_models_and_architectures.md` now correctly states MTP isn't onboarded/verified in this PR.

5. **`hy3-tiny.yml`**: added a more prominent test-config comment above `decoder_block:`.

src/maxtext/utils/maxtext_utils.py
weikuo0506: Fixed -- the comment was wrong. The real issue isn't `mlp_dim` vs `moe_mlp_dim` confusion, it's that this generic fallback has no shared-expert FLOPs term at all, and applies one MoE-shaped formula uniformly to every layer (including the dense first layer) -- it can't express a mixed dense/MoE split. Reworded the comment to say that directly. Full context in the summary comment above.

src/maxtext/utils/maxtext_utils.py
weikuo0506: You're right, that claim was wrong -- I checked and this branch is byte-identical to the generic `else` fallback below it, so there's no double-counting bug. Reworded the comment to say it's kept as an explicit branch for future-proofing (so DeepSeek/Hy3 don't silently inherit changes to that fallback), not as a functional fix.

docs/reference/models/supported_models_and_architectures.md
weikuo0506: Not onboarding/verifying MTP in this PR -- added a note to that effect right here, matching what the PR description's "Known limitation 2" already says.

src/maxtext/configs/models/hy3-tiny.yml
weikuo0506: Added a more prominent comment right above `decoder_block:` -- the file already had a top-of-file comment saying this was a test config, but that's easy to miss from this specific line; now it's called out right here too.

docs/reference/models/supported_models_and_architectures.md
weikuo0506: Update on wording: the "randomly initialized" phrasing was wrong and has been removed everywhere. Both hy3 configs leave `mtp_num_layers` at 0, so MaxText builds no MTP module and no weight is left uninitialized -- the checkpoint's layer 80 is simply not mapped, matching upstream transformers, which drops it too. The exported HF config no longer declares `num_nextn_predict_layers` either, so it does not advertise an MTP module it does not ship.

src/maxtext/utils/maxtext_utils.py
weikuo0506: No -- MFU reporting is correct. Hy3 is registered in the list right above, so it uses `calculate_routed_and_shared_ffn_tflops_per_device`, which accounts for the dense prefix and the shared experts. The comment was describing what the generic fallback *would* do if Hy3 weren't registered, which reads like a report of current behaviour. Reworded to say only what the branch does.

src/maxtext/utils/maxtext_utils.py
weikuo0506: You're right, that comment was wrong -- there is no second multiplication. `total_ffn_flops_all_layers` is multiplied by `num_decoder_layers` in exactly one place, guarded by `is_ffn_flops_already_total`, which this branch sets. Both this branch and the generic fallback just consume the value; the two formulas are in fact identical. Fixed in 28daa83.

weikuo0506:
Update: this combination is now rejected at config validation time for Hy3 (`routed_bias_update_rate > 0` with `scan_layers=false`), so it fails fast rather than relying on this note. The underlying issue in the shared `train.py` path is unchanged and still affects DeepSeek V3; it is documented at that line rather than fixed here.

PR #4735, opened 2026-08-04
add static LHS scaling support for GMM v2 FWD

CaptainO5:
# Description

This change introduces static LHS (activation) scaling support to the GMM v2 integration in `ops.py`. 

Specifically, it adds a new `_fwd_prepare_lhs_scale` helper to extract the static activation scale directly from the provided Qwix `quantization_rule` (triggering when fixed calibration bounds are utilized). The GMM v2 forward pass (`_fwd_run_tokamax_v2`) is then updated to accept the quantization rule and pass the extracted `lhs_scale` down into the underlying GMM v2 kernel to seamlessly orchestrate FP8 LHS quantization. Finally, this change also updates the underlying `gmm_v2` kernel fork to accept static LHS scale.

BUGS: 538171764, 540900176
FIXES: 538171764

# Tests

I added unit tests in `tests/unit/quantizations_test.py`.

### E2E Test

Loss is similar and perf is slightly better between after (fwd lhs quant internal static) vs. before (fwd lhs quant internal dynamic), more details: b/538171764#comment2

# Checklist

### DIFF
--- a/src/maxtext/kernels/megablox/ops.py
+++ b/src/maxtext/kernels/megablox/ops.py
@@ -330,7 +330,7 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
 def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule | None) -> jax.Array | None:
   """Extracts the static LHS (activation) scale for the GMM v2 forward pass.
 
-  GMM v2 only supports lhs_scale from symmetric fixed range calibration 
+  GMM v2 only supports lhs_scale from symmetric fixed range calibration
   (Or if lhs_scale is None, calculate dynamic scale internally).
 
   Enforces a default (1, 1) shape for per-tensor quantization kernels.
@@ -365,10 +365,9 @@ def _fwd_run_tokamax_v2(
     group_offset: jnp.ndarray | None,
     partial_sum: jnp.ndarray | None,
     transpose_rhs: bool,
-    quantization_rule: qwix.QtRule,
+    quantization_rule: qwix.QtRule | None = None,
 ) -> jnp.ndarray:
   """Executes the Tokamax GMM V2 backend for forward pass OUT = LHS @ RHS."""
-  # if transpose_rhs=False, rhs is [g, k, n], remain unchanged
   # if transpose_rhs=True, rhs [g, n, k], explicit transpose to [g, k, n]
   rhs_operand = rhs if not transpose_rhs else rhs.swapaxes(1, 2)
   rhs_scale = None
--- a/src/maxtext/kernels/megablox/ops.py
+++ b/src/maxtext/kernels/megablox/ops.py
@@ -368,6 +368,7 @@ def _fwd_run_tokamax_v2(
     quantization_rule: qwix.QtRule | None = None,
 ) -> jnp.ndarray:
   """Executes the Tokamax GMM V2 backend for forward pass OUT = LHS @ RHS."""
+  # if transpose_rhs=False, rhs is [g, k, n], remain unchanged
   # if transpose_rhs=True, rhs [g, n, k], explicit transpose to [g, k, n]
   rhs_operand = rhs if not transpose_rhs else rhs.swapaxes(1, 2)
   rhs_scale = None
--- a/src/maxtext/kernels/megablox/ops.py
+++ b/src/maxtext/kernels/megablox/ops.py
@@ -26,7 +26,6 @@ from maxtext.kernels.megablox import pallas_mosaic_tpu_v2_gmm_kernel as gmm_v2
 from maxtext.kernels.megablox import pallas_mosaic_tpu_v2_tgmm_kernel as tgmm_v2
 from maxtext.layers import quantizations
 import qwix
-from qwix._src.core import numerics
 import qwix.pallas as qpl
 import tokamax
 
@@ -331,7 +330,6 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
 def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule) -> jax.Array | None:
   """Extracts the static LHS (activation) scale for the GMM V2 forward pass.
 
-  Based on qwix's calibrate and compute_scale_zero_point functions.
   Enforces a default (1, 1) shape for per-tensor quantization kernels.
 
   Args:
@@ -347,22 +345,7 @@ def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule) -> jax.Array | None:
   if method is None or qtype is None or not method.lower().startswith("fixed"):
     return None
 
-  args = [float(a) for a in method.split(",")[1:]]
-  if len(args) not in (1, 2):
-    raise ValueError("Fixed range must be 'fixed,bound' or 'fixed,min,max'.")
-
-  if len(args) == 1:
-    args = [-args[0], args[0]]
-
-  if args[0] > 0 or args[1] < 0 or args[0] >= args[1]:
-    raise ValueError(f"The range must contain 0 and be non-empty, got: {method}")
-
-  if args[0] + args[1] == 0:
-    qmax = float(numerics.get_symmetric_bound(qtype))
-    scale_val = args[1] / qmax
-  else:
-    qmin, qmax = numerics.get_asymmetric_bound(qtype)
-    scale_val = (args[1] - args[0]) / float(qmax - qmin)
+  scale_val = quantizations.get_static_scale(qtype, method)
 
   return jnp.full((1, 1), scale_val, jnp.float32)
 
--- a/src/maxtext/kernels/megablox/ops.py
+++ b/src/maxtext/kernels/megablox/ops.py
@@ -200,7 +200,15 @@ def _gmm_fwd(
     out = _fwd_run_tokamax_v1(lhs, rhs, group_sizes, preferred_element_type, transpose_rhs, use_manual_quantization)
   elif use_tokamax_backend and use_gmm_v2:
     out = _fwd_run_tokamax_v2(
-        lhs, rhs, group_sizes, preferred_element_type, tiling, group_offset, partial_sum, transpose_rhs
+        lhs,
+        rhs,
+        group_sizes,
+        preferred_element_type,
+        tiling,
+        group_offset,
+        partial_sum,
+        transpose_rhs,
+        quantization_rule,
     )
   else:
     out = _fwd_run_megablox(
@@ -319,6 +327,36 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
   return jnp.broadcast_to(rhs_scale, (G, num_quant_blocks, 1, N))
 
 
+def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule | None) -> jax.Array | None:
+  """Extracts the static LHS (activation) scale for the GMM v2 forward pass.
+
+  If a static scale is used, GMM v2 requires it to be from a symmetric fixed-range
+  calibration (e.g., 'fixed,-max,max' or 'fixed,max'). If no static scale is
+  provided, the kernel will compute a dynamic scale on the fly.
+
+  Enforces a default (1, 1) shape for per-tensor quantization kernels.
+
+  Args:
+    quantization_rule: The Qwix quantization rule from which to extract the scale.
+
+  Returns:
+    The extracted static scale array, or None if not using purely fixed calibration.
+  """
+  if quantization_rule is None:
+    return None
+
+  method = quantization_rule.act_calibration_method
+  qtype = quantization_rule.act_qtype
+
+  # Use dynamic quantization, gmm_v2 calculates dynamic scale internally
+  if method is None or qtype is None or not method.lower().startswith("fixed"):
+    return None
+
+  scale_val = quantizations.get_static_scale(qtype, method)
+
+  return jnp.full((1, 1), scale_val, jnp.float32)
+
+
 def _fwd_run_tokamax_v2(
     lhs: jnp.ndarray | qpl.QArray,
     rhs: jnp.ndarray | qpl.QArray,
@@ -328,6 +366,7 @@ def _fwd_run_tokamax_v2(
     group_offset: jnp.ndarray | None,
     partial_sum: jnp.ndarray | None,
     transpose_rhs: bool,
+    quantization_rule: qwix.QtRule | None = None,
 ) -> jnp.ndarray:
   """Executes the Tokamax GMM V2 backend for forward pass OUT = LHS @ RHS."""
   # if transpose_rhs=False, rhs is [g, k, n], remain unchanged
@@ -354,6 +393,7 @@ def _fwd_run_tokamax_v2(
       preferred_element_type=preferred_element_type,
       partial_sum=partial_sum,
       group_offset=group_offset,
+      lhs_scale=_fwd_prepare_lhs_scale(quantization_rule),
   )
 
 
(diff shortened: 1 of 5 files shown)


src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 # Forked from:
-# https://github.com/openxla/tokamax/blob/3f332fcf85dcb87aab661d00228ed71a09b5fd56/tokamax/_src/ops/ragged_dot/pallas_mosaic_tpu_v2_gmm_kernel.py
+# https://github.com/openxla/tokamax/blob/a1105e7513c4cc8604bad5627d099dcf09430ca1/tokamax/_src/ops/ragged_dot/pallas_mosaic_tpu_v2_gmm_kernel.py
```
suhithr, line 16: To clarify the PR description, are there any divergent changes to this file in this PR?

CaptainO5: Good point, while there are no new divergent changes introduced in this pr, there is a divergence related to the partial_sum feature added earlier.

src/maxtext/kernels/megablox/ops.py
```diff
@@ -319,6 +328,45 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
...
+
+  Based on qwix's calibrate and compute_scale_zero_point functions.
+  Enforces a default (1, 1) shape for per-tensor quantization kernels.
+
+  Args:
+    quantization_rule: The Qwix quantization rule from which to extract the scale.
+
+  Returns:
+    The extracted static scale array, or None if not using purely fixed calibration.
+  """
```
shuningjin: This function seems a general processing of fixed calibration, not limited to gmm or lhs. Perhaps we can move this to `quantization.py`: similar to
https://github.com/AI-Hypercomputer/maxtext/blob/dca6781da9fac5e3b5331580d3d316e4db50a95d/src/maxtext/layers/quantizations.py#L915

```
def get_scale_for_fixed_calibration(dtype: jax.typing.DTypeLike, calibration_method: str):
```

---

> Based on qwix's calibrate and compute_scale_zero_point functions.

does it use `compute_scale_zero_point`?

tests/unit/moe_test.py
```diff
@@ -2054,5 +2055,46 @@ def test_prefuse_moe_weights_matches_unfused(self):
     )
 
 
+class LhsScaleTest(unittest.TestCase):
+  """Tests for LHS scale extraction in GMM forward."""
```
shuningjin: Thanks for adding the test! Maybe can move to `quantization_test`?

https://github.com/AI-Hypercomputer/maxtext/blob/5825af54d7d6840f097dc530712fd951e61ce811/tests/unit/quantizations_test.py#L15

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
shuningjin: Thanks for the carefully merging tokamax changes with previous maxtext changes! I have double checked. This aligns with my attempt, [diff](https://diff.googleplex.com/#key=SCNxxCkguR8V).

shuningjin (commented):
Thanks for adding the support! Overall looks good.

- Also thanks for the carefully merging new tokamax changes with previous maxtext changes! I have double checked. This aligns with my attempt, [diff](https://diff.googleplex.com/#key=SCNxxCkguR8V).
- In the PR description, can add E2E test result, e.g., loss is similar and perf is slightly better between after (fwd lhs quant internal static) vs. before (fwd lhs quant internal dynamic), with a bug link for future reference.
- Maybe the function and test can be moved to `quantization.py` and `quantization_test.py`.

src/maxtext/kernels/megablox/ops.py
CaptainO5: Thanks for the pointer, I will migrate it to quantizations.py

src/maxtext/layers/quantizations.py
```diff
@@ -912,6 +913,46 @@ def _get_max_min(target_dtype):
...
+
+  if args[0] > 0 or args[1] < 0 or args[0] >= args[1]:
+    raise ValueError(f"The range must contain 0 and be non-empty, got: {calibration_method}")
+
+  if args[0] + args[1] == 0:
+    qmax = float(numerics.get_symmetric_bound(qtype))
+    scale_val = args[1] / qmax
+  else:
+    qmin, qmax = numerics.get_asymmetric_bound(qtype)
+    scale_val = (args[1] - args[0]) / float(qmax - qmin)
```
shuningjin: Seems gmm2 only supports in-kernel lhs scale with symmetric bound.

`pallas_mosaic_tpu_v2_gmm_kernel.py` performs symmetric scale-and-clip: `block_lhs_q = jnp.clip(block_lhs * lhs_scale_inv, -dtype_max, dtype_max).astype(lhs_q_dtype)`. [here](https://screenshot-v2.corp.google.com/2trrgtl1tbbi0)

The kernel does not implement zero-point shifts. If an asymmetric range is supplied, the result can be mathematically incorrect.

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
shuningjin: Should use
```
 flat_args_preceding = (group_sizes, group_offset, lhs_in, rhs_weights)
```
otherwise can be incorrect when `partial_sum is not None and lhs_scale is not None`

CaptainO5: Great catch, thanks for the headsup

tests/unit/quantizations_test.py
```diff
@@ -716,5 +718,61 @@ def test_nnx_abstract_state_has_no_intermediates(self):
...
+      quantizations.get_static_scale(jnp.float8_e4m3fn, "fixed,1,2,3")
+
+    with self.assertRaises(ValueError):
+      quantizations.get_static_scale(jnp.float8_e4m3fn, "fixed,-200.0,224.0")
+
+    with self.assertRaises(ValueError):
+      quantizations.get_static_scale(jnp.float8_e4m3fn, "absmax")
+
+    with self.assertRaises(ValueError):
+      quantizations.get_static_scale(jnp.float8_e4m3fn, "absmax")
```
shuningjin, line 740: nit: duplicate

src/maxtext/kernels/megablox/ops.py
```diff
@@ -328,6 +359,7 @@ def _fwd_run_tokamax_v2(
     group_offset: jnp.ndarray | None,
     partial_sum: jnp.ndarray | None,
     transpose_rhs: bool,
+    quantization_rule: qwix.QtRule,
```
shuningjin: nit: `qwix_rule: qwix.QtRule | None = None`

src/maxtext/layers/quantizations.py
```diff
@@ -912,6 +912,39 @@ def _get_max_min(target_dtype):
...
+    raise ValueError(f"Only static scale quantization is supported, got {calibration_method}")
+
+  args = [float(a) for a in calibration_method.split(",")[1:]]
+  if len(args) == 1:
+    args = [-args[0], args[0]]
+
+  if len(args) != 2 or args[0] + args[1] != 0 or args[1] <= 0:
+    raise ValueError(f"Expected format: 'fixed,max' or 'fixed,-max,max'. Got: {calibration_method}")
+
+  qmax, _ = _get_max_min(qtype)
```
shuningjin: might be more accurate to use `qwix._src.core.numerics.get_symmetric_bound` (as in your previous commit). could remove `_get_max_min` function

src/maxtext/kernels/megablox/ops.py
```diff
@@ -319,6 +327,29 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
...
+  Enforces a default (1, 1) shape for per-tensor quantization kernels.
+
+  Args:
+    quantization_rule: The Qwix quantization rule from which to extract the scale.
+
+  Returns:
+    The extracted static scale array, or None if not using purely fixed calibration.
+  """
+  method = getattr(quantization_rule, "act_calibration_method", None)
+  qtype = getattr(quantization_rule, "act_qtype", None)
```
shuningjin: nit
```
if quantization_rule is None:
  return None
method = quantization_rule.act_calibration_method
qtype = quantization_rule.act_qtype
```

src/maxtext/kernels/megablox/ops.py
```diff
@@ -319,6 +327,29 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
   return jnp.broadcast_to(rhs_scale, (G, num_quant_blocks, 1, N))
 
 
+def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule) -> jax.Array | None:
+  """Extracts the static LHS (activation) scale for the GMM V2 forward pass.
+
+  Enforces a default (1, 1) shape for per-tensor quantization kernels.
```
shuningjin, line 337: nit: add note
```
gmm_v2 only supports lhs_scale from symmetric fixed range calibration (Or if lhs_scale is None, calculate dynamic scale internally).
```

src/maxtext/layers/quantizations.py
CaptainO5: This makes sense, but `qwix._src.core.numerics.get_symmetric_bound` only returns the qmax value while`quantizations.manual_quantize` requires both max and min values. I decided to use the `_get_max_min` in `get_static_scale` for consistency.

CaptainO5: Thanks, we could use -qmax for qmin and remove `_get_max_min`.


Resulting in the following bounds:
DType              | Min                  | Max                 
-------------------|----------------------|-----------------------
jnp.int4           | -7.5                   | 7.5                  
jnp.int8           | -127.5                | 127.5                 
-|-|-
jnp.float8_e4m3fn  | -448.0               | 448.0                 
jnp.float8_e5m2    | -57344.0             | 57344.0               
jnp.float16        | -65504.0             | 65504.0            
jnp.bfloat16       | -3.38953e+38         | 3.38953e+38         
jnp.float32        | -3.4028235e+38       | 3.4028235e+38

shuningjin (approved):
Great work! LGTM.

src/maxtext/kernels/megablox/ops.py
```diff
@@ -322,6 +327,35 @@ def _fwd_prepare_rhs_scale(rhs: qpl.QArray, transpose_rhs: bool = False) -> jnp.
   return jnp.broadcast_to(rhs_scale, (G, num_quant_blocks, 1, N))
 
 
+def _fwd_prepare_lhs_scale(quantization_rule: qwix.QtRule | None) -> jax.Array | None:
+  """Extracts the static LHS (activation) scale for the GMM v2 forward pass.
+
+  GMM v2 only supports lhs_scale from symmetric fixed range calibration
```
RissyRan: Nit: Why we have "only"? It see you also mentioned if we set as None, it will support dynamic scaling.

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
```diff
@@ -141,6 +141,29 @@ def get_bias(self) -> jax.Array:
...
 
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass(frozen=True)
+class LhsRef:
+  """Dataclass for the lhs value and its optional quantization scale.
+
+  Unlike `rhs`, the lhs is passed to the kernel *unquantized*. When
+  `scale` is provided, the kernel uses it to quantize the lhs (i.e.
+  `qvalue = clip(lhs / scale)` and the result is multiplied back by `scale`).
+  The scale's shape encodes the granularity (per-tensor `[1, 1]`; extensible to
```
RissyRan, line 152: Does this mean we could support sub-channel when passing [M, num_blocks]?

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
```diff
@@ -173,8 +196,21 @@ class InputConfigs:
...
   dtype: jnp.dtype
   has_bias: bool = False
+  # Whether a scale array accompanies this input. The *direction* is inferred
+  # from the dtype relationship: when the input already arrives quantized
+  # (dtype == quant_dtype) the scale dequantizes it (rhs); when it arrives
+  # unquantized (dtype != quant_dtype) the scale quantizes it online (lhs).
   has_scale: bool = False
 
+  @property
+  def should_use_external_scale(self) -> bool:
```
RissyRan, line 206: Wondering for perf reason, should we keep quantization inside of gmm v2 kernel? So that user doesn't need to handle static scaling before the call.

RissyRan (approved):
LGTM, and have a few comments. I think we could follow up static scaling inside of kernel to perf reason. Happy to chat more!

src/maxtext/kernels/megablox/ops.py
CaptainO5: GMM v2 does not support asymmetric fixed range calibration, i.e., when min != -max in `fixed,min,max`. I will rewrite the comment to make it clearer.

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
CaptainO5: Yes

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
```diff
@@ -173,8 +196,21 @@ class InputConfigs:
   quant_block_size: int | None
   dtype: jnp.dtype
   has_bias: bool = False
+  # Whether a scale array accompanies this input. The *direction* is inferred
+  # from the dtype relationship: when the input already arrives quantized
+  # (dtype == quant_dtype) the scale dequantizes it (rhs); when it arrives
+  # unquantized (dtype != quant_dtype) the scale quantizes it online (lhs).
```
suhithr, line 202: Thanks for the clarifying comment

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
```diff
@@ -466,15 +525,21 @@ def _matmul(is_first_k_step: bool, is_last_k_step: bool):
...
-          block_abs_max = jnp.max(jnp.abs(block_lhs), axis=1, keepdims=True)
-          block_scale = block_abs_max / dtype_max
-
-          # If block_scale=0, it will cause division by zero and return either
-          # NaN or Inf. Since this can cause numeric issue when downcasting to
-          # quantized value, we convert them into 0.
-          block_scale_inv = jnp.where(block_scale == 0, 0, 1 / block_scale)
-          # Convert lhs into quantized dtype.
-          block_lhs_q = (block_lhs * block_scale_inv).astype(lhs_q_dtype)
+          if should_use_external_scale:
```
suhithr, line 528: The above comment indicates that the previous version of the code depends upon a compiler optimization to avoid repeated re-computation of the `block_scale`.
Given your performance test results it seems we are getting the same CSE and avoiding recomputation with this branch?

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
suhithr: In `validate_inputs()` we explicitly check for a `[1, 1]` shape now, but this design is extensible to sub-channel quantization, correct?

suhithr (approved):
LGTM, thank you!

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
CaptainO5: True, currently only per-tensor is supported

src/maxtext/kernels/megablox/pallas_mosaic_tpu_v2_gmm_kernel.py
CaptainO5: I believe so, yes

PR #4742, opened 2026-08-05
Honor partial_rotary_factor on the MRoPE path

### LINKED ISSUE #4616: partial_rotary_factor is ignored when use_mrope is true (affects qwen3.5-35b-a3b and qwen3.5-397b-a17b)
### Summary

`partial_rotary_factor` is silently ignored whenever `use_mrope: true`. Two shipped configs set both, so RoPE is applied to the full head dimension instead of the leading quarter, with no error and no shape mismatch.

Affected:

- `src/maxtext/configs/models/qwen3.5-35b-a3b.yml` — `use_mrope: true`, `partial_rotary_factor: 0.25`
- `src/maxtext/configs/models/qwen3.5-397b-a17b.yml` — `use_mrope: true`, `partial_rotary_factor: 0.25`

### Cause

`Attention.init_rotary_embedding` routes to `Qwen3OmniMoeThinkerTextRotaryEmbedding` whenever `use_mrope` is set (`src/maxtext/layers/attentions.py`, the `elif self.use_mrope:` branch). It passes `embedding_dims=rope_embedding_dims`, which is the full `head_dim`, and never passes the partial factor — the class does not accept one.

`PartialRotaryEmbedding` implements the intended behaviour, but only the non-MRoPE path can reach it.

### Reproduction

On `main`, passing the factor fails outright:

```python
Qwen3OmniMoeThinkerTextRotaryEmbedding(
    min_timescale=1, max_timescale=10000, embedding_dims=16,
    cast_as_fprop_dtype=False, mrope_section=(2, 1, 1),
    partial_rotary_factor=0.25, rngs=nnx.Rngs(params=0),
)
# TypeError: __init__() got an unexpected keyword argument 'partial_rotary_factor'
```

and the default construction rotates every channel:

```python
layer = Qwen3OmniMoeThinkerTextRotaryEmbedding(
    min_timescale=1, max_timescale=10000, embedding_dims=16,
    cast_as_fprop_dtype=False, mrope_section=(2, 1, 1), rngs=nnx.Rngs(params=0),
)
inputs = jax.random.normal(jax.random.PRNGKey(0), (2, 8, 4, 16), dtype=jnp.float32)
positions = jnp.broadcast_to(jnp.arange(8, dtype=jnp.int32), (2, 8))
out = layer(inputs, positions)

np.allclose(out[..., 4:], inputs[..., 4:])
# -> False
# With partial_rotary_factor 0.25 only the first 4 of 16 channels may be rotated;
# the remaining 12 must pass through untouched.
```

### Suggested fix

Give the MRoPE layer the same construction `PartialRotaryEmbedding` already uses — split off the leading `rotary_dim = int(head_dim * partial_rotary_factor)` channels, build `inv_freq` over `rotary_dim` rather than `head_dim`, and concatenate the untouched remainder back:

```python
self.head_dim = embedding_dims
self.partial_rotary_factor = partial_rotary_factor
self.rotary_dim = int(self.head_dim * self.partial_rotary_factor)
super().__init__(..., embedding_dims=self.rotary_dim, ...)
```

```python
if self.rotary_dim < self.head_dim:
    inputs_rot, inputs_pass = jnp.split(inputs, [self.rotary_dim], axis=-1)
else:
    inputs_rot, inputs_pass = inputs, None
...
x_out = self.apply_rotary(inputs_rot, cos_emb, sin_emb)
if inputs_pass is not None:
    x_out = jnp.concatenate([x_out, inputs_pass], axis=-1)
```

plus one line in `init_rotary_embedding` to pass `config.partial_rotary_factor` through. The shape check should compare against `head_dim` rather than the (now reduced) `embedding_dims`.

A default of `1.0` keeps `qwen3-vl-*` and `qwen3-omni-*` bit-identical — those configs do not set the factor.

### Verification

With that change, four properties hold, checked against `PartialRotaryEmbedding` and against the unmodified layer:

1. channels beyond `rotary_dim` are passed through unchanged, leading ones are rotated;
2. a partial factor no longer produces the fully rotated result (the regression itself);
3. omitting the factor reproduces the previous full-rotation output exactly;
4. for text-only 1D positions — where MRoPE degenerates to ordinary RoPE — the MRoPE path agrees with `PartialRotaryEmbedding` channel for channel (`rtol=1e-5`).

Properties 2 and 4 fail on `main`.

adityasingh2400:
# Description

`partial_rotary_factor` was silently dropped on the MRoPE path, so the two shipped configs that set both `use_mrope: true` and `partial_rotary_factor: 0.25` applied RoPE to the full head dimension instead of the leading quarter. Reported in #4616.

The defect: `Attention.init_rotary_embedding` in `src/maxtext/layers/attentions.py` routes every `use_mrope` layer to `Qwen3OmniMoeThinkerTextRotaryEmbedding` through the `elif self.use_mrope:` branch. That branch passed `embedding_dims=rope_embedding_dims`, which is the full `head_dim`, and never passed a partial rotary factor because the class did not accept one. `PartialRotaryEmbedding` already implements the intended behavior, but only the non-MRoPE branches can reach it, and the `is_qwen3_hybrid` branch that does construct it sits below the MRoPE branch in the same chain. There was no error and no shape mismatch, so the misconfiguration was invisible at runtime.

Affected shipped configs, both `head_dim: 256`:

- `src/maxtext/configs/models/qwen3.5-35b-a3b.yml`
- `src/maxtext/configs/models/qwen3.5-397b-a17b.yml`

Both set `partial_rotary_factor: 0.25`, so 64 of the 256 head channels should be rotated. Their `mrope_section` of `[11, 11, 10]` sums to 32, which is half of that intended 64 channel slice rather than half of `head_dim`, so the frequency layout was wrong too: the interleaved MRoPE sections only reached the first 33 of the 128 angles that the full head dimension produces.

The fix teaches `Qwen3OmniMoeThinkerTextRotaryEmbedding` the construction `PartialRotaryEmbedding` already uses. It takes a `partial_rotary_factor`, computes `rotary_dim = int(head_dim * partial_rotary_factor)`, and initializes the base class with `embedding_dims=self.rotary_dim` so the timescale, and therefore the inverse frequencies, span only the rotated slice. In `__call__` the leading `rotary_dim` channels are split off, rotated, and the untouched remainder is concatenated back. The shape check now compares against `head_dim` instead of the reduced `embedding_dims`. `init_rotary_embedding` passes `config.partial_rotary_factor` through, matching what the neighboring qwen3 hybrid branch already does.

The factor defaults to 1.0. In that case no split happens and the output is unchanged, so `qwen3-vl-*` and `qwen3-omni-*`, which do not set the factor, are unaffected. I verified this directly: dumping the default layer output at `head_dim=128` with `mrope_section=(24, 20, 20)`, for both 3D multimodal positions and 2D text positions, gives bitwise identical arrays before and after the change.

# Tests

Five new cases in `tests/unit/partial_rotary_embedding_test.py`, in a `Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest` class alongside the existing partial rotary tests:

- channels beyond `rotary_dim` pass through untouched and the leading slice is rotated,
- with text-only positions MRoPE degenerates to ordinary RoPE, so the output must agree with `PartialRotaryEmbedding` channel for channel,
- a snapshot of the default full rotation output, which pins the existing behavior,
- an explicit factor of 1.0 matches the default construction,
- the two affected shipped configs are loaded through `pyconfig`, their `mrope_section` is checked against `rotary_dim // 2`, and the passthrough property is asserted at their real `head_dim` of 256.

Command:

```
python -m pytest tests/unit/partial_rotary_embedding_test.py -q
```

Fail before, with `src/maxtext/layers/embeddings.py` and `src/maxtext/layers/attentions.py` checked out from the base ref and only the test file carrying the change:

```
E     TypeError: Qwen3OmniMoeThinkerTextRotaryEmbedding.__init__() got an unexpected keyword argument 'partial_rotary_factor'

=========================== short test summary info ============================
FAILED tests/unit/partial_rotary_embedding_test.py::Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest::test_mrope_full_factor_matches_default
FAILED tests/unit/partial_rotary_embedding_test.py::Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest::test_mrope_partial_matches_partial_rotary_embedding
FAILED tests/unit/partial_rotary_embedding_test.py::Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest::test_mrope_partial_rotary_passthrough
SUBFAILED(model_name='qwen3.5-35b-a3b') tests/unit/partial_rotary_embedding_test.py::Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest::test_shipped_mrope_configs_apply_partial_rotary
SUBFAILED(model_name='qwen3.5-397b-a17b') tests/unit/partial_rotary_embedding_test.py::Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest::test_shipped_mrope_configs_apply_partial_rotary
5 failed, 10 passed in 8.98s
```

Pass after, with both source files restored:

```
.............                                                          [100%]
13 passed, 2 subtests passed in 9.37s
```

The pre-existing cases in that file, including the `PartialRotaryEmbedding` and `Gemma4PartialRotaryEmbedding` suites, pass in both runs. `tests/unit/embeddings_test.py` is also green. Ran on CPU. Formatted with the pinned pyink 24.10.1 at `--pyink-indentation=2 --line-length=122`, and pylint 3.3.8 rates the changed files 10.00/10.

FIXES: #4616

### DIFF
--- a/tests/unit/partial_rotary_embedding_test.py
+++ b/tests/unit/partial_rotary_embedding_test.py
@@ -30,7 +30,12 @@ from jax.sharding import Mesh
 from flax import nnx
 import numpy as np
 
-from maxtext.layers.embeddings import PartialRotaryEmbedding, RotaryEmbedding, Gemma4PartialRotaryEmbedding
+from maxtext.layers.embeddings import (
+    PartialRotaryEmbedding,
+    RotaryEmbedding,
+    Gemma4PartialRotaryEmbedding,
+    Qwen3OmniMoeThinkerTextRotaryEmbedding,
+)
 from maxtext.configs import pyconfig
 from maxtext.utils import maxtext_utils
 from tests.utils.test_helpers import get_test_config_path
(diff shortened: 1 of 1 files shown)


adityasingh2400:
The source fix here is superseded by #4764, which merged earlier today. I have rebased and dropped it, so this is now tests only.

Your version is the better one. It adds an instance-level override in `attentions.py`:

```python
partial_rotary_factor=(
    self.partial_rotary_factor if self.partial_rotary_factor is not None
    else self.config.partial_rotary_factor
),
```

plus validation of the factor and the rotary dim in `MRoPE.__init__`. Mine had neither, so there is nothing left worth keeping from my source changes.

What is left is the MRoPE regression coverage, which `main` does not have. The existing `tests/unit/partial_rotary_embedding_test.py` covers `PartialRotaryEmbeddingTest` and `Gemma4PartialRotaryEmbeddingTest`, but nothing exercises the Qwen3-Omni MRoPE path that #4616 reported. This adds `Qwen3OmniMoeThinkerTextRotaryEmbeddingPartialTest` with five cases:

- a partial factor leaves the unrotated tail untouched
- MRoPE partial output matches `PartialRotaryEmbedding` for the same factor
- the default rotates the full head dim
- an explicit factor of 1.0 matches the default
- every shipped MRoPE config actually applies its partial factor, checking `rotary_dim` and `sum(mrope_section) == rotary_dim // 2`

Worth saying plainly: I could not run these locally. They were written against my implementation and the repo needs `pathwaysutils` and the rest of the config chain to import, which I do not have here. They read as compatible with yours, no assertions on error strings and the default is `1.0` either way, but CI is the real check. If any of them are wrong against your version I will fix them.

Close this if the coverage is not worth carrying. The bug itself is fixed either way, and #4616 is still open if you want to close that too.

adityasingh2400:
Still relevant, so flagging this rather than letting it close.

The source fix this started as was superseded by #4764, which merged on 2026-08-07. What is left here is only the regression coverage for that fix, in `tests/unit/partial_rotary_embedding_test.py`. No source file is touched.

I re-checked it against current `main` just now. All four imports still resolve, `PartialRotaryEmbedding`, `RotaryEmbedding`, `Gemma4PartialRotaryEmbedding` and `Qwen3OmniMoeThinkerTextRotaryEmbedding` are all still in `src/maxtext/layers/embeddings.py`, and there are no failing checks.

Happy to close it if the extra coverage is not wanted. Otherwise it is ready whenever someone has a moment.

PR #4737, opened 2026-08-05
[maxtext] Add block-diffusion training primitives

ethannnnnn:
## Motivation

MaxText needs small, model-independent primitives for discrete block-diffusion
training before attention, input-pipeline, trainer, or post-training support is
introduced.

## Scope

This PR adds only:

- block corruption under `maxtext.diffusion.block_diffusion`;
- explicit target alignment for `same_position` and `shifted` logits; and
- focused unit tests for both supported contracts.

It does not change configuration, attention, model execution, training loops,
serving, SFT, distillation, or RL.

## Design

The corruption API consumes clean tokens, an explicit eligibility mask, a
logical block size, and one of two declared model contracts:

- `same_position` with an `all_masked` canvas; or
- `shifted` with a `seed_and_mask` canvas.

Each nonempty eligible block is guaranteed to retain at least one supervised
position. Padding remains untouched. Target alignment follows logical token
positions, including reordered physical layouts, instead of assuming that
array index equals sequence position.

## Compatibility

There is no behavior change for existing MaxText training or inference paths.
The new package is not activated by any configuration in this PR.

## Tests

`python -m pytest -q tests/unit/block_diffusion_test.py`

The tests cover partial blocks, per-block noise sampling, deterministic RNG,
all-invalid rows, seeded anchors, mask/loss provenance, reordered logical
positions, padding, and invalid contracts and shapes.

Current clean-head result: `13 passed`, `5 subtests passed`; Pyink, compileall,
and `git diff --check` also pass.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have added comments or docstrings where the contracts are not self-evident.
- [x] End-to-end testing is not applicable to this default-off, pure primitive layer; focused CPU tests are listed above.
- [x] No standalone documentation page or toctree change is needed; the public design document is linked below.

## Follow-ups

Block-causal attention and pre-training integration will be proposed only after
this primitive contract is reviewed and landed. CFT/SFT, OPD, exact replay,
and RL remain separate later changes.

Design document:
https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit

PR #4743, opened 2026-08-05
[maxtext] Add block-causal attention

ethannnnnn:
## Motivation

Block diffusion needs attention that is causal across blocks while remaining
bidirectional within the block currently being corrupted or denoised. This is
an attention-kernel concern and can be reviewed independently of corruption,
training objectives, rollout, and RL.

## Scope

This PR adds only:

- an opt-in `block_diffusion` attention type and positive
  `causal_block_size` configuration;
- equivalent dense dot-product and TPU Tokamax Splash block-causal masks;
- load-balanced context-parallel mask support;
- validation that rejects packing and unsupported attention/hardware
  combinations; and
- focused mask, configuration, and autoregressive-regression tests.

It does not depend on the block-diffusion primitive PR, corrupt training data,
change a loss function, add a model-specific branch, or enable diffusion by
default.

## Mask semantics

For query position `q` and key position `k`, attention is allowed when

```text
floor(q / block_size) >= floor(k / block_size)
```

Therefore, all positions in the current block can attend to one another and to
all earlier blocks, but no earlier block can attend to a later block. Segment
boundaries remain enforced. Packing is rejected because document-local block
origins are not yet represented by the attention metadata. Autoregressive
decode keeps the existing causal mask path unchanged.

## Compatibility

The default remains `attention_type=global`. The new behavior is activated only
when `attention_type=block_diffusion` is selected explicitly. Specialized
per-layer attention declarations continue to override the global configuration.

This implementation remains code-independent from the primitives landed in
#4737: it does not import or activate them. The branch is rebased onto the
current `main` containing #4737 so reviewers see only this attention change.

## Tests

```text
python -m pytest -q \
  tests/unit/attention_test.py \
  tests/unit/configs_value_test.py \
  tests/unit/tokamax_splash_attention_mask_test.py

76 passed, 48 skipped, 36 subtests passed
```

The skipped cases require unavailable platform-specific kernels. Pyink, Python
compilation, and `git diff --check` also pass for the changed files.

Focused branch coverage for the changed executable code in
`src/maxtext/layers/attention_op.py` is complete: 53/53 changed lines and
26/26 changed branches are covered.

## Checklist

- [x] I added comments or docstrings for the non-obvious block-mask invariants.
- [x] I added focused dense, Splash, load-balanced, configuration, and autoregressive-regression tests.
- [x] End-to-end model training is not applicable to this default-off attention layer; the affected CPU suites are listed above.
- [x] No standalone documentation page or toctree change is needed; the public design document is linked below.

Design document:
https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -388,12 +388,15 @@ param_scan_axis: 1
 # The attention parameter dictates the specific algorithm/methodology used to compute the attention scores
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
-attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla, full, compressed, block_diffusion
 share_kv_projections: false # Note: Not compatible with attention_type='mla'
 attention_bias: false # If true, adds a learnable bias to the query, key, and value projections
 attention_sink: false
 sliding_window_size: 0
 chunk_attn_window_size: 0
+# Token block size B for block-causal attention in Block Diffusion (arXiv:2503.09573).
+# Tokens attend bidirectionally within a block and causally across blocks.
+causal_block_size: 32
 attn_logits_soft_cap: 0.0
 final_logits_soft_cap: 0.0
 z_loss_multiplier: 0.0
@@ -1348,5 +1351,4 @@ elastic_backup_kind: "snapshot"
 elastic_timeout_seconds: 300
 elastic_max_retries: 10
 elastic_min_slice_count: -1
-
 
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -579,7 +579,7 @@ class Attention(BaseModel):
       "autoselected",
       description="The attention algorithm to use (dot_product, flash, cudnn_flash_te, vllm_rpa, vllm_batched_rpa, etc).",
   )
-  attention_type: Literal["global", "local_sliding", "chunk", "mla", "full", "compressed"] = Field(
+  attention_type: Literal["global", "local_sliding", "chunk", "mla", "full", "compressed", "block_diffusion"] = Field(
       "global", description="The variant of attention to use."
   )
   share_kv_projections: bool = Field(
@@ -626,6 +626,10 @@ class Attention(BaseModel):
   )
   sliding_window_size: NonNegativeInt = Field(0, description="The size of the sliding window for local attention.")
   chunk_attn_window_size: NonNegativeInt = Field(0, description="The window size for chunked attention.")
+  causal_block_size: PositiveInt = Field(
+      32,
+      description="The number of token positions in each bidirectional block for block-causal attention.",
+  )
   attn_logits_soft_cap: None | NonNegativeFloat = Field(
       None, description="Soft-cap value for attention logits. None means no cap."
   )
@@ -3500,6 +3504,19 @@ class MaxTextConfig(
         not isinstance(self.sliding_window_size, int) or self.sliding_window_size <= 0
     ):
       raise ValueError("`sliding_window_size` must be an integer > 0 for 'local_sliding' attention.")
+    if self.attention_type == AttentionType.BLOCK_DIFFUSION.value:
+      if self.packing:
+        # Document-local block origins inside a packed sequence are not tracked
+        # in attention metadata; packing without realignment would cause
+        # cross-document block-attention leakage.
+        raise ValueError("Block-diffusion attention does not support packing; set `packing=False`.")
+      if self.attention not in ("autoselected", "dot_product", "flash"):
+        raise ValueError("Block-diffusion attention is supported only by dot_product attention and TPU Splash attention.")
+      if self.attention in ("autoselected", "flash") and self.hardware != "tpu":
+        raise ValueError(
+            "Block-diffusion attention with attention='autoselected' or attention='flash' requires hardware='tpu'; "
+            "use attention='dot_product' on other hardware."
+        )
     if self.quantize_kvcache and not self.kv_quant_axis:
       raise ValueError("`kv_quant_axis` cannot be empty when quantize_kvcache is True.")
     if (
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -84,6 +84,16 @@ from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_
 dynamic_vector_slice_in_dim = jax.vmap(lax.dynamic_slice_in_dim, in_axes=(None, 0, None, None))
 
 
+def _resolve_attention_type(config: Config, attention_type: AttentionType | str | None) -> AttentionType:
+  configured_attention_type = AttentionType(getattr(config, "attention_type", AttentionType.GLOBAL.value))
+  if attention_type is None:
+    return configured_attention_type
+  resolved_attention_type = AttentionType(attention_type)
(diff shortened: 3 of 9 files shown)


src/maxtext/layers/attention_op.py
```diff
@@ -234,6 +288,18 @@ def _generate_chunk_attention_mask(mask_shape: tuple[int, int], chunk_size: int,
   return chunk_mask
 
 
+def _generate_block_causal_attention_mask(
+    mask_shape: tuple[int, int], causal_block_size: int, q_offset: int = 0
+) -> jax.Array:
+  """Generates a block-causal mask, bidirectional within each block."""
```
shralex: Suggestion: 
  
"""Generates a block-causal mask for Block Diffusion Language Models (BD3LMs).
  Implements the block-causal attention pattern (M_BC) described in
  Arriola et al., "Block Diffusion: Interpolating Between Autoregressive and
  Diffusion Language Models" (https://arxiv.org/abs/2503.09573).
  For block size B = causal_block_size, query index q, and key index k:
    - Within block i (floor(q/B) == floor(k/B)): tokens attend bidirectionally.
    - Across blocks (floor(q/B) > floor(k/B)): block i attends causally to
      all preceding blocks 0, 1, ..., i-1, but not to future blocks.
  Note:
    - B = 1 reduces to standard autoregressive causal attention.
    - B = L (sequence length) reduces to full bidirectional attention.
  """

src/maxtext/configs/types.py
```diff
@@ -3484,6 +3488,16 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
         not isinstance(self.sliding_window_size, int) or self.sliding_window_size <= 0
     ):
       raise ValueError("`sliding_window_size` must be an integer > 0 for 'local_sliding' attention.")
+    if self.attention_type == AttentionType.BLOCK_DIFFUSION.value:
+      if self.packing:
```
shralex, line 3508: lets explain a bit better why its an error, like:

        # Document-local block origins inside a packed sequence are not yet
        # tracked in attention metadata; packing without realignment would cause
        # cross-document block-attention leakage.

src/maxtext/layers/attentions.py
```diff
@@ -120,7 +120,7 @@ def attention_as_linen(
     float32_logits: bool = False,  # cast logits in float32 for stability.
     quant: Optional[Quant] = None,
     kv_quant: Optional[KVQuant] = None,
-    attention_type: AttentionType = AttentionType.GLOBAL,  # Default to global attention
+    attention_type: AttentionType | None = None,
```
entrpn: why is it necessary to change this default value from GLOBAL to none?

ethannnnnn: Good catch. None became unnecessary after adding _resolve_attention_type. I restored the existing AttentionType.GLOBAL default in both constructors. Configured block diffusion still activates through the resolver.

src/maxtext/layers/attention_op.py
```diff
@@ -862,6 +935,19 @@ def generate_attention_mask(
         )
       output_mask = chunk_mask * output_mask
 
+    elif self.attention_type == AttentionType.BLOCK_DIFFUSION and model_mode != MODEL_MODE_AUTOREGRESSIVE:
```
shralex, line 1000: Lets clarify why, e.g.:

    # For standard token-by-token autoregressive decoding (MODEL_MODE_AUTOREGRESSIVE),
    # keep the existing causal mask path unchanged. BD3LM generation operates via
    # block-level parallel sampling rather than token-level autoregressive decode.

src/maxtext/configs/base.yml
```diff
@@ -387,12 +387,13 @@ param_scan_axis: 1
...
 # The attention_type parameter determines the variants of attention, e.g. global or local_sliding
 attention: 'autoselected' # Supported attention: autoselected, dot_product, flash, cudnn_flash_te
-attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla
+attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla, full, compressed, block_diffusion
 share_kv_projections: false # Note: Not compatible with attention_type='mla'
 attention_bias: false # If true, adds a learnable bias to the query, key, and value projections
 attention_sink: false
 sliding_window_size: 0
 chunk_attn_window_size: 0
+causal_block_size: 32 # Number of token positions in each bidirectional block for block-causal attention
```
shralex: Suggest the following comment instead, so users can find the paper without digging into attntions.py

# Token block size B for block-causal attention in Block Diffusion (arXiv:2503.09573). Tokens attend bidirectionally within a block and causally across blocks.

shralex (approved):
Thank  you for the contribution! Added a few documentation suggestions to make it easier to navigate for future users / maintainers.

src/maxtext/layers/attention_op.py
ethannnnnn: Done. The mask helper docstring now cites the Block Diffusion paper, defines within-block and across-block attention, and documents the `B = 1` causal and `B = L` bidirectional limits.

src/maxtext/configs/types.py
ethannnnnn: Done. The validation comment now explains that document-local block origins are not tracked in attention metadata and that packing without realignment could leak block attention across documents.

src/maxtext/layers/attention_op.py
ethannnnnn: Done. The comment now clarifies that standard autoregressive decoding preserves the causal mask path, while BD3LM generation uses block-level parallel sampling.

src/maxtext/configs/base.yml
ethannnnnn: Done. The `causal_block_size` configuration comment now links the Block Diffusion paper and summarizes within-block bidirectional and across-block causal attention.

entrpn (approved):

hengtaoguo (approved):
The github actions are temporarily down so the PR tests are queued. Let's watch out for this.

tests/unit/configs_value_test.py
```diff
@@ -299,6 +300,57 @@ def test_load_balanced_chunk_context_parallel_config(self):
...
+            _BASE_CONFIG_PATH,
+            "run_name=test",
+            "steps=1",
+            "attention=dot_product",
+            "attention_type=block_diffusion",
+            "causal_block_size=7",
+            "vocab_size=256",
+            "max_target_length=2048",
+            "hardware=cpu",
+            "packing=False",
```
hengtaoguo, line 316: I'm seeing an internal test failure related to jax distributed initialization. So allow me add a new flag here to assist the bug fix `skip_jax_distributed_system=True`.

cc @entrpn The error message is [here](https://fusion2.corp.google.com/presubmit/960733080/OCL:960733080:BASE:960733443:1786085015733:60c5d675/targets/invocations/1bff1b30-8f67-413b-9001-16bca4bc6793/targets/%2F%2Fthird_party%2Fpy%2Fmaxtext:tests%2Funit%2Fconfigs_value_test/tests).

tests/unit/configs_value_test.py
```diff
@@ -299,6 +300,57 @@ def test_load_balanced_chunk_context_parallel_config(self):
...
+    base_overrides = {
+        "run_name": "test",
+        "steps": 1,
+        "attention": "dot_product",
+        "attention_type": "block_diffusion",
+        "causal_block_size": 32,
+        "vocab_size": 256,
+        "max_target_length": 2048,
+        "hardware": "cpu",
+        "packing": False,
```
hengtaoguo, line 335: Add `"skip_jax_distributed_system": True` here too.

hengtaoguo (approved):

PR #4764, opened 2026-08-06
[Fix] Apply partial MRoPE to Qwen3.5 attention

AntonyMei:
# Description

Before this fix, partial_rotary_factor is not applied in Qwen3OmniMoeThinkerTextRotaryEmbedding, resulting in repetition / corrupted output during generation. 

Added a long prompt test to `forward_pass_logit_checker.py` to catch such RoPE related issues in future

Added support for reading pkl files in `forward_pass_logit_checker.py`

FIXES: b/543115324

# Tests

- Ran this fix with vllm serve on a local TPU VM v5p-8 and ensures logits matched.
- Added a long prompt test to `forward_pass_logit_checker.py` to catch such RoPE related issues in future

```
P="The city of Lyon sits at the confluence of the Rhone and Saone rivers in eastern France. Founded as a Roman colony in 43 BC, it served for centuries as the capital of the Gauls. Its silk industry flourished in the sixteenth century, drawing weavers from across Europe. The old town preserves narrow covered passageways known as traboules, built so that silk merchants could carry bolts of fabric between workshops without exposing them to rain. "
PROMPT="$P$P$P$P$P"$'\n\n'"Based on the passage above, the traboules of Lyon were originally built so that"

python3 -m tests.assets.logits_generation.generate_hf_golden_logits \
  --model-id=Qwen/Qwen3.5-35B-A3B \
  --hf-model-path=<> \
  --hf-load-dtype=float32 \
  --prompts="$PROMPT" \
  --output-path=/mnt/disks/persist/golden_qwen35_longprompt.pkl \
  --output-format=pickle

python3 -m tests.utils.forward_pass_logit_checker src/maxtext/configs/base.yml \
  model_name=qwen3.5-35b-a3b tokenizer_path=<> \
  load_parameters_path=gs://maxtext-model-checkpoints/qwen3.5-35b-a3b/unscanned/0/items \
  scan_layers=false enable_nnx=true pure_nnx_decoder=true \
  use_multimodal=false per_device_batch_size=1 dtype=bfloat16 weight_dtype=bfloat16 \
  max_target_length=512 \
  --max_kl_div=0.1 --golden_logits_path=/mnt/disks/persist/golden_qwen35_longprompt.pkl
```
This can be used for running the tests in CI

# Checklist

### DIFF
--- a/tests/unit/qwen35_partial_mrope_test.py
+++ b/tests/unit/qwen35_partial_mrope_test.py
@@ -1,0 +1,130 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests Qwen3.5 partial multi-dimensional rotary embeddings."""
+
+import unittest
+from types import SimpleNamespace
+
+from flax import nnx
+import jax.numpy as jnp
+import numpy as np
+
+from maxtext.layers.attentions import Attention
+from maxtext.layers.embeddings import Qwen3OmniMoeThinkerTextRotaryEmbedding
+
+
+_HEAD_DIM = 256
+_PARTIAL_ROTARY_FACTOR = 0.25
+_ROTARY_DIM = 64
+_MROPE_SECTION = (11, 11, 10)
+_ROPE_THETA = 10_000_000
+
+
+def _reference_partial_mrope(inputs: np.ndarray, positions: np.ndarray) -> np.ndarray:
+  """Independent NumPy implementation of the Qwen3.5 partial-MRoPE rule."""
+  if positions.ndim == 2:
+    positions = np.broadcast_to(positions[np.newaxis, ...], (3,) + positions.shape)
+
+  inv_freq = 1.0 / (_ROPE_THETA ** (np.arange(0, _ROTARY_DIM, 2, dtype=np.float32) / _ROTARY_DIM))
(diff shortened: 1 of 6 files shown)


tests/unit/qwen35_partial_mrope_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+_HEAD_DIM = 256
+_PARTIAL_ROTARY_FACTOR = 0.25
+_ROTARY_DIM = 64
+_MROPE_SECTION = (11, 11, 10)
+_ROPE_THETA = 10_000_000
+
+
+def _reference_partial_mrope(inputs: np.ndarray, positions: np.ndarray) -> np.ndarray:
```
hengtaoguo, line 35: This test is great. Is there a huggingface equivalent that you could compare against, instead of this numpy implementation? Similar to [qwen3_5_layers_test.py](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/qwen3_5_layers_test.py). 

Also I wonder if you could move this entire test into the existing qwen3_5_layers_test above.

hengtaoguo (approved):
Thanks for the fix! Qwen3-Omni and Qwen3.5 are two sets of models (although highly similar in structure). 

Does this suggest Qwen3.5 is reusing `Qwen3OmniMoeThinkerTextRotaryEmbedding` when `use_mrope=True`? Does this bug also exist for Omni model?

tests/unit/qwen35_partial_mrope_test.py
A9isha: I am working on adding another long prompt test to forward_pass_logit_checker.py with golden logits from HF checked in for CI

A9isha (commented):
@hengtaoguo yes `Qwen3OmniMoeThinkerTextRotaryEmbedding` is shared but Omni is not affected because for Omni, `partial_rotary_factor = 1.0` makes `rotary_dim = head_dim`, so "rotate everything" — which is what the unfixed code does — is the correct answer for Omni.

tests/unit/qwen35_partial_mrope_test.py
A9isha: added forward_pass_logit_checker.py in this PR

Rohan-Bierneni (approved):
Thank you for this fix! LGTM

PR #4767, opened 2026-08-06
Support Ring Attention with DeepSeek DSA Sparse Indexer and others

zcjhao:
# Description

This PR adds end-to-end support for DeepSeek Sparse Attention (DSA) Indexer with Tokamax Ring Context Parallelism during  training. Previously, `use_indexer=True` was prohibited with Ring Attention due to a lack of dynamic per-ring-step mask extraction. This PR also solve 2 bugs concerning load balancing with Indexer and indexer losses. 
   
### Problems Solved & Key Design Choices:

1. **Enabled Dynamic Indexer Masking in Ring Splash Attention**
  - **API / Autograd**: Plumbed indexer_mask through the public API into the hardware engine, saving it in residuals for the backward pass.
  - **Hardware Injection**: Converted the 2D mask into 4D hardware-native tiles to override standard Splash masking.
  - **Ring Synchronization**: Synchronized dynamic mask shards with the K/V blocks rotating across the TPU interconnect loop.
  - **Backward Alignment**: Transposed the mask block topologies to accurately align with backward gradient shapes. 
  -  **Implementation Details & Notes**
```
    Under the hood, Tokamax Splash achieves causal masking efficiency by passing a `MaskInfo` object to the Pallas hardware. This object tells Splash which hardware tiles are completely zeroed (columns > rows) or completely active (rows > columns), meaning it only has to spend compute cycles applying exact masks to the diagonal intersection tiles. 

    To support the DSA sparse indexer, we hijacked this mechanism. We leave the `_splash_attention_forward` kernel code untouched, and instead manually repack our 2D boolean mask into a 4D tiled grid (`[q_blocks, kv_blocks, bq, bkv]`) and  inject it into `MaskInfo` dynamically.

    For standard all-gathered Splash, MaxText handles indexer masking in the following way: `attention_mla.py` creates the mask ➡️ `attention_op.py` routes it ➡️ `splash_attention_kernel.py` chops it into 4D hardware tiles using `_process_dynamic_mask()`.

    However, we could not reuse `_process_dynamic_mask()` for Ring Attention. For Ring to work, we had to build custom masking plumbing for two reasons:
    1.  Block Rotation: From the Indexer, we generate a mask of `[s_sharded, t]`. Because Ring chunks the `t` axis across the interconnect, we must map and slice the mask into `[s_sharded, t_chunked_i]` to exactly match the specific K/V block arriving on the  current Ring step. 
    2.  Coordinate Preservation: `_process_dynamic_mask()` generates a brand new `MaskInfo` object starting from local coordinate `(0, 0)`. If we called it inside the Ring loop, it would wipe out the sequence offset tracking metadata required for the Ring  progression. Instead, our custom wrapper takes the *existing* offset-tracked `MaskInfo` object and replaces the causal tiles with our indexer tiles step-by-step.

    The backward pass follows this identical blueprint, substituting the transposed topologies.
```
image

2. **Isolated Flax NNX Auxiliary Loss:**
   - In Flax NNX, `_apply_layers_sequentially` filters scanned layer states via `nnx.filter_state(..., nnx.Not((nnx.RngState, nnx.Intermediate)))`. Subclassing `nnx.Intermediate` caused `jax.lax.scan` to prune the loss variable during execution, resulting in `indexer_loss: 0.000` and broken backward gradients.
   - We define `class indexer_losses(nnx.Variable):` which inherits directly from `nnx.Variable` rather than `nnx Intermediate`. This allows the variable to pass through `jax.lax.scan` unharmed without modifying core decoder scanning logic.
    
3. **Loss Harvesting & Objective Injection:**
   - In `train.py`'s `loss_fn`, `indexer_losses` is popped before generic intermediates (mirroring the upstream MTP loss pattern).
   - The loss is extracted across scanned transformer layers and injected into the scalar optimization objective (`loss += indexer_loss`), restoring full automatic differentiation VJP  gradient flow to indexer Query/Key projection weights (`wq_b`, `wkv_b`).
   
4. **Fix Indexer Load Balancing Mask Bug in Splash Attention**
   - During Context Parallel load balancing, the `indexer_mask was` not being restored to chronological order before passing it to Splash Attention (unlike the Key and Value tensors, which were reordered).
   - We verified that without explicitly un-permuting the mask in wrap_flash_attention():
   
   ```
   indexer_mask = max_utils.reorder_sequence(tensor=indexer_mask, cp_size=cp_size, seq_dim=2, to_contiguous=True)
   ```
The newly added unit test `test_tpu_flash_attention_context_parallel_with_indexer` comparing multi-chip Flash Attention load-balancing against a single-chip baseline fails without it.

image

  [screenshot](https://screenshot-v2.corp.google.com/3ga5a9kj305e0)
  
# Tests

- Verified Ring operations in trace [xprof](http://xprof/?session_id=zjiahao-4574040906969517703) and non-zero indexer_loss from [logs](https://screenshot-v2.corp.google.com/0pv4c51bp6qt8)

image

[Image](https://screenshot-v2.corp.google.com/4ji5u82hj7960)

image

[HLO](https://screenshot-v2.corp.google.com/0r0hqiasqq5ro)

- `configs_value_test.py`
    - **Added:** `test_tpu_tokamax_ring_config_validation_accepts_indexer` to verify that `pyconfig.initialize` successfully accepts the combination of MLA, the DSA Sparse Indexer, and Tokamax Ring Attention.
    - **Updated:** Removed the outdated indexer rejection parameterization from `test_tpu_tokamax_ring_config_validation_rejects_unsupported_configs`.

- **`tokamax_ring_attention_test.py`**
  - **Added:** `test_call_ring_attention_threads_indexer_mask_without_segment_ids` to test batch vectorization and mapping threads on the dynamic `indexer_mask` without segmentation IDs.
  - **Added:** `test_call_ring_attention_threads_indexer_mask_with_segment_ids` to test batch vectorization and mapping threads on the dynamic `indexer_mask` when utilizing segmentation IDs.

- **`attention_test.py`**
    - **Added:** `test_tpu_dot_product_context_parallel_with_indexer` — Parameterized forward equivalence test verifying that standard **dot-product** MLA + Indexer calculates logits successfully and accurately across standard context parallelism permutations (testing combinations of matrix sizes and `load_balance=True/False`).
    - **Added:** `test_tpu_flash_attention_context_parallel_with_indexer` — Parameterized forward equivalence test verifying that standard **Flash Attention** MLA + Indexer matches reference outputs properly across standard context parallelism combinations.
    - **Added:** `test_tpu_flash_attention_ring_context_parallel_with_indexer` — Parameterized forward equivalence test verifying that MLA + Indexer under **Tokamax Ring Attention** identically matches the logits of single-device generic dot-product MLA + Indexer.
    - **Added:** `test_tpu_flash_attention_ring_context_parallel_grad_with_indexer` — Parameterized backward gradient equivalence test verifying that the backward input gradients and auxiliary `indexer_losses` match bit-for-bit with the reference gradients when utilizing `is_dkv=True` transposition.
  
- **`train_nnx_test.py`**
  - **Added:** `test_indexer_losses_harvested_and_injected_into_loss` to simulate multi-layer Transformer intermediate loss states and accurately validate the extraction and calculation of the `expected_indexer_loss` scalar objective.
  
  - Ran `deepseek32_vs_reference_test.py` (we werify MaxText Dot Product Baseline matches Official PyTorch version, and our previous tests show that Ring Attention version matches Dot product baseline. Because Ring Attention = Dot Product Baseline and Dot Product Baseline = PyTorch  Reference, this shows that distributed Ring Attention with Indexer matches the reference  implementation): 

```        
tests/unit/deepseek32_vs_reference_test.py::DeepseekV32IndexerTest::test_indexer_match0
  PASSED [ 12%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32IndexerTest::test_indexer_match1
  PASSED [ 25%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32IndexerTest::test_indexer_match2
  PASSED [ 37%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32MLATest::test_mla_parity_dot_product_s128_k128
  PASSED [ 50%]
    tests/unit/deepseek32_vs_reference_test.
  py::DeepseekV32MLATest::test_mla_parity_dot_product_s128_k128_c4 PASSED [ 62%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32MLATest::test_mla_parity_dot_product_s128_k4
  PASSED [ 75%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32MLATest::test_mla_parity_dot_product_s2_k4
  PASSED [ 87%]
    tests/unit/deepseek32_vs_reference_test.py::DeepseekV32MLATest::test_mla_parity_dot_product_s8_k4
  PASSED [100%]

    ======================== 8 passed in 56.02s ========================
```

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -664,7 +664,7 @@ class Attention(BaseModel):
       0,
       ge=0,
       description=(
-          "Chunk size over heads dimension for QK attention dot product in mla. "
+          "Chunk size over heads dimension for QK attention dot product in mla.  "
           "Default is 0 (no chunking). Reduces memory footprint at the cost of time."
       ),
   )
@@ -2942,6 +2942,8 @@ class MaxTextConfig(
       raise ValueError("TPU USP attention requires use_tokamax_splash=True.")
     if self.use_jax_splash:
       raise ValueError("TPU USP attention requires use_jax_splash=False.")
+    if self.use_indexer:
+      raise ValueError("TPU USP attention does not support sparse indexer masks.")
     if self.attention_type != "global":
       raise ValueError("TPU USP attention is initially supported only for global causal attention.")
     if self.packing:
@@ -2952,8 +2954,6 @@ class MaxTextConfig(
       raise ValueError("TPU USP attention does not support ragged attention.")
     if self.attention_sink:
       raise ValueError("TPU USP attention does not support attention sinks.")
-    if self.use_indexer:
-      raise ValueError("TPU USP attention does not support sparse indexer masks.")
     if self.use_chunked_prefill:
       raise ValueError("TPU USP attention does not support chunked prefill yet.")
     if self.use_multimodal:
@@ -3639,7 +3639,13 @@ class MaxTextConfig(
             f"`mla_qk_head_chunk_size` ({self.mla_qk_head_chunk_size}) must cleanly divide exactly into "
             f"`indexer_n_heads` ({self.indexer_n_heads})."
         )
+
     if self.use_indexer:
+      if self.attention_type != AttentionType.MLA.value:
+        raise ValueError(
+            f"`use_indexer=True` requires `attention_type='{AttentionType.MLA.value}'`, since only the "
+            "MLA indexer produces this mask."
+        )
       if self.q_lora_rank == 0:
         raise NotImplementedError("Sparse indexer has not implemented for q_lora_rank = 0.")
       supports_dot_product = self.attention == "dot_product"
@@ -3816,6 +3822,8 @@ class MaxTextConfig(
         raise ValueError("TPU Tokamax ring attention requires context_parallel_size > 1.")
       if self.context_sharding != "context":
         raise ValueError("TPU Tokamax ring attention requires context_sharding='context'.")
+      if self.use_indexer and self.dq_reduction_steps != 0:
+        raise ValueError("TPU Tokamax ring attention with sparse indexer mask only supports dq_reduction_steps=0.")
       if self.dq_reduction_steps not in (0, 3):
         raise ValueError("TPU Tokamax ring attention requires dq_reduction_steps to be 0 or 3.")
       if self.max_target_length % (context_parallel_size * context_parallel_size) != 0:
@@ -3844,8 +3852,6 @@ class MaxTextConfig(
         raise ValueError("TPU Tokamax ring attention does not support ragged attention.")
       if self.attention_sink:
         raise ValueError("TPU Tokamax ring attention does not support attention sinks.")
-      if self.use_indexer:
-        raise ValueError("TPU Tokamax ring attention does not support sparse indexer masks.")
       if self.use_chunked_prefill:
         raise ValueError("TPU Tokamax ring attention does not support chunked prefill yet.")
       if self.moba:
@@ -3883,6 +3889,8 @@ class MaxTextConfig(
         raise ValueError("TPU Ulysses attention requires use_tokamax_splash=True.")
       if self.use_jax_splash:
         raise ValueError("TPU Ulysses attention requires use_jax_splash=False.")
+      if self.use_indexer:
+        raise ValueError("TPU Ulysses attention does not support sparse indexer masks.")
       if self.attention_type != "global":
         raise ValueError("TPU Ulysses attention is initially supported only for global causal attention.")
       if self.context_parallel_load_balance:
@@ -3895,8 +3903,6 @@ class MaxTextConfig(
         raise ValueError("TPU Ulysses attention does not support ragged attention.")
       if self.attention_sink:
         raise ValueError("TPU Ulysses attention does not support attention sinks.")
-      if self.use_indexer:
-        raise ValueError("TPU Ulysses attention does not support sparse indexer masks.")
       if self.use_chunked_prefill:
         raise ValueError("TPU Ulysses attention does not support chunked prefill yet.")
       if self.use_multimodal:
--- a/src/maxtext/kernels/attention/tokamax_ring_attention.py
+++ b/src/maxtext/kernels/attention/tokamax_ring_attention.py
@@ -110,8 +110,6 @@ def validate_tokamax_ring_runtime(
     raise ValueError("TPU Tokamax ring attention does not support chunked prefill yet.")
   if sinks is not None:
     raise ValueError("TPU Tokamax ring attention does not support attention sinks.")
-  if indexer_mask is not None:
-    raise ValueError("TPU Tokamax ring attention does not support indexer masks.")
   if bidirectional_mask is not None:
     raise ValueError("TPU Tokamax ring attention does not support bidirectional masks.")
   if record_max_logits:
@@ -283,6 +281,7 @@ def make_sharded_ring_attention_kernel(
     ring_axis: str,
     attn_logits_soft_cap: float | None,
     maybe_shard_with_pspec: Any,
+    mask: Any = None,
 ):
   """Builds and shards the Tokamax ring attention kernel for MaxText."""
   splash_config = build_splash_config(
@@ -295,11 +294,17 @@ def make_sharded_ring_attention_kernel(
   if config.use_max_logit_estimate > 0:
     splash_config = dataclasses.replace(splash_config, max_logit_const=config.use_max_logit_estimate)
 
-  mask = _make_causal_mask(
-      (query.shape[2], key.shape[2]),
-      context_parallel_size,
-      load_balanced=config.context_parallel_load_balance,
-  )
+  if mask is None:
+    # When using the indexer, causal masking is unified into the dynamic indexer_mask
+    # and applied dynamically per block; use FullMask to avoid duplicate static masks.
+    if getattr(config, "use_indexer", False):
+      mask = tokamax_splash_mask.FullMask((query.shape[2], key.shape[2]))
+    else:
+      mask = _make_causal_mask(
+          (query.shape[2], key.shape[2]),
+          context_parallel_size,
+          load_balanced=config.context_parallel_load_balance,
+      )
 
   @functools.partial(jax.jit, static_argnames=["single_head_mask"])
   def wrap_ring_kernel(single_head_mask):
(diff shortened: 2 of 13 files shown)


RissyRan (commented):
LGTM at high level! One question about indexer loss. When onboarded, we were testing Linen instead of NNX. Could you have a run with old version to see if your new changes align with previous runs? Thanks!

```
enable_nnx: false
pure_nnx_decoder: false
pure_nnx: false
```

RissyRan:
cc @huytransformer helps take a review on kernel part

src/maxtext/layers/attention_mla.py
huytransformer, line 1293: missing `segment_positions=inputs_positions` I think may cause trouble with load balancing

huytransformer, line 367: can we raise an error in ring attention when use_indexer is true but the mask is None? If indexer exit early then ring doesn't have mask at all?

src/maxtext/configs/types.py
huytransformer, line 3657: Can we run this check for ring+indexer regardless of indexer_loss_scaling_factor? You can maybe also validate that use_indexer requires mla, since only the MLA indexer produces this mask

src/maxtext/kernels/attention/tokamax_ring_attention.py
```diff
@@ -317,11 +315,17 @@ def make_sharded_ring_attention_kernel(
...
 
-  mask = _make_causal_mask(
-      (query.shape[2], key.shape[2]),
-      context_parallel_size,
-      load_balanced=config.context_parallel_load_balance,
-  )
+  if mask is None:
+    # When using the indexer, causal masking is unified into the dynamic indexer_mask
+    # and applied dynamically per block; use FullMask to avoid duplicate static masks.
+    if getattr(config, "use_indexer", False):
```
huytransformer, line 300: Can we raise when use_indexer is set but indexer_mask is None? If the indexer returns None then would be no causal at all

src/maxtext/trainers/pre_train/train.py
huytransformer, line 119: can we also add "indexer_losses" to this list like mtp_losses, or reject indexer+linen

tests/unit/train_nnx_test.py
```diff
@@ -109,6 +109,16 @@ def __call__(self, decoder_input_tokens, decoder_positions, **kwargs):
     return out
 
 
+class _TinyDecoderIndexerLoss(_TinyDecoder):
+  """`_TinyDecoder` that also sows `indexer_loss` intermediates across its layers."""
+
+  def __call__(self, decoder_input_tokens, decoder_positions, **kwargs):
+    out = super().__call__(decoder_input_tokens, decoder_positions, **kwargs)
+    self.sow(nnx.Intermediate, "indexer_loss", jnp.array([0.25]))
```
huytransformer: This tests the old nnx.Intermediate type with no scan, so indexer_losses isn't covered. can we test the real type via scanned layers?

tests/unit/attention_test.py
```diff
@@ -2961,6 +2623,225 @@ def ring_loss(lnx):
         f"context_parallel_load_balance={context_parallel_load_balance}.",
     )
 
+  @parameterized.named_parameters(
+      {"testcase_name": "no_load_balance", "context_parallel_load_balance": False},
```
huytransformer: can we add a case with a small topk, and one cp=4 case?

zcjhao:
> LGTM at high level! One question about indexer loss. When onboarded, we were testing Linen instead of NNX. Could you have a run with old version to see if your new changes align with previous runs? Thanks!
> 
> ```
> enable_nnx: false
> pure_nnx_decoder: false
> pure_nnx: false
> ```

It seems Linen is currently broken for DSA, [Line](https://xmanager.corp.google.com/experiments/278465606), and the same configs succeed in [nnx](https://xmanager.corp.google.com/experiments/278465833) (Should I add a check that rejects linen + Indexer?). What happened was that I suddenly stopped seeing Indexer loss calculation in my traces, and I found that it seems that due to some recent changes of NNX, indexer_losses silently defaulted to 0 (breaking loss function and backward for Indexer), just like in https://github.com/AI-Hypercomputer/maxtext/pull/4525 for `mtp_loss` because https://github.com/AI-Hypercomputer/maxtext/blob/5e89a48939bffd97ff664401dacf0c052e90e24c/src/maxtext/layers/nnx_decoders.py#L1092 that dropped all `nnx.Intermediate` after a layer pass. I followed I similar approach by creating a custom subclass for indexer_losses, making sure it does not get dropped by the `nnx.Intermediate` filter and manually pop it in train.py. I believe the only part this would affect is how indexer_losses is handled, and I have checked that it correctly appears again in the logs as shown in the PR description, all other components remain the same, I believe comparison with Linen is not required.

tests/unit/attention_test.py
```diff
@@ -1371,7 +1371,7 @@ def test_tpu_flash_attention_context_parallel(
     cfg_cp = pyconfig.initialize(
         [sys.argv[0], get_test_config_path()],
         **self.config_arguments,
-        ici_context_parallelism=ici_context_parallelism,
+        ici_context_parallelism=2,
```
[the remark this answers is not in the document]
NuojCheng: why they are hardcoded as 2?

src/maxtext/trainers/pre_train/train.py
```diff
@@ -275,15 +283,16 @@ def loss_fn(model, config, data, dropout_rng, params, sparsity_state=None, is_tr
...
   indexer_loss = 0.0
   if config.use_indexer and config.indexer_loss_scaling_factor > 0.0:
-    indexer_losses = maxtext_utils.collect_intermediates_by_suffix(intermediate_outputs, "self_attention", "indexer_loss")
-    if indexer_losses:
-      indexer_loss = jnp.mean(jnp.concatenate(indexer_losses))
-      loss += indexer_loss
+    # Recursively collect per-layer indexer losses across all scanned transformer layers.
+    indexer_losses_list = maxtext_utils.collect_intermediates_by_suffix(intermediate_outputs, "indexer_loss")
+    if indexer_losses_list:
+      indexer_loss = jnp.mean(jnp.concatenate(indexer_losses_list))
```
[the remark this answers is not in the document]
NuojCheng: I think it is a solid point

tests/unit/attention_test.py
```diff
@@ -2961,6 +2961,466 @@ def ring_loss(lnx):
...
+    """Test equivalence between single-device dot_product MLA + Indexer and multi-device dot_product + CP + Indexer"""
+    config_arguments = {
+        "per_device_batch_size": 1.0,
+        "run_name": "test",
+        "enable_checkpointing": False,
+        "max_target_length": 512,
+        "attention_type": AttentionType.MLA.value,
+        "use_indexer": True,
+        "indexer_loss_scaling_factor": 0.0,
+        "indexer_topk": 256,
```
NuojCheng: "indexer_topk": indexer_topk?

NuojCheng: This command is applicable to all the tests below

src/maxtext/kernels/tokamax_splash_attention/ring_attention_kernel.py
```diff
@@ -60,8 +60,61 @@ def _validate_ring_axis_size(ring_axis: str, ring_axis_size: int, expected_ring_
...
+) -> MaskInfo:
+  """Injects a pre-sliced dynamic Indexer mask shard into local MaskInfo for the current ring step."""
+  if local_idx_mask is None:
+    return local_mask_info
+
+  bq, bkv = block_shape
+  if local_idx_mask.ndim == 3:
+    local_idx_mask = local_idx_mask[0]
+  if local_idx_mask.dtype != jnp.bool_:
+    local_idx_mask = jnp.isclose(local_idx_mask, 0.0)
```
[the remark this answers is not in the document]
NuojCheng: good point, agree

src/maxtext/kernels/tokamax_splash_attention/ring_attention_kernel.py
```diff
@@ -229,10 +308,34 @@ def rotate_kv(k_current, v_current, segment_ids_current):
...
+  else:
+    mask_4d = None
+    step0_mask = None
+    xs = jnp.arange(1, ring_axis_size)
+
+  def compute_step(i, local_idx_mask, k_current, v_current, segment_ids_current, dq_accum):
     current_kv_shard_idx = (ring_axis_idx - i) % ring_axis_size
     local_dkv_mask_info = _dynamic_slice_mask_info(dkv_mask_info, current_kv_shard_idx, ring_axis_size)
     local_dkv_mask_info = _offset_q_sequence_for_kv_shard(local_dkv_mask_info, current_kv_shard_idx, k_current.shape[-2])
+    local_dkv_mask_info = _inject_local_indexer_mask(
```
huytransformer, line 351: Can u check if dq_reduction_step=3 give correct dQ when ring attention use indexer mask? If not perhaps reject this combination for now?

src/maxtext/layers/attention_mla.py
```diff
@@ -1314,7 +1324,7 @@ def __call__(
             sparse_loss=self.config.indexer_sparse_training,
             scaling_factor=self.config.indexer_loss_scaling_factor,
         )
-        self.indexer_loss = nnx.Intermediate(indexer_loss)
+        self.indexer_loss = indexer_losses(indexer_loss)
```
huytransformer, line 1327: can u add indexer_losses to Decoder.scan_decoder_layers() and test with pure_nnx=False and scan_layers=True

huytransformer (approved):
Can u please address these 2 then this LGTM

NuojCheng (approved):

RissyRan (approved):

PR #4790, opened 2026-08-08
Gemma-4 E2B: enable image (multimodal) E2E parity — vision clipped-linears + padded-patch masking + PLE substitution (E4B vision-tower validated; E4B E2E pending)

lokic233:
## What
Enables **Gemma-4 E2B full end-to-end image (multimodal) parity** in MaxText. E4B runs through the same
contract-gated code path and its **vision tower is validated (cosine 0.999999 vs HF)**, but **E4B full decoder
E2E parity is PENDING** (no E4B E2E fixture/converted checkpoint this cycle) — E4B is **not** claimed as
E2E-validated.

## Reference contract (pinned)
- HF Transformers **5.9.0** (`transformers/models/gemma4/modeling_gemma4.py`), semantic reference:
  `Gemma4ClippableLinear` (clamp-in -> linear -> clamp-out) + the PLE block.
- Model/processor/tokenizer: gemma4-e2b -it revision as in the model yml.

The five load-bearing pieces (all required; ablation below):
1. clipped linears - per-projection checkpoint activation clip bounds (16 blocks x 7 projections = 112 modules / 448 scalar bounds).
2. padded-patch masking + real image positions (Option-S) - pad patches excluded from attention.
3. image positions / pooling - per-patch (x,y); position-based pooling to soft tokens.
4. PLE pad substitution - image placeholder rows -> pad_token_id in the token-identity PLE path (HF-faithful; "both" is HF-divergent, ablation-only).
5. causal image spans - E2B/E4B image tokens attend causally (HF use_bidirectional_attention=None).

## Validation
E2B, MaxText NNX vs pinned HF 5.9.0, teacher-forced (gate: post_image max_KL <= 1.26e-3; argmax >= 0.995):

| fixture | segment | max_KL | argmax | gate |
|---|---|---|---|---|
| E2B structured | post_image (n=75) | 4.29e-05 | 1.000 | PASS |
| E2B natural | post_image (n~72) | 4.71e-05 | 1.000 | PASS |

Ablation (E2B structured - every component load-bearing):

| arm | post_image max_KL | image_span argmax | parity |
|---|---|---|---|
| full fix | 4.29e-05 | 1.000 | PASS |
| clip OFF | 1.195 | 0.596 | FAIL |
| padded-patch masking OFF | 0.328 | 0.762 | FAIL |
| PLE substitution OFF | 8.921 | 0.000 | FAIL (dominant) |
| PLE mode "both" (HF-divergent) | 13.44 | 0.346 | FAIL |

In-tree tests (CPU): tests/unit/gemma4_clip_linears_test.py (18) + tests/unit/gemma4_mm_contract_gate_test.py (8).

**Later integration validation (composite stack - integration provenance, NOT a claim that #4790 alone fixes
checkpoint-loader gaps):** production train.py completes forward -> backward -> optimizer-apply -> checkpoint-save
on real Gemma-4 E2B weights on the composite stack #4790 + #4807 (std-fidelity candidate) + an evidence-only
full-state checkpoint remap/splice bridge + stock train.py:
- CPU: loss 14.941 -> 11.596 (finite, decreasing); theta1 fresh-reload green; 448 vision clip bounds byte-identical theta0->theta1; 751/751 non-clip trainable leaves changed.
- TPU (tpu7x single host): loss 15.917 -> 14.020 (finite, decreasing); theta1 saved with populated opt_state; exit 0.

The params-only converted checkpoint does not yet enter stock load_parameters_path cleanly (a separate loader
69-leaf drop, tracked independently); the composite evidence used the splice bridge, which is NOT presented as
an upstream-supported loader path.

## Related work (separate PRs / follow-ups - each owns its own slice; do NOT fold together)
- #4807 - Gemma-4 vision std_bias/std_scale fidelity (standalone; #4790 forward parity does NOT depend on it).
- #4860 - Gemma-4 small HF export shape map (GEMMA4_SMALL_HF_WEIGHTS_TO_SHAPE vision + clip-bound coverage);
  composes with this PR's GEMMA4_SMALL_MAXTEXT_TO_HF_PARAM_MAPPING clip-bound targets.
- Shared-layer export completeness (S13) - a serving/checkpoint-completeness invariant tracked separately.

## Limitations
- E4B: vision-tower validated; full decoder E2E parity pending.
- Multimodal packing unsupported (fail-closed).
- Multi-image E2E: not yet in the in-tree matrix.
- The composite-stack backward/save evidence is integration provenance; the stock params-only loader gap is a separate issue.

## Gemma-4 small (E2B/E4B) vs large (26B/31B) — what differs, and why the new flags

The vision **tower** is shared (`vision_encoder_block: "gemma4"`, same `image_size_for_vit`). The E2B/E4B
config flags introduced here configure the *small-model image contract*; **each defaults to a no-op**, so
26B/31B and all other models are behaviorally unchanged.

| Aspect | E2B / E4B (small) | 26B / 31B (large) | Flag (default) |
|---|---|---|---|
| Per-layer embeddings (PLE) | Yes (`hidden_size_per_layer_input=256`) — needs image-row→pad substitution | No PLE path | `ple_pad_substitute_image_rows=False`, `ple_pad_mode="identity"`, `ple_pad_token_id=0`, `image_placeholder_token_id` (= `GEMMA4_TOKEN_PLACEHOLDER`) |
| Image-span attention | **Causal** (HF `use_bidirectional_attention=None`) | **Bidirectional** image spans | `use_bidirectional_image_attn=False` |
| Vision activation clip bounds | Checkpoint ships per-projection `{input,output}_{min,max}` (112 modules / 448 scalars) | Not present | `use_clipped_linears_for_vit=False` |
| Patch handling | Padded-patch / real per-patch positions supported (native-res / pan-and-scan); pad patches masked + position-pooled | Fixed-grid all-valid | `encoder_image_position_ids` present in data → opt-in; else legacy path |
| KV sharing | `num_kv_shared_layers` (E2B=20 / E4B=…) | n/a | model yml |

The **general** MM decode path (placeholder → `get_bidirectional_mask_vision` → `merge_mm_embeddings`) is
identical to 26B/31B and unchanged. The small-model additions live in (1) the PLE substitution
(`nnx_decoders._apply_gemma4_small_layers`, small-only) and (2) the padded-patch vision path
(`gemma4_vision.Gemma4VisionEncoderLayer`, opt-in via per-patch positions). All large-model and non-Gemma4
code paths take the default no-op branch.

---
_Refreshed with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_description&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._


lokic233:
Converted to **draft** and narrowed the scope after end-to-end validation.

**What changed vs the initial version of this PR:**
- Removed the validator change that would have unblocked E2B/E4B multimodal. Clip-bounds are necessary but **not sufficient** for image parity, so unblocking would have implied working image support that isn't there yet.
- Retitled to reflect that this lands the clip-bounds **building block** only.

**E2E findings (CPU, teacher-forced, vs HF reference logits):**
- The 448 clip bounds convert + load correctly (real finite values), and the clamp math is exact — verified at the helper, forward-component, and checkpoint-load levels.
- With clip-bounds loaded, the **text path is exact** (pre-image max KL ≈ 2.7e-5) but the **image span still diverges** (post-image max KL ≈ 10, argmax ≈ 0.55 — far from parity). Clip-ON vs clip-OFF barely differ, i.e. the clip effect is swamped by a larger vision-forward divergence.
- Root cause is **not** the clip math: the current Gemma-4 vision forward is missing other pieces of the reference contract (pad-patch attention masking, external image-position threading, and likely more). Those are out of scope for this change and being worked separately.

This clip-bounds change is off by default and safe to land on its own; keeping it as draft until the remaining vision-forward pieces land, at which point E2B/E4B image parity can be demonstrated end-to-end.

lokic233:
**Update — this PR now closes E2B/E4B image parity end-to-end.**

It initially landed only the vision clipped-linears (marked draft, since clip-bounds alone are necessary but not sufficient). After root-causing the remaining image-span divergence, the missing pieces were:

1. **Padded-patch vision self-attention masking** + real per-patch position threading (phantom pad patches must be masked; pooled by real positions to the valid tokens).
2. **Decoder per-layer-embedding (PLE) image-row substitution** — E2B/E4B map image placeholder tokens → `pad_token_id` in the PLE path (HF `modeling_gemma4`); feeding the placeholder id instead was the dominant residual.
3. **Causal image spans** for E2B/E4B.

With all three plus the clip-bounds, the teacher-forced forward now matches the HF reference: post-image `max_KL 4.3e-5`, `argmax 1.0` (gate `≤ 1.26e-3` / `≥ 0.995`), vs `1.195` with clip-bounds disabled. Marking ready for review. All behavior is opt-in and defaults keep other models unchanged.

src/maxtext/models/gemma4_vision.py
```diff
@@ -30,6 +32,108 @@
...
+  for nm in ("input_min", "input_max", "output_min", "output_max"):
+    v = getattr(cb, nm).value
+    if getattr(v, "shape", ()) not in ((), (1,)):
+      raise ValueError(f"Gemma4 vision clip bound '{nm}'{(' @ '+where) if where else ''} has non-scalar "
+                       f"shape {v.shape}; expected scalar.")
+    fv = float(jnp.reshape(v, (-1,))[0])
+    if not bool(jnp.isfinite(jnp.asarray(fv))):
+      raise ValueError(f"Gemma4 vision clip bound '{nm}'{(' @ '+where) if where else ''} = {fv} is non-finite "
+                       f"(missing/NaN/Inf). use_clipped_linears_for_vit=True declares a FINITE clipped model; "
+                       f"refusing to fall back to an identity clamp.")
```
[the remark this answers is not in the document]
lokic233: Fixed in the pushed head (35ce712e). `validate_all_vision_clip_bounds(model)` now runs in `model_creation_utils.from_pretrained` AFTER the Orbax restore and BEFORE the first JIT, gated on `use_clipped_linears_for_vit and use_multimodal`. It walks the model with `nnx.iter_graph`, validates every `_ClipBounds` module (finite, scalar, min<=max), and asserts EXACTLY 112 modules / 448 scalar bounds — hard-failing on missing/NaN/Inf/wrong-count (incl. zero) rather than letting a NaN propagate into `jnp.clip`. Real-TPU: this validation PASSED (112/448) on TPU7x. Tests: `tests/unit/gemma4_clip_linears_test.py` (rejects NaN/Inf/min>max; expected-count constants).

src/maxtext/models/gemma4_vision.py
```diff
@@ -30,6 +32,108 @@
...
+
+def clip_optimizer_freeze_mask(params_tree):
+  """Bool pytree (same structure as ``params_tree``): True for TRAINABLE leaves,
+  False for the immutable clip bounds. Feed to ``optax.masked``/``multi_transform``
+  so the bounds get ``set_to_zero()`` updates. Path-based, so it survives the
+  nnx->linen->orbax round-trip regardless of leaf type erasure."""
+  flat = jax.tree_util.tree_flatten_with_path(params_tree)[0]
+  leaves_mask = [not _is_clip_bound_path(path) for path, _ in flat]
+  treedef = jax.tree_util.tree_structure(params_tree)
+  return jax.tree_util.tree_unflatten(treedef, leaves_mask)
```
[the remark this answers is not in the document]
lokic233: Fixed in the pushed head (525d38a0). `clip_optimizer_freeze_mask` is now wired into `optimizers.get_optimizer`: when `use_clipped_linears_for_vit` is set, the 448 clip-bound leaves are composed into the `optax.multi_transform` partition as `set_to_zero()` (no updates, no weight decay, no momentum/variance slots) — and this holds even when `freeze_vision_encoder_params=False`. Defense in depth: `lax.stop_gradient` is also applied at every clamp use-site (`_clip_in`/`_clip_out`). Unit tests assert the freeze-mask polarity/count and that the bounds receive zero gradient (`bounds_are_stop_gradient`).

src/maxtext/models/gemma4_vision.py
```diff
@@ -418,6 +530,116 @@ def init_rotary_embedding(self) -> Gemma4VisionRotaryEmbedding:
...
+    Called once by ``Gemma4EncoderBlock`` after construction when
+    ``config.use_clipped_linears_for_vit`` is set. Idempotent.
+    """
+    if getattr(self, "_use_clipped_linears", False):
+      return
+    self._use_clipped_linears = True
+    self.q_clip = _make_clip_state()
+    self.k_clip = _make_clip_state()
+    self.v_clip = _make_clip_state()
+    self.o_clip = _make_clip_state()
```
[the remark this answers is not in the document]
lokic233: Fixed in the pushed head (35ce712e). `Gemma4Attention.enable_vision_clipped_linears` now raises if `fused_qkv=True` (mirrors the existing `fused_mlp` guard). Also added a static config-validation gate in `configs/types.py` that hard-fails E2B/E4B multimodal with `fused_qkv=True` or `fused_mlp=True` before model construction. Config tests cover both (`tests/unit/gemma4_mm_contract_gate_test.py`: `test_fused_qkv_hard_fails`, `test_fused_mlp_hard_fails`).

lokic233:
**Update — hardening pushed to the PR head (797842b1); ready for re-review (E2B).**

The three review-bot findings are addressed with tests (replies in-thread): (1) clip validation now runs post-restore with exact 112/448 accounting; (2) clip bounds are immutable (stop_gradient at use-site + optimizer `set_to_zero` freeze, independent of `freeze_vision_encoder_params`); (3) `fused_qkv`/`fused_mlp` fail closed. Also: PLE contract adjudicated vs pinned HF 5.9.0 (identity is HF-faithful; "both" is HF-divergent, ablation-confirmed), Option-S sentinel/placeholder invariants, and a real NaN garbage-pad contamination bug found by the test and fixed.

Scope has been made honest: **E2B full E2E parity is validated** (post_image max_KL 4.29e-5/argmax 1.0 structured; 4.71e-5/1.0 natural; full component ablation in the body). **E4B is vision-tower-validated only (cosine 0.999999); full E4B decoder E2E parity is pending** and is not claimed. Real-TPU: `from_pretrained` load + 112/448 validation pass on TPU7x; a clean one-step trainer+checkpoint-roundtrip demo is pending a flax-version nnx grad buffer-lifetime issue in the demonstration harness (not the model code; the immutability mechanism is unit-proven). 26 in-tree CPU tests. Not requesting merge — inviting review.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -2233,14 +2234,47 @@ def _apply_gemma4_small_layers(
...
+      ple_tokens = decoder_input_tokens
+      ple_context = y
+      # Gemma-4 E2B/E4B build the per-layer inputs from llm_input_ids with the image placeholder
+      # tokens mapped to pad_token_id (HF modeling_gemma4.py), rather than feeding the image
+      # placeholder id / merged image features into the PLE path. Without this substitution the
+      # per-layer embeddings at the image placeholder positions diverge from the reference, which
+      # corrupts the image-span and post-image logits. Gated on ple_pad_substitute_image_rows
+      # (default False preserves the native PLE for other models).
+      if bool(getattr(cfg, "ple_pad_substitute_image_rows", False)) and multimodal_input is not None:
+        _img_id = int(getattr(cfg, "image_placeholder_token_id", 258880))
```
hengtaoguo, line 2280: Big Gemma4 models have already supported multimodal decoding, without specifically adding `image_placeholder_token_id` into the types. Why is this block change necessary here? I imagine they should share the same decoding logics?

src/maxtext/configs/base.yml
```diff
@@ -1194,6 +1194,18 @@ freeze_vision_encoder_params: true
...
+# scalar {input,output}_{min,max}). A prerequisite for Gemma-4 E2B/E4B image parity
+# (necessary but not on its own sufficient); no-op for other vision encoders. Bounds are
+# checkpoint-resident, non-trainable scalars.
+use_clipped_linears_for_vit: false
+# Gemma-4 E2B/E4B decoder image-handling (defaults preserve behavior for other models):
+use_bidirectional_image_attn: false  # E2B/E4B image spans are causal
+ple_pad_substitute_image_rows: false  # substitute pad id for image rows in the per-layer-embedding path (HF gemma4)
+ple_pad_mode: "identity"  # 'identity' (token-id path) or 'both' (also context/embedding path)
+image_placeholder_token_id: 258880  # GEMMA4_TOKEN_PLACEHOLDER
+ple_pad_token_id: 0  # E2B text_config.pad_token_id
```
hengtaoguo, line 1258: Does this imply the Gemma4-E2B/E4B have different vision logics comparing to Gemma4-26B/31B? Since the latter doesn't have these flags.

src/maxtext/configs/types.py
```diff
@@ -3610,16 +3651,84 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
...
+            raise ValueError(
+                "image_placeholder_token_id must be a valid non-negative token id when "
+                "ple_pad_substitute_image_rows=True (derive it from the model/tokenizer config)."
+            )
+          if int(self.ple_pad_token_id) < 0:
+            raise ValueError(
+                "ple_pad_token_id must be a valid non-negative token id when "
+                "ple_pad_substitute_image_rows=True (Gemma-4 E2B text_config.pad_token_id=0)."
+            )
+        # (c) Gemma-4 E2B/E4B image spans are CAUSAL. Bidirectional image attention is a Gemma-3 / 26B / 31B
```
hengtaoguo, line 4068: I believe you are referring to Gemma-4 26B/31B here? Could you also document the differences between these two sizes in the PR description? Thanks!

src/maxtext/layers/decoders.py
```diff
@@ -1679,6 +1679,18 @@ def _apply_gemma4_small_layers(
...
+      # substitution (the HF-faithful mapping of image placeholder rows -> pad in the token-identity PLE
+      # path). That logic lives in the NNX decoder (nnx_decoders._apply_gemma4_small_layers) and is exercised
+      # by the supported enable_nnx=True path. Running E2B/E4B multimodal through this pure-Linen path would
+      # silently diverge on the image-span / post-image logits, so we refuse rather than degrade quietly.
+      if bool(getattr(cfg, "ple_pad_substitute_image_rows", False)) and multimodal_input is not None:
+        raise NotImplementedError(
+            "Gemma-4 E2B/E4B multimodal PLE pad-substitution is only implemented on the NNX decoder path "
+            "(enable_nnx=True, the default). The pure-Linen decoder does not perform the image-row PLE "
+            "substitution and would silently diverge from the HF reference. Set enable_nnx=True (default) to "
+            "run Gemma-4 E2B/E4B multimodal."
```
hengtaoguo, line 1693: Since we are moving away from Linen to NNX, I think this is acceptable. How would you proceed with E2E test in this case?

src/maxtext/layers/encoders.py
```diff
@@ -106,13 +106,21 @@ def _setup_vision_encoder_layers(self):
...
+    vision_image_masks = None
     if self.vision_encoder_block.value.startswith("qwen3") and input_masks is not None:
       encoder_output = encoder(
           input_images, video_mask=input_masks, video_grid_thw=video_grid_thw, deterministic=deterministic
       )
+    elif self.vision_encoder_block == VisionEncoderBlockType.GEMMA4 and image_position_ids is not None:
+      # Gemma-4 padded-patch path: pre-patchified patches + per-patch positions (-1 = pad). The
+      # encoder returns (embeddings, image_masks); the mask marks the valid pooled tokens.
+      encoder_output = encoder(input_images, deterministic=deterministic, image_position_ids=image_position_ids)
+      embeddings, vision_image_masks = encoder_output
```
hengtaoguo, line 122: Why is the `vision_image_masks` necessary here? IIUC, in preprocessing utils, we've already processed the prompt and know here to insert the image tokens ([ref](https://github.com/AI-Hypercomputer/maxtext/blob/e591f30cbbf217ed67ef71263ec0538239f9c3b8/src/maxtext/layers/nnx_decoders.py#L1340)).

src/maxtext/layers/encoders.py
```diff
@@ -131,7 +139,7 @@ def __call__(self, input_images, input_masks=None, video_grid_thw=None, determin
     projector = getattr(self, self.projector_name)
     embeddings = projector(embeddings)
 
-    return embeddings, deep_feats
+    return embeddings, deep_feats, vision_image_masks
```
hengtaoguo, line 142: Since this line changed the return interface, we need to make sure whether it's necessary or not.

src/maxtext/layers/encoders.py
```diff
@@ -106,13 +106,21 @@ def _setup_vision_encoder_layers(self):
 
     return encoder_name, projector_name
 
-  def __call__(self, input_images, input_masks=None, video_grid_thw=None, deterministic=False):
+  def __call__(self, input_images, input_masks=None, video_grid_thw=None, deterministic=False,
+               image_position_ids=None):
```
hengtaoguo, line 110: Not sure why the `image_position_ids` matter here? IIUC only the image tensors are fed to the vision encoder, without any prompts. Is it because of the padding?

src/maxtext/optimizers/optimizers.py
```diff
@@ -239,14 +239,32 @@ def get_optimizer(config, learning_rate_schedule, model=None):
...
+  freeze_clip_bounds = bool(getattr(config, "use_clipped_linears_for_vit", False))
+
+  if freeze_mask_fn is not None or freeze_clip_bounds:
+    def _partition(params):
+      # trainable_mask: True where trainable under the whitelist (all-True if no whitelist).
+      if freeze_mask_fn is not None:
+        trainable_mask = freeze_mask_fn(params)
+      else:
+        trainable_mask = jax.tree_util.tree_map(lambda _: True, params)
+      if freeze_clip_bounds:
```
hengtaoguo, line 257: Add a one-line comment here to call out for Gemma4-small ViT?

tests/unit/gemma4_clip_linears_test.py
hengtaoguo, line 1: For the two tests added here, I would recommend move them to existing tests, such as `gemma4_small_test.py` or `gemma4_small_layers_test.py`

https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/gemma4_small_layers_test.py
https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/unit/gemma4_small_test.py

hengtaoguo (commented):
Thanks for the great contributions! Looks like Gemma4-small models have several differences comparing to the supported 26B/31B. We should call out for those differences while try to introduce minimal interface changes.

src/maxtext/layers/nnx_decoders.py
lokic233: Good question — the *general* MM decode path IS shared and unchanged. Placeholder→bidirectional-mask→
`merge_mm_embeddings` still goes through `processor.get_bidirectional_mask_vision`
(`decoder_input_tokens == GEMMA4_TOKEN_PLACEHOLDER`), exactly as 26B/31B. This block is **not** that path
— it's the **per-layer-embedding (PLE)** path, which only the small models have (`gemma4-e2b/e4b` set
`hidden_size_per_layer_input=256` / `vocab_size_per_layer_input`; 26B/31B have no PLE at all, so there is
nothing to share here).

HF `modeling_gemma4` builds the per-layer inputs from `llm_input_ids` with the image-placeholder rows
mapped to `pad_token_id` in the token-identity path (`get_per_layer_inputs`). If we feed the image
placeholder id straight into the PLE embedder instead, the per-layer embeddings at the image positions
diverge from HF and corrupt the image-span + post-image logits (ablation: PLE-substitution-OFF is the
dominant failure, post_image max_KL 8.9, image-span argmax 0.000). It's gated on
`ple_pad_substitute_image_rows` (default False → no-op for every other model).

One simplification I'm happy to take: rather than a new `258880` default, reuse the existing
`processor_gemma4.GEMMA4_TOKEN_PLACEHOLDER` as the single source of truth (the config field then only
exists to override, not to re-declare the constant). Let me know if you'd prefer that — happy to wire it.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/layers/encoders.py
lokic233: Yes — it's padding, and these are **per-patch spatial (x,y) coordinates**, not prompt/text positions. This
is the native-resolution / pan-and-scan ("Option-S" padded-patch) path: `inputs` are *already-patchified*
pixel values `[B, L, P*P*C]` where some of the `L` patch rows are **pad** (marked by a `-1` sentinel in
`image_position_ids`). The positions drive two things the encoder can't otherwise know:
1. segment-id masking — pad patches (`pos == -1`) get a distinct segment so they're excluded from
   self-attention (phantom patches don't leak into valid patches);
2. `avg_pool_by_positions` — pooling to soft tokens uses the real coordinates (a `-1` patch maps to a
   zero-weight bucket).

The stock fixed-grid Gemma-4 processor emits all-valid patches and does **not** set this key, so the legacy
all-valid path runs unchanged (there's a fail-closed guard so pre-patchified input can't silently take the
legacy branch). So it's opt-in and only active for the dynamic-N padded-patch contract.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/layers/encoders.py
lokic233: These are two orthogonal masks. The one at `nnx_decoders.py:1340` is `bidirectional_mask =
(decoder_input_tokens == GEMMA4_TOKEN_PLACEHOLDER)` — it marks **WHERE in the text sequence** the image
embeddings land (unchanged). `vision_image_masks` instead marks **WHICH of the K pooled vision tokens are
valid vs pad**, and is passed as `merge_mm_embeddings(..., token_masks=...)`.

It matters because in the padded-patch path the pooling produces a fixed `K` soft tokens but only some are
backed by real patches. `merge_mm_embeddings` uses `token_masks` to `argsort(-token_mask)` and **pack the
valid vision tokens to the front** before scattering them into the placeholder positions
(`_merge_mm_embeddings_inner`, `utils.py`). Without it, pad soft-tokens would be inserted as if valid and the
image span would be wrong. For the fixed-grid processor this is `None` → we fall back to
`encoder_image_masks`, i.e. no behavior change for existing models.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/layers/encoders.py
lokic233: The return went `(embeddings, deep_feats)` → `(embeddings, deep_feats, vision_image_masks)`. Blast radius is
exactly the two internal call sites in `models.py` (`TransformerLinenPure` and `Transformer`), both of which
do `image_masks = vision_image_masks if vision_image_masks is not None else encoder_image_masks` — so every
non-Gemma4-padded-patch caller gets `None` and keeps the prior behavior. There's no external/public API on
this signature. The third element carries the vision-derived validity mask from C, which the merge step
needs and which the encoder is the only thing positioned to compute. If you'd prefer I can wrap it in a small
result dataclass instead of growing the tuple — but it mirrors the existing 2-tuple idiom and is fully
backward compatible as-is. Open to whichever you prefer.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/layers/decoders.py
lokic233: Thanks. To be explicit: this pure-Linen branch is made **fail-closed** for the E2B/E4B multimodal PLE
substitution (raises `NotImplementedError` directing to `enable_nnx=True`, the default + supported path) —
so we never silently diverge on the Linen path. E2E is therefore validated on the **NNX** path:
- the teacher-forced KL parity fixtures in the PR (E2B post_image max_KL 4.29e-05 « the 1.26e-3 gate,
  argmax 1.000), plus the component ablation;
- I've added a Gemma-4-E2B multimodal end-to-end script at
  `tests/end_to_end/tpu/gemma4/e2b/test_gemma4_multimodal.sh`. It follows the existing E2B text scripts +
  the qwen3-vl multimodal pattern: HF→MaxText convert with the vision tower, pre-generate HF golden logits
  for an image+prompt (`generate_hf_golden_logits --image-paths`), then `forward_pass_logit_checker` with
  `use_multimodal=true use_clipped_linears_for_vit=true scan_layers=false` (NNX path) against a KL gate. The
  earlier `e2b/convert_gemma4.sh` explicitly notes multimodal was gated off; this is the counterpart now that
  #4790 enables it.

I've included it in this PR for reviewability, but happy to split it into a follow-up if you'd prefer to
keep the diff smaller. Your call.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/configs/base.yml
lokic233: The vision **tower** is the same (`vision_encoder_block: "gemma4"`, same `image_size_for_vit`); these flags
configure the E2B/E4B image **contract**, not a different encoder, and all default to a no-op so 26B/31B and
every other model are unaffected:
- `use_clipped_linears_for_vit`: E2B/E4B checkpoints ship per-projection activation clip bounds; 26B/31B don't.
- `use_bidirectional_image_attn`: E2B/E4B image spans are **causal** (HF `use_bidirectional_attention=None`);
  26B/31B use bidirectional image attention.
- `ple_pad_substitute_image_rows` / `ple_pad_mode` / `image_placeholder_token_id` / `ple_pad_token_id`: the
  **PLE** path, which only the small models have.

I'll add an explicit "E2B/E4B vs 26B/31B" differences table to the PR description (also addressing your
`types.py` comment).

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/configs/types.py
lokic233: Yes, that references the 26B/31B (bidirectional-image) family. I've added a "Gemma-4 small (E2B/E4B) vs large
(26B/31B)" differences table to the PR description covering: PLE (small-only), causal vs bidirectional image
spans, checkpoint-resident clipped linears (small-only), and the padded-patch/position contract — with the
note that every associated flag defaults to a no-op for the large models.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/optimizers/optimizers.py
lokic233: Done — added an inline comment at the `freeze_clip_bounds` branch calling out that this excludes the
Gemma4-small (E2B/E4B) ViT checkpoint-resident, non-trainable per-projection activation clip bounds from
optimizer updates / weight decay.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

tests/unit/gemma4_clip_linears_test.py
lokic233: Done — consolidated both new test files into `tests/unit/gemma4_small_layers_test.py`:
- the clip-linears pure-jax mechanism/validation tests (clip math, dtype/no-op, NaN/Inf/order hard-fail,
  112/448 counts, freeze-mask polarity, Option-S sentinel invariants), and
- the E2B/E4B multimodal static contract-gate tests.

Removed the standalone `gemma4_clip_linears_test.py` and `gemma4_mm_contract_gate_test.py`. Net −53 lines,
same 26 added test methods, and it also fixed a latent stray `if __name__ == "__main__"` mid-file in the old
clip file that had been shadowing `OptionSInvariantsTest`. Verified: all 26 pass locally on CPU (jax cpu),
and the full file collects cleanly (30 tests). One note — since `gemma4_small_layers_test.py` imports torch at
module scope, the clip mechanism tests now require torch to be *collected* (they don't use it at runtime); the
unit-test CI lane has torch, so this is fine, but if you'd rather keep the pure-jax clip tests torch-free I
can instead fold them into the lighter `gemma4_small_test.py`. Happy either way.

_Posted with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=maxtext4790&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

src/maxtext/configs/models/gemma4-e4b.yml
```diff
@@ -59,3 +60,11 @@ num_attention_heads_for_vit: 12
...
 vision_output_length: 280
 num_position_embeddings_for_vit: 10240
+# E4B image contract (decoder-side): image spans are causal and the per-layer-embedding path
+# substitutes the pad token for image placeholder rows (matches HF modeling_gemma4).
+use_bidirectional_image_attn: False
+ple_pad_substitute_image_rows: True
+ple_pad_mode: "identity"
+image_placeholder_token_id: 258880
+ple_pad_token_id: 0
+packing: false  # Gemma-4 E2B/E4B multimodal does not support packed image spans (see configs/types.py gate)
```
gagika, line 70: will this change default packing behavior for text as well?

src/maxtext/configs/base.yml
```diff
@@ -1244,6 +1244,18 @@ freeze_vision_encoder_params: true
...
+# Gemma-4 vision only: apply the per-projection activation clip bounds carried in the
+# reference checkpoint (self_attn.{q,k,v,o}_proj and mlp.{gate,up,down}_proj each have
+# scalar {input,output}_{min,max}). A prerequisite for Gemma-4 E2B/E4B image parity
+# (necessary but not on its own sufficient); no-op for other vision encoders. Bounds are
+# checkpoint-resident, non-trainable scalars.
+use_clipped_linears_for_vit: false
+# Gemma-4 E2B/E4B decoder image-handling (defaults preserve behavior for other models):
+use_bidirectional_image_attn: false  # E2B/E4B image spans are causal
+ple_pad_substitute_image_rows: false  # substitute pad id for image rows in the per-layer-embedding path (HF gemma4)
+ple_pad_mode: "identity"  # 'identity' (token-id path) or 'both' (also context/embedding path)
```
gagika, line 1256: do we need all these flags?

e.g. for
use_bidirectional_image_attn 
ple_pad_substitute_image_rows
ple_pad_mode

we can just Implement the reference behavior, unless there is a need to change them later.

src/maxtext/optimizers/optimizers.py
```diff
@@ -239,14 +239,36 @@ def get_optimizer(config, learning_rate_schedule, model=None):
...
+        trainable_mask = jax.tree_util.tree_map(lambda a, b: bool(a) and bool(b), trainable_mask, clip_trainable)
+      return jax.tree_util.tree_map(lambda x: "trainable" if x else "frozen", trainable_mask)
+
+    # Use optax.multi_transform so frozen (and clip-bound) params map to a stateless set_to_zero() optimizer
+    # and never allocate momentum/variance slots.
     return optax.multi_transform(
         {"trainable": base_opt, "frozen": optax.set_to_zero()},
-        lambda params: jax.tree_util.tree_map(lambda x: "frozen" if x else "trainable", freeze_mask_fn(params)),
+        _partition,
     )
```
gagika, line 272: `_partition` currently labels parameters **outside** `trainable_parameters_mask` as `"trainable"`: `freeze_mask_fn` was constructed with `match_returns_true=False`. Consequently, this applies `base_opt` to excluded parameters and freezes selected ones, even when vision clipping is disabled.

Please invert the freeze mask, or construct a positive trainable mask, before combining it with the clip-bound mask. The existing `test_get_optimizer_with_trainable_mask` covers the expected behavior; please also add a case with clipping enabled.

src/maxtext/utils/model_creation_utils.py
```diff
@@ -1211,6 +1211,14 @@ def _walk_align(ckpt, model_arr, axes):
             "Please ensure the checkpoint format matches the scan_layers setting."
         )
 
+    # Gemma-4 vision clipped-linears: fail-closed post-load validation BEFORE the first JIT forward/train.
+    # Asserts exactly 112 clip-state modules / 448 finite, correctly-ordered scalar bounds. A missing/NaN/
+    # unordered bound or wrong module count raises here rather than silently degrading to an identity clamp.
+    if bool(getattr(config, "use_clipped_linears_for_vit", False)) and bool(getattr(config, "use_multimodal", False)):
```
gagika, line 1217: nit:
to follow maxtext conversation use `config.use_clipped_linears_for_vit` and `config.use_multimodal` instead

src/maxtext/utils/model_creation_utils.py
```diff
@@ -1211,6 +1211,14 @@ def _walk_align(ckpt, model_arr, axes):
...
+      from maxtext.models import gemma4_vision  # pylint: disable=import-outside-toplevel
+      n_proj, n_bounds = gemma4_vision.validate_all_vision_clip_bounds(model)
+      max_logging.log(f"Gemma-4 vision clip validation OK: {n_proj} clip-state modules, {n_bounds} scalar bounds.")
```
gagika, line 1220: Should you also have similar checks for the path of 
standard train.py startup which uses setup_train_loop → setup_training_state, bypassing from_pretrained. Missing bounds can retain NaN initialization.

tests/unit/gemma4_small_layers_test.py
gagika, line 1: We have excluded this tests from CI/CD (see pytest.ini) because of torch dependancy, so basically it's manual test.
So the newly added tests, if they don't have torch dependency, please separate them so they can run in CI/CD pipeline.

tests/end_to_end/tpu/gemma4/e2b/test_gemma4_multimodal.sh
```diff
@@ -0,0 +1,71 @@
...
+export HF_MODEL=${HF_MODEL:-'path/to/your/hf/gemma-4-E2B-it'}
+
+# Step 1: HF -> MaxText conversion WITH the vision tower (use_multimodal=true).
+python3 -m maxtext.checkpoint_conversion.to_maxtext "${MAXTEXT_CONFIGS_DIR:-${MAXTEXT_REPO_ROOT:-$PWD}/src/maxtext/configs}"/base.yml \
+    model_name=${MODEL_NAME} \
+    hf_access_token=${HF_TOKEN} \
+    --hf_model_path=${HF_MODEL} \
+    base_output_directory=${MODEL_BUCKET}/${MODEL_VARIATION}/converted_mm/${idx} \
+    use_multimodal=true \
+    scan_layers=${USE_SCAN_LAYERS}
```
gagika, line 43: should you also pass?

use_clipped_linears_for_vit=true. as Clipping defaults to false.

gagika (commented):
Thanks looks good, some non blocking comments, please take a look.

PR #2546, opened 2025-10-25
Add muon optimizer

shuningjin:
# Description

This PR integrates [Muon optimizer](https://kellerjordan.github.io/posts/muon/) to MaxText training. We use implementation from [Optax](https://github.com/google-deepmind/optax/blob/main/optax/contrib/_muon.py) with reshaping interface. 

Special note for Muon optimizer: (1) Muon is used together with AdamW. (2) Muon only applies for 2D parameter, and we need reshape 3D/4D to 2D when applicable.

Fix: b/437908829

## 0 Prerequisite

The changes will need optax >=0.2.7. ~~As of 2025-12-18, we only have [optax release](https://github.com/google-deepmind/optax/releases) 0.2.6 from 2025-09. For now, we need manual install from head.~~ [Optax 0.2.7 is released on 2026-02-05](https://github.com/google-deepmind/optax/releases).

Need latest changes in [optax.contrib.muon](https://github.com/google-deepmind/optax/blob/main/optax/contrib/_muon.py). Importantly, we recently modified it to have 
  - reshape interface: https://github.com/google-deepmind/optax/pull/1407 (2025-08-25)
  - consistent rms match: https://github.com/google-deepmind/optax/pull/1435 (2025-10-23)

```
# Install the specific commit
pip install git+https://github.com/google-deepmind/optax@9858013795e22958fc2b318fb59f254bf700b10e
```
or 
```
# uninstall the old version first
pip uninstall optax
# Install the latest 'main' branch 
pip install git+https://github.com/google-deepmind/optax
```

## 1 Code change

Basic integration:
- `src/MaxText/configs/base.yml`: add muon config, while reusing adamw config
-  `src/MaxText/optimizers.py`: use optax.contrib.muon

Reshape interface (see Sec 3)
- `src/MaxText/muon_utils.py`: 
  - we use `maxtext.get_abstract_param` to get abstract structure without materializing the weight. This does not have memory or FLOP overhead.
  - from the structure, we use `get_transform_tree` to get the muon dimension number.
  - To review the muon dimension number of a model: e.g., `python3 -m MaxText.muon_utils qwen3-4b True`. This is helpful for integrating Muon with more models (see Sec 3.3)
- pass in model for `optimizers.get_optimizer` (`train_utils.py`, `train_compile.py`, `sft_trainer.py`)
- `muon_test.py`: unit test to ensure generated Muon dimension numbers match the hardcoded reference

## 2 User guide

**Model Specific Support: Reshape**
[...]

PR #4843, opened 2026-08-11
Fix muon moe dimension numbers

### LINKED ISSUE #4798: Muon: MoE expert weights may get the wrong dimension numbers on qwen3 / gpt-oss
### Bug report

`muon_utils.transform_logic` only applies the per-expert mapping when the MoE module is
named `MoeBlock_0`:

```python
if "MoeBlock_0" in path:
    # exclude gate
    if _is_path_contain_any(("wi_0", "wi_1", "wo"), path):
      return mdn((-2,), (-1,))
```

`_is_path_contain_any` compares whole path segments, and the inner `if` has no `else`, so
a non-matching path falls through to the standard `mdn((0,), (-1,))` at the end of the
function. Where `RoutedMoE` / `RoutedAndSharedMoE` is actually attached:

| model | attribute | contains `MoeBlock_0` |
|---|---|---|
| `mixtral.py:108`, `envy.py:120` | `MoeBlock_0` | yes |
| `moe.py:3195` (`RoutedAndSharedMoE`: deepseek / deepseek4 / llama4 / gemma4) | `MoeBlock_0` | yes |
| `qwen3.py:1070` (qwen3-next) | `routed_experts` | no |
| `qwen3.py:1547` (qwen3 MoE) | `moe_block` | no |
| `qwen3_custom.py:136` | `moe_block` | no |
| `gpt_oss.py:121` | `GptOssMlp` | no |

The checkpoint mappings confirm the real paths: `param_mapping.py:1471-1479`
(qwen3-next) uses `mlp-routed_experts-wi_0` / `-wi_1` / `-wo`, and
`param_mapping.py:1832-1843` (gpt-oss) uses `GptOssMlp-wi_0` / `-wo`. No `MoeBlock_0`
segment in either.

If I'm reading this right, on those four the expert stack `[E, (L,) in, out]` gets
`reduction_axis=0`, i.e. the expert axis, so Newton-Schulz orthogonalizes *across*
experts rather than within each one, and `_shape_factor` is computed from the same axes.
Nothing raises, because the reshape is still valid.

It would also explain why this isn't caught: `test_model_integration` covers deepseek2/3/4,
kimi, llama2/3, gemma3 and qwen3-0.6b — every MoE model in that list routes through
`RoutedAndSharedMoE`'s nested `MoeBlock_0`, and qwen3-0.6b is dense.

Two more cases that look like the same fallthrough, and apply even when the module name
does match:

- `wi`, the fused gate+up kernel created when `prefuse_moe_weights=True` (`moe.py:537`),
  is not in the `("wi_0", "wi_1", "wo")` tuple.
- `wi_0_bias` / `wi_1_bias` / `wo_bias`, created when `mlp_bias=True` (`moe.py:603`, used
  by gpt-oss-20b / 120b), are not caught by the `segment == "bias"` exclusion because of
  the prefix, so they appear to be orthogonalized instead of going to the AdamW branch.

For what it's worth, #4779 appears to expect the same behaviour I do — it adds
`test_qwen3_next_moe_routed_experts` asserting
`("decoder", "mlp", "routed_experts", "wi_0") -> mdn((-2,), (-1,))` — but does not change
the matcher.

## Questions

1. Is this a real bug, or is Muon not expected to be used with these models yet?
2. If it's real, would matching on the leaf segment be acceptable —
   `path[-1] in ("wi", "wi_0", "wi_1", "wo")` — instead of a per-model list of module
   names? The reference trees in `optimizers_test.py` already encode that distinction —
   the expert `wi_0` is a leaf (`mdn((-2,), (-1,))`) while a dense `wi_0` has a `kernel`
   child (`mdn((0,), (-1,))`) — because `MlpBlock` reuses those names as `DenseGeneral`
   submodules (`linears.py:493`). The gpt-oss mapping shows both side by side:
   `GptOssMlp-gate-kernel` next to `GptOssMlp-wi_0`.

Happy to send a PR — I have the change and unit tests ready.

### Logs/Output

_No response_

### Environment Information

_No response_

### Additional Context

_No response_

parsley9877:
# Description

This PR updates `maxtext.utils.muon_utils.transform_logic` to assign `MuonDimensionNumbers` to MoE expert weights across (Qwen 3, Qwen 3-next, and GPT-OSS) and ensures all bias variants are correctly excluded from Muon orthogonalization.

### Problem
1. **Missing MoE block names**: `transform_logic` previously checked `if "MoeBlock_0" in path:` to detect expert weights. However, newer MoE architectures use different module names:
   - `routed_experts` (Qwen3-next in `qwen3.py`)
   - `moe_block` (Qwen3 MoE in `qwen3.py` / `qwen3_custom.py`)
   - `GptOssMlp` (GPT-OSS in `gpt_oss.py`)
   Because these module names did not contain `"MoeBlock_0"`, their expert weight tensors fell through to the default mapping `mdn((0,), (-1,))` instead of `mdn((-2,), (-1,))`, leading to incorrect orthogonalization across the expert axis.
2. **Prefused MoE kernels**: When `prefuse_moe_weights=True`, the fused gate+up kernel is named `wi`, which was missing from the recognized expert weight kernels `("wi_0", "wi_1", "wo")`.
3. **Bias variants**: `transform_logic` checked `segment == "bias"`, which missed compound bias names such as `wi_0_bias`, `wo_bias` (used in GPT-OSS with `mlp_bias=True`), and `mlp_bias`.

### Solution
- Expanded MoE block container matching in `transform_logic` to recognize `("MoeBlock_0", "routed_experts", "moe_block", "GptOssMlp")`.
- Added `"wi"` to the expert weight kernel tuple `("wi", "wi_0", "wi_1", "wo")` returning `mdn((-2,), (-1,))`.
- Added `"bias"` to the substring exclusion tuple so all bias variations (`wi_0_bias`, `wo_bias`, etc.) return `None` and fall back to AdamW as intended.

FIXES: #4798

# Tests

Added unit test coverage in `tests/unit/muon_utils_test.py`:
- `test_bias_variant_is_excluded`: Verifies `wi_0_bias`, `wo_bias`, and `mlp_bias` return `None`.
- `test_qwen3_next_moe_routed_experts`: Verifies `routed_experts` expert kernels (`wi_0`, `wi_1`, `wo`) return `mdn((-2,), (-1,))`.
- `test_qwen3_moe_block`: Verifies `moe_block` expert kernels return `mdn((-2,), (-1,))`.
- `test_gpt_oss_mlp_moe`: Verifies `GptOssMlp` expert kernels return `mdn((-2,), (-1,))`.
- `test_moe_prefused_wi_uses_last_two_axes`: Verifies prefused `wi` kernels return `mdn((-2,), (-1,))`.
- `test_moe_gate_falls_through_to_standard`: Verifies gate routing kernels fall through to standard mapping.

Ran test suite locally:
```bash
python3 -m unittest tests/unit/muon_utils_test.py
```

# Checklist

### DIFF
--- a/tests/unit/muon_utils_test.py
+++ b/tests/unit/muon_utils_test.py
@@ -53,6 +53,14 @@ class TestTransformLogic(unittest.TestCase):
   def test_bias_is_excluded(self):
     self.assertIsNone(muon_utils.transform_logic(("decoder", "dense", "bias")))
 
+  def test_bias_variant_is_excluded(self):
+    # Bias names (e.g., in MoE with mlp_bias=True) are excluded across all expert projections.
+    self.assertIsNone(muon_utils.transform_logic(("decoder", "moe_block", "wi_0_bias")))
+    self.assertIsNone(muon_utils.transform_logic(("decoder", "GptOssMlp", "wi_0_bias")))
+    self.assertIsNone(muon_utils.transform_logic(("decoder", "GptOssMlp", "wi_1_bias")))
+    self.assertIsNone(muon_utils.transform_logic(("decoder", "GptOssMlp", "wo_bias")))
+    self.assertIsNone(muon_utils.transform_logic(("decoder", "mlp", "mlp_bias")))
+
   def test_embedding_is_excluded(self):
     self.assertIsNone(muon_utils.transform_logic(("token_embedder", "embedding")))
 
@@ -69,9 +77,27 @@ class TestTransformLogic(unittest.TestCase):
   def test_moe_wo_uses_last_two_axes(self):
     self.assertEqual(muon_utils.transform_logic(("decoder", "MoeBlock_0", "wo")), mdn((-2,), (-1,)))
 
+  def test_moe_prefused_wi_uses_last_two_axes(self):
+    self.assertEqual(muon_utils.transform_logic(("decoder", "MoeBlock_0", "wi")), mdn((-2,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "routed_experts", "wi")), mdn((-2,), (-1,)))
+
+  def test_qwen3_next_moe_routed_experts(self):
+    self.assertEqual(muon_utils.transform_logic(("decoder", "mlp", "routed_experts", "wi_0")), mdn((-2,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "mlp", "routed_experts", "wi_1")), mdn((-2,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "mlp", "routed_experts", "wo")), mdn((-2,), (-1,)))
+
+  def test_qwen3_moe_block(self):
+    self.assertEqual(muon_utils.transform_logic(("decoder", "moe_block", "wi_0")), mdn((-2,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "moe_block", "wo")), mdn((-2,), (-1,)))
+
+  def test_gpt_oss_mlp_moe(self):
+    self.assertEqual(muon_utils.transform_logic(("decoder", "GptOssMlp", "wi_0")), mdn((-2,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "GptOssMlp", "wo")), mdn((-2,), (-1,)))
+
   def test_moe_gate_falls_through_to_standard(self):
-    # 'gate' is inside MoeBlock_0 but not one of (wi_0, wi_1, wo) → standard.
+    # 'gate' is inside MoeBlock_0 but not one of (wi, wi_0, wi_1, wo) → standard.
     self.assertEqual(muon_utils.transform_logic(("decoder", "MoeBlock_0", "gate", "kernel")), mdn((0,), (-1,)))
+    self.assertEqual(muon_utils.transform_logic(("decoder", "routed_experts", "gate", "kernel")), mdn((0,), (-1,)))
 
   # --- 2.2 Self-attention ---
   def test_self_attention_out_projection(self):
@@ -118,6 +144,18 @@ class TestTransformLogic(unittest.TestCase):
     # o_a_proj projects with reduction on in_features_per_group (-2)
     # and output on out_features_per_group (-1)
     self.assertEqual(muon_utils.transform_logic(("decoder", "self_attention", "o_a_proj")), mdn((-2,), (-1,)))
+
+  def test_deepseek_v4_position_bias_is_standard_weight(self):
+    # position_bias in DeepSeek V4 compressed attention is a 2D weight matrix (not a 1D bias vector),
+    # so it receives standard Muon dimension numbers mdn((0,), (-1,)).
+    self.assertEqual(
+        muon_utils.transform_logic(("decoder", "self_attention", "csa_compressor", "position_bias")),
+        mdn((0,), (-1,)),
+    )
+    self.assertEqual(
+        muon_utils.transform_logic(("decoder", "self_attention", "hca_compressor", "position_bias")),
+        mdn((0,), (-1,)),
+    )
 
 
 class TestGetTransformTree(unittest.TestCase):
--- a/tests/unit/optimizers_test.py
+++ b/tests/unit/optimizers_test.py
@@ -219,6 +219,177 @@ QWEN3_DIMENSION_NUMBER = {
 }
 
 
+# qwen3 MoE (e.g. qwen3-30b-a3b)
+QWEN3_MOE_DIMENSION_NUMBER = {
+    "params": {
+        "decoder": {
+            "decoder_norm": {"scale": None},
+            "layers": {
+                "moe_block": {
+                    "gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+                    "wi_0": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                    "wi_1": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                    "wo": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                },
+                "post_self_attention_layer_norm": {"scale": None},
+                "pre_self_attention_layer_norm": {"scale": None},
+                "self_attention": {
+                    "query": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "key": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "value": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "out": {"kernel": mdn(reduction_axis=(0, -2), output_axis=(-1,))},
+                    "key_norm": {"scale": None},
+                    "query_norm": {"scale": None},
+                },
+            },
+            "logits_dense": {"kernel": None},
+        },
+        "token_embedder": {"embedding": None},
+    }
+}
+
+
+# qwen3 custom MoE (e.g. qwen3-custom-30b-a3b)
+QWEN3_CUSTOM_MOE_DIMENSION_NUMBER = {
+    "params": {
+        "decoder": {
+            "decoder_norm": {"scale": None},
+            "layers": {
+                "latent_norm": {"scale": None},
+                "layer_up_projection": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+                "moe_block": {
+                    "gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+                    "wi_0": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                    "wi_1": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                    "wo": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+                },
+                "post_self_attention_layer_norm": {"scale": None},
+                "pre_self_attention_layer_norm": {"scale": None},
+                "self_attention": {
+                    "query": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "key": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "value": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+                    "out": {"kernel": mdn(reduction_axis=(0, -2), output_axis=(-1,))},
+                    "key_norm": {"scale": None},
+                    "query_norm": {"scale": None},
+                },
+            },
+            "logits_dense": {"kernel": None},
+        },
+        "token_embedder": {"embedding": None},
+    }
+}
+
+
+# qwen3-next (e.g. qwen3-next-80b-a3b)
+_QWEN3_NEXT_MLP = {
+    "routed_experts": {
+        "gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wi_0": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+        "wi_1": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+        "wo": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+    },
+    "shared_expert": {
+        "wi_0": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wi_1": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wo": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    },
+    "shared_expert_gate": {"kernel": None},
+}
+
+_QWEN3_NEXT_GDN_ATTENTION = {
+    "A_log": None,
+    "conv1d": {"kernel": None},
+    "dt_bias": None,
+    "in_proj_ba": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    "in_proj_qkvz": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    "norm": {"rms_norm": {"scale": None}},
+    "out_proj": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+}
+
+_QWEN3_NEXT_FULL_ATTENTION = {
+    "attention": {
+        "query": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+        "key": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+        "value": {"kernel": mdn(reduction_axis=(0,), output_axis=(-2, -1))},
+        "out": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "key_norm": {"scale": None},
+        "query_norm": {"scale": None},
+    },
+}
+
+_QWEN3_NEXT_GDN_LAYER = {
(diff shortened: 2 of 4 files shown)


entrpn (approved):

tests/unit/muon_utils_test.py
```diff
@@ -53,6 +53,11 @@ def test_scale_is_excluded(self):
   def test_bias_is_excluded(self):
     self.assertIsNone(muon_utils.transform_logic(("decoder", "dense", "bias")))
 
+  def test_bias_variant_is_excluded(self):
```
RissyRan, line 56: Could you help explain a little bit why those bias to be excluded? or add a comment there?

Also, for a model/config, are you consider all bias? for instance, wondering why GptOss only excludes wo_bias?

RissyRan (commented):
LGTM at high level, and trying to understand a little bit on tests.

FYI @shuningjin who may have most context of Muon conv

tests/unit/muon_utils_test.py
parsley9877: I am under the impression that all kind of biases need to be excluded. These unittests are just some examples of bias tensors.

I think all bias needs to be excluded. GPTOss is just an example to test the code.

shuningjin:
Thanks for adding the fix! To support more models, I would suggest checking Section 3 in https://github.com/AI-Hypercomputer/maxtext/pull/2546. 

Files to update:
https://github.com/AI-Hypercomputer/maxtext/blob/09bc4bc10007a143bc418128b0bbe9ab4686d951/src/maxtext/configs/types.py#L3815-L3825
https://github.com/AI-Hypercomputer/maxtext/blob/c00b247a327c135ccb04eed8913f67a53adfd705/tests/unit/optimizers_test.py#L377

parsley9877:
> Thanks for adding the fix! To support more models, I would suggest checking Section 3 in #2546.
> 
> Files to update:
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/09bc4bc10007a143bc418128b0bbe9ab4686d951/src/maxtext/configs/types.py#L3815-L3825
> 
> 
> https://github.com/AI-Hypercomputer/maxtext/blob/c00b247a327c135ccb04eed8913f67a53adfd705/tests/unit/optimizers_test.py#L377

@shuningjin Thank you for your comment!

1. **`types.py`:** Added `QWEN3_MOE`, `QWEN3_CUSTOM_MOE`, `QWEN3_NEXT`, and `GPT_OSS` to the `opt_type == "muon"` decoder block whitelist.
2. **`optimizers_test.py`:** Added reference dimension number structures and integration test cases in `MuonDimensionTest` for `qwen3-30b-a3b`, `qwen3-custom-30b-a3b`, `qwen3-next-80b-a3b`, and `gpt-oss-20b`.

tests/unit/optimizers_test.py
```diff
@@ -219,6 +219,177 @@
...
+    "shared_expert": {
+        "wi_0": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wi_1": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wo": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    },
+    "shared_expert_gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+}
+
+_QWEN3_NEXT_GDN_ATTENTION = {
+    "A_log": mdn(reduction_axis=(0,), output_axis=(-1,)),
```
shuningjin: `A_log` is a scalar, mdn should be `None`
- scanned shape`'A_log': {'shape': (32, 12)`, 12 is the layer axis
- code: https://github.com/AI-Hypercomputer/maxtext/blob/c5a148fa92dcab3a0ee807c20599209c2f1d7803/src/maxtext/models/qwen3.py#L552

tests/unit/optimizers_test.py
```diff
@@ -219,6 +219,177 @@
...
+    "norm": {"rms_norm": {"scale": None}},
+    "out_proj": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+}
+
+_QWEN3_NEXT_FULL_ATTENTION = {
+    "attention": {
+        "query": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "key": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "value": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "out": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
```
shuningjin: For the full attention part, I think Q/K/V should be the same as other models, e.g., llama2
https://github.com/AI-Hypercomputer/maxtext/blob/ea3482f4e8e679bc87676fdc38c4480e16de8ed0/tests/unit/optimizers_test.py#L177-L181

Remark: 
- `"out": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},` is correct here
- qwen3 next is different from other model due to additional flatenning
https://github.com/AI-Hypercomputer/maxtext/blob/c5a148fa92dcab3a0ee807c20599209c2f1d7803/src/maxtext/layers/attentions.py#L808-L835

tests/unit/optimizers_test.py
```diff
@@ -219,6 +219,177 @@
...
+
+
+# gpt-oss (e.g. gpt-oss-20b)
+_GPT_OSS_ATTENTION = {
+    "query": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,)), "bias": None},
+    "key": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,)), "bias": None},
+    "value": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,)), "bias": None},
+    "out": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,)), "bias": None},
+    "sinks": None,
+}
```
shuningjin, line 359: For the gpt oss attention part, I think Q/K/V/out should be the same as other models, e.g., llama2
https://github.com/AI-Hypercomputer/maxtext/blob/ea3482f4e8e679bc87676fdc38c4480e16de8ed0/tests/unit/optimizers_test.py#L177-L181

shuningjin (commented):
Thanks for the update! Could you double check the reference dimension number in `optimizers_test.py`?
- Based on my initial inspection, some parts might need revision (e.g., A_log, attention). Note that my check is not exhaustive.
- My suggested [flow](https://screenshot-v2.corp.google.com/0rmujeht0bfm0) is: 
  - auto-generate a mapping with existing logic. e.g., `python3 -m maxtext.utils.muon_utils gpt-oss-20b True`
  - ask gemini for a check, 
  - and check the code (as in step4), esp. for complex model like qwen3-next. e.g., see inline comments

parsley9877:
> Thanks for the update! Could you double check the reference dimension number in `optimizers_test.py`?
> 
> * Based on my initial inspection, some parts might need revision (e.g., A_log, attention). Note that my check is not exhaustive.
> * My suggested [flow](https://screenshot-v2.corp.google.com/0rmujeht0bfm0) is:
>   
>   * auto-generate a mapping with existing logic. e.g., `python3 -m maxtext.utils.muon_utils gpt-oss-20b True`
>   * ask gemini for a check,
>   * and check the code (as in step4), esp. for complex model like qwen3-next. e.g., see inline comments

Thank you for the comment @shuningjin 

I followed the flow and verified the auto-generated mappings across models.

tests/unit/optimizers_test.py
parsley9877: Added "A_log" to the exclusion list in transform_logic step 1 so it maps to None, and updated the reference in _QWEN3_NEXT_GDN_ATTENTION.

tests/unit/optimizers_test.py
parsley9877: Updated transform_logic to handle the attention path in Qwen3-Next, setting query, key, and value to mdn(reduction_axis=(0,), output_axis=(-2, -1)) to match the standard attention models.

tests/unit/optimizers_test.py
parsley9877: Updated transform_logic to recognize GptOssAttention, setting query/key/value to mdn(reduction_axis=(0,), output_axis=(-2, -1)) and out to mdn(reduction_axis=(0, -2), output_axis=(-1,)), matching llama2.

RissyRan (approved):

tests/unit/optimizers_test.py
```diff
@@ -219,6 +219,177 @@
...
+        "wi_0": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wi_1": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wo": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    },
+    "shared_expert_gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+}
+
+_QWEN3_NEXT_GDN_ATTENTION = {
+    "A_log": None,
+    "conv1d": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
```
shuningjin: `attention.conv1d.kernel` (Shape: (4, 12, 1, 8192)) $\rightarrow$ Should be None (AdamW)
- This is a 1D depthwise convolution kernel across sequence tokens (filter size 4, 8192 channels).
- Muon’s Newton-Schulz orthogonalization is designed for dense 2D linear weight transformations. Applying matrix orthogonalization to a depthwise 1D filter of shape $(4, 8192)$ treats temporal filter taps as orthogonal matrix rows, which causes degenerate scaling and training instability. Depthwise conv kernels should be handled by AdamW.

tests/unit/optimizers_test.py
```diff
@@ -219,6 +219,177 @@
...
+        "wi_0": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+        "wi_1": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+        "wo": mdn(reduction_axis=(-2,), output_axis=(-1,)),
+    },
+    "shared_expert": {
+        "wi_0": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wi_1": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+        "wo": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
+    },
+    "shared_expert_gate": {"kernel": mdn(reduction_axis=(0,), output_axis=(-1,))},
```
shuningjin: `mlp.shared_expert_gate.kernel` (Shape: (2048, 12, 1)) $\rightarrow$ Should be None (AdamW)
- The shared expert gating projection outputs a single scalar value (output_dim = 1).
- Applying Muon to a rank-1 $(2048, 1)$ matrix reduces Newton-Schulz iterations to simple vector normalization, which interferes with proper gate logit scaling. Vector projections and scalar gates should use AdamW.

shuningjin (commented):
As a sanity check, I run these generated dimension number. Commands
```
python3 -m maxtext.utils.muon_utils qwen3-next-80b-a3b True
python3 -m maxtext.utils.muon_utils gpt-oss-20b True
python3 -m maxtext.utils.muon_utils qwen3-30b-a3b True
python3 -m maxtext.utils.muon_utils qwen3-custom-30b-a3b True
```

**Gemini-assisted Check Result**
- Please see **full report**: [doc](https://docs.google.com/document/d/1cUodi0CiqLiuHkalb_vYx7qHr4hMQ01Qmq4hCQHjma4/edit?usp=sharing)
- `gpt-oss-20b, qwen3-30b-a3b, qwen3-custom-30b-a3b` looks fine
- `qwen3-next-80b-a3b` **Issues** 
  - `attention.conv1d.kernel` (Shape: (4, 12, 1, 8192)) $\rightarrow$ Should be None (AdamW). This is a 1D depthwise convolution kernel across sequence tokens (filter size 4, 8192 channels).
  - `mlp.shared_expert_gate.kernel` (Shape: (2048, 12, 1)) $\rightarrow$ Should be None (AdamW). The shared expert gating projection outputs a single scalar value (output_dim = 1).
- **Recommended Adjustment**: In `maxtext/utils/muon_utils.py`, add filter conditions so that:
  - depthwise 1D convolutions (e.g., 'conv1d') return None.
  - Projections with an output dimension of 1 (e.g., shape[-1] == 1 or matching 'shared_expert_gate') return None.


**Remark**: The [Muon Blog](https://kellerjordan.github.io/posts/muon/) mentioned that "Muon can be used for 4D convolutional parameters by flattening their last three dimensions ([like so](https://github.com/KellerJordan/cifar10-airbench/blob/0e6f9614572d7e8e3c259905aebc7196f91d5d79/research/clean_muon.py#L95)). Keller Jordan’s technique of flattening convolutional tensors applies to standard (channel-mixing) convolutions, but behaves fundamentally differently on depthwise 1D convolutions



---
As a next step, please consider the fix and run the check again for qwen3-next-80b-a3b.

tests/unit/optimizers_test.py
parsley9877: Done. Added `conv1d` to the exclusion list in `transform_logic` so it maps to `None`, and updated the test reference.

tests/unit/optimizers_test.py
parsley9877: Done. Added `shared_expert_gate` to the exclusion list in `transform_logic` so it maps to `None`, and updated the test reference.

parsley9877:
> As a sanity check, I run these generated dimension number. Commands
> 
> ```
> python3 -m maxtext.utils.muon_utils qwen3-next-80b-a3b True
> python3 -m maxtext.utils.muon_utils gpt-oss-20b True
> python3 -m maxtext.utils.muon_utils qwen3-30b-a3b True
> python3 -m maxtext.utils.muon_utils qwen3-custom-30b-a3b True
> ```
> 
> **Gemini-assisted Check Result**
> 
> * Please see **full report**: [doc](https://docs.google.com/document/d/1cUodi0CiqLiuHkalb_vYx7qHr4hMQ01Qmq4hCQHjma4/edit?usp=sharing)
> * `gpt-oss-20b, qwen3-30b-a3b, qwen3-custom-30b-a3b` looks fine
> * `qwen3-next-80b-a3b` **Issues**
>   
>   * `attention.conv1d.kernel` (Shape: (4, 12, 1, 8192)) 
>       →
>      Should be None (AdamW). This is a 1D depthwise convolution kernel across sequence tokens (filter size 4, 8192 channels).
>   * `mlp.shared_expert_gate.kernel` (Shape: (2048, 12, 1)) 
>       →
>      Should be None (AdamW). The shared expert gating projection outputs a single scalar value (output_dim = 1).
> * **Recommended Adjustment**: In `maxtext/utils/muon_utils.py`, add filter conditions so that:
>   
>   * depthwise 1D convolutions (e.g., 'conv1d') return None.
>   * Projections with an output dimension of 1 (e.g., shape[-1] == 1 or matching 'shared_expert_gate') return None.
> 
> **Remark**: The [Muon Blog](https://kellerjordan.github.io/posts/muon/) mentioned that "Muon can be used for 4D convolutional parameters by flattening their last three dimensions ([like so](https://github.com/KellerJordan/cifar10-airbench/blob/0e6f9614572d7e8e3c259905aebc7196f91d5d79/research/clean_muon.py#L95)). Keller Jordan’s technique of flattening convolutional tensors applies to standard (channel-mixing) convolutions, but behaves fundamentally differently on depthwise 1D convolutions
> 
> As a next step, please consider the fix and run the check again for qwen3-next-80b-a3b.

Thank you for the check and suggestions @shuningjin!

- Added both `conv1d`and `shared_expert_gate` to the exclusion list in `transform_logic`.
- Updated the reference dimension numbers in `tests/unit/optimizers_test.py`.
- Re-ran `python3 -m maxtext.utils.muon_utils qwen3-next-80b-a3b True` to verify the generated structure and confirmed unit tests pass.

shuningjin (approved):
Great work on expanding model coverage! LGTM

PR #4836, opened 2026-08-11
Add TPU USP context parallelism

huytransformer:
# Description

This PR introduces `context_parallel_strategy=usp` ([USP](https://arxiv.org/abs/2405.07719), Ulysses over ring). Follow up to #4687.

Currently does not support load balancing + sequence packing.

# Tests

`python3 -m pytest tests/unit/configs_value_test.py tests/unit/usp_attention_test.py tests/unit/usp_collective_test.py`

Passed. 

`python3 -m pytest tests/unit/attention_test.py -k usp`

3 passed.

## Performance

llama3-8b, v5p (64 chips), CP64, bf16, synthetic data, global batch 1, no load-balancing

Reporting median step time (s)

|  | 64K | 128K | 256K |
|---|---|---|---|
| all-gather (CP64) | 1.31 | 4.03 | 14.33 |
| ring (CP64) | 1.68 | 4.63 | 15.35 |
| usp (16x4) | 1.31 | 4.02 | 13.92 |
| usp (8x8) | 1.44 | 4.05 | 13.65 |
| usp (32x2) | 1.65 | 4.52 | 14.72 |

Example repro command:

```
python3 -m maxtext.trainers.pre_train.train src/maxtext/configs/base.yml model_name=llama3-8b dataset_type=synthetic enable_checkpointing=False steps=20 per_device_batch_size=0.015625 packing=False attention=flash use_tokamax_splash=True use_jax_splash=False allow_split_physical_axes=True max_target_length=65536 context_parallel_strategy=usp ici_context_parallelism=16 ici_context_ulysses_parallelism=4 context_parallel_load_balance=False run_name=usp16x4_65536 base_output_directory=<output dir>
```

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #4887, opened 2026-08-14
Support packing for TPU USP attention

huytransformer:
# Description

Corollary follow up of PR #4836. This PR enables sequence packing when using TPU USP attention by lifting the config ban and adding tests.

# Tests

`python3 -m pytest tests/unit/attention_test.py -k usp`

5 passed on v4 

Verus all_gather (identical batches, shuffling off): per-step loss similar within 0.002 over 10 steps.

`python3 -m pytest tests/unit/configs_value_test.py`

Passed.

# Checklist


gobbleturk (approved):
From this PR it looks like packing was already supported, by we would throw an assertion error anyway - e.g. I don't see logic in attention.py or relevant files that needed to change to support this - is this understanding correct?

NuojCheng (approved):

huytransformer:
> From this PR it looks like packing was already supported, by we would throw an assertion error anyway - e.g. I don't see logic in attention.py or relevant files that needed to change to support this - is this understanding correct?

Yes this is true for my packing prs since we already passed segment ids to the splash kernel but never tested

PR #4776, opened 2026-08-07
[maxtext] Add block-diffusion pre-training

ethannnnnn:
## Motivation

With block-diffusion primitives and block-causal attention available, MaxText
needs a focused pre-training integration before any completion fine-tuning,
rollout, distillation, replay, or RL behavior is introduced.

## Dependencies

This change follows and depends on:

- #4737 for model-independent corruption and target alignment; and
- #4743 for opt-in block-causal attention.

It has no Tunix dependency.

## Scope

This PR adds only:

- an opt-in `training_objective=block_diffusion` configuration;
- Hugging Face text pre-training corruption with separate clean targets,
  corruption masks, and target-loss masks;
- target-aligned loss handling for Linen and NNX models;
- explicit compatibility validation for the initial pre-training surface; and
- finite zero loss and gradients when an accumulated batch has no supervised
  positions; and
- a CI guard that skips secret-dependent notebook execution for fork and
  Dependabot pull requests, where GitHub withholds `HF_TOKEN`.

It does not add SFT, DPO, rollout, serving, distillation, replay, or RL.

## Training contract

The supported model contracts are:

- `same_position` logits with an `all_masked` canvas; and
- `shifted` logits with a `seed_and_mask` canvas.

Each nonempty eligible block keeps at least one supervised position. Clean
targets and segmentation metadata remain separate from the corrupted inputs.
Loss is normalized over `targets_loss_mask` positions only. Packing, SFT,
DPO, MTP, vocabulary tiling, audio, multimodal input, and non-Hugging-Face
pipelines fail validation for this initial scope.

## Compatibility

`training_objective=causal_lm` remains the default. Its next-token data
transform, segmentation-based loss mask, and autoregressive attention path are
unchanged.

The shared gradient-accumulation helper guards a zero total-weight denominator
only for the block-diffusion objective. The causal branch retains its original
operations, including the upstream Tunix accumulation-step divisor. Focused
regression tests verify both accumulation modes against the direct full-batch
loss and gradients.

## Tests

The audited current-main replay passed:

```text
270 passed, 58 skipped, 3 deselected, 92 subtests passed
```

The gate covers block-diffusion primitives, dense and TPU mask construction,
configuration, Hugging Face objective selection, padding and corruption,
Linen and NNX loss masks, causal-loss fallback, shaped batches, context
parallel metadata, and positive- and zero-weight gradient accumulation.

Three pre-existing Hugging Face iterator integration tests are deselected in
[...]

PR #4891, opened 2026-08-14
[maxtext] Add block-diffusion CFT and SFT

ethannnnnn:
## Motivation

#4776 added the opt-in block-diffusion pre-training path. MaxText now needs a
focused text CFT and completion-only SFT integration before rollout,
distillation, replay, or RL behavior is introduced.

## Dependencies

This change follows and depends on:

- #4776 for MaxText block-diffusion pre-training; and
- google/tunix#1891 for target-aligned diffusion SFT batches and weighted
  trainer loss.

Both dependencies are merged. Current MaxText pins upstream Tunix commit
`a8d70582f1e2f1fb65973210989e0e148b5ef7ad`, a descendant of the Copybara
integration of google/tunix#1891 (`bd1b0f47606a8bfb5158fe17084503c16bb4a62e`).

## Scope

This PR adds only:

- opt-in Hugging Face text CFT and completion-only SFT for
  `training_objective=block_diffusion`;
- clean targets, role-derived completion eligibility, corruption masks, and
  explicit target-loss weights through the SFT pipeline;
- a MaxText-to-Tunix adapter for target-aligned diffusion batches and logits;
- weighted training and preaveraged evaluation metrics;
- safe trainer routing for diffusion gradient accumulation and internal neural
  network metrics; and
- fail-closed validation for prompt/completion layouts that would expose a
  later prompt to an earlier completion inside one bidirectional diffusion
  block.

It does not add DPO, multimodal or audio training, rollout, serving,
distillation, replay, or RL. Block-diffusion training continues to require an
unpacked Hugging Face text pipeline.

## Training contract

The pre-training logit and canvas contracts remain unchanged:

- `same_position` logits with an `all_masked` canvas; and
- `shifted` logits with a `seed_and_mask` canvas.

CFT corrupts and supervises all valid target positions. Completion-only SFT
keeps prompt tokens as clean context and corrupts and supervises completion
positions only. Because attention is bidirectional within a diffusion block,
each block must contain a prompt prefix followed by a completion suffix. A
later prompt may start at the next block boundary; an assistant-to-user
transition inside one block fails with an actionable error instead of leaking
future prompt text.

The Tunix adapter requires non-None position, segmentation, completion,
corruption, and loss-mask fields. It verifies mask shapes and ownership before
constructing a weighted `DiffusionTokenBatch`. Diffusion evaluation already
returns a globally weighted mean, so the SFT hooks do not divide it by the
evaluation-step count again.

Tunix NNX graph caching remains enabled for normal dense training and disabled
for MoE. It is also disabled for block diffusion when internal neural network
metrics are recorded, because those intermediate graph nodes mutate the cached
graph structure.

## Compatibility

`training_objective=causal_lm` remains the default. Its shifted targets,
MaxText SFT trainer, completion masking, metrics, and dense graph-cache path
are unchanged. Focused tests cover both objective routes and the existing
causal fallback behavior.

## Tests

The current-main replay passed the full CPU gate against the exact upstream
Tunix #1891 integration:

```text
399 passed, 98 skipped, 3 deselected, 219 subtests passed
```

The 98 skips are platform-specific. The three pre-existing Hugging Face
iterator integration tests are deselected locally because they require live
`datasets.load_dataset` access; the changed objective-selection tests in that
module pass.

The gate covers block-diffusion primitives and attention, CFT/SFT corruption,
role masks, prompt-leak rejection, Tunix batch and logits adapters, weighted
metrics, cache routing, configuration, causal regressions, and positive- and
zero-weight gradient accumulation. The Tunix adapter has 100% statement and
branch coverage.

Pyink, repository-equivalent Pylint 3.3.8 (`10.00/10`), Python compilation,
and `git diff --check` also pass. The guarded publisher reruns these
checks on freshly fetched MaxText and Tunix upstream commits before it may
push.

## Checklist

- [x] I performed a self-review of the focused CFT/SFT boundary.
- [x] Existing causal SFT behavior remains the default and has regression coverage.
- [x] Unsupported training modes and unsafe completion layouts fail closed.
- [x] No rollout, distillation, replay, or RL implementation is included.

Design document:
https://docs.google.com/document/d/1N7KcCoAIErB2CV9EJ2G1u_mdN-AQgqwNUYSvM0mtqMI/edit

### DIFF
--- a/src/maxtext/input_pipeline/hf_data_processing.py
+++ b/src/maxtext/input_pipeline/hf_data_processing.py
@@ -48,17 +48,19 @@ def _get_training_objective_transform(
     shift: bool,
     use_dpo: bool,
     use_sft: bool,
+    completion_only: bool,
     packing: bool,
     pad_id: int,
     bos_token_id: int | None,
 ) -> input_pipeline_utils.ShiftData | input_pipeline_utils.BlockDiffusionCorruption | None:
-  """Selects target preparation for causal or block-diffusion pre-training.
+  """Selects target preparation for causal or block-diffusion training.
 
   Args:
     config: Training configuration containing the objective-specific settings.
     shift: Whether causal language-model targets should be shifted by one token.
     use_dpo: Whether the pipeline is preparing direct-preference data.
     use_sft: Whether the pipeline is preparing supervised fine-tuning data.
+    completion_only: Whether SFT supervision is restricted to completion tokens.
     packing: Whether multiple examples are packed into each sequence.
     pad_id: Token ID used to pad causal language-model examples.
     bos_token_id: Beginning-of-sequence token ID, or None when unavailable.
@@ -68,16 +70,14 @@ def _get_training_objective_transform(
 
   Raises:
     ValueError: If the objective is unsupported or block diffusion is combined with
-      an incompatible post-training or packing mode.
+      DPO or packing.
   """
   objective = getattr(config, "training_objective", "causal_lm")
   if objective == "block_diffusion":
-    if use_sft:
-      raise ValueError("This block-diffusion integration currently supports pre-training only.")
     if use_dpo:
-      raise ValueError("Block-diffusion pre-training is not compatible with DPO.")
+      raise ValueError("Block-diffusion training is not compatible with DPO.")
     if packing:
-      raise ValueError("Block-diffusion pre-training requires packing=False.")
+      raise ValueError("Block-diffusion training requires packing=False.")
     return input_pipeline_utils.BlockDiffusionCorruption(
         block_size=config.causal_block_size,
         mask_id=config.block_diffusion_mask_id,
@@ -85,6 +85,7 @@ def _get_training_objective_transform(
         logit_alignment=config.block_diffusion_logit_alignment,
         canvas_policy=config.block_diffusion_canvas_policy,
         axis=1,
+        completion_only=bool(use_sft and completion_only),
     )
   if objective != "causal_lm":
     raise ValueError(f"Unsupported training objective: {objective}")
@@ -402,6 +403,7 @@ def preprocessing_pipeline(
             completion_only=sft_train_on_completion_only,
             max_target_length=max_target_length,
             unk_id=pad_id,
+            training_objective=getattr(config, "training_objective", "causal_lm"),
         )
     )
     data_column_names = ("inputs", "targets")
@@ -437,6 +439,7 @@ def preprocessing_pipeline(
       shift=shift,
       use_dpo=use_dpo,
       use_sft=use_sft,
+      completion_only=sft_train_on_completion_only,
       packing=packing,
       pad_id=pad_id,
       bos_token_id=tokenizer.bos_token_id,
--- a/src/maxtext/integration/tunix/diffusion_sft.py
+++ b/src/maxtext/integration/tunix/diffusion_sft.py
@@ -1,0 +1,182 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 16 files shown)


src/maxtext/integration/tunix/diffusion_sft.py
```diff
@@ -0,0 +1,183 @@
...
+)
+
+
+def _concrete_numpy(value):
+  """Returns a host view only when eager validation is safe."""
+  if isinstance(value, jax.core.Tracer):
+    return None
+  if isinstance(value, jax.Array) and not value.is_fully_addressable:
+    return None
+  return np.asarray(value)
```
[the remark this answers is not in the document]
ethannnnnn: Thanks for flagging this. Fixed: `_concrete_numpy` now skips every `jax.Array`, tracer, and `ShapeDtypeStruct`, so device-backed batches no longer enter NumPy validation or trigger per-step device-to-host synchronization. Shape checks still run, while detailed value validation remains enabled for host-resident NumPy batches. Regression tests cover both paths.

src/maxtext/trainers/post_train/sft/hooks.py
```diff
@@ -29,7 +29,15 @@ class SFTTrainingHooks(BaseTrainingHooks):
...
-    return jnp.sum(batch["targets_segmentation"] != 0)
+    loss_weights = batch.get("targets_loss_mask")
+    if loss_weights is None:
+      loss_weights = batch["targets_segmentation"]
+    return jnp.sum(loss_weights != 0)
+
+  @override
+  def eval_loss_is_preaveraged(self) -> bool:
+    """Whether the configured objective supplies a preaveraged eval loss."""
+    return bool(self.config.loss_is_preaveraged or self.config.training_objective == "block_diffusion")
```
hengtaoguo, line 40: I wonder whether it's really necessary to expose `loss_is_preaveraged` to end users (adds up complexity). Can we remove this flag from base/types and instead do this?

```
  @override
  def eval_loss_is_preaveraged(self) -> bool:
    """Whether the configured objective supplies a preaveraged eval loss."""
    return getattr(self.config, "training_objective", "causal_lm") == "block_diffusion"
```

src/maxtext/input_pipeline/hf_data_processing.py
```diff
@@ -48,17 +48,19 @@ def _get_training_objective_transform(
     shift: bool,
     use_dpo: bool,
     use_sft: bool,
+    completion_only: bool,
```
hengtaoguo, line 51: This also changes the calling APIs and may cause incompatibility issues. I wonder if you could directly do `completion_only = bool(use_sft and getattr(config, "sft_train_on_completion_only", False))` inside the `_get_training_objective_transform()` function.

src/maxtext/integration/tunix/diffusion_sft.py
```diff
@@ -0,0 +1,182 @@
...
+def create_batch_adapter(config):
+  """Builds a raw MaxText batch adapter for Tunix diffusion SFT."""
+  alignment = config.block_diffusion_logit_alignment
+  block_size = int(config.causal_block_size)
+  completion_only = bool(config.sft_train_on_completion_only)
+
+  def adapt(raw_batch: Mapping[str, Any]) -> diffusion_types.DiffusionTokenBatch:
+    unavailable = sorted(name for name in _REQUIRED_FIELDS if name not in raw_batch or raw_batch[name] is None)
+    if unavailable:
+      raise ValueError("block-diffusion SFT requires non-None batch fields; " f"unavailable {unavailable}")
```
hengtaoguo, line 122: nit

`raise ValueError(f"block-diffusion SFT requires non-None batch fields; unavailable {unavailable}")`

src/maxtext/trainers/post_train/sft/hooks.py
ethannnnnn: Agreed. I removed the user-facing `loss_is_preaveraged` setting. `SFTTrainingHooks.eval_loss_is_preaveraged()` now derives this directly from `training_objective == "block_diffusion"`, with tests for both causal and diffusion objectives.

src/maxtext/input_pipeline/hf_data_processing.py
ethannnnnn: Agreed. I removed the `completion_only` argument from `_get_training_objective_transform()` and now derive it inside the helper from `use_sft` and `config.sft_train_on_completion_only`. Tests cover both configured values.

src/maxtext/integration/tunix/diffusion_sft.py
ethannnnnn: Done. The error message now uses a single f-string as suggested.

PR #4892, opened 2026-08-14
Enable Quantized All-Gather over FSDP for MoE embeddings

CaptainO5:
# Description

This PR introduces the `shard_embed_moe_on_fsdp` configuration flag, allowing `embed_moe` to remain sharded so we can manually perform Quantize-All-Gather (QAG) over FSDP. 

Previously, MoE weights were automatically gathered by XLA for computations. By explicitly retaining the FSDP sharding on the `embed_moe` axis during manual QAG (when using `fixed` weight calibration), we significantly reduce communication overhead and improve throughput.

**Note:** This optimization only triggers when using static scales (fixed weight calibration).

### Key Changes
- Updated `moe.py` partition specs to support manual QAG when `shard_embed_moe_on_fsdp` is enabled.
- Added config validators to enforce mutual exclusivity across FSDP sharding flags, safely check quantization rules, and fallback to standard sharding if static quantization isn't configured.
- Updated the `moe_configuration.md` doc.

FIXES: b/544811065

# Tests

- Added unit tests in `moe_test.py` and `pyconfig_test.py`.
- Ran an E2E test which showed a ~7.3% improvement in throughput compared to the regular fp8 run. Full E2E workload and XProf results are attached in the bug comments.

# Checklist

### DIFF
--- a/docs/reference/core_concepts/moe_configuration.md
+++ b/docs/reference/core_concepts/moe_configuration.md
@@ -121,7 +121,11 @@ MaxText implements an exact, paper-aligned version of DeepSeek V4's load balanci
 
 `moe_fsdp_use_two_stage_all_gather`: If enabled, split the All-Gather operation for MoE weights into two separate stages when using FSDP/FSDP-transpose sharding. This is preferred when 3D All-Gather support is unavailable.
 
-`shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis, and recommended only when num_experts is a multiple of fsdp_parallelism.
+**MoE FSDP Sharding Strategies** (Note: At most one of the following three flags can be enabled at a time):
+
+- `shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis. Works for both unquantized and quantized. When `quantization` and `weight_quantization_calibration_method` are fixed, it performs quantized weight all gather over fsdp before gmm. This is recommended only when `num_experts` is a multiple of `fsdp_parallelism`.
+- `use_2d_fsdp_sharding`: If enabled, use fsdp and fsdp_transpose axes for sharding the MoE weights.
+- `shard_embed_moe_on_fsdp`: If enabled, keep embed_moe sharded so we can manually QAG (Quantize-All-Gather) it over FSDP. Requires `quantization` to be specified and `weight_quantization_calibration_method` to be fixed.
 
 ## 3. Performance Tuning
 
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -286,6 +286,9 @@ moe_fsdp_use_two_stage_all_gather: false
 # Shard the expert dimension of the MLP weights on the FSDP axis.
 # This configuration is recommended only when num_experts is a multiple of fsdp_parallelism
 shard_exp_on_fsdp: false
+# Keep embed_moe sharded so we can manually QAG it over FSDP.
+# Requires quantization to be specified and weight_quantization_calibration_method to be fixed (static scaling mode).
+shard_embed_moe_on_fsdp: false
 # use fsdp and fsdp_transpose axes for sharding the moe weights
 use_2d_fsdp_sharding: false
 
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -952,6 +952,10 @@ class MoEGeneral(BaseModel):
       description="Shard the expert dimension of the MLP weights on the FSDP axis, "
       "and recommended only when num_experts is a multiple of fsdp_parallelism",
   )
+  shard_embed_moe_on_fsdp: bool = Field(
+      False,
+      description="Keep embed_moe sharded so we can manually QAG it over FSDP.",
+  )
   use_2d_fsdp_sharding: bool = Field(
       False,
       description="Use `fsdp` and `fsdp_transpose` axes for 2D FSDP sharding.",
@@ -983,6 +987,17 @@ class MoEGeneral(BaseModel):
   def validate_moe_chunks(self) -> "MoEGeneral":
     if self.num_moe_token_chunks > 1 and not self.use_ring_of_experts:
       raise ValueError("num_moe_token_chunks > 1 requires use_ring_of_experts=True.")
+    return self
+
+  @model_validator(mode="after")
+  def validate_moe_sharding_strategy(self) -> "MoEGeneral":
+    """Ensure that only one MoE FSDP sharding strategy is active at a time."""
+    active_sharding_flags = sum([self.shard_exp_on_fsdp, self.use_2d_fsdp_sharding, self.shard_embed_moe_on_fsdp])
+    if active_sharding_flags > 1:
+      raise ValueError(
+          "Only one of shard_exp_on_fsdp, use_2d_fsdp_sharding, or "
+          "shard_embed_moe_on_fsdp can be True at the same time."
+      )
     return self
 
 
@@ -3167,6 +3182,18 @@ class MaxTextConfig(
     # Explicitly setting encoding removes the need for the pylint disable comment
     with open(custom_mesh_path, "r", encoding="utf-8") as f:
       return yaml.safe_load(f) or {}
+
+  @model_validator(mode="after")
+  def validate_shard_embed_moe_on_fsdp(self) -> "MaxTextConfig":
+    """Raise ValueError if shard_embed_moe_on_fsdp is used without fixed weight quantization calibration."""
+    if self.shard_embed_moe_on_fsdp and (
+        self.quantization == "" or not self.weight_quantization_calibration_method.startswith("fixed")
+    ):
+      raise ValueError(
+          "shard_embed_moe_on_fsdp requires quantization to be specified and "
+          "weight_quantization_calibration_method to be fixed (static scaling mode)."
+      )
+    return self
 
   @model_validator(mode="after")
   def set_derived_and_validate_values(self) -> "MaxTextConfig":
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1596,11 +1596,12 @@ class RoutedMoE(nnx.Module):
       # In the decoding case, the expert axis is instead replicated along the tensor's batch dimension.
       return input_activation.shape[0] > 1
 
-    def explicitly_weight_ag(shard_exp_on_fsdp):
-      if shard_exp_on_fsdp:
-        quantization_rule = qpl.get_current_rule("gmm")
-        if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
-          return True
+    def explicitly_weight_ag():
+      if not (self.config.shard_exp_on_fsdp or self.config.shard_embed_moe_on_fsdp):
+        return False
+      quantization_rule = qpl.get_current_rule("gmm")
+      if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
+        return True
       return False
 
     def maybe_aqt_partition(w0_kernel, w0_pspec, w1_kernel, w1_pspec, wo_kernel, wo_pspec):
@@ -1653,8 +1654,7 @@ class RoutedMoE(nnx.Module):
       # w0, w1, wo needs to be un sharded on fsdp / fsdp_transpose axis, so use
       # mlp_no_fsdp axis
       if self.config.shard_exp_on_fsdp:
-        quantization_rule = qpl.get_current_rule("gmm")
-        if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
+        if explicitly_weight_ag():
           # special sharding when using static scaling for weights in quantization with shard_exp_on_fsdp
           w0_pspec = self._logical_to_mesh_axes(self.wi_kernel_axes)
           w1_pspec = self._logical_to_mesh_axes(self.wi_kernel_axes)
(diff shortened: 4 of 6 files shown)


src/maxtext/layers/moe.py
```diff
@@ -1654,8 +1654,13 @@ def get_routed_moe_shardings(is_batch_sharded_by_expert, has_input_ids):
         w0_pspec = self._logical_to_mesh_axes((None, "mlp_no_fsdp", None))
         w1_pspec = self._logical_to_mesh_axes((None, "mlp_no_fsdp", None))
         wo_pspec = self._logical_to_mesh_axes((None, "mlp_no_fsdp", None))
+      elif self.config.shard_embed_moe_on_fsdp and explicitly_weight_ag():
+        # Keep embed_moe sharded so we can manually QAG it over FSDP
+        w0_pspec = self._logical_to_mesh_axes(("exp", "embed_moe", "mlp_no_fsdp"))
+        w1_pspec = self._logical_to_mesh_axes(("exp", "embed_moe", "mlp_no_fsdp"))
```
NuojCheng, line 1674: This can be confusing to readers since we are abusing logical rule notations and claim "using embed_moe rule to mlp dimension".

src/maxtext/layers/moe.py
```diff
@@ -1582,11 +1582,12 @@ def is_batch_sharded_by_ep(input_activation):
       # In the decoding case, the expert axis is instead replicated along the tensor's batch dimension.
       return input_activation.shape[0] > 1
 
-    def explicitly_weight_ag(shard_exp_on_fsdp):
-      if shard_exp_on_fsdp:
-        quantization_rule = qpl.get_current_rule("gmm")
-        if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
-          return True
+    def explicitly_weight_ag():
```
NuojCheng, line 1599: hopefully removing if shard_exp_on_fsdp branch does no affect previous benchmark regression results... Maybe still keep the previous logic but adding your own?

src/maxtext/layers/moe.py
NuojCheng: I don't see a straightforward way to avoid this without writing a new customized logical rule... Maybe just keep whatever it is...

src/maxtext/layers/moe.py
```diff
@@ -1894,8 +1899,13 @@ def get_active_sharding_axes(pspec_dim_axes, tensor_dim_index):
     def get_wi_gmm_params():
       wi_gather_axes = []
       if weight_gather:
-        # wi [Experts, In, Hidden] -> Gather Exp(0) and Hidden(2)
-        wi_gather_axes.extend(get_active_sharding_axes(w0_pspec[0], 0))
+        if self.config.shard_exp_on_fsdp:
+          # wi [Experts, In, Hidden] -> Gather Exp(0)
+          wi_gather_axes.extend(get_active_sharding_axes(w0_pspec[0], 0))
```
NuojCheng, line 1919: wait... what if shard_exp_on_fsdp=False, isn't the logic gets changed here? Why?

src/maxtext/layers/moe.py
```diff
@@ -1913,8 +1923,13 @@ def get_wi_gmm_params():
     def get_wo_gmm_params():
       wo_gather_axes = []
       if weight_gather:
-        # wo [Experts, Hidden, Out] -> Gather Exp(0) and Hidden(1)
-        wo_gather_axes.extend(get_active_sharding_axes(wo_pspec[0], 0))
+        if self.config.shard_exp_on_fsdp:
+          # wo [Experts, Hidden, Out] -> Gather Exp(0)
+          wo_gather_axes.extend(get_active_sharding_axes(wo_pspec[0], 0))
+        else:
```
NuojCheng, line 1945: same above

src/maxtext/layers/moe.py
CaptainO5: Yes, I refactored the previous logic a little by using self.config.shard_exp_on_fsdp instead of getting the same as a parameter.

src/maxtext/layers/moe.py
CaptainO5: Good question! It's actually safe because this block only runs if weight_gather is true, which requires either shard_exp_on_fsdp or shard_embed_moe_fsdp to be true. Therefore, the else implicitly handles the shard_embed case. I'll add a comment to make this clearer

Shuwen-Fang:
can you specify in description if this is only for static scales?

src/maxtext/layers/moe.py
CaptainO5: Also, I made sure only one of them are true at a time with `validate_moe_sharding_strategy` in types.py

src/maxtext/layers/moe.py
CaptainO5: Thanks, I will add a comment to clarify

docs/reference/core_concepts/moe_configuration.md
```diff
@@ -119,7 +119,11 @@ MaxText implements an exact, paper-aligned version of DeepSeek V4's load balanci
 
 `moe_fsdp_use_two_stage_all_gather`: If enabled, split the All-Gather operation for MoE weights into two separate stages when using FSDP/FSDP-transpose sharding. This is preferred when 3D All-Gather support is unavailable.
 
-`shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis, and recommended only when num_experts is a multiple of fsdp_parallelism.
+**MoE FSDP Sharding Strategies** (Note: At most one of the following three flags can be enabled at a time):
+
+- `shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis. This is recommended only when num_experts is a multiple of fsdp_parallelism.
+- `use_2d_fsdp_sharding`: If enabled, use fsdp and fsdp_transpose axes for sharding the MoE weights.
+- `shard_embed_moe_on_fsdp`: If enabled, keep embed_moe sharded so we can manually QAG (Quantize-All-Gather) it over FSDP. Requires `quantization` to be specified and `weight_quantization_calibration_method` to be fixed.
```
Shuwen-Fang, line 128: for my own understanding, is this flag required for fsdp ag quantization?

CaptainO5: Yes

NuojCheng (approved):

src/maxtext/configs/base.yml
```diff
@@ -282,6 +282,9 @@ moe_fsdp_use_two_stage_all_gather: false
 # Shard the expert dimension of the MLP weights on the FSDP axis.
 # This configuration is recommended only when num_experts is a multiple of fsdp_parallelism
 shard_exp_on_fsdp: false
+# Keep embed_moe sharded so we can manually QAG it over FSDP.
+# Requires quantization to be specified and weight_quantization_calibration_method to be fixed.
```
RissyRan: Nit: let's say something like `Requires quantization to be specified and weight_quantization_calibration_method to be fixed (static scaling mode)`

src/maxtext/configs/types.py
```diff
@@ -3159,6 +3174,20 @@ def _load_mesh_config_from_yaml(rule_value: str) -> dict:
     with open(custom_mesh_path, "r", encoding="utf-8") as f:
       return yaml.safe_load(f) or {}
 
+  @model_validator(mode="after")
+  def validate_moe_qag_requirements(self) -> "MaxTextConfig":
+    """Warn if shard_embed_moe_on_fsdp is used without fixed weight quantization calibration."""
+    if self.shard_embed_moe_on_fsdp and (
+        self.quantization == "" or not self.weight_quantization_calibration_method.startswith("fixed")
+    ):
+      logger.warning(
```
RissyRan: Suggest to make it assertion instead of warning, to avoid potential confusion in debugging.

tests/unit/moe_test.py
```diff
@@ -1053,6 +1054,86 @@ def test_moe_fsdp_two_stage_parallelism_tpu_only(self):
       actual_output, _, _ = self.get_moe_output(variables, hidden_states, cfg, mesh)
       assert_moe_close(actual_output, expected_output, cfg.dtype)
 
+  @pytest.mark.tpu_only
+  def test_shard_embed_moe_on_fsdp(self):
+    if jax.device_count() != 4:
+      self.skipTest("shard_embed_moe_on_fsdp test requires exactly 4 devices")
```
RissyRan, line 1107: Will this be silently skipped if we switch tpu type?

RissyRan (commented):
Thanks for the change! Do you know why this sharding helps 7.3%?

CaptainO5:
> Thanks for the change! Do you know why this sharding helps 7.3%?

This change ensures the XLA compiler does not gather weights prematurely, instead performing manual all gathers after quantization. This helped reduce the all-gather times and helped hide them effectively during the E2E test.

src/maxtext/configs/types.py
CaptainO5: Makes sense, sure.

tests/unit/moe_test.py
CaptainO5: Yes, it would be silently skipped if the CI ever switches to a different TPU size. I followed the pattern from test_moe_fsdp_two_stage_parallelism_tpu_only above. I think the github tests currently use 4 devices so it should run consistently. We need the number of devices to match the parallelism configs.

docs/reference/core_concepts/moe_configuration.md
shuningjin, line 128: Could you clarify why quantization needs to be specified? 
- I think it should work with bf16. But perhaps there is no point of doing so, as it is strictly worse?

src/maxtext/configs/types.py
```diff
@@ -3159,6 +3174,18 @@ def _load_mesh_config_from_yaml(rule_value: str) -> dict:
     with open(custom_mesh_path, "r", encoding="utf-8") as f:
       return yaml.safe_load(f) or {}
 
+  @model_validator(mode="after")
+  def validate_moe_qag_requirements(self) -> "MaxTextConfig":
```
shuningjin: nit: just use `validate_shard_embed_moe_on_fsdp`?

src/maxtext/layers/moe.py
```diff
@@ -1573,11 +1573,12 @@ def is_batch_sharded_by_ep(input_activation):
...
-      if shard_exp_on_fsdp:
-        quantization_rule = qpl.get_current_rule("gmm")
-        if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
-          return True
+    def explicitly_weight_ag():
+      if not (self.config.shard_exp_on_fsdp or self.config.shard_embed_moe_on_fsdp):
+        return False
+      quantization_rule = qpl.get_current_rule("gmm")
+      if quantization_rule and quantization_rule.weight_calibration_method.startswith("fixed"):
+        return True
```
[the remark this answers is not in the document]
shuningjin: nit: I feel this `defensive programming` is not needed.

RissyRan (approved):

tests/unit/moe_test.py
```diff
@@ -1053,6 +1054,86 @@ def test_moe_fsdp_two_stage_parallelism_tpu_only(self):
       actual_output, _, _ = self.get_moe_output(variables, hidden_states, cfg, mesh)
       assert_moe_close(actual_output, expected_output, cfg.dtype)
 
+  @pytest.mark.tpu_only
+  def test_shard_embed_moe_on_fsdp(self):
```
shuningjin, line 1105: I would prefer that we test both value and grad, similar to
https://github.com/AI-Hypercomputer/maxtext/blob/4e5ccd4cc369e48a9ad5201fc1e07baf7bd99c92/tests/unit/moe_test.py#L1569

In particular, the weight QAG feature has both forward and backward.
https://github.com/AI-Hypercomputer/maxtext/blob/4e5ccd4cc369e48a9ad5201fc1e07baf7bd99c92/src/maxtext/kernels/megablox/ops.py#L198
https://github.com/AI-Hypercomputer/maxtext/blob/4e5ccd4cc369e48a9ad5201fc1e07baf7bd99c92/src/maxtext/kernels/megablox/ops.py#L792-L802

docs/reference/core_concepts/moe_configuration.md
```diff
@@ -119,7 +119,11 @@ MaxText implements an exact, paper-aligned version of DeepSeek V4's load balanci
 
 `moe_fsdp_use_two_stage_all_gather`: If enabled, split the All-Gather operation for MoE weights into two separate stages when using FSDP/FSDP-transpose sharding. This is preferred when 3D All-Gather support is unavailable.
 
-`shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis, and recommended only when num_experts is a multiple of fsdp_parallelism.
+**MoE FSDP Sharding Strategies** (Note: At most one of the following three flags can be enabled at a time):
+
+- `shard_exp_on_fsdp`: If enabled, shard the expert dimension of the MLP weights on the FSDP axis. This is recommended only when num_experts is a multiple of fsdp_parallelism.
```
shuningjin: Add more comment: 

Works for both unquantized and quantized. When `quantization` and `weight_quantization_calibration_method` are specified, it performs quantized weight all gather over fsdp before gmm.

shuningjin (commented):
Thanks for adding this great feature! Overall looks good.
- Would appreciate more in-line clarifying note: why it is beneficial for quantization but not for unquantized.
- Recommend to test both value and gradient, as the feature uses custom vjp with `_fwd_gather_weight` and `_drhs_scatter_weight`

docs/reference/core_concepts/moe_configuration.md
CaptainO5: The quantization needs to be specified, because we are trying to manually gather weights after they are quantized. For bf16, like you mentioned, manually gathering doesn't make sense and it's disabled. We can let the compiler decide when to gather the weights.

src/maxtext/layers/moe.py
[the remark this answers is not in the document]
CaptainO5: Yes, changing it back, as I only added it here inconsistently

shuningjin (approved):
Great work!!

PR #4900, opened 2026-08-17
Optimize MoE routing for Gemma4

aireenmei:
# Description

Add an optional direct-token-gather path for sparse MoE routing. It gathers input tokens directly in expert order instead of materializing Top-K copies with jnp.repeat before sorting.

For Gemma4-26B, this removes the large Top-8 activation expansion while preserving expert-ID rotation and correct gradient accumulation.

The optimization is controlled by moe_use_direct_token_gather, which defaults to false and is currently enabled only for Gemma4-26B. Its performance improvement has only been validated on that model; other MoE configurations retain the existing repeat-and-sort path until they are benchmarked.

# Tests

Add unit test for numerical equivalence.

XProf confirms that the direct-gather path removes the broadcast_bitcast_fusion associated with the repeated activation, and improved end-to-end step time.
Before: [xprof](https://xprof.corp.google.com/trace_viewer/aireenmei-13534665124925283610?view_start=11352.617&view_end=11358.416), [screenshot](https://screenshot-v2.corp.google.com/40jao38emg6eo), step time: 10.5 s
Before: [xprof](https://xprof.corp.google.com/trace_viewer/aireenmei-6309662714480977089?view_start=11644.359&view_end=11652.992), [screenshot](https://screenshot-v2.corp.google.com/718p0f3ruv7uo), step time: 10.39 s

# Checklist

### DIFF
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -33,6 +33,7 @@ from jax.sharding import PartitionSpec as P
 from maxtext.common import common_types as ctypes
 from maxtext.common.common_types import ShardMode
 from maxtext.kernels import megablox as mblx
+from maxtext.kernels import sort_activations
 from maxtext.layers import attentions, linears, nnx_wrappers, quantizations
 from maxtext.layers.initializers import NdInitializer, default_bias_init, nd_dense_init, variable_to_logically_partitioned
 from maxtext.kernels.ragged.ragged_sort import a2a_ragged_sort
@@ -149,6 +150,16 @@ def _sort_activations(
     if use_custom_vjp:
       return _sort_activations_custom(inputs, sort_indices)
     return inputs[sort_indices, ...]
+
+
+def _route_activations(inputs: jax.Array, selected_experts: jax.Array) -> jax.Array:
+  """Groups token activations by expert without materializing Top-K copies."""
+  selected_experts = selected_experts.reshape((inputs.shape[0], -1))
+  # When use_gather_mosaic_kernel=False, use the general JAX backward, which
+  # gathers and sums gradients from all expert copies of each token. The
+  # optimized Mosaic gather-reduce kernel currently requires exactly 8 selected
+  # experts per token and is enabled only by the specialized batch-split path.
+  return sort_activations.route(inputs, selected_experts, use_gather_mosaic_kernel=False)
 
 
 @jax.custom_vjp
@@ -952,11 +963,13 @@ class RoutedMoE(nnx.Module):
       if roll_to_expert_id is not None:
         flatten_selected_experts = (flatten_selected_experts - roll_to_expert_id) % self.num_experts
       sorted_selected_experts = jnp.argsort(flatten_selected_experts)
-      # sort inputs for number of selected experts
-      replicated_inputs_2d = jnp.repeat(inputs_2d, self.num_experts_per_tok, axis=0)
-      sorted_inputs = _sort_activations(replicated_inputs_2d, sorted_selected_experts, use_custom_sort_vjp).astype(
-          self.dtype
-      )
+      if self.config.moe_use_direct_token_gather:
+        sorted_inputs = _route_activations(inputs_2d, flatten_selected_experts).astype(self.dtype)
+      else:
+        replicated_inputs_2d = jnp.repeat(inputs_2d, self.num_experts_per_tok, axis=0)
+        sorted_inputs = _sort_activations(replicated_inputs_2d, sorted_selected_experts, use_custom_sort_vjp).astype(
+            self.dtype
+        )
       group_size = jnp.bincount(flatten_selected_experts, length=self.num_experts)
 
     num_tokens = bsz_times_seq_len * self.num_experts_per_tok
--- a/tests/unit/moe_test.py
+++ b/tests/unit/moe_test.py
@@ -504,6 +504,53 @@ def test_sparse_matmul_repairs_batch_specs_only_without_expert_parallelism(exper
 class RoutedMoeTest(parameterized.TestCase):
   """Routed Mixture of Experts test."""
 
+  @parameterized.parameters(False, True)
+  def test_permute_direct_token_gather_matches_repeat_and_sort(self, compute_gradient):
+    inputs = jnp.arange(24, dtype=jnp.float32).reshape(2, 3, 4)
+    selected_experts = jnp.array(
+        [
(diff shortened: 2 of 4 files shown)


src/maxtext/layers/moe.py
```diff
@@ -952,11 +961,8 @@ def permute(
       if roll_to_expert_id is not None:
         flatten_selected_experts = (flatten_selected_experts - roll_to_expert_id) % self.num_experts
       sorted_selected_experts = jnp.argsort(flatten_selected_experts)
-      # sort inputs for number of selected experts
-      replicated_inputs_2d = jnp.repeat(inputs_2d, self.num_experts_per_tok, axis=0)
-      sorted_inputs = _sort_activations(replicated_inputs_2d, sorted_selected_experts, use_custom_sort_vjp).astype(
-          self.dtype
-      )
+      routing_experts = flatten_selected_experts.reshape(selected_experts.shape)
+      sorted_inputs = _route_activations(inputs_2d, routing_experts).astype(self.dtype)
```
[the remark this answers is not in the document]
aireenmei: update as suggested

tests/unit/moe_test.py
```diff
@@ -504,6 +504,34 @@ def fake_shard_map(function, *, mesh, in_specs, out_specs, check_vma):
...
+
+    if compute_gradient:
+      cotangent = jnp.arange(48, dtype=jnp.float32).reshape(12, 4)
+      expected = jax.grad(lambda x: jnp.sum(reference(x) * cotangent))(inputs)
+      result = jax.grad(lambda x: jnp.sum(actual(x) * cotangent))(inputs)
+    else:
+      expected = reference(inputs)
+      result = actual(inputs)
+
+    np.testing.assert_array_equal(result, expected)
```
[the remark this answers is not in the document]
aireenmei: This test uses small integer-valued float32 arrays, and only performs indexing and sums of exactly representable integers. There should be no floating-point approximation difference. Exact equality is better here.

src/maxtext/layers/moe.py
```diff
@@ -952,11 +963,13 @@ def permute(
       if roll_to_expert_id is not None:
         flatten_selected_experts = (flatten_selected_experts - roll_to_expert_id) % self.num_experts
       sorted_selected_experts = jnp.argsort(flatten_selected_experts)
-      # sort inputs for number of selected experts
-      replicated_inputs_2d = jnp.repeat(inputs_2d, self.num_experts_per_tok, axis=0)
-      sorted_inputs = _sort_activations(replicated_inputs_2d, sorted_selected_experts, use_custom_sort_vjp).astype(
-          self.dtype
-      )
+      if self.config.moe_use_direct_token_gather:
+        sorted_inputs = _route_activations(inputs_2d, flatten_selected_experts).astype(self.dtype)
```
NuojCheng, line 967: if sorted_selected_experts is not used here, maybe move line 954 `sorted_selected_experts = jnp.argsort(flatten_selected_experts)` in the else branch?

src/maxtext/layers/moe.py
```diff
@@ -151,6 +152,16 @@ def _sort_activations(
...
 
 
+def _route_activations(inputs: jax.Array, selected_experts: jax.Array) -> jax.Array:
+  """Groups token activations by expert without materializing Top-K copies."""
+  selected_experts = selected_experts.reshape((inputs.shape[0], -1))
+  # When use_gather_mosaic_kernel=False, use the general JAX backward, which
+  # gathers and sums gradients from all expert copies of each token. The
+  # optimized Mosaic gather-reduce kernel currently requires exactly 8 selected
+  # experts per token and is enabled only by the specialized batch-split path.
+  return sort_activations.route(inputs, selected_experts, use_gather_mosaic_kernel=False)
```
NuojCheng, line 162: if use_gather_mosaic_kernel is fixed as false when moe_use_direct_token_gather=True, we should add a check in types.py to throw early value error

NuojCheng (approved):
LGTM to me with some minor comments. Note that this change does not affect the ragged sort path.

src/maxtext/configs/models/gemma4-26b.yml
```diff
@@ -46,6 +46,7 @@ num_experts_per_tok: 8
 shared_experts: 1
 norm_topk_prob: true
 load_balance_loss_weight: 0.001
+moe_use_direct_token_gather: true
```
RissyRan: I think this is an implementation feature (perf config), but model specific config (like num_experts etc)? Recommend to remove it if this is the case.

tests/unit/moe_test.py
```diff
@@ -504,6 +504,34 @@ def fake_shard_map(function, *, mesh, in_specs, out_specs, check_vma):
 class RoutedMoeTest(parameterized.TestCase):
   """Routed Mixture of Experts test."""
 
+  @parameterized.parameters(False, True)
+  def test_route_activations_matches_repeat_and_sort(self, compute_gradient):
```
RissyRan: I think it's better to directly test `_route_activations()` with moe_use_direct_token_gather=true/false?

tests/unit/moe_test.py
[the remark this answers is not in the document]
RissyRan: +1 to `It is safer and more robust to use np.testing.assert_allclose with appropriate tolerances.` Also, let's use float32 for inputs instead of integer? This should be more similar to real situation.

RissyRan (commented):
Thanks! Minor comments

src/maxtext/layers/moe.py
aireenmei: Although not used in the `self.config.moe_use_direct_token_gather` branch. sorted_selected_experts is later returned by this function on line 1008, so it has to be outside the else branch.

src/maxtext/layers/moe.py
aireenmei: added early validation

src/maxtext/configs/models/gemma4-26b.yml
aireenmei: Makes sense, removed

tests/unit/moe_test.py
aireenmei: I think you mean testing moe_use_direct_token_gather=true/false at the `permute()` function level since `_route_activations()` is a helper function for the moe_use_direct_token_gather=true path. I've updated the test for this purpose and renamed to `test_permute_direct_token_gather_matches_repeat_and_sort`

RissyRan (approved):

PR #4905, opened 2026-08-17
MoE: zero-initialize the ragged all-to-all combine buffer when ragged_buffer_factor > 0

systalyze-ai:
# Description

Since [#4397](https://github.com/AI-Hypercomputer/maxtext/pull/4397), `ragged_buffer_factor > 0` truncates overflow but combine all-to-all writes the truncated sizes at the original offsets, leaving unwritten interior rows in a `lax.empty` buffer. Weighted combine reads those rows at full router weight, introducing uninitialized values into activations and gradients on TPU/GPU; `lax.empty` reads as zero on CPU, hiding the bug from CPU CI. This breaks [#3541](https://github.com/AI-Hypercomputer/maxtext/pull/3541)'s invariant that every `lax.empty` row is overwritten.

The combine buffer is now zero-initialized only when truncation is enabled; the default `ragged_buffer_factor=-1` path remains byte-identical. Dispatch still uses `lax.empty` because receives form a packed prefix and unwritten rows are never read.

# Tests

Added two CPU tests: one checks the emitted allocation primitive (`zeros` vs `empty`, since `empty` reads as zero on CPU), and one replays combine row copies to verify a dropped row reads zero. End-to-end: ran 24 DeepSeek-V4-Flash 284B LoRA training steps on v6e-128 with `ragged_buffer_factor=1.015625` and smooth monotonic loss. Setup: `ici_expert_parallelism=16`, `ici_tensor_parallelism=8`, `ici_fsdp_parallelism=1`, `max_target_length=16384`, `per_device_batch_size=0.125` (global batch 16), `LIBTPU_INIT_ARGS=--xla_tpu_scoped_vmem_limit_kib=98304`, dynamic-splash attention path enabled.

- `pytest tests/unit/moe_test.py` — 22 passed, 38 skipped (`origin/main`: 20 passed; this PR adds 2).
- `pyink --check --pyink-indentation=2 --line-length=122 src/maxtext/layers/moe.py tests/unit/moe_test.py` — passed.
- Mutation check: forcing the combine allocation back to `lax.empty` fails the emitted-primitive regression test.

# Checklist


richjames0 (approved):
lgtm

src/maxtext/layers/moe.py
```diff
@@ -120,6 +120,13 @@ def _truncate_matrix(all_shards_group_sizes: jax.Array, buffer_size: int) -> jax
   return jnp.diff(clamped_cumsum_extended, axis=0)
 
 
+def _ragged_all_to_all_output_buffer(shape, dtype, zero_init: bool) -> jax.Array:
```
RissyRan, line 123: Consider making this function name more generic (e.g., _initialize_buffer) rather than specific to all-to-all.

RissyRan (approved):
Thanks for the fix!

src/maxtext/layers/moe.py
RSudarsanan: Thanks, the function name has been fixed.

PR #4906, opened 2026-08-17
Flash attention: replicate the kv head axis for MQA

systalyze-ai:
# Description

Changes flash attention to replicate the K/V head axis for MQA when the mesh shards that axis. This prevents a trace failure from applying query-head sharding to K/V.

The fix applies only to MQA, where `num_kv_heads=1`. It does not replicate the K/V head axis for GQA, because each shard would pair its local query heads with the wrong K/V heads and silently compute incorrect attention. Replicating 2 K/V heads across 4 head shards produced output differing from the dense reference by up to 5.9 on an 8-device CPU mesh.

# Tests

- `PYTHONPATH=src JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=8 pytest -q tests/unit/attention_test.py`: 71 passed.

# Checklist

### DIFF
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -142,6 +142,38 @@ def validate_gpu_flash_attention(sinks: Array | None, record_max_logits: bool) -
     raise NotImplementedError("record_max_logits (QK-Clip) is not supported for GPU flash attention kernels yet.")
 
 
+def _replicate_indivisible_mqa_kv_head_axis(
+    axis_names_kv: jax.sharding.PartitionSpec | None, num_kv_heads: int, mesh: Mesh
+) -> jax.sharding.PartitionSpec | None:
+  """Replicates the sharded K/V head axis for MQA.
+
+  Replicating multiple K/V heads would change their shard-local grouping with query heads.
+  """
+  if num_kv_heads != 1 or axis_names_kv is None:
+    return axis_names_kv
+  # PartitionSpecs with unreduced/reduced metadata cannot be indexed or iterated.
+  partitions = axis_names_kv.partitions
+  if len(partitions) <= 1:
+    return axis_names_kv
+  kv_head_spec = partitions[1]
+  if isinstance(kv_head_spec, str):
+    kv_head_axes = (kv_head_spec,)
+  elif isinstance(kv_head_spec, tuple):
+    kv_head_axes = kv_head_spec
+  else:
+    return axis_names_kv
+  kv_head_shards = math.prod(mesh.shape.get(axis, 1) for axis in kv_head_axes)
+  if kv_head_shards <= 1:
+    return axis_names_kv
+  return jax.sharding.PartitionSpec(
+      partitions[0],
+      None,
+      *partitions[2:],
+      unreduced=axis_names_kv.unreduced,
+      reduced=axis_names_kv.reduced,
+  )
+
+
 # TODO(agagik): change splash_attention_mask._ComputableMask to be non protected
 class ChunkedCausalMask(splash_attention_mask._ComputableMask):  # pylint: disable=protected-access
   """Lazy chunked causal mask.
@@ -1664,6 +1696,7 @@ class AttentionOp(nnx.Module):
     axis_names_splash_kernel = self._logical_to_mesh_axes(self.flash_axis_names_splash_kernel)
     axis_names_q = self._logical_to_mesh_axes(self.flash_axis_names_q)
     axis_names_kv = self._logical_to_mesh_axes(self.flash_axis_names_kv)
+    axis_names_kv = _replicate_indivisible_mqa_kv_head_axis(axis_names_kv, key.shape[1], self.mesh)
     indexer_mask_axis_names = self._logical_to_mesh_axes((BATCH_ATTN, Q_LENGTH, KV_LENGTH))
     if use_tokamax_ring:
       context_axis = self.config.context_sharding
--- a/tests/unit/attention_test.py
+++ b/tests/unit/attention_test.py
@@ -915,6 +915,157 @@ class CudnnTePackedSequenceDescriptorTest(unittest.TestCase):
       )
 
 
+class FlashAttentionKvHeadShardingTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


richjames0 (changes requested):
Looks good but it looks like the testing could be tightened up a bit:

<bot>

The main problem: CI never runs the relevant tests. tests/conftest.py:203-207 auto-marks unmarked tests as cpu_only, and both new classes are unmarked — so they run only on the cpu-unit flavor. Nothing in pytest.ini or the workflows sets --xla_force_host_platform_device_count, so len(jax.devices()) == 1 there and FlashAttentionShardedHeadAxisTest.setUp skips (it needs ≥4). The three numerical-parity and gradient tests — the only ones that actually prove MQA is correct over a sharded head axis — never execute. Only the three fake-mesh unit tests run. Fix: mark the class tpu_only, or force the device count.

Second: the tests bypass the layer where the existing check lives. _build_op constructs AttentionOp directly. But attention_test.py:4882 already has a KVHeadShardingTest documenting the invariant "KV heads must divide the mesh axes that shard kv_heads", with the error num_kv_heads (2) … must be divisible by 4, enforced on the Attention layer's KV kernel axes. For MQA, 1 % 4 != 0. So it's unclear whether a real Attention layer with num_kv_heads=1 and a 4-way-sharded head axis even constructs — 4906 patches the flash runtime spec but not the weight-sharding path. Ask for one test through Attention, or an explanation of why the weight path is already replicated. Also: consider consolidating with that existing class rather than adding a similarly named one.

Third: the name promises a check that isn't there. _replicate_indivisible_mqa_kv_head_axis never checks divisibility — it gates purely on num_kv_heads != 1. test_keeps_kv_head_axis_when_kv_heads_divide_shards passes 8 and returns early on the != 1 branch, not on divisibility, so the test name asserts behavior that isn't being exercised. If that gate is ever relaxed, the test gives false confidence.

Fourth: the GQA rejection is incidental. test_indivisible_non_mqa_gqa_over_sharded_head_axis_is_rejected expects a ValueError, but the new code returns the spec unchanged — the error comes from downstream JAX sharding. Users get a cryptic sharding error rather than "GQA with N kv heads cannot be sharded M ways." An explicit raise with a good message would be worth the three lines, and would make the test assert the PR's own behavior.

Minor: three other _logical_to_mesh_axes(self.flash_axis_names_kv) sites (attention_op.py:667, 713, 765) feed the tokamax-ring / Ulysses / USP init-time validators, which will validate the unpatched spec while the runtime uses the patched one — only matters for MQA + context parallelism, but worth a sentence. partitions[1] is a hardcoded index that silently targets the wrong axis if a caller overrides flash_axis_names_kv; an assert on the logical name is cheap. And test_indivisible_non_mqa_gqa_is_left_untouched uses a mesh containing only {"tensor": 4} while the spec names "expert" — works via .get(axis, 1), but suboptimal.

</bot>

arminghobadi:
Thanks @richjames0 for the detailed feedback. We've addressed all the points:

First (CI never runs tests): Added `@pytest.mark.tpu_only` to `FlashAttentionShardedHeadAxisTest` ([line 1650](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-0eb6591cbdcc057eb77acc18689aaf6b95a4f4756a9908620556cf2733b3cb31R1650)) so the integration tests run on TPU hardware where we have 4+ devices.

Second (Tests bypass Attention layer): Added a new test `test_mqa_with_sharded_head_axis_constructs()` in `KVHeadShardingTest` ([line 6499](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-0eb6591cbdcc057eb77acc18689aaf6b95a4f4756a9908620556cf2733b3cb31R6499)) that validates the actual Attention layer constructs with `num_kv_heads=1` and a sharded head axis. This confirms the weight-sharding path is compatible with the fix.

Third (Divisibility check missing): Updated `_replicate_indivisible_mqa_kv_head_axis()` ([line 199](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R199)) to explicitly check `num_kv_heads % kv_head_shards != 0` and raise `ValueError` for GQA/MHA cases. The test now validates all three scenarios: indivisible cases that raise, and divisible cases that don't.

Fourth (GQA rejection incidental): Same function now raises immediately with a clear error message ([line 244](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R244-R250)) instead of letting JAX throw a cryptic error downstream. Test verifies this ([line 1629](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-0eb6591cbdcc057eb77acc18689aaf6b95a4f4756a9908620556cf2733b3cb31R1629)).

Minor issues:
- Added comments at [lines 772-775](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R772) (Tokamax), [822-825](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R822-R825) (Ulysses), [878-881](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R878-R881) (USP) documenting the validator asymmetry
- Added semantic assertion on `partitions[1]` ([line 218](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-fc2e215df5a159bfcab96c50a533c05456ef38370917bebe7f0090ac5621a685R218)) with explanation of expected schema
- Fixed mesh/spec mismatch in test by adding `"expert": 1` to mesh ([line 1633](https://github.com/AI-Hypercomputer/maxtext/pull/4906/changes#diff-0eb6591cbdcc057eb77acc18689aaf6b95a4f4756a9908620556cf2733b3cb31R1633))

PR #4908, opened 2026-08-17
DSV4: Route compressed and sliding train attention onto the splash kernels

systalyze-ai:
# Description

Adds an opt-in TPU path that routes DeepSeek-V4 `COMPRESSED` and `LOCAL_SLIDING` train attention to the existing splash kernels. This avoids materializing dense q x kv logits.

# Performance

- At seq 16384, a v6e-128 DeepSeek4-284B LoRA run required 57.5 GiB/chip without this change. The hardware had 31.25 GiB/chip available, so the model did not compile.
- At seq 512, step time fell from 0.991 to 0.637 s/step (same model on v6e-64, where the dense form still fits).

Setup: `ici_expert_parallelism=16`, `ici_tensor_parallelism=8`, `ici_fsdp_parallelism=1`, `max_target_length=16384`, `per_device_batch_size=0.125` (global batch 16), `LIBTPU_INIT_ARGS=--xla_tpu_scoped_vmem_limit_kib=98304`, dynamic-splash attention path enabled. The seq 16384 memory figure is the AOT compile requirement at that setup with this change removed.

# Tests

- `pytest tests/unit/dynamic_splash_mask_test.py tests/unit/attention_test.py`: `137 passed`.

# Checklist


richjames0 (changes requested):
There seem to be a few issues here:

1. Critical: the boolean mask is destroyed before it reaches the kernel

dynamic_splash_attention builds a boolean mask and hands it to tpu_flash_attention as indexer_mask (attention_op.py:1820-1841):

bool_mask = build_compressed_splash_mask(compressed_mask, decoder_segment_ids, segment_positions, q_seq_len, kv_seq_len, self.sliding_window_size)
...
out, max_logits = self.tpu_flash_attention(..., indexer_mask=bool_mask, ...)

Inside tpu_flash_attention, the cp_size == 1 branch converts unconditionally (attention_op.py:2387-2389):

if indexer_mask is not None:
  # Convert additive float mask (0.0=attend, negative=masked) to boolean mask for Tokamax splash kernel
  indexer_mask = _additive_to_boolean_mask(indexer_mask)

and _additive_to_boolean_mask (line 138) is return mask >= DEFAULT_MASK_VALUE * 0.5. Applied to a bool array, False promotes to 0.0, and 0.0 >= -1.19e38 is True. Every entry becomes True. Reproduced in jax:

additive -> bool : [[True, False], [False, True]]
bool     -> bool : [[True, True], [True, True]]

Consequence: with compressed_use_dynamic_splash=true, COMPRESSED layers attend to everything — no causality, no sliding window, no packing boundaries, no compressed-block selection. Training would produce garbage loss.

The path is forced, not incidental:
- the new config validator requires use_tokamax_splash=true, so the if self.config.use_tokamax_splash: branch at 2333 is always taken;
- dynamic_splash_attention:1806 raises NotImplementedError when context parallelism is on, so cp_size > 1 at 2334 is unreachable;
- LOCAL_SLIDING is unaffected — it passes indexer_mask=None (line 1544ff). Only COMPRESSED is broken.

The author did add a dtype guard — but at lines 2417-2418, thirty lines downstream, after the damage:

attn_fn = jax.vmap(dynamic_mask_splash_kernel, (0, 0, 0, 0, None, 0))
if indexer_mask.dtype != jnp.bool_:
  indexer_mask = _additive_to_boolean_mask(indexer_mask)

That guard is dead code as written. Fix: move it to line 2389, replacing the unconditional call.

Archaeology: 628a3a683 removed the two pre-existing indexer_mask == 0.0 conversions; 5e6c42ddf added the three _additive_to_boolean_mask call sites including the misplaced guard. Both are "address Gemini review comments" commits, i.e. this was introduced by review-feedback churn late in the PR, not in the original design commit.

2. Unannounced behaviour change on the existing indexer path

The keep predicate for the pre-existing (non-dynamic-splash) indexer path changed from indexer_mask == 0.0 to indexer_mask >= DEFAULT_MASK_VALUE * 0.5. For masks holding only {0.0, DEFAULT_MASK_VALUE} these agree, but any intermediate additive value (soft caps, bias terms, partial penalties) now flips from masked to attend. Not mentioned in the PR body; either justify it or restore == 0.0 for that path.

3. CI lint failure

pyink --check --pyink --pyink-indentation=2 --line-length=122 fails on attention_op.py at HEAD (50992850d), exit 1. Two lines the PR adds exceed 122 chars:

- attention_op.py:175 — segment = (decoder_segment_ids[...] == ...) & (decoder_segment_ids[...] >= 0)
- attention_op.py:206 — the raise ValueError(f"Dimension mismatch: ...")

Bisected: clean at ed84ea379, broken from 628a3a683 onward — again the review-feedback commits. dynamic_splash_mask_test.py is clean.

4. Minor: mask/query length mismatch when block_q > q_seq_len

The validation loop at 1813 uses eff = min(blk, q_seq_len), so a config with block_q > q_seq_len passes. Line 2390 then pads the mask's q dim up to a multiple of block_q while query stays at q_seq_len. Narrow, but the guard reads as if it prevents this.

5. Staleness

mergeable_state: dirty — 4908 has merge conflicts against main (319 behind).

PR #4910, opened 2026-08-17
MoE: add moe_late_tp_combine so routed experts take the tensor axis as expert parallelism

systalyze-ai:
# Description

Adds `moe_late_tp_combine` so routed experts use the tensor axis as additional expert parallelism instead of sharding the MoE hidden dimension. This removes tensor-axis collectives from expert-expanded activations.

# Performance

- On v6e-128 with DeepSeek-V4-Flash 284B LoRA at seq 16384, step time measured 6.811 s/step with the flag off and 4.632 s/step with it on (1.47x).

Setup: `ici_expert_parallelism=16`, `ici_tensor_parallelism=8`, `ici_fsdp_parallelism=1`, `max_target_length=16384`, `per_device_batch_size=0.125` (global batch 16), `LIBTPU_INIT_ARGS=--xla_tpu_scoped_vmem_limit_kib=98304`, dynamic-splash attention path enabled.

# Tests

- `JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=8 pytest tests/unit/moe_late_tp_combine_test.py` passed 3 tests.

# Checklist

### DIFF
--- a/src/maxtext/configs/base.yml
+++ b/src/maxtext/configs/base.yml
@@ -284,6 +284,8 @@ moe_fsdp_use_two_stage_all_gather: false
 shard_exp_on_fsdp: false
 # use fsdp and fsdp_transpose axes for sharding the moe weights
 use_2d_fsdp_sharding: false
+# Use the tensor axis as expert parallelism in the default sparse MoE path.
+moe_late_tp_combine: false
 
 # deepseek moe
 first_num_dense_layers: 0 # number of initial dense layers in the model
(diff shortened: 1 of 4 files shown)


richjames0 (approved):
LGTM but @RissyRan is the expert here

src/maxtext/configs/base.yml
```diff
@@ -284,6 +284,8 @@ moe_fsdp_use_two_stage_all_gather: false
 shard_exp_on_fsdp: false
 # use fsdp and fsdp_transpose axes for sharding the moe weights
 use_2d_fsdp_sharding: false
+# Use the tensor axis as expert parallelism in the default sparse MoE path.
+moe_late_tp_combine: false
```
RissyRan, line 288: It seems this use case will use TP in MLP layer as EP, is this right?

If so, I think a better way is to add your mesh as a custom one. We have a few examples, like cp-as-ep (https://github.com/AI-Hypercomputer/maxtext/tree/main/src/maxtext/configs/custom_mesh_and_rule).

Maybe we could avoid more if/else in moe.py file? Happy to chat more!

RissyRan (changes requested):
Thanks for the change! This is a great feature, and suggest to follow up custom_mesh_and_rule example. Happy to chat more. Thank you!

src/maxtext/configs/base.yml
NuojCheng: agree!!

PR #4912, opened 2026-08-17
DSV4: Add mhc_split_axis_contraction to contract the mHC rate and embed axes separately

systalyze-ai:
# Description
Adds `mhc_split_axis_contraction`, an opt-in path that keeps the mHC rate and TP-sharded embedding axes separate through RMS normalization and the fused alpha projection. The split contractions replace flatten-induced activation gathers with reductions over the sharded embedding axis.

When enabled, the checkpoint layout of the mHC parameters changes: the norm scale uses `(k, dim)` instead of `(k * dim,)`, and the alphas use `(k, dim, n)` instead of `(k * dim, n)`. Their initialized values equal the flat initialization reshaped, so converting a flag-off checkpoint is a pure reshape.

The flag defaults to false, preserving the existing RMSNorm module, parameter shapes, RNG draw order, and contraction path. DeepSeek-V4 Hugging Face conversion reshapes alpha weights for either layout, and Muon contracts both split input axes.

# Performance
Step time was 7.711 s/step with `mhc_split_axis_contraction` disabled and 6.811 s/step with it enabled, a 1.13x result.

Reproduction setup: v6e-128, DeepSeek-V4-Flash 284B, LoRA fine-tuning, `ici_expert_parallelism=16`, `ici_tensor_parallelism=8`, `ici_fsdp_parallelism=1`, `max_target_length=16384`, `per_device_batch_size=0.125` (global batch 16), `LIBTPU_INIT_ARGS=--xla_tpu_scoped_vmem_limit_kib=98304`, dynamic-splash attention path enabled, with `mhc_split_axis_contraction` toggled.

# Tests
- `JAX_PLATFORMS=cpu pytest -q tests/unit/mhc_test.py -k split_axis` — 3 passed; the split path matches Sinkhorn and mHC-lite outputs, and its initialized parameters are exact reshapes of the flat layout.
- `JAX_PLATFORMS=cpu pytest -q tests/unit/mhc_test.py` — 27 passed, 2 skipped.
- `JAX_PLATFORMS=cpu pytest -q tests/unit/param_mapping_test.py tests/unit/muon_utils_test.py` — 53 passed; split-layout Hugging Face conversion round-trips and Muon uses the intended contraction axes.

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -1775,6 +1775,14 @@ class ManifoldConstrainedHyperConnections(BaseModel):
       raise ValueError("use_mhc_pallas_kernel=True requires enable_mhc_lite=True.")
     return self
 
+  mhc_split_axis_contraction: bool = Field(
+      False,
+      description=(
+          "Whether to contract the mHC rate and embed axes separately instead of flattening them, "
+          "so the activation's TP-sharded embed dim is never all-gathered."
+      ),
+  )
+
 
 class DilocoParams(BaseModel):
   """Diloco Hyperparameters"""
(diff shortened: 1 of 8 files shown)


gobbleturk:
Does this change only have expected benefits when TP>1?

src/maxtext/configs/types.py
```diff
@@ -1837,32 +1775,25 @@ def validate_mhc_kernel(self) -> "ManifoldConstrainedHyperConnections":
       raise ValueError("use_mhc_pallas_kernel=True requires enable_mhc_lite=True.")
     return self
 
+  mhc_split_axis_contraction: bool = Field(
```
dandragona, line 1778: please update the code to throw an error if you try to use this flag without using tensor parallelism.

PR #4917, opened 2026-08-17
Fix Qwen text-only model padding check in PadOrTrimToMaxLength

subawocit:
# Description

## Problem
PadOrTrimToMaxLength is a shared Grain transform used across all models and datasets. In PadOrTrimToMaxLength.map(), any dataset example containing an "images" key (e.g., from multimodal datasets, shared test suites, or mock batches) will evoke `_pad_image_and_mask` *regardless of the model is text-only or not*.

Previously, PR #4897 added an image padding bypass for Qwen multimodal models (qwen3_vl, qwen3_omni, qwen3_5) with an unregistered model fallback:

```python
elif model_name and model_name.startswith("qwen"):
  raise ValueError(...)
```

Because if a dictionary has an "images" key, it calls `_pad_image_and_mask()` unconditionally. Therefore, `_pad_image_and_mask` itself must properly handle text-only models (vision_block is None or use_multimodal=False). Otherwise, a text-only models that has no vision encoder will fail the bypass check and hit the elif branch, incorrectly raising a ValueError: Qwen multimodal model 'qwen3-0.6b' with vision_block 'None' was not registered

## Fix
In `input_pipeline_utils.py`, allow text-only models (vision_block is None or use_multimodal=False) to bypass image padding without error. Ensure the unregistered Qwen model validation only triggers when the model is actually configured for multimodal training (use_multimodal=True).

## Tests

```
python -m unittest tests/unit/input_pipeline_utils_test.py

----------------------------------------------------------------------
Ran 13 tests in 0.004s

OK
```

```
python -m unittest tests/unit/train_compile_test.py -k 'test_qwen3'

----------------------------------------------------------------------
Ran 3 tests in 174.657s

OK
```

# Checklist

### DIFF
--- a/src/maxtext/input_pipeline/input_pipeline_utils.py
+++ b/src/maxtext/input_pipeline/input_pipeline_utils.py
@@ -795,7 +795,11 @@ class PadOrTrimToMaxLength(grain.MapTransform):
 
     vision_block = mm_processor._get_vision_block(self.config)  # pylint: disable=protected-access
     model_name = getattr(self.config, "model_name", None)
-    if vision_block in ["qwen3_omni", "qwen3_vl", "qwen3_5"]:
+    if (
+        vision_block is None
+        or not getattr(self.config, "use_multimodal", False)
+        or vision_block in ["qwen3_omni", "qwen3_vl", "qwen3_5"]
+    ):
       return preprocessed_image
     elif model_name and model_name.startswith("qwen"):
       raise ValueError(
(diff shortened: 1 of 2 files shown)


hengtaoguo (approved):

NuojCheng (approved):

src/maxtext/input_pipeline/input_pipeline_utils.py
```diff
@@ -795,6 +795,8 @@ def _pad_image_and_mask(self, preprocessed_image: mm_utils.PreprocessorOutput) -
 
     vision_block = mm_processor._get_vision_block(self.config)  # pylint: disable=protected-access
     model_name = getattr(self.config, "model_name", None)
+    if vision_block is None or not getattr(self.config, "use_multimodal", False):
+      return preprocessed_image
```
bvandermoon: Nit: New if statement repeats the same result as the if below. Might be good to merge these

bvandermoon (approved):

src/maxtext/input_pipeline/input_pipeline_utils.py
subawocit: good idea - edited!

PR #4920, opened 2026-08-17
cosmos3_nano model bring up (patch PR#4475)

lydhr:
# Description

This PR adds support for the [nvidia/Cosmos3-Nano](https://huggingface.co/nvidia/Cosmos3-Nano) model (reasoner tower) in MaxText. 

##### Key Changes
- Add Dynamic EOS Truncation in decode.py of maxengine (Applies to other models, too.)
  - Change: Truncate the generated token stream at the tokenizer's `eos_token_id` prior to decoding into text.
  - Motivation: Prevents degenerate filler/hallucinations post-completion (occurred in both the original #4475 and this PR, e.g. the output = `Move the arm ... red bottle. .com .com ... .com `). This mirrors standard behavior in inference engines like `vLLM`, ensuring clean and aligned outputs for evaluation without needing to modify injection/TPU loop logic directly.
- **Parameter Mapping**: Implemented specialized `COSMOS3_MAXTEXT_TO_HF_PARAM_MAPPING`.
  - Instead of maintaining duplicate specialized mappings and hook functions for Cosmos3 in PR#4475, this PR reuses QWEN3_VL_MAXTEXT_TO_HF_PARAM_MAPPING and QWEN3_VL_MAXTEXT_TO_HF_PARAM_HOOK_FN. 
  - A lightweight translation helper is appended to handle Cosmos3-specific naming conventions (like self_attn.to_q), improving maintainability.
- **Vision Integration**: Configured to use the `qwen3_vl` vision encoder block.

##### Reference
- Patched and adapted from an early workstream [#4475](https://github.com/AI-Hypercomputer/maxtext/pull/4475).

---

##### Checkpoint Conversion
```bash
MODEL_NAME=cosmos3-nano-reasoner
python3 -m maxtext.checkpoint_conversion.to_maxtext \
    src/maxtext/configs/base.yml \
    model_name=${MODEL_NAME?} \
    base_output_directory=${MODEL_BUCKET} \
    scan_layers=false \
    use_multimodal=true \
    hf_access_token=${HF_TOKEN?} \
    hardware=cpu \
    skip_jax_distributed_system=True \
    checkpoint_storage_use_ocdbt=False \
    checkpoint_storage_use_zarr3=False \
    --eager_load_method=safetensors \
    --lazy_load_tensors=False
```

##### Decoding / Inference (Robotic Arm Planning)
```bash
python3 -m maxtext.inference.decode \
    src/maxtext/configs/base.yml \
    model_name=${MODEL_NAME?} \
    tokenizer_path=nvidia/Cosmos3-Nano \
    tokenizer_type=huggingface  \
    load_parameters_path=${checkpoint_path} \
    per_device_batch_size=1 \
    run_name=sft_decode \
    scan_layers=false \
    use_multimodal=true \
    prompt='You are a robotic arm planner. The task is to put the flower into the red bottle. Generate a detailed plan consisting of sequential subtasks to accomplish the task.' \
    image_path='tests/assets/test_image_reasoning.jpg' \
    max_prefill_predict_length=1024 \
    max_target_length=1124 \
    ici_tensor_parallelism=4 \
    override_model_config=true \
    attention='dot_product' \
    float32_qk_product=true \
    float32_logits=true \
    hf_access_token=${HF_TOKEN}
```

**Sample Output:**
> Input `<|im_start|>user
<|vision_start|><|image_pad|><|vision_end|>You are a robotic arm planner. The task is to put the flower into the red bottle. Generate a detailed plan consisting of sequential subtasks to accomplish the task.<|im_end|>
<|im_start|>assistant
` -> `Move the arm to the flower. Grasp the flower. Move the arm to the red bottle. Place the flower in the red bottle.`

---

### Checklist
- [x] I have performed a self-review of my code. For an optional AI review, add the gemini-review label.


src/maxtext/configs/models/cosmos3-nano-reasoner.yml
```diff
@@ -0,0 +1,42 @@
...
+num_hidden_layers_for_vit: 27
+num_channels_for_vit: 3
+patch_size_for_vit: 16
+temporal_patch_size_for_vit: 2
+spatial_merge_size_for_vit: 2
+out_hidden_size_for_vit: 4096
+num_position_embeddings_for_vit: 2304
+deepstack_visual_indexes_for_vit: [8, 16, 24]
+
+vision_encoder_block: "qwen3_vl"
```
hengtaoguo, line 41: I like this! We don't have to create qwen3.5/qwen3-omni duplicates like before, just reuse `qwen3_vl` vision encoder. Can we move this line under `# Vision Encoder Configuration`?

src/maxtext/checkpoint_conversion/utils/hf_model_configs.py
```diff
@@ -1883,6 +1883,56 @@ def __init__(self, **kwargs):
 }
 qwen3_vl_30b_a3b_config = PTConfig(**qwen3_vl_30b_a3b_dict)
 
+cosmos3_nano_reasoner_dict = {
```
hengtaoguo, line 1886: A naive question: Does the original cosmos3-nano repo contain diffusion block? If so, I wonder if we should just name it `cosmos3-nano` in maxtext without `reasoner` suffix, since it's a subset of the complete model. So later when you onboard the full model, you don't need to rename this config.

src/maxtext/inference/decode.py
```diff
@@ -225,6 +226,18 @@ def main(argv: Sequence[str]) -> None:
...
+    max_logging.debug(f"raw token ids for stream {i}: {results}")
+
+    eos_token_id = None
+    if hasattr(tokenizer_model, "tokenizer") and hasattr(tokenizer_model.tokenizer, "eos_token_id"):
+      eos_token_id = tokenizer_model.tokenizer.eos_token_id
+
+    max_logging.debug(f"using eos_token_id: {eos_token_id}")
+
+    if eos_token_id is not None and eos_token_id in results:
+      max_logging.debug(f"EOS token found at index {results.index(eos_token_id)}")
```
hengtaoguo: How about a louder callout for the truncation? Something like this:

`EOS token {eos_token_id} found at index {eos_idx}; output sequence truncated to length {eos_idx + 1}.`

hengtaoguo (approved):
Thanks for the groundwork for world model in maxtext!

src/maxtext/checkpoint_conversion/utils/hf_model_configs.py
lydhr: Yes, cosmos3-nano consists of reasoner(this PR) and generator(i.e. diffusion).  
- In generator, its attention jointly takes in the KV of reasoner. 
- But the reasoner does not share this vice versa, which is why this reasoner-only PR works.

I suggest we keep `1) cosmos3-nano-reasoner`, and a `2) cosmos3-nano`
- as 1) might be called independently depending on the downstream tasks.

lydhr:
**Output degraded after rebasing** and upgrading JAX (required by a recent change using a new xla op)
- Investigating....
- Before:
> -> `Move the arm to the flower. Grasp the flower. Move the arm to the red bottle. Place the flower in the red bottle.`


- After: 
> -> `Move the arm to the flower
Move the arm to the red bottle
Place the flower in the red bottle`

------Update-------:
- **Fixed**: temporarily forcing higher precision by flipping these two flags:
  - float32_qk_product: true
  - float32_logits: true

- **Root cause**: I rebased to a recent change that lowers the precision in QK product: 
  -  Lower precision being used in the `apply_attention_dot ` path, leading to numerical truncation in decoding pipelines. 
  
- Note that using float32 for qk slows down the decoding (~12s -> 42s). In details, the overhead of _generate_jit before the loop leads to ~20s of latency.

src/maxtext/inference/decode.py
lydhr: Done. 
Also upgrade this line from logging.debug to logging.info.

src/maxtext/configs/models/cosmos3-nano-reasoner.yml
lydhr: Sure. FYI. I read that in the training of Cosmos3, partial model actually initializes from pre-trained Qwen3-vl, thus it might be highly reusable.

src/maxtext/layers/attentions.py
```diff
@@ -894,7 +894,7 @@ def init_rotary_embedding(self):
     rope_type = self.rope_type
     rope_use_scale = self.config.rope_use_scale
     if self.is_vision:
-      if self.config.model_name.startswith("qwen3"):
+      if self.config.model_name.startswith("qwen3") or self.config.model_name.startswith("cosmos3"):
         rotary_embedding = Qwen3OmniMoeVisionRotaryEmbedding(
             hidden_size=self.config.hidden_size_for_vit,
```
bvandermoon, line 899: Nit: is `Qwen3OmniMoeVisionRotaryEmbedding` still the right name for this?

bvandermoon (approved):

src/maxtext/layers/attentions.py
hengtaoguo: Good call, when we implemented Qwen3-Omni we didn't see it coming that all the following models such as Qwen3-VL/Qwen3.5/Cosmos3 are all sharing the same vision backbone. The function here is legit though. We could follow up with a more generalized rename/refactor.

lydhr: +1 similarly, `Qwen3OmniMoeVisionEncoder` is inherited by `Qwen3OmniMoeVisionEncoder`

PR #4924, opened 2026-08-17
Add HCA Static Compilation for Splash Attention [Deepseek v4]

octatrifan:
# Description
Add static Tokamax Splash Attention compilation for DeepSeek-V4 Heavily Compressed Attention (HCA).

Previously, both Compressed Sparse Attention (CSA, `compress_ratio == 4`) and HCA were dispatched through dynamic Splash Attention (`make_dynamic_splash_mqa`), requiring `DeepseekV4HCACompressor` to materialize dense `(B, 1, S, S_comp)` float mask arrays in HBM and build runtime indexer masks.

Because HCA attention patterns are strictly deterministic (local sliding window + completed preceding compression windows) and do not rely on the indexer (unlike CSA, which requires a top-k step), we can construct the block mask analytically on CPU during AOT tracing. This PR introduces `HCAStaticMask`, bypassing HBM mask allocation and dispatching directly to static Splash MQA kernels (`make_splash_mqa`).

Key changes:
- `HCAStaticMask`: Subclasses `splash_attention_mask.Mask` to evaluate coordinate-based block sparsity on CPU during tracing without runtime HBM mask materialization.
- Compressor bypass: `DeepseekV4HCACompressor` returns `None` for masks when `attention_kernel="flash"`. `generate_attention_mask` is now restricted strictly to CSA.
- MQA kernel dispatch: Routes compressed attention with `num_kv_heads == 1` to `make_splash_mqa`, stripping singleton KV head dimensions for 2D Tokamax input requirements.
- Backward ring buffer: Sets `dq_reduction_steps = 3` for `AttentionType.COMPRESSED` in `create_sa_config` to enable in-SRAM circular ring buffer accumulation for `dQ`. (Without this change, the backward pass is much slower than the dynamic path)
- Arbitrary sequence lengths & packing: Added ceiling division tile padding in `AttentionOp.tpu_flash_attention` for unaligned sequence lengths (e.g. $S=489$) and document packing boundary isolation with segment IDs.
- Validation guards: Added explicit `ValueError` checks blocking Context Parallelism (`cp_size > 1`) and Ulysses/USP when `AttentionType.COMPRESSED` is active, pending future asymmetric communication support.

### Details in b/537346777

# Performance
We conduct microbenchmarks (just on the attention layer) on a v5p TPU for DeepSeek-V4 Flash HCA.
- **Standard Context (S <= 8k)**: Static Flash and Dynamic Flash are on par across all batch sizes, delivering identical step latency while Static Flash avoids allocating dense boolean masks in device memory.
- **Long Context Speedups (S >= 16k)**: Static Flash becomes progressively faster than Dynamic Flash as sequence length increases—delivering a **1.09x** speedup at 16k, **1.20x** at 32k (with a **2.25x** speedup on the forward pass), and **1.48x** at 65k for batch size 1.
- **Memory Footprint**: By compiling mask coordinates analytically at trace-time rather than materializing dynamic boolean mask tensors on device, Static Flash eliminates the O(S^2) mask memory overhead, reducing peak HBM by up to **48%** at 65k. This allows us to scale up to 128k sequence length, where Dynamic Flash fails due to SMEM OOM.
- **Compilation Time Reduction**: When comparing against graph-captured masks, Static Flash drops JIT compilation time from **1.9 minutes to 1.7 seconds (68x reduction)** at 32k and from **7.6 minutes to 1.7 seconds (264x reduction)** at 65k. Under runtime dynamic masking, it eliminates host compiler memory spikes and SMEM lowering failures, compiling in under **5 seconds** across all sequence lengths up to 128k.

# Tests
Tested on Cloud TPU v5p:
```bash
# DeepSeek-V4 unit and reference parity tests
pytest tests/unit/deepseek_v4_vs_reference_test.py -v
# CompressedAttention compilation, numerical equivalence, and packing tests
pytest tests/unit/attention_test.py::CompressedAttentionTest -v
# Static mask bitwise parity, sliding window, and unaligned length tests
pytest tests/unit/attention_test.py::HCAStaticMaskTest -v
```

# Checklist

### DIFF
--- a/src/maxtext/layers/attention_compressed.py
+++ b/src/maxtext/layers/attention_compressed.py
@@ -425,6 +425,38 @@ def prime_prefill_cache_state(
       cache.overlap_gate.set_value(overlap_gate_to_write)
 
 
+def compute_hca_padding(
+    q_len: int,
+    comp_len: int,
+    block_q: int = 128,
+    block_kv: int = 128,
+    kv_len: int | None = None,
+) -> tuple[int, int]:
+  """Computes (pad_q, pad_kv_total) for HCA Flash Attention.
+
+  Args:
+    q_len: Unpadded query sequence length.
+    comp_len: Number of compressed KV blocks appended after the local KV tokens.
+    block_q: Query tile size the kernel is compiled for.
+    block_kv: KV tile size the kernel is compiled for.
+    kv_len: Unpadded local KV sequence length. Defaults to `q_len`, which is the
+      self-attention case DeepSeek-V4 uses; pass it explicitly if the query and
+      KV sequences can differ.
+
+  Returns:
+    A (pad_q, pad_kv_total) tuple.
+  """
+  if kv_len is None:
+    kv_len = q_len
+  pad_q = (block_q - (q_len % block_q)) % block_q
+  total_kv_len = kv_len + comp_len
+  pad_kv_total = (block_kv - (total_kv_len % block_kv)) % block_kv
+  # Ensure padding KV columns exist whenever query padding is required
+  if pad_q > 0 and pad_kv_total == 0:
+    pad_kv_total = block_kv
+  return pad_q, pad_kv_total
+
+
 class BaseDeepseekCompressor(nnx.Module):
   """Shared base class for DeepSeek-V4 long-range attention compressors.
 
@@ -524,6 +556,7 @@ class DeepseekV4HCACompressor(BaseDeepseekCompressor):
       quant: Optional[Quant] = None,
       model_mode: str = MODEL_MODE_TRAIN,
       rngs: Optional[nnx.Rngs] = None,
+      attention_kernel: str = "dot_product",
   ):
     """Initializes the HCA Compressor.
 
@@ -537,6 +570,7 @@ class DeepseekV4HCACompressor(BaseDeepseekCompressor):
       quant: Optional quantization scheme.
       model_mode: The operational mode (e.g., "train", "prefill").
       rngs: An optional Rngs instance for stochastic initializations or dropout.
+      attention_kernel: Attention kernel name ("flash", "dot_product").
     """
     super().__init__(
         config,
@@ -548,6 +582,7 @@ class DeepseekV4HCACompressor(BaseDeepseekCompressor):
         model_mode,
         rngs,
     )
+    self.attention_kernel = attention_kernel
 
   def __call__(
       self,
@@ -648,8 +683,17 @@ class DeepseekV4HCACompressor(BaseDeepseekCompressor):
           cache=cache,
       )
 
-    # Skip causal mask generation during decoding (seq_len == 1) or if no blocks were pooled
-    if seq_len == 1 or compressed_len == 0:
+    # During decoding (seq_len == 1), dot-product attention requires a zero mask
+    if seq_len == 1:
+      compressed_mask = jnp.zeros((batch_size, 1, seq_len, compressed_len), dtype=self.dtype)
+      return compressed_kv, compressed_mask
+
+    # Skip dense causal mask generation when using static flash attention (HCAStaticMask)
+    if self.attention_kernel == "flash":
+      return compressed_kv, None
+
+    # Dot-product fallback when no blocks were pooled
+    if compressed_len == 0:
       compressed_mask = jnp.zeros((batch_size, 1, seq_len, compressed_len), dtype=self.dtype)
       return compressed_kv, compressed_mask
 
@@ -1297,6 +1341,7 @@ class CompressedAttention(Attention):
           quant=self.quant,
           model_mode=self.model_mode,
           rngs=self.rngs,
+          attention_kernel=self.attention_kernel,
       )
     elif self.compress_ratio == 4:
       self.csa_compressor = DeepseekV4CSACompressor(
@@ -1582,9 +1627,14 @@ class CompressedAttention(Attention):
     # Tokamax dynamic splash tile boundary alignment. Note: Tokamax kernel inside AttentionOp additionally
     # sets inner block size as min(block_kv, key_len) during kernel invocation.
     if self.attention_kernel == "flash":
-      total_kv_len = kv.shape[1] + (compressed_kv.shape[1] if compressed_kv is not None else 0)
-      block_size = self.config.sa_block_kv
-      pad_kv_total = (block_size - (total_kv_len % block_size)) % block_size
+      comp_len = compressed_kv.shape[1] if compressed_kv is not None else 0
+      _, pad_kv_total = compute_hca_padding(
+          q_len=inputs_q.shape[1],
+          comp_len=comp_len,
+          block_q=self.config.sa_block_q,
+          block_kv=self.config.sa_block_kv,
+          kv_len=kv.shape[1],
+      )
 
       if pad_kv_total > 0:
         if compressed_kv is not None:
@@ -1608,9 +1658,9 @@ class CompressedAttention(Attention):
     if self.query_pre_attn_scalar and self.query_pre_attn_scalar != 1.0:
       q = q * self.query_pre_attn_scalar
 
-    # Build indexer mask explicitly for tokamax splash kernel
+    # Build indexer mask explicitly for tokamax splash kernel (CSA dynamic path)
     indexer_mask = None
-    if self.attention_kernel == "flash" and compressed_mask is not None:
+    if self.attention_kernel == "flash" and compressed_mask is not None and self.compress_ratio == 4:
       indexer_mask = self.attention_op.generate_attention_mask(
           q,
           unpadded_kv,
@@ -1640,6 +1690,8 @@ class CompressedAttention(Attention):
         cached_values=current_kv_cache,
         indexer_mask=indexer_mask,
         decoder_segment_ids_kv=decoder_segment_ids_kv,
+        pad_kv_total=pad_kv_total,
+        compress_ratio=self.compress_ratio,
     )
 
     # Reverse RoPE on Values
--- a/src/maxtext/layers/attention_op.py
+++ b/src/maxtext/layers/attention_op.py
@@ -86,6 +86,16 @@ from tokamax._src.ops.experimental.tpu.splash_attention import splash_attention_
 
 dynamic_vector_slice_in_dim = jax.vmap(lax.dynamic_slice_in_dim, in_axes=(None, 0, None, None))
 
+# Number of in-kernel dQ reduction steps used for AttentionType.COMPRESSED.
+#
+# The Tokamax Splash fused backward kernel only accepts `dq_reduction_steps` of 3 or None
+# (see SplashConfig.__post_init__); any other value raises at kernel config construction.
+# Selecting 3 enables a 3-slot circular ring buffer for dQ accumulation in SRAM instead of
+# writing one unreduced dQ tile per KV step to HBM, reducing unreduced dQ write traffic by a
+# factor of `kv_steps / 3`. For the DeepSeek-V4 HCA shapes benchmarked here (kv_steps = 128)
+# that is ~42x; the ratio is shape dependent, not a fixed kernel property.
+_COMPRESSED_DQ_REDUCTION_STEPS = 3
+
 
 def _resolve_attention_type(config: Config, attention_type: AttentionType | str | None) -> AttentionType:
   configured_attention_type = AttentionType(getattr(config, "attention_type", AttentionType.GLOBAL.value))
@@ -302,6 +312,126 @@ class BlockCausalMask(splash_attention_mask._ComputableMask):  # pylint: disable
             self.causal_block_size,
             self.q_sequence.tobytes() if self.q_sequence is not None else None,
         )
+    )
+
+
+class HCAStaticMask(splash_attention_mask.Mask):
+  """Static compile-time mask for DeepSeek-V4 Heavily Compressed Attention (HCA).
+
+  Defines coarse-block coordinates and deduplicated partial block patterns for
+  Splash Attention with pre-packed bitmasks loaded into VMEM/SMEM. Evaluates
+  analytical boolean slices on host CPU during compilation, eliminating runtime VPU
+  arithmetic overhead without materializing dense boolean mask arrays in HBM.
+
+  Attributes:
+    shape: (seq_len, aligned_kv_len) tuple.
+    local_kv_len: Unpadded local sequence length.
+    compressed_kv_len: Number of active compressed tokens.
+    pad_kv_total: Prepend tile padding tokens between local and compressed KV.
+    compress_ratio: Compression ratio for HCA (e.g. 128). Must be positive.
+    local_window: Local sliding-window size (e.g. 128).
+  """
+
+  def __init__(
+      self,
+      shape: tuple[int, int],
+      local_kv_len: Optional[int] = None,
+      compressed_kv_len: Optional[int] = None,
+      pad_kv_total: int = 0,
+      compress_ratio: int = 128,
+      local_window: Optional[int] = 128,
+      shard_count: int = 1,
+  ):
+    # shard_count is retained for Splash Attention Mask API compatibility;
+    # HCA static coordinates do not shard local window math across sequence axes.
+    del shard_count
+    if compress_ratio <= 0:
+      raise ValueError(f"compress_ratio must be positive, got {compress_ratio}")
+    self._shape = shape
+    self.local_kv_len = local_kv_len if local_kv_len is not None else shape[0]
+    if self._shape[0] > self.local_kv_len and pad_kv_total <= 0:
+      raise ValueError(
+          f"pad_kv_total must be > 0 when query sequence is padded (shape[0]={self._shape[0]} > local_kv_len={self.local_kv_len})."
+      )
+    self.compressed_kv_len = (
+        compressed_kv_len if compressed_kv_len is not None else max(0, self.local_kv_len // compress_ratio)
+    )
+    self.pad_kv_total = pad_kv_total
+    self.compress_ratio = compress_ratio
+    self.local_window = local_window
+
+  @property
+  def shape(self) -> tuple[int, int]:
+    return self._shape
+
+  def __getitem__(self, idx: tuple[slice, slice]) -> np.ndarray:
+    if len(idx) != 2:
+      raise NotImplementedError(f"Unsupported slice: {idx}")
+    q_slice, kv_slice = idx
+    if not isinstance(q_slice, slice) or not isinstance(kv_slice, slice):
+      raise NotImplementedError(f"Unsupported slice: {idx}")
+
+    q_slice = splash_attention_mask._fill_slice(q_slice, self.shape[0])
+    kv_slice = splash_attention_mask._fill_slice(kv_slice, self.shape[1])
+
+    rows = np.arange(q_slice.start, q_slice.stop, dtype=np.int32)[:, None]
+    cols = np.arange(kv_slice.start, kv_slice.stop, dtype=np.int32)[None, :]
+
+    if rows.size == 0 or cols.size == 0:
+      return np.empty((rows.shape[0], cols.shape[1]), dtype=np.bool_)
+
+    valid_row = rows < self.local_kv_len
+
+    diff = rows - cols
+    if self.local_window is None:
+      local_m = valid_row & (diff >= 0) & (cols < self.local_kv_len)
+    elif self.local_window > 0:
+      local_m = valid_row & (diff >= 0) & (diff < self.local_window) & (cols < self.local_kv_len)
+    else:
+      local_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)
+
+    c_idx = cols - self.local_kv_len - self.pad_kv_total
+    c_thresh = (rows + 1) // self.compress_ratio
+    comp_m = valid_row & (c_idx >= 0) & (c_idx < c_thresh) & (c_idx < self.compressed_kv_len)
+
+    # Map each padded query row 1-to-1 to a distinct padding KV column (softmax denominator = 1.0, logsumexp = 0.0).
+    # This eliminates 0 * inf = NaN in backward dkv without cross-talk or gradient bleed into real tokens.
+    if self.pad_kv_total > 0:
+      pad_cols = self.local_kv_len + ((rows - self.local_kv_len) % self.pad_kv_total)
+      pad_m = (~valid_row) & (cols == pad_cols)
+    else:
+      pad_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)
+
+    return local_m | comp_m | pad_m
+
+  def __eq__(self, other: object):
+    if not isinstance(other, type(self)):
+      return NotImplemented
+    return (
+        self.shape == other.shape
+        and self.local_kv_len == other.local_kv_len
+        and self.compressed_kv_len == other.compressed_kv_len
+        and self.pad_kv_total == other.pad_kv_total
+        and self.compress_ratio == other.compress_ratio
+        and self.local_window == other.local_window
+    )
+
+  def __hash__(self):
+    return hash(
+        (
+            type(self),
+            self.shape,
+            self.local_kv_len,
+            self.compressed_kv_len,
+            self.pad_kv_total,
+            self.compress_ratio,
+            self.local_window,
+        )
+    )
+
+  def __bool__(self) -> bool:
+    raise NotImplementedError(
+        "Conversion to bool is unsupported. Could be caused by using logical" " instead of bitwise operations on masks."
     )
 
 
@@ -1461,6 +1591,8 @@ class AttentionOp(nnx.Module):
       record_max_logits: bool = False,
   ) -> None:
     """Validates runtime constraints for the TPU Tokamax ring path."""
+    if self.attention_type == AttentionType.COMPRESSED:
+      raise ValueError("TPU Tokamax ring attention does not support AttentionType.COMPRESSED (DeepSeek-V4 HCA/CSA).")
     if getattr(self.config, "use_indexer", False) and indexer_mask is None:
       raise ValueError(
           "`indexer_mask` cannot be None when `use_indexer` is True. "
@@ -1489,6 +1621,11 @@ class AttentionOp(nnx.Module):
       record_max_logits: bool = False,
   ) -> None:
     """Validates runtime constraints for the TPU Ulysses path."""
+    if self.attention_type == AttentionType.COMPRESSED:
+      raise ValueError(
+          "TPU Ulysses attention does not support AttentionType.COMPRESSED (DeepSeek-V4 HCA/CSA) "
+          "due to MQA (num_kv_heads=1), asymmetric Q/KV sequence lengths, and dynamic/ragged masking."
+      )
     ulysses_attention.validate_ulysses_runtime(
         model_mode=model_mode,
         previous_chunk=previous_chunk,
@@ -1511,6 +1648,8 @@ class AttentionOp(nnx.Module):
       record_max_logits: bool = False,
   ) -> None:
     """Validates runtime constraints for the TPU USP path."""
+    if self.attention_type == AttentionType.COMPRESSED:
+      raise ValueError("TPU USP attention does not support AttentionType.COMPRESSED (DeepSeek-V4 HCA/CSA).")
     usp_attention.validate_usp_runtime(
         model_mode=model_mode,
         previous_chunk=previous_chunk,
@@ -1538,6 +1677,8 @@ class AttentionOp(nnx.Module):
       compressed_mask: Optional[Array] = None,
       record_max_logits: bool = False,
       decoder_segment_ids_kv: Optional[Array] = None,
+      pad_kv_total: int = 0,
+      compress_ratio: int = 0,
       *,
       qk_product_einsum: Callable[..., Array],
       wv_product_einsum: Callable[..., Array],
@@ -1546,6 +1687,9 @@ class AttentionOp(nnx.Module):
     self.check_attention_inputs(query, key, value)
     length = query.shape[-3]
     target_hardware = self.mesh.devices[(0,) * self.mesh.devices.ndim].platform
+    cp_size = self._context_parallel_size()
+    if self.attention_type == AttentionType.COMPRESSED and cp_size > 1:
+      raise ValueError(f"Context parallelism (cp_size={cp_size}) is not supported with AttentionType.COMPRESSED.")
     if (
         target_hardware == "tpu"
         and tokamax_ring_attention.is_context_parallel_ring_requested(self.config)
@@ -1633,6 +1777,8 @@ class AttentionOp(nnx.Module):
             use_ragged_attention=use_ragged_attention,
             record_max_logits=record_max_logits,
             decoder_segment_ids_kv=decoder_segment_ids_kv,
+            pad_kv_total=pad_kv_total,
+            compress_ratio=compress_ratio,
         )
         if max_logits is not None:
           self.max_logits = nnx.Intermediate(max_logits)
@@ -1814,6 +1960,8 @@ class AttentionOp(nnx.Module):
       use_ragged_attention: bool = False,
       record_max_logits: bool = False,
       decoder_segment_ids_kv: Array | None = None,
+      pad_kv_total: int = 0,
+      compress_ratio: int = 0,
   ) -> tuple[Array, Array]:
     """TPU Flash Attention."""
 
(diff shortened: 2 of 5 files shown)


src/maxtext/layers/attention_op.py
```diff
@@ -260,6 +260,120 @@ def __hash__(self):
...
+    else:
+      local_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)
+
+    c_idx = cols - self.local_kv_len - self.pad_kv_total
+    c_thresh = (rows + 1) // self.compress_ratio
+    comp_m = (c_idx >= 0) & (c_idx < c_thresh) & (c_idx < self.compressed_kv_len)
+
+    # Map each padded query row 1-to-1 to a distinct padding KV column (softmax denominator = 1.0, logsumexp = 0.0).
+    # This eliminates 0 * inf = NaN in backward dkv without cross-talk or gradient bleed into real tokens.
+    if self.pad_kv_total > 0:
```
parambole, line 343: Could you double-check the relationship between `pad_kv_total` and `pad_q` here? As
far as I can tell they're computed independently:

- `pad_q` — from `sa_block_q` and the **query** length, in `tpu_flash_attention`
- `pad_kv_total` — from `sa_block_kv` and **`local_len + comp_len`**, in `attention_compressed.py`

This branch seems to assume the pad region is wide enough to host every padded query
row. If `pad_kv_total` can be 0 while `pad_q > 0`, `pad_m` goes all-False and I'm not
sure what catches those rows.

The case I got to on paper — `compress_ratio=128`, `sa_block_q = sa_block_kv = 128`,
`seq_len=1017`:

```
comp_len     = 7
total_kv_len = 1024   ->  pad_kv_total = 0
pad_q        = 128 - (1017 % 128) = 7
```

Rows 1017–1023 are padded with segment id `-1`, and `local_m` / `comp_m` only select
KV carrying real segment ids, so once splash applies `q_seg == kv_seg` I *think* the
whole row masks out. Would that give an all-`-inf` softmax row, and does the kernel
handle that, or does it NaN into dK/dV? I couldn't run it to find out, and the forward
slice `x[:, :orig_q_len]` wouldn't cover the backward pass either way.

Possible I'm wrong that `pad_kv_total` can ever be 0 — if the config space rules that
out, ignore this. If it doesn't, an assert at construction (`pad_kv_total >= pad_q`)
would make the assumption explicit.

src/maxtext/layers/attention_op.py
```diff
@@ -1835,24 +2000,42 @@ def wrap_ulysses_splash_kernel(single_head_mask):
...
-        mask_shape = (padded_q_len, padded_kv_len)
-      else:
-        mask_shape = (query.shape[2], key.shape[2])  # (q_seq_len, kv_seq_len)
+      splash_q_seq_shards = cp_size
 
       mask_module = tokamax_splash_mask if self.config.use_tokamax_splash else splash_attention_mask
       use_load_balanced_cp = cp_size > 1 and load_balanced_context_parallel
       if self.attention_type == AttentionType.FULL:
         mask = mask_module.FullMask(mask_shape)
+      elif self.attention_type == AttentionType.COMPRESSED:
```
parambole, line 2066: Is `use_tokamax_splash=False` reachable with COMPRESSED? Reading the legacy jax-splash
branch, it looks like this would fall over there:

- **CSA** (`compress_ratio == 4`) sets `mask = None`, which I think ends up in
  `MultiHeadMask(masks=(None,) * query.shape[1])`.
- **HCA** (`compress_ratio > 4`) would hit `make_splash_mha` with `num_kv_heads=1`
  against `num_query_heads=N` — the `k[0]` / `v[0]` MQA squeeze added in this PR only
  covers the tokamax section.

I noticed the new test helpers default the flag to `False`. If the model config always forces it on should we guard it?

```python
if self.attention_type == AttentionType.COMPRESSED and not self.config.use_tokamax_splash:
    raise ValueError(
        "AttentionType.COMPRESSED flash attention requires use_tokamax_splash=True."
    )
```

src/maxtext/layers/attention_compressed.py
```diff
@@ -648,7 +651,10 @@ def hca_compressor_fn(buf_kv, buf_gate):
           cache=cache,
       )
 
-    # Skip causal mask generation during decoding (seq_len == 1) or if no blocks were pooled
+    # Skip causal mask generation when using static flash attention (HCAStaticMask),
+    # during decoding (seq_len == 1), or if no blocks were pooled.
+    if self.attention_kernel == "flash":
```
parambole: Nit: this returns before the `seq_len == 1 or compressed_len == 0` fallback
and doesn't look at `model_mode`.

`apply_attention` only routes to `tpu_flash_attention` when `target_hardware == "tpu"`;
off TPU the autoregressive path falls back to `apply_attention_dot`, which would call
`generate_attention_mask(compressed_mask=None)` → `c_len = 0` → `s_len = kv_seq_len`,
treating the compressed and `pad_kv_total` columns as ordinary local tokens. The shape
stays valid, so it'd be wrong rather than loud.

That may be a path nobody exercises for this model — if so, no action needed. If it is
live, gating on `model_mode == MODEL_MODE_TRAIN` (or moving the return below the
`seq_len == 1` branch) would keep the skip scoped to the path that actually builds
`HCAStaticMask`.

tests/unit/attention_test.py
```diff
@@ -614,6 +625,231 @@ def test_tpu_splash_selects_block_causal_masks(self):
...
+        num_kv_heads=1,
+        max_target_length=max_target_length,
+        mesh=mesh,
+        attention_kernel="dot_product",
+        attention_type=AttentionType.COMPRESSED,
+        sliding_window_size=sliding_window_size,
+    )
+
+  def test_hca_static_mask_matches_compressor_mask(self):
+    test_cases = [
```
parambole, line 908: I think every case here uses `shape = (seq_len, seq_len + pad_kv + comp_len)` with
`local_kv_len = seq_len`, so `rows` never reaches `local_kv_len` and the
`rows >= self.local_kv_len` predicate in `pad_m` stays False throughout — including
`(489, 128, 128, 23)`, which looks like it was meant to cover that.

If that's right, the padded-query-row mapping is only covered by the
`@pytest.mark.tpu_only` tests, which CI won't run. Two cases would close it:

```python
# padded query rows actually present (shape[0] = 512, local_kv_len = 489)
(489, 128, 128, 23),
# pad region too small to host them
(1017, 128, 128, 0),
```

plus an assertion that every row has at least one `True`

parambole (commented):
Nice work on this — the `HCAStaticMask` approach and the upfront block padding.

A few questions from reading through it. Caveat up front: this is a static read, I
haven't run anything on TPU, so the first two may well be non-issues if the configs
that would trigger them can't occur in practice.

| # | Kind | Location | Question |
|---|---|---|---|
| 1 | Please check | `attention_op.py` — `HCAStaticMask.__getitem__` | Can `pad_kv_total` be smaller than `pad_q`? |
| 2 | Please check | `attention_op.py` — COMPRESSED mask branch | Is `use_tokamax_splash=False` reachable here? |
| 3 | Question | `attention_compressed.py` — flash early return | Does the non-TPU AR fallback matter for this model? |
| 4 | Suggestion | `attention_test.py` — `HCAStaticMaskTest` | Test cases don't seem to reach the `pad_m` branch |

src/maxtext/layers/attention_op.py
octatrifan: I also redid the calculation on paper and I am spotting the same edgecase you are pointing out.  When the query sequence is unaligned (e.g. $S=1017 \to 1024$), padded query rows receive `q_seg = -1`.  Then the effective mask is evaluated as `mask & (q_seg == kv_seg)`. If `pad_kv_total == 0`, all KV tokens have `kv_seg >= 0`, which makes `q_seg == kv_seg` evaluate to `False` across the entire row. And that eventually causes division by 0 and NaNs.

To fix this, I added the following check in `attention_compressed.py` to explicitly guarantee at least one block of KV padding whenever query padding is active:
```python
block_q = self.config.sa_block_q
pad_q = (block_q - (inputs_q.shape[1] % block_q)) % block_q
if pad_q > 0 and pad_kv_total == 0:
  pad_kv_total = block_size

File: src/maxtext/layers/attention_op.py (hunk 2 above)
Comment (octatrifan) [reply src/maxtext/layers/attention_op.py:2066]: In practice, we already enforce this at the config validation level in [configs/types.py#L4069-L4078](https://github.com/AI-Hypercomputer/maxtext/blob/dc2fdbb9da19e37c931e8b9a6f8a99456b42d8d9/src/maxtext/configs/types.py#L4069-L4078).


However, adding the check directly inside `AttentionOp.tpu_flash_attention` (`attention_op.py`) is still valuable for unit tests & microbenchmarks (as they often instantiate `AttentionOp` or `CompressedAttention` directly).
We added the guard in `AttentionOp`:
```python
elif self.attention_type == AttentionType.COMPRESSED:
  if not self.config.use_tokamax_splash:
    raise ValueError("AttentionType.COMPRESSED flash attention requires use_tokamax_splash=True.")

File: src/maxtext/layers/attention_compressed.py (hunk 3 above)
Comment (octatrifan) [reply src/maxtext/layers/attention_compressed.py]: I swapped the two blocks so the `seq_len == 1 or compressed_len == 0` check runs before the flash bypass. 
Decode mode (`seq_len == 1`) now always returns the zero-mask before any flash skip, so the fallback to `apply_attention_dot` stays safe and correctly does `s_len = kv_seq_len - c_len`.

File: tests/unit/attention_test.py (hunk 4 above)
Comment (octatrifan) [reply tests/unit/attention_test.py:908]: We updated `test_hca_static_mask_matches_compressor_mask`:
- Parameterized test cases with `padded_q_len` so `shape = (padded_q_len, total_kv_len)`:
  - `(489, 128, 128, 23, 512)`: Query padded to 512, exercising `rows >= self.local_kv_len` on CPU.
  - `(1017, 128, 128, 128, 1024)`: Query padded to 1024 with the 1-block KV padding fix.
- Added `self.assertTrue(np.all(np.any(actual_mask, axis=-1)))` over all test cases.
- Added `test_hca_static_mask_raises_on_unpadded_kv_with_padded_query` verifying `HCAStaticMask` rejects `(1017, pad_kv=0)` with `ValueError` (since padded queries without KV padding cause backward NaNs).

File: src/maxtext/layers/attention_op.py [hunk 5]
```diff
@@ -261,6 +261,124 @@ def __hash__(self):
...
+    if compress_ratio <= 0:
+      raise ValueError(f"compress_ratio must be positive, got {compress_ratio}")
+    self._shape = shape
+    self.local_kv_len = local_kv_len if local_kv_len is not None else shape[0]
+    if self._shape[0] > self.local_kv_len and pad_kv_total <= 0:
+      raise ValueError(
+          f"pad_kv_total must be > 0 when query sequence is padded (shape[0]={self._shape[0]} > local_kv_len={self.local_kv_len})."
+      )
+    self.compressed_kv_len = (
+        compressed_kv_len if compressed_kv_len is not None else max(1, self.local_kv_len // compress_ratio)
```
shuningjin: `max(1, self.local_kv_len // compress_ratio)` forces compressed_kv_len = 1 for sequences shorter than compress_ratio when compressed_kv_len is None.

Should be `max(0, self.local_kv_len // compress_ratio)`?

tests/unit/attention_test.py
```diff
@@ -5008,7 +5322,205 @@ def test_compressed_attention_flash_vs_dot_product(self, compress_ratio):
...
+    out_flash, _ = attn_flash(
+        x, x, decoder_segment_ids=seg, inputs_positions=pos, deterministic=True, model_mode=MODEL_MODE_TRAIN
+    )
+    np.testing.assert_allclose(np.array(out_flash), np.array(out_dot), rtol=1e-2, atol=1e-2)
+
+    grad_dot = jax.grad(loss_dot)(x)
+    grad_flash = jax.grad(loss_flash)(x)
+
+    self.assertTrue(np.all(np.isfinite(np.array(grad_dot))), "Dot grad contains NaN/Inf for unpacked S=489")
+    self.assertTrue(np.all(np.isfinite(np.array(grad_flash))), "Flash grad contains NaN/Inf for unpacked S=489")
```
shuningjin, line 5801: check gradient closeness via `assert_allclose`?

src/maxtext/layers/attention_op.py
```diff
@@ -1427,6 +1556,13 @@ def apply_attention(
     self.check_attention_inputs(query, key, value)
     length = query.shape[-3]
     target_hardware = self.mesh.devices[(0,) * self.mesh.devices.ndim].platform
+    cp_size = (
```
shuningjin: nit: `cp_size = self._context_parallel_size()`

tests/unit/attention_test.py
```diff
@@ -5008,7 +5322,205 @@ def test_compressed_attention_flash_vs_dot_product(self, compress_ratio):
     out_flash = self._run_compressed_attention(compress_ratio, "flash")
     np.testing.assert_allclose(np.array(out_flash), np.array(out_dot), rtol=1e-2, atol=1e-2)
 
-  def _get_test_config(self, max_target_length, compress_ratio, attention_kernel):
+  @pytest.mark.tpu_only
+  def test_hca_flash_vs_dot_product_unaligned_489(self):
```
shuningjin: Is this the same as `test_hca_flash_vs_dot_product_unaligned_grads` case 1 forward check? If so, maybe remove?

src/maxtext/layers/attention_op.py
```diff
@@ -261,6 +261,124 @@ def __hash__(self):
...
+    if self.local_window is None:
+      local_m = (diff >= 0) & (cols < self.local_kv_len)
+    elif self.local_window > 0:
+      local_m = (diff >= 0) & (diff < self.local_window) & (cols < self.local_kv_len)
+    else:
+      local_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)
+
+    c_idx = cols - self.local_kv_len - self.pad_kv_total
+    c_thresh = (rows + 1) // self.compress_ratio
+    comp_m = (c_idx >= 0) & (c_idx < c_thresh) & (c_idx < self.compressed_kv_len)
```
shuningjin: ### Potential Issue

`local_m` and `comp_m` do not check `rows < self.local_kv_len`. For unaligned sequences with padded query rows (`rows >= self.local_kv_len`), `diff = rows - cols` falls within `[0, local_window)` for trailing real KV tokens, and `c_thresh = (rows + 1) // compress_ratio` activates preceding compressed KV blocks.

Instead of a 1-to-1 dummy connection to a padding KV column (softmax denominator = 1.0, logsumexp = 0.0), padded queries attend directly to real and compressed tokens, leaking 96.9% of their attention weight into real data and corrupting backward normalizers.

###  Example 

$S=489$, Padded to $512$, Window $128$, Ratio $128$

* Setup: Query padded to $512$ (`rows 489..511` are dummy padding). KV length is $512$ ($489$ real + $20$ pad + $3$ compressed).
* Row 489 Mask Evaluation:
  * Expected active columns: `1` (only dummy pad column `489`)
  * Actual active columns: `131` ($127$ real tokens in cols `362..488` via `local_m`, $3$ compressed tokens in cols `509..511` via `comp_m`, $1$ pad token in col `489` via `pad_m`)
* Softmax Distribution on Row 489 (when `decoder_segment_ids=None`):
  * Probability on dummy pad column `489`: `0.007634` (Expected: `1.000000`)
  * Leaked probability on real tokens `0..488`: `0.969466` (Expected: `0.000000`)


### Proposed Fix

gate `local_m`, `comp_m`, and `pad_m` by `valid_row = rows < self.local_kv_len`:

```python
    valid_row = rows < self.local_kv_len

    diff = rows - cols
    if self.local_window is None:
      local_m = valid_row & (diff >= 0) & (cols < self.local_kv_len)
    elif self.local_window > 0:
      local_m = valid_row & (diff >= 0) & (diff < self.local_window) & (cols < self.local_kv_len)
    else:
      local_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)

    c_idx = cols - self.local_kv_len - self.pad_kv_total
    c_thresh = (rows + 1) // self.compress_ratio
    comp_m = valid_row & (c_idx >= 0) & (c_idx < c_thresh) & (c_idx < self.compressed_kv_len)

    if self.pad_kv_total > 0:
      pad_cols = self.local_kv_len + ((rows - self.local_kv_len) % self.pad_kv_total)
      pad_m = (~valid_row) & (cols == pad_cols)
    else:
      pad_m = np.zeros((rows.shape[0], cols.shape[1]), dtype=np.bool_)

    return local_m | comp_m | pad_m
```

src/maxtext/layers/attention_compressed.py
```diff
@@ -648,11 +651,15 @@ def hca_compressor_fn(buf_kv, buf_gate):
...
 
-    # Skip causal mask generation during decoding (seq_len == 1) or if no blocks were pooled
+    # During decoding (seq_len == 1) or if no blocks were pooled, return a zero mask
     if seq_len == 1 or compressed_len == 0:
       compressed_mask = jnp.zeros((batch_size, 1, seq_len, compressed_len), dtype=self.dtype)
       return compressed_kv, compressed_mask
 
+    # Skip dense causal mask generation when using static flash attention (HCAStaticMask)
+    if self.attention_kernel == "flash":
+      return compressed_kv, None
```
shuningjin, line 693: move `if self.attention_kernel == "flash"` before `if seq_len == 1 or compressed_len == 0`? otherwise short seq with `compressed_len == 0` still have compressed_mask

shuningjin (commented):
Thanks for the careful design to eliminate dense mask materialization in device memory! Some comments on fix and cleanup.

src/maxtext/layers/attention_op.py
octatrifan: Correct, updated to `max(0, self.local_kv_len // compress_ratio)`

src/maxtext/layers/attention_op.py
octatrifan: Added `cp_size = self._context_parallel_size()`

tests/unit/attention_test.py
octatrifan: Added `np.testing.assert_allclose(np.array(grad_flash), np.array(grad_dot), rtol=1e-2, atol=1e-6)`

tests/unit/attention_test.py
octatrifan: Removed test as the case 1 forward already checks it.

src/maxtext/layers/attention_op.py
octatrifan: Applied the `valid_row` gating to `local_m`, `comp_m`, and `pad_m` so dummy rows strictly isolate to their 1-to-1 padding column. Also added unit test assertions to verify no weight leaks to real tokens on padded query rows.

src/maxtext/layers/attention_compressed.py
octatrifan: Split the checks so `seq_len == 1` stays handled for dot-product decode fallback, while flash attention always returns None for the causal mask regardless of `compressed_len`.

src/maxtext/layers/attention_compressed.py
```diff
@@ -1586,6 +1606,12 @@ def __call__(
       block_size = self.config.sa_block_kv
       pad_kv_total = (block_size - (total_kv_len % block_size)) % block_size
 
+      # Guarantee padding KV columns exist whenever query padding is required
+      block_q = self.config.sa_block_q
+      pad_q = (block_q - (inputs_q.shape[1] % block_q)) % block_q
+      if pad_q > 0 and pad_kv_total == 0:
+        pad_kv_total = block_size
```
shuningjin: refactor: would prefer use a helper function, which can be reused in code and tests

```python
# here
      comp_len = compressed_kv.shape[1] if compressed_kv is not None else 0
      pad_q, pad_kv_total = compute_hca_padding(
          q_len=kv.shape[1],
          comp_len=comp_len,
          block_q=self.config.sa_block_q,
          block_kv=self.config.sa_block_kv,
      )
```

```python
# helper function
def compute_hca_padding(
    q_len: int,
    comp_len: int,
    block_q: int = 128,
    block_kv: int = 128,
) -> tuple[int, int]:
  """Computes (pad_q, pad_kv_total) for HCA Flash Attention."""
  pad_q = (block_q - (q_len % block_q)) % block_q
  total_kv_len = q_len + comp_len
  pad_kv_total = (block_kv - (total_kv_len % block_kv)) % block_kv
  # Ensure padding KV columns exist whenever query padding is required
  if pad_q > 0 and pad_kv_total == 0:
    pad_kv_total = block_kv
  return pad_q, pad_kv_total
```

tests/unit/attention_test.py
```diff
@@ -818,6 +873,262 @@ def test_tpu_splash_skips_load_balancing_when_flag_is_off(self):
...
+        sliding_window_size=sliding_window_size,
+    )
+
+  def test_hca_static_mask_matches_compressor_mask(self):
+    test_cases = [
+        (512, 128, 128, 0, 512),
+        (1024, 128, 128, 64, 1024),
+        (512, 128, None, 0, 512),  # full causal local attention
+        (384, 128, 128, 128, 384),
+        (489, 128, 128, 23, 512),  # unaligned sequence length 489 with query padded to 512
```
shuningjin: refactor: these test cases needs human calculation. E.g., (23, 512) looks complex, at actual runtime it is pad_kv=20 rather than 23. can use aforementioned helper `attention_compressed.compute_hca_padding` for simplification and fidelity

```python
@@ -906,23 +906,31 @@ class AttentionTest(parameterized.TestCase):
 
   def test_hca_static_mask_matches_compressor_mask(self):
     test_cases = [
-        (512, 128, 128, 0, 512),
-        (1024, 128, 128, 64, 1024),
-        (512, 128, None, 0, 512),  # full causal local attention
-        (384, 128, 128, 128, 384),
-        (489, 128, 128, 23, 512),  # unaligned sequence length 489 with query padded to 512
-        (1017, 128, 128, 128, 1024),  # S=1017 edge case with query padded to 1024 and 1 block KV padding
+        # (seq_len, compress_ratio, local_window)
+        (512, 128, 128),
+        (1024, 128, 128),
+        (512, 128, None),  # full causal local attention
+        (384, 128, 128),
+        (489, 128, 128),   # unaligned S=489 (pad_q=23, pad_kv=20)
+        (1017, 128, 128),  # unaligned S=1017 (pad_q=7, pad_kv=121)
     ]
+    block_q = 128
+    block_kv = 128
 
-    for seq_len, compress_ratio, local_window, pad_kv, padded_q_len in test_cases:
-      with self.subTest(seq_len=seq_len, ratio=compress_ratio, window=local_window, pad=pad_kv, padded_q=padded_q_len):
+    for seq_len, compress_ratio, local_window in test_cases:
+      with self.subTest(seq_len=seq_len, ratio=compress_ratio, window=local_window):
         comp_len = max(0, seq_len // max(1, compress_ratio))
+        pad_q, pad_kv = attention_compressed.compute_hca_padding(
+            q_len=seq_len,
+            comp_len=comp_len,
+            block_q=block_q,
+            block_kv=block_kv,
+        )
+        padded_q_len = seq_len + pad_q
         total_kv_len = seq_len + pad_kv + comp_len
         shape = (padded_q_len, total_kv_len)
 
         mask = attention_op.HCAStaticMask(
```

```python
@@ -968,14 +976,21 @@ class AttentionTest(parameterized.TestCase):
 
   def test_hca_static_mask_padded_query_rows_mapping(self):
-    """Verifies that padded query rows (rows >= local_kv_len) explicitly reach and map to pad_cols via pad_m."""
+    """Verifies that padded query rows map circularly to pad_cols without bleeding into real tokens."""
     seq_len = 489
-    pad_q_len = 512
-    pad_kv = 23
-    comp_len = seq_len // 128
+    compress_ratio = 128
+    comp_len = seq_len // compress_ratio
+    pad_q, pad_kv = attention_compressed.compute_hca_padding(
+        q_len=seq_len,
+        comp_len=comp_len,
+        block_q=128,
+        block_kv=128,
+    )
+    pad_q_len = seq_len + pad_q
     total_kv_len = seq_len + pad_kv + comp_len
     mask = attention_op.HCAStaticMask(
         shape=(pad_q_len, total_kv_len),
         local_kv_len=seq_len,
         compressed_kv_len=comp_len,
         pad_kv_total=pad_kv,
-        compress_ratio=128,
+        compress_ratio=compress_ratio,
         local_window=128,
     )
     actual_mask = mask[:, :]
     for row in range(seq_len, pad_q_len):
       expected_col = seq_len + ((row - seq_len) % pad_kv)
       self.assertTrue(
           actual_mask[row, expected_col],
           f"Padded query row {row} must map to padding KV column {expected_col} in pad_m.",
       )
+      self.assertFalse(
+          np.any(actual_mask[row, :seq_len]),
+          f"Padded query row {row} must not attend to real local KV tokens.",
+      )
+      self.assertFalse(
+          np.any(actual_mask[row, seq_len + pad_kv :]),
+          f"Padded query row {row} must not attend to compressed KV tokens.",
+      )
```

tests/unit/attention_test.py
```diff
@@ -818,6 +873,262 @@ def test_tpu_splash_skips_load_balancing_when_flag_is_off(self):
...
+            attention_op,
+            "HCAStaticMask",
+            wraps=attention_op.HCAStaticMask,
+        ) as mock_mask,
+    ):
+      op.tpu_flash_attention(query, key, key, decoder_segment_ids=None, compress_ratio=128)
+      mock_mask.assert_called_once()
+      self.assertEqual(mock_mask.call_args.kwargs["compress_ratio"], 128)
+
+  def test_packed_positions_and_segment_ids_vs_generate_attention_mask(self):
```
shuningjin, line 1128: Test coverage: Incomplete Document Packing Mask Test (`test_packed_positions_and_segment_ids_vs_generate_attention_mask`)
* **Issue**: Key length is 516, but only 512 segment IDs are passed without `decoder_segment_ids_kv`. `generate_attention_mask` skips document masking on compressed columns `512..515`, leaving cross-doc compressed attention unmasked. Block 1 (`128..255`) straddles both docs but is not invalidated (`seg = -1`), and columns `512..515` are never asserted.
* **Fix**: Pass `decoder_segment_ids_kv = [seg, [1, -1, 2, 2]]` and assert cross-doc isolation on compressed columns `512..515` (including verifying that straddling Block 1 is fully masked).

```python
     comp_len = total_len // compress_ratio  # 4
     op = self._make_dot_product_op(total_len, sliding_window_size=128)
 
     pos = jnp.concatenate([jnp.arange(l1, dtype=jnp.int32), jnp.arange(l2, dtype=jnp.int32)], axis=0)[None, :]
     seg = jnp.concatenate([jnp.ones(l1, dtype=jnp.int32), jnp.full(l2, 2, dtype=jnp.int32)], axis=0)[None, :]
 
+    # Derive compressed segment IDs matching CompressedAttention runtime logic:
+    # Block 0: 0..127 (Doc 1) -> seg 1
+    # Block 1: 128..255 (Doc 1 & Doc 2) -> seg -1 (straddling block invalidated)
+    # Block 2: 256..383 (Doc 2) -> seg 2
+    # Block 3: 384..511 (Doc 2) -> seg 2
+    chunked_seg = seg.reshape((1, comp_len, compress_ratio))
+    min_seg = jnp.min(chunked_seg, axis=-1)
+    max_seg = jnp.max(chunked_seg, axis=-1)
+    comp_seg = jnp.where(min_seg == max_seg, min_seg, -1)
+    seg_kv = jnp.concatenate([seg, comp_seg], axis=1)
+
     usable_len = comp_len * compress_ratio
     block_positions = pos[:, :usable_len:compress_ratio]
     future_mask = (block_positions[:, None, None, :] + compress_ratio) > (pos[:, None, :, None] + 1)
     compressed_causal_mask = jnp.where(future_mask, DEFAULT_MASK_VALUE, 0.0)
 
     dummy_q = jnp.zeros((1, total_len, 1, 128))
     dummy_k = jnp.zeros((1, total_len + comp_len, 1, 128))
 
     mask = op.generate_attention_mask(
         dummy_q,
         dummy_k,
         decoder_segment_ids=seg,
+        decoder_segment_ids_kv=seg_kv,
         segment_positions=pos,
         model_mode=MODEL_MODE_TRAIN,
         compressed_mask=compressed_causal_mask,
         pad_kv_total=0,
     )
     mask_matrix = np.squeeze(np.array(mask == 0.0))
 
-    # Assert Doc 1 (rows 0..199) cannot attend to Doc 2 (cols 200..511)
+    # 1. Uncompressed local token cross-document isolation
     self.assertFalse(np.any(mask_matrix[:l1, l1:total_len]))
-    # Assert Doc 2 (rows 200..511) cannot attend to Doc 1 (cols 0..199)
     self.assertFalse(np.any(mask_matrix[l1:total_len, :l1]))
+
+    # 2. Compressed block cross-document isolation
+    # Doc 1 (rows 0..199) cannot attend to Block 1 (straddled) or Blocks 2, 3 (Doc 2)
+    self.assertFalse(np.any(mask_matrix[:l1, total_len + 1 :]))
+    # Doc 2 (rows 200..511) cannot attend to Block 0 (Doc 1) or Block 1 (straddled)
+    self.assertFalse(np.any(mask_matrix[l1:total_len, total_len : total_len + 2]))
+
+    # 3. Straddled boundary isolation: no query from any document can attend to Block 1
+    self.assertFalse(np.any(mask_matrix[:, total_len + 1]))
+
+    # 4. Valid compressed connections
+    self.assertTrue(np.any(mask_matrix[:l1, total_len]))
+    self.assertTrue(np.any(mask_matrix[l1:total_len, total_len + 2 :]))
```

shuningjin (commented):
Left some minor comments.

src/maxtext/layers/attention_op.py
```diff
@@ -1951,7 +2111,14 @@ def create_sa_config(config, query, key, attn_logits_soft_cap):
...
             else None,
-            dq_reduction_steps=config.dq_reduction_steps if config.dq_reduction_steps > 0 else None,
+            # Tokamax Splash backward kernel requires dq_reduction_steps in (3, None).
+            # Setting dq_reduction_steps=3 for AttentionType.COMPRESSED enables in-kernel
+            # circular ring buffer gradient accumulation, cutting unreduced dQ HBM write traffic by ~42x.
+            dq_reduction_steps=(
+                config.dq_reduction_steps
+                if config.dq_reduction_steps is not None and config.dq_reduction_steps > 0
+                else (3 if self.attention_type == AttentionType.COMPRESSED else None)
+            ),
```
parambole: The comment above it says *"Tokamax Splash backward kernel requires dq_reduction_steps in
(3, None)"*. If that's a hard kernel constraint, then this code has a hole: a user who sets
`dq_reduction_steps: 2` in their YAML gets it passed straight through to a kernel that the comment
says can't accept it. The branch only supplies the safe default when the config is unset.

Two things I'd change:

```python
if self.attention_type == AttentionType.COMPRESSED and config.dq_reduction_steps not in (3, -1, None):
  raise ValueError(
      f"AttentionType.COMPRESSED requires dq_reduction_steps in (3, unset), got {config.dq_reduction_steps}."
  )
```

and pull the `3` out into a module constant (`_COMPRESSED_DQ_REDUCTION_STEPS = 3`) with the "~42x
less unreduced dQ HBM write traffic" note attached to it. Right now that number is discoverable
only by reading this expression.

src/maxtext/layers/attention_op.py
```diff
@@ -2013,24 +2180,44 @@ def wrap_ulysses_splash_kernel(single_head_mask):
...
+      elif self.attention_type == AttentionType.COMPRESSED:
+        if not self.config.use_tokamax_splash:
+          raise ValueError("AttentionType.COMPRESSED flash attention requires use_tokamax_splash=True.")
+        if compress_ratio is None or compress_ratio <= 0:
+          raise ValueError("compress_ratio must be provided for AttentionType.COMPRESSED flash attention.")
+        if compress_ratio > 4:
+          if cp_size > 1:
+            raise NotImplementedError(
+                "Context parallelism is currently not implemented for DeepSeek-V4 HCA Flash Attention."
+            )
```
parambole: **`attention_compressed.py:1171-1176`, `attention_op.py:1559-1565`, `attention_op.py:2072-2075`**

The same restriction is now expressed three times, with three different conditions:

| Location | Condition | Error type |
|---|---|---|
| `CompressedAttention.__init__` | `cp_size > 1 and compress_ratio > 0` | `ValueError` |
| `AttentionOp.apply_attention` | `attention_type == COMPRESSED and cp_size > 1` | `ValueError` |
| `tpu_flash_attention`, HCA branch | `compress_ratio > 4 and cp_size > 1` | `NotImplementedError` |

The third is **unreachable** — `apply_attention` raises before `tpu_flash_attention` is ever
called, for the same input. I'd delete it.

The first two also disagree on how they phrase the rule (`compress_ratio > 0` vs
`attention_type == COMPRESSED`). They happen to coincide today only because `compress_ratio == 0`
rewrites `attention_type` to `LOCAL_SLIDING` a few lines earlier. That's a real coupling and it
isn't written down anywhere.

Suggestion: keep the `apply_attention` guard as the single runtime check, and move the
construction-time one up into `configs/types.py` validation. A config error is much friendlier
than a module constructor throwing during AOT compile / `train_compile.py`.

src/maxtext/layers/attention_compressed.py
octatrifan: Done, layer and tests both go through it now.

tests/unit/attention_test.py
octatrifan: Fixed both, took the negative assertions too. The tests pull their shapes from the same helper the layer uses now, so they won't catch a bug in it, so added a direct test for the helper instead.

tests/unit/attention_test.py
octatrifan: Confirmed, fixed. Test only though, production already passes the full width
decoder_segment_ids_kv and marks blocks spanning two docs as -1.
Added two positive controls as well, an all masked matrix passes all three negative assertions otherwise.

src/maxtext/layers/attention_op.py
octatrifan: Done, third one deleted and the constructor one moved into types.py.

Predicate needs more than attention_type though:
```
    attention_type == compressed
    and any(ratio > 0 for ratio in compress_ratios)
    and context_parallel_size > 1
```

compress_ratios is per layer, so all zero ratios plus cp is legal today.

I believe ordering's backwards from what you wrote, the constructor guard ran first so apply_attention is the one that never fired.

src/maxtext/layers/attention_op.py
octatrifan: Added the constant, documented as kv_steps/3 rather than 42x since that's just kv_steps=128 for
our shapes.
Went with (0, 3) not (3, -1, None) because 0 is the default in base.yml and the ring/usp/ulysses
checks all read it as unset

shuningjin (approved):
Thanks for your optimization, @octatrifan! Excellent work 🔥

parambole (approved):
LGTM Great work 🚀

PR #4925, opened 2026-08-18
Step0 reforward test scripts

entrpn:
# Description

the past.

FIXES: b/123456
FIXES: #123456

BUGS: b/123456

# Tests

commands to reproduce.

# Checklist


wenxindongwork:
thank you for these scripts! when I tried to repro these numbers I bumped into these issues, might worth double checking since they were detected by AI. 

1. The PR's weight-copy helper skips one weight 
The test builds a training layer and an inference layer, then copies weights from one to the other so they're identical. The copy function checks .kernel/.bias/.scale on each object — but the MoE down-projection wo is a bare parameter, not a module, so it silently copies nothing. The inference layer keeps its own random wo. Naturally its output is unrelated to the training layer's.  

2. The PR's full-model script doesn't sync expert weights  
For the "top-1 agreement degrades with depth" table, it syncs with nnx.update. The training model stores experts as wi_0 and wi_1; the inference model stores one fused wi. nnx.update doesn't complain — it just attaches wi_0/wi_1 as new attributes and leaves wi untouched at its random init. So the two models have different expert weights, and of course they disagree more the deeper you go. That table doesn't measure kernel drift; it's invalid.

3. A real bug in expert weights 
When MaxText hands expert weights to the inference kernel, it concatenates gate and up projections as [gate | up]. The kernel then splits that axis across TP devices. With TP=2, device 0 gets only gate columns and device 1 gets only up columns, and each then tries to split its half into gate/up again. Wrong on every device. The correct layout is per-device [gate_0|up_0][gate_1|up_1]. MaxText already does this correctly when loading a pre-fused checkpoint, but not in the code path the PR's 1-layer test uses (prefuse_moe_weights=False, attention=vllm_rpa, TP>1). I confirmed this by calling the kernel directly with no MaxText: TP=1 → correct; TP=2 with MaxText's layout → garbage; TP=2 with the per-device layout → correct. Any real serving config that takes that path would produce wrong outputs.

Two smaller issues worth knowing

- The PR's "35B" numbers aren't 35B. The 1-layer doc says qwen3.5-35b-a3b but the script overrides it to a toy: 2048 dim, 8 experts with top-8 routing (i.e. dense — every token uses every expert, so routing bugs are invisible).
- Random init makes the routed experts nearly silent. MaxText's init scales the expert weights ~20× smaller than the shared expert, so the routed-MoE output is ~1e-5 while everything else is ~1. The layer output looks fine (T25 CosSim 0.9999) even when the routed experts are total garbage, because they barely contribute. A parity test with random weights isn't really testing the experts at all.

PR #4930, opened 2026-08-18
Build the Qwen3-Next shared expert only when shared_experts > 0

WandLZhang:
`Qwen3NextSparseMoeBlock` builds a shared expert and its gate in every layer. It doesn't read `cfg.shared_experts`, which defaults to `0` in `base.yml`.

A dense configuration therefore carries one unrequested `MlpBlock` per layer. The parameter count rises and the extra weights receive gradients.

## Effect

Measured on a Qwen 3.5 dense configuration, 32 layers, `emb_dim` 4096, `moe_mlp_dim` 12288:

| | Parameters |
|---|---|
| Reported by MaxText | 11.756B |
| Published for this architecture | 6.924B |

The 4.8B difference is one unrequested MLP per layer. At 11.756B the model doesn't fit a v6e-8: HLO temporaries reach 40.50 GB against 31.24 GB of HBM. With this change the same configuration reports 6.924B and trains on 8 chips.

## Scope

The block now builds the shared expert only when `cfg.shared_experts > 0`, and skips it in the forward pass on the same condition. A configuration that sets the field keeps its current behavior, so shipped Qwen3-Next MoE configurations are unaffected.

## Tests

Adds `tests/unit/qwen3_next_shared_expert_test.py`, three cases, CPU only. With `shared_experts: 0` the expert and its gate are both `None`. With 1, both are built. And the parameter-count difference between the two equals the shared expert plus its gate, and nothing else.

On `main` the third case fails with `0 != 392` and the other two error on the missing `use_shared_expert`.

`tests/unit/qwen3_5_layers_test.py` only covers `TestQwen3_5MoeVisionEncoderEndToEnd`, so it never reaches this code and still passes.

cc @mmcsa

# Checklist

### DIFF
--- a/src/maxtext/models/qwen3.py
+++ b/src/maxtext/models/qwen3.py
@@ -1138,32 +1138,39 @@ class Qwen3NextSparseMoeBlock(nnx.Module):
         rngs=rngs,
     )
 
-    # 2. Instantiate and apply the shared expert.
-    self.shared_expert = MlpBlock(
-        config=cfg,
-        mesh=mesh,
-        in_features=cfg.emb_dim,
-        intermediate_dim=cfg.moe_mlp_dim,
-        activations=cfg.mlp_activations,
-        intermediate_dropout_rate=cfg.dropout_rate,
-        dtype=cfg.dtype,
-        weight_dtype=cfg.weight_dtype,
-        quant=self.quant,
-        model_mode=config.model_call_mode,
-        rngs=rngs,
-    )
+    # 2. Instantiate and apply the shared expert, if the config asks for one.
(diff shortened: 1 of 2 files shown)


WandLZhang:
Both attributes are now set to `None` in an `else` branch.

SurbhiJainUSC (approved):

igorts-git (approved):

igorts-git:
Please squash the commits. Unfortunately, our system cannot handle multiple commits per PR.

WandLZhang:
Squashed, it's one commit now. Ready when you are.

WandLZhang (commented):
One commit, two approvals, checks green.

PR #4451, opened 2026-07-13
Support quantization for tokamax gmm v2

shuningjin:
# Description

Integrate quantization of Tokamax GMM V2. This is joint effort with @CaptainO5, built on top of [PR#4197](https://github.com/AI-Hypercomputer/maxtext/pull/4197).

Bug: b/534817371

What this does
1. Integrate quantization in gmm v2
- `base.yml`: `use_gmm_v2=true` should work for quantization
- `src/maxtext/kernels/megablox/ops.py`: quantization and scaling integration of gmm2 / tgmm2

2. Add unit tests for quantized gmm
- `tests.integration.tokamax_test`: smoke train
- `tests.unit.moe_test`: value and gradient equivalence
  - gmm_type={megablox, tokamax gmm1, tokamax gmm2} x quant_scheme={bf16, dynamic, fp8_static_wa}

3. Significant refactor of `ops.py`: modularize GMM `fwd` and `bwd` functions 
- Previously, `_gmm_fwd` and `_gmm_bwd` were large, monolithic functions containing complex branching and quantization logic. This is difficult to read and maintain.
- Breaks them down into smaller, domain-specific sub-functions, significantly improving code readability and maintainability. Execution logic and runtime behavior remain identical, except for the aforementioned gmm v2 change.

4. Other changes
- Improve robustness of quantization. For example, `gmm` checks agaisnt empty rule to ensure quantization is applied; refactor condition `use_qwix_quantization` to be co-occur with  `quantization`; fix broken link for quantization doc.
- Add `save_checkpoint_on_start` to allow skipping for testing
- update doc for moe configuration: add `use_gmm_v2`

# Tests

### Unit test

local v5p-8

**Unit test: smoke train**
```
PYTHONPATH=.  python3 -m pytest -v --pyargs tests.integration.tokamax_test -rP -s
```

[log](https://paste.googleplex.com/6138791168442368)

**Unit test: value and gradient equivalence**
```
PYTHONPATH=. python3 -m pytest -v --pyargs tests.unit.moe_test -rP -s -k "test_gmm_grad_equivalence"
```
[log](https://paste.googleplex.com/5938345782935552)

### E2E test

We tested in b/534817371, convergence on par with bf16 for fsdp+ep or fsdp.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
[...]

PR #4957, opened 2026-08-21
Fix fp8 MoE on the sparse_matmul path

gulsumgudukbay:
# Problem

Any MoE model with `sparse_matmul=True` and an fp8 quantization crashes while the layer is being traced:

```
AttributeError: 'Fp8Quantization' object has no attribute 'quant_dg'
AttributeError: 'NANOOFp8Quantization' object has no attribute 'quant_dg'
```

`get_quantization_dtypes` reads `self.quant.quant_dg` whenever a quantization is configured, but `quant_dg` is an AQT
notion — the fp8 classes have never had one. Since `sparse_matmul=True` is the default, this is what a user hits first
when combining fp8 with a MoE model.

# Fix

`quant_dg` is how AQT spells its numerics, not what the gmm needs. All the hand-crafted rule wants is a forward dtype,
which the fp8 schemes can simply name, so they do: `float8_e4m3fn` for `fp8` and `float8_e4m3fnuz` for `nanoo_fp8`. The
read becomes "AQT describes itself through a `quant_dg`, everyone else names a dtype", and the gmm ends up quantized for
every scheme rather than only for AQT.

This keeps the three intended behaviours intact, per @shuningjin in the thread below:

1. `use_qwix_quantization=false`: every scheme quantizes the gmm through the hand-crafted rule. That now includes the
   fp8 schemes, which previously crashed here.
2. `use_qwix_quantization=true` with `fp8_full`: the gmm takes the qwix rule, untouched by this change.
3. `use_qwix_quantization=true` with any other scheme: `configure_quantization` returns `None`, so there is no dtype to
   hand over and the gmm is left unquantized, which is what a scheme declaring no gmm rule should get.

An earlier revision of this PR read `quant_dg` defensively, which stopped the crash but left fp8 with an unquantized
gmm and a warning saying so. That was the wrong side of case 1, so both are gone.

# Tests

`SparseMatmulQuantizationTest` in `moe_test.py` asserts the `lhs_quantize_dtype` and `rhs_quantize_dtype` the grouped
matmul is actually handed, across all six cases above: `fp8`, `nanoo_fp8`, `int8`, unquantized, and `fp8`/`int8` under
qwix, the last two of which must stay `None`. It intercepts `mblx.gmm`, so it costs no kernel time. The two fp8 cases
fail without this change.

Three integration tests on a tiny Mixtral with `sparse_matmul=True` cover `fp8`, `nanoo_fp8` and `int8` end to end, and
they run megablox rather than `ragged_dot` so the rule is built and used for real — which is also what confirms qwix
accepts the fnuz dtype AMD needs. A downscaled model needs tiles small enough to divide its operands, hence the tile
overrides. The two fp8 ones fail before this change, and the three take about 33s on CPU. No hardware marker: the fp8
schemes are emulated in XLA and the failure is backend independent.

# Related

- #4955 fixes the same combination on the `dense_matmul` path, where every quantization fails for an unrelated reason
  (Linen einsums with no scope to bind to after the NNX migration). The two are independent, but together they make
  quantized MoE work on both paths. There is a small textual overlap in `tests/integration/train_tests.py` and
  `tests/unit/moe_test.py`; happy to rebase whichever lands second.

# Checklist

### DIFF
--- a/tests/integration/train_tests.py
+++ b/tests/integration/train_tests.py
@@ -76,15 +76,35 @@ class TrainTests(unittest.TestCase):
       "sharding_tolerance=0.1",
   ]
 
-  # Routes the MoE layer through dense_matmul, which is what runs wherever the megablox and
-  # ragged kernels are unavailable.
-  _moe_model_overrides = [
+  _moe_expert_overrides = [
       "decoder_block=mixtral",
       "num_experts=4",
       "num_experts_per_tok=2",
       "base_moe_mlp_dim=32",
+  ]
+
+  # Routes the MoE layer through dense_matmul, which is what runs wherever the megablox and
+  # ragged kernels are unavailable.
+  _moe_model_overrides = _moe_expert_overrides + [
       "sparse_matmul=False",
       "megablox=False",
+  ]
+
+  # The sparse_matmul path, where megablox builds and uses the gmm quantization rule for real
+  # rather than falling back to ragged_dot.
+  _moe_sparse_model_overrides = _moe_expert_overrides + [
+      "sparse_matmul=True",
+      "megablox=True",
+  ]
+
+  # Every operand has to divide into its tile, and the default tiles are far larger than a
+  # downscaled model. The embedding dim is 28 or 32 depending on the device count, so 4 is
+  # the largest tile that fits it either way.
+  _megablox_tile_overrides = [
+      f"{matrix}_tile_{direction}_{dim}={size}"
+      for matrix in ("wi", "wo")
+      for direction in ("fwd", "dlhs", "drhs")
+      for dim, size in (("batch_seq", 16), ("embed_dim", 4), ("mlp_dim", 16))
   ]
 
   _qwen3_overrides = [
@@ -195,6 +215,20 @@ class TrainTests(unittest.TestCase):
       ]
       + _small_model_overrides
       + _moe_model_overrides,
+      "moe_sparse": [  # tests a MoE model on the sparse_matmul path, to be combined with a quantization
+          None,
+          get_test_config_path(),
+          f"base_output_directory={_base_output_directory}",
+          "run_name=runner_test",
+          "dataset_type=synthetic",  # use synthetic dataset_type to decrease training time
+          "steps=2",
+          "enable_checkpointing=False",
+          "enable_goodput_recording=False",
+          rf"tokenizer_path={os.path.join(MAXTEXT_ASSETS_ROOT, 'tokenizers', 'tokenizer.llama2')}",
+      ]
+      + _small_model_overrides
+      + _moe_sparse_model_overrides
+      + _megablox_tile_overrides,
       "te_fp8_delayedscaling": [  # tests base config with te_fp8_delayedscaling
           None,
           get_test_config_path(),
@@ -367,6 +401,22 @@ class TrainTests(unittest.TestCase):
   def test_moe_fp8_token_dropping(self):
     # capacity_factor > 0 adds the dispatch and combine einsums to the ones above.
     train_main(TrainTests.CONFIGS["moe"] + ["quantization=fp8", "capacity_factor=1.25"])
+
+  # The sparse_matmul tests below carry no hardware marker for the same reasons. What they cover
+  # is an attribute read during tracing rather than anything a kernel does, and megablox runs the
+  # quantized grouped matmul on CPU through its interpret mode.
+  @pytest.mark.integration_test
+  def test_moe_fp8_sparse_matmul(self):
+    train_main(TrainTests.CONFIGS["moe_sparse"] + ["quantization=fp8"])
+
+  @pytest.mark.integration_test
+  def test_moe_nanoo_fp8_sparse_matmul(self):
+    train_main(TrainTests.CONFIGS["moe_sparse"] + ["quantization=nanoo_fp8"])
+
+  # int8 takes the `quant_dg` branch of the same read, which the fp8 tests never reach.
+  @pytest.mark.integration_test
+  def test_moe_int8_sparse_matmul(self):
+    train_main(TrainTests.CONFIGS["moe_sparse"] + ["quantization=int8"])
 
   @pytest.mark.skip(reason="No runner with GPU arch >= 89 is available")
   @pytest.mark.integration_test
--- a/src/maxtext/layers/moe.py
+++ b/src/maxtext/layers/moe.py
@@ -1499,12 +1499,14 @@ class RoutedMoE(nnx.Module):
         )
 
     def get_quantization_dtypes():
-      lhs_quantize_dtype, rhs_quantize_dtype = None, None
-      if self.quant is not None:
-        quant_dg = self.quant.quant_dg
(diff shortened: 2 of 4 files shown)


Shuwen-Fang (approved):

gulsumgudukbay:
Codecov flagged one partial on the patch. The two fp8 tests only ever take the branch where there is no `quant_dg`, so the read that does find one, and the isinstance check that skips the warning, were never exercised from this PR's own tests. Added `test_moe_int8_sparse_matmul` in a46a1ad: int8 is the scheme whose dtypes actually reach the gmm, so it covers both arcs and guards the behaviour this change had to preserve. Measured locally with `coverage --branch`, the changed lines now have no missing lines or branches; the three tests run in about 28s on CPU.

shuningjin:
Hi @gulsumgudukbay, Here is the intended behavior:
1. when `use_qwix_quantization=false`, all quantization scheme use [quantized gmm with hand-crafted rule](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L94-L101)
2. when `use_qwix_quantization=true`, for `fp8_full`, it will use the [quantized gmm with qwix rule](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L88-L92)
3. when `use_qwix_quantization=true`, for non-`fp8_full` (e.g., fp8, int8), the gmm path should be an [no-op](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L88-L92) as it does not specify gmm qwix rule [here](https://github.com/AI-Hypercomputer/maxtext/blob/c8934848dd73f1a70972bac2da4c829479e515b8/src/maxtext/layers/quantizations.py#L806-L843).

I accidentally changed Behavior 3: https://github.com/AI-Hypercomputer/maxtext/pull/4451 (assert against empty gmm qwix rule) and https://github.com/AI-Hypercomputer/maxtext/pull/4842 (allow non-`fp8-full` with qwix to use quantized gmm with hand-crafted rule). We should restore to previous state (non-`fp8-full` with qwix does NOT use quantized gmm). Sorry for the confusion!

gulsumgudukbay:
> Hi @gulsumgudukbay, Here is the intended behavior:
> 
> 1. when `use_qwix_quantization=false`, all quantization scheme use [quantized gmm with hand-crafted rule](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L94-L101)
> 2. when `use_qwix_quantization=true`, for `fp8_full`, it will use the [quantized gmm with qwix rule](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L88-L92)
> 3. when `use_qwix_quantization=true`, for non-`fp8_full` (e.g., fp8, int8), the gmm path should be an [no-op](https://github.com/AI-Hypercomputer/maxtext/blob/dfd8d293d266fe224b90f7cb0b49f3e8084e9892/src/maxtext/kernels/megablox/ops.py#L88-L92) as it does not specify gmm qwix rule [here](https://github.com/AI-Hypercomputer/maxtext/blob/c8934848dd73f1a70972bac2da4c829479e515b8/src/maxtext/layers/quantizations.py#L806-L843).
> 
> I accidentally changed Behavior 3: #4451 (assert against empty gmm qwix rule) and #4842 (allow non-`fp8-full` with qwix to use quantized gmm with hand-crafted rule). We should restore to previous state (non-`fp8-full` with qwix does NOT use quantized gmm). Sorry for the confusion!

Thanks @shuningjin, that's very helpful. I had this on the wrong side of behavior 1. just pushed 5bcaf7f, which changes the fix.

`quant_dg` is how AQT spells its numerics, not what the gmm needs. The hand-crafted rule only wants a forward dtype, which the fp8 schemes can name directly, so they now do: `float8_e4m3fn` for `fp8`, `float8_e4m3fnuz` for `nanoo_fp8`. now it becomes becomes "AQT describes itself through a `quant_dg`, everyone else names a dtype", so with `use_qwix_quantization=false` the gmm is quantized for every scheme, fp8 included, where it used to crash. 

`fp8_full` under qwix is untouched, and any other scheme under qwix has no quantization object at all, so there is no dtype to pass and the gmm stays unquantized. My previous revision stopped the crash but left fp8 with an unquantized gmm and a warning saying so; both of these are gone now.

On restoring behavior 3, it seems to me like the gmm dtypes already satisfy it, since `configure_quantization` returns `None` whenever `use_qwix_quantization=true`. If what you have in mind for #4451 and #4842 goes further, I'll leave it to you

`SparseMatmulQuantizationTest` in `moe_test.py` intercepts `mblx.gmm` and asserts the dtypes it is handed for all six cases, including the two qwix ones that must stay `None`; the fp8 cases fail without the change. The integration tests now run megablox instead of `ragged_dot` so the rule is really built, which is also what confirms qwix accepts the fnuz dtype AMD needs.

tests/integration/train_tests.py
```diff
@@ -288,6 +317,21 @@ def test_gpu_fp8(self):
   def test_gpu_nanoo_fp8(self):
     train_main(TrainTests.CONFIGS["nanoo_fp8"] + ["attention=dot_product"])
 
+  # No hardware marker: the fp8 schemes do not reach the expert matmuls on this path on any
```
gobbleturk: I'm surprised this is true - isn't train_main being invoked which will run on the backend?

gulsumgudukbay: sorry you are right, that comment was stale it predates 5bcaf7f, where fp8 stopped falling past the gmm quantization and started feeding it a dtype like every other scheme. i fixed this in 664b181, along with the int8 one next to it.

regarding the marker: conftest auto-marks unmarked tests `cpu_only` and skips those on accelerator testbeds, so these only run in the CPU job, which is fine/enough because the failure is an attribute read during tracing, and fp8 is emulated in XLA, so the quantized gmm still runs under megablox's interpret mode.

tests/integration/train_tests.py
```diff
@@ -288,6 +317,22 @@ def test_gpu_fp8(self):
   def test_gpu_nanoo_fp8(self):
     train_main(TrainTests.CONFIGS["nanoo_fp8"] + ["attention=dot_product"])
 
+  # No hardware marker, so conftest runs these cpu_only. The failure they cover is an attribute
```
gobbleturk: typo conftest?

gobbleturk (approved):

tests/integration/train_tests.py
gulsumgudukbay: no, it is not a typo, conftest.py is the pytest file that auto-marks unmarked tests cpu_only. i just reworded now for it to say so properly, since "conftest runs these" was misleading anyway. it marks them, and the CPU job is what runs them sorry for the confusion.

PR #4964, opened 2026-08-21
perf(qwen3-next): nested scan over hybrid attention with per-layer remat

NuojCheng:
# Description

Qwen3-Next has a hybrid decoder stack: each period of `inhomogeneous_layer_cycle_interval` (4) layers is three linear-attention (GatedDeltaNet) layers followed by one full-attention layer. `Qwen3NextScannableBlock` instantiated those four heterogeneous sub-layers flat (`layer_0` … `layer_3`) and ran them in a Python loop, so the whole block was rematerialized as a single unit and all four sub-layers' activations were live at once.

This PR restructures the block along the same lines as `Gemma4ScannableBlock`:

- the three linear-attention layers are homogeneous, so they are stacked and run through `nnx_scan.apply_scanned_layers`;
- the single full-attention layer runs inside a trip-count-one `jax.lax.scan`, which acts as an XLA scheduling barrier (`skip-simplify-while-loops_trip-count-one`);
- each sub-layer is rematerialized on its own (`apply_internal_remat=True`), so the outer apply passes `skip_block_remat=True` and nothing is rematerialized twice.

A second, independent fix in `nnx_scan.apply_scanned_layers`: the scan body returned `nnx.state(current_layer)` — the *full* state, parameters included. `jax.lax.scan` stacks every per-iteration output, so returning parameters made XLA materialize a second copy of the stacked layer weights on every call. The body now returns only the state that was not fed in as a scan input.

Dropping every `nnx.Param` would be too blunt. Qwix materializes LoRA adapters while tracing the scan body, and `nnx.LoRAParam` is an `nnx.Param` subclass, so they would go out with the carried weights and LoRA setup would fail with `LoRA module path matched target modules, but nnx.LoRAParam is still missing`. Diffing against the parameter paths that went in keeps the memory saving and still lets anything the body creates out.

External (vLLM) KV caches arrive as a flat list with one entry per decoder layer, but the NNX scan runs over *blocks*, so `_apply_qwen3_next_scanned_blocks` regroups them into per-block tuples and writes them back afterwards — the same `prepare_kv_caches_for_scan` / `update_kv_caches_after_scan` pair Gemma4 uses. That keeps `skip_block_remat=True` on the KV-cache path instead of falling back to the generic, block-rematerialized branch. `Qwen3NextScannableBlock` now also threads one cache per sub-layer and returns the updated ones; previously the block handed the same cache object to all four sub-layers and dropped the updates. The Linen `Decoder` already did the equivalent regrouping and is unchanged in this respect.

Layer counts that are not a multiple of the cycle are handled on both decoders. The pure-NNX decoder built `num_decoder_layers // cycle` blocks and stopped there, silently dropping the trailing layers; the Linen `Decoder` did keep them, but spelled them out as individual `Qwen3NextDecoderLayerToLinen` layers, which named their parameters differently from anything the NNX side produces. Both now put the leftovers into one short `Qwen3NextScannableBlock` named `layers_remainder`, so no layer is dropped and a single checkpoint mapping still covers both trees. Gemma4 and Qwen3-Next share the code that applies it (`_apply_remainder_block`) rather than carrying two copies of the same split / `jax.checkpoint` / merge dance.

## Checkpoint conversion

The new layout needs matching conversion support, so that HF Qwen3-Next checkpoints still load into MaxText and still save back out. This follows #4530, which did the same for Gemma4.

The scanned branch of `QWEN3_NEXT_MAXTEXT_TO_HF_PARAM_MAPPING`, and its hook map, now describe the nested layout instead of the old `layer_{0..3}` keys. Writing `psa` for `param_scan_axis`, a parameter lands in one of four shapes:

| MaxText key | HF keys nest as | stacked axes |
| --- | --- | --- |
| `global_layer-*` | `[block]` | `psa` |
| `local_layers-*` | `[block][local]` | `psa, psa + 1` |
| `global_layer-…-routed_experts-w*` | `[expert][block]` | `0, 1` |
| `local_layers-…-routed_experts-w*` | `[expert][block][local]` | `0, psa, psa + 1` |

Only the last row needed new machinery. Routed experts are expert-stacked *inside* the nested local scan, so they take a third stacked axis, and the conversion helpers only handled two. `_build_multi_axis_stacked_tensor` (to_maxtext) and its inverse in `process_maxtext_param` (to_huggingface) are generalized from two axes to N, and the axis placement moves into a shared `stacked_axes` helper in `tensor_handling.py`, next to `nesting_depth` and `slice_shape`. A third copy of the same helper lived in `utils/utils.py`, unreachable — nothing calls that module's `_get_hf_loading_function` — so it is deleted here rather than updated. Nested-scan detection widens from `scanned_blocks-local_layers`, which is Gemma4's module name, to `-local_layers`, so Qwen3-Next's `layers-local_layers` matches as well; `mt_key` is threaded through the `load_dynamic` lazy path so it makes the same choice.

The Linen `Decoder` needed a fix first. `_apply_qwen3_next_scanned_blocks` still had a broadcast-arg spec copied from Gemma4, which no longer matched `Qwen3NextScannableBlock.__call__` and passed 12 positional args to a 10-argument signature, and it named the scanned module `scanned_blocks` where the pure-NNX decoder uses `layers`. With both corrected, and with the remainder unified as described above, the two decoders emit byte-identical parameter names and shapes, so a single mapping serves both.

## Results

Activation memory drops by roughly 45–60%. Every row below is the same command on both sides, changing only the checkout.

**AOT compile, real `qwen3-next-80b-a3b`, `compile_topology=v5p-256`, `per_device_batch_size=1`, `max_target_length=2048`, `remat_policy=full`, `attention=flash`, `dtype=bfloat16`:**

| `CompiledMemoryStats` | main | this PR | change |
| --- | ---: | ---: | ---: |
| `temp_size_in_bytes` | 30.25 GB | 16.10 GB | **−46.8%** |
| `argument_size_in_bytes` | 7.49 GB | 7.49 GB | — |
| `generated_code_size_in_bytes` | 312 MB | 153 MB | −50.9% |
| total (args + temp) | 37.74 GB | 23.59 GB | −37.5% |

**Real training runs on a v5p-8** (4 chips, `ici_fsdp_parallelism=4`), scaled-down qwen3-next, `max_target_length=4096`:

| | main total / temp | this PR total / temp | temp change |
| --- | ---: | ---: | ---: |
| `per_device_batch_size=2` | 26.7 / 22.1 GB | 17.0 / 12.3 GB | **−44.3%** |
| `per_device_batch_size=12` | 116.9 / 112.2 GB | 48.2 / 43.6 GB | **−61.1%** |

`argument_size` is byte-identical on both sides in all three experiments, confirming the two trees are the same architecture with the same parameter count — the saving is purely activations, which is why the gap widens with batch size.

Step time is unchanged: 1.544 s (main) vs 1.538 s (this PR) at batch 2, and 8.181 s vs 8.130 s at batch 12.

## Shortcomings and follow-ups

- **Parameter layout changes.** A block's weights move from `layer_0…layer_3` to a stacked `local_layers` plus a `global_layer`. HF checkpoints round-trip through the updated conversion, but a MaxText checkpoint saved from the old tree has to be re-converted.
- **The full-attention layer must be last in the cycle**, because the local scan runs before the global layer. That is where stock Qwen3-Next puts it, and the block's constructor raises rather than silently reordering the model if a block straddles a cycle boundary or covers more than one full-attention layer. Supporting a hybrid stack that interleaves differently means splitting the stacked local scan in two; no shipped config needs it, so the position stays derived from `inhomogeneous_layer_cycle_interval` rather than exposed as a config flag.
- **No dense-prefix support.** `first_num_dense_layers` is 0 for stock Qwen3-Next, so a scanned dense prefix is left as a follow-up.
- Gemma4 and Qwen3-Next still have near-identical block implementations. The remainder path is now shared (`_apply_remainder_block`), but factoring out a shared hybrid-block base, and unifying `nnx_scan.apply_scanned_layers` with `NNXDecoder._apply_layers_sequentially` (two appliers with two different state write-back conventions), is worth doing separately.

# Tests

## Block and decoder structure

Ten new tests in `tests/unit/nnx_decoders_test.py`, next to the analogous `TestGemma4ScannableBlock`.

`TestQwen3NextScannableBlock` asserts that the block splits a cycle into a stacked local scan plus one global layer, that the local params really are stacked along `param_scan_axis`, that **the nested scans reproduce a plain sequential unroll of the same weights** to `rtol=atol=1e-5`, that the external-KV path matches the scanned path and returns one cache per sub-layer in `local…global` order, and that a block starting off a cycle boundary -- which would straddle two periods and put the full-attention layer third of four -- is rejected rather than silently reordered.

`TestNNXDecoderQwen3Next` feeds the decoder a flat per-layer cache list with a distinct sentinel per layer and checks each one comes back on its own layer — this fails on the pre-fix code, where block *i* received `kv_caches[i]` instead of its four-cache group. It also covers the remainder: with 6 layers and a cycle of 4, `layers_remainder` has to hold the last two, and perturbing only that block's weights has to move the output, which it cannot do if the block is never applied. On the pre-fix code those two layers do not exist at all. With 8 layers there is no remainder block.

`TestQwen3NextDecoderParity` builds the parameter tree from `get_maxtext_model_info` on each decoder and asserts the two are equal, with and without a remainder. One mapping serves both decoders only for as long as that holds, and nothing was checking it.

## Checkpoint conversion

`tests/unit/param_mapping_test.py` gains three round-trip tests, modelled on #4530's `test_gemma4_local_layers_stack_unstack_roundtrip`: stacking HF weights into the MaxText layout and un-stacking them back has to be the identity. They cover the two-axis `[block][local]` case, the three-axis `[expert][block][local]` case, and the global layer's plain leading-axis MoE case. `test_qwen3_next_mapping_scanned` is rewritten for the new layout.

The mapping is also checked against the model itself. For a scanned `qwen3-next-80b-a3b`, its key set matches `get_maxtext_model_info`'s parameter tree exactly, with nothing missing and nothing extra, and every key's nesting depth matches the real tensor shape at the axes `stacked_axes` picks — on both decoders, which now produce identical trees.

## Decode

Golden-logit tests only cover the forward pass, so nothing here caught a broken prefill → autoregressive handoff, which is exactly the kind of bug a restructured block can introduce: the forward pass can be bit-correct while the block mis-threads its sub-layers' carried state between decode steps.

`tests/unit/deepseek_decode_consistency_test.py` is the existing harness for that, added by the DeepSeek "decode ran the prefill attention path under NNX" fix; its module docstring notes the property holds for any model, so new families go in the parameter list. It decodes a tiny model two ways and requires the token ids to be equal:

- **through MaxEngine** — `prefill` on the padded prompt, `init_decode_state`, `insert` at slot 0, then `generate` per step, greedy sampling;
- **teacher-forced** — a plain `MODEL_MODE_TRAIN` forward pass recomputed over the whole growing prefix each step, taking `argmax` at the last real position.

The second is the definition of what the first is supposed to be replaying, so any divergence is a decode bug rather than a tolerance question. Both sides load the same freshly initialized `nnx.Param` state, with `matmul_precision=highest`, `dtype=float32` and `attention=dot_product` so the comparison is exact rather than approximate. Prompt is 5 tokens, 5 tokens are generated.

Three Qwen3-Next cases are added, all on a scaled-down `qwen3-next-80b-a3b` (`base_emb_dim=64`, `head_dim=32`, 4 experts, `gdn_chunk_size=4`, vocab 64):

| case | layers | what it covers |
| --- | ---: | --- |
| `qwen3_next_unscanned` | 8 | the per-layer path, as a control — isolates whether a failure is the block or the layer |
| `qwen3_next_scanned` | 8 | two whole periods through the nested scan, no remainder |
| `qwen3_next_scanned_remainder` | 6 | one period plus a two-layer `layers_remainder` block |

Both attention kinds in the hybrid stack carry state across a decode step — the GatedDeltaNet recurrent and conv states, and the full-attention KV cache — and the scanned path threads them through two nested `lax.scan`s plus, in the remainder case, a separate short block outside the scan. A block that regroups caches wrongly, hands one sub-layer another's state, or drops the write-back shows up here as a diverging token. The remainder case additionally fails outright on the pre-fix code, since layers 4 and 5 were not built.

## Commands

```
pytest tests/unit/nnx_decoders_test.py tests/unit/nnx_scan_test.py \
       tests/unit/nnx_decoder_test.py tests/unit/maxtext_utils_nnx_test.py \
       tests/unit/param_mapping_test.py tests/unit/checkpointing_test.py
# 113 passed, 2 skipped

pytest tests/unit/deepseek_decode_consistency_test.py
# 7 passed

pytest tests/post_training/integration/lora_e2e_nnx_test.py
# 3 passed
```

End-to-end on a v5p-8, 8 steps of synthetic data with fixed seeds. Losses track main to within 0.002 at every step despite the two trees drawing init RNG differently:

| step | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| main | 9.505 | 9.016 | 8.585 | 8.228 | 7.956 | 7.771 | 7.660 | 7.600 |
| this PR | 9.499 | 9.013 | 8.583 | 8.228 | 7.957 | 7.771 | 7.660 | 7.601 |

Reproduce (drop `override_model_config` and the size overrides to run the real 80B):

```bash
  model_name=qwen3-next-80b-a3b override_model_config=True \
  base_output_directory=/tmp/run_out run_name=scan_check dataset_type=synthetic \
  enable_checkpointing=False steps=8 data_shuffle_seed=1234 init_weights_seed=1234 \
  base_emb_dim=2048 base_num_decoder_layers=12 base_num_query_heads=16 \
  base_num_kv_heads=2 head_dim=256 base_mlp_dim=4096 base_moe_mlp_dim=512 \
  num_experts=32 num_experts_per_tok=8 vocab_size=8192 \
  per_device_batch_size=2 max_target_length=4096 dtype=bfloat16 attention=flash \
  remat_policy=full sparse_matmul=True megablox=False ici_fsdp_parallelism=4
```

AOT numbers come from `Memory analysis: {compiled.memory_analysis()}` in `train_compile.py`:

```bash
python3 -m maxtext.trainers.pre_train.train_compile src/maxtext/configs/base.yml \
  model_name=qwen3-next-80b-a3b compile_topology=v5p-256 compile_topology_num_slices=1 \
  base_output_directory=/tmp/aot_out run_name=aot dataset_type=synthetic \
  enable_checkpointing=False skip_jax_distributed_system=True steps=3 \
  per_device_batch_size=1 max_target_length=2048 dtype=bfloat16 attention=flash \
  remat_policy=full sparse_matmul=True megablox=False
```

# Checklist


src/maxtext/layers/nnx_decoders.py
```diff
@@ -1858,6 +1882,8 @@ def __call__(
               layer_kwargs,
               kv_caches=kv_caches,
           )
+        elif self.is_qwen3_next and kv_caches is None:
+          y = self._apply_qwen3_next_scanned_blocks(y, layer_args, layer_kwargs)
```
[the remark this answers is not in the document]
NuojCheng: Good catch on the double-remat — fixed in 0857e11.

I didn't take the suggestion verbatim, though: the scan runs over *blocks*, not layers, while `kv_caches` is a flat list with one entry per decoder layer. Passing it straight through as `kv_caches_stacked` would hand block `i` only `kv_caches[i]` (a single layer's cache) instead of its four-cache group, and the block would then index into that single object.

So `_apply_qwen3_next_scanned_blocks` now does what `_apply_gemma4_scanned_blocks` does:

```python
grouped_kv_caches = maxtext_utils.prepare_kv_caches_for_scan(kv_caches, scan_length, block_length, stack=False)
y, self.layers, _ = self._apply_layers_sequentially(
    self.layers, y, *layer_args,
    length=scan_length, kv_caches_stacked=grouped_kv_caches, skip_block_remat=True, **layer_kwargs,
)
maxtext_utils.update_kv_caches_after_scan(kv_caches, grouped_kv_caches, scan_length, block_length, stacked=False)
```

and the dispatch is now `elif self.is_qwen3_next:` with `kv_caches=kv_caches`, as suggested.

Two tests cover this. `TestNNXDecoderQwen3Next.test_decoder_regroups_flat_kv_caches_per_block` passes a distinct sentinel per layer and checks each comes back on its own layer — it fails on the pre-fix code. `TestQwen3NextScannableBlock.test_external_kv_cache_matches_scanned_path` checks the block's external-KV path matches the scanned path numerically and returns one cache per sub-layer in `local…global` order.

Training is unaffected: with `kv_caches=None` both helpers pass `None` through, so that path is byte-identical to before and the memory numbers in the description still stand.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -2140,6 +2166,27 @@ def pure_gemma_fn(graphdef, state_in, y_in, kv_in):
 
     return y
 
+  def _apply_qwen3_next_scanned_blocks(self, y, layer_args, layer_kwargs):
```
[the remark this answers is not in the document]
NuojCheng: Addressed in 0857e11 — see the [reply on the dispatch thread](https://github.com/AI-Hypercomputer/maxtext/pull/4964#discussion_r3834083172). Short version: `kv_caches` is flat per-layer while the scan is per-block, so it is regrouped with `prepare_kv_caches_for_scan` / `update_kv_caches_after_scan` (as Gemma4 does) rather than passed through directly, and `skip_block_remat=True` now holds on this path too.

src/maxtext/layers/nnx_decoders.py
```diff
@@ -2140,6 +2166,27 @@ def pure_gemma_fn(graphdef, state_in, y_in, kv_in):
...
+    if scan_length == 0:
+      return y
+    y, self.layers, _ = self._apply_layers_sequentially(
+        self.layers,
+        y,
+        *layer_args,
+        length=scan_length,
+        skip_block_remat=True,
+        **layer_kwargs,
+    )
```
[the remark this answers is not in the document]
NuojCheng: Addressed in 0857e11 — see the [reply on the dispatch thread](https://github.com/AI-Hypercomputer/maxtext/pull/4964#discussion_r3834083172). Short version: `kv_caches` is flat per-layer while the scan is per-block, so it is regrouped with `prepare_kv_caches_for_scan` / `update_kv_caches_after_scan` (as Gemma4 does) rather than passed through directly, and `skip_block_remat=True` now holds on this path too.

NuojCheng:
Pushed 0857e11 addressing the review — the KV-cache path now keeps `skip_block_remat=True` and regroups the flat per-layer cache list into per-block tuples (details in the thread replies above). Tests moved into `tests/unit/nnx_decoders_test.py` next to `TestGemma4ScannableBlock`, with new coverage for both halves of the KV path.

/gemini review

src/maxtext/models/qwen3.py
```diff
@@ -43,6 +43,8 @@
 from maxtext.layers import moe
 from maxtext.layers import nnx_wrappers
 from maxtext.layers import quantizations
+from maxtext.layers import nnx_scan
+from jax.experimental import xla_metadata
```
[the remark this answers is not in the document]
NuojCheng: Done in 99cbb62 — moved up next to `from jax.ad_checkpoint import checkpoint_name` so it sits in the jax group rather than at the top of the file, keeping that group alphabetical.

src/maxtext/models/qwen3.py
```diff
@@ -1208,35 +1217,190 @@ class Qwen3NextScannableBlock(nnx.Module):
...
+        y, new_kv = self._run_layer(layer, y, layer_kwargs, current_kv)
+        updated_kvs.append(new_kv)
+        per_layer_states.append(nnx.state(layer))
+
+      stacked_state = jax.tree.map(lambda *xs: jnp.stack(xs), *per_layer_states)
+      if scan_axis != 0:
+        stacked_params, stacked_other = stacked_state.split(nnx.Param, ...)
+        stacked_params = jax.tree.map(lambda x: jnp.moveaxis(x, 0, scan_axis), stacked_params)
+        stacked_state = nnx.State.merge(stacked_params, stacked_other)
+      nnx.update(self.local_layers, stacked_state)
```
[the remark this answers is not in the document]
NuojCheng: Agreed, and applied in 99cbb62.

Worth spelling out why the `param_scan_axis` round-trip disappears along with the stacking, since it looks like a separate change: `nnx_scan.create_scanned_layers` only puts *Params* on `param_scan_axis` (`add_scan_metadata(stacked_params, param_scan_axis)`); non-Param state gets axis 0 (`add_scan_metadata(stacked_rest, 0)`). The `state` read at the top of this method is therefore already axis-0 and sliced as such, so once Params are out of `per_layer_states` the stack goes straight back on axis 0 — matching what `apply_scanned_layers` does with `scanned_rest`. Net effect here:

```python
        _, _, updated_state = nnx.split(layer, nnx.Param, ...)
        per_layer_states.append(updated_state)

      nnx.update(self.local_layers, jax.tree.map(lambda *xs: jnp.stack(xs), *per_layer_states))
```

`Qwen3NextScannableBlock` tests still pass (6/6), and the four NNX suites are at 71 passed / 2 skipped.

One note for anyone following up: `Gemma4ScannableBlock._forward_with_external_kv_cache` (`gemma4.py:633`) still has the original `nnx.state(layer)` + moveaxis version, so it keeps the extra param copy. I left it alone to keep this PR scoped to Qwen3-Next, but the same fix applies verbatim.

darisoy (approved):

src/maxtext/layers/decoders.py
```diff
@@ -1423,6 +1435,128 @@ def _apply_gemma3_scanned_blocks(
...
+          kv_caches, returned_kv_cache, num_full_blocks, block_pattern_len, stacked=True
+      )
+
+    # Layers past the last whole block go into a short block of their own. It is a
+    # Qwen3NextScannableBlock like the scanned ones, named `layers_remainder`, so the
+    # parameter paths still match the pure-NNX decoder's and one mapping serves both.
+    if remainder_layers > 0:
+      start_idx = cfg.num_decoder_layers - remainder_layers
+      remainder_kv = tuple(kv_caches[start_idx:]) if kv_caches is not None else None
+      y_and_kv = qwen3.Qwen3NextScannableBlockToLinen(
```
gobbleturk, line 1527: these are always GDN layers? If so please add a comment

src/maxtext/layers/decoders.py
```diff
@@ -1423,6 +1435,128 @@ def _apply_gemma3_scanned_blocks(
 
     return y
 
+  def _apply_qwen3_next_scanned_blocks(
```
gobbleturk, line 1440: if possible it would be great to make this function generic to this nested scan approach - I can see the "remainder" functionality might be different for different layers, but perhaps all implementations can share the core nested scan and then call into individual remainders. Doesn't have to be in this PR, but something to consider (there are pros and cons of having this shared functionality)

gobbleturk, line 1440: should we be defining this function in a qwen file - either a new qwen3 utils or the existing qwen3 file? Decoders is going to blow up implementing model specific logic otherwise

src/maxtext/layers/nnx_decoders.py
```diff
@@ -717,6 +720,48 @@ def _init_scanned_gemma4(self, decoder_block_classes, rngs, mesh):
         rngs=rngs,
     )
 
+  def _init_scanned_qwen3_next(self, rngs, mesh):
```
gobbleturk, line 766: should we be defining this function in a qwen file - either a new qwen3 utils or the existing qwen3 file? Decoders is going to blow up implementing model specific logic otherwise

tests/unit/nnx_decoders_test.py
```diff
@@ -843,6 +843,251 @@ def test_restores_local_state_and_preserves_kv_order(self):
...
+    )
+    return cfg, mesh, block
+
+  def _inputs(self, cfg):
+    inputs = jax.random.normal(jax.random.PRNGKey(1), (1, cfg.max_target_length, cfg.emb_dim), dtype=jnp.float32)
+    positions = jnp.arange(cfg.max_target_length)[None, :]
+    segment_ids = jnp.ones((1, cfg.max_target_length), dtype=jnp.int32)
+    return inputs, segment_ids, positions
+
+  def test_block_splits_cycle_into_local_stack_plus_one_global(self):
```
gobbleturk, line 902: any idea how long these tests take? This is a lot of compilation time I would imagine

gobbleturk (approved):

src/maxtext/layers/decoders.py
NuojCheng: agree, at least we should provide a call to generate this nested scan pattern for whatever hybrid models. This is a to-do (b/555915471)

NuojCheng: Right, I did this following old patterns in decoder.py. We should have a separate PR moving all these scan blocks to decoders.

NuojCheng: But I am not gonna make the changes in this PR

src/maxtext/layers/decoders.py
NuojCheng: qwen3 next model always uses GDN layer

src/maxtext/layers/nnx_decoders.py
NuojCheng: Agree, should be included as a follow-up PRs moving all such logics to model-specific files.

tests/unit/nnx_decoders_test.py
NuojCheng: Good point and many of them do take great time. 
```
76.58s call     tests/unit/nnx_decoders_test.py::TestNNXDecoderQwen3Next::test_decoder_regroups_flat_kv_caches_per_block
42.80s call     tests/unit/nnx_decoders_test.py::TestQwen3NextScannableBlock::test_external_kv_cache_matches_scanned_path
41.87s call     tests/unit/model_creation_utils_test.py::TestCreateNnxModel::test_load_checkpoint_with_custom_vision_projector
32.89s call     tests/unit/nnx_decoders_test.py::TestQwen3NextScannableBlock::test_nested_scan_matches_sequential_unroll
25.08s call     tests/unit/mhc_test.py::TestMHC::test_attention_layer_output_shape_Rate3
22.99s call     tests/unit/mhc_test.py::TestMhcKernelsBwd::test_forward_and_backward_vjp_parity_feature_tiled
21.25s call     tests/unit/multi_token_prediction_test.py::MTPLossFnZeroRegressionTest::test_mtp_loss_is_nonzero_through_real_loss_fn
```

I will removed the expensive ones that take more than 20s.

Issue #4933, opened 2026-08-18
Multi-host config validation assumes 1 device, so global_batch_size_to_train_on is derived wrong


WandLZhang:
On multi-host, config is validated before the JAX distributed system is initialized. `get_num_target_devices()` in `configs/types.py` calls `jax.devices()`, which raises at that point. The handler assumes 1 device:

```python
self.num_target_devices = 1  # Default for validation when JAX is not initialized
try:
  self.num_target_devices = get_num_target_devices()
except (RuntimeError, IndexError):
  logger.warning("JAX device system not available for config validation. Assuming 1 device.")
```

With 1 device assumed, `global_batch_size_to_train_on` becomes `per_device_batch_size` regardless of the real chip count. The run then fails inside `pjit` with an indivisibility error that names neither the batch nor the device count, so the cause isn't obvious from the message.

## Reproduce

Run any multi-host job on GKE without setting `global_batch_size_to_train_on` explicitly. A 4-host, 16-chip slice shows the warning above and then fails in `pjit`.

## Notes on a real fix

I also tried inferring the chip count from `TPU_CHIPS_PER_HOST_BOUNDS` and `TPU_WORKER_HOSTNAMES` in the `except` branch. It works on multi-host GKE, but I don't think it belongs upstream:

- It reads environment variables that multi-controller GKE happens to set. They aren't a supported interface.
- Under Pathways those variables are absent. That case is harmless, because `train.py` calls `pathwaysutils.initialize()` before `pyconfig.initialize(argv)`, so the proxy backend is registered and `jax.devices()` returns the real device list. The `except` branch never runs. I confirmed this on a v5e slice: `TPU_CHIPS_PER_HOST_BOUNDS` and `TPU_WORKER_HOSTNAMES` were both unset and `jax.devices()` still returned 8.

A better fix runs the checks that need a device count after distributed init. Or it raises an error naming the missing setting instead of quietly assuming 1. Both are bigger than something I should propose on my own.

cc @mmcsa

PR #4968, opened 2026-08-22
Shard the GatedDeltaNet sequence under ici_context_parallelism

### LINKED ISSUE #4932: GatedDeltaNet does not support ici_context_parallelism
Setting `ici_context_parallelism > 1` doesn't shard the sequence inside `Qwen3_5GatedDeltaNet`. Two separate things prevent it.

**1. Three pspecs pin the sequence axis to replicated.** In `models/qwen3.py`, `qkv_pspec`, `g_beta_pspec` and `qkvz_pspec` are built as `(KV_BATCH, None, KV_HEAD, None)`. The `None` in the sequence position tells XLA to gather the full sequence onto every device, so a `with_sharding_constraint` undoes the sharding the mesh asks for.

**2. The inter-chunk recurrence is a sequential scan.** `jax_chunk_gated_delta_rule` runs `lax.scan` over chunks. A sequential scan can't be split across devices, because each chunk needs the state the previous chunk produced.

## The recurrence is affine, so it can be parallelized

The GatedDeltaNet inter-chunk step is

```
h_new = exp(g_last) * h + k_g^T (u - w h)
      = (exp(g_last) I - k_g^T w) h + k_g^T u
      = A h + B
```

`A` and `B` don't depend on `h`. Affine maps compose associatively:

```
(A2, B2) o (A1, B1) = (A2 A1, A2 B1 + B2)
```

So each device composes its local chunks into one `(A, B)` pair with a local scan, the pairs compose across devices with a log-depth prefix scan over `ppermute`, and each device replays locally from its incoming state. `A` and `B` are 128x128 for every published Qwen 3.5 size, so the cross-device payload is small.

## Status

Filed as #4968.

- 1,048,576 tokens train end to end on 256 v5p at `ctx=256`, loss 12.923 to 10.583, 485.5 tokens/s/chip. The same configuration trains under Pathways with `enable_single_controller`.
- `tests/unit/gdn_cp_test.py` checks the composition against a sequential single-device reference on a forced 8-device CPU mesh, with a negative control. The whole path matched the stock sequential scan to 1.1e-08 in the earlier all-gather form.
- The local composition needs `jax.checkpoint` on the scan body. Without it the autodiff residuals reach 103 GB at sequence 262,144.
- An all-gather across devices is O(D) per layer, about 38 GB across 48 GatedDeltaNet layers at `ctx=256`. The prefix scan is log2(D) exchanges holding one pair.

The module is `models/gdn_cp.py`. The axis reaches `jax_chunk_gated_delta_rule` as a `cp_axis` keyword built from whichever context knob is set.

Running any of this needs `per_device_batch_size = 1/ici_context_parallelism`. See #4933.

cc @mmcsa

WandLZhang:
Fixes #4932.

Two pspecs pinned the GDN sequence axis to `None`, and the inter-chunk recurrence was a sequential `lax.scan`. Together they mean `ici_context_parallelism` can't shard the GDN sequence at all, so neither 256K nor 1M is reachable at any chip count.

The recurrence is affine in the state:

```
h' = (exp(g_last)·I − k_gᵀ·w)·h + k_gᵀ·u  =  A·h + B
```

Each device folds its local chunks into one `(A, B)` pair with a `lax.scan`, not `associative_scan`, which would materialize the running composition per chunk, 17 GB per device at a million tokens. The pairs then compose across devices with a Hillis-Steele prefix scan over `ppermute`: log2(D) exchanges holding one pair, O(1) in the device count. `A` and `B` are 128x128 at every published Qwen 3.5 size, so the payload doesn't grow with the model.

An all-gather composes the pairs correctly too, and that's what the first push of this PR did. It materializes `[D, B, H, K, K]` and `[D, B, H, K, V]` on every device, which is O(D) per layer: about 38 GB across the 48 GatedDeltaNet layers at `ctx=256`. That term is what made raising the context axis cost memory instead of saving it.

`qkvz_pspec` had the sequence axis hardcoded replicated, which told XLA to gather the full sequence onto every device: at `ctx=2`, seq 32,768, six live `bf16[2,32768,16,512]` buffers of 1.07 GB each. Found by diffing XLA buffer assignment between `ctx=1` and `ctx=2`.

**Measured** on Qwen3.5-27B, 256 v5p, `ctx=256`, `fsdp=1`, one 4,096-token sequence per device: **1,048,576 tokens**, loss 12.923 to 10.583. Also 524,288 on the same slice, and 131,072 on 64 chips.

## The launch flag this needs

Set `per_device_batch_size` to `1/ici_context_parallelism`. `num_target_devices` counts every mesh axis, so the local batch is `per_device_batch_size × ctx`, and the default grows it with the context axis. With the flag set, memory tracks tokens per device rather than the context degree. See #4933.

## Interaction with #4348

The two can't both be live. The fused kernel returns above the state composition, so `use_pallas` is forced false whenever a context axis is set, and sequence-sharded runs take the XLA scan. @bzantium verified that fallback is loss-preserving: 6e-6 relative on the output, 2e-5 on final state. The two are disjoint rather than stacked, so this goes against `main` and #4348 rebases on top.

## Size

`models/gdn_cp.py` is new, 113 lines. `models/qwen3.py` is +42/-4. Applies to a pristine checkout of `main` with no dependency on #4348 or on my other open PRs.

# Tests

`ici_context_parallelism=4` at sequence 32,768 and `ici_context_parallelism=8` at 262,144, both training with loss descending, on v5p and v6e. 1,048,576 tokens on 256 v5p. Memory is flat in `ctx` at a fixed local shard: about 30.5 GB on a 2,048-token shard whether `ctx` is 2 or 16, so the sizing extrapolates.

`tests/unit/gdn_cp_test.py` covers three things:

- `compose_local`, against the recurrence applied chunk by chunk in its original form. That reference never builds `A` or `B` and never calls `compose`, so it checks the affine reformulation rather than restating it. Reversing the chunk order is the negative control.
- `incoming_state` on a single-device mesh, where a shard has nothing before it.
- `incoming_state` against a sequential composition on a forced 8-device CPU mesh, over a 1-D context mesh, a 2-D fsdp x context mesh, and a tuple axis name. Breaking the chain is the negative control.

The multi-device case runs as a subprocess, so the forced device count takes effect before JAX initializes, following `tests/unit/usp_collective_test.py`. pyink clean at `--line-length=122`.

# Checklist

---

_2026-09-15: the 485.5 tok/s/chip this originally published is withdrawn — a two-step artifact of the step timer. Unprofiled over six steps it is 90.2 at `ctx=256` and 104.9 at `ctx=64`. The runs stand._

### DIFF
--- a/src/maxtext/kernels/attention/gdn_cp.py
+++ b/src/maxtext/kernels/attention/gdn_cp.py
@@ -1,0 +1,130 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Context-parallel evaluation of the GatedDeltaNet inter-chunk recurrence.
+
+The recurrence h_new = A @ h + B is affine in the state, so it composes
+associatively and can be split across a sharded sequence:
+
+    A_i = exp(g_last).I - k_g^T.w_i      B_i = k_g^T.u_i
+
+Each device folds its local chunks into one (A, B) pair, the pairs compose
+across devices with a prefix scan, and each device then replays its chunks from
+the state that reaches it. Composition is associative but not commutative, so
+device order is the sequence order. `context_parallel_load_balance` permutes
+that order and `configs/types.py` rejects the combination.
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+from jax import lax
+
+_PREC = jax.lax.Precision.HIGHEST
+
+
+def compose(left, right):
+  """(A_r, B_r) . (A_l, B_l) = (A_r @ A_l, A_r @ B_l + B_r)."""
+  A_l, B_l = left
+  A_r, B_r = right
+  return (
+      jnp.matmul(A_r, A_l, precision=_PREC),
+      jnp.matmul(A_r, B_l, precision=_PREC) + B_r,
+  )
+
+
+def compose_local(w, u, k, g):
+  """Fold this device's chunks into one affine map, in O(1) memory.
+
+  lax.scan rather than associative_scan on purpose: associative_scan would
+  materialize (A, B) and their running composition for every chunk, which is
+  17 GB per device at a million tokens and defeats the point. The parallelism
+  that matters here is across devices, not within one.
+  """
+  k_dim = k.shape[-1]
+  eye = jnp.eye(k_dim, dtype=jnp.float32)
+
+  # jax.checkpoint is required here. lax.scan keeps whatever
+  # the body computes as a backward residual, so A_i and B_i get stacked over
+  # every chunk even though the forward pass only ever needs one at a time. At
+  # sequence 262,144 with ctx=4 that was 103 GB of f32[1024,4,16,128,128] in the
+  # buffer dump. A_i and B_i are cheap to rebuild from w, u, k and g, which are
+  # already live, so recompute them in the backward pass instead of storing them.
+  @jax.checkpoint
+  def body(carry, x):
+    w_c, u_c, k_c, g_c = x
+    g_last = g_c[..., -1]
+    decay = jnp.exp(g_last)[..., None, None]
+    k_g = k_c.astype(jnp.float32) * jnp.exp(g_last[..., None] - g_c)[..., None]
+    k_g_T = k_g.swapaxes(-1, -2)
+    A_i = decay * eye - jnp.matmul(k_g_T, w_c.astype(jnp.float32), precision=_PREC)
+    B_i = jnp.matmul(k_g_T, u_c.astype(jnp.float32), precision=_PREC)
+    return compose(carry, (A_i, B_i)), None
+
+  lead = w.shape[1:-2]
+  init = (
+      jnp.broadcast_to(eye, lead + (k_dim, k_dim)).astype(jnp.float32),
+      jnp.zeros(lead + (k_dim, u.shape[-1]), jnp.float32),
+  )
+  (A_loc, B_loc), _ = lax.scan(body, init, (w, u, k, g))
+  return A_loc, B_loc
+
+
+def incoming_state(A_loc, B_loc, h_init, cp_axis):
+  """State entering this device, plus the final state after all devices.
+
+  log2(D) exchanges of one small matrix pair are the only cross-device traffic
+  in the scheme. Must be called inside a shard_map over `cp_axis`.
+  """
+  # A prefix scan over the device axis, not an all-gather. Gathering the D pairs
+  # materializes [D, B, H, K, K] and [D, B, H, K, V] on every device, so the
+  # composition costs O(D) per device per layer. At ctx=256 on Qwen3.5-27B that
+  # is roughly 38 GB across the 48 GatedDeltaNet layers, and it is what makes
+  # per-device memory grow with the context axis instead of shrinking.
+  #
+  # Hillis-Steele instead: log2(D) ppermute exchanges, one (A, B) pair live at a
+  # time, O(1) in D, and log2(D) backward residuals rather than D.
+  D = lax.axis_size(cp_axis)
+  idx = lax.axis_index(cp_axis)
+
+  # Inclusive prefix: each device ends holding the composition of ranks 0..idx.
+  # compose(left, right) applies left first, so the pair arriving from the
+  # earlier rank is the left operand.
+  a_run, b_run = A_loc, B_loc
+  step = 1
+  while step < D:
+    fwd = [(i, i + step) for i in range(D - step)]
+    a_recv = lax.ppermute(a_run, cp_axis, fwd)
+    b_recv = lax.ppermute(b_run, cp_axis, fwd)
+    a_cmp, b_cmp = compose((a_recv, b_recv), (a_run, b_run))
+    live = idx >= step
+    a_run = jnp.where(live, a_cmp, a_run)
+    b_run = jnp.where(live, b_cmp, b_run)
+    step *= 2
+
+  # A device starts from the exclusive prefix, which is the inclusive prefix of
+  # the device before it. One more shift by a single rank.
+  shift1 = [(i, i + 1) for i in range(D - 1)]
+  a_ex = lax.ppermute(a_run, cp_axis, shift1)
+  b_ex = lax.ppermute(b_run, cp_axis, shift1)
+  carried = jnp.matmul(a_ex, h_init, precision=_PREC) + b_ex
+  h_in = jnp.where(idx == 0, h_init, carried)
+
+  # The final state is the last device's inclusive prefix, broadcast with a
+  # masked psum rather than a gather so this stays O(1) in D too.
+  last = idx == (D - 1)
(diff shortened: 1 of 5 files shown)


src/maxtext/kernels/attention/gdn_cp.py
```diff
@@ -0,0 +1,79 @@
...
+  A_all = lax.all_gather(A_loc, cp_axis, axis=0, tiled=False)
+  B_all = lax.all_gather(B_loc, cp_axis, axis=0, tiled=False)
+  A_cum, B_cum = lax.associative_scan(compose, (A_all, B_all), axis=0)
+
+  idx = lax.axis_index(cp_axis)
+  prev = jnp.maximum(idx - 1, 0)
+  carried = jnp.matmul(A_cum[prev], h_init, precision=_PREC) + B_cum[prev]
+  h_in = jnp.where(idx == 0, h_init, carried)
+  final_h = jnp.matmul(A_cum[-1], h_init, precision=_PREC) + B_cum[-1]
+  return h_in, final_h
```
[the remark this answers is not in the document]
WandLZhang: `lax.axis_index` takes a tuple. On 8 CPU devices, `axis_index(("context",))` and `axis_index(("a","b"))` both return a flat index across the product, `all_gather` reports D=8 for both, and `ppermute` with a tuple axis and flat indices shifts as expected. The 1,048,576-token run used a 1-tuple on 256 chips.

`tests/unit/gdn_cp_test.py` covers it. The third case builds a 2x4 `("context", "context_usp_ulysses")` mesh and passes the tuple through unchanged.

On the suggestion itself: it restores the all-gather this PR replaces. Gathering the pairs materializes `[D, B, H, K, K]` and `[D, B, H, K, V]` on every device, O(D) per layer, about 38 GB across the 48 GatedDeltaNet layers at `ctx=256`. The prefix scan is log2(D) exchanges holding one pair.

igorts-git (approved):

darisoy (approved):

darisoy:
Thanks for the PR!

Regarding file organization: would it make sense to move `src/maxtext/models/gdn_cp.py` to `src/maxtext/kernels/attention/` (e.g., `src/maxtext/kernels/attention/gdn_cp.py` or `src/maxtext/kernels/attention/gated_delta_network_cp.py`)?

`src/maxtext/models/` is generally reserved for top-level model family architectures (`qwen3.py`, `deepseek.py`, etc.), whereas attention/recurrence kernels and context parallelism collective utilities live under `src/maxtext/kernels/attention/` (where `gated_delta_network.py`, `tokamax_ring_attention.py`, `ulysses_attention.py`, and `usp_attention.py` reside).

src/maxtext/kernels/attention/gdn_cp.py
```diff
@@ -0,0 +1,113 @@
+"""Context-parallel evaluation of the GatedDeltaNet inter-chunk recurrence.
```
NuojCheng, line 14: `context_parallel_load_balance` defaults as true with logic happened in https://github.com/AI-Hypercomputer/maxtext/blob/ab9f7c50c4934f75d893b295a4e9e8b3012e11da/src/maxtext/utils/train_utils.py#L308. However, context parallelism for linear attention does not need load balance. We need a form of guard to disable load balance otherwise the correctness fails.

tests/unit/gdn_cp_test.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2026 Google LLC
```
NuojCheng, line 1: please add an integration test for hybrid model using context parallelism

src/maxtext/models/gdn_cp.py
```diff
@@ -0,0 +1,113 @@
...
+  # composition costs O(D) per device per layer. At ctx=256 on Qwen3.5-27B that
+  # is roughly 38 GB across the 48 GatedDeltaNet layers, and it is what makes
+  # per-device memory grow with the context axis instead of shrinking.
+  #
+  # Hillis-Steele instead: log2(D) ppermute exchanges, one (A, B) pair live at a
+  # time, O(1) in D, and log2(D) backward residuals rather than D.
+  #
+  # D comes from the static shape of an all-gathered scalar. That is 4D bytes and
+  # gives a Python int usable as a loop bound.
+  D = int(lax.all_gather(jnp.int32(1), cp_axis).shape[0])
```
NuojCheng: please use `jax.lax.axis_size` instead of AG to compute CP rank.

src/maxtext/models/gdn_cp.py
```diff
@@ -0,0 +1,113 @@
+"""Context-parallel evaluation of the GatedDeltaNet inter-chunk recurrence.
+
+The recurrence h_new = A @ h + B is affine in the state, so it composes
+associatively and can be split across a sharded sequence. See
+apply_gdn_context_parallel.py for the derivation and the measurements.
```
NuojCheng: what is `apply_gdn_context_parallel.py`?

src/maxtext/models/qwen3.py
```diff
@@ -881,8 +899,17 @@ def extract_state(c_in, v_len):
...
-      g_beta_pspec = logical_to_mesh_axes((KV_BATCH, None, KV_HEAD), mesh=self.mesh, rules=logical_rules)
+      # LENGTH, not None. The sequence axis was hardcoded to replicated, so
+      # ici_context_parallelism could never shard the GDN sequence while still
+      # consuming the context axis from the mesh -- which is why raising ctx
+      # made memory worse instead of better. The scan handles a sharded
+      # sequence via the two-pass affine composition in models/gdn_cp.py.
+      # Either context axis can carry the sequence. LENGTH maps to both in the
+      # logical rules, so this is correct whichever one is configured.
+      _cp_on = cfg.ici_context_parallelism > 1 or getattr(cfg, "ici_context_usp_ulysses_parallelism", 1) > 1
+      _cp_len = LENGTH if _cp_on else None
```
NuojCheng: repeat with line 636-637, can we combine?

NuojCheng (commented):
The main concern is how linear attention context parallelism works with load balance for a hybrid model architecture. We might need a separate PR fixing it otherwise we can only do ulysses-style CP for now. There could be some correctness issue using current PR. 

Besides that, some minor suggestions about performance and code layout. Thank you for the contribution!

WandLZhang:
Thanks! the load balance point was a real bug. 

configs/types.py now checks qwen3_next and qwen3_5 at ctx>1 with load balance on

@darisoy: gdn_cp.py moved to kernels/attention/; 
lax.axis_size replaces the all-gather;
one gdn_context_axes() helper replaces three copies; 
the apply_gdn_context_parallel.py reference is gone

Integration test added: it runs jax_chunk_gated_delta_rule sharded against unsharded on an 8-device CPU mesh, comparing every output token and the final state. I also added gradient checks. there’s no custom_vjp here, so the backward is JAX's transpose rules composing

src/maxtext/kernels/attention/gdn_cp.py
WandLZhang: fixed in d5679bf. configs/types.py checks qwen3_next and qwen3_5 when ctx>1 and load balance is on

tests/unit/gdn_cp_test.py
WandLZhang: added in d5679bf. _run_end_to_end_checks runs jax_chunk_gated_delta_rule sharded against unsharded on an 8-device CPU mesh, comparing every output token and the final state

also added gradient checks

src/maxtext/models/gdn_cp.py
WandLZhang: done in d5679bf. D = lax.axis_size(cp_axis), works with the tuple axis

src/maxtext/models/gdn_cp.py
WandLZhang: it was in the patch series this PR came from and is not in this repo. removed in d5679bf, the docstring has the derivation now

src/maxtext/models/qwen3.py
WandLZhang: done in d5679bf, one gdn_context_axes(cfg) helper

PR #4969, opened 2026-08-22
MoE ring-of-experts: TC Pallas ring all-gather for the combine backward + x_sorted remat save

ultrons:
Stacked on #4895 (token quantized all-gather); the base branch is that PR's head, so this diff shows only the incremental change.

## What

Two default-off flags for performance improvement on expert-parallel MoE (ring-of-experts path), plus one new kernel file:

- **`moe_ring_cotangent_ag`** — the backward cotangent all-gather of the ring-of-experts combine reduce-scatter runs on a TensorCore Pallas bidirectional ring kernel instead of the XLA collective. As an XLA collective this all-gather can be placed on the SparseCore collective-offload queue and serialize behind other SC work; issuing its ICI DMAs from the TensorCore lets the scheduler overlap it with backward compute. The forward is unchanged (plain `psum_scatter`), so a remat recompute re-traces the plain collective and no Pallas DMA runs inside a rematted region. Numerically equal to `lax.all_gather` (pure tiled data move).
- **`moe_x_sorted`** (RematLocation, default `remat`) — `device` saves the routed (expert-sorted) MoE input and its small routing/metadata bundle across the remat boundary, so the backward loads them instead of re-running the dispatch token all-gather and sort. Batch-size sensitive: intended for small per-device batch; at larger per-device batch the save exceeds HBM (see repro notes).

New file `src/maxtext/kernels/ring_ag.py`: bidirectional store-and-forward ring all-gather (TC Pallas, ICI DMA), used inside the MoE `shard_map` with a caller-supplied `collective_id` (avoids barrier-semaphore collisions with other in-flight Pallas collectives) and an explicit `CostEstimate` so the latency-hiding scheduler accounts for the DMA cost.

Both compose with #4895's `moe_quantize_token_all_gather` (the `moe_x_sorted` tag is applied leaf-wise so a QArray input is saved correctly).

## Repro

DeepSeek-V3 671B on v7x, the recipe from #4895 with the new flags added:

```
python3 -m maxtext.trainers.pre_train.train maxtext/configs/base.yml \
  model_name=deepseek3-671b per_device_batch_size=1.0 max_target_length=4096 \
  ici_fsdp_parallelism=64 ici_expert_parallelism=8 \
  attention=flash sparse_matmul=true megablox=true use_tokamax_gmm=true use_gmm_v2=true \
  use_tokamax_splash=true use_ring_of_experts=true use_ragged_sort=true \
  use_custom_sort_vjp=true ragged_buffer_factor=2.0 num_moe_token_chunks=2 \
  remat_policy=custom decoder_layer_input=offload context=device out_proj=device \
  quantization=fp8_full use_qwix_quantization=true \
  weight_quantization_calibration_method=fixed,-224,224 \
  act_quantization_calibration_method=fixed,-224,224 \
  bwd_quantization_calibration_method=absmax \
  moe_quantize_token_all_gather=true \
  moe_ring_cotangent_ag=true moe_x_sorted=device \
  dataset_type=synthetic use_random_routing=true steps=20
```

- `moe_ring_cotangent_ag`: any EP degree with ring-of-experts; loss matches the flag-off run (the backward all-gather is bit-exact).
- `moe_x_sorted=device`: use with small per-device batch. AOT compile-checked: fits at `per_device_batch_size=1.0` on a 4x8x8 mesh; at `per_device_batch_size=4.0` the saved activations exceed HBM (drop this flag or keep `remat` there).
- Flags off reproduce the existing behavior byte-for-byte (`moe_x_sorted` tags are inert under `remat`; the ring branches are not taken).

## Validation

- AOT compile (virtual `tpu7x` topology, no hardware) green for: the #4895 recipe + `moe_ring_cotangent_ag` at `per_device_batch_size=4.0`, and both flags at `per_device_batch_size=1.0`.
- Ring kernel semantics validated on v7x against `lax.all_gather(..., tiled=True)` (bit-exact).
- Loss parity vs flag-off verified on a 512-chip DeepSeek-V3 run (identical lm_loss trajectory).

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -884,6 +884,10 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  moe_quantize_token_all_gather: bool = Field(
+      False,
+      description="Whether to quantize token activations to FP8 before All-Gather across EP shards in Ring of Experts.",
+  )
   moe_dispatch_no_expert_sharding: bool = Field(
       False,
       description=(
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -884,6 +884,15 @@ class MoEGeneral(BaseModel):
       False,
       description="Whether to use Ring of Experts for sparse matmul expert parallelism.",
   )
+  moe_ring_cotangent_ag: bool = Field(
+      False,
+      description=(
+          "Run the BACKWARD cotangent all-gather of the ring-of-experts combine reduce-scatter on "
+          "a TensorCore Pallas ring kernel instead of the XLA collective (which can serialize on "
+          "the SparseCore collective-offload queue). Forward is unchanged. For performance "
+          "improvement; numerically equal to lax.all_gather."
+      ),
+  )
   moe_quantize_token_all_gather: bool = Field(
       False,
       description="Whether to quantize token activations to FP8 before All-Gather across EP shards in Ring of Experts.",
@@ -1320,6 +1329,15 @@ class RematAndOffload(BaseModel):
   moe_mlpwi_0: RematLocation = Field(
       RematLocation.REMAT,
       description="Remat policy for the first part of a gated MoE's output.",
+  )
+  moe_x_sorted: RematLocation = Field(
+      RematLocation.REMAT,
+      description=(
+          "Remat policy for the routed (post-dispatch, expert-sorted) MoE input plus its small "
+          "routing/metadata bundle. 'device' saves them across the remat boundary so the backward "
+          "does not re-run the dispatch token all-gather and sort; the expert GMMs re-run from the "
+          "saved tensor. Default 'remat' recomputes (existing behavior)."
+      ),
   )
   moe_mlpwi_1: RematLocation = Field(
       RematLocation.REMAT,
@@ -3423,6 +3441,7 @@ class MaxTextConfig(
           "context",
           "mlpwi",
           "moe_mlpwi_0",
+          "moe_x_sorted",
           "moe_mlpwi_1",
           "moe_mlpwo",
           "mlpwi_0",
@@ -4564,6 +4583,7 @@ class RLConfig(
           "context",
           "mlpwi",
           "moe_mlpwi_0",
+          "moe_x_sorted",
           "moe_mlpwi_1",
           "moe_mlpwo",
           "mlpwi_0",
--- a/src/maxtext/kernels/ring_ag.py
+++ b/src/maxtext/kernels/ring_ag.py
@@ -131,14 +131,12 @@ def _kernel(w_ref, o_ref, send_sem, recv_sem, local_sem, *,
   # 1. place our own shard at its tiled slot: chunk index = sum coord_j * strides[j].
   my_chunk = sum(lax.axis_index(axes[j]) * strides[j] for j in range(len(axes)))
   my_start = my_chunk * chunk
-  pltpu.make_async_copy(
+  local_copy = pltpu.make_async_copy(
       w_ref.at[_full_index(w_ref, gather_dim, 0, chunk)],
       o_ref.at[_full_index(o_ref, gather_dim, my_start, chunk)],
-      local_sem).start()
-  pltpu.make_async_copy(
-      w_ref.at[_full_index(w_ref, gather_dim, 0, chunk)],
-      o_ref.at[_full_index(o_ref, gather_dim, my_start, chunk)],
-      local_sem).wait()
+      local_sem)
+  local_copy.start()
+  local_copy.wait()
   # 2. nested rings, innermost axis first (largest k -> smallest stride).
   for k in range(len(axes) - 1, -1, -1):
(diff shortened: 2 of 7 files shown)


src/maxtext/kernels/ring_ag.py
```diff
@@ -0,0 +1,193 @@
...
+
+  def block_start(u):
+    return base_rows + lax.rem(u + s, s) * cur_rows
+
+  prev = [None] * len(lanes)
+  for i in range(s - 1):
+    for li, (si, nb, sign, ss, sl) in enumerate(lanes):
+      start = block_start(my - i if sign < 0 else my + i)
+      if prev[li] is not None:
+        prev[li].wait()
```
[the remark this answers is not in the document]
ultrons: No race: `pltpu.make_async_remote_copy(...).wait()` waits on **both** semaphores — the send semaphore and the local `recv_sem` (the split `wait_send()` / `wait_recv()` forms exist precisely because plain `wait()` covers both). So at step `i`, `prev[li].wait()` blocks on `recv_sem.at[si]`, which is signaled exactly by the step `i-1` incoming block from the ring neighbor — the block step `i` forwards. The signal/wait accounting balances: one arrival per step per lane, one recv-wait consumed per step, and the trailing `for d in prev: d.wait()` consumes the final arrival. Adding an extra `pl.semaphore_wait(recv_sem.at[si], i)` would double-count the arrivals and deadlock. The kernel is validated bit-exact against `jax.lax.all_gather(..., tiled=True)` on v7x hardware from 8 to 512 chips.

src/maxtext/kernels/ring_ag.py
```diff
@@ -0,0 +1,193 @@
...
+  my_chunk = sum(lax.axis_index(axes[j]) * strides[j] for j in range(len(axes)))
+  my_start = my_chunk * chunk
+  pltpu.make_async_copy(
+      w_ref.at[_full_index(w_ref, gather_dim, 0, chunk)],
+      o_ref.at[_full_index(o_ref, gather_dim, my_start, chunk)],
+      local_sem).start()
+  pltpu.make_async_copy(
+      w_ref.at[_full_index(w_ref, gather_dim, 0, chunk)],
+      o_ref.at[_full_index(o_ref, gather_dim, my_start, chunk)],
+      local_sem).wait()
```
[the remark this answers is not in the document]
ultrons: There is no duplicate transfer — `make_async_copy(...)` builds a descriptor and `.wait()` only waits the semaphore (it does not issue a DMA), so the second construction was a wait-by-reconstruction, not a second copy. That said, the single-object form is clearer; applied in the latest push.

src/maxtext/layers/moe.py
```diff
@@ -57,6 +58,41 @@
 COMBINE = "combine"
 
 
+# Distinct barrier-semaphore ids for the in-MoE Pallas ring collectives; must not collide with
+# any other collective_id in flight in the same program.
+_RING_CT_AG_COLLECTIVE_ID = 55  # moe_ring_cotangent_ag: backward combine-cotangent ring all-gather
```
RissyRan, line 63: How do we know `55` is not collide with any other collective_id in flight in the same program?

src/maxtext/kernels/ring_ag.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def _neighbor(all_axes, axis, delta):
+  """device_id MESH dict: step `delta` along `axis` (wrap), hold every other mesh axis fixed."""
+  size = lax.axis_size(axis)
+  nxt = lax.rem(lax.axis_index(axis) + delta + size, size)
+  return {a: (nxt if a == axis else lax.axis_index(a)) for a in all_axes}
+
+
+def _ring_stage(o_ref, all_axes, axes, sizes, strides, k, chunk, gather_dim,
```
RissyRan, line 59: Shall we write some unit tests for this?

RissyRan (approved):
LGTM at high level, and I think we should add some unit tests for ring_ag.py.

For GitHub PR, we will need 2 approvers to trigger the pull ready label :)

src/maxtext/kernels/ring_ag.py
```diff
@@ -0,0 +1,191 @@
...
+      local_sem)
+  local_copy.start()
+  local_copy.wait()
+  # 2. nested rings, innermost axis first (largest k -> smallest stride).
+  for k in range(len(axes) - 1, -1, -1):
+    _ring_stage(o_ref, all_axes, axes, sizes, strides, k, chunk, gather_dim,
+                send_sem, recv_sem, bidi, split_dim, pipe)
+
+
+def ring_all_gather(x, mesh, gather_axes, gather_dim, collective_id, *, bidi=True, pipe=1):
```
RissyRan, line 146: It seems we haven't passed this `pipe` value anywhere?

src/maxtext/configs/types.py
```diff
@@ -1321,6 +1330,15 @@ class RematAndOffload(BaseModel):
       RematLocation.REMAT,
       description="Remat policy for the first part of a gated MoE's output.",
   )
+  moe_x_sorted: RematLocation = Field(
```
RissyRan, line 1333: Could you also help add into base.yml file? It's no op, but we usually keep 2 files aligned.

PR #4951, opened 2026-08-20
feat: Auto-detect and convert Tunix post-training checkpoints on load

hsuan-lun-chiang:
## Overview
This PR updates the checkpoint loader in MaxText to auto-detect and dynamically convert Tunix post-training checkpoints on load, allowing downstream MaxText pre-training runs to restore from Tunix DPO/SFT checkpoints seamlessly.

## Key Changes
- **Auto-Detection**: Identifies Tunix-formatted checkpoints by inspecting the checkpoint layout (presence of `model_params` collection).
- **On-the-fly PyTree Reshaping**:
  - Dynamically unwraps Optax `inject_hyperparams` wrapper states (`_drop_inject_hyperparams`).
  - Strips LoRA adapter layers (`_drop_adapter_level`) when mapping from adapter states to base pre-training states, handling dicts, lists, tuples, and namedtuples.
  - Correctly repacks base adapter levels (`_add_adapter_level`) when querying metadata or loading checkpoints with `has_base=True`.
- **`model_creation_utils.from_pretrained`**: Directs parameter inspection and single-item PyTree restore to `model_params` when loading Tunix checkpoints.

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run end-to-end tests tests and provided workload links above if applicable.
- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #4976, opened 2026-08-23
Convert Tunix-layout post-training checkpoints for pre-training

ecnal-cienet:
# Description

Stacked on #4975. Retarget to `main` once that merges.

Post-training now writes MaxText's layout, and the trainers still restore the old one, so a run resumed from an existing checkpoint rewrites itself on its next save. Nothing does that for the checkpoints nobody is going to resume, and pre-training cannot read them: it looks for weights in the Linen `params` collection, and the old layout keeps the whole model state under `model_params` instead.

`maxtext.checkpoint_conversion.tunix_to_maxtext` rewrites them in place of a training run:

```
python -m maxtext.checkpoint_conversion.tunix_to_maxtext src/maxtext/configs/base.yml \
  --source gs://bucket/run/checkpoints \
  model_name=llama3.1-8b scan_layers=True \
  base_output_directory=gs://bucket/run_converted \
  hardware=cpu skip_jax_distributed_system=True
```

It finds every step still in the old layout, including the `actor/<step>` level RL writes, and skips any that already has an `items/`. `--dry_run` reports what each step would hold without writing.

The split between `params` and `nnx_aux` is why this needs a model rather than being a rename: weights and the rng counters that drive dropout sit mixed together under `model_params`, and only the model says which is which. Weights are never materialised — an abstract model supplies the classification and the arrays move through as numpy, so it runs on CPU.

It refuses a model that does not match the checkpoint, by name and by shape. Both mistakes are easy to make and neither announces itself: splitting by the wrong model routes the leaves it does not recognise into `nnx_aux`, and a config that differs only in size has every name right and fails much later, in the middle of a read. Both happened while developing this, which is why the check is there.

# Tests

`tests/post_training/unit/tunix_to_maxtext_test.py` — 9 passed. Built on a checkpoint the test writes in the old layout, so it depends on nothing external. It covers that every leaf survives the split, that rng state stays out of the params collection, that the `inject_hyperparams` shell is stripped, that RL's `actor/<step>` steps are found, that already-converted steps are skipped, and that a mismatched model is refused.

The test that matters most loads a converted checkpoint back through `checkpointing.load_params_from_path` — pre-training's own loader — and checks the values.

Also run against a real 2026-06 post-training checkpoint on GCS (gpt3-52k, scanned): 46 arrays split into 30 aux and 16 weights, written, and loaded back through the pre-training loader as 16 tensors totalling 548,080 parameters, no NaNs.

# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/tunix_to_maxtext.py
+++ b/src/maxtext/checkpoint_conversion/tunix_to_maxtext.py
@@ -1,0 +1,265 @@
+"""Rewrites post-training checkpoints written in Tunix's layout into MaxText's.
+
+Post-training used to write
+
+  <run>/checkpoints/<step>/model_params/       the whole nnx.state(model)
+  <run>/checkpoints/<step>/optimizer_state/    opt_state and step
+
+which pre-training cannot read. It now writes
+
+  <run>/checkpoints/<step>/items/              params, opt_state, step, nnx_aux
+
+Trainers still restore the old layout, so a run resumed from one converts itself on its next
+save. This script is for the checkpoints nobody is going to resume: it rewrites them in place of
+a training run, so pre-training can pick them up.
+
+The split between `params` and `nnx_aux` is the reason this needs a model rather than being a
+rename. Both live mixed together under model_params -- weights beside the rng counters that drive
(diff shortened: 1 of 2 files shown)


ecnal-cienet:
Close as it duplicated with #4951

PR #4983, opened 2026-08-24
Fix multiprocess dataloader CPU iteration leak via weakref sweep

snehalv2002:
## Overview & Motivation

Resolves [b/532239140](https://b.corp.google.com/issues/532239140): *[Pathways Elasticity/RemoteIterator] Slice scale up using Grain Colocated Data Loading on v6e cluster creates large step times (200s+)*

In elastic training workloads on Google Cloud TPU (e.g. v6e clusters using Multi-Tier Checkpointing / Pathways elasticity), cluster preemption or worker disruption causes the framework to gracefully scale down (e.g. from 2 slices to 1 slice) and subsequently scale back up once the interrupted slice recovers.

When using **Grain with Colocated Python Data Loading** (`colocated_python_data_input=True` and `grain_worker_count > 0`), workloads experienced severe, permanent performance degradation immediately following slice scale-up:
- Healthy step times before interruption: **~1.85s / step** (~4,400 Tokens/s/device).
- Post-scale-up step times: degraded exponentially to **60s – 120s+ (and up to 200s+)**, eventually triggering timeout or hang detection.

---

## Root Cause Analysis

Grain data loading executes inside the companion **`colocated-python-sidecar`** container running alongside the TPU worker pod. 

1. **Process Reparenting to PID 1**: When a worker pod or Pathways process crashes and recovers, `main.py` inside the sidecar is restarted by the container entrypoint loop (`while true; do python /app/main.py; done`).
2. **Zombie / Orphaned Multiprocessing Workers**: While `main.py` restarts, its previous child worker processes (`multiprocessing.spawn` and `resource_tracker`) are **not** reaped by `/bin/sh`. The Linux kernel reparents them to **`PID 1`**.
3. **Hidden from Python GC**: Because these orphaned processes no longer exist in the new Python process’s `multiprocessing.process._children`, standard Python cleanup (`multiprocessing.active_children()`) cannot detect or terminate them.
4. **GCS Contention & Data Starvation**: The orphaned workers remain permanently alive in the background holding open thousands of file descriptors and TCP streams to Google Cloud Storage array-record chunks. When the new `main.py` spawns fresh workers upon scale-up, they contend with the zombie processes for bandwidth and lock files, causing `ExternalSidecar::Execute` batch fetching to block for 60s–120s per step.

---

## Proposed Changes

1. **Direct `/proc` Orphan Process Cleanup (`multihost_dataloading.py`)**:
   - Added `_cleanup_orphaned_multiprocessing_processes(force_all=False)`: Directly scans `/proc` for any `multiprocessing.spawn` or `multiprocessing.resource_tracker` processes whose parent process (`PPid`) is `1` (or dead) and terminates them via `SIGTERM` followed by `SIGKILL`.
   - Called during `RemoteIterator.__init__()` prior to `self.reset()`, and on `RemoteIterator.close()` and `MultiHostDataLoadIterator.close()`.

2. **Singleton Object Store & Global Iterator Purge**:
   - Explicitly sweeps `SINGLETON_OBJECT_STORE._storage` on worker nodes during iterator re-initialization to ensure stale `RemoteIterator` instances do not linger in memory.
   - Added `cleanup_all_iterators()` using `weakref.WeakSet` to track and force `.close()` on any active iterators across the process.

3. **Elastic Resilience Hooks (`elastic_utils.py`)**:
   - Integrated `cleanup_all_iterators()` into `maybe_elastic_scale_up()` prior to raising `ScaleUpSignalError()`.
   - Integrated `cleanup_all_iterators()` into `cleanup_iterators_and_checkpoints()` as the default `on_elastic_event_callback` for `elastic_retry()`.

---

## Empirical Verification

Tested on cluster `bodaborg-v6e-256-lcscld-c` (`southamerica-west1-a`, project `tpu-prod-env-one-vm`) with Llama 3.1 8B on 2 slices of TPU v6e:

1. **Initial 2-Slice Baseline (Steps 0–6)**:
   - Stable step duration: **1.860s – 1.865s** (~4,410 Tokens/s/device).
2. **Fault Injection & 1-Slice Degradation (Steps 0–38)**:
   - Injected `SIGILL` into `worker-0-0` at step 4+. Pathways smoothly degraded to 1 slice (16 devices) at **~1.87s/step**.
3. **Scale-Up to 2 Slices & Defect Reproduction (Steps 27–31)**:
   - Pathways re-established 2 slices.
   - Without orphan cleanup, step times immediately stalled at **85.96s (Step 27)**, **123.20s (Step 28)**, **66.48s (Step 29)**, and **120.82s (Step 30)** due to orphaned PIDs 4716 and 4718.
4. **Post-Cleanup Recovery & Steady State (Steps 32–1000)**:
   - Upon invoking `_cleanup_orphaned_multiprocessing_processes()`, orphaned processes were terminated.
   - Step times collapsed immediately from **123.2s $\to$ 1.848s – 1.95s/step** (~4,420 Tokens/s/device).
   - Training ran cleanly and steadily all the way to **step 1,000** without further degradation or `DATA_LOSS` errors.
5. **Raw Log Verification (Internal GCP Links)**
    - Patched Execution (Stable 2.0s post-recovery):** [Job Logs](https://cloudlogging.app.goo.gl/FA2KsPUuz1PZimV59)

---

## Operational Deployment Note

Because JAX `colocated_python.colocated_python_class(RemoteIterator)` serializes classes by global module reference (`('maxtext.input_pipeline.multihost_dataloading', 'RemoteIterator')`), the companion `colocated-python-sidecar` container resolves and imports this code from its own container filesystem (`/app/maxtext`).

**Action Item upon Merge**: The companion container image (`us-docker.pkg.dev/.../sidecar_mtc`) will need to be rebuilt and released from `main` so that the running sidecars incorporate these changes.

# Checklist


src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -183,12 +279,57 @@ class RemoteIterator:
   "iterator class for using colocated python class"
 
   def __init__(self, get_ds_fn, preprocessing_fn, global_shape, checkpoint_path, elastic=False):
+    # Step 1: Clean up any previous RemoteIterator stored in SINGLETON_OBJECT_STORE on the worker
+    try:
+      from jax.experimental.colocated_python.obj_backend import SINGLETON_OBJECT_STORE
```
SujeethJinesh, line 326: This seems to be cleaning up all the colocated python objects rather than just the iterators. I think this can be dangerous because the user can be running multiple things (not just data loading, e.g. MTC).

src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -46,6 +49,82 @@
 
 from maxtext.utils import max_logging
 
+_GLOBAL_REMOTE_ITERATORS = weakref.WeakSet()
+_GLOBAL_MULTIHOST_ITERATORS = weakref.WeakSet()
+
+def _cleanup_orphaned_multiprocessing_processes(force_all: bool = False):
```
SujeethJinesh: Can we restrict to terminating workers only owned by the retired data loader? I think `force_all=True` would kill normal and healthy multi-processing workers too.

src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -183,12 +279,57 @@ class RemoteIterator:
   "iterator class for using colocated python class"
 
   def __init__(self, get_ds_fn, preprocessing_fn, global_shape, checkpoint_path, elastic=False):
```
SujeethJinesh, line 323: Could you add a regression unit test to confirm that this doesn't impact MTC iterators or checkpointers that use colocated python?

SujeethJinesh (changes requested):

src/maxtext/input_pipeline/multihost_dataloading.py
snehalv2002: made a change so object eviction is restricted to objects of type `RemoteIterator`

src/maxtext/input_pipeline/multihost_dataloading.py
snehalv2002: 1. Removed force_all=True entirely from _cleanup_orphaned_multiprocessing_processes()and all call sites.
2. /proc scanning now terminates only true orphan processes reparented to PID 1 (ppid == 1). Healthy worker processes belonging to other components (ppid > 1) are never touched.
3. Added _terminate_iterator_workers() and _close_iterator() to cleanly close and terminate worker pools directly owned by the iterator/dataloader itself.
4. Stored self.dataloader as an instance attribute in RemoteIterator.reset() to allow proper teardown on close().

src/maxtext/input_pipeline/multihost_dataloading.py
snehalv2002: added 3 tests:

1. `test_singleton_object_store_cleanup_preserves_mtc_objects`: Verifies that MTC checkpointer objects in SINGLETON_OBJECT_STORE remain intact while old RemoteIterator objects are evicted and closed.
2. `test_orphan_cleanup_preserves_healthy_multiprocessing_processes`: Mocks /proc and verifies that only orphaned processes (PPid == 1) are signaled, while processes with active parents (PPid > 1) or other processes are untouched.
3. `test_terminate_iterator_workers_and_close`: Verifies targeted worker termination and nested iterator closing.

SujeethJinesh (approved):
Thank you Snehal!

src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -46,6 +50,127 @@
...
+          continue
+
+        ppid = None
+        with open(f"/proc/{pid}/status", "r", encoding="utf-8") as f:
+          for line in f:
+            if line.startswith("PPid:"):
+              ppid = int(line.split()[1])
+              break
+
+        if ppid == 1:
```
SurbhiJainUSC, line 136: should we do:
```
# Ensure we don't kill our own workers if we are running as PID 1
if ppid == 1 and current_pid != 1:
   max_logging.log(f"CLEANUP: Terminating orphaned multiprocessing process {pid} (PPid=1)")

```

src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -46,6 +50,127 @@
...
+          try:
+            os.kill(pid, signal.SIGTERM)
+          except ProcessLookupError:
+            continue
+          except Exception as e:
+            max_logging.log(f"CLEANUP: SIGTERM failed for {pid}: {e}")
+
+          time.sleep(0.05)
+          try:
+            os.kill(pid, signal.SIGKILL)
```
SurbhiJainUSC: Can we do this outside of for loop once for all the pids collected to kill?

src/maxtext/input_pipeline/multihost_dataloading.py
snehalv2002: added a check so that if the process is running as container PID 1 direct child workers are treated as active workers and not killed

src/maxtext/input_pipeline/multihost_dataloading.py
snehalv2002: done

src/maxtext/input_pipeline/multihost_dataloading.py
```diff
@@ -46,6 +50,141 @@
...
+      except Exception as e:
+        max_logging.log(f"CLEANUP: Error checking PID {pid}: {e}")
+
+    if not orphan_pids:
+      return
+
+    for pid in orphan_pids:
+      max_logging.log(f"CLEANUP: Terminating orphaned multiprocessing process {pid} (PPid=1)")
+      try:
+        os.kill(pid, signal.SIGTERM)
```
SurbhiJainUSC, line 149: We can actually send `SIGTERM` at line 137 for every process. The reason I asked to move sIGKILL out of the loop is we were adding a sleep of 0.05s for every child process. So, if there are 10 processes to kill, then we are unnecessary adding a sleep of 0.5s.
```
for entry in os.listdir('/proc'):
   if pid==1:
      os.kill(pid, signal.SIGTERM)
      orphan_pids.append(pid)

time.sleep(0.05)
for pid in orphan_pids:
   os.kill(pid, signal.SIGKILL)
```

SurbhiJainUSC (approved):

PR #5011, opened 2026-08-26
Support on-the-fly FP8 block dequantization in checkpoint utility for Qwen3.5

snehalv2002:
# Description

This PR adds support for **on-the-fly FP8 block dequantization** during Hugging Face to MaxText checkpoint conversion (`to_maxtext.py`), with implementation for the **Qwen3.5 model family** (`qwen3.5-35b-a3b`, `qwen3.5-397b-a17b`).

### Problem & Motivation
Currently, converting quantized FP8 Hugging Face checkpoints to MaxText requires dequantizing the entire model to `bfloat16` upfront prior to running the conversion utility. For large models (e.g. 35B / 397B MoE models, DeepSeek, Kimi), this intermediate step requires massive amounts of local disk space (up to 4x the compressed checkpoint size when going from FP4 to BF16, up to several terabytes) and substantial redundant I/O.

### Solution
This PR adds on-the-fly dequantization directly inside `to_maxtext.py`:
1. **FP8 Dtype Safetensors Support**: Registered `np.float8_e4m3fn`, `np.float8_e5m2`, and `np.bfloat16` for zero-copy safetensors deserialization in `to_maxtext.py`.
2. **Recursive Composite Key Support**: Added `_recursive_get_tensor` in `to_maxtext.py` and `tensor_handling.py` to recursively fetch multi-source composite tuple keys `((weight, scale_inv), ...)` from Hugging Face on-demand.
3. **2D Tile Block Dequantization**: Implemented `dequantize_fp8_block(weight, scale_inv, block_size=(128, 128))` in `param_mapping.py` to expand fine-grained $128 \times 128$ inverse scales across FP8 tiles into `bfloat16`.
4. **Dynamic Key Pairing & Hooks**: Updated `QWEN3_5_MAXTEXT_TO_HF_PARAM_MAPPING` with `hf_w(key)` to automatically pair quantized weights with `weight_scale_inv` while leaving unquantized layers intact, and updated transformation hooks to dequantize on-the-fly via `maybe_dequantize`.
5. **Automatic Config Detection**: Dynamically inspects `quantization_config` in `config.json` to enable FP8 pairing automatically.

### Scope & Model Support
- Supports the **Qwen3.5 model family** in MaxText (`qwen3.5-35b-a3b`, `qwen3.5-397b-a17b`), dynamically parameterized from Hugging Face `config.json`.
- Provides the core dequantization foundation for other block-scaled FP8 architectures (DeepSeek-V3 / DeepSeek-V3.2).

# Tests

1. **Full Parameter Parity Verification (All 673 Parameters)**:
   - Evaluated `Qwen/Qwen3.5-35B-A3B-FP8` against golden reference MaxText checkpoint `gs://maxtext-model-checkpoints/qwen3.5-35b-a3b/unscanned/0/items`:
   - **Unquantized Parameters** (333 / 333): Exact bit-for-bit match ($\text{MAE} = 0.0000\text{e}+00$, $\text{Cosine Similarity} = 1.000000$).
   - **Dequantized FP8 Parameters** (340 / 340): All passed with $\text{Cosine Similarity} \ge 0.99965$ (Mean Cos Sim: $0.999648$, Mean Relative Error: $2.65\%$), matching expected FP8 tile numerical fidelity.

2. **Full End-to-End Checkpoint Conversion**:
   - Converted `Qwen/Qwen3.5-35B-A3B-FP8` unscanned checkpoint to GCS (`gs://snehalv-data/qwen35/otfd/unscanned/0/items`) in 29 minutes with peak memory of 172 GB.

3. **End-to-End Decoding on TPU v5p**:
   - Verified execution of `decode.py` on remote v5p-8 TPU VM (`maxtext-single-host-3-v5p-8`) using the converted checkpoint:
     - Prompt: *"An attention function can be described as mapping a query and a set of key-value pairs to an output, where the query, keys, values, and outputs are all vectors. The output is "*
     - Output: *" computed as a weighted sum of the values, where the weight assigned to each value is computed by a compatibility function of the query with the corresponding key..."* (Generated coherently and accurately).

# Checklist

### DIFF
--- a/src/maxtext/checkpoint_conversion/utils/param_mapping.py
+++ b/src/maxtext/checkpoint_conversion/utils/param_mapping.py
@@ -905,18 +905,14 @@ def QWEN3_5_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=Fals
   """
   text_cfg = config.get("text_config", config)
   num_main_layers = text_cfg["num_hidden_layers"]
-  num_experts = text_cfg.get("num_experts", text_cfg.get("num_local_experts", maxtext_config.num_experts if hasattr(maxtext_config, "num_experts") else 256))
+  num_experts = text_cfg.get("num_experts")
   layer_cycle_interval = maxtext_config.inhomogeneous_layer_cycle_interval
 
   quant_cfg = config.get("quantization_config", {})
-  is_fp8 = (
-      (isinstance(quant_cfg, dict) and quant_cfg.get("quant_method") == "fp8")
-      or ("fp8" in str(getattr(maxtext_config, "hf_model_path", "")).lower())
-      or ("fp8" in str(getattr(maxtext_config, "load_parameters_path", "")).lower())
-      or ("fp8" in str(getattr(maxtext_config, "model_name", "")).lower())
-  )
+  is_fp8 = isinstance(quant_cfg, dict) and quant_cfg.get("quant_method") == "fp8"
 
   def hf_w(key: str):
+    """Maps an HF parameter key to a (weight, scale_inv) tuple if quantized, or returns the key as is."""
     if is_fp8:
       unquantized_patterns = [
           "embed_tokens",
@@ -1214,10 +1210,7 @@ def QWEN3_5_MAXTEXT_TO_HF_PARAM_HOOK_FN(config, maxtext_config, scan_layers=Fals
   """
 
   def transpose(input_tensor, target_shape=None):
-    input_tensor = maybe_dequantize(input_tensor)
-    if target_shape is not None and not saving_to_hf:
-      return input_tensor.T.reshape(target_shape)
-    return input_tensor.T
+    return maybe_dequantize(input_tensor).T
 
   def reshape_kernel(input_tensor, target_shape):
     if saving_to_hf:
--- a/src/maxtext/checkpoint_conversion/utils/param_mapping.py
+++ b/src/maxtext/checkpoint_conversion/utils/param_mapping.py
@@ -52,10 +52,60 @@ parameter from the source checkpoint and build the target checkpoint.
 """
 
 import warnings
+import ml_dtypes
 import numpy as np
+
+# Ensure safetensors and numpy recognize FP8 and BF16 types
+np.float8_e4m3fn = ml_dtypes.float8_e4m3fn
+np.float8_e5m2 = ml_dtypes.float8_e5m2
+np.bfloat16 = ml_dtypes.bfloat16
 
 import jax
 import jax.numpy as jnp
+
+
+def dequantize_fp8_block(weight, scale_inv, block_size=(128, 128)):
+  """Vectorized FP8 block dequantization in NumPy/JAX.
+
+  Args:
+    weight: quantized array of shape (..., M, N) with dtype float8_e4m3fn.
+    scale_inv: inverse scale array of shape (..., ceil(M/bk_m), ceil(N/bk_n)).
+    block_size: tuple of block dimensions, default (128, 128).
+
+  Returns:
+    Dequantized array of shape matching weight.
+  """
+  if not isinstance(weight, np.ndarray):
+    weight = np.array(weight)
+  if not isinstance(scale_inv, np.ndarray):
+    scale_inv = np.array(scale_inv)
+
+  m, n = weight.shape[-2], weight.shape[-1]
+  bk_m, bk_n = block_size
+
+  if m % bk_m == 0 and n % bk_n == 0:
+    m_blocks = m // bk_m
+    n_blocks = n // bk_n
+    prefix_shape = weight.shape[:-2]
+    w_reshaped = weight.astype(np.float32).reshape(prefix_shape + (m_blocks, bk_m, n_blocks, bk_n))
+    s_reshaped = scale_inv.astype(np.float32).reshape(prefix_shape + (m_blocks, 1, n_blocks, 1))
+    dequant = (w_reshaped * s_reshaped).reshape(weight.shape)
+  else:
+    s_expanded = np.repeat(np.repeat(scale_inv, bk_m, axis=-2), bk_n, axis=-1)
+    s_expanded = s_expanded[..., :m, :n]
+    dequant = weight.astype(np.float32) * s_expanded.astype(np.float32)
+
+  return dequant
+
+
+def maybe_dequantize(tensor):
+  """Dequantizes tensor if it is a (weight, scale_inv) tuple, otherwise returns tensor as is."""
+  if isinstance(tensor, (tuple, list)) and len(tensor) == 2 and not isinstance(tensor[0], (tuple, list)):
+    w, s = tensor
+    if hasattr(w, "shape") and hasattr(s, "shape"):
+      return dequantize_fp8_block(w, s)
+  return tensor
+
 
 
 def GEMMA3_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=False):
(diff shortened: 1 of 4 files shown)


src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -837,42 +875,72 @@ def reshape_bias(input_tensor, target_shape=None):
...
   - Handles `composite_hf_key`: MaxText key(s) map to multiple HF keys
     - attention-in_proj_qkvz-kernel: (linear_attn.in_proj_qkv.weight, linear_attn.in_proj_z.weight)
-    - attention-in_proj_ba-kernel: (linear_attn.in_proj_b.weight, linear_attn.in_proj_a.weight")
+    - attention-in_proj_ba-kernel: (linear_attn.in_proj_b.weight, linear_attn.in_proj_a.weight)
+  - Supports on-the-fly FP8 block dequantization for FP8 quantized checkpoints
   """
-  num_main_layers = config["text_config"]["num_hidden_layers"]
+  text_cfg = config.get("text_config", config)
+  num_main_layers = text_cfg["num_hidden_layers"]
+  num_experts = text_cfg.get("num_experts", text_cfg.get("num_local_experts", maxtext_config.num_experts if hasattr(maxtext_config, "num_experts") else 256))
```
snehalv2002: this is very hard to understand. also qwen 3.5 has no "num_local_experts" in its config.json. it's just "num_experts". make this more readable

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -837,42 +875,72 @@ def reshape_bias(input_tensor, target_shape=None):
...
   layer_cycle_interval = maxtext_config.inhomogeneous_layer_cycle_interval
 
+  quant_cfg = config.get("quantization_config", {})
+  is_fp8 = (
+      (isinstance(quant_cfg, dict) and quant_cfg.get("quant_method") == "fp8")
+      or ("fp8" in str(getattr(maxtext_config, "hf_model_path", "")).lower())
+      or ("fp8" in str(getattr(maxtext_config, "load_parameters_path", "")).lower())
+      or ("fp8" in str(getattr(maxtext_config, "model_name", "")).lower())
+  )
```
snehalv2002: this again has too much case work. it should only include the exact case of the config.json for the qwen 3.5 config.json. here are the important keys and values from the config from the source:
``` 
"quant_method": "fp8",
"activation_scheme": "dynamic",
"weight_per_tensor": false,
"act_per_tensor": false,
"weight_block_size": [
  128,
  128
]
```
update your code accordingly

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -837,42 +875,72 @@ def reshape_bias(input_tensor, target_shape=None):
...
+          "shared_expert_gate",
+          "mlp.gate",
+      ]
+      if any(pat in key for pat in unquantized_patterns) or key.startswith("lm_head"):
+        return key
+      scale_key = key.replace(".weight", ".weight_scale_inv")
+      if scale_key == key:
+        scale_key = key + "_scale_inv"
+      return (key, scale_key)
+    return key
```
snehalv2002, line 938: what is this function supposed to do? add a docstring. also `quant_cfg` has a key which you can access as `quant_cfg["modules_to_not_convert"]` which is a List of strings representing keys which i guess we do not convert. use that information to make this function cleaner and more robust.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -1131,18 +1208,23 @@ def QWEN3_5_MAXTEXT_TO_HF_PARAM_HOOK_FN(config, maxtext_config, scan_layers=Fals
...
-    - attention-in_proj_ba-kernel: (linear_attn.in_proj_b.weight, linear_attn.in_proj_a.weight"),
+    - attention-in_proj_ba-kernel: (linear_attn.in_proj_b.weight, linear_attn.in_proj_a.weight),
     transformed via `concat_ba_and_transpose` function
+  - Dequantizes FP8 (weight, scale_inv) tuples on-the-fly.
   """
 
   def transpose(input_tensor, target_shape=None):
+    input_tensor = maybe_dequantize(input_tensor)
+    if target_shape is not None and not saving_to_hf:
+      return input_tensor.T.reshape(target_shape)
```
snehalv2002: its not clear why we needed to add these three lines. the original code just had `return input_tensor.T`. i can see why we needed  the `maybe_dequantize` line but what's the reason for adding that `if` block? can you explain? if you think it's not needed maybe we can remove but we should be careful about deleting things since the current version of the code did pass the decode test so we know functionally it works, the comments i make are mostly style/robustness.

src/maxtext/checkpoint_conversion/utils/tensor_handling.py
```diff
@@ -48,8 +48,15 @@ def _binary_chunked_stack(tensors: List[np.ndarray], axis: int) -> np.ndarray:
   return np.concatenate([left, right], axis=axis)
 
 
+def _recursive_get_tensor(getter, key):
```
entrpn, line 51: add type annotations to getter and key.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -837,42 +875,68 @@ def reshape_bias(input_tensor, target_shape=None):
...
-  num_main_layers = config["text_config"]["num_hidden_layers"]
+  text_cfg = config.get("text_config", config)
+  num_main_layers = text_cfg["num_hidden_layers"]
+  num_experts = text_cfg.get("num_experts")
   layer_cycle_interval = maxtext_config.inhomogeneous_layer_cycle_interval
 
+  quant_cfg = config.get("quantization_config", {})
+  is_fp8 = isinstance(quant_cfg, dict) and quant_cfg.get("quant_method") == "fp8"
+
+  def hf_w(key: str):
```
entrpn, line 914: is this specific to qwen models? What if I pass another checkpoint that is not qwen 3.5?

entrpn (commented):
Made some comments, besides that you should add a check for models that are supported by this script. Maybe right after loading the quantization config and raise an exception if that model is not supported.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
snehalv2002: It is specific to qwen models, but that's why I defined it inside `QWEN3_5_MAXTEXT_TO_HF_PARAM_MAPPING`. We'll define something like this inside each model's parameter mapping.

src/maxtext/checkpoint_conversion/utils/param_mapping.py
```diff
@@ -1041,21 +1111,25 @@ def QWEN3_5_MAXTEXT_TO_HF_PARAM_MAPPING(config, maxtext_config, scan_layers=Fals
...
-              f"{prefix}-mlp-routed_experts-wo": f"model.language_model.layers.{i}.mlp.experts.down_proj",
-              (
-                  f"{prefix}-mlp-routed_experts-wi_0",
-                  f"{prefix}-mlp-routed_experts-wi_1",
-              ): f"model.language_model.layers.{i}.mlp.experts.gate_up_proj",
+              f"{prefix}-mlp-routed_experts-wo": [
+                  hf_w(f"model.language_model.layers.{i}.mlp.experts.{e}.down_proj.weight") for e in range(num_experts)
+              ],
+              f"{prefix}-mlp-routed_experts-wi_0": [
+                  hf_w(f"model.language_model.layers.{i}.mlp.experts.{e}.gate_proj.weight") for e in range(num_experts)
```
JamesDeng42, line 1128: Could this break the existing bfloat16 Qwen3.5 conversion path? please double check. 

Could we also add a small mapping regression test for both FP8 and non-FP8 Qwen3.5, for scanned and unscanned modes?

snehalv2002:
Closing this PR in favor of https://github.com/AI-Hypercomputer/maxtext/pull/5024 which adds support for converting from quantized pytorch checkpoints to quantized orbax checkpoints as well as support for loading and using quantized checkpoints in maxtext with on the fly dequantization.

PR #5014, opened 2026-08-26
Mhc expand

muskansh-google:
# Description

the past.

FIXES: b/123456
FIXES: #123456

BUGS: b/123456

# Tests

commands to reproduce.

# Checklist


src/maxtext/layers/mhc.py
```diff
@@ -186,8 +186,12 @@ def __init__(
     )
 
     # Biases
+    if self.config.enable_mhc_lite:
```
dandragona: What is this change doing?

src/maxtext/models/qwen3.py
dandragona: I can't see what `mhc_attention` or `mhc_mlp` are doing, but their names are a code smell to me.

The mHC formula is  X_{L+1} = Res X_L + H_expand (branch(H_reduce X_L))

where branch is the actual layer.

So at the start of every "layer" we reduce the expanded stream -> apply the layer -> then re-expand.

Therefore we shouldn't have any special handling for the attention and MLP blocks within the layer, at that point they should be agnostic of any mHC specialness.

src/maxtext/layers/mhc.py
muskansh-google: This was added by Jetski to match the implementation from the [original mHC-Lite paper](https://openreview.net/pdf?id=5IJX6kvOif) which mentions to avoid scrambling problem by initializing beta_res with +2 or -8 - https://screenshot-v2.corp.google.com/1u2aspuaqlcog

However, I ran the workload by commenting out this change, and it does not affect the loss convergence. So, we can ignore it for now.

PR #5015, opened 2026-08-26
fix(eval mesh): shard attention with the logical axis rules active at eval

NuojCheng:
# Description

**Fixes the eval mesh: attention is sharded with the rules that are actually active at eval, not the ones the train mesh was built from — and only load balances when the input pipeline actually reordered for it.**

`custom_mesh_and_rule_for_eval` swaps `logical_axis_rules` for the eval step, but every other mesh-derived value stays on the train rule — including `context_sharding`, which is inferred once from the *train* rules via `infer_cp_axes(logical_axis_rules)`.

That splits one invariant across two sources. In `_tpu_flash_attention` the Splash kernel is built with `q_seq_shards=cp_size` read from the static `config.context_sharding`, and then sharded with a `PartitionSpec` resolved against the **ambient** Flax rules — which at eval are `logical_axis_rules_for_eval`. With `custom_mesh_and_rule=ep-as-dp` + `custom_mesh_and_rule_for_eval=ep-as-cp`:

- `ep-as-dp` maps `activation_length` to `[]`, so `infer_cp_axes` returns nothing and `context_sharding` keeps its default `"context"` — an axis that isn't even in the ep-as-dp mesh — giving `cp_size = 1`.
- `ep-as-cp` maps `activation_q_length` to `['expert']`, so the kernel is handed to a 4-way `shard_map`.

The mask metadata is sized for one shard, so it can't be split:

```
ValueError: shard_map applied to the function 'wrap_flash_attention' was given
argument arrays with axis sizes that are not evenly divisible by the
corresponding mesh axis sizes: ... 4 does not evenly divide 10
```

## Fix 1 — resolve `cp_size` against the ambient rules

Resolve `activation_q_length` against the ambient rules — the same source that produces the `PartitionSpec` — so the shard count and the spec can no longer disagree. This fixes the class of bug, not just this rule pair.

Why not split the flag into `context_sharding` / `context_sharding_for_eval`? The consumer can't act on it: `eval_step` calls the same `loss_fn` on the same model object with only `is_train=False`, so eval runs as `MODEL_MODE_TRAIN` and `AttentionOp` has nothing to branch on. A second field would need an `is_train` bool plumbed Decoder → DecoderLayer → Attention → AttentionOp, and would still leave the invariant enforced only by convention. `context_sharding` is also not independent information — it is a cached projection of the rules.

`config.context_sharding` keeps its role as the CP **axis name** (`ring_axis`, `ulysses_axis`, the ring/USP validators). It just stops being used as a **count**.

The Ulysses axis is excluded from the product to preserve today's semantics: under USP the query sequence is split over both the ring and Ulysses axes, but `cp_size` refers to the ring extent alone. Without that exclusion USP would silently change from 2 to 4 (see the parity table below).

Also adds a warning when an eval rule declares mesh axes the mesh doesn't have — `ep-as-cp`'s `stage` axis is currently dropped in complete silence, which is only harmless while that axis would have been size 1.

## Fix 2 — load balance only when the input pipeline reordered for this `cp_size`

*(added after review — thanks @huytransformer for catching it)*

Fix 1 alone trades a crash for a wrong number. Once `cp_size` is 4 at eval, `use_load_balanced_cp` also flips on (`context_parallel_load_balance` defaults to `true`) — over a batch that was never reordered.

The load-balanced path is not a perf toggle. It is a contract with the loader:

- `LoadBalancedCausalMask` bakes the DUAL_CHUNK_SWAP permutation into a static `q_sequence`, so row *i* is masked as if it held the token from position `q_sequence[i]`.
- `wrap_flash_attention` calls `reorder_sequence(..., to_contiguous=True)` on K/V and the segment IDs to undo that permutation.

Applied to an unpermuted batch, neither raises. At `max_target_length=4096`, `cp=4` the chunk permutation is `[0,7,1,6,2,5,3,4]`:

| physical row | true position | mask allows kv ≤ | |
|---|---|---|---|
| 512 | 512 | 3584 | **+3072 future tokens visible** |
| 1023 | 1023 | 4095 | sees the entire sequence |
| 2048 | 2048 | 1024 | **−1024: can't see its own past** |

Half the rows leak the future, half are over-masked, and K/V get scrambled on top. Eval loss comes out implausibly low, with no exception and no warning.

The permutation is fixed once, at setup, by `setup_train_loop` — for `mesh[context_sharding]` under the *train* rules, for both iterators. So `max_utils.reordered_cp_size(config, mesh)` becomes the single source of truth for what order the batch is in: the loader reorders by it, and `AttentionOp._load_balanced_context_parallel()` requires the kernel's own shard count to equal it.

**Equality, not `> 1`.** A cp=2 train mesh with a cp=4 eval mesh is as wrong as 1 vs 4. On a mismatch the plain causal path takes over — correct at any shard count — and only load balancing is lost. Note the criterion is the *count*, not the axis name: a query sharded 4 ways on `expert` while the loader reordered off a 4-way `context` is fine, because DUAL_CHUNK_SWAP assigns chunk pairs to ranks 0..3 in order and the mesh axis only decides which device is which rank.

`generate_attention_mask`'s `use_segment_positions` moves to the same predicate. That path builds the mask from the positions carried in the batch, so it is self-consistent either way; routing it through the one predicate just keeps "the LB path is live" from having two spellings.

**The cuDNN/GPU path is untouched.** It reads `mesh[config.context_sharding]` directly — the same field the loader uses — so it cannot skew.

### What this costs at eval

Under `ep-as-cp` the sequence is split 4 ways for the whole model, and MoE / dense / embedding / loss stay perfectly balanced at 1024 tokens per shard regardless of token order. Only the sequence-quadratic term of causal attention is triangular, because `activation_kv_length -> []` leaves K/V replicated while q is sharded contiguously:

| | work per shard | max shard | vs. replicating `expert` |
|---|---|---|---|
| replicated | 10, 10, 10, 10 | 10 | 1.0× |
| CP-4, no LB | 1, 2, 3, 4 | 4 | **2.5×** |
| CP-4, with LB | 2.5, 2.5, 2.5, 2.5 | 2.5 | 4× |

So the eval rule keeps 62.5% of the ideal CP speedup on the quadratic term and 100% everywhere else — against a ~4× win over replication. For DSv3 at 4k, on a forward-only step dominated by MoE, the net cost is single-digit percent. A config-time warning now says so explicitly when the two rules disagree.

**Reordering the eval batch instead is not an option here** — see Fix 3. `context_parallel_load_balance: false` is also a free no-op for the MLPerf config specifically, since train runs at cp=1 under `ep-as-dp`/`dsv3-mlperf-4k` and the flag never fires there. The code fix makes that config discipline unnecessary.

## Fix 3 — MTP's CP guards were dead

`MultiTokenPredictionBlock.__call__` derived `cp_size` from `getattr(cfg, "context_parallel_size", 1)`. **There is no `context_parallel_size` config field** — the CP extent only exists on the mesh — so it always returned the default `1` and both guards below it never fired:

- MTP + CP + load balancing (DUAL_CHUNK_SWAP breaks the shift-by-one neighbour fetch in `_shift_left_one_cp_aware`)
- MTP + CP + packing with `mtp_num_layers > 1`

Pre-existing, independent of the eval-rule skew, and the same failure mode: a static value standing in for a mesh-derived one. The load-balance guard now keys off `reordered_cp_size` — the condition it is really about — and the multi-layer guard reads the mesh.

This is also what rules out the alternative fix for Fix 2. Recovering full load balancing at eval would mean reordering the eval batch with the eval rule's CP extent, but DSv3 MLPerf runs `mtp_num_layers=1` and `eval_step` reports `evaluation/mtp_loss`. On a permuted sequence MTP's "successor" is the permuted-order neighbour — a 3072-token jump at chunk boundaries. MaxText already knows this; the guard exists precisely to reject it. It just wasn't running.

## Changes

```
 src/maxtext/configs/types.py                 | 48 ++++++++++++++++++++--
 src/maxtext/layers/attention_op.py           | 49 ++++++++++++++++++++--
 src/maxtext/layers/multi_token_prediction.py | 12 ++++--
 src/maxtext/utils/max_utils.py               | 22 ++++++++++
 src/maxtext/utils/train_utils.py             | 10 +++--
 tests/integration/train_tests.py             | 13 +++++-
 tests/unit/attention_test.py                 | 77 ++++++++++++++++++++++++----
 tests/unit/multi_token_prediction_test.py    | 51 +++++++++++++++++++++
```

- `max_utils.py` — new `reordered_cp_size(config, mesh)`: the one definition of the permutation the batch carries.
- `train_utils.py` — the loader reorders by it instead of re-deriving it.
- `attention_op.py` — new `AttentionOp._context_parallel_size()` (Fix 1) and `_load_balanced_context_parallel()` (Fix 2); the two places that read `mesh.shape[config.context_sharding]` as a count now call the former, and every load-balance decision goes through the latter.
- `multi_token_prediction.py` — Fix 3.
- `types.py` — warn on eval-rule mesh axes absent from the mesh, and on a train/eval `activation_q_length` mismatch under load balancing. `infer_cp_axes` refactored onto a generic `axes_for_logical`.

# Tests

**1. `cp_size` parity — old expression vs new, across every CP configuration.** Must be identical everywhere except the skewed pair:

| case | old | new (train) | new (eval) | |
|---|---|---|---|---|
| default (no CP) | 1 | 1 | 1 | unchanged |
| cp=2 all_gather | 2 | 2 | 2 | unchanged |
| cp=2 ring | 2 | 2 | 2 | unchanged |
| ulysses ctx=2 | 2 | 2 | 2 | unchanged |
| usp ring2 × ulysses2 | 2 | 2 | 2 | unchanged |
| ep-as-dp | 1 | 1 | 1 | unchanged |
| ep-as-cp | 4 | 4 | 4 | unchanged |
| **ep-as-dp train / ep-as-cp eval** | 1 | 1 | **4** | **fixed** |

**2. Real MLPerf config at the real topology.** `deepseek3-671b`, `tpu7x-4x8x8` (256 chips / 512 devices), `ici_expert_parallelism=4`, `ici_fsdp_parallelism=-1`, `attention=flash`, `use_tokamax_splash=true`, `max_target_length=4096`, `sa_block_*=1024`, `mtp_num_layers=1`, AOT-lowered on CPU:

| | before | after |
|---|---|---|
| eval step, `ep-as-dp` + `ep-as-cp` | `ValueError` | lowers OK |
| eval step, `ep-as-dp` only (control) | lowers OK | lowers OK |
| train step | lowers OK | lowers OK |

**3. Regression coverage in `test_fractional_eval_batch_size`.** That integration test already built this exact skew (`custom_mesh_and_rule_for_eval=ep-as-cp` over default train rules) and still passed on unfixed code, so it is updated here rather than duplicated. Two things hid the bug:

- It ran the **legacy** JAX Splash kernel (`use_tokamax_splash` defaults to false), which represents a causal mask as a dense `n x n` block grid. That divides evenly by any shard count, so the skew was harmless. Tokamax — what MLPerf actually runs — stores the compressed list of `n(n+1)/2` active blocks instead, which does not. The legacy path also routes its spec through `manual_sharding_spec`, which validates divisibility; the Tokamax path applies the spec unchecked.
- The block size mattered too. At the default `sa_block_q=512`, `4096/512 = 8` blocks give 36 active — not divisible by 8, but divisible by 4. Since `tpu7x-tests` carries `github.event_name != 'pull_request'`, only the 4-device `v6e-4` runner gates PRs, so a failure that needs 8 devices would never gate anything. The test now pins `ici_expert_parallelism=4` (`fsdp=-1` takes the remainder) and `sa_block_*=1024`: 4 blocks, 10 active, and `4 does not evenly divide 10` on either runner. On `v6e-4` the pinned values are what `-1` already resolved to.

Verified on 8 devices — before the fix it fails with the production signature, after it passes:

```
* the passed args[6][<flat index 0>].active_rows of shape int8[10] ...
  corresponds to in_specs[6] of value P('expert',), which maps array axis 0
  (of size 10) to mesh axis 'expert' (of size 4), but 4 does not evenly divide 10
```

**4. Mask-dispatch parity for Fix 2** — new `test_tpu_splash_load_balances_only_when_the_loader_reordered`. Drives `tpu_flash_attention` far enough to capture the mask it built:

| q axis (eval rules) | mesh | loader reordered for | expected mask | pre-fix |
|---|---|---|---|---|
| `context` | `context=4` | 4 | `LoadBalancedCausalMask` | same |
| `expert` | `context=1, expert=4` | 1 | `CausalMask` | **`LoadBalancedCausalMask`** ❌ |
| `expert` | `context=2, expert=4` | 2 | `CausalMask` | **`LoadBalancedCausalMask`** ❌ |
| `expert` | `context=4, expert=4` | 4 | `LoadBalancedCausalMask` | same |

Plus `test_tpu_splash_skips_load_balancing_when_flag_is_off`. Confirmed both skewed rows fail on the pre-fix predicate and the two aligned rows are untouched.

**5. MTP guard coverage for Fix 3** — new `test_cp_guards_fire_on_a_cp_mesh`, one subtest per guard. Both fail with the old `getattr(cfg, "context_parallel_size", 1)` (no raise at all) and pass now.

**6. 8-device end-to-end run** (v7-8, `deepseek3-tiny`, 4 layers, ep=4 × fsdp=2, same rules and kernel flags): eval raised before Fix 1, now completes with `EXIT=0`.

**7. Config-time warnings** — verified by hand that `ep-as-dp` + `ep-as-cp` emits both the dropped-`stage`-axis warning and the new `activation_q_length` mismatch warning, and that `ep-as-dp` + `ep-as-dp` emits neither.

**8. Existing suites.**

- TPU: `attention_test.py`, `context_parallelism_test.py`, `ulysses_attention_test.py`, `tokamax_ring_attention_test.py` — **79 passed, 95 skipped, 0 failed**. (The two `attention_test.py` failures noted in an earlier revision of this description reproduced on unmodified `main` and no longer appear at the current base.)
- CPU: same four suites — 66 passed, 85 skipped. `multi_token_prediction_test.py` — 26 passed, 9 skipped. `configs_value_test.py` — 36 passed, 110 subtests passed.
- `pyink` (`--pyink-indentation=2 --line-length=122`) clean; `pylint -E` reports no new findings versus base (the one extra hit is another instance of the pre-existing pydantic `FieldInfo` false positive, from repeating `self.custom_mesh_and_rule.value`).

**Note on the reported symptom.** This was investigated from a report of the 256-chip job *hanging* at eval. Locally and in AOT the same trigger produces a deterministic trace-time `ValueError` rather than a hang; the likely explanation is one host raising while the other 63 wait in a collective. Worth grepping a failing host's stderr for `does not evenly divide` to confirm they are the same defect.

# Checklist


src/maxtext/layers/attention_op.py
```diff
@@ -1649,7 +1676,7 @@ def tpu_flash_attention(
     use_tokamax_ring = tokamax_ring_attention.is_context_parallel_ring_requested(self.config)
     use_ulysses = ulysses_attention.is_context_parallel_ulysses_requested(self.config)
     use_usp = usp_attention.is_context_parallel_usp_requested(self.config)
-    cp_size = self.mesh.shape.get(self.config.context_sharding, 1)
+    cp_size = self._context_parallel_size()
```
huytransformer, line 1696: Shouldn't we turn on load balancing in the kernel only when data loader actually shuffled? seems LB still follows the config flag while cp_size is now dynamic?

huytransformer (approved):

src/maxtext/layers/attention_op.py
NuojCheng: CP rank is still static since they only depends on values from config. I don't think it gonna impact LB

huytransformer: by "dynamic" I just meant cp_size is 1 at train but 4 at eval. the loader uses the train value and never shuffles, but the kernel uses the eval value?

NuojCheng: that's a good point. Performance it should be fine without using LB for eval, since they take relatively small portion and only happens every X steps. I am not sure about the correctness though, need to take a deeper look.

Shuwen-Fang (approved):

PR #5020, opened 2026-08-26
Add fsdp-as-dp-for-attn custom mesh and rule

NuojCheng:
# Description

Follow-up to #5007, which gave the attention weights their own `embed_attn` logical name. This PR adds a new custom mesh and rule, `fsdp-as-dp-for-attn`, that uses the new name to shard the attention weights on the **expert** physical axis:

```
FSDP acts like DP for attn, but FSDP for moe
EP acts like FSDP for attn, and EP for moe
```

The rule is a copy of `ep-as-dp.yml` — same mesh (`data`, `fsdp`, `fsdp_transpose`, `expert`), same expert-acts-as-data-parallelism philosophy outside the core MoE part — with one change, the Attention Weights section:

```yaml
# ep-as-dp.yml                  # fsdp-as-dp-for-attn.yml
['q_lora',      ['fsdp']],      ['q_lora',      ['expert']],
['kv_lora',     ['fsdp']],      ['kv_lora',     ['expert']],
['embed_attn',  ['fsdp',        ['embed_attn',  ['expert']],
                 'fsdp_transpose']],
```

All attention weights are therefore sharded on `expert` and replicated over `fsdp` / `fsdp_transpose`, so attention pays no FSDP all-gather at all. Because the expert axis carries the batch outside of MoE, sharding attention weights on it behaves like FSDP over a much smaller and faster axis — it trades HBM for a cheaper collective. MoE, dense MLP and vocab weights are untouched and keep using FSDP.

**Why this matters at large scale.** The attention weights are small, so once the FSDP rank is large (larger than 512) each shard is tiny and its all-gather is dominated by latency rather than bandwidth — the attention layer becomes latency bound and the all-gather stops paying for itself. Sharding those weights on the (much smaller) expert axis keeps the per-shard slice large enough that the collective stays bandwidth bound. MoE weights are large enough to keep scaling with FSDP, so they are left alone. DeepSeek V3 MLPerf training at 4k sequence length is one such case.

Also in this PR:

- `CustomRule.FSDP_AS_DP_FOR_ATTN` in `common/common_types.py`.

# Tests

- `python3 -m tests.utils.sharding_dump ... custom_mesh_and_rule=fsdp-as-dp-for-attn` on `deepseek2-16b` / `tpu7x-16` (`ici_fsdp_parallelism=-1 ici_expert_parallelism=2 use_ring_of_experts=true`). Diffing the resulting `named_shardings.json` against the `ep-as-dp` golden for the same model/topology: 24 of 84 entries changed, and all 24 are attention weights (`query`, `wkv_a`, `wkv_b`, `out` and their `mu` / `nu` optimizer states), going from `('fsdp', 'fsdp_transpose')` / `'fsdp'` to `'expert'`. Every MoE, dense-MLP, vocab and norm entry is bit-identical.
- `deepseek2-16b` has `q_lora_rank=0`, so the `q_lora` rule is not exercised by the dump. Checked the DeepSeek V3 kernel axis tuples directly with `flax.linen.spmd.logical_to_mesh_axes`: `wq_a ('embed_attn', 'q_lora_up_proj') -> P('expert', None)`, `wq_b ('q_lora', 'q_heads', 'kv') -> P('expert', None, None)`, `wkv_a -> P('expert', None)`, `wkv_b -> P('expert', None, None)`, `out ('heads', 'kv', 'embed_attn') -> P(None, None, 'expert')`, while `moe w0/wo`, `dense wi` and `vocab` match `ep-as-dp` exactly.
- `tests/unit/sharding_test.py`, `tests/unit/configs_value_test.py`, `tests/unit/pyconfig_test.py`: 73 passed, 1 skipped, 115 subtests passed — including `test_embed_attn_rule_accompanies_embed_rule` from #5007, which covers the new rule file.

Follow-up to #5007, now merged; this PR is rebased on top of it and targets `main`.

FIXES: b/527054211

# Checklist


src/maxtext/configs/custom_mesh_and_rule/fsdp-as-dp-for-attn.yml
```diff
@@ -0,0 +1,97 @@
...
+                      # Note activation batch and length also get used in vocab
+                      ['activation_batch', ['data', 'fsdp', 'fsdp_transpose', 'expert']],
+                      ['activation_length', []],
+                      ['activation_norm_length', []],
+                      ['activation_embed', []],
+                      ['activation_stage', []],
+                      # General Weights
+                      ['mlp', ['fsdp_transpose']],
+                      ['embed', ['fsdp', 'fsdp_transpose']],
+                      ['embed', ['fsdp']],
```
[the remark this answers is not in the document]
NuojCheng: ignore

src/maxtext/common/common_types.py
```diff
@@ -172,3 +172,4 @@ class CustomRule(enum.Enum):
   FSDP_2D = "2d-fsdp"
   EP_AS_DP = "ep-as-dp"
   SHARD_EXP_ON_FSDP = "shard-exp-on-fsdp"
+  DSV3_MLPERF_4K = "dsv3-mlperf-4k"  # ep-as-dp with attention weights sharded on expert
```
gobbleturk: This rule will be used a lot for scaling moe models, I would not include DSV3 in the name - how about fsdp-as-dp-for-attn

gobbleturk: for the comment I would be explicit:

FSDP acts like DP for attn, but FSDP for moe
EP acts like FSDP for attn, and EP for moe

gobbleturk: just a reminder we briefly discussed orthogonalizing these custom rules into axis specific - not to be addressed in this PR but perhaps we could have something like

custom_fsdp = dp_attn_fsdp_moe
custom_ep = fsdp_attn_ep_moe (this is actually default probably, not custom)

NuojCheng: Done — renamed to `fsdp-as-dp-for-attn` (enum `CustomRule.FSDP_AS_DP_FOR_ATTN`, file `fsdp-as-dp-for-attn.yml`). No DSV3 in the name; DSV3 at 4k is now just mentioned as an example workload in the file header.

NuojCheng: Done. Used your wording verbatim in both places. `common_types.py`:

```python
# FSDP acts like DP for attn, but FSDP for moe; EP acts like FSDP for attn, and EP for moe
FSDP_AS_DP_FOR_ATTN = "fsdp-as-dp-for-attn"
```

and the yml header now leads with:

```
# This rule uses data, FSDP, FSDP_transpose and expert, and assigns them as:
#
#   FSDP acts like DP for attn, but FSDP for moe
#   EP acts like FSDP for attn, and EP for moe
```

followed by the rationale (expert carries the batch outside the core MoE part, so attn weights sharded on it get FSDP-like behavior over a smaller/faster axis — trading HBM for a cheaper all-gather; helps when the attn projections are small enough to be latency bound, e.g. DSV3 at 4k).

NuojCheng: Ack — leaving the axis-specific orthogonalization (`custom_fsdp` / `custom_ep` style) out of this PR as you suggested. The new name lines up with that direction: this rule is exactly the `dp_attn_fsdp_moe` case, so it should fold in cleanly when we do split the knobs.

NuojCheng: Expanded the header to spell out when this rule is the right choice — it is aimed at large scale training:

> This is mainly useful for large scale training. The attention weights are small, so once the FSDP rank is large (larger than 512) each shard is tiny and their all-gather is dominated by latency rather than bandwidth, which makes the attention layer latency bound. Sharding the attention weights on the (much smaller) expert axis keeps the per-shard slice large enough that the collective stays bandwidth bound, at the cost of replicating those weights over fsdp / fsdp_transpose. MoE weights are large enough to keep scaling with FSDP, so they are left alone. DeepSeek V3 MLPerf training at 4k sequence length is one such case.

Same rationale added to the PR description.

gobbleturk (approved):

Shuwen-Fang (approved):

PR #5042, opened 2026-08-28
Fix dropped routing-weight gradient in ring_ragged_unsort

guowei-dev:
# Description

`_ring_ragged_unsort_bwd` returned `None` for `topk_weights_flat`:

```python
return grad_sorted_tokens, None, None, None
```

Of the four primals `(sorted_tokens_local, group_sizes_local, topk_argsort_revert_indices, topk_weights_flat)` the first three are correctly `None` — two are int32 index arrays — but the fourth is the router's top-k combine weight, which is differentiable. `jax.custom_vjp` accepts `None` there and silently substitutes a zero cotangent, so with `use_ragged_sort` + `use_ring_of_experts` the router received no gradient from the main loss and trained only on the auxiliary load-balance loss. The forward was unaffected.

This computes the missing term, `g_w[i] = <g_out[i // topk], sorted_tokens[revert[i]]>` for valid `i`, in both buffering branches. The backward's `ragged_gather` drops its fused per-slot weights so one gather feeds both gradients — the activation gradient after scaling, the weight gradient after a row-wise dot — and the forward residual gains `sorted_tokens_local`, which the `use_ragged_sort=False` path already retains for its `weight_sum` einsum.

The no-ring path combines with `jnp.einsum` and is unaffected.

FIXES: b/553024870

# Tests

`_run_ragged_sort_loss_and_grad` already compared the full gradient pytree across the flag, but under a fixed `atol=8e-2` while the loss is `jnp.mean(out**2)`, so every leaf is O(1e-5). On its own config (mixtral-8x7b, emb 7168, EP=2, bf16, v7x):

```
['gate']['kernel'] (7168, 8) | max|g_ref| 4.4373e-05  max|g_rs| 0.0000e+00 | allclose: PASS
```

An identically zero gradient passes because the tolerance is ~1800x the whole gradient; the hidden-state assertion above it is vacuous the same way (that gradient peaks at 2.85e-08). Scaling both tolerances to the magnitude being compared makes the existing test fix-sensitive without adding a new one.

| `tests/unit/moe_test.py` (TPU v7x) | result |
| --- | --- |
| `-k ragged_sort`, tolerances only, production file at upstream | 4 failed (all ring-of-experts variants), 4 passed |
| `-k ragged_sort`, this branch | 8 passed |
| `-k no_ring_of_experts`, without the fix | 4 passed — unaffected path, so the bound is not indiscriminate |
| full file, this branch | 18 passed, 57 skipped (cpu_only / AOT / 4-device), 0 failed |

Post-fix `max|diff| / max|g_ref|`: gate 9.2e-04 (was 1.000), `wi_0` 1.03e-2, `wi_1` 8.7e-3, `wo` 4.5e-3, hidden state 8.2e-3 — all inside the 8e-2 bound. Forward loss unchanged to 6 decimal places.

Also checked off-TPU, where both kernels take their JAX fallback: over 6 configurations covering both buffering branches, `<g_w, d>` is 0.0 before and matches a central-difference directional derivative to ≤3.7e-3 after. The `use_ragged_sort=False` arm agrees analytic-vs-finite-difference to ~1e-6, so the probe itself is sound.

`pre-commit` (codespell, pylint, pyink) passes on both changed files.

# Checklist

- [x] I have run end-to-end tests and provided workload links above if applicable. Unit and gradient tests only — the convergence job in b/553024870 (43 vs 49 steps) should be rerun with this patch to confirm how much of the gap it accounts for.

### DIFF
--- a/src/maxtext/kernels/ragged/ragged_sort.py
+++ b/src/maxtext/kernels/ragged/ragged_sort.py
@@ -253,7 +253,8 @@ def ring_ragged_unsort(
     during the reduction.
 
   Backward:
-    ``g_sorted_tokens[j] = w[i] * g_out[i // topk]`` where
+    ``g_sorted_tokens[j] = w[i] * g_out[i // topk]`` and
+    ``g_w[i] = <g_out[i // topk], sorted_tokens_local[j]>`` where
     ``j = topk_argsort_revert_indices[i]`` and ``j`` is in
     ``[shard_output_start, shard_output_end)``.
 
@@ -265,6 +266,7 @@ def ring_ragged_unsort(
     local_num_experts: scalar ``int`` representing the count of experts hosted on this shard.
     ep_name: ``str`` identifying the expert parallel axis name.
     topk_weights: ``[num_tokens_local * topk]`` tensor of per-slot routing weights.
+      Differentiated: its gradient is what trains the router.
 
   Returns:
     A 2D ``[num_tokens_local, hidden]`` tensor with expert outputs scattered back
@@ -351,6 +353,7 @@ def ring_ragged_unsort(
       )
 
     res = (
+        sorted_tokens_local,
         topk_argsort_revert_indices,
         topk_weights_flat,
         shard_output_start,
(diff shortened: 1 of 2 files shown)


NuojCheng (approved):
Thank you for the fix! LGTM

Shuwen-Fang (approved):

PR #5053, opened 2026-08-28
feat(layers): implement FP8 weight-only dynamic dequantization engine and types

snehalv2002:
# Description

Implements native FP8 weight-only storage with dynamic on-the-fly dequantization in MaxText's core linear and MoE layers (`DenseGeneral`, `MlpBlock`, and `RoutedMoE`).

### Motivation & Context
Open-source frontier models (e.g. Llama 3.1 8B FP8 with scalar scaling and Qwen 3.5 35B A3B FP8 with 128-block scaling) are widely distributed in quantized FP8 formats (`float8_e4m3fn` / `float8_e5m2`). Storing these checkpoints in pure BF16 requires excessive High-Bandwidth Memory (HBM), limiting rapid model bringup, inference, and logit verification on smaller TPU topologies (e.g. v4-64, v5e-8). Furthermore, TPUs prior to TPU v7 (v4, v5e, v5p) do not support native hardware FP8 matrix multiplications.

Dynamic dequantization allows weights to be stored and moved across HBM in 8-bit precision, dynamically upcasting in vector registers immediately before `lax.dot_general`, achieving a ~50% reduction in parameter memory footprint and memory-bandwidth traffic.

### Key Changes
1. **Precision Types & Configuration (`common_types.py`, `configs/types.py`, `configs/base.yml`):**
   - Registered `float8_e4m3fn` and `float8_e5m2` in the configuration schema `DType` and added `is_fp8_dtype()` helper.
   - Added `unquantized_modules` and `weight_block_size` to `Quantization` config and `base.yml`, enabling models to configure block scaling (symmetric e.g. 128x128 or asymmetric e.g. [128, 64]) and keep sensitive submodules in full precision.
   - Added `get_weight_dtype(config, module_name)` to dynamically resolve layer precision based on pattern-matched `unquantized_modules`.
2. **Quantization Configuration & Operator (`quantizations.py`):**
   - Implemented `dequantize_weight(w, scale, compute_dtype)` supporting per-tensor scalar scaling, block scaling grids (2D/3D), and general broadcastable scale shapes.
   - Added `WeightQuantConfig` dataclass and `get_weight_quant_config(config, module_name)` helper to decouple weight quantization settings from general model hyperparameters.
3. **Linear Layers & Initializers (`linears.py`, `attentions.py`, `initializers.py`):**
   - Updated `DenseGeneral` and Linen bridge `dense_general` to manage `kernel` in FP8 alongside a companion `kernel_scale` parameter, deriving partition spec shardings to match logical mesh annotations.
   - Re-exported `dequantize_weight` and `WeightQuantConfig` from `quantizations.py` for backward compatibility.
   - Updated `_compute_dot_general` / `_compute_dot_general_nnx` to dynamically upcast weights immediately before `lax.dot_general`.
   - Ensured non-FP8 weights are cast to `compute_dtype` prior to slicing and all-gather dispatch (matching unquantized AQT expectations), while keeping FP8 weights quantized until fused dequantization.
   - Added unwrapping for wrapped parameter state containers (`get_value()` / `dict`), and dequantize before slicing when `slice_bounds` is provided.
   - Wired `weight_quant` and `weight_block_size` through `Attention` projections and `MlpBlock` (`wi`, `wi_0`, `wi_1`, `wo`).
   - Updated `_default_embed_init` and `nd_dense_init` in `initializers.py` to sample distributions in `jnp.float32` and cast to target dtype when target is FP8, avoiding JAX sampling errors on 8-bit floating point dtypes.
4. **MoE Layer Support (`moe.py`):**
   - Added support for FP8 weights and companion scale parameters in `RoutedMoE` for both prefused (`wi_scale`, `wo_scale`) and unfused (`wi_0_scale`, `wi_1_scale`, `wo_scale`) layouts.
   - Supported 1D per-expert scale grids with 1D sharding `(wi_kernel_axes[0],)` to prevent rank-mismatch issues with 3D kernel axes, as well as 3D block-scaling grids inheriting full kernel axes.
   - Dynamically dequantizes expert weights during the forward pass across all execution paths (fused RPA, prefused, and unfused).
   - Routed expert gates resolve precision via `get_weight_dtype(config, "gate")` to keep routing in high precision when declared in `unquantized_modules`.
5. **Precision Isolation & Pipeline Stage Guards (`embeddings.py`, `normalizations.py`, `nnx_decoders.py`, `nnx_scan.py`):**
   - Preserved sensitive layers (token embeddings via `Embed` and untied LM heads via `logits_dense`) in `bfloat16`/`float32` when declared in `unquantized_modules`.
   - Guaranteed RMSNorm scale parameters remain in `bfloat16`/`float32`.
   - Added dimension guards (`if x.ndim > scan_axis`) in `nnx_scan.py` and `nnx_decoders.py` to prevent crashes when moving scan axes on 0D/1D parameter leaves (e.g. scales or 1D norms).

Part 1 of 5 in the FP8 Weight-Only Dynamic Dequantization series.

# Tests

Ran unit test suites covering dynamic dequantization, MoE scale initialization and execution, and decoder layer precision isolation:
```bash
pytest tests/unit/linears_test.py tests/unit/moe_test.py tests/unit/nnx_decoders_test.py -v
```

New unit tests added:
- `tests/unit/linears_test.py`:
  - `test_quantized_kernel_cast_to_compute_dtype` & `test_fp8_kernel_not_cast_before_dispatch`
  - `test_wrapped_kernel_state_does_not_raise_ellipsis_keyerror`
  - `test_dequantize_weight_scalar`, `test_dequantize_weight_per_channel`, `test_dequantize_weight_block_wise`, `test_dequantize_weight_block_divisibility`
  - `test_fp8_e4m3fn_dense_general`, `test_fp8_e5m2_dense_general`, `test_fp8_block_wise_scale`, `test_kernel_scale_sharding_inference`, `test_fp8_slice_bounds`
  - `test_weight_quant_config_scalar`, `test_weight_quant_config_block`, `test_get_weight_quant_config_helper`
- `tests/unit/moe_test.py`:
  - `test_fp8_prefused_per_expert_scale_init`
  - `test_fp8_prefused_block_scale_init`
  - `test_fp8_unfused_scale_init`
  - `test_fp8_forward_pass_dequantization`
  - `test_fp8_moe_with_weight_quant_config`
- `tests/unit/nnx_decoders_test.py`:
  - `test_fp8_weights_and_unquantized_layers`
  - `test_fp8_forward_pass_execution`

# Checklist


src/maxtext/layers/linears.py
```diff
@@ -175,34 +253,83 @@ def __init__(
...
+              rngs.params(),
+              resolved_scale_shape,
+              self.scale_dtype,
+          ),
+          sharding=resolved_scale_axes,
+      )
+    else:
+      self.scale_axes = None
+      self.kernel_scale = None
+
```
snehalv2002, line 344: a lot of casework here. can you add a comment to clear this up?

snehalv2002: Added detailed comments structuring the resolution into two distinct phases (Phase 1: Scale shape resolution across explicit overrides, block scaling grids, and per-tensor scalar scales; Phase 2: Scale mesh sharding resolution). Also de-duplicated the casework by normalizing `block_size` into a per-dimension tuple upfront.

src/maxtext/configs/base.yml
```diff
@@ -183,6 +172,10 @@ decoder_block: "llama2" # which style of decoderblock to use.
 # then you should explicitly set base_embed_dim, base_num_query_heads, base_num_kv_heads,
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
 weight_dtype: "float32"
+# List of module names or patterns to keep unquantized in model dtype (e.g. bfloat16) even when weight_dtype is FP8
+unquantized_modules: []
+# Block size for block-scaled quantized weights (e.g. 128 for 128x128 block scaling). null for per-tensor scaling.
+weight_block_size: null
```
snehalv2002, line 194: is it better to use 0 or null

Shuwen-Fang (commented):
Do you have test showing correctness of logits with a few full steps?

src/maxtext/configs/base.yml
Shuwen-Fang, line 194: are blocks always symmetric for block scaling?

src/maxtext/configs/base.yml
```diff
@@ -183,6 +172,10 @@ decoder_block: "llama2" # which style of decoderblock to use.
 # then you should explicitly set base_embed_dim, base_num_query_heads, base_num_kv_heads,
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
 weight_dtype: "float32"
+# List of module names or patterns to keep unquantized in model dtype (e.g. bfloat16) even when weight_dtype is FP8
+unquantized_modules: []
```
Shuwen-Fang, line 191: Can you list all accepted module names here?

Shuwen-Fang: can you also add to the comment that the weight precision will be specified by "dtype" if it's included here?

src/maxtext/layers/embeddings.py
```diff
@@ -126,12 +126,12 @@ def __init__(
     self.cast_input_dtype = cast_input_dtype
     self.dtype = dtype
     self.attend_dtype = attend_dtype
-
```
Shuwen-Fang, line 82: nit but wonder if it could make sense to keep an enum of all modules instead of strings

src/maxtext/configs/base.yml
```diff
@@ -183,6 +172,14 @@ decoder_block: "llama2" # which style of decoderblock to use.
 # then you should explicitly set base_embed_dim, base_num_query_heads, base_num_kv_heads,
 # base_mlp_dim, base_num_decoder_layers and/or head_dim.
 weight_dtype: "float32"
+# List of module names or patterns to keep unquantized even when weight_dtype is FP8.
```
RissyRan, line 187: Shall we indicate which part is quantized using FP8?  So empty list means all in bf16? How is ratio from recent checkpoints?

src/maxtext/configs/base.yml
```diff
@@ -183,6 +172,14 @@ decoder_block: "llama2" # which style of decoderblock to use.
...
+# For any module included here, its weights will be instantiated using `dtype` (e.g. bfloat16).
+# Accepted module names include: 'embed_tokens' (or 'token_embedder'), 'lm_head' (or 'logits_dense'),
+# 'gate' (MoE router gate), 'shared_expert_gate', 'conv1d', 'in_proj_ba', 'norm'.
+unquantized_modules: []
+# Block size for block-scaled quantized weights (e.g. 128 for symmetric 128x128 block scaling,
+# or a list/tuple like [128, 64] for asymmetric block scaling). Set to null for per-tensor scaling.
+weight_block_size: null
```
RissyRan, line 194: Shall we keep this as "[]" or "[-1, -1]"? to indicate block wise 128x128 or sub-channel, like 1x128.

src/maxtext/layers/linears.py
```diff
@@ -74,14 +74,70 @@ def canonicalize_tuple(x):
     return (x,)
 
 
-def _compute_dot_general(inputs, kernel, kernel_axes, axis, contract_ind, matmul_precision, quant):
+def dequantize_weight(
```
RissyRan: Shall we move this to src/maxtext/layers/quantizations.py?

src/maxtext/layers/moe.py
```diff
@@ -678,6 +668,61 @@ def __init__(
     else:
       self.per_expert_scale = None
 
+    if not quantizations.in_serve_mode(self.quant) and ctypes.is_fp8_dtype(self.weight_dtype):
```
RissyRan, line 790: Could you help add necessary comments for readability?

RissyRan, line 790: I didn't see related tests for this part, is this expected?

RissyRan (commented):
Thanks for the change! A few minor comments.

src/maxtext/configs/base.yml
snehalv2002: Not necessarily but the current setup allows asymmetric block sizes.

src/maxtext/configs/base.yml
snehalv2002: done

src/maxtext/configs/base.yml
snehalv2002: I noticed in the config a couple other parameters also specify null to mean disabled (e.g. quantization: null, weight_sparsity_n: null). Do you prefer  "[]" or "[-1, -1]"?

src/maxtext/configs/base.yml
snehalv2002: I used `unquantized_modules` because the huggingface checkpoints for many models like Qwen and GLM specify `modules_not_to_convert` in their `config.json`. Kimi also provides an "ignore" section under the quantization config for modules to ignore.

src/maxtext/layers/moe.py
snehalv2002: Added:
`test_fp8_prefused_per_expert_scale_init`: Verifies 1D per-expert scale initialization, shape (num_experts,), float32 dtype, and 1D sharding (model.wi_kernel_axes[0],).
`test_fp8_prefused_block_scale_init`: Verifies 3D block-grid scale initialization, shape (num_experts, in_blocks, out_blocks), and 3D sharding model.wi_kernel_axes.
`test_fp8_unfused_scale_init`: Verifies wi_0_scale, wi_1_scale, wo_scale initialization and 1D sharding.
`test_fp8_forward_pass_dequantization`: Verifies dynamic dequantization during an MoE forward pass produces valid, finite bfloat16 outputs.

snehalv2002: added comment explanation

src/maxtext/layers/linears.py
snehalv2002: moved 👍

src/maxtext/layers/embeddings.py
```diff
@@ -126,12 +126,12 @@ def __init__(
...
+    embed_weight_dtype = get_weight_dtype(self.config, "token_embedder")
```
Shuwen-Fang, line 82: nit but wonder if it could make sense to keep an enum of all modules instead of strings

Shuwen-Fang (approved):

src/maxtext/layers/embeddings.py
snehalv2002: @RissyRan do you have any preference?

src/maxtext/layers/linears.py
```diff
@@ -132,6 +165,12 @@ def __init__(
       mesh: Mesh | None = None,
       use_two_stage_all_gather: bool = False,
       debug_sharding: bool = False,
+      has_scale: bool | None = None,
+      kernel_scale_init: Initializer | None = None,
+      scale_shape: Shape | None = None,
+      scale_axes: tuple[None | str, ...] | None = None,
+      scale_dtype: DType = jnp.float32,
+      block_size: int | tuple[int, ...] | None = None,
```
shuningjin, line 174: Perhaps can decouple quantization config? e.g.,
```
@dataclasses.dataclass(frozen=True)
class WeightQuantConfig:
  quant_type: str = "none"
  weight_dtype: DType = jnp.float8_e4m3fn
  scale_dtype: DType = jnp.float32
  block_size: int | tuple[int, ...] | None = None
```

snehalv2002: Yeah seems cleaner since it already has so many parameters. I decoupled into a `WeightQuantConfig` class as you described.

src/maxtext/configs/base.yml
RissyRan: Yeah, also have a [] if you check base.yml. I think it will be better to use [] as you are expecting a list?

RissyRan (approved):

PR #5064, opened 2026-08-31
Onboard gpt3 and gpt-oss to explicit sharding

NuojCheng:
## What

Onboards the **gpt3** and **gpt-oss** decoder layers to `shard_mode=explicit`, and closes the two shared gaps that blocked them: biased `DenseGeneral` layers and YaRN rope under tensor parallelism.

## Why

`shard_mode=explicit` was already supported for `simple`, `simple_mlp`, `llama2`, `deepseek`, the Qwen3 family, Mistral/Mixtral and the Gemma family, but `decoder_block=gpt3` / `gpt_oss` were rejected by config validation. Explicit sharding is also the gate for the efficient ZeRO-1 + gradient accumulation path (`maxtext.utils.gradient_accumulation` only applies the `reduced` / `unreduced` `PartitionSpec` labels when `shard_mode == ShardMode.EXPLICIT`), so these two families were cut off from it.

## Changes

**`src/maxtext/models/gpt3.py`, `src/maxtext/models/gpt_oss.py`** — the same transformation llama2/deepseek/qwen3/mistral/gemma already use:

- `nn.with_logical_constraint(x, names)` → `self._maybe_shard_with_logical(x, names)`, a `functools.partial` of `maybe_shard_with_logical` bound to the mesh and `config.shard_mode`. It is a no-op under `ShardMode.AUTO` and emits a reshard under `ShardMode.EXPLICIT`.
- Activation `NamedSharding`s are built once in `__init__` via `create_sharding(..., rules=get_logical_axis_rules())` and threaded into the sublayers as `out_sharding=` (norms, attention, `MlpBlock`, `RoutedMoE`) and `intermediate_sharding=` (the gpt3 `MlpBlock`), so the sublayers pin their output layout instead of leaving it to GSPMD.
- `shard_mode` is threaded into `Gpt3LayerNorm` (which gains the parameter) and both gpt-oss `RMSNorm`s.

The one structural difference is gpt3's fused QKV projection. Unlike the fused path in `layers/attentions.py`, which packs q, k and v into a single flat `heads` axis and needs a `shard_map` to split them apart, gpt3 emits `(batch, length, 3, heads, head_dim)` and slices the size-3 axis. That axis is never sharded, so its `out_sharding` maps it to `None` and the existing slice stays valid under any tensor parallelism.

**`src/maxtext/layers/linears.py`** — `DenseGeneral` builds its bias in the *parameter* layout, e.g. `embed_attn -> fsdp` for an out projection, while the activation it is added to carries `activation_embed -> tensor` on that same axis. Adding them directly puts one mesh axis on two axes of the result and JAX rejects it:

```
ShardingTypeError: add operation with inputs: bf16[48@fsdp,256,256], bf16[1,1,256@fsdp]
produces an illegally sharded result: bf16[48@fsdp,256,256@fsdp]
```

The bias is now resharded onto the layout of the output before the add via `maybe_shard_with_pspec`, under `ShardMode.EXPLICIT` only (GSPMD inserts the same all-gather on its own in auto mode, and the reshard is a no-op whenever the two already agree). gpt3 is the first explicit-sharding decoder with biased projections, which is why nothing hit this before.

**`src/maxtext/layers/embeddings.py`** — `YarnRotaryEmbedding` labelled the trailing axis of its gathered frequencies `q_heads`, but that axis is `head_dim // 2`, not heads. Under any tensor parallelism the frequencies came back sharded on `tensor` and the broadcast onto the activations failed:

```
ValueError: Incompatible types for broadcasting:
input type=complex64[48,256,1,64@tensor] and requested type=complex64[48,256,8,64]
```

The frequencies are now replicated over the head dim, and the two explicit `jnp.broadcast_to(..., out_sharding=...)` calls (the complex path and the pairwise path) are deleted rather than re-targeted: once `freqs_sharding` is correct, the implicit size-1 broadcast in the elementwise op already yields the activation's own sharding, so a sharded heads axis is preserved for free.

Deleting them is also strictly more permissive. On a mesh whose sequence axis is sharded over `context_usp_ulysses` — which `activation_input_length_attn` does not list — the frequencies come back sharded on that axis while the activations are replicated on it, and `broadcast_to` cannot un-shard, so it raises where the plain multiply succeeds. A 192-configuration CPU sweep over 7 meshes x QKV / MLA-key / decode shapes x the interleave, pairwise, attention-scaling and cast variants found 12 cases that only work after the deletion, none that only work before it, and bitwise-identical values and output shardings (and identical optimized-HLO instruction counts under `ShardMode.AUTO`) everywhere else.

This also unblocks **deepseek** under `shard_mode=explicit` with tensor parallelism, which fails the same way on main today (verified below).

**`src/maxtext/configs/types.py`** — add `gpt3` and `gpt_oss` to `supported_decoders` for `shard_mode=explicit`.

**`src/maxtext/layers/nnx_decoders.py`** — the GPT3 branch of `get_norm_layer` now passes `shard_mode`, like every other branch already does.

## Tests

**`tests/integration/train_tests.py`** — `test_tpu_gpt_explicit_sharding_matches_auto` trains a tiny `gpt3-6b` and `gpt-oss-20b` for 3 steps under both shard modes and asserts the loss trajectories match. gpt3 is covered with and without the fused QKV projection, and gpt-oss under both expert parallelism (MoE dispatch) and tensor parallelism (the YaRN broadcast above).

**`tests/unit/decoder_layer_model_mode_test.py`** — `Gpt3DecoderLayer` (both QKV layouts), `GptOssDecoderLayer` and `GptOssScannableBlock` join the auto-vs-explicit forward-parity cases.

**`tests/unit/train_compile_test.py`** — AOT compilation of `gpt3-6b` (fused and unfused QKV, `v5p-8`, TP=4) and `gpt-oss-20b` (`v5p-16`, EP=8 and TP=8) under `shard_mode=explicit`.

**`tests/unit/pyconfig_test.py`** — both decoders are accepted under `shard_mode=explicit`.

### Local results (v5p-8, 4 devices)

```
[gpt3_fused_qkv] auto     losses: [145.85739135742188, 145.15592956542969, 144.79559326171875]
[gpt3_fused_qkv] explicit losses: [145.86479187011719, 145.15234375000000, 144.79580688476562]
[gpt3]           auto     losses: [145.75158691406250, 145.01895141601562, 144.65382385253906]
[gpt3]           explicit losses: [145.75024414062500, 145.02005004882812, 144.65979003906250]
[gpt_oss]        auto     losses: [8.107898712158203, 8.095435142517090, 8.088655471801758]
[gpt_oss]        explicit losses: [8.107898712158203, 8.095432281494141, 8.088606834411621]
[gpt_oss_tensor] auto     losses: [8.107820510864258, 8.095223426818848, 8.088468551635742]
[gpt_oss_tensor] explicit losses: [8.107820510864258, 8.095223426818848, 8.088468551635742]
```

gpt-oss under tensor parallelism is bit-for-bit. Under expert parallelism the grouped matmul reassociates its own reductions and drifts by 3.1e-6 relative by the third step. gpt3 is bit-for-bit at step 0 and drifts to ~5e-5 relative by step 2 in both cases, because resharding the biases reassociates the bias-gradient reductions; those two subtests use `rtol=1e-4` and the reason is documented in the test.

The four new AOT tests pass, and so do `tests/unit/decoder_layer_model_mode_test.py`, `tests/unit/pyconfig_test.py`, `tests/unit/linears_test.py` and `tests/unit/embeddings_test.py`.

For the rope fix, on main:

```
$ python3 -m maxtext.trainers.pre_train.train_compile src/maxtext/configs/base.yml \
    compile_topology=v5p-16 compile_topology_num_slices=1 model_name=deepseek3-test \
    sparse_matmul=True megablox=False attention=flash scan_layers=True \
    shard_mode=explicit ici_fsdp_parallelism=1 ici_tensor_parallelism=8
ValueError: Incompatible types for broadcasting:
input type=complex64[8,1024,1,32@tensor] and requested type=complex64[8,1024,128,32]
```

The same command compiles successfully with this change.

## Known gap

`RoutedMoE.dense_matmul` (`sparse_matmul=False`) is still not onboarded to explicit sharding — a pre-existing gap shared with mixtral and qwen3_moe — so the gpt-oss tests here use the `sparse_matmul` path.

## Checklist

- [x] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files). No doc change is needed: the set of decoders supported under `shard_mode=explicit` is not enumerated in the docs, only in the `supported_decoders` validation error message, which this PR updates.

### DIFF
--- a/src/maxtext/layers/embeddings.py
+++ b/src/maxtext/layers/embeddings.py
@@ -861,8 +861,7 @@ class YarnRotaryEmbedding(nnx.Module):
     if self.pairwise and not self.interleave:
       raise ValueError("rope_pairwise=True requires rope_interleave=True.")
 
-    # The gathered frequencies are [batch, length, half_dim]; the trailing axis is a slice of
-    # the head dimension, which is never sharded.
+    # freqs are [batch, length, half_dim]; the trailing head-dim slice is never sharded.
     self.freqs_sharding = (
         create_sharding(mesh, ("activation_batch", "activation_length", None))
         if shard_mode == ShardMode.EXPLICIT
@@ -983,10 +982,6 @@ class YarnRotaryEmbedding(nnx.Module):
         pairs = pairs.astype(jnp.float32)
         cos = jnp.real(freqs)[..., jnp.newaxis]
         sin = jnp.imag(freqs)[..., jnp.newaxis]
-        if self.shard_mode == ShardMode.EXPLICIT:
-          rotated_sharding = NamedSharding(self.mesh, jax.typeof(pairs).sharding.spec)
-          cos = jnp.broadcast_to(cos, pairs.shape, out_sharding=rotated_sharding)
-          sin = jnp.broadcast_to(sin, pairs.shape, out_sharding=rotated_sharding)
         swapped = jnp.flip(pairs, axis=-1)
         sign = jnp.asarray([-1.0, 1.0], dtype=jnp.float32)
         rotated_pairs = pairs * cos + swapped * sin * sign
@@ -1006,14 +1001,6 @@ class YarnRotaryEmbedding(nnx.Module):
 
       inputs_complex = first_half + 1j * second_half  # shape: [b, s, n, half_dim]
       # Apply the rotary transformation via complex multiplication.
-      # Broadcast the [b, s, 1, half_dim] frequencies onto the layout of the activations they
-      # multiply, so that a sharded heads axis (e.g. tensor parallelism) is preserved.
-      rotated_sharding = (
-          NamedSharding(self.mesh, jax.typeof(inputs_complex).sharding.spec)
-          if self.shard_mode == ShardMode.EXPLICIT
-          else None
-      )
-      freqs = jnp.broadcast_to(freqs, inputs_complex.shape, out_sharding=rotated_sharding)
       rotated = jnp.multiply(inputs_complex, freqs)  # shape: [b, s, n, half_dim]
 
       # Convert the complex result back to a real tensor.
--- a/src/maxtext/layers/embeddings.py
+++ b/src/maxtext/layers/embeddings.py
@@ -861,8 +861,10 @@ class YarnRotaryEmbedding(nnx.Module):
     if self.pairwise and not self.interleave:
       raise ValueError("rope_pairwise=True requires rope_interleave=True.")
 
+    # The gathered frequencies are [batch, length, half_dim]; the trailing axis is a slice of
+    # the head dimension, which is never sharded.
     self.freqs_sharding = (
-        create_sharding(mesh, ("activation_batch", "activation_length", "q_heads"))
+        create_sharding(mesh, ("activation_batch", "activation_length", None))
         if shard_mode == ShardMode.EXPLICIT
         else None
     )
@@ -982,7 +984,7 @@ class YarnRotaryEmbedding(nnx.Module):
         cos = jnp.real(freqs)[..., jnp.newaxis]
         sin = jnp.imag(freqs)[..., jnp.newaxis]
         if self.shard_mode == ShardMode.EXPLICIT:
-          rotated_sharding = create_sharding(self.mesh, ("activation_batch", "activation_length", None, None, None))
+          rotated_sharding = NamedSharding(self.mesh, jax.typeof(pairs).sharding.spec)
           cos = jnp.broadcast_to(cos, pairs.shape, out_sharding=rotated_sharding)
           sin = jnp.broadcast_to(sin, pairs.shape, out_sharding=rotated_sharding)
         swapped = jnp.flip(pairs, axis=-1)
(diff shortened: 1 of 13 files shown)


src/maxtext/layers/embeddings.py
```diff
@@ -981,10 +982,6 @@ def __call__(self, inputs: Array, position: None | Array = None) -> Array:
         pairs = pairs.astype(jnp.float32)
         cos = jnp.real(freqs)[..., jnp.newaxis]
         sin = jnp.imag(freqs)[..., jnp.newaxis]
-        if self.shard_mode == ShardMode.EXPLICIT:
```
Shuwen-Fang, line 984: why are we deleting this?

NuojCheng: The deleted code are dead. freqs was mislabelled with `q_heads` and now this issue is fixed.

PR #5073, opened 2026-08-31
Fix standalone torchax converter for Qwen3.5, and enable ATTN_DP and EP

wenxindongwork:
# Description

Makes `use_standalone_converter` work end to end for Qwen3.5 against current tpu-inference, so RL rollouts run on vLLM's **native** Qwen3.5 (torchax) model with weight sync from MaxText.

- **`torchax_converter/qwen35_moe.py`**: updates the converter to the current tpu-inference runner layout (the drift behind #4713) — transposes on the fused projections (and none on `o_proj` / GDN `out_proj` / shared `down_proj`), KV-head replication when attention shards exceed KV heads, current + legacy routed-expert names, `A_log` kept float32.
- **`maxtext_vllm_rollout.py`**: actually applies the converter — `_create_model_converter` instantiates it when `use_standalone_converter` is set, and `MaxTextVllmSampler.update_params` syncs via a new `_sync_standalone_converted` (KV-cache teardown → convert → chunked reshard → in-place runner-state assignment → KV-cache rebuild).
- **EP + attention DP**: the converter is parameterized on the sampler's sharding (`attn_dp_size`, `enable_expert_parallel`, filled automatically from the rollout config): attention/GDN/shared-expert shard `tp // attn_dp` ways, and routed experts use the GMM_EP layout under expert parallelism. Defaults reproduce plain TP bit-for-bit.

To use in the e2e: set `use_standalone_converter=True` and remove the `MaxTextForCausalLM` hf_overrides + `maxtext_config` so vLLM runs its native model.

Caveat: this mirrors tpu-inference's internal layout, so it must be revalidated when tpu-inference changes weight processing, and FP8 checkpoints are not covered. The canonical-layout path (#5034 + vllm-project/tpu-inference#3477 + google/tunix#2014) removes that coupling.

# Tests

- Unit tests (`tests/post_training/unit/qwen35_standalone_converter_test.py`, 25 cases): the real `convert()` on a small synthetic scanned state in all three sharding modes (plain TP, attention DP, EP), checked against expectations computed directly from the inputs — KV-head replication, GMM_TP/GMM_EP expert layouts, untransposed kernels, dtypes, expert-name aliases, full key coverage — plus the converter selection/hint plumbing and the `_sync_standalone_converted` paths (routing, in-place update, chunked/fallback reshard, layout-drift and non-dict-state errors) with a mocked runner.

- gsm8k **0.9621/0.9629** (`lm_eval`, TP=8 + `--enable-expert-parallel` + `attn_dp_size=4`); the synced state is bit-identical to the HF-loaded state that scored it.
- Live e2e on TPU (v7x-8): MaxText ckpt → sampler sync into a running, randomized native vLLM engine (TP=8 + attn_dp=4 + EP): all dumped tensors bit-identical to the HF-loaded reference, 4/4 greedy generations identical, sync 36 s.
- CPU bit-validation of `convert()` against runner-state dumps in both plain TP=8 (all 573 tensors) and TP=8 + attn_dp=4 + EP; coverage check confirms every runner weight is produced.

# Checklist

### DIFF
--- a/src/maxtext/integration/vllm/maxtext_vllm_rollout.py
+++ b/src/maxtext/integration/vllm/maxtext_vllm_rollout.py
@@ -38,7 +38,9 @@ from flax import nnx
 from flax.traverse_util import flatten_dict, unflatten_dict
 
 from tunix.generate import mappings
+from tunix.generate import utils as tunix_gen_utils
 from tunix.generate.vllm_sampler import VllmConfig, VllmSampler
+from tunix.rl import reshard as tunix_reshard
 from tunix.rl.rollout import base_rollout, vllm_rollout
 
 from maxtext.integration.vllm.convert_utils import _sharding_summary
@@ -46,10 +48,10 @@ from maxtext.integration.vllm.weight_converter import (
     WeightConverter,
     MODEL_TO_CONVERSION_RULES,
 )
+from maxtext.integration.vllm.torchax_converter.base import BaseMaxTextToVLLMConverter
 from maxtext.integration.vllm.torchax_converter.gemma4_moe import Gemma4MaxTextToVLLMConverter
 from maxtext.integration.vllm.torchax_converter.qwen35_moe import Qwen35MaxTextToVLLMConverter
 from maxtext.integration.vllm.torchax_converter.qwen3_moe import Qwen3MaxTextToVLLMConverter
-
 
 # Sentinel distinguishing "this model has no entry" from "this model has an
 # entry whose value is None", which means direct-sync-only.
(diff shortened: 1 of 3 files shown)


YixuanWang-99 (approved):

src/maxtext/integration/vllm/maxtext_vllm_rollout.py
```diff
@@ -73,10 +75,27 @@ def _create_model_converter(
     mesh: jax.sharding.Mesh,
     use_hf_mapping: bool = False,
     use_weight_converter: bool = False,
+    use_standalone_converter: bool = False,
+    sharding_hints: Optional[dict] = None,
```
hengtaoguo, line 79: What does this `sharding_hints` contain/mirror from the tpu-inference? Is it allowing any customized keys?

hengtaoguo (approved):
Thanks for the update fix!

src/maxtext/integration/vllm/maxtext_vllm_rollout.py
wenxindongwork: thanks for the thorough review. currently sharding_hints contains only two params

```
sharding_hints = {
    "attn_dp_size":     int
    "enable_expert_parallel":  True/False
}
```

PR #5079, opened 2026-08-31
Add StreamIndex Kernel for CSA [Deepseek v4]

octatrifan:
# Description

Adds a fused Pallas TPU kernel (`csa_streamindex_score`) for DeepSeek-V4 Compressed Sparse Attention (CSA) indexer scoring (for the forward pass)
BUGS: b/552514390

### Summary
- Fuses the $Q \cdot K^T$ matmul, ReLU, head weighting, and causal masking into on-chip VMEM to avoid materializing large intermediate tensors in HBM.
- Supports head-major layout $[B, H, S, D]$ and wraps with `@jax.custom_vjp`.
- Gated behind `use_csa_streamindex_kernel: bool = False` in `base.yml` with automatic fallback to standard einsum.
- Wrapped with `@jax.custom_vjp`: executes the fused Pallas kernel on the forward pass, and falls back to reference einsum autodiff (chunked) on the backward pass.

### Implementation
Tiles query tokens in blocks of 128 and compressed keys in blocks of 1024. For each block, it computes `relu(q @ k.T) * weights`, sums across all 64 heads directly in VMEM, applies the causal mask, and writes out the final `[B, S, W]` scores. This avoids allocating the 4D `[B, H, S, W]` scores tensor in HBM entirely.  It does **not include** the top-k at the end in-kernel.

# Tests

- Unit tests in [`tests/unit/csa_streamindex_test.py`](file:///usr/local/google/home/octatrifan/repos/maxtext/tests/unit/csa_streamindex_test.py):
  - Numerical parity vs reference einsum (with and without causal mask).
  - Custom VJP gradient parity.
  - Layer integration and fallback checks.
  - JAXPR graph verification (`pallas_call` vs `dot_general`).

Reproduce:
```bash
pytest tests/unit/csa_streamindex_test.py
```

# Performance
Microbenchmarks on a single TPU, just for the Index Scoring part (not e2e results). Compared against both the unchunked baseline einsum and the orthogonal head-chunking scan proposed in PR #5116 (which is a pure JAX solution to the OOM problem, but which requires more memory and is slower than StreamIndex)

StreamIndex Pallas Kernel vs. Head Chunking (PR #5116) vs. Baseline
Tested on single-chip TPU v5p. $H=64, D=128, W=S/4$:
| Sequence Length ($S$) | Batch ($B$) | Total Tokens | Baseline Einsum | Head Chunking (PR #5116, $c=16$) | StreamIndex Kernel | Kernel Speedup |
|---|---|---|---|---|---|---|
| 16,384 | 2 | 32,768 | 16.31 ms | 8.45 ms | **5.06 ms** | **1.67x vs PR / 3.22x vs Base** |
| 32,768 | 2 | 65,536 | 64.01 ms | 32.02 ms | **19.77 ms** | **1.62x vs PR / 3.24x vs Base** |
| 65,536 | 2 | 131,072 | OOM (274 GB req) | 125.90 ms (20.09 GiB) | **78.42 ms (18.04 GiB)** | **1.61x vs PR** |
| 65,536 | 4 | 262,144 | OOM | 251.84 ms (40.17 GiB) | **156.84 ms (36.08 GiB)** | **1.61x vs PR** |
| 65,536 | 8 | 524,288 | OOM | OOM | 313.53 ms (72.16 GiB) | **StreamIndex Only** |
| 131,072 | 1 | 131,072 | OOM | 247.27 ms (36.09 GiB) | **156.59 ms (34.04 GiB)** | **1.58x vs PR** |
| 131,072 | 2 | 262,144 | OOM | OOM| **313.14 ms (68.08 GiB)** | **StreamIndex Only** |
| 131,072 | 4 | 524,288 | OOM | OOM | OOM | Physical Limit (>96 GB) |
---
* **Speed**: StreamIndex is **1.6x faster** than Head Chunking and **3.2x faster** than baseline einsum across all context lengths.
* **Scalability**: StreamIndex supports **2x larger contexts** on a single TPU v5p, successfully running at $S=65\text{k}, B=8$ and $S=131\text{k}, B=2$ where Head chunking crashes with OOM.
* **Peak Memory**: StreamIndex uses less HBM by keeping intermediate score tensors on-chip in VMEM rather than HBM.

# Checklist

### DIFF
--- a/src/maxtext/kernels/attention/csa_streamindex.py
+++ b/src/maxtext/kernels/attention/csa_streamindex.py
@@ -1,0 +1,281 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Fused Pallas TPU kernel for DeepSeek-V4 CSA StreamIndex Score Computation.
+
+Computes:
+  index_scores = sum_h(ReLU(q_h @ comp^T) * softmax_scale * w_h)
+
+The kernel fuses the dot product, ReLU activation, softmax scaling and
+head-weight contraction so the intermediate [B, H, S, W] tensor stays in VMEM.
+Masking is the caller's responsibility; see `csa_streamindex_score_kernel`.
+"""
+
+import functools
+import jax
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+import jax.numpy as jnp
+
+
+# Default Pallas tile sizes, exported so tests and callers share one definition.
+# 128x1024 measured fastest on v6e at H=64; smaller head counts get larger tiles
+# to keep the MXU fed. Tuning data in PR #5079.
+DEFAULT_BLOCK_Q_LARGE_H = 128
+DEFAULT_BLOCK_W_LARGE_H = 1024
+DEFAULT_BLOCK_Q_SMALL_H = 256
+DEFAULT_BLOCK_W_SMALL_H = 2048
+
+# Heads per backward-pass scan iteration. Not `csa_qk_head_chunk_size`: that one
+# trades speed for HBM in the forward einsum and defaults to off, whereas this
+# controls reverse-mode fusion and must stay on or the backward OOMs at 32k.
+# 1 measured strictly best (fastest and lowest HBM) across B in {1,2,4} and
+# H in {16,32,64} on v6e -- at 1 XLA fuses the scan body and the residual never
+# reaches HBM.
+DEFAULT_BWD_HEAD_CHUNK = 1
+
+
+def csa_streamindex_score_kernel(
+    q_ref,  # [num_heads, block_q, head_dim]
+    k_ref,  # [block_w, head_dim]
+    w_ref,  # [block_q, num_heads]
+    out_ref,  # [block_q, block_w]
+    *,
+    softmax_scale: float,
+):
+  """Fused Pallas TPU kernel with 2D MXU matmul.
+
+  Masking is deliberately NOT done here: the kernel has no `position_ids` and
+  would have to synthesize positions from grid indices, which is wrong for
+  packed sequences. The caller masks, using the `future_mask` it already builds.
+  """
+  num_heads, block_q, head_dim = q_ref.shape
+  block_w, _ = k_ref.shape
+
+  # Reshape Q to 2D: (num_heads * block_q, head_dim) for native 2D systolic array MXU contraction
+  q_2d = q_ref[...].reshape(num_heads * block_q, head_dim)
+  k_2d = k_ref[...]
+
+  # 2D MXU matmul: (num_heads * block_q, head_dim) @ (block_w, head_dim)^T -> (num_heads * block_q, block_w)
+  s_2d = jnp.einsum("nd,md->nm", q_2d, k_2d, preferred_element_type=jnp.float32)
+
+  # Reshape to (num_heads, block_q, block_w) and apply ReLU
+  s = s_2d.reshape(num_heads, block_q, block_w)
+  s = jnp.maximum(s, 0.0)
+
+  # Multiply by weights and sum across heads in VMEM
+  w = w_ref[...].astype(jnp.float32).transpose(1, 0)[:, :, None]
+  s_weighted = jnp.sum(s * w, axis=0) * softmax_scale
+
+  out_ref[...] = s_weighted.astype(out_ref.dtype)
+
+
+def _csa_streamindex_score_pallas_fwd(
+    q: jax.Array,
+    compressed: jax.Array,
+    weights: jax.Array,
+    *,
+    softmax_scale: float,
+    block_q: int | None = None,
+    block_w: int | None = None,
+    interpret: bool = False,
+) -> jax.Array:
+  """Forward implementation using fused Pallas TPU kernel for head-major [B, H, S, D] q."""
+  batch_size, num_heads, seq_len, head_dim = q.shape
+  _, compressed_len, comp_head_dim = compressed.shape
+  assert comp_head_dim == head_dim, f"{comp_head_dim=} != {head_dim=}"
+  assert weights.shape == (batch_size, seq_len, num_heads), f"{weights.shape=} != {(batch_size, seq_len, num_heads)=}"
+
+  if block_q is None:
+    block_q = DEFAULT_BLOCK_Q_LARGE_H if num_heads >= 32 else DEFAULT_BLOCK_Q_SMALL_H
+  if block_w is None:
+    block_w = DEFAULT_BLOCK_W_LARGE_H if num_heads >= 32 else DEFAULT_BLOCK_W_SMALL_H
+
+  # Blocks may exceed the array; padding keeps the grid uniform. This also
+  # covers seq_len < block_q, which pads up to a single query block.
+  padded_s = ((seq_len + block_q - 1) // block_q) * block_q
+  padded_w = ((compressed_len + block_w - 1) // block_w) * block_w
+
+  if padded_s > seq_len:
+    pad_s = padded_s - seq_len
+    q = jnp.pad(q, ((0, 0), (0, 0), (0, pad_s), (0, 0)))
+    weights = jnp.pad(weights, ((0, 0), (0, pad_s), (0, 0)))
+  if padded_w > compressed_len:
+    pad_w = padded_w - compressed_len
+    compressed = jnp.pad(compressed, ((0, 0), (0, pad_w), (0, 0)))
+
+  grid = (batch_size, padded_s // block_q, padded_w // block_w)
+
+  in_specs = [
+      pl.BlockSpec((None, num_heads, block_q, head_dim), lambda b, i, j: (b, 0, i, 0)),
+      pl.BlockSpec((None, block_w, head_dim), lambda b, i, j: (b, j, 0)),
+      pl.BlockSpec((None, block_q, num_heads), lambda b, i, j: (b, i, 0)),
+  ]
+  out_specs = pl.BlockSpec((None, block_q, block_w), lambda b, i, j: (b, i, j))
+
+  out = pl.pallas_call(
+      functools.partial(
+          csa_streamindex_score_kernel,
+          softmax_scale=softmax_scale,
+      ),
+      in_specs=in_specs,
+      out_specs=out_specs,
+      grid=grid,
+      compiler_params=pltpu.CompilerParams(
+          # Every (b, i, j) program writes a disjoint output block, so all three
+          # axes are parallel and Mosaic may reorder or split them.
+          dimension_semantics=("parallel", "parallel", "parallel"),
+      ),
+      out_shape=jax.ShapeDtypeStruct((batch_size, padded_s, padded_w), jnp.float32),
+      interpret=interpret,
+  )(q, compressed, weights)
+
+  return out[:, :seq_len, :compressed_len]
+
+
+def csa_indexer_scores_jax(
+    q: jax.Array,
+    compressed: jax.Array,
+    weights: jax.Array,
+    *,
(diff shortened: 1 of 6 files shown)


src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+  return dq, dk, dw
+
+
+csa_streamindex_score.defvjp(
+    _csa_streamindex_score_fwd, _csa_streamindex_score_bwd
+)
+
+
+def reference_csa_streamindex_score(
+    q: jax.Array,
```
dandragona, line 212: We should depend on the real pure JAX CSA StreamIndex implementation in MaxText so that this implementation doesn't diverge from what the source.

src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+  grid = (batch_size, padded_s // block_q, padded_w // block_w)
+
+  in_specs = [
+      pl.BlockSpec((None, num_heads, block_q, head_dim), lambda b, i, j: (b, 0, i, 0)),
+      pl.BlockSpec((None, block_w, head_dim), lambda b, i, j: (b, j, 0)),
+      pl.BlockSpec((None, block_q, num_heads), lambda b, i, j: (b, i, 0)),
+  ]
+  out_specs = pl.BlockSpec((None, block_q, block_w), lambda b, i, j: (b, i, j))
+
+  out = pl.pallas_call(
```
dandragona, line 115: Swap this with an emit_pipeline, then make a standalone test which activates:

from absl import flags

FLAGS = flags.FLAGS
FLAGS.set_default("xla_xprof_register_llo_debug_info", "True")
FLAGS.set_default("xla_enable_custom_call_region_trace", "True")
FLAGS.set_default("xla_enable_transpose_trace", "True")
FLAGS.set_default("xla_xprof_enable_custom_call_tracing", "True")

and get us the xprof.

src/maxtext/layers/attention_compressed.py
dandragona, line 903: (I know this is pre-existing code) This line makes me sad since we are a pretraining library, is there a situation where future_mask is not None?

Maybe we can fold 909:911 into the else block above?

src/maxtext/layers/attention_compressed.py
```diff
@@ -874,20 +875,28 @@ def indexer_compressor_fn(buf_kv, buf_gate):
...
     q = self.rotary_emb(q, position_ids, unsqueeze_dim=1)
-
-    q = q.astype(jnp.float32)
-    compressed_kv = compressed_kv.astype(jnp.float32)
-
-    scores = jnp.einsum("bhsd,bhwd->bhsw", q, compressed_kv)
-    scores = jax.nn.relu(scores) * self.softmax_scale
     weights = self.weights_proj(hidden_states).astype(jnp.float32) * self.weights_scaling
-    index_scores = jnp.einsum("bhsw,bsh->bsw", scores, weights)
+    if self.config.use_csa_streamindex_kernel:
```
Rohan-Bierneni: Lets make sure that this kernel is only run in training mode. It will break both decode and prefill modes. Can you add a check where this is called only when model_mode == TRAIN

src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+  w = w_ref[...].astype(jnp.float32).transpose(1, 0)[:, :, None]
+  s_weighted = jnp.sum(s * w, axis=0) * softmax_scale
+
+  # In-VMEM causal future masking
+  if compress_rate > 0:
+    i = pl.program_id(1)
+    j = pl.program_id(2)
+    q_indices = i * block_q + jnp.arange(block_q, dtype=jnp.int32)[:, None]
+    k_indices = (j * block_w + jnp.arange(block_w, dtype=jnp.int32)[None, :]) * compress_rate
+    future_mask = (k_indices + compress_rate) > (q_indices + 1)
```
Rohan-Bierneni: This breaks in MaxText whenever packed sequences or document boundaries are used (where position_ids reset to 0 mid-sequence). 

Also, do we need to do masking in the kernel? attention_compressed.py already materializes future_mask in HBM bc [combined_invalid](https://github.com/AI-Hypercomputer/maxtext/blob/1814305e731d31fc9acd4eee7289285869a28df2/src/maxtext/layers/attention_compressed.py#L913) uses it top invalidate top-k outputs.

I think doing masking outside the kernel is fine, just also need to remove the check in https://github.com/AI-Hypercomputer/maxtext/blob/1814305e731d31fc9acd4eee7289285869a28df2/src/maxtext/layers/attention_compressed.py#L910 as well as this code here. 

Actually, have you tested this edge case where S < block_q with the current setup. I think it might fail

src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+  """Forward implementation using fused Pallas TPU kernel for head-major [B, H, S, D] q."""
+  batch_size, num_heads, seq_len, head_dim = q.shape
+  _, compressed_len, comp_head_dim = compressed.shape
+  assert comp_head_dim == head_dim, f"{comp_head_dim=} != {head_dim=}"
+  assert weights.shape == (batch_size, seq_len, num_heads), f"{weights.shape=} != {(batch_size, seq_len, num_heads)=}"
+
+  if block_q is None:
+    block_q = 128 if num_heads >= 32 else 256
+  if block_w is None:
+    block_w = 1024 if num_heads >= 32 else 2048
```
Rohan-Bierneni: Did you do any tile size tuning to set these specific default tile sizes? For DsV4, the num_heads=64, so

M = 64 * 128 = 8192
N = 1024
s_2d = 8192 * 1024 * 4 = 32MB

While this is fine for tpu limits when you set the scoped_vmem flag, what is the impact if we did something like block_q=64 or 128 & block_w=512 or 256?

src/maxtext/kernels/attention/csa_streamindex.py
Rohan-Bierneni: +1 to having an xprof trace and a percale trace if possible

src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+      functools.partial(
+          csa_streamindex_score_kernel,
+          softmax_scale=softmax_scale,
+          compress_rate=compress_rate,
+      ),
+      in_specs=in_specs,
+      out_specs=out_specs,
+      grid=grid,
+      compiler_params=pltpu.CompilerParams(
+          dimension_semantics=("parallel", "parallel", "arbitrary"),
```
Rohan-Bierneni: Marking the compressed key window (calling this W) as arbitrary would tell mosaic to process these in sequential order. 

But IIUC every W writes to an independent slice in out_specs so is it necessary to set it as "arbitrary" instead of "parallel"?

For ref: https://docs.jax.dev/en/latest/_autosummary/jax.experimental.pallas.tpu.CompilerParams.html#jax.experimental.pallas.tpu.CompilerParams.dimension_semantics

src/maxtext/kernels/attention/csa_streamindex.py
```diff
@@ -0,0 +1,233 @@
...
+      softmax_scale=softmax_scale,
+      compress_rate=compress_rate,
+      block_q=block_q,
+      block_w=block_w,
+      interpret=interpret,
+  )
+  return out, (q, compressed, weights)
+
+
+def _csa_streamindex_score_bwd(
```
Rohan-Bierneni, line 233: This should not be using the bwd jax baseline from the initial bringup, but the head chunking jax implementation that should cut the peak HBM by a fact of num_head_chunks (set to 8 i believe). 

Then, it would not materialize the full [B, H, S, W] matrix but [B, num_head_chunks, S, W] at a time.

tests/unit/csa_streamindex_test.py
```diff
@@ -0,0 +1,275 @@
...
+    w = s // 4
+    scale = d**-0.5
+    key1, key2, key3 = jax.random.split(self.key, 3)
+    q = jax.random.normal(key1, (b, h, s, d), dtype=jnp.bfloat16)
+    compressed = jax.random.normal(key2, (b, w, d), dtype=jnp.bfloat16)
+    weights = jax.random.normal(key3, (b, s, h), dtype=jnp.float32)
+
+    fn = jax.jit(
+        lambda q, k, w: csa_streamindex.csa_streamindex_score(
+            q, k, w, softmax_scale=scale, block_q=128, block_w=512, interpret=False
```
Rohan-Bierneni: Match this with the default tile sizes set in the code for block_q & block_w.

src/maxtext/layers/attention_compressed.py
```diff
@@ -874,20 +875,28 @@ def indexer_compressor_fn(buf_kv, buf_gate):
...
+      index_scores = csa_streamindex.csa_streamindex_score(
```
Rohan-Bierneni: How come this is not wrapped with jax.shard_map? Don't we need this to enable compatability with the mesh defined. Seems like it was there but removed in a recent commit

Rohan-Bierneni (commented):
Thanks for the pallas kernel implementation! I have left some comments PTAL

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Added `csa_indexer_scores_jax()` as the single definition. The layer's non-kernel path and the kernel's VJP both call it now, so there's nothing left to drift.

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Skipping for now out of scope for this PR, as discussed with the commenter. XProf added in the buganizer link.

src/maxtext/layers/attention_compressed.py
octatrifan: Done. `use_kernel = config.use_csa_streamindex_kernel` and `model_mode == MODEL_MODE_TRAIN`.

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Removed, the kernel was synthesizing positions from grid indices, which is wrong any time positions `≠ arange(S)`. The caller masks unconditionally now, which is free since future_mask is already materialized for the validity check.

Covered by `test_parity_seq_len_smaller_than_block_q`, which sweeps (S,W) = (1,1), (7,3), (64,16), (127,63).

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Swept on v6e (B=1, H=64, D=128), The smaller tiles you suggested are all slower — 64×256 by 1.7×. 256×1024 is 4% faster but doubles the VMEM footprint, and 128×1024 already has little headroom under the 32 MiB scoped limit at H=64. Keeping 128×1024.

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Changed to all-parallel and verified on v6e.

src/maxtext/kernels/attention/csa_streamindex.py
octatrifan: Done, added `lax.scan` over head chunks, which bounds the saved intermediate to [B, chunk, S, W] instead of [B, H, S, W]

tests/unit/csa_streamindex_test.py
octatrifan: Exported as `DEFAULT_BLOCK_{Q,W}_{LARGE,SMALL}_H`, and the tests reference the constants instead of hardcoding.

src/maxtext/layers/attention_compressed.py
octatrifan: Added, batch-only. It raises if the batch isn't divisible by the shard count. Context/sequence parallelism isn't supported and isn't attempted here.

src/maxtext/layers/attention_compressed.py
octatrifan: `future_mask` is a local, only assigned in the AR branch (`~is_valid`, from cache validity). Train and prefill always hit the None path so it's dead for pretraining, live for decode, where it marks unpopulated cache slots and so can't come from position_ids.

Folding into the else is equivalent, but the if `compressed_len == 0`: return sits between them, so it'd compute a `[b, s, 0]` mask the early return discards.

Rohan-Bierneni (approved):
LGTM. Thanks for the changes!

shuningjin (approved):

dandragona (approved):

PR #5088, opened 2026-09-01
Match Tunix peft_trainer_v2 performance in MaxTextTrainingEngine

NuojCheng:
# Description

`MaxTextTrainingEngine` was slower than Tunix's `PeftTrainer` v2 on the same model, the same loss function and the same `optax` transformation, and the gap turned out to be entirely host-side. This PR closes it, and fixes two correctness problems the measurement turned up along the way.

Five files, no new scripts:

| file | why |
| --- | --- |
| `src/maxtext/training_engine/maxtext_engine.py` | the step path, the accumulation normalization, the gradient norm |
| `src/maxtext/training_engine/inflight_throttler.py` | queue a scalar per step instead of the whole train state |
| `src/maxtext/training_engine/metrics.py` | deferred metrics write |
| `tests/post_training/unit/maxtext_engine_test.py` | coverage for all of the above |
| `tests/post_training/integration/maxtext_engine_grpo_loss_test.py` | size its batch to the mesh, per the behaviour change below |

## Performance

- **Trace the accumulation and update kernels under `nn_partitioning.axis_rules`**, so gradient accumulation is fused into the compiled program instead of running as an eager `jax.tree.map(add)` over every gradient leaf on every micro-batch.

  This is **the one behaviour change in the PR**, as opposed to a timing change. Those rules live in a context variable, so a kernel traced without them saw an empty rule set and every `maybe_shard_with_logical` inside the MaxText layers was silently a no-op; making them real means `micro_batch_size_to_train_on` must now be a multiple of `data x fsdp`. A batch those devices cannot split gets padded out by XLA, and the padded lanes are all-zero sequences that mask themselves out of attention entirely — their contribution comes back as NaN on the pad token's embedding row, giving unusable gradients under a finite loss. MaxText's splash kernel already asserts on exactly this ratio ("Batch dimension should be shardable among the devices in data and fsdp axis", `attention_op.py`); `dot_product` does not, so there it surfaces as a NaN rather than an error. Callers relying on an undersized micro-batch are the group affected, and in practice such a batch was already leaving devices idle. This is how it showed up: the GRPO integration test hard-coded a batch of 2 and ran on a 4-device host, hence the fifth file above.
- **Carry a pure `nnx.State` mirror across steps** in place of a per-step `nnx.split` of the module graph. The split was two graph traversals per step, scaling with node count — 21.3 ms on an unscanned qwen3-0.6b's 310 parameter leaves — and it allocated a large short-lived object graph twice per step, whose GC pauses landed on whichever step was unlucky (a 464 ms worst step against a 161.6 ms median).
- **Defer the metrics write until after the next dispatch**, so its device-to-host read overlaps live work rather than blocking on it.

## Correctness

- **Accumulate gradients unreduced and divide once by the summed denominator**, rather than averaging per-micro-batch means. The two agree only when every micro-batch carries the same number of tokens; on ragged batches, mean-of-means is simply not the gradient of the batch loss. The denominator rides along in checkpoint metadata so a mid-step resume can finish the division, and is backfilled from the restored per-micro-batch losses for checkpoints written before it was tracked.
- **Compute the gradient norm on every update**, not only under `skip_step_on_spikes`. `learning/grad_norm` is a metric `trainers/pre_train/train.py` reports every step and `metrics._METRICS_TO_LOG` already lists, but with spike-skipping off — base.yml's default — it read NaN. Tunix computes the same quantity unconditionally inside `peft_trainer_v2._update_step`. **`202a89ab8` landed the same fix on main while this was in review**; rebasing onto it, I kept main's placement — *before* clipping — over this branch's post-clipping one, because that is where Tunix's `optax.global_norm` sits (its clipping, when a caller configures any, is a later link in the optax chain). In `train.py`'s vocabulary that makes it `raw_grad_norm` rather than `learning/grad_norm`. What this PR still adds on top of main's version is the float32 reduction — a sum of squares over bf16 leaves overflows on production-size models — and the norm doubling as the throttler's handle. It agrees with Tunix bit for bit:

  | | MaxText `gradient_norm` | Tunix `grad_norm` |
  | --- | --- | --- |
  | GA=1, batch 8 | 59.77279281616211 | 59.77279281616211 |
  | GA=8, batch 8 | 18.25168800354004 | 18.25168800354004 |

  It costs ~0.6 ms per update — 0.7% of a GA=1 step, 0.1% of a GA=8 one, and below the noise floor on device (568.0 vs 570.7 ms TPU-busy per optimizer step at GA=8), because XLA fuses the reduction into an update kernel that already streams every gradient leaf.

  The norm also gives `InflightThrottler` a scalar to block on. It previously queued the train state itself, which pinned three parameter trees alive for as long as the entry sat in the queue, and — once the update kernel donates its state argument — could raise `Array has been deleted` out of `jax.block_until_ready` when an entry was popped after a later step had reused those buffers.

## Results

qwen3-0.6b, batch 8 × seq 1024, f32, `optax.sgd(1e-5)`, no clipping, measured as an A/B on these four files alone — the "before" arm is `git checkout origin/main -- src/maxtext/training_engine/` on this same commit, so nothing else differs. 4 × v6e, `fsdp=4`, unscanned, median of 19 steps after warmup, no profiler attached:

| | before (`main`) | after (this PR) | speedup |
| --- | --- | --- | --- |
| GA=8, per optimizer step | 4443.3 ms | **613.0 ms** | **7.25x** |
| GA=8, per micro-batch | 555.4 ms | **76.6 ms** | **7.25x** |
| GA=8, 23-step loop | 135.8 s | **24.2 s** | 5.61x |
| GA=8, process wall clock | 161.3 s | **49.8 s** | 3.24x |
| GA=1, per step | 516.8 ms | **84.5 ms** | **6.12x** |
| GA=1, 23-step loop | 21.1 s | **8.7 s** | 2.43x |
| GA=1, process wall clock | 46.6 s | **34.1 s** | 1.37x |

Gradient accumulation multiplies the win, because the removed `nnx.split` sat in `fwd_bwd`, which runs once per micro-batch rather than once per optimizer step. The process-wall-clock rows are lower than the rest because ~25 s of interpreter start, imports, config and engine build is fixed cost this PR does not touch.

The saving is a fixed quantity of host time, so what it is worth also depends on how big the NNX module graph is — `scan_layers` collapses a decoder stack into one stacked node set, so a *scanned* qwen3.5-35b-a3b (70 parameter leaves) sees 2322.5 → 2314.0 ms/step, a 1.004x median, while its worst step still drops 2641.8 → 2314.6 ms as the GC pauses go away — and on how much device time there is to hide host work behind.

Against Tunix's trainer at GA=8, driving the *same* MaxText model, so only the trainer differs (measured on an earlier base, where this PR's arm read 589.6 ms rather than 613.0):

| GA=8, qwen3-0.6b | ms/optimizer step | ms/micro-batch | TPU-busy/step | utilization |
| `MaxTextTrainingEngine` | **589.6** | **73.7** | 568.0 ms | **96%** |
| `PeftTrainer`, MaxText model | 1225.9 | 153.2 | 617.7 ms | 50% |

Two independent checks that the win is host-only: TPU-busy time across the GA=8 A/B is 3408.12 vs 3408.13 ms over the six optimizer steps each trace covers — agreement to one part in 340,000 — and `jit_accum_kernel` carries the same XLA module hash (`13160937502384084184`) on both sides.

Full write-up, methodology and xplane traces: #5060.

# Tests

`tests/post_training/unit/maxtext_engine_test.py`, extended here to cover:

- the cached pure-state path end to end, on a model with non-`Param` state and over two steps, which is where a stale or wrongly-partitioned cache would show up;
- unreduced accumulation and its denominator, including a mid-step checkpoint round-trip;
- the throttler queueing one scalar rather than the state's leaves;
- `gradient_norm` being recorded with `skip_step_on_spikes` off — the configuration that used to produce nothing.

```
JAX_PLATFORMS=cpu pytest tests/post_training/unit/maxtext_engine_test.py \
  tests/post_training/unit/maxtext_engine_constructor_test.py \
  tests/post_training/unit/maxtext_engine_e2e_test.py \
  tests/post_training/unit/router_replay_engine_test.py
# 59 passed
```

On TPU, `tests/post_training/integration/maxtext_engine_grpo_loss_test.py` passes against the real GCS checkpoint (1 passed in 51.67 s), and the numerical parity harness in #5060 shows identical loss and gradients agreeing to `rel_l2` 3.0e-4 against Tunix at GA=1. Two things in that test are now load-bearing: its batch comes from `mesh.shape["data"] * mesh.shape["fsdp"]` rather than a hard-coded 2, for the reason above; and `matmul_precision=highest` is pinned, because its KL assertion compares log-probs from two code paths — Tunix's `compute_per_token_logps` for the reference against the engine's sharded forward for the policy — which at bf16 disagree by ~2e-2 on values near -12.6, and `low_var_kl` squares that into a KL of ~1.2e-4, an order of magnitude above the tolerance the assertion is meant to police.

# Checklist

### DIFF
--- a/src/maxtext/training_engine/maxtext_engine.py
+++ b/src/maxtext/training_engine/maxtext_engine.py
@@ -55,8 +55,8 @@ import numpy as np
 # same situation. Real buffers are identified by their train step, so this cannot collide.
 EMPTY_METRICS_BUFFER_ID = -1
 
-# Where `nnx.split(TrainStateNNX(...))` puts the model's own state. See
-# `_check_pure_state_reusable` for why the engine verifies this rather than assuming it.
+# Where `nnx.split(TrainStateNNX(...))` puts the model's own state; verified, not assumed,
+# by `_check_pure_state_reusable`.
 _MODEL_STATE_KEY = "model"
 
 _PURE_STATE_FALLBACK_WARNING = (
@@ -371,18 +371,15 @@ class MaxTextTrainingEngine(abstract_engine.AbstractTrainingEngine):
     else:
       self._model = model_or_model_mesh_pair
     self._state: Any = None
-    # Pure-pytree mirror of the model and the train state, carried across steps so the step
-    # path never re-walks the module graph. `None` means "not cached", which is also how the
-    # fast path is switched off; see `_refresh_pure_state`.
+    # Pure-pytree mirror of the model and train state, carried across steps so the step path
+    # never re-walks the module graph. `None` means "not cached".
     self._params_pure: Any = None
     self._rest_pure: Any = None
     self._state_pure: Any = None
     self._pure_state_warned: bool = False
     self._accumulated_grads: Any = None
-    # Sum of the per-micro-batch loss denominators behind `_accumulated_grads`. Tracked
-    # alongside them because the gradients are accumulated unreduced: this is the divisor
-    # `update()` applies once, and it is not `micro_step_count` unless every micro-batch
-    # happened to carry the same token count.
+    # Summed loss denominators behind `_accumulated_grads`, which are unreduced: this is the
+    # divisor `update()` applies once.
     self._accumulated_denominator: Any = None
     self._micro_step_count = 0
     # Set when this run resumed from an intra-step checkpoint, cleared once the step it
@@ -519,19 +516,12 @@ class MaxTextTrainingEngine(abstract_engine.AbstractTrainingEngine):
   def _sharding_ctx(self):
     """Activates the mesh and logical axis rules the MaxText layers are written against.
 
-    MaxText modules place their own sharding constraints through
-    `nn_partitioning.get_axis_rules()` (`sharding.maybe_shard_with_logical` and friends).
-    Those rules live in a context variable, so a kernel traced outside this context sees an
-    empty rule set: every logical constraint silently becomes a no-op and XLA is left to
-    guess the partitioning for activations and gradients. It guesses badly -- on
-    llama3.1-8b/fsdp=8 the same fwd/bwd measured 1012 ms untraced-in-context against 581 ms
-    inside it, with no numerical difference. `train.py` wraps its own `jax.jit` the same way
-    (`with jax.set_mesh(mesh), mesh, nn_partitioning.axis_rules(config.logical_axis_rules)`),
-    which is why the standalone trainer never hit this. Its middle `mesh` is left out here:
-    `Mesh.__enter__` is deprecated and `jax.set_mesh` already covers it.
-
-    Entered around the *call*, not around `jax.jit(...)`: jit is lazy, so the rules must be
-    live when the first call triggers tracing.
+    The rules live in a context variable, so a kernel traced outside this context sees an
+    empty rule set, every `maybe_shard_with_logical` becomes a no-op and XLA guesses the
+    partitioning -- badly: 1012 ms against 581 ms for the same fwd/bwd on llama3.1-8b/fsdp=8.
+    `train.py` wraps its own `jax.jit` the same way. Entered around the *call*, since jit is
+    lazy and the rules must be live when tracing happens. Note a batch that data x fsdp
+    cannot divide gives NaN gradients once the constraints are real.
     """
     if self._mesh is None:
       yield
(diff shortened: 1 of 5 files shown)


src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -484,8 +512,156 @@ def with_gen_model_input_fn(self, gen_model_input_fn: Callable[[Any], dict[str,
...
+    self._params_pure = None
+    self._rest_pure = None
+    self._state_pure = None
+
+  def _disable_pure_state(self, reason: str) -> None:
+    """Falls back to re-splitting the module graph on every step, saying so once."""
+    self._invalidate_pure_state()
+    if not self._pure_state_warned:
+      self._pure_state_warned = True
+      logging.warning(_PURE_STATE_FALLBACK_WARNING, reason)
```
gobbleturk, line 547: is there a fix we can recommend? in this case?

gobbleturk (approved):

igorts-git (approved):

A9isha (approved):

PR #5097, opened 2026-09-02
Training engine: implement eval_step API

SurbhiJainUSC:
# Description

* Implements `eval_step()` API in `MaxTextTrainingEngine` mirroring tunix's peft_trainer_v2. 
* MetricsRecorder takes a `mode` argument and the engine owns a second mode="eval" instance, keeping eval numbers out of the train buffer and out of the checkpoint; MetricsBuffer.mode then flows through write_metrics into both the console line and the TensorBoard tag, which becomes {mode}/{name}. 
* eval_context() accumulates across every micro-batch into one buffer (WeightedMetric appends sum and denominator, so compute() is the weighted mean over the whole eval set) and writes it exactly once on exit.

# Tests

CI tests

# Checklist

### DIFF
--- a/src/maxtext/training_engine/maxtext_engine.py
+++ b/src/maxtext/training_engine/maxtext_engine.py
@@ -157,6 +157,53 @@ _UNCOMPARABLE_STRUCTURE_HINT = (
 )
 
 
+def _normalize_loss_output(out: Any, has_aux: bool) -> abstract_engine.LossOutput:
+  """Normalizes whatever a loss function returned a `LossOutput`.
+
+  Shared by the training and evaluation kernels.
+
+  Args:
+    out: The loss function's return value.
+    has_aux: Whether the caller considers a 2-tuple's second element to be auxiliary
+      output worth recording.
+
+  Returns:
+    The equivalent `LossOutput`.
+
+  Raises:
+    TypeError: If `out` matches none of the accepted shapes.
+  """
+  if isinstance(out, abstract_engine.LossOutput):
+    return out
+  if isinstance(out, abstract_engine.WeightedMetric):
+    return abstract_engine.LossOutput(primary_loss=out, aux_metrics={})
+  if isinstance(out, (tuple, list)) and len(out) == 2:
+    loss_val, aux = out
+    if isinstance(loss_val, abstract_engine.WeightedMetric):
+      primary_loss = loss_val
+    elif isinstance(aux, dict) and "xent_sum" in aux and "total_weights" in aux:
+      primary_loss = abstract_engine.WeightedMetric(
+          unreduced_sum=aux["xent_sum"],
+          denominator=aux["total_weights"],
+      )
+    else:
+      raise TypeError(
+          f"Cannot construct WeightedMetric from 2-tuple loss return with elements "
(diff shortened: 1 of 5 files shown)


khatwanimohit (approved):

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1142,29 +1207,59 @@ def update(self, **kwargs: Any) -> int:
...
+    if self._compile_requested:
+      dynamic_batch, static_batch = _split_static_and_dynamic(batch)
+      signature = _batch_signature(dynamic_batch, static_batch)
+      if self._compiled_eval is None or self._needs_recompile(signature, self._compiled_eval_signature):
+        self._compile_eval_for_batch(dynamic_batch, static_batch)
+      loss, aux = self._compiled_eval(params, rest, dynamic_batch)
+    else:
+      loss, aux = self._eval_kernel(params, rest, batch)
+
+    if isinstance(loss, abstract_engine.WeightedMetric):
```
igorts-git, line 1265: Are there situations where loss won't be WeightedMetric? Should we support those cases as well? Otherwise we are not printing anything.

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1142,29 +1207,59 @@ def update(self, **kwargs: Any) -> int:
...
+      self.record_metrics("loss", loss, mode=metrics_module.Mode.EVAL)
```
igorts-git, line 1266: Remind me, do this log immediately (and blocks on the JAX to finish execution) or only adds the loss to the buffer?

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1142,29 +1207,59 @@ def update(self, **kwargs: Any) -> int:
...
     """
-    if not self._eval_step_warned:
-      self._eval_step_warned = True
-      logging.warning(
-          "MaxTextTrainingEngine.eval_step is not implemented: it evaluates nothing and "
-          "records no metrics, so any eval result reported for this run is meaningless. "
-          "Logged once per engine instance."
-      )
+    batch = self._prepare_batch(payload)
+
```
igorts-git, line 1242: Should we use `self._throttler.wait_for_all()` here and `self._throttler.add_computation()` at the end of this function?

igorts-git (approved):

src/maxtext/training_engine/maxtext_engine.py
SurbhiJainUSC: It only adds the compuation to the buffer, wihtout blocking.

src/maxtext/training_engine/maxtext_engine.py
SurbhiJainUSC: I think we initially added this to support MaxText default loss function with this training engine, where the loss value is not an object of `WeightedMetric`.

src/maxtext/training_engine/maxtext_engine.py
SurbhiJainUSC: Good catch, added rate-limiting in eval to prevent Host OOMs on large eval sets.

igorts-git (approved):

PR #5100, opened 2026-09-02
Add RL sequence-packing and vLLM block-size knobs

gutianyu-google:
# Description

Tunix packs RL training sequences into fixed-token rows when its training config carries `max_seq_token_per_tpu`, and it forwards the packed segment ids to the model only when the model's call signature has a parameter literally named `segment_ids`. MaxText exposed neither, so packing could not be turned on from an RL config: the CLI rejected the key (not in `RLConfig`), and the adapter would never have received the packed ids.

This PR adds:

- `max_seq_token_per_tpu` to `RLConfig` (default `0` = unpacked, one padded row per sequence, i.e. today's behavior), plumbed into `RLTrainingConfig`. When set, a maximal sequence (`max_prefill_predict_length` + generation length) must fit in one row; Tunix validates this at startup.
- A `segment_ids` parameter on `TunixMaxTextAdapter.__call__`, mapped to MaxText's `decoder_segment_ids`, so packed rows keep per-sequence attention isolation. Without it the adapter falls back to synthesizing a pad mask and packed sequences would attend to each other.
- `vllm_block_size` to `RLConfig` (default `None`, backend-chosen as today). tpu_inference derives the KV-cache page size from the engine shape, so unrelated engine changes (`max_model_len`, TP/DP layout) silently move it; pinning it keeps generation comparable across such experiments. It is forwarded through `rollout_vllm_kwargs` only when set.

Measured on a Qwen3-0.6B GRPO run on TPU v7x (2048 sequences/step, 12288-token rows, prompts ~150 tokens, completions ~4400 tokens): packing cut `actor_train_time` from 15.2 s to 6.7 s per step, with rewards and completion lengths unchanged. Both defaults leave existing configs untouched.

**Note on the adapter change:** the `segment_ids` parameter takes effect on every Tunix batch, not only packed ones. Once the gate passes, Tunix sends its non-pad mask as `segment_ids` on unpacked batches too, so the adapter's own pad-id fallback no longer fires on that path and pad-mask authority moves from the adapter's `pad_id` to Tunix's. Both derive from the tokenizer pad id today, so values are unchanged; "defaults leave existing configs untouched" applies to the config knobs, not to this.

# Tests

- `tests/post_training/unit/tunix_adapter_test.py`: `segment_ids` is forwarded as `decoder_segment_ids`, and takes precedence when both are given.
- `tests/post_training/unit/train_rl_test.py`: `RLConfig` defaults and explicit values for both knobs.

```
python -m pytest tests/post_training/unit/tunix_adapter_test.py tests/post_training/unit/train_rl_test.py
```

# Checklist

### DIFF
--- a/src/maxtext/configs/types.py
+++ b/src/maxtext/configs/types.py
@@ -2727,6 +2727,11 @@ class VLLM(BaseModel):
   async_scheduling: bool = Field(False, description="Enable asynchronous scheduling in vLLM.")
   max_num_batched_tokens: Optional[int] = Field(None, description="Max number of batched tokens in vLLM.")
   max_num_seqs: Optional[int] = Field(None, description="Max number of sequences in vLLM.")
+  vllm_block_size: Optional[int] = Field(
+      None,
+      gt=0,
+      description="KV-cache block (page) size for vLLM. None lets the backend pick it from the engine shape.",
+  )
   stop_strings: Optional[list[str]] = Field(None, description="List of stop strings for vLLM decoding.")
   vllm_additional_config: dict[str, Any] = Field(default_factory=dict, description="Additional vLLM config options.")
   vllm_hf_overrides: dict[str, Any] = Field(
@@ -2858,6 +2863,15 @@ class RLDataset(BaseModel):
   train_fraction: float = Field(1.0, gt=0.0, le=1.0, description="Fraction of the dataset to be used for training.")
   train_micro_batch_size: int = Field(-1, description="Micro batch size for training.")
   rollout_micro_batch_size: int = Field(-1, description="Micro batch size for rollout.")
+  max_seq_token_per_tpu: int = Field(
+      0,
+      ge=0,
+      description=(
+          "Token budget per packed training row (Tunix `max_seq_token_per_tpu`). When > 0, rollout sequences are packed "
+          "into rows of this many tokens for the actor/reference passes instead of one padded row per sequence; a maximal "
+          "sequence (max_prefill_predict_length + generation length) must fit in one row. 0 disables packing."
+      ),
+  )
   dataset_processor_path: str = Field(
       "",
       description=(
@@ -5164,6 +5178,16 @@ class RLConfig(
     model_name = getattr(self, "model_name", None)
     if model_name is None:
       raise ValueError("model_name is not set. Please pass model_name in your command.")
+
+    # With sequence packing on, a maximal sequence (prompt cap + generation
+    # cap = max_target_length) must fit in one packed row. Tunix checks this
+    # too, but only in the learner, after the models are already on the
+    # accelerators; fail here, before any of that work.
+    if 0 < self.max_seq_token_per_tpu < self.max_target_length:
+      raise ValueError(
+          f"max_seq_token_per_tpu ({self.max_seq_token_per_tpu}) must be at least "
+          f"max_target_length ({self.max_target_length}) when sequence packing is enabled."
+      )
 
     # Set tokenizer_path based on model_name if not explicitly provided.
     tokenizer_path = getattr(self, "tokenizer_path", None)
--- a/src/maxtext/integration/tunix/tunix_adapter.py
+++ b/src/maxtext/integration/tunix/tunix_adapter.py
@@ -109,10 +109,18 @@ class TunixMaxTextAdapter(nnx.Module):
       decoder_segment_ids: Optional[Array] = None,
       output_hidden_states: bool = False,  # ignored
       forced_routed_experts: Optional[Array] = None,
+      segment_ids: Optional[Array] = None,
   ) -> Tuple[Array, None]:
     """Forward compatible with Tunix Trainers default loss.
     Returns logits, None.
+
+    `segment_ids` is the name Tunix uses for packed-sequence segment ids: it
+    forwards them only to models whose call signature has a parameter of that
+    exact name. They are MaxText's `decoder_segment_ids`; when both are given,
+    `segment_ids` wins so packed rows keep per-sequence attention isolation.
     """
+    if segment_ids is not None:
+      decoder_segment_ids = segment_ids
     if decoder_segment_ids is None and self._pad_id is not None:
       decoder_segment_ids = (input_tokens != self._pad_id).astype(jnp.int32)
     logits = self.base(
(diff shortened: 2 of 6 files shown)


khatwanimohit (approved):

src/maxtext/configs/types.py
```diff
@@ -2659,6 +2659,11 @@ class VLLM(BaseModel):
   async_scheduling: bool = Field(False, description="Enable asynchronous scheduling in vLLM.")
   max_num_batched_tokens: Optional[int] = Field(None, description="Max number of batched tokens in vLLM.")
   max_num_seqs: Optional[int] = Field(None, description="Max number of sequences in vLLM.")
+  vllm_block_size: Optional[int] = Field(
+      None,
+      gt=0,
+      description="KV-cache block (page) size for vLLM. None lets the backend pick it from the engine shape.",
+  )
```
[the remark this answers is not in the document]
gutianyu-google: vLLM itself only requires `block_size > 0`, and the TPU backend's real constraints are dtype-dependent and validated at engine construction with explicit errors, so a power-of-two check here would reject valid configurations; keeping `gt=0` to match vLLM's own contract.

src/maxtext/trainers/post_train/rl/train_rl.py
```diff
@@ -436,6 +436,18 @@ def create_rl_components(  # pylint: disable=too-many-positional-arguments
...
+      "hf_overrides": trainer_config.vllm_hf_overrides,
+      "enable_expert_parallel": sampler_config.enable_expert_parallel,
+      "enable_prefix_caching": rollout_prefix_caching_enabled(trainer_config),
+      # Ensures vLLM model initializes with correct dtype (not float32 default)
+      "dtype": trainer_config.weight_dtype.value,
+  }
+  if trainer_config.vllm_block_size is not None:
+    # Pin the KV-cache page size; left unset, the backend derives it from the
+    # engine shape, so unrelated engine changes can move it.
+    rollout_vllm_kwargs["block_size"] = trainer_config.vllm_block_size
```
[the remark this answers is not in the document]
gutianyu-google: Agreed, done in d0937c1f5: the check lives in `RLConfig`'s after-validator (next to the other cross-field constraints) so it fails at config time, with a unit test in `train_rl_test.py`.

tests/post_training/unit/tunix_adapter_test.py
```diff
@@ -116,6 +116,32 @@ def test_passes_through_explicit_segment_ids_unchanged(self):
 
     np.testing.assert_array_equal(np.asarray(self.base.captured["decoder_segment_ids"]), np.asarray(explicit_seg))
 
+  def test_segment_ids_alias_maps_to_decoder_segment_ids(self):
```
A9isha, line 122: These pass `segment_ids=` explicitly, so they pin the plumbing but not the gate — the reason the parameter must have this exact name. `_call_contains_by_type` also counts a `**kwargs` as a match, so a later refactor can keep these green while tunix stops forwarding the ids and every packed row silently reverts to whole-row attention: finite loss, no error. One line closes it:

```python
self.assertTrue(common.model_call_contains(TunixMaxTextAdapter, "segment_ids"))
```

tests/post_training/unit/tunix_adapter_test.py
```diff
@@ -116,6 +116,32 @@ def test_passes_through_explicit_segment_ids_unchanged(self):
...
+    synthesizing a pad mask."""
+    adapter = TunixMaxTextAdapter(base_model=self.base, pad_id=99)
+
+    input_tokens = jnp.array([[10, 11, 12, 20, 21]], dtype=jnp.int32)
+    positions = jnp.array([[0, 1, 2, 0, 1]], dtype=jnp.int32)
+    packed_seg = jnp.array([[1, 1, 1, 2, 2]], dtype=jnp.int32)
+
+    adapter(input_tokens, positions, None, None, segment_ids=packed_seg)
+
+    np.testing.assert_array_equal(np.asarray(self.base.captured["decoder_segment_ids"]), np.asarray(packed_seg))
```
A9isha, line 134: Both new tests assert into the stub's `captured` dict, so they show the kwarg is forwarded, not that packed rows are attention-isolated. The 15.2 → 6.7 s run is the real evidence, but it isn't in CI.

A packed-vs-unpacked check through a small real Transformer — pack two short sequences into one row, compare per-segment log-probs against running them separately — catches mask-polarity and boundary-leak regressions. I have that written and green; happy to send it as a follow-up rather than grow this PR.

src/maxtext/integration/tunix/tunix_adapter.py
```diff
@@ -109,10 +109,18 @@ def __call__(
...
   ) -> Tuple[Array, None]:
     """Forward compatible with Tunix Trainers default loss.
     Returns logits, None.
+
+    `segment_ids` is the name Tunix uses for packed-sequence segment ids: it
+    forwards them only to models whose call signature has a parameter of that
+    exact name. They are MaxText's `decoder_segment_ids`; when both are given,
+    `segment_ids` wins so packed rows keep per-sequence attention isolation.
     """
+    if segment_ids is not None:
```
A9isha, line 122: Worth a line in the description: this is unconditional, so it changes unpacked runs too. Once the gate passes, `tunix/rl/common.py:486-490` sends `segment_ids=input_seg_ids` — the non-pad mask from `process_ids` — on *every* batch, not just packed ones. So the adapter's own `(input_tokens != self._pad_id)` fallback just below stops firing on the tunix path, and authority for the pad mask moves from the adapter's `pad_id` to tunix's. Both are the tokenizer pad id today, so the values agree and I don't think anything breaks — but "defaults leave existing configs untouched" is true of the config knobs and not of this.

A9isha (approved):
Thank you!

tests/post_training/unit/tunix_adapter_test.py
gutianyu-google: Done, thanks!

tests/post_training/unit/tunix_adapter_test.py
gutianyu-google: Thanks, let's go with the follow up PR path as u suggested.

src/maxtext/integration/tunix/tunix_adapter.py
gutianyu-google: Done, thanks!

hengtaoguo (approved):

hengtaoguo (approved):

PR #5110, opened 2026-09-02
Track tunix API changes in the MaxText training engine

A9isha:
# Description

Three compat fixes, all needed to import and run `MaxTextTrainingEngine` against current tunix. Each one is an unconditional break today.

1. Freeze `RouterReplayTrainerPayload`.

Tunix declares `TrainerPayload` as
`@flax.struct.dataclass(frozen=True, kw_only=True)` (tunix/experimental/common/datatypes.py). Python refuses to derive a non-frozen dataclass from a frozen one, so importing maxtext_engine raises at class-creation time:

    TypeError: cannot inherit non-frozen dataclass from a frozen one

Match the base: `struct.dataclass` rather than `dataclasses.dataclass`, so the subclass is frozen too and stays registered as a pytree, which it has to be to cross the `jit` boundary in the engine's step functions.

tests/post_training/unit/router_replay_engine_test.py passes unchanged (7 tests), so nothing was relying on mutating a payload after construction.

2. Drop the `host_stage=` kwarg to `RaidenSynchronizer`.

It no longer exists, so constructing the synchronizer fails outright:

    TypeError: RaidenSynchronizer.__init__() got an unexpected keyword
    argument 'host_stage'

The kwarg used to pull proxy-backed (Pathways) params into client host memory before binding. The synchronizer now detects the proxy runtime itself and binds via the Raiden FFI without the host round-trip, so the `is_pathways` probe that fed it is dead too and goes with it.

3. Drop the `release_host_arrays()` call in `release_weight_sync`.

Same rewrite removed the method:

    AttributeError: 'RaidenSynchronizer' object has no attribute
    'release_host_arrays'

The synchronizer clears its staged references at the start of the next bind, so there is nothing left to release between rounds.

# Tests

Verified end to end on v5p-8 (4 chips) with the tunix `math_gsm8k_dist` GRPO demo, Qwen3-0.6B, Raiden weight sync between the MaxText trainer and a vLLM rollout. Source and destination checksums match across the sync: 310 tensors, 596049920 elements, grand total 13217973.204956055.

# Checklist

### DIFF
--- a/src/maxtext/training_engine/maxtext_engine.py
+++ b/src/maxtext/training_engine/maxtext_engine.py
@@ -27,6 +27,7 @@ from typing import Any
 
 from absl import logging
 from flax import nnx
+from flax import struct
 from flax.traverse_util import flatten_dict
 from flax.traverse_util import unflatten_dict
 import jax
@@ -145,7 +146,13 @@ _UNCOMPARABLE_STRUCTURE_HINT = (
 )
 
 
-@dataclasses.dataclass(kw_only=True)
+# Frozen to match tunix's `TrainerPayload`, which is a
+# `flax.struct.dataclass(frozen=True, kw_only=True)`. Python refuses to derive a
+# non-frozen dataclass from a frozen one, so a plain `dataclasses.dataclass`
+# here fails at import time with "cannot inherit non-frozen dataclass from a
+# frozen one". `struct.dataclass` also keeps the subclass registered as a
+# pytree, so it can still cross a `jit` boundary.
+@struct.dataclass(frozen=True, kw_only=True)
 class RouterReplayTrainerPayload(abstract_engine.TrainerPayload):
   """A TrainerPayload extension carrying forced router-replay expert decisions.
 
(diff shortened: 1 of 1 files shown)


src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1295,7 +1299,9 @@ def release_weight_sync(self, **kwargs: Any) -> Any:
     if self._raiden_syncs:
       for sync in self._raiden_syncs:
         logging.vlog(1, "Trainer Raiden metrics: %s", sync.metrics())
-        sync.release_host_arrays()
+        # `release_host_arrays()` used to drop the staged references here. The
```
SurbhiJainUSC, line 1302: Do we need this comment?

khatwanimohit, line 1302: nit: can we remove this comment, without that function now we don't need it at all?

khatwanimohit (approved):

PR #5112, opened 2026-09-02
Add ahead-of-time (XAOT) compilation for MaxTextTrainingEngine

NuojCheng:
# Description

Stacked on #5104 (merged into #5099's branch). One reviewable commit.

`trainers/pre_train/train_compile.py` lets a pre-training configuration be costed and memory-checked against a target topology from a host that does not own it. `MaxTextTrainingEngine` had no equivalent, so every iteration on a sharding or a batch size meant booking the hardware.

This adds `training_engine/maxtext_engine_compile.py`, which does the same for the engine's three kernels -- first forward/backward, accumulating forward/backward, update -- and reports each one's cost and memory. Nothing is materialized: weights and optimizer moments are `jax.ShapeDtypeStruct`s and the mesh is a topology description. It reuses the pre-train path's `get_topology_mesh` and `save_compiled`, so `compile_topology`, `compile_xla_flags` and `compiled_trainstep_file` keep the meanings they already have; the three executables are written to three files suffixed by kernel name.

```
python3 -m maxtext.training_engine.maxtext_engine_compile src/maxtext/configs/base.yml \
  model_name=qwen3-0.6b run_name=engine_aot_qwen3 \
  compile_topology=v6e-4 compile_topology_num_slices=1 \
  per_device_batch_size=4 max_target_length=2048 \
  ici_fsdp_parallelism=4 attention=flash enable_checkpointing=false
```

qwen3-0.6b for a v6e-4, 38s on a host with no v6e allocation:

| kernel | argument | output | alias | temp |
| `fwd_bwd` | 596 MB | 596 MB | 0 | 6.41 GB |
| `fwd_bwd_accum` | 1.19 GB | 596 MB | 596 MB | 6.59 GB |
| `update` | 2.39 GB | 1.79 GB | 1.79 GB | 4.7 MB |

Two supporting pieces in the engine:

- **`materialize_weights=False`** builds the train state abstractly. The moments cannot come from `nnx.Optimizer`, which allocates them with `zeros_like`, so the state is built under a trace instead: `nnx.eval_shape` for the module graph, and `jax.eval_shape` over an all-`Explicit` view of the mesh for the layouts, since that is where JAX carries a parameter's sharding into the moment allocated from it. Such an engine can be lowered but not run -- `fwd_bwd`, `update` and the checkpoint methods raise rather than return something plausible.
- **`lower()`** routes through the same `_compile_for_batch` the live engine calls on its first `fwd_bwd`, so the ahead-of-time path cannot drift from the live one by construction.

### Two bugs found on the way

**`_is_jax_dynamic` did not count `jax.ShapeDtypeStruct`.** Left alone, an AOT batch is classified *static*, closed over as a constant, and the kernel lowered with no batch argument at all -- a smaller HLO and a memory report missing every activation, with nothing anywhere saying so.

**`_update_kernel` was compiling twice per run.** `nnx.Optimizer` builds optax's `count` and its own `step` with `jnp.zeros` under no mesh, so they reach the first update uncommitted on device 0 and come back from it committed across the mesh. That is a second argument signature and a second full compile of the largest kernel in the engine, for a program that runs once. `_place_state_on_mesh` settles them up front (three scalar `device_put`s). A pre-existing startup cost independent of AOT, but it also decided whether an AOT report describes the steady state or only step one.

# Tests

New: `tests/post_training/unit/maxtext_engine_xaot_test.py`, 21 tests, on four simulated CPU devices plus a v6e-4 topology.

The HLO of all three kernels is asserted byte-identical between a live engine that has actually stepped and an abstract one -- StableHLO compared exactly, optimized HLO after the repo's usual source-location normalization -- across data parallelism, Zero-1, FSDP, `shard_mode=auto` and bfloat16 gradients. Plus:

- `test_the_comparison_can_fail` -- a byte-equality assertion is only as strong as its ability to tell two things apart, so this perturbs the model width and requires every comparison to notice. Width, not sequence length: `_update_kernel` never sees a sequence, so the obvious knob would leave one of the three kernels silently untested.
- `test_lowering_reproduces_the_kernels_training_ran` -- licenses the test rig itself, which re-lowers from recorded argument avals.
- `test_every_step_after_the_first_runs_the_same_kernels` -- pins the double-compile fix.
- qwen3-0.6b compiled for a v6e-4 topology, and `main` end to end writing one executable per kernel.

```
pytest tests/post_training/unit/maxtext_engine_xaot_test.py          # 21 passed, 80s
                                                                     # (26 with libtpu, for the v6e-4 topology tests)
pytest tests/post_training/unit/maxtext_engine_test.py \
       tests/post_training/unit/maxtext_engine_deferred_all_reduce_test.py \
       tests/post_training/unit/maxtext_engine_zero1_test.py \
       tests/post_training/unit/maxtext_engine_constructor_test.py   # 90 passed
```

No end-to-end workload: this change compiles but never executes a training step, and the engine's execution paths are covered by the four suites above.

# Checklist

### DIFF
--- a/src/maxtext/training_engine/maxtext_engine.py
+++ b/src/maxtext/training_engine/maxtext_engine.py
@@ -61,6 +61,10 @@ _MODEL_STATE_KEY = "model"
 # Where the same split puts the `nnx.Optimizer`, including the optax state Zero-1 shards.
 _OPTIMIZER_STATE_KEY = "optimizer"
 
+# The kernels an update is made of, in the order a step runs them. `compile_kernels()` is
+# keyed by these, and so is the ahead-of-time entry point in `maxtext_engine_compile`.
+KERNEL_NAMES = ("fwd_bwd", "fwd_bwd_accum", "update")
+
 _PURE_STATE_FALLBACK_WARNING = (
     "Cannot keep the train state as a pure pytree across steps (%s), so every fwd_bwd and "
     "update will re-walk the NNX module graph. That is correct but slow -- the two "
@@ -75,14 +79,16 @@ def _is_jax_dynamic(value: Any) -> bool:
   A `gen_model_input_fn` returns the loss function's keyword arguments, and only some of
   them are arrays. Tunix's GRPO adapter, for instance, returns a `TrainExample` alongside
   an `algo_config` object and integer `pad_id`/`eos_id`. The arrays must be traced; the
-  rest must be closed over, or `jax.jit` rejects the call outright.
+  rest must be closed over, or `jax.jit` rejects the call outright. `jax.ShapeDtypeStruct`
+  counts too: `compile_kernels()` drives the whole path on shapes alone, and an aval closed over as a
+  constant cannot be traced at all.
   """
   leaves = jax.tree.leaves(value)
   if not leaves:
     # An all-`None` subtree (e.g. an unset `ref_per_token_logps`) flattens to nothing. It
     # carries no data either way, so tracing it is harmless and keeps the treedef intact.
     return True
-  return any(isinstance(leaf, (jax.Array, np.ndarray, np.generic)) for leaf in leaves)
+  return any(isinstance(leaf, (jax.Array, jax.ShapeDtypeStruct, np.ndarray, np.generic)) for leaf in leaves)
 
 
 def _split_static_and_dynamic(batch: Any) -> tuple[Any, dict[str, Any]]:
@@ -348,6 +354,13 @@ def _normalize_loss_output(out: Any, has_aux: bool) -> abstract_engine.LossOutpu
   )
 
 
+def _to_aval(value: Any) -> Any:
+  """Returns `value` as a `jax.ShapeDtypeStruct` on the sharding it carries, or unchanged if it has no shape."""
+  if not hasattr(value, "shape") or not hasattr(value, "dtype"):
+    return value
+  return jax.ShapeDtypeStruct(value.shape, value.dtype, sharding=getattr(value, "sharding", None))
+
+
 @struct.dataclass(frozen=True, kw_only=True)
 class RouterReplayTrainerPayload(abstract_engine.TrainerPayload):
   """A TrainerPayload extension carrying forced router-replay expert decisions.
@@ -538,26 +551,15 @@ class MaxTextTrainingEngine(abstract_engine.AbstractTrainingEngine):
     # every eager caller compile or make a deferred compile never happen.
     self._compile_requested = False
     self._compiled_signature: Any = None
+    # `{kernel name: the jitted wrapper}` staged by the last `_compile_for_batch`; empty until
+    # then, and only ever read straight after one, since a recompile replaces every entry.
+    self._jitted_kernels: dict[str, Any] = {}
     self._compiled_eval: Any = None
     self._compiled_eval_signature: Any = None
     self._signature_compare_warned: bool = False
     if not training_config.model_name:
       raise ValueError("training_config.model_name must be specified")
-    model_or_model_mesh_pair = model_creation_utils.from_pretrained(
-        config=self._config,
-        mesh=self._mesh,
-        model_mode=common_types.MODEL_MODE_TRAIN,
-        rng_key=self._init_rng,
-        wrap_with_tunix_adapter=wrap_with_tunix_adapter,
-        tokenizer_pad_id=tokenizer_pad_id,
-    )
-    # `from_pretrained` returns `(model, mesh)` when it had to derive the mesh itself, and just the model
-    # when one was supplied. Adopt the derived mesh so `self._model` is always a module and `compile()` can
-    # still build shardings.
-    if self._mesh is None:
-      self._model, self._mesh = model_or_model_mesh_pair
-    else:
-      self._model = model_or_model_mesh_pair
+    self._model = self._build_model(wrap_with_tunix_adapter, tokenizer_pad_id)
     self._state: Any = None
     # Pure-pytree mirror of the model and train state, carried across steps so the step path
     # never re-walks the module graph. `None` means "not cached".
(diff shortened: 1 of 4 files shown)


src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -295,6 +300,50 @@ def _conform_accumulator(value: Any, target: jax.sharding.NamedSharding) -> Any:
...
+def _to_aval(value: Any) -> Any:
+  """Returns `value` as a `jax.ShapeDtypeStruct`, keeping whatever sharding it carries.
+
+  `jax.jit(...).lower()` accepts arrays and avals interchangeably, so `lower` converts
+  everything first: the ahead-of-time path then hands the kernels exactly the objects the
+  live path does, and the two cannot diverge on the array-vs-aval distinction.
+  """
+  if value is None:
+    return None
+  return jax.ShapeDtypeStruct(value.shape, value.dtype, sharding=getattr(value, "sharding", None))
```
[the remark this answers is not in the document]
NuojCheng: Already handled, in a revision after the one you reviewed. At 273b078 `_to_aval` guarded only `value is None`; it now guards on the attributes themselves, which subsumes the `None` case:

```python
if not hasattr(value, "shape") or not hasattr(value, "dtype"):
  return value
```

Same shape as the guard in `_place_state_on_mesh` you point at, so no change needed here.

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -295,6 +300,50 @@ def _conform_accumulator(value: Any, target: jax.sharding.NamedSharding) -> Any:
...
+  """Returns `aval` with its sharding spec re-expressed on `mesh`.
+
+  Propagation hands back `NamedSharding`s on whichever mesh was active during the trace --
+  an `AbstractMesh`, or the explicit view above. `_mesh_sharding` compares meshes by
+  equality to decide whether a leaf belongs to this engine's, so a spec that is right but
+  homed elsewhere would be silently discarded and replaced by a replicated one.
+  """
+  spec = getattr(getattr(aval, "sharding", None), "spec", None)
+  target = jax.sharding.NamedSharding(mesh, spec) if spec is not None else None
+  return jax.ShapeDtypeStruct(aval.shape, aval.dtype, sharding=target)
```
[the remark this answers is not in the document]
NuojCheng: Taken, in ecdcf56 — but as consistency rather than as a crash fix, and it is worth saying which, since the two helpers are not equally exposed.

I instrumented `_rehome_aval` and recorded the type of every leaf it is handed, across `adamw`, `adam_pax`, `sgd` and Zero-1:

```
adamw/dp: {'ShapeDtypeStruct': 159}
zero1:    {'ShapeDtypeStruct': 159}
sgd:      {'ShapeDtypeStruct': 116}
adam_pax: {'ShapeDtypeStruct': 158}
```

No non-array leaf, and one of the two call sites cannot produce one by construction: `state_pure` is the output of `jax.eval_shape`, which returns `ShapeDtypeStruct`s and nothing else. The other, `model_pure`, comes from `create_nnx_abstract_model`, which is itself built out of avals. So the `AttributeError` is not reachable on any path I can construct today — optax's `count` and the optimizer's `step`, the scalars this would most plausibly be about, are `jnp.zeros([], int32)` and so are arrays.

Still worth adding. `_to_aval` and `_place_state_on_mesh` both guard, and `_rehome_aval` being the one that does not is the kind of asymmetry that reads as an oversight later. Returning the leaf untouched is also the right answer semantically: something with no shape has no sharding to re-home.

src/maxtext/training_engine/maxtext_engine_compile.py
```diff
@@ -0,0 +1,312 @@
...
+from maxtext.common import train_state_nnx
+from maxtext.configs import pyconfig
+from maxtext.trainers.pre_train import train_compile as pre_train_compile
+from maxtext.training_engine import maxtext_engine
+from maxtext.utils import gcs_utils
+from maxtext.utils import max_utils
+from maxtext.utils import maxtext_utils
+from maxtext.utils import model_creation_utils
+
+KERNEL_NAMES = ("fwd_bwd", "fwd_bwd_accum", "update")
```
gobbleturk: can we give users the choice of which to compile, maybe just one at a time? compilation can be slow, a user might only be interested in one of these, not all of them

gobbleturk (approved):

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1359,6 +1421,50 @@ def compile(self, dummy_data: abstract_engine.TrainerPayload) -> None:
     dynamic_batch, static_batch = _split_static_and_dynamic(self._prepare_batch(dummy_data))
     self._compile_for_batch(dynamic_batch, static_batch)
```
igorts-git: please add actual pre-compilation here.

src/maxtext/training_engine/maxtext_engine.py
igorts-git, line 1407: also compile here?

igorts-git (approved):

src/maxtext/training_engine/maxtext_engine_compile.py
NuojCheng: I am hesitant on adding more additional maxtext flags for this feature... Let's do that in followup PRs.

NuojCheng: Unless there is a way to avoid adding maxtext flags for this.

src/maxtext/training_engine/maxtext_engine.py
NuojCheng: add

src/maxtext/training_engine/maxtext_engine.py
NuojCheng: added compile eval

PR #5128, opened 2026-09-03
feat(kda): integrate KDA attention with tokamax backend and CP support

chiaotung97:
# Description

This PR integrates **KDA (Kimi Delta Attention, a recurrent linear attention mechanism)** into MaxText end to end: the attention layer, the tokamax-backed Pallas TPU kernel, context parallelism (CP), and **decoder wiring via `attention_type='kda'`**. KDA updates its recurrent state with the Delta Rule:

```
S' = S * exp(g_t)
residual = v_t - k_t^T @ S'
S = S' + beta_t * k_t ⊗ residual
o_t = scale * q_t^T @ S
```

The layer follows the Megatron KDA reference and delegates kernel execution to tokamax's Pallas TPU implementation, keeping MaxText free of low-level kernel code.

## Key Changes

| File | Description |
|---|---|
| `src/maxtext/layers/attention_kda.py` | New `KimiDeltaAttention` layer and `ShortConvolution`: QKV/beta/gate/output-gate projections, depthwise causal 1D convolution, SiLU activation, always-on QK L2 normalization, per-head RMSNorm + output gate, gate parameters `A_log` / `dt_bias` (matching the Megatron reference); also hosts `halo_exchange_for_conv` for CP |
| `src/maxtext/kernels/kda/` | New `chunk_kda` entry point + tokamax adapter: `[B,T,H,D]` ↔ `[H,B,T,D]` layout translation, lazy import of the KDA API, explicit `mosaic` implementation selection inside `shard_map` |
| `src/maxtext/common/common_types.py` | `AttentionType.KDA` |
| `src/maxtext/configs/types.py` | New `KdaAttention` config (`linear_conv_kernel_dim`, `use_kda_safe_gate`, `kda_lower_bound`, reserved `use_kda_lora`) + validators; `attention_type` Literal gains `"kda"`; `attention_type='kda'` with `scan_layers=true` is rejected at config time |
| `src/maxtext/configs/base.yml` | KDA flags registered; `attention_type` supported list |
| `src/maxtext/layers/nnx_decoders.py` | `NNXDecoderLayer` gains a per-layer `attention_type` override and a KDA branch: builds `KimiDeltaAttention` (no KV cache — recurrent state is carried by the kernel); non-KDA types are untouched |
| `tests/unit/kda_attention_test.py` | Layer/kernel/CP suite (see below) |
| `tests/unit/kda_decoder_integration_test.py` | Decoder-integration suite: config guards + a real-decoder training run |
| `docs/reference/kda_cp_support.md` | Design doc, linked into the Reference toctree |

## Using KDA

Select it like any other attention variant — this mirrors how `attention_type='mla'` is consumed:

```yaml
attention_type: kda
scan_layers: false        # required (KDA layers are not validated in scanned stacks yet)
use_kda_safe_gate: true   # recommended: bounded-decay gate, keeps the recurrence stable
kda_lower_bound: -5.0     # common choice, required in [-5, 0) when the safe gate is on
```

Notes:
- KDA always L2-normalizes Q/K — this is part of the KDA architecture (the Delta-Rule recurrence diverges with unbounded q/k), independent of the shared `use_qk_norm` flag which belongs to dot-product attention.
- `context_parallel_load_balance` is rejected with KDA+CP (token order is load-bearing for the recurrent state).
- KDA has no KV cache; autoregressive decoding is not implemented yet (raises a clear `NotImplementedError`). Train/prefill paths are supported.

## Context Parallelism Support

- The CP mesh axis is taken from `cfg.context_sharding` (default `"context"`; `"expert"` works for expert-as-context) and is threaded consistently through the conv halo exchange, the T-axis partition-spec injection, and `ContextParallelMetadata`
- Under CP without user segments, an all-ones segment tensor is synthesized outside `shard_map` so the kernel can always derive per-rank `cu_seqlens` / chain fields
- `ShortConvolution` pulls `kernel_size-1` left-context tokens from the previous CP rank via `ppermute` inside a dedicated `shard_map`; a clear `ValueError` guards the `kernel_size-1 > T_local` case (multi-rank receptive fields are not implemented)

## Tests and Validation

**Unit tests — 35 TPU tests + 23 CPU tests, all passing** (4×TPU v6e, 2026-09-04):

- Layer/kernel: forward/backward (activations + weights), determinism, padding, packed-segment isolation (cross-row and within-row), precision vs a pure-XLA recurrent reference (FP32/BF16, ULP-based fallback)
- CP: kernel-level equivalence parametrized CP=2/CP=4, CP backward, full-layer CP without segments (dummy synthesis path), **full-layer CP with real packed segments** (segment spanning the rank boundary + boundary exactly at the split; fwd + input/weight grads vs non-CP), oversized-halo rejection, load-balance rejection, CP metadata-missing refusal
- Composition: full-layer **Mosaic vs tokamax XLA reference parity** (same weights, fwd + grads)
- Kernel guards: `initial_state` / `output_final_state` rejected with clear errors
- Markers: `tpu_only` is applied per-test, so the pure config/pure-op tests also run in regular CPU CI

**Decoder integration** (`tests/unit/kda_decoder_integration_test.py`):

- CPU-runnable: config acceptance, `scan_layers` guard, dispatch of `NNXDecoderLayer` to `KimiDeltaAttention` (and unchanged default for other types)
- TPU: a real-decoder training run (`decoder_block=default`, 2 layers, delayed-copy task — i.i.d. tokens with `t[i] = t[i-delay]`, `delay=5` beyond the conv receptive field, so only the recurrent state can solve it; masked loss): **loss collapses from 5.28 to < 1.0 in 400 steps** through the actual decoder

### Reproducing

On a TPU host (Python ≥ 3.12; verified on 4×TPU v6e):

```bash
# 1. TPU-capable JAX (validation used JAX 0.11.0 + libtpu 0.0.44.1)
pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html

# 2. MaxText with TPU dependencies
pip install -e <maxtext checkout>
pip install -r <maxtext checkout>/src/dependencies/requirements/generated_requirements/tpu-requirements.txt

# 3. tokamax with the KDA kernels — the KDA API has landed on openxla/tokamax
#    main but no release contains it yet; install from source:
git clone https://github.com/openxla/tokamax.git && pip install -e tokamax
#    (or the equivalent branch validated here: -b antgroup/kda-pallas-kernel, tip 939da5c)

# 4. Tests (kernel/CP tests are tpu_only and skip cleanly elsewhere)
pytest tests/unit/kda_attention_test.py tests/unit/kda_decoder_integration_test.py -v
```

## Dependencies

- The KDA kernels come from [openxla/tokamax#1103](https://github.com/openxla/tokamax/pull/1103) (Kimi Delta Attention Pallas kernels, by @Fred33146). **Status**: the code has landed on openxla/tokamax `main` (signature-compatible with this PR's adapter, verified), but **no public tokamax release contains it yet**. MaxText's pin therefore cannot express the dependency; the adapter imports the KDA API lazily, so installs without it are unaffected until KDA is actually used (clear `ImportError` with instructions at that point). Once the first release ships, the `tokamax>=` pin and derived requirement files will be bumped.
- Validated with JAX 0.11.0 + libtpu 0.0.44.1

## Hardware / Shape Constraints (mosaic kernel)

The adapter selects the `"mosaic"` Pallas implementation explicitly (no silent fallback). Constraints — surfaced as clear `NotImplementedError`s from tokamax at bind time:

- TPU generation ≥ 6 (validated on v6e)
- Key dimension ≤ 256
- Under CP: key and value head dims must be multiples of 128
- Sequence length padded internally to a multiple of chunk size 64

## Known Limitations / Follow-ups

- Autoregressive/inference path not implemented (train/prefill only; AR raises a clear error)
- `initial_state` / `output_final_state` not yet supported (explicit `NotImplementedError`)
- `scan_layers` must be `false` with `attention_type='kda'` (scanned KDA stacks not validated yet)
- bf16 training-recipe tuning for KDA models (the integration test validates in fp32 with the safe gate)
- Hybrid per-layer KDA + softmax models (e.g. Ling-style interleaving) and a production KDA model config
- tokamax version pin bump once the first public release with the KDA API ships

# Checklist


scripts/dev/kda_e2e_smoke.py
```diff
@@ -0,0 +1,235 @@
...
+    parser.error(
+        f"--delay must exceed the short-conv receptive field "
+        f"(linear_conv_kernel_dim={cfg.linear_conv_kernel_dim}), got {args.delay}; "
+        "otherwise the target can be copied by the convolution without KDA state"
+    )
+  devices = jax.devices()
+  print(f"devices: {devices}")
+  mesh = jax.sharding.Mesh(np.array(devices), ("x",))
+  rngs = nnx.Rngs(0)
+  with mesh:
```
NuojCheng: with jax.set_mesh(mesh):

with mesh is deprecated

scripts/dev/kda_e2e_smoke.py
```diff
@@ -0,0 +1,235 @@
+# Copyright 2026 Ant Group. All Rights Reserved.
```
NuojCheng: could you either update this as a test, or move it to src/maxtext/examples/.

src/maxtext/configs/types.py
```diff
@@ -746,6 +746,81 @@ class CompressedAttention(BaseModel):
   )
 
 
+class KdaAttention(BaseModel):
```
NuojCheng, line 749: please also add new flags in configs/base.yml

src/maxtext/utils/cp_utils.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2026 Ant Group. All Rights Reserved.
```
NuojCheng: if this util is not for kda, I suggest merging it into attention_kda.py

NuojCheng (commented):
Thank you for contribution! Some general architecture comments before diving deep

chiaotung97:
> Thank you for contribution! Some general architecture comments before diving deep

Thank you for detailed comments. All four addressed in the latest push: 
(1) dropped the deprecated with mesh: the mesh is passed explicitly and no ambient context is needed.
(2) the smoke script is now a proper test `TestKdaE2eSmoke::test_delayed_copy_loss_collapses`.
(3) the KDA flags are registered in `configs/base.yml`.
(4) `halo_exchange_for_conv` is KDA-specific, so it's merged into `attention_kda.py`. Details in the inline replies.

scripts/dev/kda_e2e_smoke.py
chiaotung97: Removed entirely rather than migrating to jax.set_mesh: the mesh is passed explicitly to the layer/shard_map.
It forces Manual axis types that clash with the layer's explicit Auto-typed shard maps. Consistent with upstream CP tests, which also use no ambient mesh context.

scripts/dev/kda_e2e_smoke.py
chiaotung97: Converted to a test: `TestKdaE2eSmoke::test_delayed_copy_loss_collapses` (tpu_only, ~1 min on v6e). It keeps the history-dependent delayed-copy task (delay chosen to exceed the conv receptive field so only the KDA recurrent state can solve it) and asserts the loss collapses. The `scripts/dev/` script is removed.

src/maxtext/utils/cp_utils.py
chiaotung97: Yes, it's KDA-specific, so it is good to merge into `attention_kda.py` and removed `utils/cp_utils.py`, Thanks.

chiaotung97:
> ## [Codecov](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer) Report
> ❌ Patch coverage is `77.77778%` with `58 lines` in your changes missing coverage. Please review.
> 
> [Files with missing lines](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer)	Patch %	Lines
> [src/maxtext/layers/attention_kda.py](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer#diff-c3JjL21heHRleHQvbGF5ZXJzL2F0dGVudGlvbl9rZGEucHk=)	77.88%	[37 Missing and 9 partials ⚠️ ](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer)
> [src/maxtext/kernels/kda/tokamax.py](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer#diff-c3JjL21heHRleHQva2VybmVscy9rZGEvdG9rYW1heC5weQ==)	50.00%	[7 Missing and 1 partial ⚠️ ](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer)
> [src/maxtext/kernels/kda/__init__.py](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer#diff-c3JjL21heHRleHQva2VybmVscy9rZGEvX19pbml0X18ucHk=)	77.77%	[1 Missing and 1 partial ⚠️ ](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer)
> [src/maxtext/utils/cp_utils.py](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer#diff-c3JjL21heHRleHQvdXRpbHMvY3BfdXRpbHMucHk=)	92.85%	[1 Missing and 1 partial ⚠️ ](https://app.codecov.io/gh/AI-Hypercomputer/maxtext/pull/5128?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AI-Hypercomputer)
> 📢 Thoughts on this report? [Let us know!](https://github.com/codecov/feedback/issues/255)

The patch number reflects what CI *can* measure, not what the code covers: every KDA kernel-invoking path is exercised by `tpu_only` tests, additionally gated by `skipif(KDA API available)`. CI installs the public tokamax release, which does **not** contain the KDA module yet (the kernels landed on openxla/tokamax `main` but no release has been cut — see the Dependencies section), so those lines are correctly *skipped* in CI rather than executed.

Running the full suite on an actual TPU host (4×TPU v6e, both TPU mode and CPU mode) measures **99% line coverage on the new KDA source files**:

| File | Coverage |
|---|---|
| `src/maxtext/kernels/kda/__init__.py` | 100% |
| `src/maxtext/kernels/kda/tokamax.py` | 100% |
| `src/maxtext/layers/attention_kda.py` (layer + CP halo exchange) | 99% |

The only uncovered lines are the `except ImportError` fallback for tokamax-less installs, which is unreachable by construction whenever the KDA API is present. The decoder-wiring path (`attention_type='kda'`) is covered by `tests/unit/kda_decoder_integration_test.py` (CPU guards on every runner + a real-decoder training run on TPU).

Once tokamax ships the first release containing the KDA API and CI can install it, the kernel paths will be exercised in CI as well and the patch number will reflect them.

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,672 @@
...
+
+      kda_args = (q, k, v, g, beta, self.A_log.value, delta_time_bias_2d)
+      if has_seg:
+        if decoder_segment_ids is not None:
+          kda_args = kda_args + (decoder_segment_ids,)
+        else:
+          # CP without varlen: pass None; the shard_map function
+          # synthesises a dummy seg internally.
+          kda_args = kda_args + (None,)
+      o = _shard_map_chunk_kda(*kda_args)
```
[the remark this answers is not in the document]
chiaotung97: `has_seg` is now simply `decoder_segment_ids is not None` (CP without user segments synthesizes the tensor earlier), so `kda_args` never appends `None`.

src/maxtext/configs/types.py
chiaotung97: Done, `linear_conv_kernel_dim`, `use_kda_safe_gate`, `kda_lower_bound`, `use_kda_lora` are registered in `configs/base.yml` next to the other attention options, matching the `KdaAttention` field names/defaults in `types.py`.

RissyRan:
I see @NuojCheng is OOO. @Rohan-Bierneni and @dandragona Could you help review and assist customer to merge? Thank you!

src/maxtext/configs/base.yml
```diff
@@ -409,10 +408,16 @@ param_scan_axis: 1
...
+attention_type: 'global' # Supported attention_type: global, local_sliding, chunk, mla, kda, full, compressed, block_diffusion
 share_kv_projections: false # Note: Not compatible with attention_type='mla'
 attention_bias: false # If true, adds a learnable bias to the query, key, and value projections
 attention_sink: false
+# KDA (Kimi Delta Attention) options, consumed by the KimiDeltaAttention layer
+# selected via attention_type: 'kda'.
+linear_conv_kernel_dim: 4 # Kernel size of the causal depthwise short convolution applied to Q/K/V in KDA; 0 disables it
+use_kda_safe_gate: false # If true, uses the sigmoid lower-bound ("safe") gate path of the KDA kernel
+kda_lower_bound: 0.0 # Lower bound of the KDA sigmoid gate; only used when use_kda_safe_gate=true (must be in [-5, 0); -5.0 is a common choice)
+use_kda_lora: false # Reserved: the KDA LoRA path is not implemented and must stay false
```
dandragona, line 420: Is this necessary if it is not implemented?

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,747 @@
...
+from maxtext.kernels.kda import chunk_kda
+
+# KDA depends on tokamax at runtime, but import should succeed because
+# tokamax is a mandatory dependency for KDA models.
+try:
+  from tokamax._src.ops.experimental.kda.cp_utils import (
+      ContextParallelMetadata as TokamaxContextParallelMetadata,
+  )
+except ImportError:
+  TokamaxContextParallelMetadata = None
```
dandragona, line 47: Do we expect a situation where ImportError will be raised?

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,747 @@
...
+
+def halo_exchange_for_conv(
+    x: jax.Array,
+    halo_size: int,
+    axis_name: str = "context",
+    seq_axis: int = 1,
+) -> jax.Array:
+  """Prepend ``halo_size`` tokens from the previous CP rank for causal conv.
+
+  KDA's ``ShortConvolution`` is the only user today; the helper lives in this
```
dandragona, line 83: This function description is hard to follow with a lot of jargon. IIUC this is just dealing with the convolution window difficulties when dealing with a sharded context axis?

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,747 @@
...
+      offset = self.kernel_size - 1 - k
+      x_slice = x_padded[:, offset : offset + T, :]
+      if segment_ids is not None:
+        x_slice = x_slice * masks[k]
+      output = output + x_slice * self.kernel[k]
+
+    return output.astype(self.dtype)
+
+
+class KimiDeltaAttention(nnx.Module):
```
dandragona, line 207: Is this mostly a 1:1 port from Tokamax? If so can we include that in the function description with a list of the changes (if any) we have added that deviate from the Tokamax version?

dandragona (changes requested):
Thank you for the CL! It looks good to me, but want to clean some minor things and explain some stuff a bit more

src/maxtext/layers/attention_kda.py
chiaotung97: Thanks, I have rewrote the docstring to lead with the problem instead of the mechanism.

src/maxtext/layers/attention_kda.py
chiaotung97: Not a port of tokamax.
Tokamax provides only the op. Its base.KimiDeltaAttention is the pure-JAX reference implementation of the recurrence (q/k/v/g/beta → o), while this class is a full transformer layer around it: projections, short conv, gating, normalization and sharding are all MaxText's. The only thing we take from tokamax is api.kimi_delta_attention  (implementation="mosaic", Pallas/TPU) plus ContextParallelMetadata for cross-rank recurrent state, note tokamax explicitly leaves conv-side CP to the caller, so halo_exchange_for_conv is ours.

src/maxtext/layers/attention_kda.py
chiaotung97: Yes, it's reachable now. The first tokamax release containing `tokamax._src.ops.experimental.kda` is **0.0.14**. MaxText's locks say `tokamax>=0.0.13` and CI installs with `--resolution=lowest`, so CI lands on 0.0.13, the decoupled job pins `==0.0.12` deliberately.

And `layers/nnx_decoders.py` imports this module unconditionally, so a bare import would break every NNX model on those environments. I've rewritten the comment to say exactly that.

src/maxtext/configs/base.yml
chiaotung97: Thanks, "Reserved" didn't convey much, so the `base.yml` comment, field description and layer docstring now say what the flag actually selects.

It does exist upstream (fla builds both gate projections as a two-stage `nn.Sequential`; sglang splits them into `f_a_proj`/`f_b_proj`). We implement the full-rank variant because that's what the published Ling-3.0-flash checkpoint ships:  [inclusionAI/Ling-3.0-flash](https://huggingface.co/inclusionAI/Ling-3.0-flash) sets `no_kda_lora: true`, and its `attention.f_proj.weight` is a single `[4096, 2560]` tensor rather than a bottleneck pair, so this variant can be checked against a public checkpoint, while the low-rank one has no reference in MaxText to validate against.

I kept the field because it fails loudly. The validator rejects `true` with an explanation, rather than ignoring it.

src/maxtext/layers/attention_kda.py
dandragona: tysm!

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,828 @@
...
+    # KDA Delta Rule relies on sequential recurrent state S_t = f(S_{t-1}, ...).
+    # load_balance's DUAL_CHUNK_SWAP reorder breaks token order, invalidating
+    # the sequential dependency. Reject this combination up front.
+    if cp_size > 1 and getattr(cfg, "context_parallel_load_balance", False):
+      raise ValueError(
+          "KDA CP does not support context_parallel_load_balance. "
+          "Recurrent state S depends on exact token order; DUAL_CHUNK_SWAP "
+          "reorder breaks the sequential dependency. Set "
+          "context_parallel_load_balance=false when using KDA with CP."
+      )
```
Rohan-Bierneni, line 567: Could we also mirror the context_parallel_load_balance check in MaxTextConfig.validate_all() in src/maxtext/configs/types.py?

Right now it's guarded in KimiDeltaAttention.__call__, but because context_parallel_load_balance: true is the default in base.yml, jobs running multi-node CP with KDA won't fail until the first forward step (after all device allocation and model weight initialization). Validating it early in types.py ensures immediate feedback at config parse time.

Example for GDN: https://github.com/AI-Hypercomputer/maxtext/blob/176b82712f4030f7ae0f41581cc159d9050acfe0/src/maxtext/configs/types.py#L4824

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,828 @@
...
+    # `max_segments_per_seq`, which defaults to -1 (unset) — fail fast with
+    # a config-level message instead of erroring deep in kernel binding.
+    if decoder_segment_ids is not None and cfg.max_segments_per_seq <= 0:
+      raise ValueError(
+          "KDA with packed sequences (decoder_segment_ids) requires "
+          f"`max_segments_per_seq` to be a positive integer, got "
+          f"{cfg.max_segments_per_seq}. Set `max_segments_per_seq` in your "
+          "config to a static upper bound on the number of packed segments "
+          "per sequence."
+      )
```
Rohan-Bierneni, line 584: Also here add a pyconfig check when packing=True and attention_type='kda'?

max_segments_per_seq defaults to -1 in base.yml. While the runtime check in attention_kda.py is great defense-in-depth, checking self.packing and self.attention_type == 'kda' and self.max_segments_per_seq <= 0 in MaxTextConfig.validate_all() will catch unconfigured packed runs before dataset pipeline creation or weight loading.

Example: https://github.com/AI-Hypercomputer/maxtext/blob/176b82712f4030f7ae0f41581cc159d9050acfe0/src/maxtext/configs/types.py#L4777

src/maxtext/kernels/kda/tokamax.py
```diff
@@ -0,0 +1,220 @@
...
+      beta=beta_h,
+      a_log=a_log,
+      delta_time_bias=delta_time_bias,
+      scale=scale,
+      segment_ids=segment_ids,
+      use_gate_in_kernel=use_gate_in_kernel,
+      use_qk_l2norm=use_qk_l2norm,
+      lower_bound=lower_bound,
+      max_num_segments=max_num_segments,
+      implementation="mosaic",  # tokamax's Pallas Mosaic TPU kernel ("mosaic" resolves to the mosaic_tpu impl)
```
Rohan-Bierneni: Is there a reason we hardcode "mosaic" and do not allow "xla" option? 

Tokamax also provides implementation="xla" (the reference implementation tested in test_full_layer_mosaic_vs_xla_parity). Could we make implementation configurable (e.g. defaulting to "mosaic", but allowing "xla" via a config flag or mapping attention: dot_product -> xla)?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -109,34 +117,45 @@ def __init__(
...
-        model_mode=model_mode,
-        rngs=rngs,
-    )
+    if self.attention_type == AttentionType.KDA:
+      # KDA is a self-contained recurrent attention layer (its own QKV/conv/
+      # gate/beta stack + tokamax Delta-Rule kernel); it carries recurrent
+      # state instead of a KV cache.
+      self.self_attention = attention_kda.KimiDeltaAttention(
+          config=self.config,
+          layer_idx=0,
```
Rohan-Bierneni: layer_idx=0 is hardcoded when instantiating KimiDeltaAttention. While KimiDeltaAttention currently doesn't use self.layer_idx, all layers in a multi-layer stack end up tagged as layer 0.

Could we accept layer_idx: int = 0 in NNXDecoderLayer.__init__ and forward lyr from _init_sequential_generic (similar to how QWEN3_5 and DEEPSEEK4 do it)? Or if it's intentionally unused, default layer_idx: int = 0 in KimiDeltaAttention.__init__

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,828 @@
...
+
+    B, T_orig, _ = hidden_states.shape
+    T = T_orig
+
+    if T % _KDA_CHUNK_SIZE != 0:
+      pad_len = _KDA_CHUNK_SIZE - (T % _KDA_CHUNK_SIZE)
+      hidden_states = jnp.pad(hidden_states, ((0, 0), (0, pad_len), (0, 0)))
+      if decoder_segment_ids is not None:
+        decoder_segment_ids = jnp.pad(decoder_segment_ids, ((0, 0), (0, pad_len)), constant_values=0)
+      T = hidden_states.shape[1]
```
Rohan-Bierneni, line 607: Question on an edge case: In attention_kda.py, the sequence padding T % _KDA_CHUNK_SIZE != 0 operates on T_global.

If T_global = 128 with cp_size = 4, T_global is divisible by 64 (so no padding occurs), but the local slice per rank is T_local = 32. Inside Tokamax, check_inputs_support enforces chunk size 64. It currently passes under CP only because line 680 synthesizes dummy segment_ids, triggering Tokamax's _align_seqs to internally pad segments to 64.

Should we enforce in types.py that max_target_length % (64 * cp_size) == 0, or pad to _KDA_CHUNK_SIZE * cp_size globally in attention_kda.py?


Example: https://github.com/AI-Hypercomputer/maxtext/blob/176b82712f4030f7ae0f41581cc159d9050acfe0/src/maxtext/configs/types.py#L3569

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,828 @@
...
+  pad_width = [(0, 0)] * x.ndim
+  pad_width[seq_axis] = (halo_size, 0)
+  zero_padded = jnp.pad(x, pad_width)
+
+  if not _has_named_axis(axis_name):
+    return zero_padded
+
+  cp_size = jax.lax.psum(1, axis_name=axis_name)
+  if cp_size == 1:
+    return zero_padded
```
Rohan-Bierneni: In halo_exchange_for_conv, zero_padded = jnp.pad(x, pad_width) is constructed unconditionally before checking _has_named_axis(axis_name) and cp_size == 1.

When running under multi-rank CP (cp_size > 1), this padded array is immediately discarded in favor of jnp.concatenate([halo, x], axis=seq_axis)

This should be:

```
if halo_size <= 0:
    return x
def _zero_pad():
  pad_width = [(0, 0)] * x.ndim
  pad_width[seq_axis] = (halo_size, 0)
  return jnp.pad(x, pad_width)
if not _has_named_axis(axis_name):
  return _zero_pad()
cp_size = jax.lax.psum(1, axis_name=axis_name)
if cp_size == 1:
  return _zero_pad()
```

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,828 @@
...
+        rngs=rngs,
+    )
+
+    # Gate parameters, named and initialized as in the KDA reference
+    # implementation. Params keep the reference names (A_log / dt_bias) but
+    # are passed to tokamax as `a_log` and `delta_time_bias` inside the
+    # shard_map below.
+    # A_log: [num_key_heads] — log of the diagonal decay matrix.
+    # Initialized as log(U(1, 16)), the reference implementation's default.
+    A_init_range = (1.0, 16.0)
```
Rohan-Bierneni: Question on initialization: In the pytorch ref [fla/layers/kda.py](https://github.com/fla-org/flash-linear-attention/blob/e52dbc0ea19d3a40d7ab7f9eed855d2b473994d2/fla/layers/kda.py#L175), when safe_gate=True, A_log is initialized to zeros (torch.zeros(...), so A=exp(0)=1.0), whereas in standard softplus mode it is initialized to log(U(1,16)).

In Tokamax's _activate_gate, the decay is computed as lower_bound * sigmoid(exp(a_log) * g_f). Initializing A_log to log(U(1,16)) scales the gate input by up to 16× at step 0, sharpening the sigmoid.

Was this intentional for the MaxText recipe, or should self.A_log initialize to zeros when use_kda_safe_gate=True to match the reference?

src/maxtext/layers/nnx_decoders.py
```diff
@@ -194,16 +213,26 @@ def __call__(
...
-        decoder_positions,
-        decoder_segment_ids=decoder_segment_ids,
-        deterministic=deterministic,
-        model_mode=model_mode,
-        kv_cache=kv_cache,
-        attention_metadata=attention_metadata,
-    )
+    if self.attention_type == AttentionType.KDA:
+      # KDA has no KV cache; its recurrent state is carried by the kernel.
+      attention_lnx, _ = self.self_attention(
```
Rohan-Bierneni, line 220: I think we can explicitly set kv_cache=None after the attention intialization here for kda.

Rohan-Bierneni (commented):
Thanks for the changes. I have left a few comments PTAL

dandragona (approved):

src/maxtext/layers/nnx_decoders.py
chiaotung97: Done, the KDA branch now sets `kv_cache = None` explicitly rather than relying on the incoming value passing through.

src/maxtext/layers/attention_kda.py
chiaotung97: Thanks, this was a real divergence from the reference, changed. `A_log` is now initialized per gate path: zeros (A = 1) under the safe gate, `log(U(1, 16))` for the softplus gate, with the reasoning in a comment (`lower_bound` already bounds the decay in the safe-gate path, so A only sharpens the sigmoid there).

Since this changes the init, I re-ran the end-to-end validation: the delayed-copy decoder training still collapses the loss on 4×v6e. `test_a_log_init_follows_gate_path` pins the rule.

src/maxtext/layers/attention_kda.py
chiaotung97: Done as suggested, `_zero_pad()` is now built lazily, so the CP path doesn't construct an array it immediately discards.

src/maxtext/layers/attention_kda.py
chiaotung97: I traced this in tokamax 0.0.14, your reading of the mechanism is exactlyright, but I'd leave the padding as it is:

- `check_inputs_support` requires `seq_len % chunk_size == 0` **only when `segment_ids is None`** (`pallas_mosaic_tpu.py:165`). With segment_ids it takes the varlen path, where `_align_seqs(..., align=chunk_size)` (`utils.py:155`) pads each segment to a chunk multiple.
- Under CP we always have segment_ids — synthesized when the user supplies none, so `T_local < 64` is handled. Your example is already covered: `test_kda_cp_equivalence[cp_size=4]` runs T_global=128, i.e. T_local=32, and passes on 4×v6e.

So `max_target_length % (64 * cp_size) == 0` would reject a configuration that works and is tested today, and padding globally to `64 * cp_size` would double the token count in that same example. What I did instead: the synthesis site now documents that it's load-bearing for this reason, so dropping it later is a visible decision rather than a silent breakage.

src/maxtext/layers/nnx_decoders.py
chiaotung97: Good catch. `NNXDecoderLayer` now accepts `layer_idx: int = 0` and `_init_sequential_generic` forwards the real `lyr` for KDA layers, the same pattern as QWEN3_5 / DEEPSEEK4, so a multi-layer KDA stack no longer tags every layer as 0.

src/maxtext/layers/attention_kda.py
chiaotung97: Done, added next to the GatedDeltaNet guard in `types.py` (`set_derived_and_validate_values`), computing the CP size the same way (`ici_context_parallelism * dcn_context_parallelism`). Since `context_parallel_load_balance` defaults to true, this now fails at config parse time instead of at the first forward step, after device allocation and weight init. The layer keeps its runtime check as a second line of defence. Covered by `test_kda_cp_rejects_load_balance_at_config_time`.

src/maxtext/layers/attention_kda.py
chiaotung97: Done, same place, mirroring the existing `cudnn_flash_te` + packing guard: `attention_type == "kda" and packing and max_segments_per_seq <= 0` now raises at config parse time, so it fails before the dataset pipeline is built. The runtime check in the layer stays. Covered by `test_kda_packing_requires_max_segments_at_config_time`.

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,672 @@
...
+          for k in range(self.kernel_size - 1, -1, -1)
+      ]
+
+    output = jnp.zeros((B, T, F), dtype=x.dtype)
+    for k in range(self.kernel_size):
+      offset = self.kernel_size - 1 - k
+      x_slice = x_padded[:, offset : offset + T, :]
+      if segment_ids is not None:
+        x_slice = x_slice * masks[k]
+      output = output + x_slice * self.kernel[k]
```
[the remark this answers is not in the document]
chiaotung97: Checked this against the flax version the PR was validated on (0.12.9), and the suggestion would go the wrong way:

- `param[0]` → works
- `param.value[0]` → `DeprecationWarning: '.value' access is now deprecated.
For
  Variable[Array] instances use: variable[...]`
- `param.get_value()[0]` → works

So `self.kernel[k]` is already the supported form. The conv path is exercised by `test_short_conv_no_cp` and `test_short_conv_cp_halo`.

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,672 @@
...
+    scale = self.key_head_dim**-0.5
+    safe_gate = cfg.use_kda_safe_gate
+    lower_bound = cfg.kda_lower_bound if safe_gate else None
+    if not safe_gate and cfg.kda_lower_bound != 0.0:
+      warnings.warn(
+          f"kda_lower_bound={cfg.kda_lower_bound} is ignored because use_kda_safe_gate=False. "
+          "Set use_kda_safe_gate=True to enable lower_bound clamping.",
+          stacklevel=2,
+      )
+    n_max = cfg.max_segments_per_seq if cfg.max_segments_per_seq > 0 else None
```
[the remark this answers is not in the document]
chiaotung97: In place: when CP is active and the caller supplied no segmentation, an all-ones `decoder_segment_ids` is synthesized **outside** the `shard_map` (with `n_max = 1`), so `shard_map` never receives `None` for a sharded input. Covered by `test_kda_cp_full_layer_dummy_segments`, which runs the full layer under CP with no user segment ids.

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,672 @@
...
+        q, k, v, g, beta, a_log, delta_time_bias_2d, *rest = args
+        seg = rest[0] if rest else None
+        max_num_segments = n_max
+
+        # CP: provide a dummy seg (all-ones) so the kernel has
+        # segment_ids to derive cu_seqlens from, even when the user
+        # hasn't supplied real segmentation info.
+        if seg is None and cp_size > 1:
+          seg = jnp.ones(q.shape[:2], dtype=jnp.int32)
+          max_num_segments = 1
```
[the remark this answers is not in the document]
chiaotung97: Done, that is the current implementation: the helper unpacks `seg = rest[0] if rest else None` and no longer builds a local dummy array.

scripts/dev/kda_e2e_smoke.py
```diff
@@ -0,0 +1,235 @@
...
+        f"(linear_conv_kernel_dim={cfg.linear_conv_kernel_dim}), got {args.delay}; "
+        "otherwise the target can be copied by the convolution without KDA state"
+    )
+  devices = jax.devices()
+  print(f"devices: {devices}")
+  mesh = jax.sharding.Mesh(np.array(devices), ("x",))
+  rngs = nnx.Rngs(0)
+  with mesh:
+    model = TinyKdaLM(cfg, mesh, args.num_layers, rngs=rngs)
+  n_params = sum(v.size for v in jax.tree.leaves(nnx.state(model)) if isinstance(v, (jax.Array, np.ndarray)))
```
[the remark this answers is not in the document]
chiaotung97: This file no longer exists, the script became `tests/unit/kda_attention_test.py::TestKdaE2eSmoke`, which asserts on the loss trajectory rather than counting parameters, so the `nnx.state` / `Variable` issue doesn't arise. (For the record, under flax 0.12.9 the supported accessors are `variable[...]` and `variable.get_value()`, not `.value`.)

src/maxtext/kernels/kda/tokamax.py
chiaotung97: Done in part, `chunk_kda` / `tokamax_chunk_kda` now take an `implementation` argument defaulting to `"mosaic"`, so tests and debugging can pass `"xla"` explicitly. That's how the parity test already drives it.

I kept it off the config surface because of tokamax's fallback semantics: `implementation=None` doesn't mean "pick one for me", it means the sequence `("mosaic", "xla")` — so when the Pallas kernel declines the shapes or hardware it drops to the pure-JAX reference recurrence. Results stay correct, it runs far slower, and nothing logs the switch. A config that can reach that default would let a training run degrade silently.

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+    cfg = self.config
+
+    # KDA head dimensions derived from global config. KDA uses one head count
+    # and one head dim for q, k and v alike (no GQA-style grouping):
+    #   key_head_dim = value_head_dim = config.head_dim (kv_channels)
+    #   num_key_heads = num_value_heads = config.base_num_query_heads (num_attention_heads)
+    self.key_head_dim = cfg.head_dim
+    self.value_head_dim = cfg.head_dim
+    self.num_key_heads = cfg.base_num_query_heads
+    self.num_value_heads = cfg.base_num_query_heads
```
NuojCheng: use `cfg.num_query_heads` instead

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+      )
+    else:
+      self.q_conv = None
+      self.k_conv = None
+      self.v_conv = None
+
+    # QKV projections
+    # Separate projections for Q, K, V (not fused) to allow independent conv
+    self.q_proj = linears.DenseGeneral(
+        in_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+        weight_dtype=cfg.weight_dtype,
+        kernel_axes=("embed", "heads", "kv"),
+        use_bias=cfg.attention_bias,
+        shard_mode=cfg.shard_mode,
+        matmul_precision=cfg.matmul_precision,
+        rngs=rngs,
+    )
+
+    self.k_proj = linears.DenseGeneral(
+        in_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+        weight_dtype=cfg.weight_dtype,
+        kernel_axes=("embed", "heads", "kv"),
+        use_bias=cfg.attention_bias,
+        shard_mode=cfg.shard_mode,
+        matmul_precision=cfg.matmul_precision,
+        rngs=rngs,
+    )
+
+    self.v_proj = linears.DenseGeneral(
+        in_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+        use_bias=cfg.attention_bias,
+        shard_mode=cfg.shard_mode,
+        matmul_precision=cfg.matmul_precision,
+        rngs=rngs,
+    )
+
+    # Output projection
+    self.o_proj = linears.DenseGeneral(
+        in_features_shape=(self.num_value_heads, self.value_head_dim),
+        out_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+        use_bias=False,
+        shard_mode=cfg.shard_mode,
+        matmul_precision=cfg.matmul_precision,
+        rngs=rngs,
+    )
+
+    # Beta projection for generating beta (Delta rule mixing coefficient)
+    # beta has shape [B, T, H] - per-head scalar
+    self.b_proj = linears.DenseGeneral(
+        in_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+
+    # Q/K L2 normalization is applied in this layer, before chunk_kda (see
+    # __call__), so the kernel is passed use_qk_l2norm=False.
+
+    # Output gate projection: full-rank gate of shape [B, T, H, V] (`g_proj` in
+    # public KDA checkpoints). The low-rank bottleneck variant is not
+    # implemented — see the class docstring; use_kda_lora=True is rejected at
+    # config time.
+    self.gate_proj = linears.DenseGeneral(
+        in_features_shape=cfg.base_emb_dim,
```
NuojCheng: use `cfg.emb_dim`

src/maxtext/configs/types.py
```diff
@@ -4943,6 +4934,29 @@ def calculate_global_batch_sizes(per_device_batch_size, expansion_factor, num_de
           f"QK-Clip is only supported when attention_type='mla', but found attention_type='{self.attention_type}'."
       )
 
+    if self.attention_type == "kda" and self.scan_layers:
+      raise ValueError(
+          "attention_type='kda' requires scan_layers=false: KDA layers have not been validated "
+          "inside a scanned layer stack. Set scan_layers: false."
+      )
+
+    kda_context_parallel_size = self.ici_context_parallelism * self.dcn_context_parallelism
```
NuojCheng: let's update this line to
```
kda_context_parallel_size = getattr(self, f"ici_{self.context_sharding}_parallelism", 1) * getattr(self, f"dcn_{self.context_sharding}_parallelism", 1)
```

src/maxtext/layers/attention_kda.py
NuojCheng: I think we should push for upgrading tokamax depencency instead adding this logic in the code.

NuojCheng: I will check with on-duty

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+    padding here would only add a discarded array to the graph.
+    """
+    pad_width = [(0, 0)] * x.ndim
+    pad_width[seq_axis] = (halo_size, 0)
+    return jnp.pad(x, pad_width)
+
+  if not _has_named_axis(axis_name):
+    return _zero_pad()
+
+  cp_size = jax.lax.psum(1, axis_name=axis_name)
```
NuojCheng: use `jax.lax.axis_size(axis_name=axis_name)`

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+        f"sequence length ({t_local}) on the '{axis_name}' axis. The causal "
+        "convolution receptive field would span multiple CP ranks, which is "
+        "not implemented. Use a smaller linear_conv_kernel_dim, a longer "
+        "sequence, or a smaller CP size."
+    )
+
+  # Forward ring: each rank sends its tail to the next rank.
+  tail = jax.lax.dynamic_slice_in_dim(x, x.shape[seq_axis] - halo_size, halo_size, axis=seq_axis)
+  perm = [(i, (i + 1) % cp_size) for i in range(cp_size)]
+  halo = jax.lax.ppermute(tail, axis_name=axis_name, perm=perm)
```
NuojCheng, line 162: Why the context parallelism logic seems splitted into tokamax and maxtext? It looks like the majority of context parallelism logic are [in tokamax](https://github.com/openxla/tokamax/blob/main/tokamax/_src/ops/experimental/kda/cp_utils.py), which uses an all-gather and collective cost scales with sequence length. Why we need a separate shardmap here for ppermute?

src/maxtext/layers/attention_kda.py
```diff
@@ -0,0 +1,854 @@
...
+
+
+class KimiDeltaAttention(nnx.Module):
+  """Kimi Delta Attention (KDA) layer.
+
+  KDA is a linear attention mechanism that uses the Delta Rule for state
+  correction:
+    S' = S * exp(g_t)
+    residual = v_t - k_t^T @ S'
+    S = S' + beta_t * k_t (x) residual
```
NuojCheng, line 240: for my own understanding, are residuals hardcoded in the tokamax kernel? Can we tune them by only changing maxtext code?

docs/reference/kda_cp_support.md
```diff
@@ -0,0 +1,205 @@
...
+### Plan 2: ShortConvolution CP Wrapper (`layers/attention_kda.py`)
+
+`ShortConvolution.__call__` internally calls `halo_exchange_for_conv`, which requires the CP axis scope (`cfg.context_sharding`). Inside `KimiDeltaAttention.__call__`, when CP is enabled, wrap the q/k/v conv calls in an independent `jax.shard_map`.
+
+Change location: the conv call segment after QKV projection in `KimiDeltaAttention.__call__`.
+
+Key design decisions:
+
+- **conv shard_map and chunk_kda shard_map are independent**: two separate `jax.shard_map` invocations, freeing conv's ppermute buffer in between
+- `check_vma=False`: FlashAttention custom rules may falsely report VMA errors
```
NuojCheng: MaxText actually encourages using `check_vma=True`. I won't call it "falsely report VMA errors". Could you please remove this line, and we will followup on this.

docs/reference/kda_cp_support.md
```diff
@@ -0,0 +1,205 @@
...
+    if TokamaxContextParallelMetadata is None:
+        raise ImportError(...)  # refuse to run: CP would silently break state
+    cp_ctx = TokamaxContextParallelMetadata(mesh=self.mesh, axis_name=cp_axis_name)
+```
+
+`ContextParallelMetadata` is a frozen dataclass. `mesh` and `axis_name` are set at construction time; chain metadata fields are populated internally by `chunk_kda`. The `axis_name` comes from `cfg.context_sharding`, so expert-as-context meshes bind the metadata to the `"expert"` axis.
+
+#### 3b. Partition Spec Injection
+
+`nnx.logical_to_mesh_axes` may map the T axis to `None` (or to a mesh axis that does not carry the sequence shard) due to Flax rule priority + size-1 axis stripping, but shard_map requires the T axis to carry the CP axis:
```
NuojCheng: I don't see `nnx.logical_to_mesh_axes` gets used anywhere in this PR. If we use maxtext's own `logical_to_mesh_axes`, this problem should be solved.

docs/reference/kda_cp_support.md
```diff
@@ -0,0 +1,205 @@
...
+CP (cp_size > 1):
+  [B, T/cp, E] → QKV proj → SHARD_MAP(ShortConv w/ halo)  ← independent conv shard_map
+                           → SiLU + L2Norm
+                           → ContextParallelMetadata(mesh, cfg.context_sharding)  ← constructed outside shard_map
+                           → _inject_cp_axis_on_T + _wsc     ← partition spec fixup
+                           → SHARD_MAP(chunk_kda)            ← context_parallel_metadata passed in
+                           → [B, T/cp, E]
+```
+
+Key difference from MLA CP: MLA relies on splash attention kernel internally doing implicit all_gather K/V → local attention; KDA does not rely on all_gather. Instead, `ContextParallelMetadata` lets the kernel coordinate recurrent state across ranks during forward/backward.
```
NuojCheng: This statement seems differring from the collective code in tokamax KDA, see https://github.com/openxla/tokamax/blob/e99898210069f7b291065f4b3794ead25373204d/tokamax/_src/ops/experimental/kda/pallas_mosaic_tpu_fwd_kernel.py#L411-L414.

src/maxtext/layers/attention_kda.py
Shuwen-Fang: Will work on upgrading it

src/maxtext/layers/attention_kda.py
chiaotung97: Two different collectives at two different scales. 

Tokamax's side: 
* The recurrent state. ContextParallelMetadata drives state hand-off inside the kernel, and it does not gather the sequence, it gathers each rank's per-segment affine summary. 
* Forward: S_ext_local` [H,B,K,V] and M_local [H,B,K,K] (pallas_mosaic_tpu_fwd_kernel.py:411-414; chunk_gated_delta_rule_fwd_h_pre_process's return types are [H,B,K,V] / [H,B,K,K], no T). 
* Backward: dS_ext + dM packed into a single gather, packed [B,H,K,V+K] (..._bwd_kernel.py:1752-1754, with the comment that packing halves the CP collective cost). Plus one 2×cp_size int32 gather for chain metadata (cp_utils.py:260-261). So this cost is constant in sequence length, O(cp_size · H · K · (V+K)).

MaxText's side: 
* The conv halo. The depthwise causal conv runs before chunk_kda, in MaxText code, so its left-boundary context can't come from the state exchange, it needs actual neighbour activations. tokamax reserves fields for exactly this (cp_utils.py:111-112, conv1d_kernel_size / pre_num_conv_tokens) but they're documented "Reserved for conv1d CP.  None for KDA" (:101-102), unimplemented. KDA in tokamax assumes q/k/v arrive already convolved. Our halo is kernel_size-1 tokens (3 by default), one ppermute hop, independent of both T and cp_size. If tokamax implements conv-CP, ours is the part that disappears.

Why a second shard_map. ppermute needs the CP axis bound, and shard_map is what binds it. The conv sits between the projections and SiLU / L2-norm / the gate+beta projections, all MaxText-side and auto-sharded, so folding it into the kernel's shard_map would mean dragging those  stages inside with it. Two shard_maps also let the halo buffer die before the kernel's collectives allocate.

src/maxtext/layers/attention_kda.py
chiaotung97: Yes. The residual and state update live inside the kernel, and api.kimi_delta_attention exposes no hook for them, so changing the formula is a tokamax change rather than a MaxText one.

MaxText does control every input to that formula. The gate comes from g_proj together with A_log and dt_bias, and is bounded by use_kda_safe_gate and kda_lower_bound. Beta is sigmoid(b_proj(x)). The layer also decides where Q/K L2 normalization happens and applies the gated RMSNorm after the kernel. Decay and gate parameterization are therefore tunable here, while the Delta Rule residual is not.

One knob is still hardcoded. scale is fixed at key_head_dim ** -0.5 on line 713, which matches the kernel default.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. num_key_heads and num_value_heads both read cfg.num_query_heads now.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. I changed all seven projections: lines 351, 364, 377, 392, 407, 422 and 442. Line 407 (g_proj) had the same pattern and you did not flag it, so it is fixed too.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. Covered in the note on line 351.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. Covered in the note on line 351.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. Covered in the note on line 351.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. Covered in the note on line 351.

src/maxtext/layers/attention_kda.py
chiaotung97: Done. Covered in the note on line 351.

src/maxtext/configs/types.py
chiaotung97: Done. The same validator already computes context_parallel_size at line 4625 using exactly that expression, so the KDA guard reuses it instead of naming ici_context_parallelism directly. Expert-as-context is now covered.

src/maxtext/layers/attention_kda.py
chiaotung97: Done, now jax.lax.axis_size(axis_name=axis_name).

docs/reference/kda_cp_support.md
chiaotung97: Removed that line, you are right that it does not hold up. I also measured both KDA shard_maps with check_vma=True on 4x v6e: 3 failed and  9 passed. The conv halo shard_map is fine with True. The kernel shard_map fails for two reasons, and both are in tokamax.
1. cp_utils.py:297. The fori_loop in _derive_cp_metadata_from_segment_ids enters with a replicated carry and returns one that varies on context.  JAX rejects the scan and suggests jax.lax.pcast(..., ('context',),  to='varying') on the initial carry.
2. jax/_src/pallas/core.py:1888 requires manual_axis_type on jax.ShapeDtypeStruct, which the KDA Pallas launcher does not set.

Until those land, False matches the rest of MaxText. attention_op.py  lines 1778, 1843 and 2298 plus splash_attention_kernel.py:2155 all  hardcode it, and config.check_vma is read only by moe.py:2625.

src/maxtext/layers/attention_kda.py
chiaotung97: > I think we should push for upgrading tokamax depencency instead adding this logic in the code.

That works out well, because tokamax 0.0.14 is on PyPI now and it does ship tokamax._src.ops.experimental.kda. I diffed the published wheel against the branch this PR was validated on. api.py, cp_utils.py and arg_specs.py are byte identical, and the only differences in base.py and common.py are internal. It requires jax>=0.11.0, which matches our baseline, so raising the floor to >=0.0.14 and regenerating the locks should not drag JAX along.

Once the bump lands I will delete the availability path: _MIN_TOKAMAX_VERSION, _predates_kda and kda_api_unavailable in kernels/kda/tokamax.py, the try/except at attention_kda.py:47-52, and the design doc section. CI will also stop skipping the kernel tests, which fixes the coverage number by itself.

One thing I need a call on. decoupled-requirements.txt:204 pins tokamax==0.0.12 on purpose, and 0.0.14 requires xprof>=2.23.1, so that env cannot take the raise. It only breaks because nnx_decoders.py:42 imports attention_kda at module scope.

docs/reference/kda_cp_support.md
chiaotung97: Done. KDA does gather, it just gathers a fixed size state summary instead of K/V. 
  
Forward gathers `S_ext [H, B, K, V]` and `M [H, B, K, K]` at `pallas_mosaic_tpu_fwd_kernel.py:411-414`. Backward packs `dS_ext` and `dM` into a single `[B, H, K, V + K]` gather at `pallas_mosaic_tpu_bwd_kernel.py:1752-1754`. Chain metadata adds `2 x cp_size` int32 at `cp_utils.py:260-261`. None of the three carries T, since `chunk_gated_delta_rule_fwd_h_pre_process` returns `[H, B, K, V]` and `[H, B, K, K]` whatever `T_local` is, so the CP cost is `O(cp_size * B * H * K * (V + K))` per step.
  
The paragraph is now a table of those three gathers, and the MLA contrast is about the payload rather than about whether a collective exists.

docs/reference/kda_cp_support.md
chiaotung97: Done, that line is gone. I measured it instead of asserting it. On 4xTPU v6e, `check_vma=True` on the kernel shard_map gives 9 passed and 3 failed, and enabling it on the conv side as well gives the identical result, so the conv side tolerates it and the failure is kernel side.
  
Both causes are in tokamax. First, `cp_utils.py:297`, the `fori_loop` in `_derive_cp_metadata_from_segment_ids` enters with a replicated carry (`bool[]`, `int32[]`) and returns one that varies on the CP axis, and JAX's own message suggests `pcast(..., ('context',), to='varying')` on the initial carry. Second, `jax/_src/pallas/core.py:1888` needs `manual_axis_type` on every `ShapeDtypeStruct`, which the KDA Pallas launcher does not set. One of the three failures runs at `cp_size=1`, so the second cause is not CP specific.
  
The doc now records that measurement, and notes that `False` matches `layers/attention_op.py` lines 1778, 1843 and 2298 plus `splash_attention_kernel.py:2155`, with `config.check_vma` read only by `layers/moe.py:2625`.

docs/reference/kda_cp_support.md
chiaotung97: Done, that sentence is gone. You are right that `nnx.logical_to_mesh_axes` is not used anywhere in this PR, we import MaxText's own at `attention_kda.py:57`.
  
I measured what it actually resolves to. With `ici_context_parallelism=2` the T axis is `P('fsdp', 'context', None)`, and with `ici_context_parallelism=4` it is `P(None, 'context', None)`, so on the default axis the overwrite is a no-op and the problem I described did not exist. It still matters for expert-as-context, where the same call gives `P(('fsdp', 'expert'), None, None)` because `activation_norm_length` has no `expert` rule at `configs/types.py:1383`, and the shard_map would then run replicated over the axis its collectives use. 
  
The doc has all three rows tabulated, and the `_inject_cp_axis_on_T` docstring is rewritten to state only that.

src/maxtext/layers/attention_kda.py
Shuwen-Fang: https://github.com/AI-Hypercomputer/maxtext/pull/5317 merged, we are now upgraded to 0.0.14

src/maxtext/layers/attention_kda.py
NuojCheng: Got it, thanks!

docs/reference/kda_cp_support.md
NuojCheng: Yes trivial rank gonna bring issues and may need to tune `remove_size_one_mesh_axis_from_type` in `jax.config`. Thank you for the update!

docs/reference/kda_cp_support.md
NuojCheng: Make sense and this all gather does seem cheap! It yields slightly worse arithematic intensity compared to softmax attention CP due to the `cp_size` scalar but does yield under the same level of magnitude.

NuojCheng (approved):
Thank you for the contribution!! After our tokamax dependency gets upgraded we could update the import statement.

PR #5148, opened 2026-09-04
Register LazyTensor with the Orbax v1 leaf-handler registry

### LINKED ISSUE #5106: checkpoint_conversion: save fails with Orbax NoEntryError when --lazy_load_tensors is combined with a single simulated device
### Bug report

`save_weights_to_checkpoint` in `src/maxtext/checkpoint_conversion/utils/utils.py` fails deterministically when lazily-loaded tensors are combined with a single simulated device. No checkpoint is produced, so every conversion using that configuration is blocked.

**Steps to reproduce**

On any machine that can run MaxText checkpoint conversion (the failing step is CPU-only; we reproduced it on an 8xH100 x86_64 node and on aarch64 GB200/GB300 nodes, but no accelerator is required to hit it):

1. Install MaxText at `main` (2026-09-02 or later) with `orbax-checkpoint` 0.12.4.
2. Fetch any supported Hugging Face checkpoint, e.g. `meta-llama/Llama-3.1-8B-Instruct`.
3. Run a `checkpoint_conversion` entry point that calls `save_weights_to_checkpoint`, with lazy loading on and exactly one simulated device:

```bash
JAX_PLATFORMS=cpu python3 -m maxtext.checkpoint_conversion.to_maxtext \
  src/MaxText/configs/base.yml \
  model_name=llama3.1-8b \
  base_output_directory=/tmp/llama31-8b-orbax \
  --hf_model_path=/path/to/Llama-3.1-8B-Instruct \
  --lazy_load_tensors=true \
  --simulated_cpu_devices_count=1 \
  --save_dtype=bfloat16
```

The default `--simulated_cpu_devices_count=16` hides the bug, because it routes through the sharding path. Setting it to `1` is the only change needed to trigger the failure.

**Expected behavior**

The conversion completes and writes the converted Orbax checkpoint to `base_output_directory`, exactly as it does today with `--simulated_cpu_devices_count > 1` and exactly as it did before `common/checkpointing.py` moved to the Orbax v1 API. `--simulated_cpu_devices_count` is documented as controlling how many CPU devices are simulated for sharding; it is not expected to change whether a checkpoint can be saved at all. The run should log `Conversion complete. Checkpoint saved to ...` and exit 0, and the resulting checkpoint should be byte-equivalent in content to one produced with more simulated devices, since the single-device path differs only in that it skips sharding.

**Actual behavior**

The save aborts with an Orbax `NoEntryError` raised from `resolve_handler_for_save` while resolving a handler for the `"items"` checkpointable, before any file is written. The process exits non-zero and `base_output_directory` is left empty. In our runs this happened about 15 seconds into the conversion step, deterministically, on every attempt and on both x86_64 and aarch64.

**Root cause**

The lazy proxies are never materialized on the single-device path:

- `to_maxtext.py` places `LazyTensor` proxies as weight-tree leaves under `--lazy_load_tensors`, and registers a handler for them through the **Orbax v0** API (`orbax.checkpoint.type_handlers.register_type_handler`).
- The only surviving code that converts those proxies to NumPy is inside `shard_jax_weights` (`if not isinstance(arr, (np.ndarray, jax.Array)): arr = np.array(arr)`).
- `save_weights_to_checkpoint` skips `shard_jax_weights` when `device_count == 1` (logging `Single device: Skip sharding`), so unmaterialized proxies end up in the `TrainState` passed to `save_checkpoint`.
- `common/checkpointing.py` now uses **Orbax v1** (`from orbax.checkpoint import v1 as ocp`) and calls `save_checkpointables` with `{"items": state}`. Orbax v1 resolves a handler by checkpointable name or `handler.is_handleable(value)`, never by the value's top-level type. `"items"` is unbound in the default global registry, so `PyTreeHandler.is_handleable` tree-maps over the **leaves** against `StandardLeafHandlerRegistry` (`jax.Array`, `np.ndarray`, `int`, `float`, `bytes`, `str`) with `issubclass`. `LazyTensor` is a plain proxy implementing `__array__` and is not an `np.ndarray` subclass, so no leaf handler matches. The underlying `UnregisteredTypeError` is swallowed by a bare `except Exception` and re-surfaces as an opaque `NoEntryError`.

The Orbax v0 and v1 leaf registries are disjoint, so the v1 migration of `common/checkpointing.py` orphaned the v0 `LazyTensor` registration in `to_maxtext.py` — the registration still runs (it still logs its `np.ndarray` collision warning) but is invisible to the v1 save path.

**Suggested fix**

Materialize the lazy leaves on the single-device branch, reusing the conversion `shard_jax_weights` already performs. We validated this experimentally as an in-image patch on 8xH100 with Llama-3.1-8B: the previously-failing conversion completes, the checkpoint is written, downstream decode passes, and inference throughput and step times are unchanged within run-to-run noise (about 0.1-0.4%) against a pre-regression baseline. The diff is below.

```diff
--- a/src/maxtext/checkpoint_conversion/utils/utils.py
+++ b/src/maxtext/checkpoint_conversion/utils/utils.py
@@ def shard_jax_weights(...)
   return jax_weights
 
 
+def materialize_lazy_weights(jax_weights, mem_info):
+  """Materializes lazily-loaded weight proxies into NumPy arrays."""
+  start = time.time()
+  jax_weights_flat, jax_weights_struct = tree.flatten(jax_weights)
+  del jax_weights
+  gc.collect()
+
+  jax_weights_new = []
+  jax_weights_flat.reverse()
+  num_weights = len(jax_weights_flat)
+  for _ in tqdm(range(num_weights)):
+    jax_weight = jax_weights_flat.pop()
+    if not isinstance(jax_weight, (np.ndarray, jax.Array)):
+      # materialize lazy tensor
+      jax_weight = np.array(jax_weight)
+    jax_weights_new.append(jax_weight)
+    del jax_weight
+    gc.collect()
+    logging.debug("Memory usage: %f GB", mem_info.memory_info().rss / (1024**3))
+
+  jax_weights = tree.unflatten(jax_weights_struct, jax_weights_new)
+  max_logging.log(f"Elapse for lazy weight materialization: {(time.time() - start) / 60:.2f} min")
+
+  return jax_weights
+
+
 def save_weights_to_checkpoint(
@@ in save_weights_to_checkpoint
   else:
     # If number of simulated devices is 1, SKIP sharding and SKIP jax conversion.
     max_logging.log("Single device: Skip sharding")
+    # Sharding is what would otherwise materialize lazily-loaded weights, so do
+    # it here: the save path below has no handler for lazy proxy leaves.
+    max_logging.log("Single device: Materializing lazy weights")
+    jax_weights = materialize_lazy_weights(jax_weights, mem_info)
```

It mirrors `shard_jax_weights`' reverse-pop plus `gc.collect()` loop, so peak memory stays roughly one tensor above the converted tree, and checkpoint contents are unchanged: this is the same conversion the multi-device path already performs and the same conversion the pre-migration `LazyTensorHandler.serialize` performed.

Alternatively, port `LazyTensorHandler` to an Orbax v1 `LeafHandler` registered through `PyTreeOptions`/`Context`, which would preserve streaming into serialization but pins MaxText to the v1 leaf-handler API and only helps `to_maxtext.py`.

The blast radius is wider than one entry point: `save_weights_to_checkpoint` serves several conversion entry points, and any of them would hit this if they produced lazy leaves. Today only `to_maxtext.py` constructs `LazyTensor`, so only it is affected.

### Logs/Output

Relevant log lines, in order, from a failing conversion run:

```
Type handler registry overriding type "<class '__main__.LazyTensor'>" collision on np.ndarray
Lazy loading ENABLED
Single device: Skip sharding
```

followed by:

```
File ".../checkpoint_conversion/utils/utils.py", line 1302, in save_weights_to_checkpoint
    checkpointing.save_checkpoint(checkpoint_manager, step, state_new, config=config)
File ".../common/checkpointing.py", line 992, in save_checkpoint
    checkpoint_manager.save_checkpointables(...)
File ".../orbax/checkpoint/experimental/v1/_src/saving/saving.py", line 321, in ...
    resolve_handler_for_save(...)
File ".../orbax/checkpoint/experimental/v1/_src/handlers/registration.py", line 623, in resolve_handler_for_save
    raise NoEntryError(
```

No checkpoint files are produced. After applying the suggested fix, the same run instead logs `Single device: Materializing lazy weights`, then `Elapse for lazy weight materialization: 1.04 min`, then `Conversion complete. Checkpoint saved to ...`, and downstream decode succeeds with no traceback anywhere in the log.

### Environment Information

- **MaxText version / commit:** `main` as of 2026-09-02. Relevant blobs: `src/maxtext/checkpoint_conversion/utils/utils.py` = `26677a0`, `src/maxtext/common/checkpointing.py` = `a11681b`, `src/maxtext/checkpoint_conversion/to_maxtext.py` = `ce2eecb`.
- **orbax-checkpoint:** 0.12.4
- **jax:** 0.11.2.dev20260901
- **Model:** `meta-llama/Llama-3.1-8B-Instruct` (bf16 safetensors)
- **Conversion step:** CPU-only (`JAX_PLATFORMS=cpu`), `--simulated_cpu_devices_count=1`, `--lazy_load_tensors=true`, `--save_dtype=bfloat16`, `async_checkpointing=false`
- **Operating system / hardware:** Linux; reproduced on `x86_64` (8xH100 node) and `aarch64` (GB200/GB300 nodes). The failure is platform-independent — it happens before any device work.

### Additional Context

Reported by NVIDIA. This is a regression relative to the pre-Orbax-v1 behaviour: the same configuration worked before `common/checkpointing.py` moved to the v1 API, because the v0 `LazyTensorHandler` was still consulted at save time. It began failing in our nightly runs on the first day that picked up the v1 migration commit, and the last passing run used MaxText from the day before.

---

_This issue was drafted with assistance from the `opus` AI model._

lokic233:
## What

Registers the converter's `LazyTensor` leaf type with the Orbax **v1** leaf-handler
registry, so HF→MaxText checkpoint conversion works again at
`--simulated_cpu_devices_count=1`.

Fixes #5106.

## Why

`checkpoint_conversion/to_maxtext.py:297` registers `LazyTensor` through the Orbax
**v0** API (`type_handlers.register_type_handler`). Since c41b2ae3a *("Refactor
MaxText's checkpointing system to primarily use the Orbax v1 API",  2026-09-01)* the
save path runs through `ocp.training.Checkpointer.save_checkpointables`, which
resolves handlers against the **v1** registry only.

The two registries are disjoint, and the compatibility bridge runs **one way**: a v1
`PyTreeHandler` builds a per-instance `StandardLeafHandlerRegistry` and *derives* its
v0 registry from itself. So a v0 global registration is structurally unreachable from
a v1 save. Conversion dies with `NoEntryError`.

The error names the *checkpointable* (`"items"` / `TrainState`), never the leaf type,
which is why this is easy to misdiagnose — the real failure is one leaf deeper.

c41b2ae3a touched zero files under `checkpoint_conversion/`, which is why this
regressed silently.

## Scope — measured, not inferred

The failure requires `simulated_cpu_devices_count=1`. At 2+, `shard_jax_weights`
materializes every leaf at `checkpoint_conversion/utils/utils.py:1194-1196` before the
tree ever reaches Orbax. Same tree, same model, same `lazy_load_tensors=True`, one flag
apart:

| `simulated_cpu_devices_count` | result on `main` |
|---|---|
| 1 | `NoEntryError` at `save_checkpointables` |
| 16 (the default) | rc=0, conversion succeeds |

**This is also why CI is green.**
`tests/integration/checkpoint_conversion_test.py::test_qwen3_30b_a3b_roundtrip_conversion`
is a real round-trip test collected by the `tpu-integration` and `gpu-integration`
flavors — but it never passes `--simulated_cpu_devices_count`, so it always runs at 16,
the one setting that cannot fail. Not "nobody runs it"; it runs constantly, at the safe
setting.

## How

Purely additive — 4 files, +68 −0. The existing v0 registration at `:297` is left in
place; removing it is orthogonal.

- `LazyTensorLeafHandler(NumpyLeafHandler)` — materializes via `np.asarray` in
  `serialize`, then defers to the numpy handler.
- `LazyTensorPyTreeHandler(PyTreeHandler)` — injects a per-instance
  `StandardLeafHandlerRegistry` carrying that handler. (`registry.add` takes a *type*,
  not an instance, hence the subclass.)
- `build_lazy_tensor_checkpointables_registry()` — a `local_registry` with
  `include_global_registry=True`, bound to checkpointable name `"items"`.
- Threaded through `checkpoint_context.build_context` /
  `checkpointing.create_orbax_checkpoint_manager` /
  `checkpoint_conversion/utils/utils.save_weights_to_checkpoint` as an optional
  `checkpointables_registry` kwarg, default `None` — no behavior change for any
  existing caller.

### The load-bearing detail: `secondary_typestrs=["np.ndarray"]`

Without it, the checkpoint records the *custom handler's* typestr, and a standard v0
restore then fails with `Unknown type: "...LazyTensorLeafHandler"`. The "fix" would
silently write **unloadable checkpoints** — strictly worse than the loud failure it
replaces. An earlier prototype did exactly that, and it was caught only because the
validator restores through the v0 `PyTreeCheckpointer` rather than just checking that
the save returned 0.

Reviewers: this is the line to scrutinize.

## Testing

CPU logits diff against the `transformers` reference at reduced dimensions
(hidden 256, 8 layers, vocab 512), every parameter randomized — default init leaves
RMSNorm scales at 0, which hides a wrong norm mapping. Run on a tree carrying the
patch, not a monkeypatch, at `simulated_cpu_devices_count=1`:

| arm | rc | max\|diff\| | cosine | argmax agreement |
|---|---|---|---|---|
| `lazy_load_tensors=True`, `scan_layers=False` | 0 | 1.66893e-06 | 1.00000000 | 100% |
| `lazy_load_tensors=True`, `scan_layers=True` | 0 | 1.57952e-06 | 1.00000000 | 100% |
| `lazy_load_tensors=False` (eager reference) | 0 | 1.66893e-06 | 1.00000000 | 100% |

Both lazy arms fail with `NoEntryError` without this patch; the eager arm is the
unchanged reference. Lazy now matches eager to the digit.

Both `scan_layers` settings are exercised because they are different code paths and
`scan_layers=True` is what training consumes.

The harness is sensitive to real corruption, not just to crashes: a same-shape
`gate_proj`↔`up_proj` swap in the same converter scores cosine **0.729** / argmax 18.8%.
Shape and name assertions cannot see that swap; only a value diff can.

## Notes

Two imports reach into `_src`
(`orbax...experimental.v1._src.serialization.numpy_leaf_handler` and
`..._src.serialization.registry`). If there is a public path for these I'd rather use
it — happy to switch.

### DIFF
--- a/src/maxtext/checkpoint_conversion/to_maxtext.py
+++ b/src/maxtext/checkpoint_conversion/to_maxtext.py
@@ -50,6 +50,7 @@ Example Usage:
 """
 
 import argparse
+import dataclasses
 from functools import partial
 import json
 import os
@@ -76,6 +77,10 @@ from maxtext.utils import max_logging, max_utils, maxtext_utils
 from maxtext.utils.globals import HF_IDS
 import numpy as np
 from orbax.checkpoint import type_handlers
+from orbax.checkpoint import v1 as ocp_v1
+from orbax.checkpoint.experimental.v1._src.handlers import pytree_handler
+from orbax.checkpoint.experimental.v1._src.serialization import numpy_leaf_handler
+from orbax.checkpoint.experimental.v1._src.serialization import registry as leaf_handler_registry
 from safetensors import safe_open
 
 try:
@@ -295,6 +300,51 @@ class LazyTensorHandler(type_handlers.NumpyHandler):
 # Register LazyTensor with the custom handler.
 # It's safe to register this globally even if eager loading is used.
 type_handlers.register_type_handler(LazyTensor, LazyTensorHandler(), override=True)
+
+
+class LazyTensorLeafHandler(numpy_leaf_handler.NumpyLeafHandler):
+  """Orbax v1 leaf handler for LazyTensor.
+
+  The v0 registration above cannot serve the v1 save path. A v1 ``PyTreeHandler``
+  resolves leaves through a per-instance ``LeafHandlerRegistry`` and *derives* its
+  v0 registry from itself, so the compatibility bridge only runs v1 -> v0; a v0
+  global registration is unreachable from v1.
+
+  Like its v0 counterpart this masquerades as the standard numpy handler --
+  ``secondary_typestrs`` below writes ``np.ndarray`` as the leaf's typestr -- so
+  the checkpoint is indistinguishable from one a plain ``NumpyLeafHandler``
+  produced and restores in a standard MaxText instance.
+  """
+
+  async def serialize(self, params, serialization_context):
+    # MATERIALIZE: trigger the lazy load (__array__) explicitly before saving.
+    # This ensures the parent NumpyLeafHandler receives real np.ndarrays.
+    params = [dataclasses.replace(param, value=np.asarray(param.value)) for param in params]
+    return await super().serialize(params, serialization_context)
+
+
+class LazyTensorPyTreeHandler(pytree_handler.PyTreeHandler):
+  """``PyTreeHandler`` whose leaf registry additionally accepts ``LazyTensor``."""
+
+  def __init__(self, **kwargs):
+    if "leaf_handler_registry" not in kwargs:
+      registry = leaf_handler_registry.StandardLeafHandlerRegistry()
+      registry.add(
+          LazyTensor,
+          numpy_leaf_handler.NumpyShapeDtype,
+          LazyTensorLeafHandler,
+          override=True,
+          secondary_typestrs=["np.ndarray"],
+      )
+      kwargs["leaf_handler_registry"] = registry
+    super().__init__(**kwargs)
+
+
+def build_lazy_tensor_checkpointables_registry():
+  """Registry that saves MaxText's "items" checkpointable with LazyTensor support."""
+  registry = ocp_v1.handlers.local_registry(include_global_registry=True)
+  registry.add(LazyTensorPyTreeHandler, checkpointable_name="items")
+  return registry
 
 
 def get_maxtext_model_info(config):
@@ -1093,6 +1143,7 @@ def main(
       config.checkpoint_storage_use_ocdbt,
       config.checkpoint_storage_use_zarr3,
       config=config,
+      checkpointables_registry=build_lazy_tensor_checkpointables_registry(),
   )
 
   print_ram_usage("Program Ends")
--- a/src/maxtext/checkpoint_conversion/utils/utils.py
+++ b/src/maxtext/checkpoint_conversion/utils/utils.py
@@ -53,6 +53,7 @@ from maxtext.common.gcloud_stub import gcs_storage
 from maxtext.checkpoint_conversion.utils.tensor_handling import nesting_depth, stacked_axes
 from maxtext.utils import max_logging
 import orbax.checkpoint as ocp
+from orbax.checkpoint import v1 as ocp_v1
 
 _storage = gcs_storage()
 Client = _storage.Client
@@ -1238,6 +1239,7 @@ def save_weights_to_checkpoint(
     use_ocdbt: bool,
     use_zarr3: bool,
     config=None,
+    checkpointables_registry: ocp_v1.handlers.CheckpointableHandlerRegistry | None = None,
 ):
   """Saves model weights to a MaxText-compatible checkpoint with optional sharding.
 
@@ -1254,6 +1256,10 @@ def save_weights_to_checkpoint(
           (OCDBT) format for improved metadata handling.
       use_zarr3: If True, uses the Zarr3 storage format for the underlying array data.
       config: Optional config to save along with checkpoint metadata.
+      checkpointables_registry: Optional Orbax v1 handler registry, for converters whose
+          weight pytrees hold leaf types Orbax does not know natively. Only matters when
+          device_count is 1; above that shard_jax_weights has already materialized every
+          leaf into a sharded jax.Array.
   """
   mem_info = psutil.Process()
   logging.debug("Memory usage: %f GB", mem_info.memory_info().rss / (1024**3))
@@ -1281,6 +1287,7 @@ def save_weights_to_checkpoint(
       save_interval_steps,
       use_ocdbt=use_ocdbt,
       use_zarr3=use_zarr3,
+      checkpointables_registry=checkpointables_registry,
   )
   if checkpoint_manager is None:
     raise RuntimeError("Failed to create Orbax checkpoint manager.")
--- a/src/maxtext/common/checkpoint_context.py
+++ b/src/maxtext/common/checkpoint_context.py
@@ -94,6 +94,7 @@ def build_context(
     colocated_python_checkpointing: bool = False,
     partial_load: bool = False,
     checkpoint_layout: ocp.options.CheckpointLayout | None = None,
+    checkpointables_registry: ocp.handlers.CheckpointableHandlerRegistry | None = None,
 ) -> ocp.Context:
   """Builds an Orbax v1 ``Context`` from MaxText checkpoint flags.
 
@@ -119,6 +120,10 @@ def build_context(
       equivalent of v0 ``partial_restore=True``).
     checkpoint_layout: On-disk layout (``ORBAX`` or ``SAFETENSORS``) for
       loading.
+    checkpointables_registry: Handler registry deciding which
+      ``CheckpointableHandler`` serves each named checkpointable. Callers whose
+      pytrees hold leaf types Orbax does not know natively pass a registry
+      carrying a handler for them; ``None`` leaves the global registry in place.
 
   Returns:
     A configured, unfrozen ``ocp_v1.Context``.
@@ -171,5 +176,8 @@ def build_context(
 
   if checkpoint_layout is not None:
     ctx.checkpoint_layout = checkpoint_layout
+
+  if checkpointables_registry is not None:
+    ctx.checkpointables.registry = checkpointables_registry
 
   return ctx
--- a/src/maxtext/common/checkpointing.py
+++ b/src/maxtext/common/checkpointing.py
@@ -375,6 +375,7 @@ def create_orbax_checkpoint_manager(
     todelete_subdir: str | None = None,
     todelete_full_path: str | None = None,
     ocdbt_target_data_file_size_bytes: int | None = None,
+    checkpointables_registry: ocp.handlers.CheckpointableHandlerRegistry | None = None,
 ):
   """Returns an Orbax v1 training ``Checkpointer``, or None if checkpointing is disabled."""
   if not enable_checkpointing:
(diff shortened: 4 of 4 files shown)


janbernloehr:
We hit #5106 independently in our nightly Llama-3.1-8B inference CI, which converts a
Hugging Face checkpoint to MaxText before serving it, and reached the same root cause
as this PR. Some extra reproduction detail in case it helps review.

**Failure**

The conversion step runs

```
JAX_PLATFORMS=cpu python3 -m maxtext.checkpoint_conversion.to_maxtext \
  model_name=llama3.1-8b scan_layers=true \
  --lazy_load_tensors=true --simulated_cpu_devices_count=1 \
  --hf_model_path=<local Llama-3.1-8B-Instruct safetensors dir> ...
```

and dies at save time with

```
NoEntryError: Could not identify a valid handler for the checkpointable: "items" and
checkpointable type=<class 'flax.training.train_state.TrainState'>
```

The message names the checkpointable, not the leaf, which — as the PR body says —
makes it easy to misread as a `TrainState` problem rather than a `LazyTensor` one.

**Regression bracket**

- Last good: nightly images built from `main` immediately before c41b2ae3a2586429...
  ("Refactor MaxText's checkpointing system to primarily use the Orbax v1 API", merged
  2026-09-01) landed.
- First bad: the next nightly, i.e. the first image containing that commit. Every
  nightly since has failed the same way, on every platform we run this on.
- `orbax-checkpoint` is 0.12.4 on both sides of the bracket, so this is not an Orbax
  version bump — it is exactly the v0-registration / v1-save-path mismatch this PR
  describes.

**Scope confirmation**

We independently observed the `simulated_cpu_devices_count` dependence stated in the
PR body: at 1 the `LazyTensor` proxies reach the v1 save unmaterialized; at 2+
`shard_jax_weights` converts every leaf first, so the same tree saves fine. Our CI
uses 1, which is why it has been red continuously while the round-trip integration
test stays green at the default.

**Environment**

- Conversion runs CPU-only (`JAX_PLATFORMS=cpu`); the converted checkpoint is then
  served on a single 8x H100 80GB SXM node, Linux/x86_64.
- Nightly JAX + MaxText container image, `orbax-checkpoint` 0.12.4, saved
  checkpointable type `flax.training.train_state.TrainState`.
- Source checkpoint: local Llama-3.1-8B-Instruct HF safetensors; bf16 and fp8 serving
  variants, both blocked by the same conversion failure.

**Validation of this PR**

We applied this PR's diff unmodified to the MaxText source in our nightly image and
re-ran the two previously failing full-size inference workloads (bf16 and fp8),
changing nothing else.

**Alternative, if the `_src` imports are a blocker**

Your "Notes" section flags that the handler reaches into
`orbax...experimental.v1._src`. Before this PR existed we validated a narrower
alternative against the same failing workload: materialize the lazy leaves inside the
`device_count == 1` branch of `save_weights_to_checkpoint` — the branch that today only
logs `"Single device: Skip sharding"` — mirroring what `shard_jax_weights` already does
at 2+ devices. It touches one file, needs no Orbax-internal imports, and does not change
what is written to disk, but it also does not make `LazyTensor` generally saveable
through v1, so it is strictly the smaller hammer. Happy to share that diff if
maintainers would rather avoid the `_src` dependency.

PR #5153, opened 2026-09-06
[Stacked PR 3/6] Add decoupled GDN Pallas backward pass kernel and parity tests

Rohan-Bierneni:
# Description

This is **PR 3 of 6** in the stacked series enabling the Pallas Gated Delta Net (GDN) backward pass kernel in MaxText.

This PR introduces the decoupled GDN Pallas backward pass kernel and complete analytical gradient parity testing:
1. **Decoupled Backward Architecture (`gdn_bwd_pallas.py`)**:
   - `pallas_gdn_bwd_kernel`: Executes 40+ GDN adjoint matrix recurrences via `pltpu.emit_pipeline` with vectorized head processing and zero HBM intermediate spills.
   - `conv1d_silu_bwd`: Executes immediately after Pallas via unrolled native JAX shift-multiply. Decoupling Conv1D backward avoids binding Conv1D gradient live ranges with GDN adjoint matrix state buffers, eliminates vector register spills, and keeps peak VMEM well within hardware boundaries.
   - `gdn_decoupled_conv1d`: Custom VJP unifying forward execution (reusing $T^{-1}$ cached in residuals from PR 2) and backward execution. **Named `gdn_decoupled_conv1d` (rather than `fused`) to accurately convey that Conv1D is decoupled in native JAX rather than fused with GDN recurrence in Pallas.**
   - Aliases: Provides `gdn_kernel` and backward-compatible aliases.
2. **Exhaustive Analytical Parity Test Suite (`tests/unit/gdn_bwd_pallas_test.py`)**:
   - Validates analytical gradients against pure JAX autodiff across sequence lengths (128 to 2048), batch sizes, and multi-head settings.
   - Asserts strict numerical parity across $dQ, dK, dV, deta, d^Glpha, dW_{conv}, db_{conv}$.

# Files Changed
- `src/maxtext/models/kernels/gdn/__init__.py`: Package export for `gdn_decoupled_conv1d`, `gdn_kernel`, `pallas_gdn_bwd_kernel`.
- `src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py`: Decoupled backward pass kernel and custom VJP implementation (`gdn_decoupled_conv1d`).
- `tests/unit/gdn_bwd_pallas_test.py`: Full analytical gradient parity test suite against JAX autodiff.

# Tests
- Verified with `pre-commit run --files ...` (all hooks passed: `codespell`, `pylint`, `pyink`).
- Parity test suite: `pytest tests/unit/gdn_bwd_pallas_test.py` validates analytical gradient equivalence to machine precision (13/13 unit tests passed in 137s).

# Checklist
- [X] I have run end-to-end tests and verified pre-commit linters pass.


src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+import functools
+from typing import Any, Optional, Tuple
+
+import jax
+from jax.ad_checkpoint import checkpoint_name
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+import jax.numpy as jnp
+
+try:
```
dandragona, line 18: I think all of this special handling can get removed. We can just raise an error if someone tries to use this kernel in types.py without the necessary dependencies.

src/maxtext/models/kernels/gdn/__init__.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2026 Google LLC
```
suhithr, line 1: Is there a requirement that the gdn kernel is in `models/` otherwise can we move this to `src/maxtext/kernels/`?

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+  return (grad_t,)
+
+
+invert_triangular_matrix.defvjp(
+    _invert_triangular_matrix_fwd,
+    _invert_triangular_matrix_bwd,
+)
+
+
+def chunk_forward(
```
suhithr: As a code organization pointer, I think we can move these JAX implementations of the GDN chunking logic into a separate file and leave `gdn_bwd_pallas.py` to be focused on the pallas kernel.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+  # State update: decayed previous state + rank-1 update from chunk with v_new
+  state_prev_decayed = state_prev * gating_last
+  k_scaled = k_h * gating_backward
+  state_new_intra = jnp.matmul(jnp.swapaxes(k_scaled, -1, -2), v_new)
+  state_new = state_prev_decayed + state_new_intra
+
+  return out, state_new, A
+
+
+def chunk_state_forward_with_cached_tinv(
```
suhithr: Given that `chunk_state_forward_with_cached_tinv` and `chunk_state_forward_with_tinv` perform nearly identical logic save for the `t_inv` calculation could they be factored into one implementation with a separate function to handle the `tinv` calculation?

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+    conv_weight_3d = conv_weight.astype(jnp.float32)
+  else:
+    conv_weight_3d = conv_weight[:, None, :].astype(jnp.float32)
+
+  conv_input = jnp.pad(qkv.astype(jnp.float32), ((0, 0), (kernel_size - 1, 0), (0, 0)))
+  conv_out = sum(conv_input[:, k : k + seq_len, :] * conv_weight_3d[k, 0, :] for k in range(kernel_size))
+  if conv_bias is not None:
+    conv_out = conv_out + conv_bias.astype(jnp.float32)
+  qkv_conv = jax.nn.silu(conv_out)
+  return conv_out, qkv_conv.astype(qkv.dtype)
```
suhithr: As we are recomputing `conv_out` below in the backwards pass, why are we returning it here? If it's not needed should we modify this function to only return `qkv_conv`?

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+  return state_new
+
+
+# ==============================================================================
+# SECTION 2: Conv1D + SiLU Forward & Backward Primitives (Pure JAX)
+# ==============================================================================
+
+
+def conv1d_silu_fwd(
+    qkv: jax.Array,
```
suhithr: It would be great if in these argument comments we denoted the shapes of the inputs (similar to http://shortn/_TCc6yQGC8C). This can help ensure we sum across the right axis in the manual `bwd`.

tests/unit/gdn_bwd_pallas_test.py
```diff
@@ -0,0 +1,979 @@
...
+    num_v_heads = 4
+    head_k_dim = 128
+    head_v_dim = 128
+    conv_kernel_size = 4
+    dim_size = num_k_heads * head_k_dim * 2 + num_v_heads * head_v_dim
+
+    key = jax.random.PRNGKey(789)
+    k1, k2, k3, k4, k5, k6, k7, k8 = jax.random.split(key, 8)
+
+    qkv = jax.random.normal(k1, (batch_size, seq_len, dim_size), dtype=jnp.float32)
```
suhithr, line 178: Lets use a helper function to initialize these fwd inputs, and one for the backwards inputs as well.

suhithr (commented):
Initial pass covering mostly changes around code organization and cleanup.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1801 @@
...
+  t_inv_val = t_inv_ref[...].astype(jnp.float32)
+  a_log_val = a_log_ref[...][0, :num_v_heads].astype(jnp.float32)
+  dt_bias_val = dt_bias_ref[...][0, :num_v_heads].astype(jnp.float32)
+
+  scale = 1.0 / jnp.sqrt(kq_head_dim)
+  mask_cumsum = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32))
+  mask_strict = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32), k=-1)
+  mask_causal = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32), k=0)
+
+  if use_qk_norm_in_gdn:
```
dandragona: here and below: when do we not use qk norm? I think we can simplify this.

tests/unit/gdn_bwd_pallas_test.py
```diff
@@ -0,0 +1,979 @@
...
+      rel_diff = abs_diff / (float(jnp.max(jnp.abs(exp_g))) + 1e-7)
+      status = "✅ MATCH" if rel_diff < 1e-3 else "❌ DIVERGED"
+      print(f"  {name:<28}: MaxAbsDiff = {abs_diff:.2e} | RelDiff =" f" {rel_diff:.2e} | {status}")
+      self.assertLess(
+          rel_diff,
+          1e-3,
+          f"{name} relative difference {rel_diff:.2e} exceeds tolerance 1e-3",
+      )
+
+    print("✅ All 7 parameter gradients match Pure JAX autodiff within 0.1% on" " CPU!")
```
suhithr, line 263: Should we add the pytest tag to run this on TPU especially if correctness is important for this kernel on TPU? Let me know if MaxText CI doesn't support this or you don't think it's necessary.

tests/unit/gdn_bwd_pallas_test.py
```diff
@@ -0,0 +1,979 @@
...
+          chunk_size=chunk_size,
+          use_qk_norm_in_gdn=True,
+          compute_dtype=jnp.float32,
+      )
+      return jnp.sum(out * do)
+
+    grads = jax.grad(loss_fn, argnums=(0, 1, 2, 3, 4, 5))(qkv, b, a, conv_weight, a_log, dt_bias)
+    for g in grads:
+      self.assertIsNotNone(g)
+      self.assertFalse(np.any(np.isnan(np.array(g))))
```
suhithr, line 314: Do we need this test that just checks if any gradients are missing / their values are NaN? Perhaps it could be folded into one of the above tests that exercises comparing the gradient values with the JAX autodiff.

src/maxtext/models/kernels/gdn/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Local GDN forward kernel package with triangular inverse matrix caching."""
+
+from . import compute_conv1d
```
suhithr, line 18: Let's use absolute imports.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=unused-import,unused-wildcard-import,wildcard-import
+"""Canonical Gated Delta Net (GDN) backward pass facade re-exporting from .gdn_bwd."""
+
+try:
+  from maxtext.models.kernels.gdn import gdn_bwd
+  from maxtext.models.kernels.gdn.gdn_bwd import *  # pylint: disable=wildcard-import
```
suhithr, line 20: Why do we need this wildcard import?

suhithr (commented):
Gave the `_test` file a pass. Lets compress number of tests perhaps with parametrization & folding some of the tests that test the same thing into one.

Additionally I think we should run at least a couple of the correctness tests of this kernel on TPU, I've flagged the ones I think we should tag.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+
+try:
+  from maxtext.models.kernels.gdn.gdn_bwd.bwd_memory_ref import make_bwd_block_specs
+  from maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import ensure_cpu_interpret_registered
+except (ImportError, ModuleNotFoundError):
+  try:
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.bwd_memory_ref import make_bwd_block_specs
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import ensure_cpu_interpret_registered
+  except (ImportError, ModuleNotFoundError):
+    from .bwd_memory_ref import make_bwd_block_specs
```
suhithr: I think we should fail if we cannot import from either of the above paths. Same as above for `normalizations`.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+  del kernel_size, pad_len
+  c = pl.program_id(1)
+  repeats = num_v_heads // num_kq_heads
+  q_size = num_kq_heads * kq_head_dim
+  k_size = num_kq_heads * kq_head_dim
+  v_size = num_v_heads * v_head_dim
+
+  @pl.when(c == 0)
+  def _init():
+    d_state_scr[...] = jnp.zeros((num_v_heads, kq_head_dim, v_head_dim), dtype=jnp.float32)
```
dandragona, line 83: I think this needs to be the final state gradient

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+  t_inv_val = t_inv_ref[...].astype(jnp.float32)
+  a_log_val = a_log_ref[...][0, :num_v_heads].astype(jnp.float32)
+  dt_bias_val = dt_bias_ref[...][0, :num_v_heads].astype(jnp.float32)
+
+  scale = 1.0 / jnp.sqrt(kq_head_dim)
+  mask_cumsum = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32))
+  mask_strict = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32), k=-1)
+  mask_causal = jnp.tril(jnp.ones((chunk_size, chunk_size), dtype=jnp.float32), k=0)
+
+  if use_qk_norm_in_gdn:
```
dandragona, line 111: Here and elsewhere, I think we always want to normalize, so let's not support the else branch IIUC

src/maxtext/models/kernels/gdn/gdn_bwd/api.py
```diff
@@ -0,0 +1,513 @@
...
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.compute_conv1d_bwd import conv1d_silu_bwd, conv1d_silu_fwd
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.compute_bwd_gdn import _compute_forward_conv_and_states, pure_jax_decoupled_conv1d_gdn
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.pallas_mosaic_tpu_bwd import pallas_gdn_bwd_kernel
+  except (ImportError, ModuleNotFoundError):
+    from .compute_conv1d_bwd import conv1d_silu_bwd, conv1d_silu_fwd
+    from .compute_bwd_gdn import _compute_forward_conv_and_states, pure_jax_decoupled_conv1d_gdn
+    from .pallas_mosaic_tpu_bwd import pallas_gdn_bwd_kernel
+
+
+def decoupled_conv1d_gdn_bwd_kernel(
```
dandragona, line 60: I think this can be dropped? and pallas_mosaic_tpu_bwd entry points moved into this?

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+try:
+  from maxtext.models.kernels.gdn.gdn_bwd.bwd_memory_ref import make_bwd_block_specs
+  from maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import ensure_cpu_interpret_registered
+except (ImportError, ModuleNotFoundError):
+  try:
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.bwd_memory_ref import make_bwd_block_specs
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import ensure_cpu_interpret_registered
+  except (ImportError, ModuleNotFoundError):
+    from .bwd_memory_ref import make_bwd_block_specs
+    from .runtime_utils import ensure_cpu_interpret_registered
```
dandragona, line 42: I think we can remove this exception handling and just import directly

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+
+
+def _bwd_gdn_pipeline_body(
+    qkv_conv_ref: Any,
```
dandragona, line 46: Refactor these parameters into a dataclass like we do in the mhc kernel

src/maxtext/models/kernels/gdn/gdn_bwd/compute_bwd_gdn.py
```diff
@@ -0,0 +1,435 @@
...
+    from maxtext.src.maxtext.layers import normalizations
+  except (ImportError, ModuleNotFoundError):
+    pass
+
+try:
+  from maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import invert_triangular_matrix
+except (ImportError, ModuleNotFoundError):
+  try:
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import invert_triangular_matrix
+  except (ImportError, ModuleNotFoundError):
```
suhithr: We should fail if these don't import

src/maxtext/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+  try:
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.bwd_memory_ref import make_bwd_block_specs
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.runtime_utils import ensure_cpu_interpret_registered
+  except (ImportError, ModuleNotFoundError):
+    from .bwd_memory_ref import make_bwd_block_specs
+    from .runtime_utils import ensure_cpu_interpret_registered
+
+
+def _bwd_gdn_pipeline_body(
+    qkv_conv_ref: Any,
```
suhithr: +1

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,666 @@
...
+  return (
+      dy_conv_flat,
+      d_b_flat,
+      d_a_flat,
+      d_a_log_reduced,
+      d_dt_bias_reduced,
+  )
+
+
+pallas_gdn_bwd_computation = pallas_gdn_bwd_kernel
```
suhithr: Do we need this remapping?

src/maxtext/kernels/gdn/gdn_bwd/bwd_memory_ref.py
```diff
@@ -0,0 +1,103 @@
...
+) -> Tuple[list[pl.BlockSpec], list[pl.BlockSpec], int, int]:
+  """Constructs reverse-scan Pallas emit_pipeline in_specs and out_specs for GDN backward."""
+  del batch_size, kernel_size, pad_len, g
+  if padded_num_v_heads is None:
+    padded_num_v_heads = ((num_v_heads + 127) // 128) * 128
+
+  def rc(c):
+    return num_chunks - 1 - c
+
+  in_specs = [
```
suhithr, line 74: This is a bit challenging to verify, an inline comment/docstring for `in_specs` could help.

src/maxtext/models/kernels/gdn/gdn_bwd/bwd_memory_ref.py
```diff
@@ -0,0 +1,103 @@
...
+    num_v_heads: int,
+    kq_head_dim: int,
+    v_head_dim: int,
+    padded_num_v_heads: int | None = None,
+    g: Any = None,
+    kernel_size: int = 4,
+    pad_len: int = 8,
+) -> Tuple[list[pl.BlockSpec], list[pl.BlockSpec], int, int]:
+  """Constructs reverse-scan Pallas emit_pipeline in_specs and out_specs for GDN backward."""
+  del batch_size, kernel_size, pad_len, g
```
suhithr: if we don't need these could we just not have them as function arguments? I don't see them used

src/maxtext/kernels/gdn/gdn_bwd/jax_compute_gdn_states.py
```diff
@@ -0,0 +1,435 @@
+# Copyright 2026 Google LLC
```
suhithr, line 1: This file seems to mostly do fwd computations, that are used in the backward pass? Perhaps we rename it from `compute_bwd_gdn.py` to something like `jax_compute_gdn_states` or `jax_compute_gdn_chunks` or something of your choosing.

I had some previous comments on refactoring some of these functions, PTAL at those as well.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,666 @@
...
+      a_log_g = a_log[:, vh_start:vh_end]
+    else:
+      a_log_g = a_log[:, :, vh_start:vh_end]
+
+    if dt_bias.ndim == 1:
+      dt_bias_g = dt_bias[vh_start:vh_end]
+    elif dt_bias.ndim == 2:
+      dt_bias_g = dt_bias[:, vh_start:vh_end]
+    else:
+      dt_bias_g = dt_bias[:, :, vh_start:vh_end]
```
suhithr: Are there cases where `dt_bias`, and `a_log` can have these 3 different shapes?
I think we should support only the cases we are using.

src/maxtext/models/kernels/gdn/gdn_bwd/api.py
```diff
@@ -0,0 +1,513 @@
...
+      d_conv_weight,
+      d_conv_bias,
+      d_a_log,
+      d_dt_bias,
+  )
+
+
+decoupled_conv1d_gdn_bwd_computation = decoupled_conv1d_gdn_bwd_kernel
+pallas_fused_conv1d_gdn_bwd_kernel = decoupled_conv1d_gdn_bwd_kernel
+pallas_fused_conv1d_gdn_bwd_computation = decoupled_conv1d_gdn_bwd_kernel
```
suhithr: this renaming does not seem necessary, we could remove these

src/maxtext/models/kernels/gdn/gdn_bwd/api.py
```diff
@@ -0,0 +1,513 @@
...
+      d_recurrent_state_out,
+  )
+
+
+gdn_decoupled_conv1d.defvjp(
+    _gdn_decoupled_conv1d_fwd,
+    _gdn_decoupled_conv1d_bwd,
+)
+
+# Backward compatibility aliases
```
suhithr: As above I don't think we need these aliases, I imagine these are an artifact from testing.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,666 @@
...
+  #
+  # Replacement:
+  #   The 2 lines below:
+  #     dA = jnp.matmul(du, jnp.swapaxes(v_beta, -1, -2)) + jnp.matmul(dw, jnp.swapaxes(k_beta_g, -1, -2))
+  #     dS = jnp.tril(-jnp.matmul(jnp.matmul(A_T, dA), A_T), k=-1)
+  #   can be replaced with:
+  #     dS = jnp.tril(-jnp.matmul(d_v_beta, jnp.swapaxes(v_new, -1, -2)), k=-1)
+  #
+  # This eliminates 3 out of 4 matrix multiplications in this block and removes
+  # the need to allocate intermediate buffer dA in VMEM.
```
suhithr: If we are not using this optimization then perhaps we remove this comment and create a b/ with it?
Or we could use it, eyeballing the comment the math looks fine to me?

src/maxtext/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,666 @@
...
+  tile_v_heads = None
+  for candidate in range(max_possible, 0, -1):
+    if num_v_heads % candidate == 0 and candidate % repeats == 0:
+      tile_v_heads = candidate
+      break
+  if tile_v_heads is None:
+    tile_v_heads = repeats if num_v_heads % repeats == 0 else num_v_heads
+  num_groups = num_v_heads // tile_v_heads
+  tile_kq_heads = tile_v_heads // repeats
+
```
suhithr, line 497: I think we should add something like `validate_inputs` where we can validate shapes, dtypes of inputs to the kernel. The advantage is that it creates an explicit record of what inputs the kernel supports.

see: https://github.com/openxla/tokamax/blob/39e236ed715187e4822a81c5245b0b65d4f8901e/tokamax/_src/ops/experimental/gmm_v2/gmm_v2.py#L1155

suhithr (commented):
Thanks for this, I started looking at the math until `dS` and it is looking good. I appreciate the additional documentation you have provided.

src/maxtext/kernels/gdn/__init__.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2026 Google LLC
```
Rohan-Bierneni: Done. Relocated the entire package from `src/maxtext/models/kernels/gdn/` to `src/maxtext/kernels/gdn/`.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Done. Separated the pure JAX recurrence and chunk state calculation logic into `src/maxtext/kernels/gdn/gdn_bwd/jax_compute_gdn_states.py`, keeping `gdn_bwd_pallas.py` focused strictly on the Pallas kernel dispatch and wrappers.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Done. Factored into a single unified implementation in `jax_compute_gdn_states.py` that accepts an optional precomputed `t_inv`. If `t_inv is None`, it calculates the triangular inverse; otherwise, it directly reuses the cached tensor.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Modified the forward helper to return only `qkv_conv`. `conv_out` is no longer returned or retained in the carry tuple.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Added explicit shape annotations (e.g., `[batch, seq_len, num_k_heads, head_k_dim]`) to all argument docstrings in `gdn_bwd_pallas.py` and `jax_compute_gdn_states.py`.

tests/unit/gdn_bwd_pallas_test.py
Rohan-Bierneni: Done. Factored input generation into reusable `_init_fwd_inputs` and `_init_bwd_inputs` helpers across the tests.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Simplified to always enforce QK L2 normalization with $\epsilon = 1\text{e-}6$ matching the forward pass kernel.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Simplified to always enforce QK L2 norm

tests/unit/gdn_bwd_pallas_test.py
Rohan-Bierneni: Added `@pytest.mark.tpu` and a standalone TPU benchmark in `tests/unit/gdn_benchmark_test.py`. Although still finalizing what exact tests to include since we have others in the downstream stacked prs. For now I have made this fix.

tests/unit/gdn_bwd_pallas_test.py
Rohan-Bierneni: Done. Folded the non-null and non-NaN checks directly into `test_gdn_bwd_gradient_equivalence` assertions against JAX autodiff.

src/maxtext/kernels/gdn/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Local GDN forward kernel package with triangular inverse matrix caching."""
+
+from . import compute_conv1d
```
Rohan-Bierneni: Converted all relative imports to absolute imports (`from maxtext.kernels.gdn...`).

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
Rohan-Bierneni: Replaced wildcard imports with explicit symbol imports

src/maxtext/models/kernels/gdn/gdn_bwd/compute_bwd_gdn.py
Rohan-Bierneni: Same response as above comment removed these fallback blocks.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Removed fallback try/except blocks in internal kernel modules; imports are now direct and will fail fast if modules are missing.

src/maxtext/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
```diff
@@ -0,0 +1,643 @@
...
+  del kernel_size, pad_len
+  c = pl.program_id(1)
+  repeats = num_v_heads // num_kq_heads
+  q_size = num_kq_heads * kq_head_dim
+  k_size = num_kq_heads * kq_head_dim
+  v_size = num_v_heads * v_head_dim
+
+  @pl.when(c == 0)
+  def _init():
+    d_state_scr[...] = jnp.zeros((num_v_heads, kq_head_dim, v_head_dim), dtype=jnp.float32)
```
Rohan-Bierneni: Updated to return the accumulated final recurrent state gradient `d_recurrent_state` rather than an intermediate slice

src/maxtext/kernels/gdn/gdn_bwd/api.py
```diff
@@ -0,0 +1,513 @@
...
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.compute_conv1d_bwd import conv1d_silu_bwd, conv1d_silu_fwd
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.compute_bwd_gdn import _compute_forward_conv_and_states, pure_jax_decoupled_conv1d_gdn
+    from maxtext.src.maxtext.models.kernels.gdn.gdn_bwd.pallas_mosaic_tpu_bwd import pallas_gdn_bwd_kernel
+  except (ImportError, ModuleNotFoundError):
+    from .compute_conv1d_bwd import conv1d_silu_bwd, conv1d_silu_fwd
+    from .compute_bwd_gdn import _compute_forward_conv_and_states, pure_jax_decoupled_conv1d_gdn
+    from .pallas_mosaic_tpu_bwd import pallas_gdn_bwd_kernel
+
+
+def decoupled_conv1d_gdn_bwd_kernel(
```
Rohan-Bierneni: Done. Consolidated entry points into `api.py` and removed redundant wrapper layers

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Removed all the try/except imports

src/maxtext/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Done. Created `GDNBackwardConfig` dataclass in `src/maxtext/kernels/gdn/config.py` bundling `(num_k_heads, num_v_heads, head_k_dim, head_v_dim, conv_kernel_size, chunk_size, compute_dtype)` following the pattern in the MHC kernel.

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Removed the redundant dimension remapping; operands are passed directly in standard layout `[B, S, H, D]`

src/maxtext/kernels/gdn/gdn_bwd/bwd_memory_ref.py
Rohan-Bierneni: Added a detailed docstring explaining the VMEM block spec mappings and grid tiling for `in_specs` and `out_specs`

src/maxtext/models/kernels/gdn/gdn_bwd/bwd_memory_ref.py
Rohan-Bierneni: Pruned unused arguments from the memory reference constructor

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Restricted shape validation to 1D `(num_v_heads,)` matching the qwen3.5 model configuration

src/maxtext/models/kernels/gdn/gdn_bwd/api.py
Rohan-Bierneni: Removed testing aliases and kept only original function names

src/maxtext/models/kernels/gdn/gdn_bwd/pallas_mosaic_tpu_bwd.py
Rohan-Bierneni: Done. Removed the inline comment and filed **b/562944370** to track the decay optimization for future revisions.

Although most likely I will try to implement this fix in this pr and update again

src/maxtext/models/kernels/gdn/gdn_bwd/api.py
Rohan-Bierneni: Removed testing aliases and kept only original function names

src/maxtext/kernels/gdn/gdn_bwd/jax_compute_gdn_states.py
Rohan-Bierneni: Done. Renamed `compute_bwd_gdn.py` to `jax_compute_gdn_states.py` to accurately reflect that it manages the forward recurrence states and chunking representations needed by the backward pass.
Also addressed the associated refactoring comments:
1. Unified `chunk_state_forward_with_cached_tinv` and `chunk_state_forward_with_tinv` into a single consolidated function that takes an optional precomputed `t_inv`.
2. Removed `conv_out` from the return tuple so it returns only `qkv_conv`.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1784 @@
...
+    r_q = jnp.sqrt(jnp.sum(q_orig**2, axis=-1, keepdims=True) + 1e-12)
+    q_unit = q_orig / r_q
+    d_q = (d_q_scaled - q_unit * jnp.sum(d_q_scaled * q_unit, axis=-1, keepdims=True)) / r_q
+
+    r_k = jnp.sqrt(jnp.sum(k_orig**2, axis=-1, keepdims=True) + 1e-12)
+    k_unit = k_orig / r_k
+    d_k = (d_k_proj - k_unit * jnp.sum(d_k_proj * k_unit, axis=-1, keepdims=True)) / r_k
+  else:
+    d_q = d_q_proj * scale
+    d_k = d_k_proj
```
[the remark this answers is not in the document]
Rohan-Bierneni: Done. Aligned the L2 norm backward gradient epsilon to `1e-6` matching `normalizations.l2norm` in the forward pass to guarantee mathematical consistency and training stability.

src/maxtext/models/kernels/gdn/gdn_bwd_pallas.py
```diff
@@ -0,0 +1,1784 @@
...
+      do_5d,
+      chunk_states,
+      t_inv_5d,
+      a_log_3d,
+      dt_bias_3d,
+      reset_hbm,
+  )
+
+  d_a_log_reduced = jnp.sum(d_a_log_chunks[..., 0, :num_v_heads], axis=(0, 1)).astype(a_log.dtype)
+  d_dt_bias_reduced = jnp.sum(d_dt_bias_chunks[..., 0, :num_v_heads], axis=(0, 1)).astype(dt_bias.dtype)
```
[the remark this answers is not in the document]
Rohan-Bierneni: Aligned with Suhith's review feedback to support only the 1D parameter shapes `(num_v_heads,)` actually used by the model architecture, enforced via explicit rank assertions (`assert a_log.ndim == 1` and `assert dt_bias.ndim == 1`)

PR #5158, opened 2026-09-07
Train & Eval on entire dataset when pdbs < 1

muskansh-google:
# Description

This PR fixes a bug where data was silently dropped when running with fractional per-device batch sizes (`per_device_batch_size < 1.0` or `eval_per_device_batch_size < 1.0`). In the past, the dataloader loaded at least one sample per device, but `loss_fn` truncated the input batch down to `micro_batch_size_to_train_on` or `micro_batch_size_to_eval_on`, discarding the remaining samples. With this change, the slicing logic is removed, and all loaded samples are processed across microbatches using gradient accumulation during training and forward-only metric accumulation during evaluation.

### Why This Solution
Instead of discarding data, setting `per_device_batch_size < 1.0` is treated as implicit microbatch accumulation:
- For training, gradient accumulation is automatically computed over `num_microbatches = global_batch_size_to_load // micro_batch_size_to_train_on`. All loaded samples are processed across microbatches and accumulated into the gradients.
- For evaluation, a dedicated forward-only scan loop (`_fractional_batch_eval`) processes each microbatch and aggregates metrics without performing unnecessary backpropagation.
- No data is thrown away, batch size calculations are consistent, and throughput matches the loaded dataset volume.

# Tests

Added unit tests to verify batch sizing calculations and microbatch accumulation:
- `tests/unit/pyconfig_test.py`:
  - `test_calculate_global_batch_sizes_unified`: Verifies calculation logic across fractional batch sizes, gradient accumulation steps, and expansion factors.
  - `test_fractional_per_device_batch_size_calculations`: Verifies config initialization with fractional train and eval batch sizes.
- `tests/unit/gradient_accumulation_nnx_test.py`:
  - `test_fractional_per_device_batch_size_accumulates_all_samples`: Verifies gradient accumulation across all 4 samples without data dropping when `per_device_batch_size = 0.25`.
  - `test_fractional_eval_microbatch_reshaping_and_accumulation`: Verifies microbatch reshaping and metric accumulation for eval without data dropping.

Commands to run the tests:
```bash
pytest tests/unit/pyconfig_test.py -k "calculate_global_batch_sizes_unified or fractional_per_device_batch_size"
pytest tests/unit/gradient_accumulation_nnx_test.py -k "fractional"

### CODE REVIEW

File: src/maxtext/trainers/pre_train/train.py [hunk 1]
```diff
@@ -624,20 +624,96 @@ def diff_wrapper(curr_params, custom_params, rest, config, data):
...
-  if config.mtp_eval_target_module > 0:
-    mtp_acceptance_rate = calculate_mtp_acceptance_rate(aux["intermediate_outputs"], config)
+  def single_eval_fn(d):
+    return loss_fn(
+        eval_model,
+        config,
+        d,
+        eval_rng,
+        eval_params,
+        sparsity_state=sparsity_state,
```
khatwanimohit, line 697: these variables don't exist?

src/maxtext/trainers/pre_train/train.py
```diff
@@ -624,20 +624,96 @@ def diff_wrapper(curr_params, custom_params, rest, config, data):
...
+      "indexer_loss": acc["indexer_loss"] / num_microbatches,
+      "mtp_loss": acc["mtp_loss"] / num_microbatches,
+  }
+  return loss, aux
+
+
 def eval_step(model, config, state, data, dropout_rng=None):
   """eval_step no backprop and new state compared with train_step."""
   del dropout_rng  # unused for NNX (kept for jit signature parity)
   state = nnx.merge(model, state)  # reconstruct TrainStateNNX
```
khatwanimohit, line 688: this state is not used anymore

src/maxtext/trainers/pre_train/train.py
```diff
@@ -140,13 +145,6 @@ def loss_fn(model, config, data, dropout_rng, params, sparsity_state=None, is_tr
       if data[mask_name].shape != target_shape:
         raise ValueError(f"{mask_name} must match targets shape; got {data[mask_name].shape} and {target_shape}")
 
-  # decimate proportion of data when per_device_batch_size<1
```
khatwanimohit, line 143: if `expansion_factor_real_data>1` then removing these lines would include dummy data, can we check if there's a unit test which checks the condition of expansion_factor_real_data > 1

PR #5179, opened 2026-09-09
Fix: Numeric divergence between the mHC layer and its Pallas kernel

denis-mil:
Bug report: https://github.com/AI-Hypercomputer/maxtext/issues/5195

# Description

**Motivation:**
The reference path had its own `mapping` helper that cast `alpha_scale` and
`beta` down to `self.dtype` before computing `sigmoid(alpha_scale * h + beta)`,
while the Pallas path called `common.compute_sigmoid_gate`, which keeps the
scale and bias in float32. With `dtype: bfloat16` the two branches therefore
gated on different numbers, so toggling `use_mhc_pallas_kernel` changed the
layer's output beyond accumulation noise.

**Implementation:**
Drop `ManifoldConstrainedHyperConnections.mapping` and call the shared
`compute_sigmoid_gate` from both pre- and post-mapping, casting the float32
result back to `self.dtype` so the following GEMM is unchanged. Export the
helper from `maxtext.kernels.mhc` so the layer can reach it.

**Tests:**
- `test_pallas_kernel_matches_reference_path` runs the same weights through
  both branches and asserts they agree; the `_tpu` variant repeats it against
  the Mosaic-compiled kernel instead of interpret mode.
- `test_sigmoid_gate_computes_in_float32` pins the helper's precision directly,
  since the end-to-end comparison cannot resolve a bfloat16 regression there.
- Fold the four copies of the interpret-mode kernel patching into the
  `_interpreted_mhc_kernel` context manager, and the repeated module/MLP
  construction into `_build_mhc_and_mlp`.

# Tests

Unit tested on CPU and TPU v6e-1

# Checklist


bvandermoon (approved):

RissyRan (approved):

denis-mil:
@bvandermoon, thanks for the review. There are 4 CI workflows currently awaiting approval. Could you please re-approve them so the updated tests can run?

I’ve fixed the test that failed in CI and verified it on a single-chip TPU and on CPU using `XLA_FLAGS=--xla_force_host_platform_device_count=4` to simulate the CI's `fsdp=4` mesh.

**Root cause of the failure:**
*   **Hardcoded batch:** The CI host has 4 chips (`fsdp=4`, `axes=Auto`), but `_assert_pallas_matches_reference` hardcoded `"per_device_batch_size": 1` (`mhc_test.py:490`).
*   **Dimension mismatch:** `self.x` used the per-device batch size, making the input batch dimension 1.
*   **Sharding conflict:** The MLP output is `(1, 256, 512)`. `MlpBlock` constrains this to `P('fsdp', None, None)` (`linears.py:623`). A batch size of 1 cannot be split across 4 shards, which throws an `IndivisibleError`.
*   **Misleading log:** The shim (`tunix_adapter.py:41-45`) catches the error and attempts to reshard, but it fails because reshard rejects `Auto` axes, ultimately producing the confusing error trace.

denis-mil:
@bvandermoon, the CI `pre-commit` check failed on a file unrelated to this PR (I ran `pre-commit` on all files, and `pyink` simply added a missing blank line). 

There are currently 2 CI workflows awaiting approval. Could you please re-approve them so the updated tests can run? 

Thank you!

matrawybarte-lgtm (commented):
الييبررز

matrawybarte-lgtm:
هل تم انتهاء المشكله

PR #5198, opened 2026-09-11
Packing adapter segment ids

A9isha:
# Description

Sequence packing for the RL trainer landed on main in `ed64d6c7d` (#5100). This PR does not change that implementation. It fixes three adjacent defects on the path packing runs through, adds the test coverage that #5100 went in without, and repairs one pre-existing test break that is unrelated to either.

Three commits, each passing its own tests:

| Commit | Contents |
| --- | --- |
| `f3ba3bbe2` | Repairs `test_shared_types_are_tunix_classes`, broken at base. |
| `ae2e35e1c` | The three fixes, plus the test migration the API removal forces. |
| `c0cbd6cee` | The packing test suite. |

**1. The adapter accepted `attention_mask` and discarded it**, rebuilding an equivalent mask from `pad_id`. That is two independent sources of truth about which tokens are padding, with no mechanism keeping them in agreement: a `pad_id` differing from the one Tunix padded with, or a `pad_id` left `None` for a padded batch, trains on padding and produces well-formed logits with incorrect log-probs. Segment ids now derive from the mask, so there is one source. Priority is explicit segment ids, then the mask, then `pad_id` synthesis.

Nothing is lost by consuming the mask rather than the dense array. `tunix/sft/utils.py:make_causal_attn_mask` builds `input_mask[..., None, :] * tril`, applying the mask on the key side only, so the `[B, L, L]` array carries no more information than a `[B, L]` pad mask. Plumbing the dense array to attention would restrict the model to dot-product attention — flash and splash take a `SequenceDescriptor` built from segment ids — for identical numerics. Packing itself passes `attention_mask=None` and lands on the `pad_id` fallback exactly as before; the unpacked Tunix path is the one whose behaviour changes.

**2. MaxText declared its own `RouterReplayTrainerPayload`**, redeclaring `token_ids`/`token_mask` on the assumption they were inherited. Tunix `984e6ce7` emptied `TrainerPayload` — the arrays live on its subclasses now — so the redeclaration had become the definition, and the payload MaxText read no longer resembled the one Tunix sends. It now consumes Tunix's `RLTrainerPayload`, re-exported from `abstract_engine`.

Two behavioural consequences beyond the type name:

- `inputs_position` was a global cumsum over the mask, so the second sequence in a packed row began where the first ended, and every one of its tokens was rotated by the first sequence's length. The function was already half packing-aware — `targets_segmentation` masked the seams correctly — and the existing test built exactly the packed row that triggers it but asserted only the mask. Positions now restart per segment.
- `segment_positions` was never read. The packed assembler computes it, and re-deriving it produces a second value that can disagree with the positions the rollout used. It is used when present, with the segment-aware derivation as the fallback.

Consuming `RLTrainerPayload` means joining the prompt and completion halves, which is uniform across both assemblers rather than a special case: `SequencePackedBatchAssembler` emits a zero-width prompt with the whole packed row in the completion, so the concatenation is a no-op there. Tunix types `segment_ids` as `[B, T]` or `[B, C]`, which coincide only because of that zero-width prompt; a completion-width array against a real prompt would broadcast or truncate silently, so it is now rejected.

**3. Three engine docstrings described the code inaccurately** once the adapter took packed `segment_ids`. `tokenizer_pad_id`'s docstring and the matching `ValueError` both stated the adapter "cannot build decoder_segment_ids" without it; it can, from two higher-priority sources. `_shard_micro_batch`'s note offered "a sequence-packed micro-batch, always size 1" as its example of a batch dimension that fails to divide the mesh axis, and neither half holds: the distributed packer takes its row count from Tunix's `train_micro_batch_size`, which does default to 1 (`rl_program.py:140`), while the colocated packer sizes rows as `fsdp * dp` (`rl/utils.py:401`) and so divides the axis by construction.

**The separate repair.** `test_shared_types_are_tunix_classes` constructs `RLTrainerPayload(token_ids=, token_mask=, advantages=, loss_mask=)`. Tunix `984e6ce7` deleted three of those four fields, so the test raises `TypeError` against any tunix at or after that commit. CI does not see it: `post_train_github_deps.txt` pins `google-tunix @ 1b0e3c5e` (2 Sep), two days before the refactor and confirmed not an ancestor of it. Anyone developing against tunix HEAD sees it red on main today. The replacement fields (`prompt_ids`, `prompt_mask`, `completion_ids`, `completion_mask`, `advantages`) exist in both versions, so the repair holds at the pin and at tunix main; the same is true of every field the production code reads, so this PR does not require the pin to move.

# Tests

`tests/post_training/unit/maxtext_engine_packing_test.py` is new; three existing suites are extended. 95 tests and 6 subtests pass on CPU in ~170 s:

```
JAX_PLATFORMS=cpu python -m pytest \
  tests/post_training/unit/maxtext_engine_packing_test.py \
  tests/post_training/unit/maxtext_engine_test.py \
  tests/post_training/unit/router_replay_engine_test.py \
  tests/post_training/unit/tunix_adapter_test.py
```

The new suite covers, in order: the adapter contract packing depends on (Tunix gates its entire segment path on an exact-name lookup for a `segment_ids` parameter on the model's `__call__`; the adapter offered `decoder_segment_ids`, so the gate returned False and a packed batch fell back to causal-only masking over the whole row, at a finite loss and normal throughput); packed-versus-unpacked gradients over the whole parameter tree rather than a sample; an absolute hand-computed loss ported from `tunix/tests/rl/algo_core_test.py:49`; the gradient-accumulation identity over four partitions of 8–32 examples; the accumulate-then-apply cadence; and the compiled branch of `fwd_bwd`.

Segment lengths within a packed row are unequal throughout, and that is required rather than incidental: at uniform lengths, an implementation that pools all scored tokens is bit-identical to one that averages per-segment means, so a uniform fixture executes the segmented aggregation on every step without ever distinguishing it. Injecting exactly that mutation passes a uniform fixture and fails the ragged one.

## Results

Three axes, each covering what the other two cannot. All three pass.

**1. Unit tests, CPU.** As above — 17 tests and 6 subtests in the new file, ~126 s. The only axis that runs in CI.

**2. One trainer, two layouts of byte-identical data.** `--compare-layouts` in `compare_tunix_trainer.py`, held on the `packing-compare-harness` branch (see below). 4× TPU v5, Qwen3-0.6B, seq 128, batch 24, 3 steps. The metric is the step-0 gradient's relative L2 distance from the unpacked arm, across Tunix's four packing budgets, against a deliberately broken arm (all tokens in one segment, so attention crosses every boundary) as a negative control:

| | budgets | control | margin |
| --- | --- | --- | --- |
| `MaxTextTrainingEngine`, ragged segments | 6.86e-03 – 7.92e-03 | 5.17e-01 | 65× |
| tunix `PeftTrainer` v2, uniform segments | 1.46e-02 – 1.56e-02 | 5.16e-01 | 33× |

The two controls agree to four significant figures (0.5165510 and 0.5164846), which is the strongest evidence that the harness measures the same quantity on both sides: a step-0 gradient belongs to the model, the data and the loss, not to the trainer.

Read the loss column as provenance only. The broken arm's step-1 loss is −0.055719450 against the unpacked arm's −0.055719163 — agreeing to seven decimals while its gradient is 52% away — because `ref_per_token_logps` comes from the same forward pass the loss then scores, so the contamination cancels from numerator and denominator alike. A packed run can therefore report a plausible loss curve while training on cross-segment attention.

**3. Two trainers, the same packed data.** `--packed --trainer both`, same hardware and model, 12 rows × 2 segments × 128 tokens. `MaxTextTrainingEngine` against tunix `PeftTrainer` v2: effective gradient `rel_l2 = 7.12e-08`, `max_abs_diff = 5.96e-08`, identical loss (−0.055719416588544846 on both), identical accumulated denominator (24.0), gradient norms 8.16370964 and 8.16370869. That is float32 epsilon.

Axis 3 establishes that packing composition — segmented aggregation, the summed denominator, accumulation — is identical across the two trainers. It cannot establish that the attention mask is honoured, because both arms share `TunixMaxTextAdapter` and a gate regression would corrupt them identically. Axis 2 is the arrangement in which a shared defect is visible.

**End-to-end.** The GSM8K distributed GRPO demo was run on a single host through the real `SequencePackedBatchAssembler` (4 rollouts collapsing into 1 packed micro-batch), exercising the assembler seam and the engine's packed path in a real orchestrator. This is plumbing evidence only: every advantage in that run is 0 and so is the loss, which is correct GRPO behaviour when all rewards in a group agree.

**Not measured.** There is no throughput comparison yet; every result above is correctness. Throughput is the reason to pack, so that measurement should follow.

**The TPU harness is not in this PR.** `compare_tunix_trainer.py`, which produces axes 2 and 3, is the single commit of the `packing-compare-harness` branch, stacked on this one so its imports resolve. It is held back because it requires a TPU, has no automated caller, and carries unlanded code from #5060.

# Checklist

### DIFF
--- a/src/maxtext/integration/tunix/tunix_adapter.py
+++ b/src/maxtext/integration/tunix/tunix_adapter.py
@@ -69,6 +69,34 @@ jax.lax.with_sharding_constraint = _compat_wsc
 jax.lax.top_k = _compat_top_k
 
 
+def _segment_ids_from_attention_mask(attention_mask: Array, input_tokens: Array) -> Array:
+  """Recovers Tunix's per-token non-pad mask from its `[B, L, L]` attention mask.
+
+  MaxText takes no precomputed mask: it derives one per layer from
+  `decoder_segment_ids` plus causality, and flash and splash take a
+  `SequenceDescriptor` built from those ids rather than a dense array. Using
+  Tunix's mask therefore means converting it, not passing it through.
+
+  Nothing is lost. `tunix/sft/utils.py:make_causal_attn_mask` builds
+  `input_mask[..., None, :] * tril`, masking the key side only, so the last query
+  row carries the full mask. Deriving the ids here rather than from `pad_id` also
+  leaves one source of truth about which tokens are padding instead of two.
+
+  Raises:
+    ValueError: If `attention_mask` is not `[B, L, L]` against `input_tokens`.
+      Shapes are static, so this raises under `jit` as well as eagerly.
+  """
+  attention_mask = jnp.asarray(attention_mask)
+  batch, seq_len = input_tokens.shape
+  if attention_mask.shape != (batch, seq_len, seq_len):
+    raise ValueError(
+        f"attention_mask has shape {attention_mask.shape}, expected {(batch, seq_len, seq_len)} "
+        f"for input_tokens of shape {input_tokens.shape}. The last query row is read as Tunix's "
+        "per-token non-pad mask, which requires that shape."
+    )
+  return attention_mask[:, -1, :].astype(jnp.int32)
+
+
 class TunixMaxTextAdapter(nnx.Module):
   """Adapter exposing Tunix Trainer call signature over a Transformer model."""
 
@@ -87,14 +115,14 @@ class TunixMaxTextAdapter(nnx.Module):
         use_standalone_mappings,
     )
     self.use_no_op_mappings = use_no_op_mappings
-    # When `pad_id` is provided AND tunix passes `decoder_segment_ids=None`,
-    # synthesize per-token segment_ids (1 for non-pad, 0 for pad). MaxText's
-    # segment-based attention mask then blocks queries at non-pad positions
-    # from attending to pad-position keys. Without this, the adapter forwards
-    # `decoder_segment_ids=None` and MaxText falls back to causal-only masking,
-    # so padded tokens get attended to as if they were real input — silently
-    # corrupting trainer log-probs on every batch. (Rollout-side vLLM does the
-    # right thing because it batches without padding via its scheduler.)
+    # Lowest-priority source of segment ids, used only when Tunix supplies
+    # neither `segment_ids` nor an `attention_mask` (`__call__` has the order).
+    # Synthesizes them per token -- 1 for non-pad, 0 for pad -- so MaxText's
+    # segment-based mask stops non-pad queries attending to pad keys. Without it
+    # the adapter forwards `decoder_segment_ids=None`, MaxText falls back to
+    # causal-only masking, and padding is attended to as real input, silently
+    # corrupting trainer log-probs on every batch. Rollout-side vLLM is
+    # unaffected: its scheduler batches without padding.
     self._pad_id = pad_id
 
   # ------------------------------------------------------------------ #
@@ -118,9 +146,14 @@ class TunixMaxTextAdapter(nnx.Module):
     forwards them only to models whose call signature has a parameter of that
     exact name. They are MaxText's `decoder_segment_ids`; when both are given,
     `segment_ids` wins so packed rows keep per-sequence attention isolation.
+
+    Three sources can supply the segment ids, in descending priority:
+    explicit segment ids, Tunix's `attention_mask`, then `pad_id` synthesis.
     """
     if segment_ids is not None:
       decoder_segment_ids = segment_ids
+    if decoder_segment_ids is None and attention_mask is not None:
+      decoder_segment_ids = _segment_ids_from_attention_mask(attention_mask, input_tokens)
     if decoder_segment_ids is None and self._pad_id is not None:
       decoder_segment_ids = (input_tokens != self._pad_id).astype(jnp.int32)
     logits = self.base(
--- a/src/maxtext/training_engine/maxtext_engine.py
+++ b/src/maxtext/training_engine/maxtext_engine.py
@@ -27,11 +27,9 @@ from typing import Any, Optional
 
 from absl import logging
 from flax import nnx
-from flax import struct
 from flax.linen import partitioning as nn_partitioning
 import jax
 import jax.numpy as jnp
-from jax.typing import ArrayLike  # pylint: disable=g-importing-member
 from maxtext.common import common_types
 from maxtext.common import train_state_nnx
 from maxtext.configs import pyconfig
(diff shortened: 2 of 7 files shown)


YixuanWang-99 (approved):

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -373,64 +359,89 @@ def _to_aval(value: Any) -> Any:
...
+    # pinned JAX.
+    segment_start_count = jax.lax.cummax(jnp.where(starts, running, -1), axis=running.ndim - 1)
+    positions = jnp.maximum(running - segment_start_count, 0).astype(jnp.int32)
+
+  # roll(-1) wraps token 0 into the last position, which is not its real next token, so
+  # mask that position out. Same at every packed segment end: the next token there
+  # belongs to a different sequence.
   targets_segmentation = token_mask.at[:, -1].set(0)
   same_segment = segment_ids[:, :-1] == segment_ids[:, 1:]
   targets_segmentation = targets_segmentation.at[:, :-1].multiply(same_segment.astype(token_mask.dtype))
```
YixuanWang-99, line 460: Should targets_segmentation train with token_mask or loss_mask?

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -373,64 +359,89 @@ def _to_aval(value: Any) -> Any:
...
-      constructor keyword arguments.
-    token_mask: See TrainerPayload.
-    segment_ids: See TrainerPayload.
-    forced_routed_experts: Optional `[batch, seq, top_k]` (or `[batch, seq,
-      num_layers, top_k]` for a distinct routing per layer) array of expert
-      indices to replay, overriding the model's normal top-k routing. `-1` marks
-      a padded/unused slot. See `check_forced_routing_support` in
-      `maxtext.configs.types` for which decoder_blocks accept this.
+def _row_and_mask(payload: abstract_engine.RLTrainerPayload) -> tuple[jax.Array, jax.Array]:
+  """Returns the `[batch, seq]` model row and its mask, prompt half then completion half.
```
igorts-git: I think that the word "half" is misleading here. Half generally means exactly 50%. Maybe call it "parts"

igorts-git (approved):

src/maxtext/integration/tunix/tunix_adapter.py
```diff
@@ -69,6 +69,33 @@ def _compat_top_k(operand, k, axis=-1):
 jax.lax.top_k = _compat_top_k
 
 
+def _segment_ids_from_attention_mask(attention_mask: Array, input_tokens: Array) -> Array:
+  """Recovers Tunix's per-token non-pad mask from its `[B, L, L]` attention mask.
+
+  MaxText takes no precomputed mask: it derives one per layer from
+  `decoder_segment_ids` plus causality, and flash and splash take a
+  `SequenceDescriptor` built from those ids rather than a dense array. Using
```
lc5211, line 77: What's the benefit on doing this?

src/maxtext/training_engine/maxtext_engine.py
A9isha: `targets_segmentation` is the loss mask — it weights the cross-entropy (train.py:254) and is the denominator (train.py:264). So yes, good question, but swapping in `completion_mask` is a no-op on the packed path: `SequencePackedBatchAssembler` ships a zero-width prompt, so `token_mask` already is the action mask.

The real defect is that the mask is applied unshifted against `targets = roll(token_ids, -1)`. Position `i` predicts token `i+1`, so `i` should be scored iff `i+1` is an action token. With `segment_ids=[1,1,1,2,2,2]` and `mask [0,1,1,0,1,1]` we produce `[0,1,0,0,1,0]`; correct is `[1,1,0,1,1,0]` — each segment's first completion token never becomes a target.

Missed because no test supplies `segment_ids` alongside a mask with zeros inside a segment. Without `segment_ids` it falls back to `token_mask`, so mask and segment boundaries coincide and the existing `same_segment` guard is already correct.

Separately, on `PaddedBatchAssembler` the prompt is scored. Whether that's wrong depends on router replay's intended loss.

Fix is the left-shift plus a test with `segment_ids` and interior zeros. I am already planning a follow up PR for  router-replay packed batch.

src/maxtext/training_engine/maxtext_engine.py
A9isha: SG - changed to parts

src/maxtext/integration/tunix/tunix_adapter.py
```diff
@@ -69,6 +69,33 @@ def _compat_top_k(operand, k, axis=-1):
...
+  `input_mask[..., None, :] * tril`, masking the key side only, so the last query
+  row carries the full mask. Deriving the ids here rather than from `pad_id` also
+  leaves one source of truth about which tokens are padding instead of two.
+
+  Raises:
+    ValueError: If `attention_mask` is not `[B, L, L]` against `input_tokens`.
+      Shapes are static, so this raises under `jit` as well as eagerly.
+  """
+  batch, seq_len = input_tokens.shape
+  if attention_mask.shape != (batch, seq_len, seq_len):
```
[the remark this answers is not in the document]
A9isha: done

src/maxtext/integration/tunix/tunix_adapter.py
A9isha: Yes let me share the context. Before this, the adapter accepted `attention_mask` and ignored it; `pad_id` synthesis was the only source of `decoder_segment_ids`. This adapter which originated last year during Tunix MaxText integration has been kept as light weight as possible and reuse as MaxText's pre-existing (pretraining) functionality

PR #5207, opened 2026-09-11
feat(quantization): add native FP8 inference (serve_fp8_weight) for DenseGeneral and GMM v2

Shuwen-Fang:
# Description

Implements native FP8 inference (`quantization="serve_fp8_weight"`) for Qwen3.5 (and other FP8-quantized architectures) to execute forward passes directly using quantized FP8 weights without on-the-fly dequantization ([b/557372531](http://b/557372531)).

Previously, weights stored/loaded in FP8 were dequantized on-the-fly to BF16 prior to compute, causing substantial memory bandwidth and runtime overhead during inference and RL rollouts. This change allows weights to stay in FP8 and compute directly via native FP8 GEMMs across DenseGeneral (supporting per-tensor, per-channel, and block-wise scaling) and RoutedMoE (supporting both tokamax v2 FP8 GMM and vLLM fused MoE via `tpu_inference`).

---

# Benchmark & Performance Results

Evaluated on **TPU v6e-8** (8 chips, GhostLite) using the per-tensor quantized `qwen3.5-35b-a3b-fp8` checkpoint (`/dev/shm/maxtext_qwen3.5_35b_fp8_pertensor_v3`), running a standalone autoregressive decode forward loop simulating RL rollout forward inference (batch size = 1, sequence length = 64 tokens, `scan_layers=True`, `attention="dot_product"`).

### Performance Summary

| Metric | Baseline (`quantization=""`) | Diff (`quantization="serve_fp8_weight"`) | Delta / Speedup |
| :--- | :--- | :--- | :--- |
| **Average Forward Step Time (15 steps)** | **471.59 ms** (±3.07 ms) | **259.80 ms** (±3.25 ms) | **1.81x faster (-45.0% latency)** |
| **Step 0 (JIT Compilation)** | 5.12 s | 1.32 s | **3.88x faster compile** |
| **Output Logits Finite** | Yes (all finite) | Yes (all finite) | Parity verified |

### XProf Trace Viewer Links
* **Baseline Session (`quantization=""`):**  
  http://xprof.corp.google.com/?session_id=shuwenf-5004636990561406008  
  *Session ID:* `shuwenf-5004636990561406008`
* **Diff Session (`quantization="serve_fp8_weight"`):**  
  http://xprof.corp.google.com/?session_id=shuwenf-16130608832799170934  
  *Session ID:* `shuwenf-16130608832799170934`

**Persisted GCS Traces:**
* Baseline: `gs://test-maxtext-output/shuwenf/xprof/20260911_run/xprof/baseline/`
* Diff: `gs://test-maxtext-output/shuwenf/xprof/20260911_run/xprof/diff/`

---

# Tests

### Unit Tests
* `python -m unittest tests/unit/native_fp8_dot_general_test.py` (**23/23 passed** in 42.8s):
  - Per-tensor, per-channel, and block-wise numerical verification against independent references across all Qwen3.5 weight families (`self_attn`, `linear_attn`, MoE expert gate/up/down).
  - Multi-axis output kernel rank-matching tests (e.g. fused QKV `(embed, heads, head_dim)`).
  - `DequantizeWeightPerChannelTest`: square matrix 2D per-channel scaling & MoE 3D per-expert scaling.
  - `MoENativeFp8PerChannelTest`: GMM v2 native FP8 execution with per-channel `qpl.QArray`.
* `pytest tests/unit/moe_test.py -k "Native"`:
  - Validates GMM v2 (per-tensor match and K-axis fallback to dequantize) and fused MoE (`vllm_rpa` standard and prefused weights).
* `python tests/unit/configs_value_test.py` (**3/3 passed** in 7.1s):
  - Validates `serve_fp8_weight` configuration, FP8 weight requirement, and Qwix mutual exclusivity.
* `python -m unittest tests/unit/nnx_decoders_test.py` (**53/53 passed** in 91.9s):
  - Full decoder forward pass and configuration suite.
* Pre-commit hooks clean: `codespell`, `pyink`, `pylint` (10.00/10), and `yamllint` all passed.

### Numerical Parity & Alignment (Native FP8 OFF vs. ON)
Evaluated full `NNXDecoder` forward pass across 128 tokens (`batch_size=4`, `seq_len=32`, `vocab_size=1024`) on TPU v6e comparing Post-PR Native OFF (`quantization=""` / dequantize) vs. Post-PR Native ON (`quantization="serve_fp8_weight"`):

| Metric | Post-PR Native OFF vs. ON |
| :--- | :--- |
| **Mean $D_{\text{KL}}$** | **$4.35 \times 10^{-3}\text{ nats}$** (Max: $1.96 \times 10^{-2}$) |
| **Top-1 Token Match Rate** | **89.06%** (114 / 128 tokens) |
| **Top-5 Candidate Overlap** | **78.64%** |
| **Logit Cosine Similarity** | **0.9956** (99.56%) |
| **SQNR** | **20.60 dB** |

BUGS: b/557372531

# Checklist


src/maxtext/layers/moe.py
```diff
@@ -3283,8 +3271,24 @@ def __call__(
...
+    native_gmm = isinstance(self.quant, quantizations.ServeFp8WeightQuantization) and cfg.sparse_matmul and cfg.use_gmm_v2
+
+    def _maybe_native_gmm_weight(kernel, scale):
+      if native_gmm and ctypes.is_fp8_dtype(kernel.dtype) and scale is not None:
+        # gmm_v2 only natively broadcasts per-tensor/per-channel scale, not a full 2D
+        # block grid; only per_tensor is handled here, rest falls back to dequantize.
+        scheme, _ = quantizations.infer_scale_granularity(scale.shape[1:])
+        if scheme == "per_tensor":
+          per_expert_scale = jnp.max(scale.reshape(scale.shape[0], -1), axis=1).reshape(-1, 1, 1, 1)
+          return qpl.QArray(qvalue=kernel, scale=per_expert_scale)
```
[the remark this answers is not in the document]
Shuwen-Fang: qpl.QArray (the Pallas/gmm_v2 variant used here) doesn't actually require scale.ndim == qvalue.ndim

src/maxtext/layers/quantizations.py
```diff
@@ -87,6 +87,106 @@ def einsum(self, dtype: DType = jnp.float32):
...
+  rhs_axis = contract_ind[0]
+
+  # qwix.QArray requires scale.ndim == qvalue.ndim. DenseGeneral's own true
+  # per-tensor convention (block_size=None) stores a genuine 0-d scalar scale
+  # (linears.py: `resolved_scale_shape = ()`), which is rank-mismatched against
+  # the kernel -- reshape it to an all-1s shape of the kernel's rank first (a
+  # size-1 scalar reshapes losslessly either way); infer_scale_granularity reads
+  # the same on both, since an all-1s shape has no axes with size > 1 either way.
+  if kernel_scale.ndim != quantized_kernel.ndim:
+    kernel_scale = kernel_scale.reshape((1,) * quantized_kernel.ndim)
```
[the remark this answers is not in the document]
Shuwen-Fang: Already handled — the pushed code branches on `kernel_scale.size == 1` (all-1s reshape) vs. the general per-channel case (reshape against the kernel's actual non-contracted output shape, with a ValueError if it doesn't match), rather than unconditionally reshaping to all-1s. See `native_fp8_dot_general` lines 193-208 at HEAD.

snehalv2002:
Some review notes from a pass over this PR. Six items, roughly in order of importance.

### 1. `out_sharding` is dropped in the native FP8 path

In `_compute_dot_general_nnx` (`src/maxtext/layers/linears.py`), the `return quantizations.native_fp8_dot_general(...)` happens before the `if out_sharding is not None:` block, and `native_fp8_dot_general` doesn't take an `out_sharding` argument. `out_sharding` is non-`None` under `ShardMode.EXPLICIT`, which the Qwen3.5 GDN path uses, so this is on the target path rather than a corner case. `matmul_precision` is dropped by the same early return.

Suggest threading `out_sharding` (after `truncate_out_sharding`) and `precision` into `native_fp8_dot_general`, or raising when `out_sharding is not None`.

### 2. The block-wise tests don't exercise the native FP8 kernel

`tests/unit/nnx_decoders_test.py` sets `weight_block_size: 16`. In qwix:

```
qwix/_src/core/dot_general.py:404   MIN_TILE_SIZE_TO_DEQUANT_ON_OUTPUT = 128
qwix/_src/core/dot_general.py:470   if tile_size < MIN_TILE_SIZE_TO_DEQUANT_ON_OUTPUT:
                                      use_fast_dot_general = False
```

With tile size 16 < 128, `qwix.dot_general` dispatches to `_slow_dot_general`, which dequantizes both operands before the matmul. So `test_matches_dequantize_baseline` compares a dequantized forward pass against a dequantized baseline, and native FP8 compute isn't actually covered.

Suggest sizing the test dims to multiples of 128 with `weight_block_size: 128`, and asserting `_fast_dot_general` is reached.

### 3. Silent BF16 fallbacks with no logging

`_maybe_native_gmm_weight` (`src/maxtext/layers/moe.py`) handles only per-tensor and N-axis per-channel scales. A 128x128 block-scaled MoE weight classifies as `block_wise` and falls through to `dequantize_weight`, so with `sparse_matmul=True, use_gmm_v2=True` the expert weights run dequantized BF16. The same applies to multi-axis contractions in `_compute_dot_general_nnx` (the `len(axis) == 1` guard) and to any layer where `kernel_scale is None`.

As it stands there's no way to tell which layers ran native. Suggest a one-time log naming the module and the reason, or an opt-in strict mode that raises.

Question: is block-wise MoE on `gmm_v2` blocked by anything other than `_fwd_prepare_rhs_scale` using `jnp.broadcast_to` where a `jnp.repeat` over the N blocks would work? The kernel looks like it already handles `(E, K_blocks, 1, N)`.

### 4. Activation quantization granularity is tied to the weight scale shape

In `native_fp8_dot_general`, the `per_tensor` branch sets `channelwise_axes = []`, so activations get a single absmax over the entire (B, S, D) tensor. The activation scale factors out of the contraction regardless of the weight's granularity, and the `per_channel` branch immediately below shows per-token works fine.

Suggest always using `channelwise_axes = [d for d in range(inputs.ndim) if d != lhs_axis]` and letting the weight scheme decide only `tiled_axes`.

Some of the reported 20.60 dB SQNR / 89.06% top-1 gap is structural, since the baseline is W8A16 and this is W8A8, but whole-tensor activation scaling is likely costing a few dB on top of that. A per-token variant would separate the two, and a per-layer-family attribution would help given how many layers currently fall back per item 3.

### 5. `infer_scale_granularity` infers from the scale shape alone

It never sees `kernel.shape`, `block_size`, or `contract_ind`. A scale varying only along the contracted axis is returned as `"per_channel"`, which is the opposite of what the name implies, and `moe.py` already has to special-case `channel_axis == 1` to work around it. A block-wise weight with a single block along N (scale shape `(32, 1)`) also classifies as `per_channel`.

Related: the activation tile size comes from the config `block_size` rather than the weight's own scale grid. `quantized_kernel.shape[rhs_axis] // kernel_scale.shape[rhs_axis]` is the one value that can't disagree with the checkpoint, and it's what qwix computes internally at `dot_general.py:469`. That would also cover `in_proj_ba` in `Qwen3NextGatedDeltaNet`, which isn't passed `block_size` (both sibling projections are), so `tiled_axes` becomes `{lhs_axis: None}` under a block-wise checkpoint.

### 6. Test tolerances are loose

The parity assertions use `max|a-b| / max|b| < 0.1`. A 10% max-normalized error will pass with a genuinely mis-scaled layer, and normalizing by the global max hides per-element error. Cosine similarity or SQNR thresholds would be tighter, and the PR description already computes both.

Shuwen-Fang:
> Some review notes from a pass over this PR. Six items, roughly in order of importance.
> 
> ### 1. `out_sharding` is dropped in the native FP8 path
> In `_compute_dot_general_nnx` (`src/maxtext/layers/linears.py`), the `return quantizations.native_fp8_dot_general(...)` happens before the `if out_sharding is not None:` block, and `native_fp8_dot_general` doesn't take an `out_sharding` argument. `out_sharding` is non-`None` under `ShardMode.EXPLICIT`, which the Qwen3.5 GDN path uses, so this is on the target path rather than a corner case. `matmul_precision` is dropped by the same early return.
> 
> Suggest threading `out_sharding` (after `truncate_out_sharding`) and `precision` into `native_fp8_dot_general`, or raising when `out_sharding is not None`.
> 
> ### 2. The block-wise tests don't exercise the native FP8 kernel
> `tests/unit/nnx_decoders_test.py` sets `weight_block_size: 16`. In qwix:
> 
> ```
> qwix/_src/core/dot_general.py:404   MIN_TILE_SIZE_TO_DEQUANT_ON_OUTPUT = 128
> qwix/_src/core/dot_general.py:470   if tile_size < MIN_TILE_SIZE_TO_DEQUANT_ON_OUTPUT:
>                                       use_fast_dot_general = False
> ```
> 
> With tile size 16 < 128, `qwix.dot_general` dispatches to `_slow_dot_general`, which dequantizes both operands before the matmul. So `test_matches_dequantize_baseline` compares a dequantized forward pass against a dequantized baseline, and native FP8 compute isn't actually covered.
> 
> Suggest sizing the test dims to multiples of 128 with `weight_block_size: 128`, and asserting `_fast_dot_general` is reached.
> 
> ### 3. Silent BF16 fallbacks with no logging
> `_maybe_native_gmm_weight` (`src/maxtext/layers/moe.py`) handles only per-tensor and N-axis per-channel scales. A 128x128 block-scaled MoE weight classifies as `block_wise` and falls through to `dequantize_weight`, so with `sparse_matmul=True, use_gmm_v2=True` the expert weights run dequantized BF16. The same applies to multi-axis contractions in `_compute_dot_general_nnx` (the `len(axis) == 1` guard) and to any layer where `kernel_scale is None`.
> 
> As it stands there's no way to tell which layers ran native. Suggest a one-time log naming the module and the reason, or an opt-in strict mode that raises.
> 
> Question: is block-wise MoE on `gmm_v2` blocked by anything other than `_fwd_prepare_rhs_scale` using `jnp.broadcast_to` where a `jnp.repeat` over the N blocks would work? The kernel looks like it already handles `(E, K_blocks, 1, N)`.
> 
> ### 4. Activation quantization granularity is tied to the weight scale shape
> In `native_fp8_dot_general`, the `per_tensor` branch sets `channelwise_axes = []`, so activations get a single absmax over the entire (B, S, D) tensor. The activation scale factors out of the contraction regardless of the weight's granularity, and the `per_channel` branch immediately below shows per-token works fine.
> 
> Suggest always using `channelwise_axes = [d for d in range(inputs.ndim) if d != lhs_axis]` and letting the weight scheme decide only `tiled_axes`.
> 
> Some of the reported 20.60 dB SQNR / 89.06% top-1 gap is structural, since the baseline is W8A16 and this is W8A8, but whole-tensor activation scaling is likely costing a few dB on top of that. A per-token variant would separate the two, and a per-layer-family attribution would help given how many layers currently fall back per item 3.
> 
> ### 5. `infer_scale_granularity` infers from the scale shape alone
> It never sees `kernel.shape`, `block_size`, or `contract_ind`. A scale varying only along the contracted axis is returned as `"per_channel"`, which is the opposite of what the name implies, and `moe.py` already has to special-case `channel_axis == 1` to work around it. A block-wise weight with a single block along N (scale shape `(32, 1)`) also classifies as `per_channel`.
> 
> Related: the activation tile size comes from the config `block_size` rather than the weight's own scale grid. `quantized_kernel.shape[rhs_axis] // kernel_scale.shape[rhs_axis]` is the one value that can't disagree with the checkpoint, and it's what qwix computes internally at `dot_general.py:469`. That would also cover `in_proj_ba` in `Qwen3NextGatedDeltaNet`, which isn't passed `block_size` (both sibling projections are), so `tiled_axes` becomes `{lhs_axis: None}` under a block-wise checkpoint.
> 
> ### 6. Test tolerances are loose
> The parity assertions use `max|a-b| / max|b| < 0.1`. A 10% max-normalized error will pass with a genuinely mis-scaled layer, and normalizing by the global max hides per-element error. Cosine similarity or SQNR thresholds would be tighter, and the PR description already computes both.

 * **1 (`out_sharding` / `precision`)**: Added! Moved `truncate_out_sharding` before the native FP8 check,
  threaded `out_sharding` and `precision` into `native_fp8_dot_general` and added unit tests in `tests/unit/native_fp8_dot_general_test.py`, good catch.
    * **2 & 3 (Block-wise execution & fallbacks)**: This PR is scoped to native FP8 execution for per-
  tensor and per-channel weights. 
    * **4 (Activation scaling)**: Kept aligned with current dynamic FP8 quantization standards for now; per-
  token activation experiments can be evaluated alongside block-wise.
    * **5 & 6 (Scale inference & tolerances)**: In practice, `infer_scale_granularity` handles all current
  target checkpoint layouts. The test tolerances match existing MaxText FP8 vs BF16 parity thresholds.

snehalv2002 (approved):

src/maxtext/configs/base.yml
```diff
@@ -132,6 +132,7 @@ dtype: "bfloat16"
 # 'fp8' for 8-bit floating-point gemms on nvidia gpus.
 # 'nanoo_fp8' for 8-bit floating-point gemms on amd mi300/mi325 gpus.
```
gobbleturk, line 133: I am unclear what should be specified in the quantization vs weight_dtype

src/maxtext/layers/linears.py
```diff
@@ -116,6 +116,8 @@ def _compute_dot_general_nnx(
     out_sharding: NamedSharding | None = None,
     kernel_scale: Array | None = None,
     compute_dtype: DType | None = None,
+    native_fp8_compute: bool = False,
+    block_size: int | tuple[int, ...] | None = None,
```
gobbleturk: can we name it quantize_block_size or scale_block_size?

gobbleturk (approved):

src/maxtext/configs/base.yml
Shuwen-Fang: weight dtype is the full precision dtype for weights that we are not quantizing (e.g. parts of the attention module for qwen 3.5)

src/maxtext/layers/linears.py
Shuwen-Fang: updated

PR #5172, opened 2026-09-09
[Raiden Weight Sync 7/7] Clean up rollout path and remove redundant weight sync logic

YixuanWang-99:
## Overview
Part of the stacked Raiden weight-sync enablement PR chain replacing #5089.

**Stack:**
- [M7] https://github.com/AI-Hypercomputer/maxtext/pull/5166: Cross-repo drift guard
- [M1] https://github.com/AI-Hypercomputer/maxtext/pull/5167: Inhomogeneous layer cycle support in raiden_unscan
- [M2] https://github.com/AI-Hypercomputer/maxtext/pull/5168: MoE 128-lane layout and padding (pairs with Tunix T5)
- [M3] https://github.com/AI-Hypercomputer/maxtext/pull/5169: Target-free streaming weight conversion
- [M5] https://github.com/AI-Hypercomputer/maxtext/pull/5170: Shared expert gate weight_dtype fix
- [M4] https://github.com/AI-Hypercomputer/maxtext/pull/5171: TrainingEngine Raiden FFI integration (pairs with Tunix T2a)
- **[M6] AI-Hypercomputer/maxtext (this PR)**: Rollout cleanup

## Details
- Net −260 lines.
- Removes deprecated and redundant rollout and weight sync duplication in `maxtext_vllm_rollout.py`, now that the logic is cleanly owned by M3/M4.
- Remaps hybrid KV cache groups in `adapter.py` using `layer_name_to_kvcache_index` before passing to decoder layers.
- Adds `rollout_tensor_parallelism` field to VLLM config.
- To keep CI happy we also update the google-tunix commit hash in `post_train_github_deps.txt`. This pulls in the latest API changes to the raiden_synchronizer from Tunix that are required for unit tests to pass.
- 
# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code. For an optional AI review, add the `gemini-review` label.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files).

PR #5216, opened 2026-09-14
fix(rl): restore MoE `wo` in tunix's zero-pad key set

chiajunglien:
## Description

The Gemma4-26B and Qwen3-30B-A3B RL e2e pipelines fail during the **initial weight sync into the vLLM rollout**, before any RL step runs, with a Tunix `ShapeMismatchError` on the MoE `wo` weight. This restores the one-key patch that makes Tunix zero-pad `wo` instead of trying to repeat it.

```                                                                                                                  
tunix.generate.utils.ShapeMismatchError: Cannot align axis 1 for                                                                       
decoder.layers_0.mlp.moe_block.MoeBlock_0.wo: src=704 -> tgt=1024 is not an                                                            
integer multiple and the key is not a recognized MoE pattern.                                                                          
```

Affected nightly tasks:                                                                                                                
- `maxtext_e2e_tpu_post_training.rl-gemma4-26b.wait_for_workload_completion`                                                           
- `maxtext_e2e_tpu_post_training.rl-qwen3-30b.wait_for_workload_completion`

[DAG Error Log](https://efdb2a1d6c2c435b8b7de77690c286a9-dot-us-central1.composer.googleusercontent.com/dags/maxtext_e2e_tpu_post_training/grid?tab=logs&task_id=gemma4-26b.rl-gemma4-26b.rl-v5p-32.run_model.wait_for_workload_completion&dag_run_id=auto_gha__34793910551__2026-09-14T00%3A56%3A48Z__post_training)

## Root Cause

`tunix.generate.utils._align_per_axis` resolves every shape mismatch down one of two branches: MoE MLP weights are **zero-padded**, everything else is **repeated** (KV-head expansion). The branch is chosen by a leaf-key lookup, and Tunix's set omits `wo` even though its own docstring lists it:


```python                                                                                                                              
# tunix/generate/utils.py (pinned a8d7058, and google/tunix main)                                                                      
_MOE_MLP_WEIGHTS = frozenset({'wi', 'wi_0', 'wi_1'})   # `wo` missing                                                                  
``` 

MaxText's own copy of that set (`integration/vllm/convert_utils.py`) is correct and includes `wo`.

Only `wo` fails: both pipelines run with `prefuse_moe_weights=true`, so the target carries a fused `wi`, and `wi` is already in Tunix's set.                                                                              
                                                                                                                                           
This is a regression. #5025 patched the key set inside `train_rl.py::_tpu_inference_compat_patches()`; #5172 (`d043c5b8e`) removed that context manager in favour of the Raiden FFI path. But RL does not use that path yet — `MaxTextTrainingEngine.prepare_weight_sync` currently has no production caller, and `train_rl.py` still syncs through `MaxTextVllmRollout` → `VllmSampler.load_checkpoint` → Tunix `transfer_state_directly`. The shim was removed one step before its replacement was wired up.

## Solution

Re-apply the patch in `maxtext_vllm_rollout.py`, at module scope. Two deliberate differences from #5025:

- **Patch Location**: Moved the patch into the module that constructs the Tunix sampler rather than relying on `train_rl.py`'s (now removed) context manager. Module-level scope is required because the initial weight sync occurs inside `MaxTextVllmRollout.__init__` via `load_checkpoint`.
- **Dynamic Key Alignment (No Hardcoded Keys)**: Instead of hardcoding keys, it unions Tunix's set with MaxText's own `convert_utils.MOE_MLP_WEIGHTS`. This ensures both stay synchronized and allows the patch to no-op automatically once upstream `google-tunix` carries the fix (or if the internal symbol is renamed, which gracefully logs a warning and skips).


# Tests

Gemma4-26b RL E2E: https://cloudlogging.app.goo.gl/ryN9cRnkrcNzWNG38

Qwen3-30b RL E2E: https://cloudlogging.app.goo.gl/rNo1dZPCPDmAGR2S6

```bash
JAX_PLATFORMS=cpu pytest tests/post_training/unit/vllm_rollout_tunix_patch_test.py -v
```

# Checklist


YixuanWang-99:
Could you add a unit test that fails if this fix is removed?

#5172 deleted the previous version of it and CI stayed green, let's guard it with a unit test.

YixuanWang-99 (approved):

hengtaoguo (approved):

chiajunglien:
> Could you add a unit test that fails if this fix is removed?
> 
> #5172 deleted the previous version of it and CI stayed green, let's guard it with a unit test.

Done — added `tests/post_training/unit/vllm_rollout_tunix_patch_test.py`.

PR #5236, opened 2026-09-15
Plumb tunix gc_collect_after_weight_sync into the RL trainer

gutianyu-google:
## What

tunix exposes `ClusterConfig.gc_collect_after_weight_sync` (skip the full host `gc.collect()` after every weight sync). It cannot be set from a MaxText RL config, so the RL trainer always runs the per-step collection.

Adds the `RLConfig` field (default equals tunix's default, so existing configs are unchanged) and forwards it to the tunix `ClusterConfig`, filtered through `_kwargs_supported_by`: kwargs the installed tunix's constructor does not accept (checked via `inspect.signature`) are dropped with a log line instead of failing, so the pinned tunix (`post_train_github_deps.txt`, a8d70582, which predates the field) keeps working; once the pin is bumped the setting takes effect without further changes.

Also logs the resolved `GrpoConfig` on the standard (non-agentic) path, matching the agentic path, so the applied knob values are visible in the logs.

*Split out of the original two-knob version of this PR per review: the `reward_num_workers` half needs a picklable reward-fn wrapper to actually take effect (the current closure makes tunix's reward pool fall back to serial with a `PicklingError`), and follows in its own PR.*

## Effect

On a Qwen3-0.6B GRPO run the gc knob was worth ~1.2 s per step.

## Tests

`tests/post_training/unit/train_rl_test.py`: the new field defaults to tunix's default and accepts overrides (also declared in `configs/post_train/rl.yml`); `_kwargs_supported_by` additionally passes everything through for classes taking `**kwargs` or with no inspectable signature; `_kwargs_supported_by` keeps accepted kwargs (dataclass or plain class), drops unknown ones with a log line, and stays silent when nothing is dropped.

```
PYTHONPATH=src python -m pytest tests/post_training/unit/train_rl_test.py -k "tunix_knob or kwargs_supported"
```

## Checklist

- [x] I have added tests that prove the change works.
- [x] Defaults leave existing configs untouched.


src/maxtext/trainers/post_train/rl/train_rl.py
```diff
@@ -304,6 +304,19 @@ def build_reward_fns(trainer_config: Any, make_reward_fn: Callable) -> list:
...
+  """Keeps only the kwargs that the installed tunix `cls` declares.
+
+  Lets MaxText forward newer tunix knobs while still running against an older
+  pinned tunix that does not know them; dropped settings are logged.
+  """
+  names = {f.name for f in dataclasses.fields(cls)}
+  dropped = sorted(k for k in kwargs if k not in names)
+  if dropped:
+    max_logging.log(f"Installed tunix {cls.__name__} does not support {dropped}; those settings are ignored.")
+  return {k: v for k, v in kwargs.items() if k in names}
```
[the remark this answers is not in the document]
gutianyu-google: Done.

khatwanimohit (approved):

src/maxtext/configs/types.py
```diff
@@ -2785,6 +2785,14 @@ class RLCluster(BaseModel):
   use_pathways_reshard: bool = Field(
       True, description="Legacy experimental GRPO: use Pathways resharding to move policy params to the sampler."
   )
+  gc_collect_after_weight_sync: bool = Field(
+      True,
+      description=(
+          "Run a full host gc.collect() after every trainer->sampler weight sync "
+          "(tunix ClusterConfig.gc_collect_after_weight_sync). Disable on colocated setups where the "
```
A9isha: tunix's own docstring is explicit that the weight-sync copies accumulate and the run can OOM over time, and that disabling is only safe "provided the memory headroom is there." Colocated is the setup with the least headroom, so as written it points people at an OOM a few hundred steps in. Worth naming the consequence in both types.py and rl.yml.

tests/unit/csa_streamindex_test.py
```diff
@@ -346,4 +346,5 @@ def interpret_kernel_fn(q, compressed, weights, softmax_scale, *_args):
 
 if __name__ == "__main__":
   from absl.testing import absltest
+
```
A9isha, line 349: nit: please remove this

A9isha (approved):

src/maxtext/configs/types.py
gutianyu-google: Done.

tests/unit/csa_streamindex_test.py
gutianyu-google: Done.

PR #5239, opened 2026-09-15
Fix Qwen weight unrolling on MaxText

niting:
# Description

A recent change in MaxText caused the weight unrolling logic for Qwen models to be broken. This is needed if MaxText trainer runs against a VLLM sampler where scanned checkpoint weights are unscanned/unrolled before running on vllm. This adds back the calls to the unroll functions - which are still available.

We also fix some issues with dimension mismatches. We were using unpadded dimensions where padded dimensions should have been used and this causes responses from the model to be gibberish. 

# Tests

Ran Qwen3.5-35B on DeepSWE stack against TPU v5p-256. This seemed to output coherent responses.

# Checklist


khatwanimohit (approved):

SurbhiJainUSC (approved):

src/maxtext/layers/moe.py
```diff
@@ -689,9 +687,9 @@ def __init__(
       # During aqt convert state we delete kernel weight from params to save
       # memory. Instead they are retrieved from the tensors stored in the 'aqt'
       # collection.
-      self.wi_0 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
-      self.wi_1 = jnp.zeros((num_experts, self.moe_expert_input_dim, intermediate_dim))
-      self.wo = jnp.zeros((num_experts, intermediate_dim, self.moe_expert_input_dim))
+      self.wi_0 = jnp.zeros((num_experts, self.moe_expert_input_dim, moe_intermediate_dim))
+      self.wi_1 = jnp.zeros((num_experts, self.moe_expert_input_dim, moe_intermediate_dim))
+      self.wo = jnp.zeros((num_experts, self.intermediate_dim, self.moe_expert_input_dim))
```
YixuanWang-99, line 692: Do you think wo should be `self.wo = jnp.zeros((num_experts, moe_intermediate_dim, self.moe_expert_input_dim))`? 

I thought the down-projection kernel wo maps from the intermediate activation space back to moe_expert_input_dim. Its contracting dimension must equal the output dimension of wi (i.e., moe_intermediate_dim).

src/maxtext/layers/moe.py
```diff
@@ -708,7 +706,7 @@ def __init__(
               self.rngs.params(),
               (
                   self.num_experts,
-                  moe_intermediate_dim,
+                  self.intermediate_dim,
                   self.moe_expert_input_dim,
```
YixuanWang-99, line 710: Similar to above, might use `moe_intermediate_dim` instead

src/maxtext/layers/moe.py
YixuanWang-99, line 684: we defined the moe_intermediate_dim here

YixuanWang-99:
I am bit worried about the scenario that `padded_base_moe_mlp_dim` is explicitly set (e.g., padding intermediate dimensions like 1408 to 1536 to satisfy GMM v2 TPU kernel alignment requirements). But we can deprioritize the concerns for now

YixuanWang-99 (approved):

src/maxtext/layers/moe.py
khatwanimohit, line 692: I think we can remove these set of changes cause they are under quantization block and that too for aqt quantization, we'll not fall under these conditions for RL workload

PR #5259, opened 2026-09-17
feat: add moe_spread_experts_over_fsdp to enable spreading MoE experts over FSDP axis

systalyze-ai:
## Description

This PR adds support for spreading MoE expert weights over both the `expert` and `fsdp` mesh axes, enabling better memory efficiency when combining Mixture of Experts (MoE) with FSDP parallelism.

### Problem Solved

Under base rules, running an MoE model with FSDP > 1 either:
1. Replicates the expert stack across FSDP shards (embed-dim sharding only, OOM at scale), or
2. Collides logical axes that both resolve to the fsdp mesh axis (duplicate-axis sharding error)

### Solution

Spreads the expert dimension over [expert, fsdp] and releases the MoE embed dimension from fsdp so no tensor claims the same mesh axis twice. Unlike `shard_exp_on_fsdp` (which requires `ici_expert_parallelism = ici_tensor_parallelism = 1`), this composes with expert/tensor parallelism.

### Implementation Details

The feature provides three modes controlled by `moe_spread_experts_over_fsdp`:

- **'never'** (default): Keep base rules, preserves existing behavior
- **'auto'**: Apply when `num_experts` is divisible by `(ici_expert_parallelism * ici_fsdp_parallelism)`
- **'always'**: Apply and fail loudly if divisibility requirement not met

The implementation:
- Adds validation for divisibility checks
- Handles auto-fill case (fsdp == -1) by declining in 'auto' mode, proceeding in 'always' mode
- Transforms logical_axis_rules dynamically during config initialization
- Includes proper logging for debugging

### Changes

1. **src/maxtext/configs/base.yml** - Added `moe_spread_experts_over_fsdp` config parameter with documentation
2. **src/maxtext/configs/types.py** - Added `moe_spread_experts_over_fsdp` field to `MoEGeneral` pydantic model
3. **src/maxtext/configs/pyconfig.py** - Implemented `_maybe_spread_moe_experts_over_fsdp()` function with full validation logic
4. **tests/unit/pyconfig_test.py** - Added 14 comprehensive unit tests

## Tests

Comprehensive test coverage includes:

- ✓ All three modes ('never', 'auto', 'always')
- ✓ Divisibility validation (pass and fail cases)
- ✓ Auto-fill FSDP handling (-1 case)
- ✓ Edge cases (non-MoE models, trivial FSDP axis)
- ✓ Invalid mode error handling
- ✓ Both tuple and string axis format support

All syntax validated (Python and YAML). Tests can be run with:
```bash
python -m pytest tests/unit/pyconfig_test.py::PyconfigTest -k "moe_spread" -v
```

Checklist


gobbleturk:
Does using the custom rule custom_rule=shard-exp-on-fsdp https://github.com/AI-Hypercomputer/maxtext/blob/09e98cf342c6d346ed87439b0bf53b3f215bbb70/src/maxtext/configs/custom_mesh_and_rule/shard-exp-on-fsdp.yml not work? I think it makes the same logical axis rule change as in your PR. If FSDP + EP doesn't work with this custom rule, is there a way to fix it instead of creating a new feature?

src/maxtext/configs/pyconfig.py
```diff
@@ -185,6 +185,66 @@ def _apply_rules(base_rules, new_rules, config):
   return _merge_logical_axis_rules(base_rules, new_rules)
 
 
+def _maybe_spread_moe_experts_over_fsdp(raw_keys):
+  """Rewrites logical_axis_rules so MoE expert weights shard over [expert, fsdp].
+
+  Under base rules, running an MoE model with fsdp > 1 either replicates the expert
+  stack across the fsdp shards (embed-dim sharding only, OOM at scale) or resolves
```
gobbleturk, line 192: I'll note sharding on embed dimension or expert dimension does not change the memory pressure, either way each shard has a memory of weight_size/(FSDP * EP) before the FSDP all gather and weight_size/EP after the FSDP all gather, which dimension you shard on does not affect size. It does affect some copy overheads and muon though

PR #5272, opened 2026-09-18
Fix pipeline_utils mutating axes_to_remove and add unit tests

harshaa765:
# Description

Fixes `remove_gathered_mesh_axes` in `utils/pipeline_utils.py` mutating its `axes_to_remove` argument, and adds the first direct unit tests for `pipeline_utils`.

**The bug.** For every non-MoE leaf, `remove_gathered_mesh_axes` did `axes_to_remove.append("expert")` on the caller's list. `derive_stage_weight_partition_specs` (called from `NNXCircularPipeline.from_repeat_weights_to_bsw`) passes that same list to every leaf, so once any non-MoE leaf has been visited, `"expert"` is also stripped from the routed MoE block (`MoeBlock_0`) of every leaf visited afterwards. That contradicts the docstring ("The 'expert' mesh axis is only retained for the routed block").

**When it shows up.** JAX flattens dicts in sorted key order. In a single decoder layer `MoeBlock_0` sorts before its lowercase siblings (`self_attention`, `mlp`, the norms, `shared_experts`), so it is visited before the list gets polluted, which is why this is easy to miss. With `num_layers_per_pipeline_stage > 1` the stage is an `NNXSequentialPipelineStage` with children `layers_0`, `layers_1`, ...; `layers_0`'s non-MoE leaves are then visited before `layers_1`'s `MoeBlock_0`, and the routed-expert weights of every sub-layer after the first lose their `expert` axis in the BSW partition specs. Reproduction at the `PartitionSpec` level, using a Mixtral-style layer:

```
num_layers_per_pipeline_stage=2
  layers_0 MoeBlock_0/wi_0 -> P('expert', None, 'tensor')
  layers_1 MoeBlock_0/wi_0 -> P(None, None, 'tensor')   # expected P('expert', None, 'tensor')
```

**The fix.** Build a new list (`[*axes_to_remove, "expert"]`) instead of appending to the argument. Results are unchanged for single-layer stages and for non-MoE models; only the multiple-sub-layers-per-stage MoE case changes.

I have only reproduced this at the spec level. I have not run a multi-layer-per-stage MoE pipeline on TPU/GPU, so I can't quantify the runtime effect (I'd expect extra gathering and memory for the expert weights).

**Tests.** `pipeline_utils.py` had no direct unit tests. `tests/unit/pipeline_utils_test.py` (CPU only, `cpu_only` marker) covers:
- the partition-spec helpers (`get_mesh_axis_dim_indices`, `remove_gathered_mesh_axes`, `derive_stage_weight_partition_specs`, `strip_pipeline_repeat_logical_axis`, `is_spec_leaf`), including regression cases for the bug above;
- `is_static_param`, `flatten_nnx_state` / `unflatten_nnx_state`, `advance_rng_state`, the Linen-collection helpers and `PipelineContext`;
- `create_gradient_accumulation_scan`: outputs and gradients (eager and under `jit`) match a plain `jax.lax.scan` over the same step, including the accumulated `bsw` cotangent.

# Tests

```
python -m pytest tests/unit/pipeline_utils_test.py
```

- 44 tests pass on CPU (jax 0.11.1, flax 0.12.9, Python 3.12).
- With only the `pipeline_utils.py` change reverted, 5 failures appear (subtests counted separately): the `non_moe_block` case of `RemoveGatheredMeshAxesTest.test_does_not_modify_axes_to_remove`, `DeriveStageWeightPartitionSpecsTest.test_does_not_modify_axes_to_remove`, the `layers_1` and `layers_2` subtests of `test_moe_block_keeps_expert_axis_in_every_sublayer_of_a_stage`, and `test_moe_result_does_not_depend_on_sibling_traversal_order_sibling_sorts_before`. The rest pass either way.
- `pyink` (indentation 2, line length 122), `pylint` (flags from `.pre-commit-config.yaml`) and `codespell` are clean on both files.
- Not run: `tests/integration/pipeline_parallelism_test.py` and any TPU/GPU workload (they need the full dependency stack and accelerators).

# Checklist

- [X] I have run end-to-end tests tests and provided workload links above if applicable. (N/A: no end-to-end or accelerator run; the change is covered by the CPU unit tests above.)
- [X] I have made or will make corresponding changes to the doc if needed, including adding new documentation pages to the relevant Table of Contents (toctree directive) as explained in [our documentation](https://maxtext.readthedocs.io/en/latest/development.html#adding-new-documentation-files). (N/A: no user-facing behavior change.)


harshaa765:
@SurbhiJainUSC can you  please review this PR?

PR #5276, opened 2026-09-18
Resolve kv_tp_size and moe_mlp_tp_size with fallback to rollout TP.

niting:
In the existing state, TP was not taking effect as the default value of 1 always evaluted to True. Fixes the computation to utilize a value overriden by the user.

I ran this change on a v5p cluster for Qwen3.5-35B and it seems to work. Witohut this change, we were generating gibberish responses.


src/maxtext/integration/vllm/convert_utils.py
```diff
@@ -744,3 +744,31 @@ def resolve_prefuse_moe_weights(config: Any, prefuse_moe_weights: Optional[bool]
...
+  """Resolves rollout KV TP from explicit override, config, or rollout TP fallback."""
+  if kv_tp_size is not None and int(kv_tp_size) > 0:
+    return int(kv_tp_size)
+
+  if config is not None:
+    raw = getattr(config, "kv_tp_size", 0)
+    if raw and int(raw) > 0:
+      return int(raw)
+
+  resolved_tp = resolve_rollout_tp(config, tp)
```
YixuanWang-99, line 759: What if rollout_expert_parallelism (ep)>1?

based on my understanding, kv_tp_size should be  tp * ep

igorts-git (approved):

PR #5285, opened 2026-09-18
Fix issues with trainable_parameters_mask in MaxTextTrainingEngine

niting:
# Description

When `trainable_parameters_mask` is specified, `get_optimizer()` wraps the optimizer with `optax.multi_transform` to map non-matching parameters to `optax.set_to_zero()`. While this prevents updates to frozen weights, autodiff still computes non-zero gradients for all parameters.

In `MaxTextTrainingEngine._update_kernel`, `grad_norm` and `apply_gradient_clipping` were computed over the unmasked gradients before the optimizer zeroed the updates. On models with frozen components (such as freezing router gate kernels in MoE RL training), this caused:
1. `grad_norm` to be inflated by frozen weights.
2. `apply_gradient_clipping` to aggressively over-clip trainable weights based on the inflated global norm.
3. Spikes on frozen parameters to trigger false-positive step skips when `skip_step_on_spikes` is enabled.

This change:
- Masks out frozen parameter gradients in `_update_kernel` before computing `grad_norm` and applying gradient clipping.
- Caches `_freeze_mask_fn` and `_freeze_mask` across eager and compiled execution paths in `MaxTextTrainingEngine`.
- Extracts `apply_trainable_parameters_mask` in `optimizers.py` with diagnostic logging of frozen parameter counts and sample paths.
- Adds unit tests in `maxtext_engine_test.py` covering eager and compiled execution with `trainable_parameters_mask`.

# Tests

Tests were added.

# Checklist


src/maxtext/optimizers/optimizers.py
```diff
@@ -258,20 +259,57 @@ def get_optimizer(config, learning_rate_schedule, model=None, mesh=None):
...
   # If a whitelist of trainable parameters is provided, freeze everything else.
+  return apply_trainable_parameters_mask(base_opt, config)
+
+
+def apply_trainable_parameters_mask(base_opt, config):
+  """Freeze every parameter not matched by `config.trainable_parameters_mask`.
+
+  Returns `base_opt` unchanged when the mask is empty, so callers can apply this
+  unconditionally. Wrap only the optimizer itself, not a chain that includes
+  gradient clipping: clipping must see the whole gradient tree so the global
```
A9isha, line 270: >clipping must see the whole gradient 

This states the opposite of what the PR does — the change in _update_kernel exists so the global norm is not the same with and without freezing. Please rewrite this

tests/post_training/unit/maxtext_engine_test.py
```diff
@@ -252,6 +253,87 @@ def test_gradient_norm_is_over_the_accumulated_normalized_gradient(self):
...
+      t.update()
+
+      np.testing.assert_allclose(np.asarray(t.model.router_gate.get_value()), [10.0, 10.0])
+      self.assertFalse(np.allclose(np.asarray(t.model.weights.get_value()), [1.0, 2.0]))
+
+      metrics = t.get_metrics(clear_cache=True)
+      recorded_norm = float(np.asarray(metrics.scalar_metrics["gradient_norm"]).reshape(-1)[0])
+      # Without masking, the norm would include router_gate: sqrt(1^2 + 1^2 + 5^2 + 5^2) = sqrt(52) = 7.211.
+      # With masking, it reflects only the weights (around sqrt(2) = 1.414).
+      self.assertLess(recorded_norm, 2.0)
```
A9isha, line 294: Both values are exactly determined and identical in eager and compiled mode — measurable at  `gradient_norm == 1.4142135` and `post-update weights == [0.92928934, 1.9292893]`. `assertLess(recorded_norm, 2.0)` passes for any implementation that merely shrinks the norm; please use `assert_allclose` against `np.sqrt(2)` and the exact weights.

Also: nothing asserts the over-clipping fix or sets `skip_step_on_spikes=true`, so two of the three problems listed in the description are untested. The two tests are identical apart from t.compile(payload), including a duplicated MoEModel — worth parameterizing.

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -1230,6 +1235,14 @@ def _update_kernel(self, state_pure, accumulated_grads, accumulated_denominator,
           lambda g: jnp.where(has_weights, g / safe_denominator.astype(g.dtype), jnp.zeros_like(g)),
           accumulated_grads,
       )
+      freeze_mask = self._freeze_mask
```
A9isha, line 1238: Compiled is the path that matters here, and _compile_for_batch already populates self._freeze_mask before the trace, so the eager gaps can wait. Rather than fix them in this PR, please leave a marker so they are not lost:

```
# TODO(mazumdera): the eager path never assigns self._freeze_mask, so the regex mask is
# rebuilt every step, and optax's partition.update_fn re-runs the frozen-parameter report
# in optimizers.py on every step as well. Both are compile-time-only on the jitted path.
```

PR #5287, opened 2026-09-18
fix(attention test): give the DSV4 compressed mask test the cp axis rule

gulsumgudukbay:
# Description

- Under load-balanced context parallelism a sequence is split across devices and token order is shuffled, so the mask has to be built from token **positions** rather than row indices.
- `_load_balanced_context_parallel()` decides whether that applies by resolving `activation_q_length` against the *ambient* flax rules -- deliberately, so eval under `logical_axis_rules_for_eval` cannot claim an order the input pipeline never applied.
- `test_generate_attention_mask_compressed_load_balanced_context_parallel` never enters an `axis_rules` context, so the name resolves to `P(None)`, `cp_size` is 1, load balancing is skipped, and `AttentionType.COMPRESSED` returns a plain index-order mask while the test asserts the position-derived one.
- Fix: make that one rule ambient around the op and set `ulysses_context_sharding`, which the cp-size computation reads once the query axis resolves. Same bug and same fix as #5134, which landed just before this test was added.
- Only `activation_q_length` is made ambient, not the whole stub config: its `segment_ids_batch -> context` rule would shard this test's batch-1 segment ids across the 4-way context mesh.
- The test skips below 4 devices and is auto-marked `cpu_only`, so the 4-GPU `gpu-unit` job deselects it -- which is why CI has not been executing it at all.

# Tests

Reproduce the failure on `main` and the fix on this branch with:

    XLA_FLAGS=--xla_force_host_platform_device_count=4 python -m pytest \
      tests/unit/attention_test.py -k compressed_load_balanced_context_parallel

Verified on 4x AMD Instinct MI355X (ROCm 10, jax 0.11.1): fails on `main` (plain causal mask) and passes on this branch.

# Checklist


tests/unit/attention_test.py
```diff
@@ -5688,25 +5689,28 @@ def test_generate_attention_mask_compressed_load_balanced_context_parallel(self)
...
-        config=config,
-        num_query_heads=1,
-        num_kv_heads=1,
-        max_target_length=kv_len,
-        mesh=mesh,
-        attention_kernel="dot_product",
-        attention_type=AttentionType.COMPRESSED,
-        sliding_window_size=sliding_window_size,
-    )
+    # Only the query-length rule: `segment_ids_batch` -> context would shard this batch of 1
```
shralex: Gemini proposes this comment instead:

"Only the query-length rule (activation_q_length -> context) can shard this batch of 1 across the 4-way context mesh, since the segment_ids_batch rule cannot partition a batch size of 1."

shralex (approved):

gobbleturk (approved):

tests/unit/attention_test.py
gulsumgudukbay: thanks, i took this into account but didnt use it directly as it has two axes swapped. 
I adjusted the axes - activation_q_length shards the query length (16 tokens here), which is what the cp gate resolves; the batch of 1 is what segment_ids_batch cant partition. updated comment:

  ```
  # Only `activation_q_length` -> context is made ambient: it shards the query length, which is
  # what the cp gate reads. The config's `segment_ids_batch` -> context rule is left out because
  # it cannot partition this test's batch of 1 across the 4-way context mesh.
```

two other places in this file carry the older version, i also aligned them

PR #5289, opened 2026-09-18
Add free_kv_cache_during_weight_sync to the RL config and forward it to tunix

gutianyu-google:
## What

tunix frees the vLLM KV cache before every trainer->sampler weight sync and re-creates it afterwards. google/tunix#2231 made that optional (`RolloutConfig.rollout_vllm_free_kv_cache_during_weight_sync` -> `VllmConfig.free_kv_cache_during_weight_sync`): with it disabled the pool stays allocated and only its prefix-cache entries are reset.

MaxText cannot reach the option today: `MaxTextVllmRollout` builds its `VllmConfig` field by field and never passes it, and `RLConfig` rejects unknown keys.

This PR:
- adds the `RLConfig` field `free_kv_cache_during_weight_sync` (default `true` = tunix's default, so existing configs are unchanged) and declares it in `configs/post_train/rl.yml`;
- passes it into the tunix `RolloutConfig` through `_kwargs_supported_by`, so the currently pinned tunix (which predates the field) keeps working and the setting takes effect once the pin is bumped;
- forwards it from `MaxTextVllmRollout` into `VllmConfig` only when the rollout config carries the field **and** the installed `VllmConfig` accepts it, so a rollout config that has the attribute cannot break an older `VllmConfig`.

## Effect

Measured on a colocated Qwen3-0.6B GRPO run (64 tpu7x chips, `hbm_utilization_vllm=0.3`, two concurrent 20-step runs, TensorBoard `perf/train/global_step_time`, median of steps 3-20): 60.28 s with the default, 57.92 s with `free_kv_cache_during_weight_sync=false`. The run with the default logs `Deleting kv-cache` and `Reinitializing kv-cache` 21 times each; the run with `false` logs neither. HBM in use per chip goes 28.8 GiB -> 1.1 GiB -> 28.8 GiB around every sync with the default, and stays at 28.8 GiB with `false`.

The weight sync materializes a second copy of the sampler weights on HBM, so `false` is only safe when that copy fits next to the KV pool. The default `hbm_utilization_vllm` (0.72) leaves much less room than the 0.3 used above. Both config descriptions say so.

## Notes for the reviewer

- With the currently pinned tunix the option is a no-op: `_kwargs_supported_by` drops it and logs that the setting is ignored. It takes effect once the tunix pin includes google/tunix#2231.
- Both `update_params` overrides in `maxtext_vllm_rollout.py` delegate to tunix's `VllmSampler.update_params`, so all sync paths honor the option.
- `test_free_kv_cache_option_names_match_the_installed_tunix` is skipped in CI until the pin is bumped; I ran it locally against tunix main (61 passed).

## Tests

- `tests/post_training/unit/train_rl_test.py`
  - the field defaults to tunix's default and accepts an override;
  - `create_rl_components` (tunix pieces faked) passes `rollout_vllm_free_kv_cache_during_weight_sync` to `RolloutConfig` with the configured value when `RolloutConfig` declares it, and drops it cleanly for a `RolloutConfig` that predates it. This test fails if the `train_rl.py` forwarding is removed or the tunix field name is misspelt.
- `tests/post_training/unit/vllm_rollout_unroll_test.py`
  - `MaxTextVllmRollout` forwards both values into `VllmConfig`;
  - omits the kwarg when the rollout config lacks the field, and when `VllmConfig` is a strict class that does not accept it (fails with `TypeError` without the signature check);
  - pins the two tunix field names against the installed tunix (skipped while the pin predates them), since a misspelt name would be dropped without an error. The existing forwarding test now shares a construction helper with the new ones.
  - forwards the option when the `VllmConfig` signature cannot be inspected, matching what `_kwargs_supported_by` does.

Ran both files against the pinned-era tunix (60 passed, 1 skipped) and against tunix main (61 passed).

```
PYTHONPATH=src python -m pytest tests/post_training/unit/vllm_rollout_unroll_test.py tests/post_training/unit/train_rl_test.py
```

# Checklist


A9isha (approved):
Thank you!

One thing to change before merge: the field belongs in the `VLLM` mixin, not `RLCluster`. `RLCluster` is the mixin for tunix's `ClusterConfig`, this one maps to `RolloutConfig/VllmConfig`, and `RLConfig` already has `vllm: VLLM` ("Configuration for rollouts") holding `hbm_utilization_vllm`. the flat YAML key still resolves, the default is preserved, and nested access just becomes config.vllm.free_kv_cache_during_weight_sync.

Optional: `_accepts_kwarg` drops silently, and `integration/vllm/hybrid_cache_utils.py` already has `call_with_supported_kwargs` doing the same signature check with a log line (used by `maxtext_vllm_adapter/adapter.py`). Low stakes since `_kwargs_supported_by` already logs the `RolloutConfig` side — the only silent case is a tunix where RolloutConfig has the field but VllmConfig does not.

gutianyu-google:
Done. Also added a log line for the silent case.

SurbhiJainUSC (approved):

PR #5291, opened 2026-09-19
Add m3 package scaffolding with architecture rules and layout test

bvandermoon:
# Description

First PR of the m3 (Modern MaxText Models) workstream: adds the `maxtext.m3`
package skeleton so subsequent model/config/transform PRs have a place to land.

This is scaffolding only — no behavior change. The package is not imported by
any MaxText entry point, and `src/maxtext/layers/` remains the production path.

Contents:
- `src/maxtext/m3/` with `core/`, `infra/`, `configs/`, `models/`, `train/`.
  Each `__init__.py` documents the modules that will live there.
- `src/maxtext/m3/README.md` with the architecture rules (pure NNX, models own
  their architecture, sharding as annotation, infra applied post-construction)
  plus the invariants m3 models must hold to stay drivable by
  `maxtext_engine` / the RL stack.
- `tests/unit/m3/m3_scaffolding_test.py` pinning the layout.

No dependency changes: flax, qwix, tokamax, orbax and grain are already pinned
at the versions m3 needs.

# Tests

`tests/unit/m3/m3_scaffolding_test.py` verifies `import maxtext.m3` works, that
every declared subpackage imports and ships an `__init__.py`, and that `__all__`
stays in sync with the directories on disk. It is collected by the existing
pytest config, so no workflow changes were needed.

# Checklist


tests/unit/m3/m3_scaffolding_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+import importlib
+import os
+
+from absl.testing import absltest
+from absl.testing import parameterized
+
+import maxtext.m3
+
+_SUBPACKAGES = ("configs", "core", "infra", "models", "train")
```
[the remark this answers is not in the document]
bvandermoon: Done. Switched to disk-derived discovery for the __all__ check. Kept an explicit designed-layout tuple too, since a purely dynamic check can't catch a subpackage being deleted from both disk and __all__.

src/maxtext/m3/README.md
hengtaoguo, line 1: Is `m3` intended as the permanent package name, or is it a temporary staging directory that we plan to rename once the migration is complete?

src/maxtext/m3/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+There is no `layers/` directory: NNX *is* the layer library.
+
+## Architecture rules
+
+1. **Use NNX built-ins directly.** Compose `nnx.Linear`, `nnx.Einsum`, `nnx.Embed`, `nnx.RMSNorm`,
+   `nnx.Dropout` and `nnx.List` as-is — no wrapper layer classes, no Linen, no bridges.
+2. **Models own their architecture.** Each model file defines its own attention, MLP and decoder
+   block. No shared attention class parameterized for every variant, and no model-name branching.
+   Some duplication across model files is cheaper than a 60-parameter shared layer.
```
hengtaoguo, line 47: +1 to Rule 2 — getting rid of the complicated shared `attentions.py` is a huge move but also significantly worthwhile. 

To make sure I understand the intended boundary: models will own their attention layer (Q/K/V projections, RoPE, QK-norm, GQA), while low-level kernel dispatch (Splash/Flash/Ring) lives as shared components in `core/`?  Or is the goal for each model family to be fully self-contained down to the kernel/Pallas call, accepting some kernel-level duplication for maximum readability?

hengtaoguo (approved):
This is an awesome and long-overdue upgrade, thanks for setting up the foundation!

src/maxtext/m3/README.md
bvandermoon: Good question. It probably makes sense to think of a better name for this, but I don't want to block on it currently. Let me know if you have any suggestions. I will try to get some from Gemini as well

src/maxtext/m3/README.md
bvandermoon: Good question. Models will generally own their own architecture. But there will still be some amount of sharing for things like kernels, MoE helper logic, etc. This [prototype](https://github.com/AI-Hypercomputer/maxtext/compare/main...bvandermoon-m3-proto#diff-fdd965889c455fe830ad154a460508c49c60d40892af0260cf5c8c3ad8c43dec) has a high-level idea for how we will reuse kernels. There is also other logic in the prototype if you are interested

cgarciae (approved):

src/maxtext/m3/models/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+"""m3 model garden: one self-contained directory per model family.
+
+Each family owns its architecture end to end:
+
+    models/
+        llama3/
+            llama3.py           # attention, MLP, decoder layer, model
+            llama3_sharding.py  # logical-axis -> mesh-axis rules
+            llama3_8b.yml       # size-specific config overrides
+
```
lydhr: Thanks for the details!
For incoming diffusion models, can we put a schedulers/ folder here, referring to that in [HF Diffusers](https://github.com/huggingface/diffusers/tree/main/src/diffusers/schedulers).

```
models/
    schechulers/
        scheduling_ddim.py
        scheduling_ddpm.py
        ......
   llama3/
   ....
```

cc @hengtaoguo @juansear

PR #5308, opened 2026-09-21
fix: false-positive CI test duration regression alerts

mesakhcienet:
# Description

Resolves false-positive CI "Track Test Duration" regression alerts that have been failing every PR.

## Problem

The "Track Test Duration" job compares each test's duration against the last scheduled main run. `pytest-split` reshuffles tests across 3 workers on every run. The first test in each worker that uses a model family pays an XLA cold-compile cost (~30–130s), while subsequent tests reuse the cache (~4–25s). Because the layout changes every run, a different test absorbs the penalty each time, causing the CI to falsely report massive (300%+) regressions. This has failed on **15 out of 15** PR runs where the check executed.

## Root Cause Analysis

An investigation across 19 CI runs (15 PR + 4 scheduled main) confirmed that only two modules are affected: `moe_test` and `attention_test`. No other module has ever triggered an alert. The third-highest swing across all other modules is +12.8s — below the 15s absolute threshold.

Tests within these modules DO share compilation cache (3–10x speedup when a sibling runs first), contrary to the initial assessment that they generate completely unique computational graphs. The compile cost includes shared sub-computations (reference attention, MoE forward pass, mesh setup).

Previous attempts to fix this by adding `__hash__`/`__eq__` to `HyperParameters` and custom attention kernels were ineffective — `HyperParameters` is partialed out before jitting (never a jit argument), and the kernel classes already had proper pytree registration.

## Fix

1. **Per-test exclusion list for noisy modules:** Adds `EXCLUDED_MODULES_PER_TEST` in `process_test_results.py` containing `moe_test` and `attention_test`. Per-test regressions in these modules are logged as `::warning::` for visibility but do not fail the CI run.
2. **Full enforcement retained for all other modules:** All ~19 other modules (~46% of TPU test time) retain the original per-test regression enforcement with the same thresholds (20% relative AND 15s/30s absolute).
3. **Deterministic benchmark arrays:** Sorts the glob results and dictionary iterations in `process_test_results.py` for stable macro-level dashboard tracking.
4. **No workflow changes:** `track_performance.yml` is unchanged from main.

# Tests

- Added unit tests verifying excluded modules (`moe_test`, `attention_test`) return `failed=False` on regression.
- Added unit test verifying non-excluded modules return `failed=True` on regression.
- Added unit tests for skipped and failed testcase handling.
- Verified all 8 tests pass locally.
- Validated deterministic output array generation.

# Checklist


tests/utils/process_test_results.py
```diff
@@ -179,10 +215,37 @@ def main():
       print(f"Error parsing or processing {xml_file}: {e}")
       has_errors = True
 
+  # Check per-module regression
```
xibinliu, line 218: If a PR adds new tests (or adds parameters to an existing test) in an existing test file, it can trigger a false-positive for the module regression and fail the CI. I think we should remove this check until we have a better solution.

PR #5324, opened 2026-09-21
feat(remat): add te_quantization knob for custom remat policy

kocchop:
# Description

Rename TransformerEngine's quantization-scale checkpoint tag from "quantization" to "te_quantization" so it no longer collides with the `quantization` config field, and expose it as a custom-remat knob (remat/device/offload) alongside the existing tensor names.

# Tests

commands to reproduce.

# Checklist


src/maxtext/layers/nnx_decoders.py
```diff
@@ -1266,7 +1266,7 @@ def minimal_policy(self, with_context=False, with_quantization=False):
     if with_context:
       names.append("context")
     if with_quantization:
-      names.append("quantization")
+      names.append("te_quantization")
```
jberchtold-nvidia, line 1269: It might be safer to do this:
```
names.append("quantization")
names.append("te_quantization")
```
To keep the behavior identical to before for non-TE quantizations for these pre-built remat policies

https://github.com/AI-Hypercomputer/maxtext/blob/c28166ca7de7f3b3cbca34ea3989fe50bb8c8c46/src/maxtext/layers/nnx_decoders.py#L1284
and
https://github.com/AI-Hypercomputer/maxtext/blob/c28166ca7de7f3b3cbca34ea3989fe50bb8c8c46/src/maxtext/layers/nnx_decoders.py#L1293


In practice, I don't know if anybody is using these `minimal_with_quantization` remat policies with non-TE quantization. I was the one who added `minimal_with_quantization` and `minimal_with_context_and_quantization` originally. At least on our internal fork, I can't recall who upstreamed it. So it's definitely possible that nobody is really using `minimal_with_quantization` and are instead using a custom remat policy, which in hindsight is what I should've done had I known about custom remat policies at the time.

src/maxtext/layers/quantizations.py
```diff
@@ -1370,7 +1370,7 @@ def generate_quantizer_set(
         return super().generate_quantizer_set(  # pytype: disable=wrong-keyword-args
             postfix=postfix,
             variable_collection=OVERWRITE_WITH_GRADIENT,
-            quantization_checkpoint_name="quantization",
+            quantization_checkpoint_name="te_quantization",
```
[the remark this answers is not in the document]
jberchtold-nvidia: @kocchop if you don't mind, could you make this suggestion as well? This was not introduced by this PR and it is currently not causing any issues since we never override these in the MaxText integration, but it is something I missed and it could lead to some confusion in the future if we, for example, split grouped quantization and Dense quantization into two separate checkpoint names

src/maxtext/configs/base.yml
```diff
@@ -435,6 +435,7 @@ mla_q: 'remat'
 mla_kv: 'remat'
 attention_out: 'remat'
 engram: 'remat'
+te_quantization: 'remat' # TransformerEngine quantization scale tensors
```
jberchtold-nvidia, line 438: What about `te_quantization_remat_location`? I feel `te_quantization` is similar to `quantization` and `te_gmm_quantization` and from the name alone, `te_quantization` isn't immediately clear that it's a remat policy rather than a quantization recipe configuration

PR #5329, opened 2026-09-22
fix(training_engine): prevent 397B every-step checkpointing + Raiden weight-sync OOM

YixuanWang-99:
## Summary
Fixes CPU RAM OOM (`exitCode: 137`) during 397B Orbax/Pathways checkpointing and weight synchronization.

### Changes
1. **Pathways Persistence `gs://` Guard (`src/maxtext/training_engine/checkpointing.py`)**:
   - Validate in `CheckpointManager.__init__` (before `_maybe_register_pathways_persistence()`) that `checkpoint_dir` starts with `gs://` whenever `ENABLE_PATHWAYS_PERSISTENCE=1`, so every `CheckpointManager` instantiation validates the path even after `_PATHWAYS_PERSISTENCE_REGISTERED` is set.
2. **Release Staged Sync Buffers Before Checkpoint (`src/maxtext/training_engine/maxtext_engine.py`)**:
   - In `MaxTextTrainingEngine.prepare_weight_sync()`, call `self._raiden_sync.release_buffers()` (and `gc.collect()`) immediately after `checksums()` completes and before `save_checkpoint()`, reclaiming ~794 GB of temporary converted weight buffers prior to checkpoint serialization (while always saving full model + optimizer state).
3. **Unit Tests**:
   - `tests/unit/checkpointing_test.py`
   - `tests/post_training/unit/prepare_weight_sync_test.py`


src/maxtext/training_engine/maxtext_engine.py
igorts-git, line 2220: I don't know why we have this condition here and whether we ever take it, but your newly added `_purge_raiden_buffers` inside `release_weight_sync` is likely going to break it.

src/maxtext/training_engine/maxtext_engine.py
```diff
@@ -2349,14 +2229,18 @@ def prepare_weight_sync(
         )
         return self._staged_metadata
 
-      # 1. Drain all in-flight TPU computations to ensure weights are fully updated
+      # 1. Drain in-flight checkpoint persistence and TPU computations, and drop
+      # any leftover converted tree from the previous cycle BEFORE convert().
+      if self._checkpoint_manager is not None:
+        self._checkpoint_manager.wait_until_finished()
```
igorts-git: Are you sure this is still needed? I thought that your change to add `checkpoint_storage_device_host_concurrent_gb` significantly reduces the amount of RAM needed for checkpointing.

src/maxtext/training_engine/checkpointing.py
```diff
@@ -40,22 +40,63 @@ class CheckpointState:
...
+# means Orbax fell back to the controller-side handler, which stages every shard through
+# proxy-pod host RAM and OOMs at 397B scale.
+_PATHWAYS_PERSISTENCE_HANDLERS = ("CloudPathwaysArrayHandler", "PathwaysPersistenceArrayHandler")
+
+
+class PathwaysCheckpointingUnavailableError(RuntimeError):
+  """Pathways checkpointing was explicitly requested but could not be registered."""
+
+
+def _maybe_register_pathways_persistence(impl_name: str = _PERSISTENCE) -> None:
```
sadikneipp, line 61: I believe this section is redundant with the init we do in pathwaysutils: https://github.com/AI-Hypercomputer/pathways-utils/blob/main/pathwaysutils/_initialize.py#L95-L104

PR #5331, opened 2026-09-22
Add scheduled-only label to allow running scheduled tests on PRs if required

chiajunglien:
# Description

Tests marked `scheduled_only` (and the whole TPU7X suite) only run in the periodic scheduled CI, so regressions in them are only caught hours after a PR lands, and a PR author who knowingly touches that area has no way to validate it before merging.

This PR adds a `scheduled-only` PR label that re-runs the CI pipeline as if it were a scheduled run:

- `ci_pipeline.yml` now listens to the `pull_request: [labeled]` event. Every entry job (`analyze_code_changes`, `code_quality_check`, `docs_build_check`) is gated on the added label being `scheduled-only`, so adding any other label (`pull ready`, `gemini-review`, etc.) leaves all jobs skipped and costs no CI capacity.
- `is_scheduled_run` is set to true when the label is present, so the `not scheduled_only` pytest filter is dropped, `cpu-unit` installs torch-cpu, and coverage is uploaded under the `scheduled` flag — identical to a cron run.
- `analyze_code_changes.sh` gains a `FORCE_ALL_TESTS` escape hatch so a labeled PR runs every suite instead of only the ones matching the changed files.
- TPU7X tests, previously hard-skipped on PRs, run when the label is present.

`notify_failure` / `investigate_failure` deliberately stay `schedule`-only, so a failing labeled PR does not open or comment on tracking issues.

The label is opt-in: remove and re-add it to trigger another run.

# Tests

https://github.com/AI-Hypercomputer/maxtext/actions/runs/35696264188

# Checklist


.github/workflows/ci_pipeline.yml
```diff
@@ -70,17 +79,28 @@ jobs:
...
           GITHUB_BASE_REF: ${{ github.base_ref }}
+          # A `scheduled-only` PR mirrors a scheduled run, so run every suite
+          # regardless of which files the PR touched.
+          FORCE_ALL_TESTS: ${{ contains(github.event.pull_request.labels.*.name, 'scheduled-only') }}
         run: |
           bash .github/scripts/analyze_code_changes.sh
 
   code_quality_check:
     name: Code Quality Check
+    if: >-
```
SurbhiJainUSC: This job should run irrespective of scheduled run or PR run.  Same comment for docs_build_check and analyze_code_changes job. Please remove `if` block.

chiajunglien: Thanks for the comment! 

Just to clarify, the current `if` condition actually **already runs** on both scheduled runs and regular PR runs:

* **On regular PR runs / new commits (`synchronize`, `opened`)**: `github.event.action != 'labeled'` evaluates to `true`, so the jobs run normally.
* **On scheduled runs (`schedule`)**: `github.event_name != 'pull_request'` evaluates to `true`, so the jobs run as well.

The reason it's needed is that this PR also adds `labeled` to the `pull_request` trigger, and GitHub doesn't support filtering by label name in `on:` — so the event fires for **every** label, not just `scheduled-only`. The only scenario this condition skips is when someone adds an **unrelated PR label** (like `gemini-review`).

If we remove the `if` block entirely, adding any label to a PR would trigger these three checks **and everything downstream of them** — the package build plus whichever TPU/GPU/CPU/Pathways suites that PR's changes require.

Does this make sense? If you prefer cleaner job definitions without duplicating this logic, I can move the check onto `build_and_upload_maxtext_package` instead — then these three jobs always run, and only the expensive test matrix is gated. Let me know what you think!

SurbhiJainUSC (approved):

PR #5334, opened 2026-09-22
m3-modernization of qwen3-dense

hengtaoguo:
# Description

Trying to modernizing qwen3 dense family in an isolate format (`use_m3_model=True`).

# Tests

20-step training:

- qwen3_06b_old: [xprof](https://xprof.corp.google.com/hlo_op_profile/hengtaoguo-3971942883812682315)
- qwen3_06b_m3: [xprof](https://xprof.corp.google.com/hlo_op_profile/hengtaoguo-10266671168751941728)

# Checklist


lydhr (approved):

src/maxtext/m3/models/__init__.py
```diff
@@ -17,12 +17,25 @@
...
-        llama3/
-            llama3.py           # attention, MLP, decoder layer, model
-            llama3_sharding.py  # logical-axis -> mesh-axis rules
-            llama3_8b.yml       # size-specific config overrides
-
-This module will also hold the single dispatch point (`create_model`) and the
-model registry. There is no shared `Decoder`, no shared attention class, and no
-model-name branching
+        qwen3/
+            modeling_qwen3.py  # attention, MLP, decoder layer, model
```
lydhr, line 21: - Are we putting the other qwen3 models (4b, 30b) in the same qwen3/ folder?
  - If yes, can we rename this to qwen3_06B.py?
- Are the other qwen3 models reusing the attention, MLP, decoder?
  - If yes, how would them reuse the basic blocks, such as a qwen3_attention.py?
- The original comment listed llama3_8b.yml here. Are we keeping a copy and route the to this yml?

The above questions can be demonstrated in a follow up PR, adding a another qwen3 model, e.g. qwen3_6b.

lydhr: Qoute the M3 design doc, which seems having everything in qwen3.py and customize only via config.yml.
```
llama3
            llama3.py        # LLaMA: config, attention, MLP, decoder, model
            llama3_8b.yml    # Model-specific overrides
            llama3_70b.yml   # Model-specific overrides
            llama3_405b.yml  # Model-specific overrides
            llama3_sharding.py # LlamaSharding(Sharding) rules
```

hengtaoguo: Thanks for raising the concerns!

1. Models in the same family like qwen3-X should be able to share one modeling file here like `modeling_qwen.py`. During the initialization, each block will have access to the config, whose hyper-parameters will configure the number of layers and attention heads etc. This is intentionally similar to huggingface style like [modeling_qwen3.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3/modeling_qwen3.py)
2. Yes, ideally this current file should support any qwen3-dense model. We can validate and tweak that with a follow up PR adding a qwen3-4B registry.
3. This is the part I'm not super sure: whether the model configs lives here or in another centralized config folder. We should have a follow up study on that, based on the design we can keep them in this folder.

hengtaoguo: After double checking, we should consider write the MOE models in a separate folder like [qwen3_moe](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_moe/modeling_qwen3_moe.py). While the dense models should still live inside qwen3. We could always check HF grouping pattern for reference.

src/maxtext/m3/README.md
```diff
@@ -21,9 +21,9 @@ defines its own attention, MLP and decoder blocks instead of routing through a s
 class and a central `Decoder`, so a model can be read top to bottom in one place. Scalability
 concerns are applied as post-construction transforms rather than branched on inside model code.
 
-> **Status: scaffolding.** This package is not yet wired into any MaxText entry point and nothing
-> here is importable by the legacy stack. Legacy `src/maxtext/layers/` remains the production path
-> until the migration milestones complete.
+> **Status: in progress.** Model construction can opt in with `use_m3_model=true`. The m3
+> registry currently supports `qwen3-0.6b`, while unregistered models are rejected. The flag defaults
+> to false; legacy `src/maxtext/layers/` remains the production path until migration milestones complete.
```
bvandermoon: Can this comment be re-worded? `qwen3-0.6b` has been tested for correctness here, but it does not provide performance features yet

src/maxtext/m3/models/qwen3/modeling_qwen3.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+class Qwen3Model(nnx.Module):
+  """Top-level Qwen3 autoregressive language model in m3 format."""
+
+  def __init__(
+      self,
+      config: Config,
+      mesh: Optional[Mesh] = None,
+      model_mode: str = MODEL_MODE_TRAIN,
```
bvandermoon: Is `model_mode` actually used anywhere by m3 in this PR? If not, can we just remove it for now? We might need to add it or something similar back soon, but I would like to hold off until we see that it's really needed to avoid complications. I don't see any branches based on `model_mode` currently, but let me know if I am missing it somewhere

src/maxtext/m3/models/qwen3/modeling_qwen3.py
```diff
@@ -0,0 +1,518 @@
...
+      inputs: jax.Array,
+      decoder_segment_ids: Optional[jax.Array] = None,
+      decoder_positions: Optional[jax.Array] = None,
+      deterministic: bool = False,
+      model_mode: str = MODEL_MODE_TRAIN,
+      **kwargs,
+  ) -> jax.Array:
+    """Applies pre-norm attention and pre-norm MLP blocks with residual connections."""
+    if isinstance(inputs, tuple):
+      inputs = inputs[0]
```
bvandermoon: Is this actually needed?

I see this message from Antigravity:

```
This is the legacy `nn.scan` carry convention leaking into the model body. `infra/scan` is a
post-construction transform, so if a carry protocol is needed it belongs in `infra/`, not in each
model's decoder layer. Can we drop it?
```

src/maxtext/m3/models/qwen3/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+)
+
+__all__ = [
+    "Qwen3Attention",
+    "Qwen3Decoder",
+    "Qwen3DecoderLayer",
+    "Qwen3MLP",
+    "Qwen3Model",
+    "create_qwen3_model",
+]
```
bvandermoon: Is it possible to delete this? It shouldn't be needed. You might have taken this example since I added it to RoPE earlier, but I am about to submit a PR to delete it from there too

src/maxtext/m3/models/__init__.py
```diff
@@ -17,12 +17,28 @@
...
+model-name branching.
 """
+
+from collections.abc import Callable
+from maxtext.m3.models.qwen3 import create_qwen3_model
+
+
+MODEL_REGISTRY: dict[str, Callable] = {
+    "qwen3-0.6b": create_qwen3_model,
+}
```
cgarciae, line 34: alternatively we could create a `register_model` function and use it as a decorator e.g.

```python
@register_model("qwen3-0.6b")
def create_qwen3_model(...):
```

src/maxtext/m3/README.md
hengtaoguo: Just rephrased it, we are still in the scaffolding space and introduced new flag `use_m3_model`. qwen3-0.6b only had correctness validation, legacy path should still serve as the source of truth right now.

src/maxtext/m3/models/qwen3/modeling_qwen3.py
hengtaoguo: Good call! This `model_mode` is never used anywhere, I've removed it from the code and rerun correctness tests (still passes).

src/maxtext/m3/models/qwen3/modeling_qwen3.py
hengtaoguo: Removed, thanks for the comments.

src/maxtext/m3/models/qwen3/__init__.py
hengtaoguo: This route was intended to enable 

```
from maxtext.m3.models.qwen3 import create_qwen3_model
```

I've removed this extra layer, so the registry will call this now:

```
from maxtext.m3.models.qwen3.modeling_qwen3 import create_qwen3_model
```

src/maxtext/m3/models/__init__.py
hengtaoguo: Good call, I will keep this in mind!

Right now in the scaffolding stage with only one example model, using the decorator may impact readability a little. I plan to add a qwen3_moe soon, and in that follow-up PR I will try the decorator approach.

